PR #120, opened 2014-06-30
Update SignoffDAO.inc.php

besamelsosu:
parameter ordering fix, causes problems otherwise.


asmecher:
SignoffDAO has been deprecated and removed from master.

PR #204, opened 2014-10-28
*8980* fix generation of ul.interests in ojs register page only when it ...

eirikhanssen:
...will have contents  ##eirikhanssen/validation-2_4_5##


asmecher:
Reverting -- see http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8980#c17

PR #208, opened 2014-10-30
*8994* Avoid triggering public events twice

beghelli:


asmecher:
Calling this a WONTFIX, I think. The current JS framework is deprecated in favour of Vue & co, and I don't think this quirk is causing any evident problems.

Issue #245, opened 2014-12-05
Large log messages in the FileLoader can exceed PHP's memory_limit


ctgraham:
When processing large files with many execution log entries (large external log files in OJS 2.4 UsageStatsLoader is a good example), the accumulated execution log can exceed PHP's memory_limit, killing the process unceremoniously and losing the execution log email notification.

asmecher:
Back to you, Bruno.

beghelli:
Porting completed.

beghelli:
Porting completed.

PR #250, opened 2014-12-09
Monitor memory usage with each file processed by FileLoader::executeActions()

ctgraham:
Abort early if usage might exceed the PHP memory limit.  Resolves pkp/pkp-lib#245.

### DIFF
--- a/classes/task/FileLoader.inc.php
+++ b/classes/task/FileLoader.inc.php
@@ -20,6 +20,7 @@ define('FILE_LOADER_PATH_STAGING', 'stage');
 define('FILE_LOADER_PATH_PROCESSING', 'processing');
 define('FILE_LOADER_PATH_REJECT', 'reject');
 define('FILE_LOADER_PATH_ARCHIVE', 'archive');
+define('FILE_LOADER_MEMORY_TOLERANCE', 0.90);
 
 class FileLoader extends ScheduledTask {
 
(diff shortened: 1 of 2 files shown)


classes/task/FileLoader.inc.php
```diff
@@ -126,7 +127,39 @@ function executeActions() {
 		if (!$this->checkFolderStructure()) return false;
 
 		$foundErrors = false;
+		// get setting value
+		$memoryLimit = ini_get('memory_limit');
+		// multiply for shorthand notation
+		switch (strtolower($memoryLimit[strlen($memoryLimit)-1])) {
```
asmecher: Doesn't this try to perform math on the number without first stripping off the suffix?

classes/task/FileLoader.inc.php
```diff
@@ -126,7 +127,39 @@ function executeActions() {
...
+			//  3) A string form of this log which is preprocessed by the mailer (e.g. CR -> CRLF)
+			//  4) A string or array form of this log as it is sent in the mailer object
+			// Since this is a serial process operating on potentially large files, keep track of the log size
+			// and this process's own memory increase to preempt out-of-memory errors
+			if (memory_get_usage() - $baseMemoryUsage > $largestMemoryIncrease) {
+				$largestMemoryIncrease = memory_get_usage() - $baseMemoryUsage;
+			}
+			$baseMemoryUsage = memory_get_usage();
+			// if the largest known memory increase were added in again, and if the logsize were quadrupled as described above, and if this would exceed our memory tolerance, break out of processing early.
+			if ($memoryLimit > 0 && memory_get_usage() + $logSize * 4 + $largestMemoryIncrease > $memoryLimit * FILE_LOADER_MEMORY_TOLERANCE) {
```
asmecher: Why *4?

classes/task/FileLoader.inc.php
ctgraham: I don't think so.  We're just getting the units character for the switch statement here.

classes/task/FileLoader.inc.php
ctgraham: See the comment above for the (at least) four memory representations of this object.

classes/task/FileLoader.inc.php
asmecher: I meant the next bits where we multiply. The suffix is never stripped, I think.

ctgraham: True.  It works because PHP interprets "[0-9]+[a-z]" in numeric context as just the number.  Could have been clearer, but I think we're dropping this whole approach in favor of https://github.com/pkp/pkp-lib/pull/354 .

ctgraham:
Superseded by https://github.com/pkp/pkp-lib/pull/354

PR #287, opened 2015-01-14
*8849* Permit reader self-registration

asmecher:
See http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8849


bozana:
Alec, I've tested and reviewed it ;-) and everything looks perfect for me :-)
There is maybe just one more thing that could be considered: upgrade.-- In my installation I've already created a role with the permission level 'Reader' and have also assigned it a stage. Now the stage assignment isn't possible any more, but the old entries are still in the database, I think in the table user_group_stage. These should be removed at upgrade, right?

asmecher:
Excellent! I've added an upgrade script to remove the roles and will merge. Thanks for the review.

Issue #276, opened 2015-01-10
SSL redirect ignores base_url[<journal_name>]


jprk:
Hello,

I have a OJS 2.4.5 installation with maybe non-standard single journal setup (that is, the base url http://ojs.nnw.cz goes directly to the journal homepage, not to the OJS/index homepage). With a bit of URL rewriting and `restful_urls=On` in the configuration file, it works flawlessly until I try to force SSL for logins by setting `force_login_ssl=On`. For the time being I am ignoring the links to the OJS/index parts as _My Journals_, this can be handled by another set of URL rewriting rules.

Now, although I have set up `base_url[nnw]=http://ojs.nnw.cz` in the configuration file, and all the HTTP links are generated correctly (e.g. http://ojs.nnw.cz/about and so on), the correct login link http://ojs.nnw.cz/login gets redirected to https://ojs.nnw.cz/nnw/login instead of assumed https://ojs.nnw.cz/login.

I am not very familiar with the design of OJS, but my guess is that while handling the `force_login_ssl` configuration option in [PKPLoginHandler.index()] (https://github.com/pkp/pkp-lib/blob/abd8adfe791498379c7c7aa46d6475956a1566eb/pages/login/PKPLoginHandler.inc.php#L37), the internal request value shall be once again remapped using `base_url[<journal_name>]`.

It is not something crucial, it can be handled by yet another set of rewrite rules that redirect the incorrect redirect once again, but it is not very convenient ... 

-- jprk

ctgraham:
I tested this in dev-2_4 with rewrites of:
```
RewriteEngine On
RewriteCond %{REQUEST_URI} !/index.php/
RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} !-f
RewriteRule ^/index/(.*)$ /index.php/index/$1 [L]

RewriteCond %{REQUEST_URI} !/index.php/
RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} !-f
RewriteRule ^(.*)$ /index.php/testjournal$1 [L]
```

and a config.inc.php of:
```
base_url = "http://my.test.journal/index"
base_url[testjournal] = "http://my.test.journal/"
```
and either `force_ssl` or `force_login_ssl` set to `On`.

My experience verifies the issue.  I haven't gotten down to the cause or fix yet.

ctgraham:
[...]

PR #314, opened 2015-02-04
pkp/pkp-lib#276: Defer processing of REQUEST_URI and QUERY_STRING in a protocol redirect until the target page.

ctgraham:
A protocol redirect should simply pass the request unmodified from HTTP to HTTPS (or vice-versa).  Processing the request path and query string (and appropriate hook calls) can happen at the target page.


asmecher:
Thanks, looks good -- we'll have to wait and see if we've missed any edge cases by handling things this way, but I think it's a better solution than what we were doing before.

PR #320, opened 2015-02-06
*8759* Submissions list improvements

beghelli:


classes/controllers/grid/feature/GeneralPagingFeature.inc.php
```diff
@@ -0,0 +1,169 @@
...
+ * @ingroup controllers_grid_feature
+ *
+ * @brief Base class that implements common functionality for
+ * paging features.
+ *
+ */
+
+import('lib.pkp.classes.controllers.grid.feature.GridFeature');
+
+class GeneralPagingFeature extends GridFeature{
```
asmecher: (missing space)

classes/controllers/grid/feature/GeneralPagingFeature.inc.php
```diff
@@ -0,0 +1,169 @@
+<?php
+
+/**
+ * @file classes/controllers/grid/feature/GeneralPagingFeature.inc.php
+ *
+ * Copyright (c) 2014 Simon Fraser University Library
```
asmecher: All new copyrights should be 2015.

classes/controllers/grid/feature/GeneralPagingFeature.inc.php
```diff
@@ -0,0 +1,169 @@
...
+
+import('lib.pkp.classes.controllers.grid.feature.GridFeature');
+
+class GeneralPagingFeature extends GridFeature{
+
+	/** @var ItemIterator */
+	private $_itemIterator;
+
+	/**
+	 * Constructor.
```
asmecher: Parameter description

classes/controllers/grid/feature/InfiniteScrollingFeature.inc.php
```diff
@@ -0,0 +1,132 @@
...
+	function InfiniteScrollingFeature() {
+		parent::GeneralPagingFeature('infiniteScrolling');
+	}
+
+
+	//
+	// Extended methods from GridFeature.
+	//
+	/**
+	 * @see GridFeature::getJSClass()
```
asmecher: Use @copydoc rather than @see unless you're re-specifying the @param / @return stuff

classes/submission/SubmissionDAO.inc.php
```diff
@@ -405,9 +414,13 @@ function getUnpublishedByUserId($userId, $contextId = null, $rangeInfo = null) {
 				' . $this->_getFetchColumns() . '
 				FROM	submissions s
 				LEFT JOIN published_submissions ps ON (s.submission_id = ps.submission_id)
+				LEFT JOIN submission_settings ss ON (s.submission_id = ss.submission_id)
```
asmecher: Shouldn't this be joined only if a title is specified?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -452,6 +465,114 @@ function getByStatus($status, $stageUserId = null, $reviewUserId, $contextId = n
...
+	 * @param $title string|null optional Filter by title.
+	 * @param $author string|null optional Filter by author.
+	 * @param $stageId int|null optional Filter by stage id.
+	 * @param $rangeInfo DBResultRange optional
+	 * @return DAOResultFactory
+	 */
+	function getAssignedToUser($userId, $contextId = null, $title = null, $author = null, $stageId = null, $rangeInfo = null) {
+		$params = $this->_getFetchParameters();
+		$userId = (int) $userId;
+		array_push($params, $userId, STATUS_DECLINED, $userId, $userId, ROLE_ID_AUTHOR, $userId);
```
asmecher: Need int casts whererever possible.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -452,6 +465,114 @@ function getByStatus($status, $stageUserId = null, $reviewUserId, $contextId = n
...
+				LEFT JOIN signoffs so ON (sf.file_id = so.assoc_id) 
+				LEFT JOIN user_groups g ON (g.user_group_id = so.user_group_id)
+				LEFT JOIN review_assignments ra ON (s.submission_id = ra.submission_id)
+				' . ($title?' LEFT JOIN submission_settings ss ON (s.submission_id = ss.submission_id)':'') . '
+				' . ($author?' LEFT JOIN authors au ON (s.submission_id = au.submission_id)':'')
+				. $this->_getFetchJoins() .
+			' WHERE s.date_submitted IS NOT NULL AND ps.date_published IS NULL AND s.user_id <> ? AND s.status <> ?
+				AND (sa.user_id = ? OR (so.user_id = ? AND g.role_id <> ? ) 
+				OR ra.reviewer_id = ?)'
+				. (($contextId)?' AND s.context_id = ?':'')
```
asmecher: (extra brackets)

classes/submission/SubmissionDAO.inc.php
```diff
@@ -452,6 +465,114 @@ function getByStatus($status, $stageUserId = null, $reviewUserId, $contextId = n
...
+				' . $this->_getFetchColumns() . '
+			FROM submissions s
+				LEFT JOIN published_submissions ps ON (s.submission_id = ps.submission_id)
+				LEFT JOIN stage_assignments sa ON (s.submission_id = sa.submission_id)
+				LEFT JOIN submission_files sf ON (s.submission_id = sf.submission_id)
+				LEFT JOIN signoffs so ON (sf.file_id = so.assoc_id) 
+				LEFT JOIN user_groups g ON (g.user_group_id = so.user_group_id)
+				LEFT JOIN review_assignments ra ON (s.submission_id = ra.submission_id)
+				' . ($title?' LEFT JOIN submission_settings ss ON (s.submission_id = ss.submission_id)':'') . '
+				' . ($author?' LEFT JOIN authors au ON (s.submission_id = au.submission_id)':'')
```
asmecher: What if there are two authors for the same submission whose names both match? That'll cause it to be listed twice, I think.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -452,6 +465,114 @@ function getByStatus($status, $stageUserId = null, $reviewUserId, $contextId = n
...
+	 * @param $rangeInfo DBResultRange optional
+	 * @return DAOResultFactory
+	 */
+	function getAssignedToUser($userId, $contextId = null, $title = null, $author = null, $stageId = null, $rangeInfo = null) {
+		$params = $this->_getFetchParameters();
+		$userId = (int) $userId;
+		array_push($params, $userId, STATUS_DECLINED, $userId, $userId, ROLE_ID_AUTHOR, $userId);
+		if ($contextId) $params[] = $contextId;
+
+		if ($title) {
```
asmecher: What if the user is browsing in a locale that doesn't have a specific translation? That'll never be found. OTOH I think it's better to search titles in all languages, and if any match, show the submission.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -452,6 +465,114 @@ function getByStatus($status, $stageUserId = null, $reviewUserId, $contextId = n
...
+		}
+		if ($author) array_push($params, $authorQuery = '%' . $author . '%', $authorQuery, $authorQuery);
+		if ($stageId) $params[] = $stageId;
+
+		$result = $this->retrieveRange($sql =
+			'SELECT s.*, ps.date_published,
+				' . $this->_getFetchColumns() . '
+			FROM submissions s
+				LEFT JOIN published_submissions ps ON (s.submission_id = ps.submission_id)
+				LEFT JOIN submission_files sf ON (s.submission_id = sf.submission_id)
```
asmecher: Won't this join cause submissions with multiple files to be listed multiple times?

controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php
```diff
@@ -64,20 +64,52 @@ function getCellState($row, $column) {
...
-			$submission = $row->getData();
+		$submission = $row->getData();
+		if ($column->getId() == 'editor') {
+			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
+			$editorAssignments = $stageAssignmentDao->getEditorsAssignedToStage($submission->getId(), $submission->getStageId());
+			$assignment = current($editorAssignments);
+			if (!$assignment) return array();
+			$user = $request->getUser();
+			$userDao = DAORegistry::getDAO('UserDAO'); /* @var $userDao UserDAO */
+			$editor = $userDao->getById($assignment->getUserId());
```
asmecher: OTOH, I think the EditAssignment object contains enough identifying information that the User object fetch might be unnecessary.

controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php
```diff
@@ -104,9 +136,25 @@ function getTemplateVarsFromRowColumn($row, $column) {
...
+					$primaryAuthor = $submission->getPrimaryAuthor();
+					$authors = $submission->getAuthors();
+					if (!isset($primaryAuthor)) {
+						if (sizeof($authors) > 0) {
+							$primaryAuthor = $authors[0];
+						}
+					}
+
+				$authorsInTitle = $primaryAuthor->getLastName();
+				if (count($authors) > 1) $authorsInTitle = $authorsInTitle . ' et al.';
```
asmecher: "Et al" needs to be moved into a locale key.

asmecher: (...plus IMO we should do this in a standard way throughout the system, like getAuthorString does. If we need to cases -- a long and a short author string, where the short one uses "et al" -- then that should be implemented in the same place as getAuthorString.)

controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php
```diff
@@ -104,9 +136,25 @@ function getTemplateVarsFromRowColumn($row, $column) {
...
+				$title = $authorsInTitle . '; ' . $title;
+				return array('label' => $title);
 			case 'context':
```
asmecher: I think the dashboard is always going to be inside a context now, no?

controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php
```diff
@@ -243,13 +251,10 @@ static function getPageAndOperationByUserRoles($request, $submission, $userId =
 	 * @param $submission Submission
 	 * @return LinkAction
 	 */
-	function _getCellLinkAction($request, $page, $operation, $submission) {
+	function _getCellLinkAction($request, $page, $operation, $submission, $title) {
```
asmecher: $title needs documenting

controllers/grid/submissions/SubmissionsListGridHandler.inc.php
```diff
@@ -96,11 +92,13 @@ function initialize($request) {
 
 		$this->addColumn(
 			new GridColumn(
-				'author',
-				'submission.authors',
-				null,
+				'id',
 				null,
-				$cellProvider
+				'ID',
```
asmecher: Needs to be localized

classes/submission/SubmissionDAO.inc.php
beghelli: I think that this will be avoided by the group by

classes/submission/SubmissionDAO.inc.php
beghelli: Group by will avoid that, I think (same below)

classes/submission/SubmissionDAO.inc.php
asmecher: Ah -- but this will cause problems with PostgreSQL at least; PostgreSQL requires all SELECTed columns to be included in GROUP BY until version 9.1 or so, when it got a little smarter. Even with 9.1 I think you'd need to add ps.article_id and maybe others to the GROUP BY.

Instead, I'd suggest using JOINs to do the disambiguation, and removing the GROUP BY clause. I can scare up an example of this if you're not sure what I mean.

beghelli: we did grouped by all selected columns here right: https://github.com/pkp/ojs/blob/master/classes/article/ArticleDAO.inc.php#L467 ? I can try to think harder to disambiguate or do the same that we already did...

asmecher: Hmm, that will cause problems with PostgreSQL <9.1.something, as prior to that PostgreSQL required us to actually specify every column. Since then they added smarts to know that grouping by a PRIMARY KEY column was effectively the same thing. I didn't know we were already doing this, but rather than fixing it now, let's keep it consistent.

Just FYI, an alternate approach that would work without the PostgreSQL requirement is to do two joins:
...LEFT JOIN mytable t ON (...)
...LEFT JOIN mytable t2 ON (... AND t2.my_id > t.my_id)
WHERE
t2.my_id IS NULL

The first join brings in the actual content you're looking for, and the second makes sure it's a unique result.

asmecher:
Woo! Happy to see this, @beghelli.

PR #349, opened 2015-02-18
Update common.xml

hauschke:
fixed typo "Benachrichtung*" to "Benachrichtigung*"


mtub:
Thanks, Christian. This has been unspotted for many years. Will fix it in other applications and branches, too.

PR #354, opened 2015-02-21
pkp/pkp-lib#245 Use files to store scheduled task execution logs

beghelli:


classes/scheduledTask/ScheduledTaskHelper.inc.php
```diff
@@ -165,6 +200,7 @@ function notifyExecutionResult($id, $name, $result, $message = '') {
 	function _sendEmail($message, $subject) {
 		$mail = $this->getMail();
 		$mail->addRecipient($this->_contactEmail, $this->_contactName);
+		$mail->setFrom('bruno.beghelli@gmail.com', 'Bruno Beghelli');
```
beghelli: oups, this doesn't belongs here. :) I needed this to correctly send emails using smtp, will remove

classes/scheduledTask/ScheduledTask.inc.php
```diff
@@ -81,17 +96,26 @@ function getName() {
 	 * SCHEDULED_TASK_MESSAGE_TYPE... constants.
 	 */
 	function addExecutionLogEntry($message, $type = null) {
-		$log = $this->_executionLog;
+		$logFile = $this->_executionLogFile;
+		$log = null;
```
asmecher: This is unnecessary, I think.

classes/scheduledTask/ScheduledTask.inc.php
```diff
@@ -81,17 +96,26 @@ function getName() {
...
 		}
 
-		$this->_executionLog = $log;
+		$fp = fopen($logFile, 'ab');
+		if (flock($fp, LOCK_EX)) {
+			fwrite($fp, $log . PHP_EOL);
+			flock($fp, LOCK_UN);
+		} else {
+			// Couldn't lock the file.
+			assert(false);
```
asmecher: I prefer to use fatalError in cases like this. If assertions are disabled on the server, it'll break without warning anyone if the lock can't be made. It's open to debate, though.

PR #367, opened 2015-02-27
Update ru_RU locale after #245

ppv1979:


mtub:
merged. thanks, Pavel!

Issue #377, opened 2015-03-03
Improve theme capabilities


asmecher:
* Add a class to body element to permit targeting of specific pages in themes
* Permit easy overrides of built-in templates with theme-supplied templates
* Improve contract for front-end templates (e.g. can we auto-document smarty templates?)

asmecher:
Pulls:
https://github.com/pkp/omp/pull/105
https://github.com/pkp/pkp-lib/pull/378

asmecher:
@jnugent, mind reviewing this?

asmecher:
A third PR (this one for OJS master): https://github.com/pkp/ojs/pull/442

jnugent:
Looks great, Alec.  I had a few comments, but no quibbles with the code.  Merge away.

asmecher:
Thanks, Jason! Merged.

PR #378, opened 2015-03-04
pkp/pkp-lib#377 Improve theme capabilities

asmecher:


classes/plugins/ThemePlugin.inc.php
```diff
@@ -33,31 +33,23 @@ function register($category, $path) {
...
+			HookRegistry::register('PageHandler::displayCss', array($this, '_displayCssCallback'));
+
+			// Add the stylesheet.
 			$dispatcher = $request->getDispatcher();
 			$templateManager->addStyleSheet($dispatcher->url($request, ROUTE_COMPONENT, null, 'page.PageHandler', 'css', null, array('name' => $this->getName())), STYLE_SEQUENCE_LATE);
+
+			// If this theme uses templates, ensure they're given priority.
+			array_unshift(
+				$templateManager->template_dir,
+				$path = Core::getBaseDir() . DIRECTORY_SEPARATOR . $this->getPluginPath() . '/templates'
```
jnugent: Moot point because / is fine, but why use DIRECTORY _SEPARATOR in one instance and '/templates' in the other?

classes/template/PKPTemplateManager.inc.php
```diff
@@ -90,44 +90,51 @@ function PKPTemplateManager($request) {
...
+			'useMinifiedJavaScript' => Config::getVar('general', 'enable_minified'),
+			'toggleHelpOnText' => __('help.toggleInlineHelpOn'),
+			'toggleHelpOffText' => __('help.toggleInlineHelpOff'),
+			'currentContext' => $this->_request->getContext(),
+			'currentLocale' => $locale,
+			'pageTitle' => $application->getNameKey(),
+			'applicationName' => __($application->getNameKey()),
+			'exposedConstants' => $application->getExposedConstants(),
+			'jsLocaleKeys' => $application->getJSLocaleKeys(),
+		));
```
jnugent: Nice use of array() to optimize.

classes/plugins/ThemePlugin.inc.php
asmecher: I went on a rampage some years ago replacing "/" with DIRECTORY_SEPARATOR. It's way less readable and as far as I know has never been necessary. We're currently using a mix of the two, and I grabbed this from someplace that got rampaged. Generally speaking, "/" is a lot nicer to read, but mixing hasn't been a problem.

classes/template/PKPTemplateManager.inc.php
asmecher: I've been itching to do that for a while. The itching has subsided.

PR #379, opened 2015-03-04
*8615* Deprecate submissions.user_id in favour of stage assignments

asmecher:


classes/controllers/modals/editorDecision/PKPEditorDecisionHandler.inc.php
```diff
@@ -391,7 +399,7 @@ protected function _saveEditorDecision($args, $request, $formName, $redirectOp =
 			$notificationMgr->updateNotification(
 				$request,
 				$notificationTypes,
-				array($submission->getUserId()),
+				$authorUserIds,
```
beghelli: Again, this might send notifications to users that are not direct part of this review process, I think.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -322,21 +321,28 @@ function getAssignedById($submissionId, $userId, $contextId = null, $useCache =
 			unset($submission);
 		}
 
-		$params = $this->_getFetchParameters();
-		$params[] = (int) $submissionId;
-		$params[] = (int) $userId;
+		$params = array_merge(
+			array((int) ROLE_ID_AUTHOR),
```
beghelli: is this necessary? I mean, the constant value comes from our own code right?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -322,21 +321,28 @@ function getAssignedById($submissionId, $userId, $contextId = null, $useCache =
...
+			'SELECT	s.*, ps.date_published,
 				' . $this->_getFetchColumns() . '
-				FROM	submissions s
+			FROM	submissions s
 				LEFT JOIN published_submissions ps ON (s.submission_id = ps.submission_id)
+				LEFT JOIN stage_assignments asa ON (asa.submission_id = s.submission_id)
+				LEFT JOIN user_groups aug ON (asa.user_group_id = aug.user_group_id AND aug.role_id = ?)
 				' . $this->_getFetchJoins() . '
-				WHERE	s.submission_id = ?
-				AND s.date_submitted IS NOT NULL AND ps.date_published IS NULL AND s.user_id <> ? AND s.status <> ' .  STATUS_DECLINED .
```
beghelli: with this change if an user is assigned as author, his submission will not appear in the assigned submissions list anymore right? It will now appear in the my submissions grid. I think that previously this was different. My submissions was only the one that had the same user_id (the one that submitted originally). But that makes more sense now.

controllers/api/file/PKPManageFileApiHandler.inc.php
```diff
@@ -171,11 +180,20 @@ function saveMetadata($args, $request) {
 			$metadataForm->execute($args, $request);
 			$submissionFile = $metadataForm->getSubmissionFile();
 
+			// Get a list of author user IDs
+			$authorUserIds = array();
+			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+			$submitterAssignments = $stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR);
+			while ($assignment = $submitterAssignments->next()) {
+				$authorUserIds[] = $assignment->getUserId();
+			}
```
beghelli: maybe we can create a method in ArticleDAO to retrieve the authors user id's and avoid duplicating those lines?

controllers/modals/editorDecision/form/EditorDecisionWithEmailForm.inc.php
```diff
@@ -79,9 +78,8 @@ function initData($args, $request, $actionLabels) {
 		$email = new SubmissionMailTemplate($submission, $emailKeys[$this->getDecision()]);
 
 		$paramArray = array(
-			'authorName' => $submitter->getFullName(),
+			'authorName' => $submission->getAuthorString(),
 			'editorialContactSignature' => $user->getContactSignature(),
-			'authorUsername' => $submitter->getUsername(),
```
beghelli: did you removed this from the email template?

classes/controllers/modals/editorDecision/PKPEditorDecisionHandler.inc.php
asmecher: You're thinking about translators, correct? Hmm, maybe we need to talk to the UI/UX group about this.

classes/submission/SubmissionDAO.inc.php
asmecher: The extra int cast doesn't hurt anything, but yes, it's unnecessary.

classes/submission/SubmissionDAO.inc.php
asmecher: Unless I got the queries wrong, it should now appear in My Submissions if I'm assigned on the author stage, regardless if I'm the one who submitted it or not. This query gets the Assigned Submissions list, which should exclude stuff I'm assigned to as an author, because we don't want submissions to appear in both grids.

controllers/api/file/PKPManageFileApiHandler.inc.php
asmecher: I thought about that, but it's only 3 cases, and I don't like when DAO methods mix a lot of concerns together about different tables/objects.

classes/controllers/modals/editorDecision/PKPEditorDecisionHandler.inc.php
beghelli: yes, who knows what else, because user groups and stages are flexible...

classes/submission/SubmissionDAO.inc.php
beghelli: sure, I checked the queries and it seems right.

controllers/modals/editorDecision/form/EditorDecisionWithEmailForm.inc.php
asmecher: Oops, fixed.

classes/controllers/modals/editorDecision/PKPEditorDecisionHandler.inc.php
asmecher: We talked about this during UI/UX and decided that at least for the moment it's OK for everyone with an author role to receive the notifications. Translators are maybe the only confusing situation at the moment -- others (in OMP in particular) are logical: chapter authors, series editors, etc.

PR #407, opened 2015-03-16
updated README: link to new forum

mtub:


asmecher:
Thanks, @mtub, I've committed similar changes to the other branches where the old forum was mentioned.

PR #409, opened 2015-03-17
updated locales for OJS 2.4.6

mtub:


asmecher:
Thanks, Marco, I must've missed a few!

PR #415, opened 2015-03-22
Ukrainian translation for OCS 2.3

KuznetsovIllya:
Ukrainian translation for OCS 2.3


mtub:
Thanks, @KuznetsovIllya. Will merge and modify the headers after that.

Issue #412, opened 2015-03-20
Possibility to set up the cover thumbnail size and position the thumbnails in the catalog


bozana:
Remove the white background for thumbnails in order to be able to position them (left, middle, right) in the catalog.
Implement the possibility to set up the max with and height of the cover thumbnails and also the possibility to resize the existing cover thumbnails.

asmecher:
Bozana, have you created a pull request for the pkp-lib changes?

asmecher:
Oops! Yes, both are here:
https://github.com/pkp/omp/pull/109
https://github.com/pkp/pkp-lib/pull/413

asmecher:
Merged, thanks!

PR #422, opened 2015-03-24
pkp/pkp-lib#412 possibility to define the size of cover thumbnails

bozana:
I moved the code to master and separated the system specific code... :-)
The other pull request can be removed... :-)

### DIFF
--- a/templates/controllers/tab/settings/appearance/form/appearanceForm.tpl
+++ b/templates/controllers/tab/settings/appearance/form/appearanceForm.tpl
@@ -114,6 +114,8 @@
 	{/fbvFormArea}
 	{* end Layout *}
 
+	{$thumbnailsSizeSettings}
+
 	{* Lists *}
 	{fbvFormArea id="advancedAppearanceSettings" title="manager.setup.lists" class="border"}
 		{fbvFormSection description="manager.setup.listsDescription"}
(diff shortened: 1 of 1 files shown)


templates/controllers/tab/settings/appearance/form/appearanceForm.tpl
```diff
@@ -114,6 +114,8 @@
 	{/fbvFormArea}
 	{* end Layout *}
 
+	{$thumbnailsSizeSettings}
```
asmecher: Calling the variables this way is still OMP-specific. Better to use a generic name that could be used for different purposes, including (but no limited to) the thumbnail stuff in OMP.

bozana: Ah, me crazy... :-\
Would the name 'additionalAppereanceSettings' be OK?

asmecher: Yes, that sounds fine.

bozana:
Alec, I've just seen that the changes for OMP are merged, but this should be merged too, in order for everything to function :-)
Should I in such cases also commit the submodule reference in OMP/OJS?

asmecher:
Sorry, I missed this one. Merged now.

Generally speaking, yes, it's a good idea to commit the submodule hash too -- and see the PKP submodule changes section here: http://pkp.sfu.ca/wiki/index.php?title=Information_for_Developers
If you follow that convention for the commit message for the submodule update, then travis testing will know how to pull the right lib/pkp submodule from your fork.

bozana:
Great, thanks! Sorry for not having done it this time, but next time for sure :-)

Issue #429, opened 2015-03-26
Announcements don't respect date_posted ("published") setting


asmecher:
See http://forum.pkp.sfu.ca/t/published-field-under-announcements-doesnt-work

asmecher:
Pulls:
https://github.com/pkp/pkp-lib/pull/430
https://github.com/pkp/ojs/pull/461
Bruno, mind taking a quick look?

beghelli:
I couldn't open the forum thread but I left some comments there.

asmecher:
Merged!

PR #430, opened 2015-03-26
pkp/pkp-lib#429 For lib/pkp submodule, respect announcement publication date

asmecher:


classes/announcement/PKPAnnouncementDAO.inc.php
```diff
@@ -339,6 +339,7 @@ function &getAnnouncementsNotExpiredByAssocId($assocType, $assocId, $rangeInfo =
 			WHERE assoc_type = ?
 				AND assoc_id = ?
 				AND (date_expire IS NULL OR date_expire > CURRENT_DATE)
+				AND (date_posted < CURRENT_DATE)
```
beghelli: is not null would make more sense right? Can we set the date posted to anything higher than  the current date?

classes/announcement/PKPAnnouncementDAO.inc.php
```diff
@@ -370,16 +372,18 @@ function &getNumAnnouncementsNotExpiredByAssocId($assocType, $assocId, $numAnnou
 	}
 
 	/**
-	 * Retrieve most recent announcement by Assoc ID.
+	 * Retrieve most recent published announcement by Assoc ID.
 	 * @param $assocType int
 	 * @return Announcement
 	 */
-	function &getMostRecentAnnouncementByAssocId($assocType, $assocId) {
+	function &getMostRecentPublishedAnnouncementByAssocId($assocType, $assocId) {
```
beghelli: I think we don't need the reference anymore...

classes/announcement/PKPAnnouncementDAO.inc.php
asmecher: Bruno, are you talking about the date_expired check or the date_posted one?

classes/announcement/PKPAnnouncementDAO.inc.php
asmecher: This is for OJS 2.4.x, and I'd rather not change reference semantics until 3.0. While technically I suspect we no longer have good PHP4 compatibility, there's no sense in disturbing stable code.

beghelli: oups, I was thinking about master, sorry

classes/announcement/PKPAnnouncementDAO.inc.php
beghelli: date_posted one

asmecher: The bug reporter's goal was to permit setting a future date and having the announcement not get published until then.

Issue #433, opened 2015-03-26
Make a user group required on registration


asmecher:
Also ensure that the reader role is available in OJS by default.

asmecher:
PR: https://github.com/pkp/pkp-lib/pull/434

asmecher:
Mind reviewing the PR quickly, Jason?

PR #434, opened 2015-03-26
pkp/pkp-lib#433 Require a user group on registration

asmecher:


jnugent:
Looks good, Alec.  Merging!

Issue #437, opened 2015-03-27
Strict ValidatorInSet type checking breaks subscription type creation


asmecher:
(...and possibly other forms)

Problem report: http://pkp.sfu.ca/support/forum/viewtopic.php?f=8&t=14388
Bug introduced: https://github.com/pkp/pkp-lib/commit/72240b59f0846c68f9c70b0277a02d634f2ef4b1#commitcomment-10437232
Motivation: https://github.com/pkp/pkp-lib/pull/214#discussion_r19964582

jnugent:
we originally made this strict because the original pull request broke the tests on Clinton's VM.  What do we really want this to do?  It's a simple matter to remove the argument that forces strict type checking, so long as this still means that the validator works as it should.  Will switching it break any of our unit tests?

ctgraham:
I think switching back will only "break" the new unit tests I added, particularly https://github.com/pkp/pkp-lib/blob/72240b59f0846c68f9c70b0277a02d634f2ef4b1/tests/classes/validation/ValidatorInSetTest.php#L29 .

As you pointed out earlier, the real "issue" is probably the inclusion of the 0 value in the test set, illustrating one of PHP's "eccentricities".  If we switch back to loose type checking, we document the eccentricity; if we keep strict type checking, we need to modify the subscription (and possibly other components) to conform.

jnugent:
If we're okay with documenting this particular eccentricity of PHP, and reducing the amount of work we need to do to bring things back into form, I'm fine with allowing loose checking for the in_array function call.  We can adjust that one test for ValidatorInSet.  Sound good?

ctgraham:
C.f. https://github.com/pkp/pkp-lib/pull/442

asmecher:
Jason, could you also cherry-pick this into ojs-stable-2_4_6? Does it need forward-porting to the master branch?

jnugent:
Cherry picked to ojs-stable-2_4_6.  I am inclined to leave the strict checking on for master because we don't have subscriptions in master yet, and probably don't need the backwards compatibility if that is the only thing that strict checking affects.  Closing, but feel free to reopen if you feel otherwise.

asmecher:
Jason, we're likely to have this affect master as well. Round-tripping numeric IDs via request parameters is a pretty common operation. May as well save us an hour's head-scratching when this does come up in master; I don't think it'll just be subscriptions.

jnugent:
[...]

PR #442, opened 2015-03-30
pkp/pkp-lib#437 : strict in_array() breaks legacy subscription functionality ...

ctgraham:
... which depends on equalities of int($v) = "$v". Revert to loose in_array() checking.

Not keen on it, but here it is.


jnugent:
Thanks, Clinton!

Issue #497, opened 2015-04-27
Unexpected behavior using "+" operator to append items to arrays in subclasses


gpieri:
OJS and pkp-lib are using the "+" operator for appending items to arrays in subclasses of DAO and Form (e.g. in ControlledVocabEntryDAO).

However the "+" operator not always works as expected. It considers the keys in the merging process, replacing items instead of appending them (more info on http://php.net/manual/en/function.array-merge.php).

For instance:

	php > var_dump(array("a", "b") + array("c"));
	array(2) {
		[0]=>
		string(1) "a"
		[1]=>
		string(1) "b"
	}

The expected behavior  for merging lists, in theses subclasses, is the one provided by array_merge function.

For instance, in ControlledVocabEntryDAO: 

	function getLocaleFieldNames() {
		return parent::getLocaleFieldNames() + array('name');
	}

should be replaced by

	function getLocaleFieldNames() {
		return array_merge(parent::getLocaleFieldNames(), array('name'));
	}

asmecher:
Thanks, @lepidus -- could you include a quick description of the use case that led to this problem being discovered?

asmecher:
Pulls:
https://github.com/pkp/pkp-lib/pull/498
https://github.com/pkp/ojs/pull/484

gpieri:
Sure, @asmecher.

I discovered this issue while developing a plugin for a brazilian journal.

They had a use case where articles belonging to one specific section were book reviews, and they wanted the reference to the reviewed book to be included in the article metadata and shown in the issue summary (http://www.historiadahistoriografia.com.br/revista/issue/view/19 section named "Resenha").

While developing the plugin I added a new property to ArticleDAO using the hook articledao::getLocaleFieldNames. After some testing I discovered that the title wasn't been updated anymore, because instead of appending the new property to the list, it replaced the title (first property in the array) due to this behavior of the + operator.

TL;DR When using the hook articledao::getLocaleFieldNames in a plugin for adding a new property, the article's title wasn't updated anymore by ArticleDAO.

Would you like that I include this info in the commit message, or in the Pull Request?

asmecher:
@gpieri, just making sure I knew the whole picture. Merged, thanks!

PR #498, opened 2015-04-27
pkp/pkp-lib#497 Use array_merge instead of + for merging arrays

lepidus:
Fix issue pkp/pkp-lib#497


beghelli:
Just to make sure, array merge will not preserve numeric key values... not sure if it makes any difference here.

asmecher:
Yes, it's OK -- we've already moved to using array_merge more consistently in master, though I've grepped and fixed a few more that I'll commit soon.

PR #524, opened 2015-05-16
Adding Arabic Language Pack to pkp-lib dev 2_4

vormia:
Greetings;
This is the Arabic translation files of the 'pkp-lib' directory containing all the files in one commit.
If this approach works fine, I would use it with OJS repository as well.


mtub:
Thank you, @vormia. Your css file contains only two changes and shouldn't include the rest of the code you've copied from common.css. But here's another idea: Do you mind having a look at the rtl.css file already in the master branch of pkp-lib (https://github.com/pkp/pkp-lib/blob/master/styles/rtl.css)? We could port that file to the ojs-dev-2_4 branch, too, and patch it with the centered alignment of the header image text you've suggested. Could you check if your translation works with that file and maybe exchange commonAr.css for rtl.css?

vormia:
Hi @mtub;
After trying the file you suggested, I found it **not** enough to satisfy the need of the Arabic user, but when adding the following lines to it:

__div {
	text-align: right; direction: rtl;
}__

it would do...
"Center alignment of the header image text" could *be merged with these changes* to become a good adjustment. I think this would work well with Persian language also, since it is right-to-left orientation language like Arabic.
Yes, if what was mentioned above done, you could **exchange commonAr.css for rtl.css**.

mtub:
Hi @vormia, could you please remove commonAr.css from the pull request?

I've updated the rtl.css file as suggested, see #525.

vormia:
Pardon me @mtub;
But could you please give me a hint how to *remove commonAr.css from the pull request*.
I get the following message: "You must be on a branch to make...."

howto

mtub:
Hi @vormia, you could remove the file locally:

git rm styles/commonAr.css

then commit the change and push to update your pull request.

mtub:
Nice. Ok, this looks good, I will merge it now.

vormia:
Thanks @mtub.

Issue #511, opened 2015-05-04
Payment description HTML support inconsistent


asmecher:
See http://forum.pkp.sfu.ca/t/bug-report-html-not-supported-in-the-manual-fee-payment-message for details.

asmecher:
For OJS 2.4.6, apply these:

- https://github.com/pkp/ojs/commit/29f07fa16a979b0373ed07ef0bfbc18120d95930.diff
- https://github.com/pkp/ojs/commit/61d5919fd5624a289633dd6beab5c94143527f6c.diff

PR #542, opened 2015-06-02
Allow implicit authentication to be optional

davidjb:
To be referenced by an upcoming pull request in OJS.


davidjb:
Updated in line with suggestions from #511.

PR #547, opened 2015-06-04
Remove spurious duplicates of the grid.action.restoreDefaults key

ctgraham:
These keys either duplicate the existing value, or provide an English value in a non-English locale.

The **de_DE also contains a duplicate key**, but the values are non-unique:
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/locale/de_DE/grid.xml#L31
vs.
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/locale/de_DE/grid.xml#L60


mtub:
Thanks, @ctgraham . Will take care of the de_DE file.

PR #613, opened 2015-07-22
Fix minor typing mistake "iff"

samueloph:
I don't know what is the best approach, since this is a very little thing (the type mistake).

If I see another thing like this, should I wait to send more than one together (even them being different files and corrections)? Should I send a PR containing the little change? Or maybe should I open an issue and wait till there's more things to fix?

### DIFF
--- a/classes/file/FileManager.inc.php
+++ b/classes/file/FileManager.inc.php
@@ -53,7 +53,7 @@ class FileManager {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


samueloph:
I have just realized that I should've explained which file I was changing, sorry for the messy and low-detailed PR

asmecher:
Thanks, @samueloph -- actually, "iff" is shorthand for "if and only if" and is sometimes (rarely) used in software.

Issue #674, opened 2015-08-10
cherry-pick locale updates to OJS 2 stable branch


mtub:
* apply latest locale additions and updates from ojs-dev-2_4 to ojs-stable-2_4_6 branches

PR #675, opened 2015-08-10
apply latest locale additions/updates to OJS stable branch

mtub:
* see pkp/pkp-lib#674
* includes cherry-pick of removing a duplicate key from en_US locale


asmecher:
Thanks thanks thanks, @mtub -- that looks like it was a lot of work.

PR #704, opened 2015-08-26
Proposed change to RU locale

ojsbsb:
From one of our library's OJS projects, we have three changes to the Russian locale that might be of general interest: 1f36a4a

### DIFF
--- a/classes/validation/ValidatorISSN.inc.php
+++ b/classes/validation/ValidatorISSN.inc.php
@@ -67,7 +67,7 @@ class ValidatorISSN extends ValidatorRegExp {
 	 * @return string
 	 */
 	static function getRegexp() {
-		return '/(\d{4})-(\d{4})/';
+		return '/^(\d{4})-(\d{3}[\dX])$/';
 	}
 }
 


ppv1979:
These translations of terms are shorter, but unfortunately they doesn't express the same meaning as previous translations. For example, "Index Terms" is not the same as "Keywords".

mtub:
I agree with Pavel, who is maintaining the Russian locale. 

The key is used in the sidebar search menu and also in the enhanced search menu. This displays as:

**Index terms**                           
* Discipline(s) 	
* Keyword(s) 	                           
* Type (method/approach) 	
* Coverage 	
* All index term fields

The proposed change identifies the category (index terms, in key *search.indexTerms*) with one of the subcategories (keywords, in key *search.subject*). 

(And looking at how this is being handled in the other locales helped me see that this line is misleading in the de_DE locale, that also uses "keywords".)

@ojsbsb , I think @ppv1979 is right on this one and will not merge the pull request. However, thank you for sharing your suggestion.

PR #707, opened 2015-08-28
Remove OMP-only  from pkp-lib PageHandler.inc.php

NateWr:
This is getting dumped back in via OMP's template manager.


asmecher:
Happy to see this kind of thing go.

Issue #734, opened 2015-09-10
Usage stats loader task doesn't warn about not being able to move files


beghelli:
See: http://forum.pkp.sfu.ca/t/problems-with-usagestats-metrics-not-updated-v2-4-6-files-not-moved/3297/7

The problem is not appearing on the scheduled tasks logs because the usage stats loader task is not using the _moveFile that's implemented by the FileLoader class, which notifies if it can't move the files.

beghelli:
Fixed and ported to master

PR #735, opened 2015-09-10
pkp/pkp-lib#734: make sure system is logging when not being able to move usage stats log files

beghelli:


asmecher:
Looks good -- please go ahead and merge (porting around as necessary).

Issue #685, opened 2015-08-12
Data Privacy Option for Statistic Plugin


bozana:
Add the data privacy option to the statistic plugin:
-- the logged IP address shall be hashed with a daily randomly generated salt,
-- the user shall be informed about tracking,
-- there should be an opt-out possibility for the users.

beghelli:
I reviewed the code already, but I think that Bozana is going to change one thing still right @bozana?

bozana:
Yes, true, this change is coming end of this week...

bozana:
@beghelli: I changed some things: The salt can now be saved in a file and the information for the users/readers is improved. Could you please review it too?

asmecher:
@beghelli, it looks like this one was merged but not closed. If it is complete, please move back to 2.4.7 and close.

beghelli:
Needs to forward port to master

beghelli:
This is already integrated in both stable and master.

PR #738, opened 2015-09-11
data privacy option for OMP usage stats plugin

bozana:
I would like to make the pull request for this branch, if someone would like to use it immediatelly...
I will then port it to master too, although the master branch seems to currently be in havy development...

### DIFF
--- a/plugins/generic/usageStats/PKPUsageStatsPlugin.inc.php
+++ b/plugins/generic/usageStats/PKPUsageStatsPlugin.inc.php
@@ -47,6 +47,7 @@ class PKPUsageStatsPlugin extends GenericPlugin {
 		if ($this->getEnabled() && $success) {
 			// Register callbacks.
 			HookRegistry::register('PluginRegistry::loadCategory', array($this, 'callbackLoadCategory'));
+			HookRegistry::register('LoadHandler', array($this, 'callbackLoadHandler'));
 
 			// If the plugin will provide the access logs,
 			// register to the usage event hook provider.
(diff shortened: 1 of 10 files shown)


asmecher:
@beghelli, suddenly a flurry of activity on stuff you know best... Mind taking a look? Thanks, @bozana!

bozana:
could you cherry-pick the first commit (the second is already there) to the master or shall I make a new pull request?

plugins/generic/usageStats/README
```diff
@@ -0,0 +1,21 @@
+Data Privacy Option
+--------------------
+
+REQUIREMENTS:
+In order to use the data privacy option, insert the section "[usageSats]" in the
```
beghelli: typo, usageSats instead of usageStats

plugins/generic/usageStats/PKPUsageStatsPlugin.inc.php
```diff
@@ -186,6 +185,10 @@ function callbackLoadCategory($hookName, $args) {
 			$this->import('UsageStatsReportPlugin');
 			$plugin = new UsageStatsReportPlugin();
 		}
+		if ($category ==  'blocks') {
+			$this->import('UsageStatsOptoutBlockPlugin');
+			$plugin = new UsageStatsOptoutBlockPlugin($this->getName());
```
beghelli: I didn't see that before, but this block will always appear, no matter if the manager has activated the data privacy option or not right? Is this what you wanted?

bozana: Ah, I could check here if the option is chosen, right? Thanks Bruno!!! I will then change those two things (the typo too) and let you know...

asmecher:
@beghelli and @bozana, I'm guessing this still needs to be forward-ported to the master branch? If so, and if you're not working on that at the moment, could one of you file an issue so it doesn't get forgotten?

beghelli:
we can track that on https://github.com/pkp/pkp-lib/issues/685 I think...

PR #747, opened 2015-09-16
Install plugins from xz compressed tarball

samueloph:
This PR makes OJS 3 accept plugins installations from a xz compressed tarball.

The .xz format has a better compression ratio than bz, reducing network traffic during the upload of the plugins.


ctgraham:
My reading is that this PR enforces xz as a requirement rather than allowing for xz as an option.  A change such as this shouldn't break current functionality.  Can you re-work to allow both formats?

samueloph:
You're right!
I'm gonna try to make it backwards compatible.

samueloph:
Done!
Closing this PR in favour of a better one backwards compatible.

Thanks for the comments @ctgraham!

PR #751, opened 2015-09-18
Add support for xz compressed plugins

samueloph:
This PR adds support for xz compressed plugins, as seem in https://github.com/pkp/pkp-lib/pull/747.

I removed the "z" argument from tar extraction since tar auto detects wether its a .gz or a .xz and this also removes the need to check the file extension at the extraction routine.

Please let me know if you think there's a better approach to keep backwards compatibility (to keep .tar.gz support).

### DIFF
--- a/classes/plugins/PluginHelper.inc.php
+++ b/classes/plugins/PluginHelper.inc.php
@@ -35,6 +35,21 @@ class PluginHelper {
 	}
 
 	/**
+	 * Match plugin name to perform decompress
+	 * @param $matches array
+	 * @param $originalFileName string Original filename of plugin archive
+	 */
+	private function matchPluginName(&$matches, $originalFileName) {
+		$path_parts = pathinfo($originalFileName);
+		switch($path_parts['extension']) {
+		  case "gz":
+		  String::regexp_match_get('/^[a-zA-Z0-9_.-]+/', basename($originalFileName, '.tar.gz'), $matches);
+		  case "xz":
+		  String::regexp_match_get('/^[a-zA-Z0-9_.-]+/', basename($originalFileName, '.tar.xz'), $matches);
+		}
+	}
+
+	/**
 	 * Extract and validate a plugin (prior to installation)
 	 * @param $filePath string Full path to plugin archive
 	 * @param $originalFileName string Original filename of plugin archive
@@ -45,13 +60,7 @@ class PluginHelper {
 		// equal plugin directory name and plugin files must be in a
 		// directory named after the plug-in (potentially with version)
 		$matches = array();
-		$path_parts = pathinfo($originalFileName);
-		switch($path_parts['extension']) {
-		  case "gz":
-		  String::regexp_match_get('/^[a-zA-Z0-9]+/', basename($originalFileName, '.tar.gz'), $matches);
-		  case "xz":
-		  String::regexp_match_get('/^[a-zA-Z0-9]+/', basename($originalFileName, '.tar.xz'), $matches);
-		}
+		matchPluginName($matches, $originalFileName);
 		$pluginShortName = array_pop($matches);
 		if (!$pluginShortName) {
 			$errorMsg = __('manager.plugins.invalidPluginArchive');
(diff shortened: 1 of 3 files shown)


samueloph:
Oh, by the way, could somebody explain me why there are tests for a 'cli' binary like this one https://github.com/samueloph/pkp-lib/blob/master/classes/plugins/PluginHelper.inc.php#L66 ?

I don't know what this 'cli' is, it looks like some old binary name for tar.

samueloph:
Oh, please don't merge before I update this string https://github.com/pkp/ojs/blob/master/locale/pt_BR/manager.xml#L563.

How should I do it for other languages?

ctgraham:
Re: https://github.com/pkp/pkp-lib/pull/751#issuecomment-141333846, CLI stands for Command Line Interface.  In the line you reference, it is the configuration section in which the `tar` configuration item is located.

ctgraham:
Re: https://github.com/pkp/pkp-lib/pull/751#issuecomment-141426289

In master, the key "manager.plugins.upgradeDescription" has been migrated to the shared library:
https://github.com/pkp/pkp-lib/blob/master/locale/en_US/manager.xml#L374

In 2.4.6, the locale key "manager.plugins.upgradeDescription" used in:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_6/templates/manager/plugins/managePlugins.tpl#L65
is defined in:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_6/locale/en_US/manager.xml#L120

See the documentation here: http://pkp.sfu.ca/wiki/index.php?title=Translating_OxS for information on how we coordinate translations.

asmecher:
BTW, I've filed an issue entry for this: https://github.com/pkp/pkp-lib/issues/752

samueloph:
Is it ready to merge now?
I've fixed the strings and checked file extension on decompression.

I still gonna do some tests, but would like to now if I missed anything.

classes/plugins/PluginHelper.inc.php
```diff
@@ -45,7 +45,13 @@ function extractPlugin($filePath, $originalFileName, &$errorMsg) {
 		// equal plugin directory name and plugin files must be in a
 		// directory named after the plug-in (potentially with version)
 		$matches = array();
-		String::regexp_match_get('/^[a-zA-Z0-9]+/', basename($originalFileName, '.tar.gz'), $matches);
+		$path_parts = pathinfo($originalFileName);
```
ctgraham: I think this code should be a private method so it does not have to be repeated again on https://github.com/pkp/pkp-lib/pull/751/files#diff-192390af7614d51a83a27e64641372b6R83

classes/plugins/PluginHelper.inc.php
```diff
@@ -59,7 +65,7 @@ function extractPlugin($filePath, $originalFileName, &$errorMsg) {
 		// Test whether the tar binary is available for the export to work
 		$tarBinary = Config::getVar('cli', 'tar');
 		if (!empty($tarBinary) && file_exists($tarBinary)) {
-			exec($tarBinary.' -xzf ' . escapeshellarg($filePath) . ' -C ' . escapeshellarg($pluginExtractDir));
+			exec($tarBinary.' -xf ' . escapeshellarg($filePath) . ' -C ' . escapeshellarg($pluginExtractDir));
```
ctgraham: The use of compression autodetection is specific to GNU tar and will fail on other tars, for example Solaris tar.

samueloph: oh no, I let this one pass, sorry!

Let me fix it and double check everything.

samueloph:
Ok, I've made a few changes. I still need to test them.
Should I stop using my already PR'd branch to make this changes and do them on another branch instead?

If you don't mind this way, I think it's easier for you to see if I ain't doing anything wrong.

Also, I hope my mistakes don't cause disturb, considering that PHP isn't really my focus, I just wanna try to help.

ctgraham:
> Should I stop using my already PR'd branch to make this changes and do them on another branch instead?

Test your changes locally on your branch(es) until you are happy with them, then push these changes to github.com branch that you've created the PR against.  Commits can be squashed for clarity.

> I don't know if I should do an exception treatment on decompression

I think we're generally holding off of Exceptions in the current dev branch (though I've started sneaking them into plugins).  I've seen `assert()`s and `trigger_error()`s used in other places. @asmecher would probably be better able to describe current error handling strategies.

> There's still a problem with ...

How about just:
```php
String::regexp_match_get('/^[a-zA-Z0-9_.-]+/', basename($originalFileName, '.tar.'.$path_parts['extension']), $matches);
```
and
```php
exec($tarBinary.' -x'.($path_parts['extension']) == 'xz' ? 'J' : 'x').'f ' . escapeshellarg($filePath) . ' -C ' . escapeshellarg($pluginExtractDir));
```
?
This doesn't handle .tgz or .txz or future options, so there might be a better solution still.

samueloph:
Closing this due to inactivity, I hope to get free time and finish this this month.

Issue #765, opened 2015-09-22
inconsistent language names in registry/locales.xml


bozana:

bozana:
The language names are sometimes in the given language and sometimes in English. An user asked if there could be "Русский" instead of "Russian". If so, I can make a pull request with the change...

mtub:
@bozana We've planned to have the names in the respective language and replace the English names accordingly. I've changed a few of them when new new translations came in. Have to make sure that we use the same name in locale files and in the registry. Can go through all of them and patch accordingly. Do you mind if I take over?

bozana:
@mtub, of course not :-) please go ahead... :-)

mtub:
Will deal with other applications later on.

asmecher:
Merged all 4 PRs, thanks!

PR #775, opened 2015-09-30
Standardized locale names

mtub:
All locales (translations) are now named in their language. Following https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes.

This also addresses pkp/pkp-lib#765.

### DIFF
--- a/locale/da_DK/admin.xml
+++ b/locale/da_DK/admin.xml
@@ -8,7 +8,7 @@
   * Copyright (c) 2000-2015 John Willinsky
   * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
   *
-  * Localization strings for the da_DK (Danish) locale.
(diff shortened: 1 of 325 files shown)


asmecher:
Merged, thanks. I'd suggest that we consider removing some of the cases where the language name is mentioned in order to decrease the maintenance on this kind of thing, in cases where the language name doesn't really add anything useful -- the comments at the top of the files come to mind.

mtub:
Thanks for merging. I think this was rather a one-time operation, also no manual editing involved when checking/fixing. Having the language name in the file (comment) can be helpful. But I have no strong feelings about it.

asmecher:
I suspect we're going to continue getting new translation contributions that need this correction too -- but as long as they get corrected on the way in, we'll stay consistent. No strong feelings either so let's leave it as is for now.

Issue #781, opened 2015-09-30
Forward-port hashing improvements to OMP 1.2 / OJS 3.0


asmecher:
Forward-port https://github.com/pkp/pkp-lib/issues/695

asmecher:
Pulls:
https://github.com/pkp/pkp-lib/pull/785
https://github.com/pkp/ojs/pull/627

@ctgraham and @UTSLibraryITS, if you have the time and interest, would you mind taking a look? It differs from the fix for issue #695 in a few ways:

- The majority of the changes are moved into the lib/pkp submodule, where they can also apply to OMP etc.
- I'm using Composer instead of a git submodule to bring in the support library
- A bunch of the compatibility checks were removed because master-branch releases have a better baseline of PHP support.

PR #785, opened 2015-10-01
pkp/pkp-lib#781 Improve password hashing

asmecher:


classes/security/Validation.inc.php
```diff
@@ -263,15 +307,55 @@ static function generatePassword($length = 8) {
...
+			$expiry = time() + $expires;
+		}
+
+		// use last login time to ensure the hash changes when they log in
+		$data = $user->getUsername() . $user->getPassword() . $user->getDateLastLogin() . $expiry;
+
+		// generate hash and append expiry timestamp
+		$algos = hash_algos();
+
+		foreach (array('sha256', 'sha1', 'md5') as $algo) {
```
ctgraham: Why list md5 here if we're going to fallback below?

ctgraham:
This pull will also need a change to
https://github.com/pkp/pkp-lib/blob/5c1bbdc627ad79b4cced0c00af7a3087d915846f/pages/login/PKPLoginHandler.inc.php#L247-L248
along the lines of https://github.com/pkp/pkp-lib/pull/833

classes/security/Validation.inc.php
asmecher: I'm a little foggy on the details, but this is MD5 using HMAC and the PHP `hmac` function set, vs. plain old MD5 using a legacy function. They will behave a little differently but it's harmless, I think.

Issue #784, opened 2015-10-01
replace a missing translation with the english text


bozana:
Instead of ##..key..## the English text should be displayed, when a translation is missing.

bozana:
S. https://github.com/pkp/pkp-lib/pull/786

ojsbsb:
@bozana I think this is really useful - just one question: Will en_US be hardcoded as the only fallback language? I am asking because depending on the journal, it might make sense to have a custom language set as primary fallback (for example de_DE, if the journal is published in, say, de_DE and cs_CZ) and use en_US only if the translation isn't defined for the primary fallback language either.

mtub:
Hi @ojsbsb, see https://github.com/pkp/defaultTranslation.

We've thought about falling back to another language, but there's only one locale that we can safely consider as complete: en_US. Cascading through multiple languages seems a little bit too much. But if you see a need for this feature, we can take the discussion to the [Forum](http://forum.pkp.sfu.ca/c/translations).

PR #786, opened 2015-10-02
changes needed for the replacement of missing translations with english texts

bozana:
S.https://github.com/pkp/pkp-lib/issues/784
These are the changes needed for the plugin that will replace the missing translations with the English texts. The plugin is coming too.


asmecher:
Committed and cherry-picked to master.

Issue #794, opened 2015-10-05
XML Writer could support Comments


ctgraham:
XML comments could be useful sometimes, say for example if you were working on EZID and needed [a good way to reuse Crossref's code and structure](https://github.com/chasehd/EzidDOI/commit/6088bd0fae8fd86fa9a82d01401c2df3fd320a9b#commitcomment-13598683) with minimum disruption.

ctgraham:
PR is https://github.com/pkp/pkp-lib/pull/795

ctgraham:
@asmecher, the inheritance here is corrected.  Since the new functionality won't really be called from anything in core, perhaps it could be moved up to 2.4.7?

asmecher:
Moved up to 2.4.7 and merged. I'll also cherry-pick to master, though I hope the custom XML stuff won't last long there. Thanks, @ctgraham.

PR #795, opened 2015-10-05
pkp/pkp-lib#794: Enable XMLCustomWriter support for XML comment nodes

ctgraham:
Pretty lightweight change to allow XMLCustomWriter support for comment nodes, I think.

### DIFF
--- a/classes/xml/XMLComment.inc.php
+++ b/classes/xml/XMLComment.inc.php
@@ -1,0 +1,112 @@
+<?php
+
+/**
+ * @file classes/xml/XMLComment.inc.php
+ *
+ * Copyright (c) 2013-2015 Simon Fraser University Library
+ * Copyright (c) 2000-2015 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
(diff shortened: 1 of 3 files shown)


classes/xml/XMLComment.inc.php
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @class XMLComment
+ * @ingroup xml
+ *
+ * @brief Extension of XMLNode for a simple DOM-style comment.
+ */
+
+import ('lib.pkp.classes.xml.XMLNode');
+
+class XMLComment {
```
asmecher: Have you considered making this extend XMLNode? In DOM implementations, OTOH, comments are typically nodes; the way this is implemented there's an unstated assumption that both XMLNode and XMLComment will behave similarly. An interface or class extension would make that explicit.

Note that this whole set of classes will be removed (with great pleasure) in favour of the standard PHP5 toolset at some point. But that won't be done for OJS 2.x.

ctgraham: Oops.  It was supposed to extend XMLNode.

PR #812, opened 2015-10-13
Fixed undefined variable $userGroupDao

selwyntcy:
I came across this undefined variable $userGroupDao when accessing a submission's workflow as a user with the Section Editor role.


asmecher:
Merged, thanks!

Issue #831, opened 2015-10-22
Password reset fails with addition of expiration token


ctgraham:
Validation::generatePasswordResetHash() now adds an expiration token for OJS.
https://github.com/pkp/ojs/blob/ojs-dev-2_4/classes/security/Validation.inc.php#L307-L341

We have a lovely method which validates a password reset hash and expiration.
https://github.com/pkp/ojs/blob/ojs-dev-2_4/classes/security/Validation.inc.php#L349-L359

We never use it.

Instead, we compare a newly generated reset hash against the former hash, expiry and all.
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/pages/login/PKPLoginHandler.inc.php#L240-L241

C.f.: http://forum.pkp.sfu.ca/t/ojs-2-4-7-reset-password-link-expired/4768

asmecher:
Thanks for reporting, @ctgraham. Are you working on this, or shall I?

ctgraham:
Testing a change to PKPLoginHandler::resetPassword() now.

asmecher:
Thanks, @ctgraham. I'll looking into writing a test up to prevent regressions like this once a fix is in place.

asmecher:
For instructions on how to apply the patch to an OJS 2.4.7 installation, see here:
https://github.com/pkp/pkp-lib/pull/833#issuecomment-150335585

PR #833, opened 2015-10-22
pkp/pkp-lib#831: Validate password reset hash via new Validation method

ctgraham:
Tested this on my VM to confirm success with a correct hash/expiry and failure with an incorrect hash/expiry, but then the VM decided not to talk to Github.  So, copy-pasted one line change to a repo that would talk, but doesn't actually have a web front end.

Resolves https://github.com/pkp/pkp-lib/issues/831 (I hope).

### DIFF
--- a/pages/login/PKPLoginHandler.inc.php
+++ b/pages/login/PKPLoginHandler.inc.php
@@ -237,8 +237,7 @@ class PKPLoginHandler extends Handler {
 
 		$templateMgr =& TemplateManager::getManager();
 
-		$hash = Validation::generatePasswordResetHash($user->getId());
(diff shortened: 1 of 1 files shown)


asmecher:
Confirmed that this works for me, @ctgraham. Thanks for your work on this.
The direct patch link, if anyone needs it, is: https://github.com/pkp/pkp-lib/pull/833

You can test applying this to an OJS 2.4.7 installation by running...

    cd /path/to/OJS
    cd lib/pkp
    wget -O - "https://github.com/pkp/pkp-lib/pull/833.diff" | patch -p1 --dry-run

If that appears to work, you can actually applying it by then running...

PR #837, opened 2015-10-26
pkp/pkp-lib#793 Remove unused LinksHandler page

NateWr:


asmecher:
Very happy to see more red than green in PRs these days. Thanks!

PR #849, opened 2015-10-29
pkp/pkp-lib#1086 Version control for OJS3

felixgruender:


felixgruender:
See also:
https://github.com/pkp/ojs/pull/651

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -105,6 +105,11 @@ function initData($submission) {
 			// include all submission metadata fields for submissions
 			$this->_parentForm->setData('submissionSettings', array('all' => true));
 			$this->_parentForm->setData('abstractsRequired', $this->_getAbstractsRequired($submission));
+			
+			$submissionDao = Application::getSubmissionDAO();
+			$submissionSettingsRevisions = $submissionDao->getSubmissionRevisions($submission->getId(), true, true, 'DESC');
```
asmecher: You can use the `SORT_DIRECTION_...` constants defined in `DAO.inc.php` instead of ASC/DESC strings.

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -147,6 +152,25 @@ function execute($submission, $request) {
...
+			$publicationTabFieldNames = $submissionDao->getAdditionalFieldNames();
+			
+			foreach($publicationTabFieldNames as $publicationTabFieldName) {
+				$submission->setData($publicationTabFieldName, $revisionSubmissionData->getData($publicationTabFieldName));
+			}
+		} else {
+			$revision = $submissionDao->getLatestRevisionId($submission->getId());
+		}
+
+		if ($request->getUserVar('saveAsRevision') && ($request->getUserVar('saveAsRevision') == 1)) {
```
asmecher: Wouldn't just...

    if ($request->getUserVar('saveAsRevision') ...

...be sufficient? Much more readable IMO.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
+	function getSubmissionSettings($submission, $submissionRevision = null) {
+
+		if ($submissionRevision == null) {
+			$sql = "SELECT * FROM submission_settings WHERE submission_id = ?";
+			$params = array($submission->getId());
+		} else {
+			$sql = "SELECT * FROM submission_settings WHERE submission_id = ? AND submission_revision = ?";
+			$params = array($submission->getId(), $submissionRevision);
+		}
+		$result = $this->retrieve($sql, $params);
```
asmecher: Consider writing this block as a single SQL statement, e.g.:

    $params = array((int) $submission->getId());
    if ($submissionRevision) $params[] = (int) $submissionRevision;
    $result = $this->retrieve(
        'SELECT * FROM submission_settings WHERE submission_id = ?' .
        ($submissionRevision?' AND submission_revision = ?':''), 
        $params
    );

(Also note the added int cast for submission revision.)

This duplicates less code and if we later had to add another potential condition it would scale better.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
 
 		return $submission;
 	}
+	
+	/**
+	 * Get contents of the submission_settings table
+	 * @param $submission Submission object
+	 * @param $submissionRevision int optional
+	 */
+	function getSubmissionSettings($submission, $submissionRevision = null) {
```
asmecher: The `get` prefix suggests it's going to return a value, but it loads it into the `$submission` instead. Suggest calling the function `loadSubmissionSettings` instead?

asmecher: (I wonder if we're going to have to scale out revisions beyond just the submissions table, e.g. to include author names, galley information, etc. In that case it might be worth making this an optional feature of DAO::getDataObjectSettings.)

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
+	 * 1) Retrieve only the previous revisions, not the current one
+	 * 2) Get only the id's; optionally the title can be retrieved, too
+	 * 3) Default sorting order is ASC
+	 * @param $submissionId int
+	 * @param $showAll boolean
+	 * @param $withTitle boolean
+	 * $param $order string
+	 * @return array
+	 */
+	function getSubmissionRevisions($submissionId, $showAll = false, $withTitle = false, $order = 'ASC') {
```
asmecher: (Use constant for sort order; also probably elsewhere)

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
+		$result->Close();
+	}
+	
+	/**
+	 * Get all revisions for a submission; default settings: 
+	 * 1) Retrieve only the previous revisions, not the current one
+	 * 2) Get only the id's; optionally the title can be retrieved, too
+	 * 3) Default sorting order is ASC
+	 * @param $submissionId int
+	 * @param $showAll boolean
```
asmecher: These parameters could use some inline documentation, e.g.:

    * @param $showAll boolean false to exclude current revision; true to fetch all revisions including current

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
+		if ($withTitle == false) {
+			$sql = 'SELECT DISTINCT submission_revision FROM submission_settings WHERE submission_id = ? ORDER BY submission_revision ' . $order;
+			$params = array($submissionId);
+		} else {
+			$sql = 'SELECT DISTINCT submission_revision, setting_value FROM submission_settings 
+				    WHERE submission_id = ? AND locale = ? AND setting_name = ? ORDER BY submission_revision ' . $order;
+			$params = array($submissionId, AppLocale::getPrimaryLocale(), 'title');
+		}
+		
+		$result = $this->retrieve($sql, $params);
```
asmecher: As suggested above, consider writing this query out once with ternary operators for optional parts.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -111,10 +112,104 @@ function _fromRow($row) {
...
+		return $submissionRevisions;
+	}
+	
+	/**
+	 * Get the current revision id for a submission
+	 * @param $submissionId int
+	 * @param $journalId int
+	 * @return int
+	 */
+	function getLatestRevisionId($submissionId, $journalId = null) {
```
asmecher: This is in pkp-lib so shouldn't refer to OJS-specific aspects (e.g. journalId)

templates/controllers/wizard/fileUpload/form/submissionFileMetadataForm.tpl
```diff
@@ -30,6 +30,25 @@
 		{fbvFormSection title="submission.upload.noteToAccompanyFile"}
 			{fbvElement type="textarea" id="note" height=$fbvStyles.height.SHORT}
 		{/fbvFormSection}
+		{if $submissionSettingsRevisions}
+			<label for="submissionSettingsRevision">{translate key="submission.upload.assignMetadata"}:</label>
+			<select name="submissionSettingsRevision">
```
asmecher: Suggest writing this with `{fbvElement type="select" ...}` rather than a custom `<select>` element. Not that we're totally happy with this approach but it is consistent with the rest of the codebase.

templates/controllers/wizard/fileUpload/form/supplementaryFileMetadataForm.tpl
```diff
@@ -38,6 +38,25 @@
 			{fbvElement label="common.date" inline=true size=$fbvStyles.size.SMALL type="text" id="dateCreated" value=$submissionFile->getDateCreated(null) class="datepicker"}
 			{fbvElement label="common.language" inline=true size=$fbvStyles.size.SMALL type="text" id="language" value=$submissionFile->getLanguage() maxlength="255"}
 		{/fbvFormSection}
+		{if $submissionSettingsRevisions}
+			<label for="submissionSettingsRevision">{translate key="submission.upload.assignMetadata"}:</label>
+			<select name="submissionSettingsRevision">
```
asmecher: (Use fbvElement)

templates/submission/submissionMetadataFormTitleFields.tpl
```diff
@@ -8,7 +8,15 @@
...
  * submission metadata.
  *}
+{if $formParams.revision && ($formParams.revision < $latestRevisionId)}
+	{assign var=rich value=null}
+{else}
+	{assign var=rich value="extended"}
+{/if}
+
 {fbvFormArea id="generalInformation" title="submission.submit.titleAndSummary"}
+	{fbvElement type="hidden" name="submissionRevision" id="submissionRevision" value=$formParams.revision|escape}
```
asmecher: When using fbvElement, escaping is taken care of.

templates/submission/submissionMetadataFormTitleFields.tpl
```diff
@@ -21,6 +29,6 @@
 	{/fbvFormSection}
 	{fbvFormSection description="common.subtitle.tip"}{/fbvFormSection}
 	{fbvFormSection title="common.abstract" for="abstract" required=$abstractsRequired}
-		{fbvElement type="textarea" multilingual=true name="abstract" id="abstract" value=$abstract rich="extended" readonly=$readOnly}
+		{fbvElement type="textarea" multilingual=true name="abstract" id="abstract" value=$abstract rich=$rich readonly=$readOnly}
```
asmecher: Is this change intentional?

xml/oai2.xsl
```diff
@@ -598,7 +598,7 @@ p.intro {
   <xsl:choose>
     <xsl:when test='starts-with(.,"http" )'>
       <xsl:choose>
-        <xsl:when test='string-length(.) &gt; 50'>
+        <xsl:when test='string-length(.) &gt; 80'>
```
asmecher: Is this change intentional? It's 3rd-party code, which we try not to distribute modified copies of if possible.

classes/submission/SubmissionFileDAODelegate.inc.php
```diff
@@ -359,6 +362,18 @@ function _deleteDependentObjects($submissionFile) {
 				ASSOC_TYPE_SUBMISSION_FILE, $submissionFile->getFileIdAndRevision()
 		);
 	}
+	
+	/**
+	 * @copydoc DAO::getAdditionalFieldNames()
+	 */
+	function getAdditionalFieldNames() {
```
asmecher: I think I'd prefer to use a new column in the relevant tables rather than storing this as a setting. It will be a core feature, and there are likely to be performance implications that require some optimization, and that'll be easier if it's a full-fledged column.

templates/submission/submissionMetadataFormTitleFields.tpl
felixgruender: Yes, that was intentional. For the old revisions, I deactivate all input fields - and I realized that deactivating a textarea is only possible when it's not in the "rich" mode. That means that I have to turn off the "rich" mode first when I want to make it read-only.

asmecher: I see. That'll be a problem if there is rich content, because turning off `rich` will result in HTML tags etc. being escaped and displayed to the reader literally.

I can think of two approaches, both best solved in the form builder vocab code:
- Configure TinyMCE to behave read-only as expected when `readonly` is set, or
- have the form builder vocab present a `<div>` element or something, with the contents of the field inside it, and not use TinyMCE at all for cases where `readonly` is set

asmecher:
Superseded by https://github.com/pkp/pkp-lib/pull/1471

Issue #892, opened 2015-11-12
Fix creation of author biography entries from user biographies


asmecher:
See http://forum.pkp.sfu.ca/t/omp-displaying-bio-in-catalog-book-abstracttab/662/5

asmecher:
PR: https://github.com/pkp/pkp-lib/pull/893

PR #893, opened 2015-11-12
pkp/pkp-lib#892 Fix creation of author biography entries from user bios

asmecher:


asmecher:
Cherry-picked to omp-dev-1_1

PR #930, opened 2015-11-27
Implement Pub ID plugins for OJS 3.0 / OMP 1.2

asmecher:
(PR created for code review purposes, not merging yet)

### DIFF
--- a/classes/plugins/PKPPubIdPlugin.inc.php
+++ b/classes/plugins/PKPPubIdPlugin.inc.php
@@ -1,0 +1,410 @@
+<?php
+
+/**
+ * @file classes/plugins/PKPPubIdPlugin.inc.php
+ *
+ * Copyright (c) 2014-2015 Simon Fraser University Library
+ * Copyright (c) 2003-2015 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * @class PKPPubIdPlugin
+ * @ingroup plugins
+ *
+ * @brief Abstract class for public identifiers plugins
+ */
+
+import('lib.pkp.classes.plugins.Plugin');
+
+abstract class PKPPubIdPlugin extends LazyLoadPlugin {
+
+	/**
+	 * Constructor
+	 */
+	function PKPPubIdPlugin() {
+		parent::LazyLoadPlugin();
+	}
+
+
+	//
+	// Implement template methods from Plugin
+	//
+	/**
+	 * @copydoc Plugin::register()
+	 */
+	function register($category, $path) {
+		if (!parent::register($category, $path)) return false;
+		// Enable storage of additional fields.
+		foreach($this->getDAOs() as $publicObjectType => $daoName) {
+			HookRegistry::register(strtolower_codesafe($daoName).'::getAdditionalFieldNames', array($this, 'getAdditionalFieldNames'));
+		}
+		return true;
+	}
+
+	/**
+	 * @copydoc Plugin::getActions()
+	 */
+	function getActions($request, $actionArgs) {
+		$router = $request->getRouter();
+		import('lib.pkp.classes.linkAction.request.AjaxModal');
+		return array_merge(
+				$this->getEnabled()?array(
+						new LinkAction(
+								'settings',
+								new AjaxModal(
+										$router->url($request, null, null, 'manage', null, $actionArgs),
+										$this->getDisplayName()
+										),
+								__('manager.plugins.settings'),
+								null
+								),
+				):array(),
+				parent::getActions($request, $actionArgs)
+				);
+	}
+
+ 	/**
+	 * @copydoc Plugin::manage()
+	 */
+	function manage($args, $request) {
+		$user = $request->getUser();
+		$router = $request->getRouter();
+		$context = $router->getContext($request);
+
+		$settingsFormName = $this->getSettingsFormName();
+		$settingsFormNameParts = explode('.', $settingsFormName);
+		$settingsFormClassName = array_pop($settingsFormNameParts);
+		$this->import($settingsFormName);
+		$form = new $settingsFormClassName($this, $context->getId());
+		$notificationManager = new NotificationManager();
+		if ($request->getUserVar('save')) {
+			$form->readInputData();
+			if ($form->validate()) {
+				$form->execute();
+				$notificationManager->createTrivialNotification($user->getId(), NOTIFICATION_TYPE_SUCCESS);
+				return new JSONMessage(true);
+			} else {
+				return new JSONMessage(true, $form->fetch($request));
+			}
+		} elseif ($request->getUserVar('clearPubIds')) {
+			$contextDao = Application::getContextDAO();
+			$contextDao->deleteAllPubIds($context->getId(), $this->getPubIdType());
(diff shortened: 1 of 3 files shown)


classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	//
+	// Implement template methods from Plugin
+	//
+	/**
+	 * @copydoc Plugin::register()
+	 */
+	function register($category, $path) {
+		if (!parent::register($category, $path)) return false;
+		// Enable storage of additional fields.
+		foreach($this->getDAOs() as $publicObjectType => $daoName) {
```
asmecher: I think this should only happen if the plugin is enabled, right?

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+
+ 	/**
+	 * @copydoc Plugin::manage()
+	 */
+	function manage($args, $request) {
+		$user = $request->getUser();
+		$router = $request->getRouter();
+		$context = $router->getContext($request);
+
+		$settingsFormName = $this->getSettingsFormName();
```
asmecher: Why not create a

    abstract function instantiateSettingsForm();

...then implement it in any subclasses of the pub ID plugin? That would save us parsing out names from strings.

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+		$settingsFormNameParts = explode('.', $settingsFormName);
+		$settingsFormClassName = array_pop($settingsFormNameParts);
+		$this->import($settingsFormName);
+		$form = new $settingsFormClassName($this, $context->getId());
+		$notificationManager = new NotificationManager();
+		if ($request->getUserVar('save')) {
```
asmecher: Generally I prefer a pattern like

    switch ($request->getUserVar('verb')) {
        case 'save':
            ...
        case 'clearPubIds':
            ...
    }

asmecher: (See e.g. https://github.com/pkp/ojs/blob/master/plugins/generic/webFeed/WebFeedPlugin.inc.php which is a little different than what I quoted above)

asmecher: Part of what this improves is the fallback behavior, e.g.:

    function manage(...) {
        switch (something) {
            case 'flarm':
                ... handle it ...
                return new JSONMessage(true); // Execution ends here!
        }
        return parent::manage(...); // Parent class gets called to deal with anything else.
    }

Hope that's making some sense.

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	 * @return string
+	 */
+	abstract function constructPubId($pubIdPrefix, $pubIdSuffix, $contextId);
+
+	/**
+	 * Public identifier type, see
+	 * http://dtd.nlm.nih.gov/publishing/tag-library/n-4zh0.html
+	 * @return string
+	 */
+	abstract function getPubIdType();
```
asmecher: Nice to finally be able to use abstract functions instead of

    fatalError('WHY ARE WE HANDLING ABSTRACTS THIS WAY?!');

:)

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	 */
+	abstract function isObjectTypeEnabled($pubObjectType, $contextId);
+
+	/**
+	 * Should the object be excluded from
+	 * assigning it the pub id
+	 * @param $pubObject object
+	 * @return boolean
+	 */
+	function isObjectExcluded($pubObject) {
```
asmecher: I'm hoping we can get rid of some of the exclusion stuff if we do what we discussed -- including the assignment step somewhere in the publication workflow. It'll be nice to reduce the number of abstract methods here.

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	/**
+	 * Verify form data.
+	 * @param $fieldName string The form field to be checked.
+	 * @param $fieldValue string The value of the form field.
+	 * @param $pubObject object
+	 * @param $contextId integer
+	 * @param $errorMsg string Return validation error messages here.
+	 * @return boolean
+	 */
+	function verifyData($fieldName, $fieldValue, &$pubObject, $contextId, &$errorMsg) {
```
asmecher: Not sure the reference on pubObject is required (though I guess the one on errorMsg still is).

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	 * Set and store a public identifier.
+	 * @param $pubObject object
+	 * @param $pubObjectType string As returned from self::getPubObjectType()
+	 * @param $pubId string
+	 * @return string
+	 */
+	function setStoredPubId(&$pubObject, $pubObjectType, $pubId) {
+		$daos = $this->getDAOs();
+		$dao = DAORegistry::getDAO($daos[$pubObjectType]);
+		$dao->changePubId($pubObject->getId(), $this->getPubIdType(), $pubId);
```
asmecher: Do we want to use an interface or subclass of DAOs that defines some of these methods that are assumed to exist? Might be nice.

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+	// Public API
+	//
+	/**
+	 * Check for duplicate public identifiers.
+	 * @param $pubId string
+	 * @param $pubObject object
+	 * @param $contextId integer
+	 * @return boolean
+	 */
+	function checkDuplicate($pubId, &$pubObject, $contextId) {
```
asmecher: Where possible, remove reference use. It's not needed on $pubObject anymore. (PHP5 always treats variables referring to objects as references. PHP4 would create copies all over the place.)

classes/plugins/PKPPubIdPlugin.inc.php
```diff
@@ -0,0 +1,410 @@
...
+
+			unset($objectsToCheck);
+		}
+
+		// We did not find any ID collision, so go ahead.
+		return true;
+	}
+
+	/**
+	 * Get the journal object.
```
asmecher: Make sure we're not using OJS language in the PKP library.

classes/plugins/PKPPubIdPluginHelper.inc.php
```diff
@@ -0,0 +1,134 @@
...
+				$form->readUserVars(array($clearFormFieldName));
+			}
+		}
+	}
+
+	/**
+	 * Set the additional data from public identifier plugins.
+	 * @param $form object
+	 * 	OJS IssueForm, MetadataForm, or ArticleGalleyForm i.e. OMP ...
+	 * @param $pubObject object An Article, Issue, or ArticleGalley
```
asmecher: I think we have a new level to consider: issue, article, galley, and file.

asmecher:
@bozana, do you remember what the state of this PR was?

PR #974, opened 2015-12-15
pkp/pkp-lib#972: clean up submission lists from site wide aspect leftovers

beghelli:


asmecher:
Like all those removals.

Issue #935, opened 2015-11-30
Poor SQL performance for ReviewFormDAO::getActiveByAssocId()


ctgraham:
When testing access to editing a Journal Section in OJS 2.4, I'm finding slow performance on the query in `ReviewFormDAO::getActiveByAssocId()`.

In a sample with 37 review_forms records and 12,133 review_assignments records, on MySQL 5.1, the [as-is query with two LEFT JOINS](https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/classes/reviewForm/ReviewFormDAO.inc.php#L279-L292) runs for 1:18.93.

A modified query with a single LEFT JOIN and an OR condition runs in 0:00.05.
```
SELECT rf.*
   FROM review_forms rf
    LEFT JOIN review_assignments rac ON (
     rac.review_form_id = rf.review_form_id AND
     (rac.date_confirmed IS NOT NULL OR (rac.date_confirmed IS NULL AND rac.date_notified IS NOT NULL))
    )
   WHERE rf.assoc_type = ? AND assoc_id = ? AND rf.is_active = 1
   GROUP BY rf.assoc_type, rf.assoc_id, rf.review_form_id, rf.seq, rf.is_active
   ORDER BY rf.seq
```

A modified query with a subquery runs in 0:00.0006.
```
SELECT rf.*
   FROM review_forms rf
   WHERE rf.assoc_type = ? AND assoc_id = ? AND rf.is_active = 1 AND rf.review_form_id IN
    (SELECT rac.review_form_id FROM review_assignments rac WHERE
     rac.review_form_id = rf.review_form_id AND
     (rac.date_confirmed IS NOT NULL OR (rac.date_confirmed IS NULL AND rac.date_notified IS NOT NULL))
    )
   ORDER BY rf.seq
```

Can someone with Postgres validate these alternate syntaxes and runtimes?

asmecher:
@ctgraham, reviewing the SQL and this change looks good to me.

Both formulations work in PostgreSQL (testing for syntax, not for correct results).

Personally I try to avoid subqueries but I suspect that's a historical aversion and no longer relevant.

ctgraham:
One potential problem with subqueries is that MySQL and Postgres differ on whether `NOT IN (NULL)` is true or false/undefined.  That oddity is still relevant, to my knowledge.

[...]

PR #1008, opened 2016-01-04
pkp/pkp-lib#935: replace multiple LEFT OUTER JOINs in Review Forms

ctgraham:
Change LEFT OUTERs to EXISTS / NOT EXISTS clauses if needed; remove one unused LEFT OUTER JOIN set.  This is done to improve performance on large sets.

Resolves https://github.com/pkp/pkp-lib/issues/935

### DIFF
--- a/classes/reviewForm/ReviewFormDAO.inc.php
+++ b/classes/reviewForm/ReviewFormDAO.inc.php
@@ -278,17 +278,7 @@ class ReviewFormDAO extends DAO {
 		$result =& $this->retrieveRange(
 			'SELECT	rf.*
 			FROM	review_forms rf
-				LEFT JOIN review_assignments rac ON (
-					rac.review_form_id = rf.review_form_id AND
-					rac.date_confirmed IS NOT NULL
-				)
-				LEFT JOIN review_assignments rai ON (
-					rai.review_form_id = rf.review_form_id AND
-					rai.date_notified IS NOT NULL AND
-					rai.date_confirmed IS NULL
-				)
 			WHERE	rf.assoc_type = ? AND assoc_id = ? AND rf.is_active = 1
-			GROUP BY rf.assoc_type, rf.assoc_id, rf.review_form_id, rf.seq, rf.is_active
 			ORDER BY rf.seq',
 			array((int) $assocType, (int) $assocId), $rangeInfo
 		);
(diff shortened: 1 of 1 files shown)


beghelli:
Hi @ctgraham ,

I am no expert, but a quick research and I saw this: http://explainextended.com/2009/09/18/not-in-vs-not-exists-vs-left-join-is-null-mysql/ and http://explainextended.com/2009/09/16/not-in-vs-not-exists-vs-left-join-is-null-postgresql/

If I got it correct and according to that material, when you use LEFT JOIN together with IS NULL or  IS NOT NULL on the query, both MySQL and Postgres are able to optimize that, and the performance is the same or even faster than EXISTS. Were you able to test and see the performance improvement? Thanks.

ctgraham:
I'm testing against MySQL 5.1, as the default version shipped on RHEL.  The difference in performance with NOT EXISTS was similar to the [dramatic difference on NOT IN](https://github.com/pkp/pkp-lib/issues/935#issue-119522899).

classes/reviewForm/ReviewFormDAO.inc.php
```diff
@@ -278,17 +278,7 @@ function &getActiveByAssocId($assocType, $assocId, $rangeInfo = null) {
...
-					rac.review_form_id = rf.review_form_id AND
-					rac.date_confirmed IS NOT NULL
-				)
-				LEFT JOIN review_assignments rai ON (
-					rai.review_form_id = rf.review_form_id AND
-					rai.date_notified IS NOT NULL AND
-					rai.date_confirmed IS NULL
-				)
 			WHERE	rf.assoc_type = ? AND assoc_id = ? AND rf.is_active = 1
-			GROUP BY rf.assoc_type, rf.assoc_id, rf.review_form_id, rf.seq, rf.is_active
```
beghelli: maybe you forgot to add the EXISTS sql here?

ctgraham: No, this removal is intentional.  See https://github.com/pkp/pkp-lib/issues/935#issuecomment-160763433

All these joins do is slow the query!  :-)

beghelli: :) great. @asmecher, do you want to test with postgres this?

asmecher: It's a trivial enough query -- I ran a quick smoke test:

    SELECT rf.* FROM review_forms rf WHERE rf.assoc_type = 1 AND assoc_id = 1 AND rf.is_active = 1 ORDER BY rf.seq;

...and it looks OK to me.

Issue #1011, opened 2016-01-04
Types in submission.citations.filter.unsupportedPublicationType should not be translated


asmecher:
The symbolic names `book`, `journal`, `conf-proc`, and `thesis`, specified as options in the English-language version of the `submission.citations.filter.unsupportedPublicationType` locale key, appear to have been translated (at least in the case of the Spanish locale). These are symbolic names; ensure that they aren't translated.

See http://forum.pkp.sfu.ca/t/problem-with-publication-type-field-in-citation-tool-spanish-language/6459

mtub:
I'll take care of this. Almost every locale but de_DE and ru_RU has this. We've had similar cases when translators translated code examples, symbolic names, etc. At some point, we might think about not having these static kinds of strings in the locale strings, but reading them from another file would make the code much less readable, so that's probably not an option.

mtub:
When checked and merged, I can cherry-pick this for master and omp-dev-1_1 branches.

asmecher:
@mtub, one approach that would prevent this, if you think it's useful:

In the code...

   $translatedKey = __('submission.citations.filter.unsupportedPublicationType', array('book' => 'book', 'journal' => 'journal', 'conf-proc' => 'conf-proc', 'thesis' => 'thesis'));

In the locale XML file...

    <message key="submission.citations.filter.unsupportedPublicationType">Unsupported or missing publication type. Please manually edit the citation and enter a publication type (allowed types are '{$book}', '{$journa}', '{$thesis}' and '{$conf-proc}').</message>

Using variables might be a good hint not to translate. But I'm not 100% convinced it's worth the trouble.

mtub:
I think it's too much. We are now having a closer look at translations. I will try to include a note on "Watch out for code, variables, etc. when translating" in the translation documentation.

asmecher:
Sounds good to me, thanks.

asmecher:
@mtub, is this ready to be closed?

asmecher:
Presuming a yes -- correct me if I'm wrong, @mtub.

PR #1014, opened 2016-01-05
removed translated pubtype names

mtub:
* fix for #1011 
* when possible, replaced the pubtype names by static English names
* when not possible, removed the sentence or the complete entry


beghelli:
Thanks Marco, merged.

PR #1018, opened 2016-01-05
removed translated pubtype names

mtub:
cherry-picked, see #1014 and https://github.com/pkp/pkp-lib/issues/1011


asmecher:
Thanks, @mtub -- I've added the pkp-lib issue numbers to the titles of the PRs. That makes it easier to track all the related information.

Issue #1039, opened 2016-01-11
[OJS] Announcements are not visible at the published date


lepidus:
Announcements are not visible at the published date. They only become visible the day after.

Announcement should be visible at the published date.

asmecher:
Master branch port:
https://github.com/pkp/pkp-lib/commit/3d6029a40d060141375530aab2c68cf09bd7c2a8
(Typo: Commit message includes wrong bug number)

Thanks, @lepidus!

gpieri:
You're welcome, @asmecher.

(Regarding the typo... I spotted it after submitting the PR, I thought had fixed it, sorry. We're using the issue number on the commit message, in this case 1039. Should we start using the PR number as in the master branch port?)

asmecher:
No, that was my typo, not yours! Sorry, I wasn't clear.

gpieri:
I see. I misinterpreted it. :-)

gpieri:
Hi, @mfelczak. I updated the PR with the suggested changes. Let us know if there is any additional improvement.

asmecher:
Merged the revised commit; will port around.

PR #1040, opened 2016-01-11
pkp/pkp-lib#1039 Show announcements at the published date

lepidus:
Fixes issue pkp/pkp-lib#1039

### DIFF
--- a/classes/announcement/PKPAnnouncementDAO.inc.php
+++ b/classes/announcement/PKPAnnouncementDAO.inc.php
@@ -338,7 +338,7 @@ class PKPAnnouncementDAO extends DAO {
 			FROM announcements
 			WHERE assoc_type = ?
 				AND assoc_id = ?
-				AND (date_expire IS NULL OR date_expire > CURRENT_DATE)
+				AND (date_expire IS NULL OR date_expire >= CURRENT_DATE)
 				AND (date_posted < CURRENT_DATE)
 			ORDER BY announcement_id DESC',
 			array((int) $assocType, (int) $assocId),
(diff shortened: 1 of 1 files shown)


classes/announcement/PKPAnnouncementDAO.inc.php
```diff
@@ -338,7 +338,7 @@ function &getAnnouncementsNotExpiredByAssocId($assocType, $assocId, $rangeInfo =
 			FROM announcements
 			WHERE assoc_type = ?
 				AND assoc_id = ?
-				AND (date_expire IS NULL OR date_expire > CURRENT_DATE)
+				AND (date_expire IS NULL OR date_expire >= CURRENT_DATE)
```
beghelli: What does date_expire have to do with the publish date? I am not sure if I followed your fix, but it seems that you changed the way the expire date is working, getting announcements that would be expired if the equal operator was not used.

classes/announcement/PKPAnnouncementDAO.inc.php
```diff
@@ -338,7 +338,7 @@ function &getAnnouncementsNotExpiredByAssocId($assocType, $assocId, $rangeInfo =
...
 				AND (date_posted < CURRENT_DATE)
```
beghelli: I think that the fix should go here, date_posted <= CURRENT_DATE, right? And thanks for your bug report and the pull request :)

gpieri: Whoops -- You're absolutely right, @beghelli. I cherry-picked the wrong commit from my development branch for creating the PR. Sorry. 

I'm not sure how I should proceed now. Should I update this PR or create a new one for you guys?

asmecher: And this is my fault too for skimming the code so quickly. @gpieri, can you submit a PR that corrects the earlier one? I'll port it around.

gpieri: Sure, @asmecher. I will do it now.

PR #1060, opened 2016-01-17
Locale Standardization

samueloph:
Removes unnecessary empty lines and standarizes the comment section with a break line before and after it.
I'm currently updating some of the pt_br locales and found some unnecessary empty lines, so i've also standarized a empy line before and after the comment header of the files.

Not really *usefull*, but anyway here it go.

### DIFF
--- a/locale/en_US/admin.xml
+++ b/locale/en_US/admin.xml
@@ -35,7 +35,6 @@
 	<message key="admin.languages.download.cannotOpen">Cannot open language descriptor from PKP web site.</message>
 	<message key="admin.languages.download.cannotModifyRegistry">Cannot add the new locale to the locale registry file, typically "registry/locales.xml".</message>
 	<message key="admin.languages.cantDisable">This locale is the primary language of the site. You can't disable it until you choose another primary locale.</message>
-
 	<message key="admin.languages.confirmReload">Are you sure you want to reload this locale? This will erase any existing locale-specific data such as customized email templates.</message>
 	<message key="admin.languages.installedLocales">Installed Locales</message>
 	<message key="admin.languages.installLanguages">Manage Locales</message>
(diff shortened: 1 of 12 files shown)


mtub:
Thanks, @samueloph. As much as I like standardization, I'd rather not merge this at the moment: It doesn't add any specific value, but breaks the comparability of locales (as long as we don't change it in every locale). Having content on the same line in different locales enables us to run a diff quite easily.

Regarding removing empty lines and sorting alphabetically: Often someone put some meaning into the structure, grouping strings that cover a new feature, etc. Even if there are some files that are sorted, I don't think this is especially useful. I'd rather align the locale files to the en_US locale files (taking into account that some of them are quite messed up) then sort every file alphabetically.

samueloph:
I think that sorting keys (keeping their sections* integrity) is good for the translation proccess because I did not see diff comparability being enforced at least on the pt_br files, which I found strings out of order (comparing to en_us, not alphabetically) and then I had to to pick each one of the mis-positioned strings and check where they should be with the us files (because there wasn't a sorting schema proposed, eg. alphabetically).

Changing to alphabetical sorting of the strings would make it easier to update and fix translation, as it wouldn't be needed to check where the new strings are and which/how many of them already are in the wrong position (I just checked 2 files of the pt_br, there could be a lot more), one could only add them and sort again.

You can see in [this commit](https://github.com/samueloph/pkp-lib/commit/a7e3ab8bd1d46659036277e78e2cdab526abd0a1) (before sorting the pt_br translation) that I almost ended retranslating a lot of strings that already were in the file because they were out of order (not alphabetically, but they simply didn't followed the en_us corresponding order). And I couldn't even just sort all the pt_br strings to automatically position them, I had to check each one and where they were on en_us. 
When I found this lack of integrity in the order of the strings and realize that my diffs were showing lack of strings that were already translated I had to change my approach so I didn't make work that already has been done.

I believe the meaning to the strucure can be made using sections and not just empty lines, then we would know each section must be sorted individually (It happens that the files already look like this was the default and they were dropped some time ago) and the section name, even considering that in most of the cases a section isn't needed because the alphabetical sorting would already keep them together and made it easier to found them without ctrl+f.

I know that this change would make translations shows the same problems because the us_en would be in a different order too (en_us alph. sorted, others not), but it's not like we can't sort the others too and make compatible. And even formalize a translation guideline to OJS.

But anyway, it's not a major improvment, so it's up to you. Though I really think this would make it easier to the translators, considering that could be a lot more translated strings that were accepted in the wrong order (not equal to us_en order).

Just to be sure, considering that I'm not a native speaker: I didn't mean to be rude or anything like that.

*sections are delimited by a commented line, [like this](https://github.com/pkp/pkp-lib/blob/master/locale/en_US/common.xml#L342)

mtub:
@samueloph, not rude at all, it's a reasonable suggestion. :) - I've been struggling  with the structure of the xml files myself. I just think that sorting alphabetically without sections etc. would remove the last bit of context that translators have when looking at a particular string. Often reading the context in which the lines have been introduced helps at least a bit.

That being said, it's probably got to not read too much meaning into the order of lines - often keys are used in different contexts, etc. Personally, I'd really like to see more comments/sections in the en_US locale files.

But it's only recently that there are more than 2, 3 translators working directly with the XML files, using git and GitHub, etc. - a development that I really like. Some translators work with the translator plugin (https://github.com/pkp/translator) with nothing more than the English line that they then try to translate, sometimes without a deeper understanding of the purpose of that line. So we always have to serve different kinds of translators. I will keep your suggestion in mind and think about ways to achieve better English locale files.

But I think we agree that the root is the English locale, and whether sorted alphabetically, structured with sections, or left as they are, the other locales should be in the same order to make diffs easy. With that in mind, how shall we proceed with your other pull request (pkp/pkp-lib#1061), given that I probably will not merge this?

Your work and suggestion is really appreciated, thanks!

samueloph:
I can revert the sorting commit from that branch and then you can merge it, is that ok?

Issue #1092, opened 2016-01-26
OAI-PMH interface strips HTML, but leaves HTML entities


ctgraham:
To display record content, OAI-PMH strips the HTML tags via `stripAssocArray()`:
https://github.com/pkp/pkp-lib/blob/a39b1328db9ec06d2902e67b337dc5a63c60a496/includes/functions.inc.php#L331

This, however, just calls strip_tags, and so HTML entities are left intact.  These HTML entities are then (appropriately) escaped for XML output and passed to the client.

If plaintext output is intended for something like the dc.description, then it really should be plaintext.

asmecher:
Thanks, @ctgraham!

PR #1093, opened 2016-01-26
pkp/pkp-lib#1092: Add HTML entity decode to stripAssocArray() for conversion of HTML to plaintext.

ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/1092, at least to the degree that PHP is natively able, by decoding known entities into the specified client character set.

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -322,13 +322,19 @@ function arrayClean(&$array) {
 
 /**
  * Recursively strip HTML from a (multidimensional) array.
+ * Assumes entity decoding to UTF-8 (primarily for OAI-PMH)
  * @param $values array
  * @return array the cleansed array
  */
-function stripAssocArray($values) {
+function stripAssocArray($values, $useClientCharset = false) {
 	foreach ($values as $key => $value) {
 		if (is_scalar($value)) {
 			$values[$key] = strip_tags($values[$key]);
+			if ($useClientCharset && strtolower(Config::getVar('i18n', 'client_charset')) !== 'utf-8') {
+				$values[$key] = html_entity_decode($values[$key], ENT_QUOTES, Config::getVar('i18n', 'client_charset'));
+			} else {
+				$values[$key] = String::html2utf($values[$key]);
+			}
 		} else {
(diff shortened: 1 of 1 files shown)


includes/functions.inc.php
```diff
@@ -328,7 +328,7 @@ function arrayClean(&$array) {
 function stripAssocArray($values) {
 	foreach ($values as $key => $value) {
 		if (is_scalar($value)) {
-			$values[$key] = strip_tags($values[$key]);
+			$values[$key] = html_entity_decode(strip_tags($values[$key]), ENT_QUOTES, Config::getVar('i18n', 'client_charset'));
```
asmecher: The OAI interface is fixed to use UTF8, regardless of what the client charset is set to, per the OAI-PMH spec.

ctgraham: Even better.  Will fix.

ctgraham:
This new change is overkill.  This function is currently used only in the context of OAI-PMH, but is defined outside of the OAI-PMH context.  I felt dirty hardcoding the function to UTF-8, so now it has an unused parameter.

includes/functions.inc.php
```diff
@@ -322,13 +322,14 @@ function arrayClean(&$array) {
...
+ * Assumes entity decoding to UTF-8 (primarily for OAI-PMH)
  * @param $values array
  * @return array the cleansed array
  */
-function stripAssocArray($values) {
+function stripAssocArray($values, $useClientCharset = false) {
 	foreach ($values as $key => $value) {
 		if (is_scalar($value)) {
-			$values[$key] = strip_tags($values[$key]);
+			$values[$key] = html_entity_decode(strip_tags($values[$key]), ENT_QUOTES, $useClientCharset ? Config::getVar('i18n', 'client_charset') : 'UTF-8');
```
mfelczak: @ctgraham, we also have String::html2utf as an alternative to PHP's built-in html_entity_decode.

ctgraham: Sure, I can adapt it to that.

ctgraham:
Transitioned the default usage to the more consistent `String::html2utf()` per @mfelczak.

PR #1107, opened 2016-02-02
Fix TinyMCE variable replacement regex

selwyntcy:
Avoid matching variables in HTML tag attributes, such as `<a href="{$submissionReviewUrl}">{$submissionReviewUrl}</a>`

### DIFF
--- a/js/controllers/SiteHandler.js
+++ b/js/controllers/SiteHandler.js
@@ -195,7 +195,7 @@
 					'#' + tinyMCEObject.id);
 
 			e.content = e.content.replace(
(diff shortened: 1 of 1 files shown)


asmecher:
Nice solution, @selwyntcy -- I was planning on working around this PHP-side, which would have been a lot messier. Thanks for contributing!

selwyntcy:
Not sure if it is foolproof though. It relies on negative lookahead only, as javascript regex doesn't support lookbehind.

And thank you and your team for your enormous effort in the project indeed!

Issue #1106, opened 2016-02-01
[OMP] Can’t edit contributors to a book


crism:
Against `HEAD`: in the Catalog Entry modal pop-up, on the Monograph tab, contributors can be added and reordered, but there seems to be no way to edit their names, e-mail addresses, roles, or status as primary contact or browse-list presence.

crism:
This may be due to badly-constructed HTML; the check-boxes, for instance, have:
```html
<div id="isChecked">
  <div href="#" class="pkp_helpers_container_center checked"></div>
</div>
```
The multiple `div` elements with the same `id` are not valid, nor is the use of `href` on a `div`. This may lead to JavaScript events misfiring… I’m investigating.

asmecher:
(Note that this may also be affecting the Chapters grid -- I recently noticed that the category row actions aren't showing.)

crism:
The chapters grid has its own handler, but curiously, has an `id` of `authorsGridContainer-...`.

What they may have in common is that the grids are getting set to read-only, which disables a bunch of controls… but I’m still tracing the code, so that’s just a gut guess right now.

asmecher:
Just let me know if you want a hand, @crism.

crism:
Why does `PKPAuthorGridHandler` have `editAuthor()` _and_ `updateAuthor()`? That looks suspicious.

Overall, @asmecher, this section of code looks like it wasn’t finished. There are what looks like cut-and-paste errors in both the PHP and the Smarty templates. It might benefit from a review from you.

asmecher:
@crism, I just ran a quick test and I'm able to edit authors as I'd expect. Maybe something specific to your setup -- would you be able to send me a DB dump along with a few notes on duplicating the problem (e.g. what account you're logged in with, and what submission ID you're looking at)?

crism:
Hm. I thought it might have something to do with the published status of the submission in question, but a different, unpublished submission has the same problem, in both Chrome for Windows and Firefox for Linux. Will try to prepare a database snapshot for you.

asmecher:
Thanks, @crism. What's happening is that the grid is deciding you don't have permission to edit authors because you're not assigned to the submission, but in doing a little bit of related clean-up I'm down a rabbit-hole. I'll post a fix once I dig out.

asmecher:
Pulls:
https://github.com/pkp/omp/pull/208
https://github.com/pkp/pkp-lib/pull/1110

[...]

PR #1110, opened 2016-02-02
pkp/pkp-lib#1106 Fix author grid authorization inconsistency/duplication

asmecher:


controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -227,11 +223,29 @@ function getRequestArgs() {
...
 
 	/**
-	 * Determines if there should be an 'add user' action on this grid.
-	 * Overridden by child grids.
+	 * Determines if there should be add/edit actions on this grid.
+	 * @param $user User
 	 * @return boolean
 	 */
-	function canAdminister() {
+	function canAdminister($user) {
```
beghelli: I think that centralizing here this logic you also fixed a bug that were always adding the ordering feature to ROLE_ID_SUB_EDITOR even if they were not assigned to the submission... Nice.

PR #1132, opened 2016-02-11
fix JSLint complaints

bozana:


asmecher:
Thanks! I'll take care of submodule pushes.

PR #1134, opened 2016-02-12
Add missing Reviewer Form email variables

selwyntcy:


controllers/grid/users/reviewer/form/ReviewerForm.inc.php
```diff
@@ -241,12 +242,16 @@ function fetch($request) {
 			'reviewDueDate' => __('reviewer.submission.reviewDueDate'),
 			'submissionReviewUrl' => __('common.url'),
 			'reviewerUserName' => __('user.username'),
+			'contextName' => $context->getLocalizedName(),
+			'contextUrl' => __('common.url'),
+			'editorialContactSignature' => strip_tags($user->getContactSignature(), "<br>"),
+			'submissionTitle' => strip_tags($submission->getLocalizedTitle()),
```
asmecher: I suspect, since we're now supporting HTML emails, we'll have to run many of our variables through htmlspecialchars. Consider e.g. a `<` in the article title. Article titles shouldn't support HTML (at least not until we have a single-line rich editor).

controllers/grid/users/reviewer/form/ReviewerForm.inc.php
```diff
@@ -241,12 +242,16 @@ function fetch($request) {
 			'reviewDueDate' => __('reviewer.submission.reviewDueDate'),
 			'submissionReviewUrl' => __('common.url'),
 			'reviewerUserName' => __('user.username'),
+			'contextName' => $context->getLocalizedName(),
+			'contextUrl' => __('common.url'),
+			'editorialContactSignature' => strip_tags($user->getContactSignature(), "<br>"),
```
asmecher: For cases where we want to let limited HTML in, use String::stripUnsafeHtml.

controllers/grid/users/reviewer/form/ReviewerForm.inc.php
```diff
@@ -241,12 +242,16 @@ function fetch($request) {
 			'reviewDueDate' => __('reviewer.submission.reviewDueDate'),
 			'submissionReviewUrl' => __('common.url'),
 			'reviewerUserName' => __('user.username'),
+			'contextName' => $context->getLocalizedName(),
```
asmecher: Since some variables are going to be available to all email templates, I suspect it would make sense to put this into a superclass so that we don't have to re-code them for every specific case.

bozana: This one is apparently already in MailTemplate class, s. the function assignParams https://github.com/pkp/pkp-lib/blob/master/classes/mail/MailTemplate.inc.php#L144

controllers/grid/users/reviewer/form/ReviewerForm.inc.php
selwyntcy: I was not familiar with the convention here, so I followed how `SubmissionMailTemplate` handles these variables.

https://github.com/pkp/pkp-lib/blob/cb6fa16ad663618fe2133dc6dec556bb0e8535fa/classes/mail/SubmissionMailTemplate.inc.php#L51-54

controllers/grid/users/reviewer/form/ReviewerForm.inc.php
selwyntcy: ditto

controllers/grid/users/reviewer/form/ReviewerForm.inc.php
selwyntcy: I need some guidance in how I should properly pass these variables along through the `MailTemplate` class.

Take `editorialContactSignature` as an example. In `initData()`, it is assigned to the `SubmissionMailTemplate` object `$template` using `assignParams()`, but the `$template` object is not passed along entirely to the form; only its body is passed as `personalMessage`.
https://github.com/pkp/pkp-lib/blob/cb6fa16ad663618fe2133dc6dec556bb0e8535fa/controllers/grid/users/reviewer/form/ReviewerForm.inc.php#L206-L211

Therefore, during `readInputData()` and `execute()`, there seems to be no way to retrieve `editorialContactSignature`.

Is the following appropriate?
- use `$this->setData()` to supply `$editorialContactSignature` to the form
- add `$editorialContactSignature` as a hidden input to the `reviewerFormFooter.tpl`
- retrieve it back in `readInputData()` to make it availabe to `execute()`

asmecher: I'd suggest having the `emailVariables` describe the contents rather than reiterating them. Think...

> Dear `Recipient Name`,

...rather than...

> Dear `John Doe`,

For the implementation, I don't love static functions, but I also don't love instantiating and populating a MailTemplate object in the `display` function. Perhaps add a static function called e.g. `getEmailVariables` in ArticleMailTemplate, then this class could augment them e.g. something like:

    $templateMgr->assign('emailVariables',
        array_merge(
            ArticleMailTemplate::getEmailVariables(),
            array(
 		$templateMgr->assign('emailVariables', array(
 			'reviewerName' => __('user.name'),
                ...
            )
        )
    );

Does that make sense? Suggestions for improvement are welcome, of course.

Issue #1139, opened 2016-02-15
[OMP] Users offered inaccessible URLs


crism:
From both #1122 and #1127, both against `HEAD`.

#1122: `SubmissionsListGridCellProvider::getUrlByUserRoles()` produces a URL which the user does not have access to, if the submission has moved on. The user should instead get the most recent stage to which they have permission.

#1127: When a review is begun, the author is notified that their submission has changed status. However, the notification includes a link to the workflow (`.../index.php/workflow/access/4`), which the author does not have permission to see (“The current role does not have access to this operation”)

asmecher:
@crism, the attached patch should at least correct the problem of authors getting `workflow` URLs.

Do you have a set of reproduce steps for the other issue, where users can't access stages after the workflow has moved on?

crism:
Interesting… I’ll test f17c3ee here, thanks.

As for reproduction… I’ll double-check, but I think a reviewer with no other role is locked out once the work is in production.

asmecher:
Ah, I see that problem with reviewers. Looking now.

asmecher:
OK, I think these have it covered. Closing for now -- but please test if you get the chance, @crism!

crism:
This looks pretty good to me. Nice work!

Issue #1122, opened 2016-02-05
[OMP] reviewer status is too sticky


crism:
[Using `HEAD` of OMP.] When a user looks at their assignments in the dashboard, if they have ever been a reviewer on a submission, then the action URL is to the reviewer action (_e.g._ `.../reviewer/submission/2`). There is no way (aside from knowing the correct URL) to get to the submission’s actual current state.

crism:
This problem appears to trace to `SubmissionsListGridCellProvider::getCellActions()`, specifically:
```php
$reviewAssignmentDao = DAORegistry::getDAO('ReviewAssignmentDAO');
$reviewAssignment = $reviewAssignmentDao->getLastReviewRoundReviewAssignmentByReviewer($submission->getId(), $user->getId());
if (is_a($reviewAssignment, 'ReviewAssignment')) return array(new LinkAction(...));
```
Either `$reviewAssignment` should be `null` if the review is finished (which seems inappropriate), or there needs to be additional logic to see if the user has some subsequent role to play with respect to this submission.

crism:
Indeed, it looks like someone started down this road… `$userGroupDao` is created at [`controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php:97`](https://github.com/pkp/pkp-lib/blob/master/controllers/grid/submissions/SubmissionsListGridCellProvider.inc.php#L97), but then never used.

crism:
Though I think the thing is to look at the round associated with the review; if the round is over, then use `getUrlByUserRoles()`… I’ll try that and see what it does.

crism:
That resolves the problem for the multi-role reviewer, but now the reviewer-only gets, “You don’t currently have access to that stage of the workflow.” What’s the correct logic to determine the last stage to which a user had access?

asmecher:
I'm not sure if this adequately answers your question, but you should be able to use `$reviewAssignment->getStageId()` to determine what stage the review assignment applies to.

crism:
Ah… yes, I’m going to claim that this is fixed; the resulting problem is related to a problem found in #1127, which I’ll file separately.

asmecher:
Thanks, @crism, that sounds great -- is there a PR attached? I don't see a link here.

PR #1140, opened 2016-02-15
Issue/1122

crism:
Fixes the described problem, in favor of #1139 (which needs fixed anyway).


crism:
This is for #1122—sorry for not mentioning that!

asmecher:
Thanks, @crism -- that looks like a nice, clean solution.

PR #1160, opened 2016-02-18
Fix FBV textarea "required" attribute

selwyntcy:
The `required` attribute is a `boolean attribute`. Even when the `default` switch case renders it as `required="0"`, the text area is still marked as required. So I add a case to add it only when it is `true`.

### DIFF
--- a/classes/form/FormBuilderVocabulary.inc.php
+++ b/classes/form/FormBuilderVocabulary.inc.php
@@ -493,6 +493,7 @@ class FormBuilderVocabulary {
 					}
 					break;
 				case 'id': break; // if we don't do this, the textarea ends up with two id attributes because FBV_id is also set.
+				case 'required': if ($value) $textAreaParams .= $key . ' '; break;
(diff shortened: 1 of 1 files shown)


asmecher:
@selwyntcy, I wasn't able to reproduce the problem; looking e.g. at Settings > Journal > Masthead, none of the fields are marked required. What form were you seeing this behavior on?

selwyntcy:
I came across this in Review Form. There, `Extended text box` items are marked with either `required="0"` or `required="1"` attributes. jquery.validate would complain both as required.
Now I see that for other required fields, like those in Masthead, validations are marked up with something like `validation="required"`.

asmecher:
The problem is that the `getRequired()` function is returning an int or string value, which is not getting interpreted PHP-side as a boolean, but instead getting passed through to markup-land as a string.

I think this same problem applies to other input types, so a broader fix might be to change the `FBV_required` handling in the `smartyFBVElement` function, which is currently...

    $smarty->assign('FBV_required', isset($params['required']) ? $params['required'] : false);

It looks like the handling of that flag is divided messily between the `smartyFBVElement` function and the various `_smartyFBV...` functions it calls; I think some general cleanup is probably the only way to make sure we've got this fixed consistently. Do you want to take a crack at it, @selwyntcy? It's our mess, so I won't mind looking at it if you'd rather no!

PR #1162, opened 2016-02-19
Propose a new Review Form element type Instruction Text

selwyntcy:
Sometimes we may want to add instructions midway in a review form to introduce different sections.

For example:

> **Section A: Please rate the following qualities of the manuscript (5 - excellent ... 1 - poor).**
> 
> 1. Methodology (radio buttons)
> 2. Originality (radio buttons)
> 3. Clarity (radio buttons)
> 
> **Section B: Please give you comments and suggestions.**
> 
> 1. Comments (textarea)
> 2. Suggestions (textarea)

Although the instructions can be put in the question texts of the first questions, it is semantically better to allow a separate element type for pure texts.


asmecher:
Agreed, @selwyntcy, that makes logical sense.

vedranvucic:
I came across the same issue.This may be relevant to sum sub-specialist issues and methodology/instrumentation aspects that may be involved in reviewers task.  Thus, review form item can have space to define task and another to respond to it as radio button, text box etc.

t4x0n:
I support this idea! In our review forms, we need subtitles or sections, so we use something like that: 
image
with repeated information in every line. So, is not a big problem but I hope that idea can see the light in the future :)

PR #1164, opened 2016-02-19
Issue/1149

### LINKED ISSUE #1149: [OMP] error when re-enrolling a user as a reviewer
From #1127, against `HEAD` (but with more detail):

User 2 is a member of group 18 (external reviewers). User 2 participated in the first round of external reviews. In the second round, under “Locate a Reviewer,” User 2 is not listed. When “Enroll Existing User” is selected, User 2 can be found, but on attempting to add him as a reviewer, the notification dialog spins indefinitely, and an error message is logged:
```
DB Error: Duplicate entry '18-2' for key 'user_user_groups_pkey', referer: .../index.php/test/workflow/index/2/5
```
Either the user should be permitted to participate in multiple review rounds (my preference), or should be explicitly disallowed, even through enrollment. In either case, adding a user to a group of which they are already a member should never cause an error; the pre-existence should probably just be taken as a success. (Maybe use `UPDATE IGNORE`?)

crism:
Fixes #1149—but has possible knock-on security or workflow implications, so a review from an experienced eye would be welcome.


bozana:
@crism, thanks!!!
@asmecher, I think this can be merged. Would you also like to take a look?

PR #1182, opened 2016-02-23
fix test for null reply-to address

jnugent:
Test the array element for null instead of the array itself.  This will keep the reply-to header out of the Mail envelope entirely.


asmecher:
Thanks, @jnugent; can you cherry-pick to ojs-stable-2_4_8 and master? And if the SPF stuff gets ported over to OCS, there too.

PR #1189, opened 2016-02-24
Fix review form item reorder

selwyntcy:
It wasn't working due to a mismatched argument name.


asmecher:
Thanks, good spotting!

PR #1190, opened 2016-02-24
Add missing asterisk for required items in review form

selwyntcy:
Copied from https://github.com/pkp/pkp-lib/blob/06d64660e64312ccdb9a13efcb35910f9e8ddbf4/templates/manager/reviewForms/previewReviewForm.tpl#L14

### DIFF
--- a/templates/reviewer/review/reviewFormResponse.tpl
+++ b/templates/reviewer/review/reviewFormResponse.tpl
@@ -18,6 +18,7 @@
 	{/if}
 
 	{fbvFormSection translate=false title=$reviewFormElement->getLocalizedQuestion() list=$list}
+		<p>{$reviewFormElement->getLocalizedQuestion()}{if $reviewFormElement->getRequired()}*{/if}</p>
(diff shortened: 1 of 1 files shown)


asmecher:
@selwyntcy, doesn't this result in presenting both the question and the "required" status twice?

The question is presented [here](https://github.com/pkp/pkp-lib/pull/1190/files#diff-50662552da31709d8bf64717faf84af4L20), and the requied status as an attribute e.g. [here](https://github.com/pkp/pkp-lib/pull/1190/files#diff-50662552da31709d8bf64717faf84af4L22).

selwyntcy:
@asmecher, the "required" status is only shown retrospectively, after an event on that element triggers the validate js. Before the user clicks on a required element and then clicks another element, he has no visual clue to which items are required.

Besides, there is a note `* Denotes required field` at the bottom.

asmecher:
@selwyntcy, ah, I think that's a bug -- the "required" status should always be visible, just like elsewhere (see e.g. the "Profile" form for a few examples). The `* Denotes required field` should be displayed only if there is at least one required field. I'm happy to look into the cause if you like.

PR #1133, opened 2016-02-12
Add missing Review Reminder email variables

selwyntcy:
- remove extra `emailSignature`
- pass missing `reviewDueDate` to the form in `initData()`
- add missing email variables

Corresponding changes to `dateDue` in Review Reminder form template:
- rename to `reviewDueDate` for consistency with other email templates
- change `disabled` attribute to `readonly`, since disabled inputs are not posted in the request
- remove `datepicker` class for consistency with `readonly` attribute

These changes are kept minimal to make the Review Reminder functional.
- Should I also change the other dates (`dateNotified`, `dateConfirmed` and `responseDue`) to `readonly` and remove their `datepicker` class?
- Should these dates be also passed to the form in `initData()`?
- I'm not sure the proper way to pass those email variables to TinyMCE, as there is no `fetch()` implemented here.

PR #1197, opened 2016-02-26
Add missing Review Reminder email variables to both the email templat…

selwyntcy:
…e and TinyMCE

Rebased from PR #1133.

I tried to follow @asmecher and @bozana's suggestions to implement `fetch()` here, following how it is done in `Reviewer Form`, to supply the email variables to TinyMCE, and to modify the few dates in the same way.

I guess I might have copied more than necessary from `Reviewer Form` to here. Please advise.

### DIFF
--- a/controllers/grid/users/reviewer/form/ReviewReminderForm.inc.php
+++ b/controllers/grid/users/reviewer/form/ReviewReminderForm.inc.php
@@ -19,6 +19,9 @@ class ReviewReminderForm extends Form {
 	/** The review assignment associated with the reviewer **/
 	var $_reviewAssignment;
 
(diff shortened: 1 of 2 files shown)


bozana:
Hi @selwyntcy 
Yes, I believe you copied more than necessary. I believe you will only need something like this in the fetch function:
`
$templateMgr->assign('emailVariables', array(
	'dateNotified ' => ... ,
	'dateConfirmed ' => ...,
	'responseDue ' => ...,
));
return parent::fetch($request);
`
If you still have problems, don't worry, you can ignore it and I can then do it after the merge of your other changes... 
THANKS A LOT!!!

PR #1214, opened 2016-03-02
Update admin.xml

alonsoj:
All keys synchronized with en_US and almost all of them translated.


locale/es_ES/admin.xml
```diff
@@ -6,98 +6,196 @@
...
+
+
+
+
+
+
+
+
+
+
```
beghelli: Thanks Alonso. Why so many empty lines?

alonsoj: I used a wrong template for translation.

PR #1221, opened 2016-03-03
Update user.xml

alonsoj:
Modified abbreviations for male and female


mtub:
Thanks @alonsoj and @marcbria .

Issue #1225, opened 2016-03-06
[OMP] add Catalan (ca_ES) locale


mtub:
Add a new locale to OMP: ca_ES (Català). 

Files provided by @marcbria, credits: see https://pkp.sfu.ca/wiki/index.php?title=OMP:_Catalan_(ca_ES)

Update OMP repository and OMP plugins.

mtub:
Phew. Completed for now.

PR #1226, opened 2016-03-06
new locale: ca_ES

mtub:
files provided by @marcbria

pkp/pkp-lib#1225


locale/ca_ES/default.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../dtd/locale.dtd">
+
+<!--
+  * locale/ca_ES/default.xml
+  *
+  * Copyright (c) 2014-2015 Simon Fraser University Library
```
asmecher: Copyright dates need updating.

mtub: Everything is set to the copyright dates in the corresponding en_US locale files. Couldn't we just update all copyright dates in all files after merging these changes?

asmecher: I already did batch changes for the branches we're likely to release. I suspect we won't release anything further from the omp-dev-1_1 branch, and yes, I'll correct the dates there before releasing if we do.

mtub:
@asmecher, this should be good then…

asmecher:
Phew! Thanks, @mtub.

marcbria:
Thanks you both!! ;-)

Alec, this translation could fail with a very few missing chains because it's before the very last version, but we plan to will check it very soon (for OCS and OMP) with the translation module.
But IMHO you don't need to worry is we won't be able to arrive for this release because the translations are perfectly usables.

PR #1253, opened 2016-03-10
set return-path header instead of from if config set

jnugent:
This adjustment requires no change to the application config file and just sets the Return-Path header instead of the From header, which is the header that is actually examined during SPF tests.

### DIFF
--- a/classes/mail/Mail.inc.php
+++ b/classes/mail/Mail.inc.php
@@ -394,7 +394,6 @@ class Mail extends DataObject {
 
 	/**
 	 * Return a string containing the from address.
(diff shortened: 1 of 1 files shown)


asmecher:
This is only a partial resolution for the SPF stuff you've been working on, correct? Most journals won't be using that config option.

jnugent:
That's  correct.  If a journal is not doing any rewriting, then things work as they have always been (From unaltered, no return-path override).  

There will be other PRs to fix some encoding issues that Rune Hansen at CD has noticed as well.

PR #1282, opened 2016-03-17
pkp/pkp-lib#975 small layout fix for submit in my role as section

bozana:
S. https://github.com/pkp/pkp-lib/issues/975#issuecomment-197970187


asmecher:
Merged, thanks!

PR #1322, opened 2016-03-30
pkp/pkp-lib#986: Implement non orderable items warning for grids

beghelli:


classes/controllers/grid/feature/OrderItemsFeature.inc.php
```diff
@@ -21,13 +21,23 @@ class OrderItemsFeature extends GridFeature{
...
 	 * Constructor.
+	 * @param $overrideRowTemplate boolean This feature uses row
+	 * actions and it will force the usage of the gridRow.tpl.
+	 * If you want to use a different grid row template file, set this flag to
+	 * false and make sure to use a template file that adds row actions.
+	 * @param $nonOrderableItemMessage string A translated message to be used
+	 * when user tries to move a non orderable grid item.
 	 */
-	function OrderItemsFeature($overrideRowTemplate) {
+	function OrderItemsFeature($overrideRowTemplate, $nonOrderableItemMessage) {
```
asmecher: Should this perhaps have a default for `$nonOrderableItemsMessage`? Most grids using `OrderItemsFeature` won't have non-orderable items, I would expect, so shouldn't be forced to specify this.

beghelli: yes, we can use it, but the default right now is null, but it's defined by the child class, which is wrong. I will move it to here. Thanks

Issue #1155, opened 2016-02-16
[OMP] half-entered production discussions look weird to other users


crism:
From #1127, against `HEAD`:

When one user starts a Production Discussion, other users will see an odd, half-baked entry, until the first user finishes submitting it. It seems like the item should be invisible to other users until submission is complete. (The discussion is visible, with a name of just a dash and no other metadata. Presumably, a stub is made in the database; it should either be deferred, or marked as incomplete and skipped by other users.)

crism:
Ooh, fun. Try this, from two computers, with two logins:
 1. From computer A, logged in as user A, click Add discussion.
 1. From computer B, logged in as user B, refresh the relevant page (say, external review).
 1. From computer B, find the stub discussion. Select the expand arrow.
 1. From computer B, delete the discussion.
 1. From computer A, fill in the subject and message for the new (and now deleted) discussion.
 1. From computer A, save the discussion.

Sadly, no fireworks, but computer A gets `##user.authorization.invalidQuery##` message.

In any case, this demonstrates a potentially serious race condition, especially if any users are hostile, cranky, or just curious, making this more than just a UI curiosity.

crism:
This is not something inherent to grids; file and reviewer grids do not write to the database upon starting to add items. This appears to be unique to discussions (via `QueriesGridHandler`).

crism:
The problem appears to stem from the `Que,ryForm` constructor, which calls `QueryDAO::insertObject()` at line 57 and `NoteDAO::insertObject()` at line 70.

Either this needs not to happen (and I can see a bunch of reasons why that might not work), or others’ views need to filter out transient stubs.

asmecher:
Filtering out the stubs seems like a logical solution.

crism:
That pull request can be acted on, but doesn’t quite close the issue—there remains the problem that, in a discussion with a non-null message, any other null messages are then displayed within the discussion widget.

asmecher:
@crism, I'm not sure if I'm getting the issue right, but the goal here is to hide stub queries, and the PR looks like it'll do that. We don't use stubs for comments except the initial one, so I don't think we'll see issues beyond that. Unless I'm missing a case?

crism:
Yeah, @asmecher—if user A starts to add a message to an existing conversation, and user B looks at the conversation before user A saves changes, user B will see a null message from user A in the thread. That should likewise be suppressed.

asmecher:
[...]

PR #1331, opened 2016-04-01
Issue/1155

crism:
The first commit (e151f10) should be uncontroversial; it hides stub messages from conversations (pkp/pkp-lib#1155). It hides them from _all_ users, even if the user who created the stub. This potentially clutters the database if it happens a lot.

The second commit (8f63b58) allows admin users to see the stubs (the better to delete them with). It involves changes to more classes and methods, and might be objectionable…

### DIFF
--- a/classes/note/NoteDAO.inc.php
+++ b/classes/note/NoteDAO.inc.php
@@ -76,7 +76,7 @@ class NoteDAO extends DAO {
 	 * @param $sortDirection int Optional sorting order constant: SORT_DIRECTION_...
 	 * @return object DAOResultFactory containing matching Note objects
 	 */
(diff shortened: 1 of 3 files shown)


asmecher:
Personally I don't see the harm in letting the occasional stub comment hang around, but likewise I don't see the harm in allowing someone to clean them up if they see fit and if the chances of a collision (cleaning up while someone composes) are low. An alternate approach would be to have a periodic task that deletes null comments that are older than a certain age -- that could also work nicely for stub queries.

However, I would suggest changing `isAdmin` to `canManage` or similar. Administrator has a particular meaning and would suggest to me that this flag says the user is a site administrator.

asmecher:
Merged; thanks, @crism.

PR #1345, opened 2016-04-06
fix warning if there are no optional columns in the usage stats form

bozana:
PHP Warning:  in_array() expects parameter 2 to be array, null given


asmecher:
Merged, thanks!

Issue #1381, opened 2016-04-14
Calling mkdirtree when open_basedir prevents access to a folder triggers infinite recursion


pmarrone:
The code in classes/file/FileManager.inc.php mkdirtree relies on file_exists() to end the recursive exploration of the directory structure.

```
	function mkdirtree($dirPath, $perms = null) {
		if (!file_exists($dirPath)) {
			if ($this->mkdirtree(dirname($dirPath), $perms)) {
				return $this->mkdir($dirPath, $perms);
			} else {
				return false;
			}
		}
		return true;
	}
```

If the open_basedir directive is set (or also, theoretically, safe mode) and the current directory in mkdirtree actually exists but is outside the accessible directories, file_exists returns false and the recursion goes all the way to the root directory ('/' in my case).

As the parent directory (`dirname($dirpath)`) of '/' is also '/', the recursion goes until the script stops (I thought it should have thrown a stack overflow error, but this managed to produce over 2GB of logs and kept running at max CPU usage).

Adding error checking for this allows for better error recovery and cleaner logs.

I'll try to send a Pull Request with this fix.

asmecher:
Thanks, @pmarrone, I've merged and cherry-picked this to the relevant branches. Closing!

PR #1382, opened 2016-04-14
pkp/pkp-lib#1381 Prevent mkdirtree infinite recursion

pmarrone:
Prevent infinite recursion when mkdirtree is called with a path where its first existing directory is outside those allowed by the open_basedir directive.

This is done by detecting when mkdirtree is about to make a recursive call to itself with the same parameters. That is, when the current given path is its own parent, which happens with the root directory ('/').

### DIFF
--- a/classes/file/FileManager.inc.php
+++ b/classes/file/FileManager.inc.php
@@ -348,7 +348,11 @@ class FileManager {
 	 */
 	function mkdirtree($dirPath, $perms = null) {
 		if (!file_exists($dirPath)) {
-			if ($this->mkdirtree(dirname($dirPath), $perms)) {
+			//Avoid infinite recursion when file_exists reports false for root directory
+			if ($dirPath == dirname ( $dirPath )) {
+				trigger_error ( 'There are no readable files in this directory tree. Is safe mode or open_basedir active?', E_USER_ERROR );
+				return false;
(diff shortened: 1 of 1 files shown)


classes/file/FileManager.inc.php
```diff
@@ -348,7 +348,11 @@ function rmtree($file) {
 	 */
 	function mkdirtree($dirPath, $perms = null) {
 		if (!file_exists($dirPath)) {
-			if ($this->mkdirtree(dirname($dirPath), $perms)) {
+			//Avoid infinite recursion when file_exists reports false for root directory
+			if ($dirPath == dirname ( $dirPath )) {
+				trigger_error ( 'There are no readable files in this directory tree. Is safe mode or open_basedir active?', E_USER_ERROR );
```
asmecher: Thanks for contributing, @pmarrone! Two minor changes to suggest for this line:
- Could you change `trigger_error` to `fatalError` for consistency with the rest of the codebase?
- Just a typo, and sorry for being pedantic: "Is safe mode or open_basedir active?" should be "_Are_ safe mode or open_basedir active?"

pmarrone: No problem. I changed trigger_error to fatalError and 'Is' to 'Are' in the next commit.

Issue #1391, opened 2016-04-18
Allow protocol-relative URLs in compiled CSS


asmecher:
Otherwise the protocol (e.g. http://) gets compiled into the CSS, and requests via a different protocol (e.g. https://) can cause warnings.

asmecher:
PR: https://github.com/pkp/pkp-lib/pull/1392

@jmacgreg, this causes cached CSS to use protocol-relative URLs.

I checked the Google CDN URLs and the OMP 1.2 release already supports protocol-relative URLs. If you plan to upgrade to OMP 1.2, it'll be fixed there; if it'll take a while, you can just disable CDNs in `config.inc.php` to cause OMP to use local copies (which will be fetched according to the user's protocol).

I've assigned the CSS PR to @beghelli, so once he's reviewed it I can commit it around and, if you plan to take awhile upgrading to 1.2, back-port to OMP 1.1.1.

jmacgreg:
Hi Alec, thanks - I've currently modified some cache files (not a permanent solution, obvs), but will look at the CDN flag as well. They'll be upgrading to 1.2 shortly though, so rather than forget that we enabled that, I may just let this be. Thanks!

beghelli:
@asmecher  oups, didn't meant to close, just merged the pull request.

asmecher:
Actually, that's fine with me! Thanks, @beghelli.

PR #1392, opened 2016-04-18
pkp/pkp-lib#1391 Permit protocol-relative URLs in compiled CSS

asmecher:
@beghelli, do you mind reviewing this one?

### DIFF
--- a/classes/core/PKPRequest.inc.php
+++ b/classes/core/PKPRequest.inc.php
@@ -42,9 +42,6 @@ class PKPRequest {
 	/** @var string server host */
 	var $_serverHost;
 
-	/** @var string base url */
-	var $_baseUrl;
-
 	/** @var string request protocol */
 	var $_protocol;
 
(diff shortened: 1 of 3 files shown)


classes/core/PKPRequest.inc.php
```diff
@@ -154,24 +151,26 @@ function getIfModifiedSince() {
...
 	/**
 	 * Get the base URL of the request (excluding script).
+	 * @param $allowProtocolRelative boolean True iff protocol-relative URLs are allowed
 	 * @return string
 	 */
-	function getBaseUrl() {
+	function getBaseUrl($allowProtocolRelative = false) {
 		$_this =& PKPRequest::_checkThis();
 
-		if (!isset($_this->_baseUrl)) {
```
beghelli: we are not saving the base url in the object anymore because the same request might need protocol relative equals true and equals false?

asmecher: Yes, that's correct. Practically speaking I don't think re-executing this code every time (as opposed to caching) will have a performance impact.

Issue #1426, opened 2016-05-04
$_SERVER[HOSTNAME] should not be used for host identification


ctgraham:
In OJS 2.4.x, `$_SERVER['HOSTNAME']` is considered an option for host identification if `$_SERVER['HTTP_X_FORWARDED_HOST']` and `$_SERVER['HTTP_HOST']` are unavailable.
https://github.com/pkp/pkp-lib/blob/ojs-stable-2_4_8/classes/core/PKPRequest.inc.php#L338

Are there any know web engines where the actual HOSTNAME would be legitimate?  The HTTP_\* variables make sense, but if they are absent it probably means the script is running CLI and it is highly unlikely that the server's actual hostname would bear any resemblance to the virtual hosts served by the web engine.

_My scenario:_
A virtual machine (ojs-prod-04) serving multiple virtualhosts.  No `base_url[]` overrides in config.inc.php.  Cron execution of runScheduledTasks.php.

_Result:_
Cron's call to PKPRequest::getServerHost() returns "ojs-prod-04" instead of the actual virtualhost as configured in Apache and in config.inc.php's `base_url`.

Issue #1423, opened 2016-05-02
PKP PLN Network Status is confused by URL aliases


ctgraham:
@ubermichael, see [my question](http://forum.pkp.sfu.ca/t/pkp-pln-status-question/10932/15?u=ctgraham) and [@mjordan's reply](http://forum.pkp.sfu.ca/t/pkp-pln-status-question/10932/16?u=ctgraham) regarding a journal which has multiple legitimate DNS names.

A problem seems to arise in the use of `$journal->getUrl()` instead of `$base_url`.

ctgraham:
If I'm understanding the concern correctly, we did something similar with usageEvents:
https://github.com/ulsdevteam/ojs/blob/uls-install/plugins/generic/usageEvent/UsageEventPlugin.inc.php#L289-L302

Perhaps this kind of "canonicalUrl" disambiguation should be promoted to the `PKPRequest` object?

asmecher:
Grepping through for uses of `$journal->getUrl`, pretty much all of those use cases should be replaced with the more consistent and dependable `Request::url`. Once that's done, I'd suggest deprecating or removing `$journal->getUrl` entirely.

ctgraham:
Replacing usages of $journal->getUrl() with Request::url() doesn't resolve the disambiguation of multiple base_urls, does it?

asmecher:
Using `Request::url` should cause OJS to forcibly generate URLs per `base_url[...]` overrides, regardless of where the user is coming from.

ctgraham:
Something is not quite right there.  In dev-2_4, `$journal->getUrl()` is just `return Request::url($this->getPath());`  So, if it is nomalizing on base_url or base_url[], the assumption of the issue as reported above is wrong.

ctgraham:
Found it.  Request::url() honors the base_url[] overrides, but ignores the base_url.  In my case, I have just a single journal install, and only base_url is set, no base_url[] overrides.  Apache, however, will direct 5 different DNS entries to this install.

asmecher:
When `base_url[...]` isn't supplied (that's different than `base_url` without `[...]`), OJS will respond according to whatever incoming URL it's invoked from. If you supply `base_url[...]` directives, then that'll force OJS to use whatever is configured as canonical.

ctgraham:
That makes sense to me.

I don't think, however, that as a single journal install I should need to enter a base_url[] override for the PKP PLN to work.  :smile:

ubermichael:
[...]

PR #1427, opened 2016-05-04
pkp/pkp-lib#1426: use only HTTP_ prefixed variables for server hostname detection.

ctgraham:
`$_SERVER['HOSTNAME']` is unlikely to return expected results in modern architectures.

This resolves https://github.com/pkp/pkp-lib/issues/1426 and helps to resolve https://github.com/pkp/pkp-lib/issues/1423, at the expense of anyone who is really still running an install on a dedicated physical server with a physical hostname matching the OJS install's DNS name, but not matching the config.inc.php base_url.

### DIFF
--- a/tests/classes/core/PKPRequestTest.php
+++ b/tests/classes/core/PKPRequestTest.php
@@ -254,6 +254,18 @@ class PKPRequestTest extends PKPTestCase {
 		$_SERVER = array(
 			'HOSTNAME' => 'hostname'
 		);
+		self::assertEquals('localhost', $this->request->getServerHost());
+	}
+
+	/**
+	 * @covers PKPRequest::getServerHost
+	 * @depends testGetServerHostLocalhost
+	 */
+	public function testGetServerHostWithServerName() {
+		// if HOSTNAME is set then return it
+		$_SERVER = array(
+			'SERVER_NAME' => 'hostname'
+		);
 		self::assertEquals('hostname', $this->request->getServerHost());
 	}
 
--- a/classes/core/PKPRequest.inc.php
+++ b/classes/core/PKPRequest.inc.php
@@ -335,7 +335,7 @@ class PKPRequest {
 		if (!isset($_this->_serverHost)) {
 			$_this->_serverHost = isset($_SERVER['HTTP_X_FORWARDED_HOST']) ? $_SERVER['HTTP_X_FORWARDED_HOST']
 				: (isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST']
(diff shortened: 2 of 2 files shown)


tests/classes/core/PKPRequestTest.php
```diff
@@ -254,7 +254,7 @@ public function testGetServerHostWithHostname() {
 		$_SERVER = array(
```
asmecher: It looks like this assignment could be removed, since we're no longer using `HOSTNAME`.

asmecher:
I take your point regarding the hostname being usually irrelevant, but it is intended as a less-than-ideal fallback for when the normal stuff isn't available. By my reading, falling through to nothing rather than falling through to `HOSTNAME` is guaranteed to be bad, where falling through to `HOSTNAME` was usually bad. Does this solve a problematic case, or just remove a potential source of confusion? In both cases the resolution is to add the missing configuration directives, is it not?

ctgraham:
Currently, if config.inc.php only has a base_url defined, and no base_url[] overrides defined (as is the case for our ULS journals) the existing logic in [`PKPRouter::_urlGetBaseAndContext()`](https://github.com/pkp/pkp-lib/blob/ojs-stable-2_4_8/classes/core/PKPRouter.inc.php#L430) falls through to `PKPRouter::getIndexUrl()`, which calls `PKPRequest::getBaseUrl()`, which calls `PKPRequest::getServerHost()`, which works if the request is being served by Apache, but gives unwanted results if the request is a CLI call.

`PKPRequest::getBaseUrl()` [expects to handle this case](https://github.com/pkp/pkp-lib/blob/ojs-stable-2_4_8/classes/core/PKPRequest.inc.php#L153), but it never gets the opportunity.  So, now that I'm typing this, perhaps the better solution is to prioritize the check for a CLI call in `PKPRequest::getBaseUrl()`.

tests/classes/core/PKPRequestTest.php
ctgraham: I think the point of this test would now be to ensure that we *aren't* picking up the (provided) HOSTNAME from the `$_SERVER` environment.

ctgraham:
fd0a06e is just in testing and is *not* ready for merging.

ctgraham:
This code has been running in our environment and has resolved the problems we experienced with running the PLN deposits and CrossRef exports via cron.

asmecher:
All right -- looks good, let's give this a crack -- though I'm a little nervous about outlying webserver environments.

asmecher:
Merged and cherry-picked over to the master branch.

Issue #1507, opened 2016-06-02
RECAPTCHA breaks on SSL if SSL is optional


ctgraham:
In OJS 2.x, if a site is viewed over HTTPS, but config.inc.php's `force_ssl` is unset, RECAPTCHA is served over HTTP instead of HTTPS, and any sane browser will block the component, making validation impossible.

See:
https://github.com/pkp/ojs/blob/397061d442f66cd788f4b1d50b16be000e769a0c/classes/user/form/RegistrationForm.inc.php#L114-L115
and
https://github.com/pkp/ojs/blob/dbe779f2bd97e9a0f2842c4357fb56496b7179e4/classes/comment/form/CommentForm.inc.php#L131-L132
and
https://github.com/pkp/pkp-lib/blob/a39b1328db9ec06d2902e67b337dc5a63c60a496/classes/notification/form/NotificationMailingListForm.inc.php#L76-L77

This check should not be based on `Config::getVar('security', 'force_ssl')`, but on whether or not the Request itself is HTTPS.

asmecher:
@ctgraham, I've merged both PRs, thanks -- does this fully resolve the issue?

ctgraham:
Yes, to my knowledge.

asmecher:
Great -- thanks as always!

ctgraham:
Can this be cherry-picked to stable-2_4_8 for 2.4.8-1?

asmecher:
Done, thanks!

PR #1512, opened 2016-06-03
pkp/pkp-lib#1507: Use SSL for RECAPTCHA if protocol is HTTPS or if config requires SSL

ctgraham:
Partially resolves https://github.com/pkp/pkp-lib/issues/1507

Allows backwards compatibility in the unlikely case of SSL being forced, but the request coming in over HTTP (a proxy perhaps?) but primarily ensures if the Request is HTTPS, then SSL should be used.


asmecher:
Merged -- thanks, @ctgraham.

PR #1518, opened 2016-06-10
Use protocol-relative URLs for CC license badges

selwyntcy:


asmecher:
Good call, @selwyntcy, thanks!

PR #1521, opened 2016-06-13
Made fuctions in PKPOAIDAO abstract.

WaDelma:
These functions where missed in 76b00bdb15b97d322b34f719c58ad75e97511679


WaDelma:
I see that there are other cases where `assert(false)` is used inside functions. These should be evaluated if they can be updated to be abstract.

asmecher:
Merged, thanks; these are cosmetic so it's not a high priority to grep and replace, but I'm happy to review and merge.

PR #1523, opened 2016-06-14
Refactor temporary file moving in PluginHelper.

WaDelma:
This refactors temporary file moving to it's own function and improves temporary file deletion.

### DIFF
--- a/classes/plugins/PluginHelper.inc.php
+++ b/classes/plugins/PluginHelper.inc.php
@@ -109,23 +109,7 @@ class PluginHelper {
 
 			// Copy the plug-in from the temporary folder to the
 			// target folder.
-			// Start with the library part (if any).
-			$libPath = $path . '/lib';
-			$fileManager = new FileManager();
-			if (is_dir($libPath)) {
-				if(!$fileManager->copyDir($libPath, $pluginLibDest)) {
-					$errorMsg = __('manager.plugins.copyError');
-					return null;
-				}
-				// Remove the library part of the temporary folder.
-				$fileManager->rmtree($libPath);
-			}
-
-			// Continue with the application-specific part (mandatory).
-			if (!$fileManager->copyDir($path, $pluginDest)) {
-				$errorMsg = __('manager.plugins.copyError');
-				return null;
-			}
+			_cutAndPasteTemporaryFile($path, $pluginDest, $pluginLibDest, $errorMsg);
 
 			// Remove the temporary folder.
 			$fileManager->rmtree(dirname($path));
@@ -240,25 +224,7 @@ class PluginHelper {
 
 			// Copy the plug-in from the temporary folder to the
 			// target folder.
-			// Start with the library part (if any).
-			$libPath = $path . '/lib';
-			if (is_dir($libPath)) {
-				if(!$fileManager->copyDir($libPath, $pluginLibDest)) {
-					$errorMsg = __('manager.plugins.copyError');
-					return null;
-				}
-				// Remove the library part of the temporary folder.
-				$fileManager->rmtree($libPath);
-			}
-
-			// Continue with the application-specific part (mandatory).
-			if(!$fileManager->copyDir($path, $pluginDest)) {
-				$errorMsg = __('manager.plugins.copyError');
-				return null;
-			}
-
-			// Remove the temporary folder.
-			$fileManager->rmtree(dirname($path));
+			_cutAndPasteTemporaryFile($path, $pluginDest, $pluginLibDest, $errorMsg, $fileManager);
 
 			$upgradeFile = $pluginDest . '/' . PLUGIN_UPGRADE_FILE;
 			if($fileManager->fileExists($upgradeFile)) {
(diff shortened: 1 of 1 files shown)


asmecher:
@WaDelma, this definitely won't work yet -- the `_cutAndPasteTemporaryFile` function is declared within the class but doesn't have `$this->` or `self::` prefixes when called. When you get your development system, give it a crack and let me know.

asmecher:
Thanks, @WaDelma -- have you had a chance to test this yet?

WaDelma:
Well I have been able to the tests that there currently are, but I am not sure if they would catch bugs in this... The testing story for OJS and pkp-lib is bit hazy for me.

WaDelma:
I am currently trying to test this. Is there any test plugins anywhere I can test this on?    
I found a [page](https://pkp.sfu.ca/wiki/index.php?title=Reference_Plugins) for reference plugins, but there doesn't seem to be one that is packaged right for "Upload A New Plugin".    
And there doesn't seem to be any info how to do the packaging. I tried to just upload zip file you can download from github, but that complains:
> The uploaded plugin archive does not contain a folder that corresponds to the plugin name.

asmecher:
You could try the tarball release of the Hypothes.is plugin, available here: https://github.com/asmecher/hypothesis/releases

This is how plugins should be packaged.

WaDelma:
I tried that plugin and it worked.
But I found bugs that exists without my modifications:
  - If I try to upgrade that plugin with itself it complains:
 >The uploaded plugin does not fit the category of the upgraded plugin.

    and if you try to upgrade again with same file selected it says:
> The uploaded plugin archive does not contain a folder that corresponds to the plugin name. 

  - It doesn't seem like it removes temporary files after it's done.
  - When I removed the plugin it says:

 >Plugin "##plugins.generic.hypothesis.name##" successfuly deleted

WaDelma:
This should be ready. Or should I try to tackle problems I mentioned last comment?

WaDelma:
Is there still something blocking this from being merged/cherrypicked?

WaDelma:
I found the cause of:
> The uploaded plugin does not fit the category of the upgraded plugin.

When you click upgrade on plugin list it sends GET:
> http://example.com/index.php/sad/$$$call$$$/grid/settings/plugins/settings-plugin-grid/upgrade-plugin?plugin=hypothesisplugin&category=generic

But when you actually click save to upgrade it sends GET:
> http://example.com/index.php/sad/$$$call$$$/grid/settings/plugins/settings-plugin-grid/save-upload-plugin?function=upgrade

The execute method of UploadPluginForm tries to getUserVar for both plugin and category which return empty because they are not present.

WaDelma:
Is there still something blocking this from being merged?

asmecher:
Simply that I haven't had time to work with it properly -- sorry, @WaDelma! @bozana, do you mind if I give this to you?

classes/plugins/PluginHelper.inc.php
```diff
@@ -103,13 +106,14 @@ function installPlugin($path, &$errorMsg) {
...
 
+		$fileManager = new FileManager();
 		if (!$installedPlugin) {
 			$pluginLibDest = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/' . strtr($pluginVersion->getProductType(), '.', '/') . '/' . $pluginVersion->getProduct();
 			$pluginDest = Core::getBaseDir() . '/' . strtr($pluginVersion->getProductType(), '.', '/') . '/' . $pluginVersion->getProduct();
 
 			// Copy the plug-in from the temporary folder to the
 			// target folder.
-			$this->_cutAndPasteTemporaryFile($path, $pluginDest, $pluginLibDest, $errorMsg);
+			$this->_cutAndPasteTemporaryFile($path, $pluginDest, $pluginLibDest, $errorMsg, $fileManager);
```
bozana: Shouldn't here and at the other place where this function is called the `return null` be considered e.g. in case there was an error when copying the file, like in the original version?

classes/plugins/PluginHelper.inc.php
```diff
@@ -256,22 +267,30 @@ function _cutAndPasteTemporaryFile($path, $pluginDest, $pluginLibDest, &$errorMs
...
+	}
 
-		// Remove the temporary folder.
-		$fileManager->rmtree(dirname($path));
+	/**
+	 * Cuts and pastes source to target.
+	 * @param $manager The file manager to use
+	 * @param $src The source directory
+	 * @param $trg The target directory
+	 * @return Was it successful
```
bozana: The types are missing for `@param` and `@return` variables -- e.g. `@param $manager FileManager`. Also for the function `_cutAndPasteTemporaryFile` above.

bozana:
@WaDelma, I had a first look at your changes and have a few questions/comments. I would also like to test it, but I because I am traveling I will be able to do that first next week. Sorry for the delay! Thanks a lot!

bozana:
@WaDelma, I tested it and everything is fine. Thus, if you could just consider those two comments above, so I can then merge it, i.e.
-- To consider the return value for the function "_cutAndPasteTemporaryFile" i.e. where it is called and to return null if the copy error occurred, because else, the function will continue with the processing...
-- To adjust the functions parameter description i.e. to add the variable type.
I also fixed the upgrade problem that you mentioned above and will then add that fix once your changes are merged. 
Thanks a lot!

WaDelma:
Ah yeah. This thing. Completely forgot you left those comments. Sorry for that.

I'll do the thing now.

bozana:
Thanks a lot @WaDelma!

asmecher:
Thanks to you both -- the delay in resolving this was my fault!

PR #1551, opened 2016-06-23
Only allow to have recaptcha only on HTTPS

rcaap-fccn:
To prevent the problem of using a force_ssl off and ssl on sign in on.
When this occurs, the sign on over ssl, the user will have a ssl session and the recaptcha will be requested using http protocol instead of https. If the recaptcha is always required in https, then the issue disapears.


ctgraham:
Better would be to forward port https://github.com/pkp/pkp-lib/issues/1507

This avoids modifying the library code.

PR #1563, opened 2016-06-27
default to sending mail from User in MailTemplate

jnugent:
This fixes up the default From header in the MailTemplate class in pkp-lib, so mail appears is sent from the logged in user first, before defaulting to site and then context.

the send() method already included a test that called getEnvelopeSender, which tests to see if envelope forcing is turned on before setting the Sender property of the mailer, so there was no need to adjust that.  OMP and OJS master both include force_default_envelope_sender options already.

### DIFF
--- a/classes/mail/MailTemplate.inc.php
+++ b/classes/mail/MailTemplate.inc.php
@@ -93,9 +93,8 @@ class MailTemplate extends Mail {
 
 		// Default "From" to user if available, otherwise site/context principal contact
 		if ($user) {
-			$this->setReplyTo($user->getEmail(), $user->getFullName());
-		}
-		if (!$context) {
+			$this->setFrom($user->getEmail(), $user->getFullName());
+		} elseif (is_null($context) || is_null($context->getSetting('contactEmail'))) {
 			$site = $request->getSite();
 			$this->setFrom($site->getLocalizedContactEmail(), $site->getLocalizedContactName());
 		} else {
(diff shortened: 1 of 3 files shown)


jnugent:
@asmecher I think this solves it.  Default to an empty email header, but pull the context one if the mail template deems it appropriate.

asmecher:
I'm wondering if it wouldn't be better to remove the email header entirely; it was added for the sake of the SPF stuff. Do you foresee a use for it with that changed?

asmecher:
Ah, I see you've already described it here: https://github.com/pkp/ojs/pull/832#issuecomment-229141693

The email header gets pretty ambiguous this way; it's still described this way:
> The prepared emails that are sent by the system on behalf of the journal will begin with the following header. These emails will be addressed from the Principal Contact, so it's important to clarify that the Primary Contact is not necessarily responsible for the message content, which may be sent on behalf of a different user.

If it's only going to be relevant for a few emails, I still think it's better to remove it, and then perhaps to work on those specific emails (maybe just clarifying the text if they don't come from the user one might expect).

jnugent:
@asmecher I've updated thisPR with commits that completely remove the emailHeader entirely.  

OMP would also have this that needs to go in:  https://github.com/pkp/omp/pull/299/commits/bf71ca6651eef9228bc4eaf29f31afcdfa00566d

OJS:  https://github.com/pkp/ojs/pull/900

Issue #1568, opened 2016-06-28
enable the display of the xml validation messages


bozana:
This will be used in import/export plugins.

asmecher:
Merged, thanks!

PR #1569, opened 2016-06-28
pkp/pkp-lib#1568 enable the display of xml validation messages

bozana:
S. https://github.com/pkp/pkp-lib/issues/1568

### DIFF
--- a/classes/filter/Filter.inc.php
+++ b/classes/filter/Filter.inc.php
@@ -416,14 +416,14 @@ class Filter extends DataObject {
 	 * NB: sub-classes will not normally override
 	 * this method.
 	 *
-	 * @param mixed an input value that is supported
+	 * @param $input mixed an input value that is supported
 	 *  by this filter
-	 * @displayErrors boolean weather the value
+	 * @param $returnErrors boolean weather the value
 	 *  should be returned also if an error occurred
 	 * @return mixed a valid return value or null
 	 *  if an error occurred during processing
 	 */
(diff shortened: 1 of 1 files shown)


classes/filter/Filter.inc.php
```diff
@@ -418,10 +418,12 @@ function isCompatibleWithRuntimeEnvironment() {
 	 *
 	 * @param mixed an input value that is supported
 	 *  by this filter
+	 * @displayErrors boolean weather the value
```
asmecher: Syntax should be

    @param $displayErrors boolean ...description...

(The other param is also described incorrectly.)

Also, maybe better to call it `returnErrors` rather than `displayErrors`.

bozana:
Thanks @asmecher, so OK?

classes/filter/Filter.inc.php
```diff
@@ -416,12 +416,14 @@ function isCompatibleWithRuntimeEnvironment() {
 	 * NB: sub-classes will not normally override
 	 * this method.
 	 *
-	 * @param mixed an input value that is supported
+	 * @param $input mixed an input value that is supported
 	 *  by this filter
+	 * @param $returnErrors boolean weather the value
```
asmecher: `weather` should be `whether`, but now I'm just getting picky :)

bozana:
Hihihihi :-) It's a pity my editor doesn't have language check! :-P Now OK?

PR #1592, opened 2016-07-09
removed duplicate key in locale files

mtub:
`user.login.activated` was a duplicate in user.xml, removed from all locales that had inherited the duplication


asmecher:
Merged, thanks!

Issue #1596, opened 2016-07-11
Mail::getReplyToString() is noisy


ctgraham:
In OJS 2.4.x, the "replyTo" property is likely null:
https://github.com/pkp/pkp-lib/blob/ojs-stable-2_4_8/classes/mail/Mail.inc.php#L338
But we treat is as an array when checking it for getReplyToString()
https://github.com/pkp/pkp-lib/blob/ojs-stable-2_4_8/classes/mail/Mail.inc.php#L356

This throws as warning.

PR #1597, opened 2016-07-11
pkp/pkp-lib#1596: Check if $replyTo is array before passing it to array_key_exists

ctgraham:
Resolves pkp/pkp-lib#1596.


asmecher:
Cherry-picked also to `ojs-stable-2_4_8`. Thanks!

Issue #383, opened 2015-03-05
Implement/test Crossref support for OJS 3.0


asmecher:

AhemNason:
Happy to help here, Alec.

PR #1598, opened 2016-07-11
pkp/pkp-lib#383 submit button layout

bozana:
CSS for the primary submit link action in crossref export plugin


asmecher:
Merged, thanks!

Issue #1527, opened 2016-06-17
Public identifiers for OMP


bozana:
Migrate pub ids changes from OJS master to OMP master, s. https://github.com/pkp/pkp-lib/issues/1457

- [x] a) add "identifiers" tab to the metadata/edit modal of each publication object (monograph, publication format, file)
- [x] b) introduce "publisher-id" (custom URL identifier) field, add it to the new "identifiers" tab
- [x] c) consider "publisher-id" (custom URL identifier) in the reader interface, s. also issue 1074
- [x] d) implement DOI plugin 
- [x] e) implement URN plugin
- [x] f) implement a different pub ids assignment concept -- the pub ids shouldn't be assigned automatically as till now, but by the editor within the publishing process: 
- on the "identifiers" tab,
- publication formats tab in the catalog-entry modal
- when approving a publication format and file 
- [x] g) UI/UX improvements, s. below

UI/UX:
- Display what is missing in the pub id suffix, why it cannot be assigned.
- Use something like pkp_button_link_offset (like Cancel buttons in modals) for "Clear DOI"/"Clear URN"
- Make "Clear Issue Objects DOIs"/"Clear Issue Objects DOIs" as buttons, something like pkp_button_offset
- Add headings "URN" and "DOI" for the pub id assignment parts on the issue publishing confirmation modal, when publishing an issue.

bozana:
PRs:
pkp-lib: https://github.com/pkp/pkp-lib/pull/1570
omp: https://github.com/pkp/omp/pull/300
ojs: https://github.com/pkp/ojs/pull/897

asmecher:
Thanks, the 3 PRs in https://github.com/pkp/pkp-lib/issues/1527#issuecomment-229285757 look good -- just a couple trivial typos. Do these need to be merged in coordination with the CrossRef stuff?

bozana:
Thanks @asmecher, yes it would be better to merge them at the same time or first, but... also not a very big deal/problem if not :-)

PR #1609, opened 2016-07-13
pkp/pkp-lib#1527 public url identifiers

bozana:
Public URL identifiers for OMP, s. c) here: https://github.com/pkp/pkp-lib/issues/1527
and a bug fix for pluck_files.

This part in the PKP LIB is only for files, and it could be used for OJS too. -- Will have to check/implement it there too.

### DIFF
--- a/classes/submission/PKPSubmissionFileDAO.inc.php
+++ b/classes/submission/PKPSubmissionFileDAO.inc.php
@@ -72,10 +72,10 @@ abstract class PKPSubmissionFileDAO extends PKPFileDAO implements PKPPubIdPlugin
 	 * @param $settingName string
 	 * @param $settingValue mixed
 	 * @param $submissionId int optional
-	 * @param $pressId int optional
+	 * @param $contextId int optional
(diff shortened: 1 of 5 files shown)


classes/submission/PKPSubmissionFileDAO.inc.php
```diff
@@ -67,6 +67,90 @@ function getRevision($fileId, $revision, $fileStage = null, $submissionId = null
 		return $this->_checkAndReturnRevision($revisions);
 	}
 
+	/**
+	 * Find file IDs by querying file settings.
+	 * @param $settingName string
+	 * @param $settingValue mixed
+	 * @param $submissionId int optional
+	 * @param $pressId int optional
```
asmecher: Need to use app-agnostic variables.

bozana:
@asmecher, thanks a lot for the comments -- very useful as always! Luckily you take a look... :-))) 
I've made the changes, s. "pub url id fixes after code review".

asmecher:
Merged too, thanks!

Issue #1560, opened 2016-06-25
[OJS] update all locales before OJS 2.4.8-1 release


mtub:
Make sure that the latest translation updates from `ojs-dev-2_4` are included in `ojs-stable-2_4_8` for the release of OJS 2.4.8-1.

mtub:
There's another update coming in...

asmecher:
@mtub, is this ready to close from your perspective?

mtub:
sorry… closing now.

PR #1610, opened 2016-07-13
pt_PT locale: updated locale files

mtub:
Another update by José Castro Silva

@asmecher , how do you feel about the languages file? It's not translated...

#1560


asmecher:
Untranslated language list: not ideal, but OK for now.

PR #1611, opened 2016-07-14
pt_PT locale: updated locale files

mtub:
cherry-picked from ojs-dev-2_4, see #1610

pkp/pkp-lib#1560


asmecher:
Merged, thanks thanks!

Issue #1410, opened 2016-04-22
Add a "Schedule for Publication" button


stranack:
On the Production stage, can we consider adding a "Schedule for Publication" button just above the Participants grid? This would trigger a modal with the Publication Metadata information.

It took me forever to find where to schedule the submission!

This would also be consistent with other stages, where the buttons above the Participants grid trigger a significant action.

It would also simplify the "Metadata" modal.

asmecher:
Incidentally, this is one of the cases in OMP where we added a banner-style notification to guide the user towards the next step after hitting a dead-end.

Tagging @bozana, because this has implications for her work on e.g. DOIs. Scheduling is one point at which DOIs could get stamped, so there may be a need for a modal that pops up with an opportunity for modals to augment the built-in fields.

asmecher:
@bozana, just reviewing open issues -- I think you were interested in tackling this one as a button like the other editorial decisions.

bozana:
PR:
pkp-lib: https://github.com/pkp/pkp-lib/pull/1622
ojs: https://github.com/pkp/ojs/pull/936

asmecher:
Merged both, thanks -- can this be closed?

PR #1622, opened 2016-07-19
pkp/pkp-lib#1410 add schedule for publication button

bozana:
S. https://github.com/pkp/pkp-lib/issues/1410
plus consider pub id assignment after the scheduling the publication

### DIFF
--- a/classes/plugins/PKPPubIdPluginHelper.inc.php
+++ b/classes/plugins/PKPPubIdPluginHelper.inc.php
@@ -162,15 +162,20 @@ class PKPPubIdPluginHelper {
 	 * @param $contextId integer
 	 * @param $form object
 	 * @param $pubObject object
+	 * @param $save boolean Whether the pub id shall be saved here
 	 * 	Submission, Representation, SubmissionFile + OJS Issue
 	 */
-	function assignPubId($contextId, $form, $pubObject) {
+	function assignPubId($contextId, $form, $pubObject, $save = false) {
 		$pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $contextId);
 		if (is_array($pubIdPlugins)) {
 			foreach ($pubIdPlugins as $pubIdPlugin) {
 				if ($form->getData($pubIdPlugin->getAssignFormFieldName())) {
 					$pubId = $pubIdPlugin->getPubId($pubObject);
-					$pubObject->setStoredPubId($pubIdPlugin->getPubIdType(), $pubId);
+					if ($save) {
+						$pubIdPlugin->setStoredPubId($pubObject, $pubIdPlugin->getPubObjectType($pubObject), $pubId);
+					} else {
+						$pubObject->setStoredPubId($pubIdPlugin->getPubIdType(), $pubId);
+					}
 				}
 			}
 		}
--- a/controllers/grid/pubIds/form/PKPAssignPublicIdentifiersForm.inc.php
+++ b/controllers/grid/pubIds/form/PKPAssignPublicIdentifiersForm.inc.php
@@ -36,18 +36,25 @@ class PKPAssignPublicIdentifiersForm extends Form {
 	var $_confirmationText;
 
 	/**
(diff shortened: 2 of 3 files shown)


controllers/grid/pubIds/form/PKPAssignPublicIdentifiersForm.inc.php
```diff
@@ -115,14 +131,17 @@ function readInputData() {
 
 	/**
 	 * Assign pub ids.
-	 * @copydoc Form::execute()
+	 * @param $request PKPRequest
+	 * @param $save boolean
+	 *  true if the pub id shall be saved here
+	 *  false if this form is integrated somwhere else, where the pub object will be updated.
```
asmecher: *somewhere

controllers/grid/pubIds/form/PKPAssignPublicIdentifiersForm.inc.php
```diff
@@ -66,6 +73,7 @@ function fetch($request) {
 		$templateMgr->assign('pubObject', $this->getPubObject());
 		$templateMgr->assign('approval', $this->getApproval());
 		$templateMgr->assign('confirmationText', $this->getConfirmationText());
+		$templateMgr->assign('formParams', $this->getFormParams());
```
asmecher: This appears to be an OJS-only concern (e.g. it's never used in the pkp-lib templates). Is it possible to keep it in OJS only?

bozana: I think it would be possible to add a new form subclass in OJS that will consider this attribute only. This is relative generic parameter and eventually it could be used in OMP (but it is not at the moment). Would you recommend to have it in the OJS subclass?

asmecher: Alternately, is it possible to make some of the templates generic and share them in lib/pkp? It just seems clunky to have the variable provided only in the pkp-lib PHP code, and used only in the OJS template code.

bozana: OK, OK, I'll create a new small subclass for OJS :-)

bozana:
@asmecher, I considered your comments and moved the formParams into the new OJS form class, s. the last commit "move formParams into application specific form class" above.

classes/plugins/PKPPubIdPluginHelper.inc.php
```diff
@@ -162,15 +162,20 @@ function execute($contextId, $form, $pubObject) {
...
 	 */
-	function assignPubId($contextId, $form, $pubObject) {
+	function assignPubId($contextId, $form, $pubObject, $save = false) {
 		$pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $contextId);
 		if (is_array($pubIdPlugins)) {
 			foreach ($pubIdPlugins as $pubIdPlugin) {
 				if ($form->getData($pubIdPlugin->getAssignFormFieldName())) {
 					$pubId = $pubIdPlugin->getPubId($pubObject);
-					$pubObject->setStoredPubId($pubIdPlugin->getPubIdType(), $pubId);
+					if ($save) {
```
asmecher: There isn't a PR for OMP, correct? Since the default for `$save` is false, won't this result in the pub ID not being saved in OMP anymore?

bozana: No, this was/is the default behaviour -- the pub object (e.g. Representation) is saved in the other form, and the pub id is only assigned to the object here prior to saving it in that other form.
Now, in OJS, using the "schedule for publication", this was the other way aroung -- first the article is saved (e.g. scheduled to an issue) and then the pub id is eventually assigned, thus in this case it has to be saved here.

asmecher: Ah, OK, thanks!

PR #1650, opened 2016-07-27
Fix upgrading of plugins

WaDelma:
As I noticed in #1523 upgrading plugins is currently broken.  
The plugin and it's category isn't currently propagated to the point of upgrade and this pr fixes that.  
This doesn't fix the upgrading completely however: If I try to upgrade plugin after these changes it says:

> The current role does not have access to this operation.

even though it's the admin account.

### DIFF
--- a/classes/controllers/grid/plugins/PluginGridHandler.inc.php
+++ b/classes/controllers/grid/plugins/PluginGridHandler.inc.php
@@ -318,7 +318,9 @@ abstract class PluginGridHandler extends CategoryGridHandler {
 	 */
 	function saveUploadPlugin($args, $request) {
 		$function = $request->getUserVar('function');
-		$uploadPluginForm = new UploadPluginForm($function);
+		$plugin = $request->getUserVar('plugin');
+		$category = $request->getUserVar('category');
+		$uploadPluginForm = new UploadPluginForm($function, $plugin, $category);
 		$uploadPluginForm->readInputData();
 
 		if($uploadPluginForm->validate()) {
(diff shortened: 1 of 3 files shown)


WaDelma:
I have hard time deciphering whats going wrong with permissions: adding and removing plugins works right. And I don't see what it does differently for the upgrade case.

WaDelma:
Could anyone shine light on how the permission system works? I am bit lost with it.

WaDelma:
I was able to find the problem.
After I made it propagate the category and the plugin it started [adding](https://github.com/pkp/ojs/blob/95cfc304dd6045201c1eda180815daca9f5ce56c/controllers/grid/settings/plugins/SettingsPluginGridHandler.inc.php#L77) `OjsPluginAccessPolicy` which contains `RoleBasedHandlerOperationPolicy` which in turn denies the access.    
The weird thing is that actual the actual check `_checkUserRoleAssignment` succeeds, but `_checkOperationWhitelist` fails. This is because the action that it actually tries to do is `saveUploadPlugin` which is not defined for `PluginGridHandler`.    
For some reason uploading plugin doesn't add `OjsPluginAccessPolicy` so it succeeds even though the action is the same.

WaDelma:
Ah. Obviously uploading cannot know the plugin name or category.    
So should I fix this by changing the action to something else that exists, by adding the action as an action, by removing `OjsPluginAccessPolicy` or what?

asmecher:
@WaDelma, I'm sorry for not being more active on this request -- we're currently swamped with preparing OJS 3.0 for release and will be for at least a month. I do have this conversation flagged for follow-up once I get a breather.

WaDelma:
Not a problem. I understand that OJS 3.0 is taking lots of your attention. I am eagerly waiting it myself :smiley:

asmecher:
@bozana, would you mind taking a look at this one too? Related to your recent work on plugin  upgrades.

bozana:
@WaDelma, would it be possible for you to see if this problem is solved in the current master or ojs-stable-3_0_0 branch? -- Because, I believe it should be solved in this issue: https://github.com/pkp/pkp-lib/issues/1872. Thanks a lot!

bozana:
@WaDelma, can this PR be closed? THANKS!

WaDelma:
Yeah I think it can

PR #1657, opened 2016-07-29
add getDAO to DataObject

bozana:
according to the code review comments...

### DIFF
--- a/classes/plugins/PKPPubIdPluginHelper.inc.php
+++ b/classes/plugins/PKPPubIdPluginHelper.inc.php
@@ -172,7 +172,7 @@ class PKPPubIdPluginHelper {
 				if ($form->getData($pubIdPlugin->getAssignFormFieldName())) {
 					$pubId = $pubIdPlugin->getPubId($pubObject);
 					if ($save) {
-						$pubIdPlugin->setStoredPubId($pubObject, $pubIdPlugin->getPubObjectType($pubObject), $pubId);
+						$pubIdPlugin->setStoredPubId($pubObject, $pubId);
 					} else {
 						$pubObject->setStoredPubId($pubIdPlugin->getPubIdType(), $pubId);
 					}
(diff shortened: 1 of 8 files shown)


classes/plugins/PKPPubIdPluginHelper.inc.php
```diff
@@ -195,11 +195,9 @@ function clearPubId($contextId, $pubIdPlugInClassName, $pubObject) {
 				if (get_class($pubIdPlugin) == $pubIdPlugInClassName) {
 					// clear the pubId:
 					// delete the pubId from the DB
-					$pubObjectType = $pubIdPlugin->getPubObjectType($pubObject);
-					$daos = $pubIdPlugin->getDAOs();
-					$dao = $daos[$pubObjectType];
+					$dao = $pubObject->getDAO();
 					$pubObjectId = $pubObject->getId();
-					if ($pubObjectType == 'SubmissionFile') {
+					if (is_a($pubObject, 'SubmissionFile')) {
```
asmecher: I think this exception was necessary because we had some confusing inheritance cases -- see the note in `SubmissionFile.inc.php`'s `getFileId` function:

    // WARNING: Do not modernize getter/setters without considering
    // ID clash with subclasses ArticleGalley and ArticleNote!

I think those cases have both been removed, so we can move to using `getId`/`setId` on `SubmissionFile` subclasses instead of `getFileId`/`setFileId`. I think that means this special case can be removed and `getFileId`/`setFileId` can be removed from `SubmissionFile`, letting calls there fall back to `PKPFile`, which uses the more standard `getId` and `setId`.

classes/submission/Submission.inc.php
```diff
@@ -1157,6 +1157,12 @@ function isCCLicense() {
 		return preg_match('/creativecommons\.org/i', $this->getLicenseURL());
 	}
 
+	/**
+	 * @copydoc DataObject::getDAO()
+	 */
+	function getDAO() {
+		return Application::getSubmissionDAO();
```
asmecher: At some point it would be nice to find a way of removing these functions from `Application` -- but that's a subject for another time.

classes/plugins/PKPPubIdPluginHelper.inc.php
asmecher: (I've merged the PRs, but think this is a worthwhile tweak for separate submission. Mind tinkering with it, or if you're working on other stuff, filing it for later?)

bozana: OK. 
I am not 100% sure, but somehow I think I had to do it this way, at least at that time, when I was implementing it some time ago. I will double check it all and change accordingly... THANKS!!!

classes/submission/Submission.inc.php
bozana: OK

PR #1706, opened 2016-08-09
standardize filter display names

bozana:
just for all of them to use the similar notation, but actually not so important...
would changing this mean something for the upgrade?


asmecher:
Merged -- I don't think this has any ramifications needing consideration.

asmecher:
(Except that these should probably be localized sometime soon.)

PR #1752, opened 2016-08-24
banner notifications

bozana:
S. https://github.com/pkp/pkp-lib/issues/1710
This fixes also https://github.com/pkp/pkp-lib/issues/1682


classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.inc.php
```diff
@@ -444,10 +444,17 @@ function reviewRead($args, $request) {
...
 		if (!$reviewAssignment->getDateCompleted()) {
 			// Editor completes the review.
 			$reviewAssignment->setDateConfirmed(Core::getCurrentDate());
 			$reviewAssignment->setDateCompleted(Core::getCurrentDate());
+			$updateReviewNotifications = true;
+		}
+
+		$this->_updateReviewRoundStatus($reviewAssignment);
+
+		if ($updateReviewNotifications) {
```
asmecher: Is this just to avoid unnecessarily changing the notifications? I would assume that `updateNotifications` won't cause problems if called unnecessarily. (If that's right, no need to change it -- just checking.)

bozana: Yes, you are right :-) I will remove it then...

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
```diff
@@ -0,0 +1,218 @@
...
+	 */
+	function EditingStatusNotificationManager($notificationType) {
+		parent::NotificationManagerDelegate($notificationType);
+	}
+
+	/**
+	 * @copydoc PKPNotificationOperationManager::getNotificationMessage()
+	 */
+	public function getNotificationMessage($request, $notification) {
+		switch ($notification->getType()) {
```
asmecher: Suggest adding a `default: assert(false);` here and below, just because.

bozana: OK :-)

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
```diff
@@ -0,0 +1,218 @@
...
+		$submission = $submissionDao->getById($submissionId);
+
+		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+		$editorStageAssignments = $stageAssignmentDao->getEditorsAssignedToStage($submissionId, $submission->getStageId());
+
+		// One way to check if a copyeditor is assigned
+		// Get the 'copyedit assignment' notifications
+		$copyEditorAssigned = false;
+		$notificationDao = DAORegistry::getDAO('NotificationDAO'); /* @var $notificationDao NotificationDAO */
+		$copyEditAssignmentNotificationFactory = $notificationDao->getByAssoc(
```
asmecher: I worry about making some notifications dependent on others; if we change the handling of copyeditor assignments (which is likely, since the relationship between discussions/notifications is currently a source of confusion in UI/UX testing), we'll have to remember to change this as well. What do you think about checking to see if a discussion exists with at least one assistant involved in it instead?

asmecher: (I see this is more or less what you've got below. Note that notifying a copyeditor automatically creates a query -- so the query check should be enough just by itself. I think that means this can be simplified.)

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
```diff
@@ -0,0 +1,218 @@
...
+
+		$notificationType = $this->getNotificationType();
+
+		foreach ($editorStageAssignments as $editorStageAssignment) {
+			switch ($submission->getStageId()) {
+				case WORKFLOW_STAGE_ID_PRODUCTION:
+					// Remove 'assign a copyeditor' and 'awaiting copyedits' notification
+					$this->_removeNotification($submissionId, $editorStageAssignment->getUserId(), $notificationType, $contextId);
+					break;
+				default:
```
asmecher: Better, I think, to specify `case WORKFLOW_STAGE_ID_EDITING` and add a `default: assert(false);` for other cases.

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
bozana: I did that too i.e. both checks. Shall I then remove this one?

bozana: I think the query is not created automatically, but I will double check...

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
bozana: Yes, good point! :-)

classes/notification/managerDelegate/PendingRevisionsNotificationManager.inc.php
```diff
@@ -97,6 +97,7 @@ public function updateNotification($request, $userIds, $assocType, $assocId) {
 		$userId = current($userIds);
 		$submissionId = $assocId;
 		$stageData = $this->_getStageDataByType();
+		if ($stageData == null) return;
```
asmecher: I see a few conditions added here relating to `$stageData` not being present. What's the use case for that? I think the stage data should reliably describe the application, no? (I don't know this aspect of the codebase well.)

classes/notification/managerDelegate/ProductionStatusNotificationManager.inc.php
```diff
@@ -0,0 +1,210 @@
...
+
+	/**
+	 * @copydoc PKPNotificationOperationManager::getNotificationMessage()
+	 */
+	public function getNotificationMessage($request, $notification) {
+		switch ($notification->getType()) {
+			case NOTIFICATION_TYPE_ASSIGN_PRODUCTIONUSER:
+				return __('notification.type.assignProductionUser');
+			case NOTIFICATION_TYPE_AWAITING_REPRESENTATIONS:
+				return __('notification.type.awaitingRepresentations');
```
asmecher: Suggest adding (as elsewhere): `default: assert(false);`

classes/notification/managerDelegate/PendingRevisionsNotificationManager.inc.php
bozana: WORKFLOW_STAGE_ID_INTERNAL_REVIEW is also in this class, because the NOTIFICATION_TYPE_PENDING_INTERNAL_REVISIONS is just in pkp-lib s. below. Thus in OJS the WORKFLOW_STAGE_ID_INTERNAL_REVIEW will not be found, creating a php notice when _getStageDataByType() below is called.

classes/notification/managerDelegate/ProductionStatusNotificationManager.inc.php
```diff
@@ -0,0 +1,210 @@
...
+		$submissionId = $assocId;
+		$submissionDao = Application::getSubmissionDAO();
+		$submission = $submissionDao->getById($submissionId);
+
+		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+		$editorStageAssignments = $stageAssignmentDao->getEditorsAssignedToStage($submissionId, $submission->getStageId());
+
+		// One way to check if a production user is assigned
+		// Get the 'layout assignment' notifications
+		$layoutEditorAssigned = false;
```
asmecher: As above: suggest making this only dependent on queries, not notifications. And since that will make these two blocks of code basically identical, maybe they could be rolled into one?

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
bozana: just because it can happen, that a copyeditor/layouteditor is assigned and no discussion exist...

controllers/grid/queries/QueriesGridHandler.inc.php
```diff
@@ -320,6 +320,22 @@ function deleteQuery($args, $request) {
 		$notificationDao = DAORegistry::getDAO('NotificationDAO');
 		$notificationDao->deleteByAssoc(ASSOC_TYPE_QUERY, $query->getId());
 
+		// Update submission notifications
+		$notificationsToUpdate = array(
```
asmecher: If the notification manager delegates are aligned into one class (once the notification stuff is removed), maybe it'll be possible to remove the stage-specific stuff from here? The query grid classes so far don't need to behave differently based on what stage they're in, and that would be nice to preserve.

classes/notification/managerDelegate/EditingStatusNotificationManager.inc.php
asmecher: The query should be created when notification happens.

bozana: OK, I will then change it now as you suggests, and the we can then solve the issue with queries later, OK?

asmecher: I'm just thinking out loud, and not sure that thought is coherent :) Have you not been able to replicate the behavior with queries being automatically created when e.g. a copyeditor is notified? If not, I can check it here.

bozana: No queries are automatically created. I will however put everything in one class, I think this will work well. I could also leave the notification check for now and remove it later, when we realized what to do with and implemented automatic queries. Just let me know what do you think would be best for now: with or without notification check?

asmecher: Ah! Sorry, it looks like I wrote that but dropped it sometime before it got merged:
https://github.com/pkp/pkp-lib/issues/1432#issuecomment-219175029

I'll dust it off and merge it in later today. Meanwhile, go ahead and make the notifications dependent on just the queries; I'll let you know when I've got my part merged in and we can make sure the two play nice.

bozana: OK :-)

asmecher:
OK, I've committed the changes to get notifications to automatically create queries!

bozana:
@asmecher, the last commit above are my changes according to the code review. Everything works good with the new queries... :-)

Issue #1663, opened 2016-08-04
Remove Author Names/Reviewer Names from Filenames


stranack:
Currently, uploaded filenames automatically include the name of the user. This is generally useful, except for authors and reviewers during the review process, as it undermines the double blind process.

Can we ensure that author names and reviewer names are not included in filenames?

bozana:
@asmecher, here is the PR: https://github.com/pkp/pkp-lib/pull/1764
All files uploaded in the review stage will have the anonymous default file name. Also the files copied from the submission to the review stage (by "Send to Review" action), if they have the default file names -- if they have the default file names, the names will be change to the anonymous default file names. Hope this is OK...

asmecher:
Closing!

marcbria:
@asmecher any plan to port this to the LTS version?
We all agreed that LTS won't get new features but what about relevant fixes like this one?

asmecher:
@marcbria, this was coded up way back in 2016, so should be included in the LTS. Maybe it's a different issue or a regression?

PR #1764, opened 2016-08-29
pkp/pkp-lib#1663 anonymous default file names in review stage

bozana:
S. https://github.com/pkp/pkp-lib/issues/1663

### DIFF
--- a/classes/submission/SubmissionFile.inc.php
+++ b/classes/submission/SubmissionFile.inc.php
@@ -570,7 +570,7 @@ class SubmissionFile extends PKPFile {
 
 	/**
 	 * Generate a user-facing name for the file
-	 * @param $anonymous boolen Whether the user name should be excluded
+	 * @param $anonymous boolean Whether the user name should be excluded
 	 * @return string
 	 */
 	function _generateName($anonymous = false) {
--- a/classes/submission/SubmissionFile.inc.php
+++ b/classes/submission/SubmissionFile.inc.php
@@ -570,9 +570,10 @@ class SubmissionFile extends PKPFile {
 
 	/**
 	 * Generate a user-facing name for the file
+	 * @param $anonymous boolen Whether the user name should be excluded
 	 * @return string
 	 */
-	function _generateName() {
+	function _generateName($anonymous = false) {
 		$genreDao = DAORegistry::getDAO('GenreDAO');
 		$genre = $genreDao->getById($this->getGenreId());
 		$userGroupDAO = DAORegistry::getDAO('UserGroupDAO');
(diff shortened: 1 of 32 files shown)


classes/submission/SubmissionFile.inc.php
```diff
@@ -570,17 +570,19 @@ function _generateFileName() {
 
 	/**
 	 * Generate a user-facing name for the file
+	 * @param $anonymous boolen Whether the user name should be excluded
```
asmecher: *boolean

locale/en_US/common.xml
```diff
@@ -142,6 +142,7 @@
 	<message key="common.false">False</message>
 	<message key="common.file">File</message>
 	<message key="common.file.namingPattern">{$username}, {$userGroup}, {$originalFilename}</message>
+	<message key="common.file.anonymousNamingPattern">{$userGroup}, {$originalFilename}</message>
```
asmecher: Could you duplicate this throughout the other locales? Better to have it present (and accurate for most locales) than to have `##common.file.anonymousNamingPattern##` showing up as a universal default file title :)

classes/submission/SubmissionFileDAODelegate.inc.php
```diff
@@ -89,8 +89,19 @@ function insertObject($submissionFile, $sourceFile, $isUpload = false) {
 			$submissionFile->setFileId($this->_getInsertId('submission_files', 'file_id'));
 		}
 
+		$reviewStage = $submissionFile->getFileStage() == SUBMISSION_FILE_REVIEW_FILE ||
```
asmecher: $reviewStage = in_array($submissionFile->getFileStage(), array(
        SUBMISSION_FILE_REVIEW_FILE, SUBMISSION_FILE_REVIEW_ATTACHMENT, SUBMISSION_FILE_REVIEW_REVISION,
    ));

asmecher: (slight preference :)

asmecher:
I just thought of something -- this covers new uploads, but what about detected revisions of existing uploads (a common case), or files directly promoted forward, which will already have the author's name in the title? The only easy way to cover both of those, I think, is to just go with the user group name throughout and remove the author name entirely. What do you think, @bozana?

bozana:
Hmmm... I tested it and what I could figure out is that on each place at the review stage only the review stage files are provided for selection, in review files grid, in reviewers grid, new review round, in revisions grid, when requesting revisions and asking for resubmission for review. And those are all anonymized, if they have default names. Thus it seems to be safe/OK, but I am not 100% sure if I oversee something...

asmecher:
Isn't it possible to promote files forward into the review files grid when sending them from the submission stage, when recording a "send to review" decision?

bozana:
Yes, but that I havel covered, within the second l se statement, when the file name exists and has default nam. Thus those are anonymized. (From in between flights :-))

bozana:
I've just rebased and added your suggestions. I will leave the decision to you, if this way or everywhere/everything anonymous. In that second case, I think just the current locale key will have to be changed and eventually copied everywhere.

asmecher:
Thanks, @bozana! Have a great vacation.

PR #1806, opened 2016-09-09
Bump up the max length of the user's email on the form to reset passw…

ptrckb:
It increases the maximum length for the email field on the reset password form to the value of the same field on other form (especially the registration form).


ctgraham:
There's no good reason for there to be a maxlength on this field at all.  I suggest removing the attribute.

ptrckb:
Problem fixed for pkp/pkp-lib#2143.

PR #1830, opened 2016-09-19
define column metric as integer instead of small integer

bozana:
S. http://forum.pkp.sfu.ca/t/ojs-upgrade-failed-2-4-8-1-to-3-0-0/20226/7
the PR for ojs-stable-3_0_0 is coming too.
I am not sure if the cherry-picking to omp-stable-1_2 and omp-dev-1_2 is also necessary? :-)


asmecher:
Cherry-picked over to `omp-stable-1_2_0` and `omp-dev-1_2`, thanks!

PR #1841, opened 2016-09-22
Check is_array again

gonetil:
If $this->_data[$key] isn't an array, then you might get something like:

`stderr: PHP Warning:  Cannot use a scalar value as an array in ojs3/lib/pkp/classes/core/DataObject.inc.php on line 134, referer: http://ojs3.localhost//admin/settings
`

This causes the settings page, in the first tab (settings), to be blank (and the "loading" icon to stay forever loading). This is related with custom locales (we are using es_AR, which seems to be very incomplete at the time).
I am not entirely sure what to do if is_array returns false. Maybe create it?

### DIFF
--- a/classes/core/DataObject.inc.php
+++ b/classes/core/DataObject.inc.php
@@ -130,7 +130,8 @@ class DataObject {
 					if (empty($this->_data[$key])) unset($this->_data[$key]);
 				}
 			} else {
-				$this->_data[$key][$locale] = $value;
+				if (is_array($this->_data[$key])
+					$this->_data[$key][$locale] = $value;
 			}
 		}
 	}


asmecher:
Hi @gonetil -- I suspect this is happening because OJS is encountering an unexpected mix of localized (array) and non-localized (scalar) data. Can you add some `error_log` calls to determine what `$key` is in this situation?

gonetil:
Hi @asmecher ,
seems to be with keys _abbrev_ and _name_ for most journals for es_AR locale. Maybe updating our journals' settings will make it work, or just disabling es_AR locale for the time. I think we're going for that solution.
Anyway, the fix prevents the JSON from being sent _broken_ (which makes OJS unusable). Perhaps, instead of ignoring the error, there should be a flash message or notification in order to let the user know that there are missing keys but to let him/her keep using OJS.

asmecher:
@gonetil, check to ensure that the `locale` column is always consistently set for those settings -- you should never see a case where some entries have it containing a locale (e.g. `es_AR`) and other rows have it empty or null. If you see a mix, remove the empty/null rows. Then flush your settings cache (e.g. `rm cache/*.php`) to ensure that the new data is used. That should resolve the issue.

gonetil:
Yep, that did the trick! Thanks @asmecher

asmecher:
Glad you got it going!

PR #1850, opened 2016-09-27
Ojs stable 3 0 0

knjigor:


bozana:
@knjigor, this is probably against the wrong PKP branch (master), it should be ojs-stable-3_0_0?

knjigor:
Yes, my mistake, I will correct it tomorrow.

PR #1851, opened 2016-09-27
Translation sr sr

knjigor:
Second iteration


mtub:
Hi @knjigor,

I am closing this and other PRs, as we still have to come to a decision regarding the cyrillic and the latin version of the Serbian locale (and the PR would essentially replace the latin version with the cyrillic version). Don't worry, we will get your translation online, I am pretty sure. @vedranvucic is reviewing and updating the files right now.

Thank you!

Issue #1709, opened 2016-08-09
Native Import Issues for OJS and OMP


bozana:
- [x] a) If there was an import error the submission is still partly imported. If an error occurs, remove already processed/imported submissions and issues. Sections will not be removed.
- [x] b1) Consider source files -- At the moment they seem to be ignored i.e. the inserted source_file_id and source_revision is always = null. 
- [x] b2) Different revisions of a file are inserted with different file_id. Instead they should be inserted with the same file_id.
- [x] b3) Consider query files
- [x] c) OJS: Only issue(s) can be imported. Allow article(s) import too
- [x] d) OJS: consider issue reference in articles import XML 
- [x] e) "locale" attribute for localized nodes is optional, which leads to inserting data into DB without locale, if the locale attribute is missing.
- [x] f) OJS: check if a section being imported exists: check titles and abbreviations (as in OJS 2.4.x)
- [x] g) consider some errors catching and displaying them to the user, instead of fatalError.
- [ ] h) file import using href element
- [x] i) provide export XML for download, instead of displaying it in the browser
- [x] j) OJS: add pages element for the article element
- [x] k) Consider controlled vocabularies (keywords, agencies, disciplines, subjects) -- no matter how the journal/press workflow submission settings are i.e. what vocabulary is selected
- [x] l) Fail the import if the revision `uploader` or `user_group_ref` do not exist in the system (moved from https://github.com/pkp/pkp-lib/issues/1728)

bozana:
@asmecher, do you think some of those issues should be addressed for OJS 3.0?

asmecher:
I think we can leave these for 3.0.1, but if we get spare time before the release (ha!) it might be a worthy addition to consider.

bozana:
@asmecher, when a published article is being imported, there have to be an issue associated with it. At the moment we do not have this information. Is there maybe a good way to have a reference to an issue in the import XML file? For example an additional attribute "issue_id" for the "article" element, that would be considered if there is no issue in deployment (i.e. if not the whole issue is being imported, so that we in that case know which issue)? I could add a drop-down list of issues in the import form, for user to select an issue in that case, but maybe it would be easier and also more flexible to have the information in the XML file -- e.g. this way articles from different issues could be imported at once.

bozana:
[...]

PR #1876, opened 2016-10-12
pkp/pkp-lib#1709 provide export XML for download

bozana:
S. i) here: https://github.com/pkp/pkp-lib/issues/1709
(This PR considers also other export plugins)

### DIFF
--- a/classes/plugins/ImportExportPlugin.inc.php
+++ b/classes/plugins/ImportExportPlugin.inc.php
@@ -127,45 +127,22 @@ abstract class ImportExportPlugin extends Plugin {
 
 	/**
 	 * Return the plugin export directory.
-	 *
-	 * This will create the directory if it doesn't exist yet.
-	 *
-	 * @return string|array The export directory name or an array with
-	 *  errors if something went wrong.
+	 * @return string The export directory path.
 	 */
 	function getExportPath() {
-		$exportPath = Config::getVar('files', 'files_dir') . '/' . $this->getPluginSettingsPrefix();
-		if (!file_exists($exportPath)) {
-			$fileManager = new FileManager();
-			$fileManager->mkdir($exportPath);
-		}
-		if (!is_writable($exportPath)) {
-			$errors = array(
-					array('plugins.importexport.common.export.error.outputFileNotWritable', $exportPath)
-			);
-			return $errors;
-		}
-		return realpath($exportPath) . '/';
+		return Config::getVar('files', 'files_dir') . '/temp/';
 	}
 
 	/**
 	 * Return the whole export file name.
+	 * @param $basePath string Base path for temporary file storage
 	 * @param $objectsFileNamePart string Part different for each object type.
 	 * @param $context Context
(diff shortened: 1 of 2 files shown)


bozana:
@asmecher, also here, I did the change, to use TemporaryFileManager. Could you please take a look (the last commit)? THANKS!!!

classes/file/TemporaryFileManager.inc.php
```diff
@@ -57,6 +57,16 @@ function deleteFile($fileId, $userId) {
 	}
 
 	/**
+	 * Delete a file by path.
+	 * For temporary files that are not in the DB, e.g. exported xml files.
+	 * @param $filePath string the location of the file to be deleted
+	 * @return boolean returns true if successful
+	 */
+	function deleteFileByPath($filePath) {
```
asmecher: Why not let the TemporaryFileManager persist the file in the database? Then you could get rid of `deleteFileByPath`, `downloadFileByPath`, and even `ImportExportPlugin::getExportFileName`, and let the TemporaryFileManager handle more of the details.

bozana: Yes, I wasn't sure -- it seemed to me to be to much 'action' to write and read from the database and act like this would be like a file upload (considering those information as at a file upload), because this case is a little bit different -- those exported files are not like file uploads, that are usually treated by this manager -- they just write a content to a file. Do you think it would be better to do so?

asmecher: I agree that persisting to the database isn't useful in this case, but it does permit the `TemporaryFileManager` to take care of managing the file storage, cleanup, etc. The less we re-implement those concerns the better, I think.

For one-shot cases like this there might be a role for a simpler file manager class, but the extra overhead from persisting to the database is so little that I'm not sure it's worth the extra code.

asmecher: (I don't love the TemporaryFileManager in general, and I'm now directing your work without much recent experience with it -- so if you think we're going down a rabbit hole, please stop me :)

bozana: Hmmm... Somehow I don't feel this is good ;-) -- I have first to create a file (from XML string) in order then to divide it apart, to fit the TemporaryFile structure, to get and save information into the DB, that I don't need at all (newFileName, fileType, fileSize, originalFileName). Thus, maybe then to:
1) write a new class (ExportFileManager?) that will extend the PrivateFileManager and just add the new base path = '/temp/'. Or to 
2) use just a normal FileManager, giving it the manually constructed path to that temp folder? -- I actually only need that base path from the TemporaryFileManager (and to mark it as a temporary file for the code readers) :-P Everything else what I need is in the FileManager (writeFile, downloadFile, deleteFile -- that's it). 
What do you think? Maybe the #2?

bozana:
@asmecher, I now changed the commit above to use just the normal FileManager...

asmecher:
Merged, thanks!

PR #1895, opened 2016-10-18
Make datepicker respect date formats

defstat:


asmecher:
I think this went against the wrong branch or needs a rebase -- there are a lot of commits in the PR that shouldn't be here.

PR #1945, opened 2016-11-02
ImportExport js object added to Helper

defstat:
Added the $.pkp.plugins.importexport object into Helper.js. Needed for the QuickSubmitPlugin (and other importexport plugins) 

https://github.com/pkp/pkp-lib/issues/1792

Issue #1792, opened 2016-09-02
Port QuickSubmit plugin to OJS 3.0


asmecher:

defstat:
I am on that....

asmecher:
Assigned!

bozana:
I think, the milestone here can be changed -- it is ready for OJS 3.0.2, right? Maybe the next milestone would be OMP?

asmecher:
I need to either add the plugin as a submodule or (better) add it to the plugin gallery; just leaving it open until then so I don't forget. I don't think we need this plugin for OMP, or at least not yet.

asmecher:
Added to plugin gallery!

bozana:
Coooooool! :-)

marcbria:
I can't find this plugin in the plugin gallery of the last OMP (3.1.1.4).
It this plugin supersede by any OMP core feature?

Sorry in advance if this is not he right place/post to ask.

asmecher:
@marcbria, there isn't an OMP QuickSubmit plugin (yet); just an OJS one.

Riznor:
The plugin is in my gallery but I can't install it.

asmecher:
@Riznor, the best place to get technical support is on our support forum at https://forum.pkp.sfu.ca. Include information on what version of OJS you're using, how you tried to install the plugin, and what the result was.

Riznor:
Thank you, @asmecher.

PR #1949, opened 2016-11-02
ImportExport js object added to Helper

bozana:
s. https://github.com/pkp/pkp-lib/pull/1945#issuecomment-257985786

s. https://github.com/pkp/pkp-lib/issues/1792


bozana:
Thanks a lot @defstat!

PR #1987, opened 2016-11-14
pkp/pkp-lib#1709 native import error handling

bozana:
s. a), g) and l) here: https://github.com/pkp/pkp-lib/issues/1709

### DIFF
--- a/classes/plugins/importexport/PKPImportExportDeployment.inc.php
+++ b/classes/plugins/importexport/PKPImportExportDeployment.inc.php
@@ -27,6 +27,9 @@ class PKPImportExportDeployment {
 	/** @var Submission The current import/export submission */
 	var $_submission;
 
+	/** @var array The processed import objects IDs */
+	var $_processedObjectsIds;
+
 	/** @var array Connection between the file and revision IDs from the XML import file and the DB file IDs */
 	var $_fileDBIds;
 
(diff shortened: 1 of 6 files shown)


classes/submission/SubmissionDAO.inc.php
```diff
@@ -134,6 +134,12 @@ function deleteById($submissionId) {
 		// Delete submission files.
 		$submission = $this->getById($submissionId);
 		assert(is_a($submission, 'Submission'));
+		// 'deleteAllRevisionsBySubmissionId' has to be called before 'rmtree'
+		// because SubmissionFileDaoDelegate::deleteObjects checkts the file
```
asmecher: *checks

classes/plugins/importexport/PKPImportExportDeployment.inc.php
```diff
@@ -121,6 +126,39 @@ function getSubmission() {
...
+	function addProcessedSubmissionId($submissionId) {
+		$this->_processedSubmissionsIds[$submissionId] = array();
+	}
+
+	/**
+	 * Add the error message to the processed submission ID.
+	 * @param $submissionId integer
+	 * @param $errorMsg string
+	 */
+	function addSubmissionError($submissionId, $errorMsg) {
```
asmecher: Since we may also need to clean up other dependent types, maybe this would make a good general feature rather than just submission-specific. Maybe something like `addError($assocType, $assocId, $errorMsg)` etc? Likewise `set`/`getProcessedSubmissionsIds`.

Issue #1793, opened 2016-09-02
Test/fix PostgreSQL upgrades


asmecher:
See http://forum.pkp.sfu.ca/t/upgrade-failure-postgresql/19215

asmecher:
Patches to apply:
 - In `lib/pkp`: https://github.com/pkp/pkp-lib/commit/7e33ad63f2ee4c2b3063923d70a36e3fab23f07e
 - In root: https://github.com/pkp/ojs/commit/69ed23aba986a0206ff9eb4f92dd21f30b6edccd
 - In root: https://github.com/pkp/ojs/commit/625a73622375ff07242452a32fc939dfbcca5496
 - To correct `ERROR: type "serial" does not exist`: Download patch from https://pkp.sfu.ca/support/forum/viewtopic.php?t=11896#p45922

asmecher:
Closing for now.

mittinatten:
The patch for "ERROR: type serial.." can't be applied as is, had to remove the last change in the set (which was just a new-line).

asmecher:
@mittinatten OK, thanks for the heads-up (it's quite an old patch). Unfortunately that's a patch to the ADODB library which is getting very long in the tooth; I suspect we're going to have to replace it, which is not going to be exciting. Otherwise, did the set of patches above fix the upgrade process for you?

mittinatten:
Unfortunately not, do you want me to post the rest here, in the PKP-forum thread, or as a new issue?

asmecher:
@mittinatten, thanks for the details. What version are you upgrading from? Would you be willing to provide me with a database dump from the old version for local testing over here? (If so, feel free to alter whatever you want before sending it, particularly `users.password`, `users.email`, and `authors.email`.)

mittinatten:
Upgrading from 2.4.2. Can see what I can do with the dump when I'm back at work tomorrow, where can I send it?

asmecher:
I'm alec at smecher point bc point ca, thanks!

mittinatten:
Sent it yesterday, just checking to verify the email didn't get caught in any spam filters.

mittinatten:
I tried upgrading the same database from 2.4.2 to 2.4.8 and that worked without error messages.

asmecher:
@mittinatten, just a heads-up that I received the database, thanks -- just trying to scrape together the time to give it a test.

vvkorotkikh:
[...]

PR #2065, opened 2016-12-06
Fix for some Postgres problems

mittinatten:
This is related to #1793. I need to maintain a fork of pkp-lib for just these few lines. Could it be considered for inclusion in future versions?


asmecher:
@mittinatten, that looks like an incomplete form of this patch:
http://wayback.archive-it.org/7100/20160821102046/https://pkp.sfu.ca/support/forum/viewtopic.php?t=11896#p45922

I'm hesitant to merge modifications into ADODB because that's 3rd-party code, and inhibits us from packaging e.g. in Debian using an external dependency for ADODB. A better solution would be to upgrade the ADODB library that we're using and pitch into the upstream project, or to stop using ADODB and work with another library instead. (I'm not aware of other replacements that cover schema management features we currently need e.g. for upgrades, and my interest in replacing ADODB is decreasing as I've recently seen some signs of life from the project.)

asmecher:
(This is secondary, but the branch could use a rebase -- there are lots of other unintentional commits included here.)

mittinatten:
It is based on the patch you link to yes. I understand your hesitation, I will cross my fingers that there'll be an update to ADODB in a not too distant future. It would simplify things significantly if our fork of OJS wasn't also dependent on a fork of pkp-lib, especially when it's such a tiny difference.

(Right, I only looked at the diff, not the history, will attempt a rebase)

Issue #1726, opened 2016-08-12
Submission Locale


bozana:
Provide a possibility to change the submission locale at any point in the workflow. S http://forum.pkp.sfu.ca/t/doaj-plugin-and-multi-language-articles/17210/6

bozana:
@asmecher, could you take a look at this PR: https://github.com/pkp/ojs/pull/1012
Should this feature also be migrated to OJS 3.x?

asmecher:
Yes, this should go to 3.x too, thanks!

asmecher:
Merged the 2.x PR; just needs porting to OJS 3.0.x / OMP 1.2.x.

bozana:
PRs:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/2066
ojs master: https://github.com/pkp/ojs/pull/1124 (remove unused localele keys + submodule update)
omp master: https://github.com/pkp/omp/pull/365 (just submodule update)

bozana:
@asmecher, could you take a look at the PRs above?

asmecher:
Just one comment, otherwise looks good!

bozana:
It is necessary to separate submission locale filed from the title and abstract fields, not to be displayed in the submission step 3 (where title and abstract are displayed), because it should be displayed in the submission step 1.

bozana:
PRs:
pkp-lib: https://github.com/pkp/pkp-lib/pull/2094
ojs: https://github.com/pkp/ojs/pull/1142
omp: https://github.com/pkp/omp/pull/367

bozana:
@asmecher, the tests are still running, but could you please take a look at these new 3 RRs? -- I had to separate submission locale field in a new template, to be able to use it in different places. Also, I added the possibility to change the monograph locale, which wasn't there -- is this OK i.e. could you think of any reason why this is not a good idea? THANKS!!!

asmecher:
All looks good and the tests passed! Just one trivial comment, please go ahead and merge.

asmecher:
(The monograph locale not updating was a bug, thanks!)

PR #2066, opened 2016-12-06
pkp/pkp-lib#1726 possibility to change the submission locale

bozana:
s. https://github.com/pkp/pkp-lib/issues/1726


classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -139,6 +140,9 @@ function execute($submission, $request) {
 		$submission->setSource($this->_parentForm->getData('source'), null); // Localized
 		$submission->setCitations($this->_parentForm->getData('citations'));
 
+		// Update submission locale
+		$submission->setLocale($this->_parentForm->getData('locale'));
```
asmecher: It looks like this value is currently trusted without validation -- it needs to be one of the supported locales.

Issue #1861, opened 2016-10-01
Add tool to clean reviewer interests


asmecher:
Write a tool to remove user interests that are not referenced by any user accounts. This can be useful for cleaning out the controlled vocabulary after removing spam user registrations.

asmecher:
Related tables: `users`; `user_interests`; `controlled_vocab_entries`; `controlled_vocab_entry_settings`.

asmecher:
@kaschioudi, I have an OMP user who would like to run this -- how much do you have left to do?

kaschioudi:
@asmecher: I just pushed the remaining changes
https://github.com/kaschioudi/pkp-lib/commit/c2ff57ad7ba59f5d0466caa697994bdf2abc2f93
https://github.com/kaschioudi/ojs/commit/fd0076aa3362d58aba005d7880460024d2082bc5

Here's the PR for pkplib: https://github.com/pkp/pkp-lib/pull/2186

Oddly enough I am unable to create a PR for ojs (https://github.com/pkp/ojs/compare/master...kaschioudi:1861-fix?expand=1). Not sure why..

asmecher:
It won't let you create another pull request while one is already open -- but the open one has your latest commits pushed to it, so that's OK.

asmecher:
Merged and ported over to OMP `master` and `omp-stable-1_2_0`, thanks!

PR #2070, opened 2016-12-07
bug fix: when getByControlledVocabId() gets called without filter…

kaschioudi:
…  no need to group by controlled_vocab_entry_settings.id #1861


kaschioudi:
@asmecher

asmecher:
@kaschioudi, OTOH that might not work for PostgreSQL -- but using `cve.controlled_vocab_entry_id` as `GROUP BY` should work, and would also solve https://github.com/pkp/pkp-lib/issues/1856. Committing a fix to that issue instead of using this PR: https://github.com/pkp/pkp-lib/commit/f52aa5a669e0937ebe2a494dfab73478d0adb900

PR #2080, opened 2016-12-10
proposed Bitcoin ISO 4217 code (XTC) added

Geremia:
cf. [https://en.bitcoin.it/wiki/Bitcoin_symbol#Currency_code](https://en.bitcoin.it/wiki/Bitcoin_symbol#Currency_code)


asmecher:
Thanks, @Geremia -- we've run into controversy a number of times in our locale data (e.g. country lists), and try as much as possible to keep to the respective standards. As you note, ISO4217 doesn't yet include a Bitcoin entry. If you have a local need for this, I'd suggest maintaining it as a modification for the time being.

PR #2094, opened 2016-12-13
pkp/pkp-lib#1726 create new template for submission locale field

bozana:
separate submission locale field from others -- in order not to be included in the submission step 3, but in the submission step 1
s. https://github.com/pkp/pkp-lib/issues/1726#issuecomment-266748463

### DIFF
--- a/templates/submission/submissionLocale.tpl
+++ b/templates/submission/submissionLocale.tpl
@@ -1,0 +1,21 @@
+{**
+ * submission/submissionLocale.tpl
+ *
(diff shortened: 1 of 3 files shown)


templates/submission/submissionLocale.tpl
```diff
@@ -0,0 +1,21 @@
+{**
+ * submission/submissionLocale.tpl
```
asmecher: Use the full pathname here.

PR #2161, opened 2017-01-08
Update ru_ru translation and fix en_US typo

ppv1979:

PR #2162, opened 2017-01-08
ru_RU locale update

mtub:
- see discussion of these changes at the now superseded PR #2161 
- lots of files changed, including en_US locale files
- have reviewed the changes made by @ppv1979 at #2161


locale/ru_RU/countries.xml
```diff
@@ -35,7 +32,7 @@
 	<country name="Бахрейн" code="BH"/>
 	<country name="Бангладеш" code="BD"/>
 	<country name="Барбадос" code="BB"/>
-	<country name="Беларусь" code="BY"/>
+	<country name="Белоруссия" code="BY"/>
```
asmecher: Does this correspond to the Debian ISO 3166-1 iso-codes package, which we're generally using for translations of this file? I'm nervous about inadvert politics. (See the first paragraph of https://en.wikipedia.org/wiki/Belarus)

asmecher:
Wow, that is a ton of work, and looks very cleanly done.

locale/ru_RU/countries.xml
ppv1979: https://ru.wikipedia.org/wiki/ISO_3166-1 — calls it «Белоруссия», 
the russian version of the article  you mentioned (https://en.wikipedia.org/wiki/Belarus)also says «Белоруссия» — https://ru.wikipedia.org/wiki/Белоруссия.

asmecher: Thanks, @ppv1979 -- Frankly the issue is that I'm ill-informed about geopolitics and have been bitten by it before :) Basically, as long as we're following ISO 3166-1 conventions I'm happy. Obviously users are free to adapt their installations if they're not satisfied with that but I generally try to take the least controversial approach by default.

ppv1979: @asmecher, I see.  I've checked translation of other countries with Debian package and, you know, that some of that translations are exact, but difficult to find. For example, UK is "Соединенное королевство" (exact translation of the "United Kingdom"), but in Russia we don't use such name. We usually say "Великобритания" ("Great Britain"), but people didn't match "Соединенное королевство" with it. Another example, US is "Соединённые штаты" (exact translation of the "United States"), but in Russia we use "Соединённые штаты Америки" ("United States Of America"), "США" ("USA"), or "Америка" (America), so "Соединённые штаты" also is the name that people do not find.  Should we use the exact translation in this situations? I don't think so. I think we should use the way that people know that countries.

As for Belarus — it's not a problem, we can use "Беларусь", because people understand it.

Should I add the commit with fixed files or @mtub can do it?

mtub: I can patch this, will do it right now. As for the further question on country names, I tend to agree with @asmecher – even if the Debian list is not perfect, it is something that we can refer to without making every single decision on a name or spelling ourselves. We haven't come with a plan to generate the country list automatically from the Debian package yet.

mtub:
It is, Pavel has done some great work here. We will have some important locale updates for the next release :)

Thanks again, @ppv1979 !

PR #2168, opened 2017-01-10
Update ISO lists according to the current ISO standards

ppv1979:
After discussion in PR #2162 with @asmecher I’ve checked the lists of countries, languages and currencies with ISO standards. 

Here is the update for en_US and translation for ru_RU (the translation is based on the Russian standards mentioned in the header of each file).

### DIFF
--- a/locale/en_US/countries.xml
+++ b/locale/en_US/countries.xml
@@ -8,7 +8,7 @@
   * Copyright (c) 2000-2016 John Willinsky
   * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
   *
-  * Localized list of countries. Based on ISO 3166-1.
+  * Localized list of countries. Based on ISO 3166-1:2006.
   -->
 <countries>
 	<country name="Afghanistan" code="AF"/>
@@ -37,7 +37,7 @@
 	<country name="Benin" code="BJ"/>
 	<country name="Bermuda" code="BM"/>
 	<country name="Bhutan" code="BT"/>
-	<country name="Bolivia, Plurinational State of" code="BO"/>
+	<country name="Bolivia (Plurinational State of)" code="BO"/>
 	<country name="Bonaire, Sint Eustatius and Saba" code="BQ"/>
 	<country name="Bosnia and Herzegovina" code="BA"/>
 	<country name="Botswana" code="BW"/>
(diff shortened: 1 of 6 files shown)


locale/en_US/countries.xml
```diff
@@ -62,15 +62,15 @@
...
+	<country name="Congo (the Democratic Republic of the)" code="CD"/>
 	<country name="Cook Islands" code="CK"/>
 	<country name="Costa Rica" code="CR"/>
 	<country name="Côte d'Ivoire" code="CI"/>
 	<country name="Croatia" code="HR"/>
 	<country name="Cuba" code="CU"/>
 	<country name="Curaçao" code="CW"/>
 	<country name="Cyprus" code="CY"/>
-	<country name="Czech Republic" code="CZ"/>
+	<country name="Czechia" code="CZ"/>
```
asmecher: Well I'll be! As of September 2016. https://en.wikipedia.org/wiki/Name_of_the_Czech_Republic#English-language_name

asmecher:
Many thanks, @ppv1979! Out of curiosity, what procedure did you use to update these? There are some (rarely used) tools in the `tools/` subdirectory, but I'm not sure if that's what you used.

locale/en_US/countries.xml
mtub: Still think it would be nice to have an easy update-lists-from-ISO process…like isoquery-to-XML...

ppv1979:
@asmecher, I use old, slow and time consuming way — manual comparing using my eyes :-) It's the best way in this situation.

locale/en_US/countries.xml
ppv1979: @mtub, yes it would be easy, but I think that if you try to compare en_US files with Debian iso-package you will see that they are not the same, because I've used here the original materials from ISO. As for Russian translation (I've checked it with Debian files) they have many differences.

asmecher:
@ppv1979, remind me to buy you a drink if the opportunity arises. One day we'll hopefully move our translations to a standard .po file toolset or else something that isn't homegrown -- and when that happens we can just use other standards without having to convert them.

ppv1979:
@asmecher, I'll write down about the drink :-)  You know, a clean XML file for me is quite good, because I open two files in TextWrangler, if necessary I can compare these files and copy necessary strings from en_US to ru_RU and translate it after that. So, I am translating looking in two files: in en_US and ru_RU.

PR #2183, opened 2017-01-11
pkp/pkp-lib#2182 Add CSRF token when deleting a note

NateWr:


asmecher:
Ah -- thanks, @NateWr; I'll also cherry-pick this around.

PR #2188, opened 2017-01-12
Remove nl2br from submissions template for author guidelines

ajnyga:
The template here https://github.com/pkp/pkp-lib/blob/master/templates/frontend/pages/submissions.tpl
calls for the author guidelines setting. However, it includes the nl2br function which adds new linebreaks. The text is already edited with wysiwyg.


asmecher:
Also cherry-picked over to `ojs-stable-3_0_1`, thanks!

Issue #2076, opened 2016-12-08
Review page counting code: MEDRA and EndNote


ctgraham:
The MEDRA export tries to count pages here:
https://github.com/pkp/ojs/blob/c3c29f172c9a5b2d579d969f4044101617fd7921/plugins/importexport/medra/filter/ArticleMedraXmlFilter.inc.php#L199-L216

In 2.4.x, so did the EndNote citation format:
https://github.com/pkp/ojs/blob/dbe779f2bd97e9a0f2842c4357fb56496b7179e4/plugins/citationFormats/endNote/citation.tpl#L36-L44
But now it just expresses a range:
https://github.com/pkp/ojs/blob/4af018f3a9c5908df9ca9d8b1c8de80471acff79/plugins/citationFormats/endNote/citation.tpl#L34

The `Article::getStartingPage()` and `Article::getEndingPage()` enforce numeric return values and thus are used for page counting, but this is loose logic.

ctgraham:
I'm pretty sure MEDRA supports a repeating `PageRun` element, which describes contiguous sequences of pages.  See: https://www.medra.org/schema/onix/DOIMetadata/2.0/ONIX_DOIMetadata_2.0.xsd
```
<xs:element name="TextItem" minOccurs="0">
  <xs:complexType>
    <xs:choice>
      <xs:sequence>
        <xs:element name="TextItemType" type="SerialArticleTextItemTypeCode"/>
        <xs:element name="PageRun" type="PageRun" minOccurs="0" maxOccurs="unbounded"/>
        <xs:element name="NumberOfPages" type="NumberOfPagesNumber" minOccurs="0"/>
      </xs:sequence>
      <xs:sequence>
        <xs:element name="PageRun" type="PageRun" maxOccurs="unbounded"/>
        <xs:element name="NumberOfPages" type="NumberOfPagesNumber" minOccurs="0"/>
      </xs:sequence>
      <xs:element name="NumberOfPages" type="NumberOfPagesNumber"/>
    </xs:choice>
  </xs:complexType>
</xs:element>
```

Since we don't capture the number of pages, and can't reliably calculate it, this would be preferable.

ctgraham:
What little documentation I've found on the Refer/BibIX format we're using for EndNote also suggests that the current expression of a page range rather than a page count is legitimate.

http://web.archive.org/web/20070905162121/http://www.ecst.csuchico.edu/~jacobsd/bib/formats/refer.html

Issue #333, opened 2015-02-10
CrossRef XML Export - Ability to handle roman numerals.


bdgregg:
As discussed on the forum here: http://pkp.sfu.ca/support/forum/viewtopic.php?f=8&t=12529&p=52679#p52679 the CrossRef XML Export needs to support the ability to handle roman numerals in the pages XML output.  

Additionally, we should include some guidance on the Editor page where you are editing the Table of Contents such that we provide a link to some help (maybe via a mouse over an ? icon) to accepted valid options for this column such as "i-iv", "a-s", "1-34", etc.  

If this is not resolved any CrossRef XML export is considered invalid (as OJS removes the first_page, other_pages attributes from the XML) if the journal chooses to use the CrossRef documented and allowed roman numerals in the pages field.

asmecher:
This field is not just used for CrossRef, so the impact would need to be checked against other areas. For example:
- Resolver plugin: plugins/gateways/resolver/ResolverPlugin.inc.php
- URN Pub ID plugin: plugins/pubIds/urn/URNPubIdPlugin.inc.php
- DOI Pub ID plugin: plugins/pubIds/doi/DOIPubIdPlugin.inc.php
- DOAJ export: plugins/importexport/doaj/DOAJExportDom.inc.php
- DataCite export: plugins/importexport/datacite/classes/DataciteExportDom.inc.php
- PubMed export: plugins/importexport/pubmed/PubMedExportDom.inc.php
- MEDRA export: plugins/importexport/medra/classes/O4DOIExportDom.inc.php
- NLM OAI: plugins/oaiMetadataFormats/nlm/OAIMetadataFormat_NLM.inc.php

(These are the areas where OJS might try to tease apart the format of the page numbering; there are other areas, like DC11, native import/export, etc., where the contents of that field are passed along verbatim.)

marcbria:
I agree with Clinton comment in forum: With CrossRef we are in schema violation.
http://www.crossref.org/schemas/common4.3.0.xsd

> [xsd:element name="pages"]
> The container for information about page ranges.
> When an entity has non-contiguous page information, you should capture the first page range in first_page and last_page. Any additional page information should be captured in other_pages.
> Punctuation is only allowed in other_pages. It should not appear in first_page and last_page.
> Page number letter prefixes or suffixes should be included. Roman numeral pages are permitted in both upper case and lower case.
> Data may be alpha, numeric or a combination.

Is it possible to isolate the change to CrossRef plugin?

[...]

PR #2194, opened 2017-01-12
pkp/pkp-lib#333, pkp/pkp-lib#2076: update Submission::getPageArray() to deal appropriately with empty page values.

ctgraham:
This is a small bugfix for pkp/pkp-lib#333, and affects pkp/pkp-lib#2076.

The problem was that `Submission::getPageArray()` was returning `array(array(''))` when pages has been set to an empty string.  That certainly was not my intent and seems counter intuitive.  The return should have been just an empty array.


bozana:
Thanks a lot @ctgraham!

PR #2196, opened 2017-01-13
pkp/pkp-lib#333, pkp/pkp-lib#2076: forward port Submission::getPageArray() to master.

ctgraham:
Helps to forward port pkp/pkp-lib#333 to master and provides foundation for resolving pkp/pkp-lib#2076.

Both issues still need application-specific pull requests for full resolution.

### DIFF
--- a/classes/submission/Submission.inc.php
+++ b/classes/submission/Submission.inc.php
@@ -924,6 +924,37 @@ abstract class Submission extends DataObject {
 	}
 
 	/**
+	 * get pages as a nested array of page ranges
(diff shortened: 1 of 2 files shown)


bozana:
Thanks a lot @ctgraham!

bozana:
Ah, the test is unfortunately not working in master :-( The Submission is abstract class there, so the object cannot be created.
Maybe we could either use the Application::getSubmissionDAO and then create the appropriate object (if ArticleDAO -> new Article(), if MonographDAO -> new Monograph()), or write another test class in each application, that would define $this->submission. @ctgraham, what do you think would be better?

ctgraham:
Hmmm.... both approaches feel kind of awkward to me.  Who in our group has the most experience with unit testing to offer a recommendation?

bozana:
maybe @asmecher :-)

bozana:
I temporarily solved it, for tests not to fail, but when @asmecher suggests a good solution, I can change it again...

PR #2200, opened 2017-01-14
Save comments to editor to a note

ajnyga:
This is a suggestion for saving the comments for the editor to the Pre-Review Discussions.
https://github.com/pkp/pkp-lib/issues/1873

### DIFF
--- a/classes/submission/Submission.inc.php
+++ b/classes/submission/Submission.inc.php
@@ -233,14 +233,6 @@ abstract class Submission extends DataObject {
 	}
 
 	/**
-	 * Get comments to editor.
-	 * @return string
-	 */
-	function getCommentsToEditor() {
-		return $this->getData('commentsToEditor');
-	}
-
-	/**
 	 * Return option selection indicating if author should be hidden in issue ToC.
 	 * @return int AUTHOR_TOC_...
 	 */
@@ -254,14 +246,6 @@ abstract class Submission extends DataObject {
 	 */
 	function setHideAuthor($hideAuthor) {
 		$this->setData('hideAuthor', $hideAuthor);
-	}
-
-	/**
-	 * Set comments to editor.
-	 * @param $commentsToEditor string
-	 */
-	function setCommentsToEditor($commentsToEditor) {
-		$this->setData('commentsToEditor', $commentsToEditor);
 	}
 
 	/**
--- a/classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
+++ b/classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
@@ -114,7 +114,7 @@ class PKPSubmissionSubmitStep1Form extends SubmissionSubmitForm {
 			$query = $this->getCommentsToEditor($this->submissionId);			
 			$this->_data = array_merge($data, array(
 				'locale' => $this->submission->getLocale(),
-				'commentsToEditor' => $query->getHeadNote()->getContents(),
+				'commentsToEditor' => $query ? $query->getHeadNote()->getContents() : '',
 			));
 		} else {
 			$supportedSubmissionLocales = $this->context->getSupportedSubmissionLocales();
@@ -162,6 +162,10 @@ class PKPSubmissionSubmitStep1Form extends SubmissionSubmitForm {
 
 	/**
 	 * Add or update comments to editor
+	 * @param $submissionId int
+	 * @param $commentsToEditor string
+	 * @param $userId int
+	 * @param $query Query optional
 	 */
 	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $query = null) {
 		$queryDao = DAORegistry::getDAO('QueryDAO');
@@ -197,19 +201,22 @@ class PKPSubmissionSubmitStep1Form extends SubmissionSubmitForm {
 				$note->setDateModified(Core::getCurrentDate());
 				$noteDao->updateObject($note);
 			}
-
+			
 		}
-
+		
 	}
 	
 	/**
 	 * Get comments to editor
+	 * @param $submissionId int
+	 * @return null|Query
 	 */
 	function getCommentsToEditor($submissionId) {
+		$query = null;
 		$queryDao = DAORegistry::getDAO('QueryDAO');		
 		$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $submissionId);
-		$query = $queries->next();
-		return $query; 		
+		if ($queries) $query = $queries->next();
+		return $query;
 	}
 
 	/**
@@ -232,11 +239,7 @@ class PKPSubmissionSubmitStep1Form extends SubmissionSubmitForm {
 			// Add or update comments to editor
 			if ($this->getData('commentsToEditor')){
 				$query = $this->getCommentsToEditor($this->submissionId);				
-				if (isset($query)){
-					$this->setCommentsToEditor($this->submissionId, $this->getData('commentsToEditor'), $user->getId(), $query);
-				} else{
-					$this->setCommentsToEditor($this->submissionId, $this->getData('commentsToEditor'), $user->getId());
-				}
+				$this->setCommentsToEditor($this->submissionId, $this->getData('commentsToEditor'), $user->getId(), $query);
 			}
 			$submissionDao->updateObject($this->submission);
 		} else {
--- a/classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
+++ b/classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
@@ -161,6 +161,51 @@ class PKPSubmissionSubmitStep1Form extends SubmissionSubmitForm {
 	}
 
 	/**
+	 * Add or update comments to editor
+	 */
+	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $update = null) {
+		$queryDao = DAORegistry::getDAO('QueryDAO');
+		$noteDao = DAORegistry::getDAO('NoteDAO');
+
+		if (!isset($update)){
+			$query = $queryDao->newDataObject();
+			$query->setAssocType(ASSOC_TYPE_SUBMISSION);
+			$query->setAssocId($submissionId);
+			$query->setStageId(WORKFLOW_STAGE_ID_SUBMISSION);
+			$query->setSequence(REALLY_BIG_NUMBER);
+			$queryDao->insertObject($query);
+			$queryDao->resequence(ASSOC_TYPE_SUBMISSION, $submissionId);
+			$queryDao->insertParticipant($query->getId(), $userId);
+			$queryId = $query->getId();
+
+			$note = $noteDao->newDataObject();
+			$note->setUserId($userId);
+			$note->setAssocType(ASSOC_TYPE_QUERY);
+			$note->setTitle('Comments to editor'); // Localize?
+			$note->setContents($commentsToEditor);
+			$note->setDateCreated(Core::getCurrentDate());
+			$note->setDateModified(Core::getCurrentDate());			
+			$note->setAssocId($queryId);
+			$noteDao->insertObject($note);
+
+		} else{
+			$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $submissionId);
+			$query = $queries->next();
+			$queryId = $query->getData('id');
+
+			$notes = $noteDao->getByAssoc(ASSOC_TYPE_QUERY, $queryId);
+			if (!$notes->wasEmpty()) {
+				$note = $notes->next();
+				$note->setContents($commentsToEditor);
+				$note->setDateModified(Core::getCurrentDate());
+				$noteDao->updateObject($note);
+			}
+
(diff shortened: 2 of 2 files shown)


classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -176,11 +223,20 @@ function execute($args, $request) {
 				$this->submission->stampStatusModified();
 				$this->submission->setSubmissionProgress($this->step + 1);
 			}
+			// Add or update comments to editor
+			if ($this->getData('commentsToEditor')){
+				$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $this->submissionId);
+				$query = $queries->next();
+				if (isset($query)){
+					$this->setCommentsToEditor($this->submissionId, $this->getData('commentsToEditor'), $user->getId(), 1);
```
bozana: How about giving the `$query` object as the last parameter to the `setCommentsToEditor `function, instead of `$update`? If $query is set than it would be updated and if not a new would be created...

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -161,13 +161,60 @@ function setSubmissionData($submission) {
...
+			$note->setContents($commentsToEditor);
+			$note->setDateCreated(Core::getCurrentDate());
+			$note->setDateModified(Core::getCurrentDate());			
+			$note->setAssocId($queryId);
+			$noteDao->insertObject($note);
+
+		} else{
+			$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $submissionId);
+			$query = $queries->next();
+			$queryId = $query->getData('id');
```
bozana: you can use `getId()` here

bozana:
@ajnyga, this works good :-) I have just a few question, s. the commit... 
Also: in the function `initData` there is this line: 
`'commentsToEditor' => $this->submission->getCommentsToEditor()`. 
Could we here use the discussion content, instead of `$this->submission->getCommentsToEditor()` (that will not exist any more).
Maybe one more note: I suppose there is only one discussion entry for the given submission at this place -- because the submission is just starting, right?

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: Great idea!

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: I think I had difficulties in getting getId() to work there, I am not sure why. I will try again.

ajnyga:
I will look at the init. There is also the question of chancing the database table, there is a field for the comments at the moment. I do not know whether you want these two forms of saving the comments to co-exist for now? 

Regarding the last question, I think that at this stage there is no possibility that there would be other discussions concerning the submission. That would be only possible after submitting. And after that this thread would be handled in the pre-review discussions.

bozana:
I think we should then provide an upgrade and move all current comments to the discussions and remove that column `comments_to_ed` in the DB table `submissions`. @asmecher, would you agree?
@ajnyga, if you would like and you could (because of the time) work on that upgrade script, I would be more than happy, but it is no problem if you do not want or cannot -- I will do it...

ajnyga:
Yes, the upgrade script would definitely need changes as well. 

I can take a look later today.

asmecher:
(Agreed -- if we're using discussions, best to remove the comments to editor column!)

ajnyga:
@bozana here are the changes you suggested.

I also edited submission.inc.php and removed set/get for the comments to editor field. 

I will open a new branch in OJS for the upgrade script changes and do another pr for that. What I do not know is how to handle the removal of the database table field during the upgrade. But I will start with the upgrade script.

bozana:
@ajnyga, to remove the column from the DB table you would remove it from the table schema definition i.e. from here: https://github.com/pkp/ojs/blob/master/dbscripts/xml/ojs_schema.xml#L450
In the main upgrade script that schema is called here: https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L68
I.e. your script/function, that moves the existing comments into the discussions, would have to be called before that line in the upgrade.xml, because then the column will be removed. I.e. somewhere here https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L42 there should be something like:
```
<upgrade minversion="3.0.1.0" maxversion="3.0.1.9">
<data file="dbscripts/xml/upgrade/3.0.2_preupdate.xml" />
</upgrade>

```
or if it is a function in the Upgrade.inc.php:

```
<upgrade minversion="3.0.1.0" maxversion="3.0.1.9">
<code function="moveCommentsToEditor" />
</upgrade>
```

bozana:
Ah sorry @ajnyga, now I realized that this should be also done for the migration from OJS 2.4.x, as well as for 3.0.0 and 3.0.1. Thus: here the `articles` table is renamed into the `submissions` table: https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L29. I.e. your script could work only with `submissions` table, because it would apply and be the same for all versions, the user are migrating from, 2.4.x, 3.0.0 and 3.0.1. It must be then positioned after the `3.0.0_preupdate.xml` call and before the before the `ojs_schema.xml` call (that will remove the column). 
For 2.4.x migration you could position the function call here: https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L41
And for 3.0.0 and 3.0.1 you could add somewhere here https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L42 something like:
```
<upgrade minversion="3.0.0.0" maxversion="3.0.1.9">
<code function="moveCommentsToEditor" />
</upgrade>
```
THANKS!!!

ajnyga:
I think that all the above is now ready here: https://github.com/pkp/ojs/pull/1228

Changing to use the submissions table was harder than expected because that table does not have the user_id available anymore. So feel free to criticize my sql there :D

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -204,6 +201,15 @@ function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $update
...
 
 	}
+	
+	/**
+	 * Get comments to editor
+	 */
+	function getCommentsToEditor($submissionId) {
+		$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $submissionId);
+		$query = $queries->next();
+		return $query;
```
bozana: what will be returned if there is no such query? null?

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -225,10 +231,9 @@ function execute($args, $request) {
 			}
 			// Add or update comments to editor
 			if ($this->getData('commentsToEditor')){
-				$queries = $queryDao->getByAssoc(ASSOC_TYPE_SUBMISSION, $this->submissionId);
-				$query = $queries->next();
+				$query = $this->getCommentsToEditor($this->submissionId);				
 				if (isset($query)){
```
bozana: I believe you do not need this `if-else` -- you can just call the `setCommentsToEditor` with the `$query`, because the `$query` is then checked there.

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -204,6 +201,15 @@ function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $update
 		}
 
 	}
+	
+	/**
+	 * Get comments to editor
+	 */
```
bozana: `@param` and `@return` function description would be nice...

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -156,18 +157,17 @@ function readInputData() {
 	 */
 	function setSubmissionData($submission) {
 		$this->submission->setLanguage(PKPString::substr($this->submission->getLocale(), 0, 2));
-		$this->submission->setCommentsToEditor($this->getData('commentsToEditor'));
 		$this->submission->setLocale($this->getData('locale'));
 	}
 
 	/**
 	 * Add or update comments to editor
 	 */
```
bozana: @param and @return function description would be nice...

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -156,18 +157,17 @@ function readInputData() {
...
-	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $update = null) {
+	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $query) {
```
bozana: you could set $query = null per default, or?

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -111,9 +111,10 @@ function fetch($request) {
 	 */
 	function initData($data = array()) {
 		if (isset($this->submission)) {
+			$query = $this->getCommentsToEditor($this->submissionId);
```
bozana: `$query` could be `null` here, I believe, so the method call below (`$query->getHeadNote()`) cannot be done, or? Maybe then to use something like `$query ? $query->getHeadNote() : ''` below?

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
```diff
@@ -163,7 +163,7 @@ function setSubmissionData($submission) {
 	/**
 	 * Add or update comments to editor
 	 */
-	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $query) {
+	function setCommentsToEditor($submissionId, $commentsToEditor, $userId, $query = null) {
```
bozana: ah, I see it is null per default here, so forget the comment from the commit before... :-)

classes/submission/Submission.inc.php
```diff
@@ -233,14 +233,6 @@ function setLicenseURL($licenseURL) {
 	}
 
 	/**
-	 * Get comments to editor.
-	 * @return string
-	 */
-	function getCommentsToEditor() {
```
bozana: This function is called on a few other places as well, but I can do that once this PR is merged:
s. `classes/article/ArticleDAO.inc.php` and `lib/pkp/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php`

classes/submission/Submission.inc.php
```diff
@@ -257,14 +249,6 @@ function setHideAuthor($hideAuthor) {
 	}
 
 	/**
-	 * Set comments to editor.
-	 * @param $commentsToEditor string
-	 */
-	function setCommentsToEditor($commentsToEditor) {
```
bozana: The same here -- this is a note for me: 
s. `lib/pkp/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php` and `lib/pkp/classes/submission/SubmissionDAO.inc.php`

bozana:
@ajnyga, thanks a lot! Just a few comments here... else everything fine :-)

ajnyga:
thanks, I am doing the changes. Probably it would had been quicker for you just do the changes youself, but I appreciate the comments!

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: of course...

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: check

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: check

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: check, return is still empty?

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: check

classes/submission/form/PKPSubmissionSubmitStep1Form.inc.php
ajnyga: check

ajnyga:
I think that this is safe to close now @bozana @asmecher ?

I just upgraded to 3.0.2. Everything went smoothly, including this comments bit, thank you for your hard work!

bozana:
Yes, please do. Also the other related PR (for upgrade) :-) Thanks a lot!!!

Issue #850, opened 2015-10-29
Implement CRUD API for DOAJ


asmecher:
See https://doaj.org/api/v1/docs for details.

asmecher:
@bozana, I think you were interested in working on this.

ajnyga:
@bozana, I was planning of doing something similar as well. Basically I was thinking of taking of course the current DOAJ plugin as a starting point and maybe check the functionality in the Crossref plugin if the automatic deposits function could be used or at least some part of it.

bozana:
@ajnyga, great! Yes, I think you could orient yourself on Crossref or DataCite or mEDRA plugin. All of hem share a lot of code/functions and it is similar with DOAJ. Crossref has a slightly different registration -- more complex. When would you start to work on this? I am happy to help!

ajnyga:
I already had to postpone this once, but I have it scheduled now for next week. I could perhaps have something ready on Tuesday. I will let you know.

bozana:
@ajnyga, I was hoping to work on this on the PKP sprint, that is tomorrow and the day after tomorrow, but lets see if I will be able to or if I will chose to work on another topic. However, I restructured the shared code for those export plugins, s. https://github.com/bozana/ojs/commit/13d513e16affbebba0ea7d2b5e73dcffb02dad0a. Thus, the next and major step would be to implement the depositXML function in DOAJExportPlugin.inc.php, s. https://github.com/bozana/ojs/blob/850fix/plugins/importexport/doaj/DOAJExportPlugin.inc.php#L95.
As far as I could quickly see, the DOAJ API expects a JSON, that has to be constructed. Eventually we could use the filter framework for that, similar to the filter for the export in XML (s. https://github.com/bozana/ojs/tree/850fix/plugins/importexport/doaj/filter), but I am not sure i.e. will have to take a further look.
When you find some time to work on this, please use my 850fix branch (https://github.com/bozana/ojs/tree/850fix) to build on that. If you eventually have some time the next two days, during the sprint, you could participate remotely i.e. we could eventually work together on this. Else, I will keep you informed about any further step I undertake.

ajnyga:
Hi @bozana,

I am planning to participate in the sprint. However, due to the time difference it is a bit difficult for me to work at the same time as most of the sprinters. It is 4 already 4 PM here when the work starts in Canada. So, maybe I could suggest the DOAJ API as a project to the sprint in a while and do my work while you guys sleep.

[...]

PR #2214, opened 2017-01-20
doaj buttons css

bozana:
s. https://github.com/pkp/pkp-lib/issues/850

### DIFF
--- a/styles/pages/workflow.less
+++ b/styles/pages/workflow.less
@@ -121,11 +121,11 @@
 	}
 }
 
-.pubids_export_actions {
+.export_actions {
 	padding-left: @double;
 	padding-right: @double;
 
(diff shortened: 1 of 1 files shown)


styles/pages/workflow.less
```diff
@@ -121,11 +121,11 @@
 	}
 }
 
-.pubids_export_actions {
+.export_actions {
 	padding-left: @double;
 	padding-right: @double;
 
-	.pubids_export_action {
+	.export_action {
```
asmecher: Hmm, we're going to have trouble modifying the .less files unless we cause the CSS cache to be removed on upgrade. (The CSS isn't recompiled unless the cache is manually deleted.) Mind adding an upgrade task to remove the cached CSS?

PR #2236, opened 2017-01-25
Initial commit for Finnish locale in pkp-lib

ajnyga:
Finnish locale for pkp-lib, tagging @mtub


ajnyga:
We did the translation using Memsource. It seems that the export function there made those changes to the header. I will do another commit and fix these tomorrow.

ajnyga:
@mtub headers should be ok now

mtub:
Thank you, @ajnyga !

PR #2239, opened 2017-01-26
Translation updates

mtub:
Updates from our translators, already reviewed and slightly corrected.

* it_IT
* nb_NO

Some fields contain English text that apparently was entered intentionally (group names like "layout editor")


asmecher:
Merged -- thanks, @mtub!

Issue #2249, opened 2017-01-30
[OJS] OJS 3.0.1 Add reviewer modal not loading


ajnyga:
Hi,

One of our journals has an submission that has been made using OJS 3.0.1. They have started the review stage, but can not assign reviewers because the add reviewer modal is not responding. Firefox asks if you want to continue loading the script, but that does not help.

There are no errors in php error_log on no errors in javascript log. It seems that the script is stuck here:
![screenshot_94](https://cloud.githubusercontent.com/assets/16347527/22415439/1552fd8c-e6d0-11e6-83f4-83561cc53e2f.png)

Any idea how to debug this?

ajnyga:
It seems to be something in the submission abstract that is breaking the script, because removing the abstract solves the problem. I will test what string is the reason and let you know.

ajnyga:
Tagging @NateWr and @asmecher 

Short story short: If you have a string like `<15` in the abstract the above problem with adding reviewers will emerge. This does not happen if you have `< 15`.

No idea why.

ajnyga:
This is maybe the same issue: 
https://github.com/tinymce/tinymce/issues/2567

asmecher:
@ajnyga, is this an abstract that was entered into the system via the TinyMCE GUI (and thus should be properly escaped) or came in some other way (thus might not be properly escaped)?

Check in the database to see whether that abstract has the `<` encoded as `<`, or as `&lt;`.

ajnyga:
Hi,

This is an abstact that was sent among the first ones after we moved to OJS3.
If I look at the source with tinymce, I see "&lt;"

Spocke in the above tinymce thread writes that "Properly encode your content. &amp;lt; not &lt; I guess we need to add some docs somewhere about how to entity encode HTML this issue seems to be reported from time to time."

ajnyga:
edit: I checked the generated source at the metadata page and it has "`&amp;lt;15`"

But as I said, although I have entered that using tinymce and it seems to be escaped correctly, it is still causing the script mentioned above to time out.

ajnyga:
Another detail. If there is not much content after `<15` you do not get this problem. But if there is for example a full paragraph of text following that, the script times out.

I was able to replicate this will a clean install of 3.0.1.

ajnyga:
@asmecher,
The html variable that the loop has contains the review request letter, which in turn contains the abstract. This is what is sent to that script: 

[...]

PR #2251, opened 2017-01-31
remove html2text from SubmissionMailTemplate

ajnyga:
Described here: https://github.com/pkp/pkp-lib/issues/2249

Using `PKPString::stripUnsafeHtml` insted, not sure if necessary, but no harm I guess.

Also, maybe worth checking other uses as well, like https://github.com/pkp/pkp-lib/blob/master/classes/mail/Mail.inc.php ?


asmecher:
The use of `html2text` in `Mail.inc.php` is fine. Merged -- thanks as always, @ajnyga!

Issue #1955, opened 2016-11-04
[OJS] 3.0.0 schemaValidate, proxy settings and allow_url_fopen


knjigor:
Refer to http://forum.pkp.sfu.ca/t/ojs-3-crossref-error/23770/14?u=knjigor

bozana:
For XML validation, the PHP method `schemaValidate` is used. It gets the Crossref schema URL as parameter. I don't know how this method then uses the URL and how to solve the proxy problem :-( @asmecher, do you think `stream_context_set_default` (like for example described here) would work? 
Or, does anybody know how we could download localy all the "millions" ;-) schemas, that the Crossref schema includes?

asmecher:
It looks like `stream_context_create` is one way to do it -- looking at http://php.net/manual/en/context.http.php -- but it would need some testing. Another option would be to add an option to disable validation completely -- this would work around this problem but also would be useful for other purposes, such as production deployment if/where validation was too time-consuming. But I would rather we didn't mirror all the CrossRef stuff locally.

bozana:
The option to disable validation would be also good if the `allow_url_fopen = Off` -- we use schemas URLs, so the `schemaValidate` uses `fopen` to get the schema. If `allow_url_fopen` is disabled, this will not work. S. http://forum.pkp.sfu.ca/t/ojs3-0-2-and-crossref-plugin-validation-error/28369/10.

bozana:
PRs:
pkp-lib ojs-stable-3_0_2: https://github.com/pkp/pkp-lib/pull/2264
ojs ojs-stable-3_0_2: https://github.com/pkp/ojs/pull/1248

bozana:
@asmecher, could you please take a look at the PRs above?

bozana:
Everything merged, thus closing...

PR #2264, opened 2017-02-06
pkp/pkp-lib#1955 no XML validation option for the export plugins

bozana:
s. https://github.com/pkp/pkp-lib/issues/1955
this PR introduces the option not to validate XML when exporting...
A note for an easier code review: The function `isCompatible` is copied from the parent class with the only difference that she is calling the `checkType` with the new parameter

### DIFF
--- a/classes/xslt/XMLTypeDescription.inc.php
+++ b/classes/xslt/XMLTypeDescription.inc.php
@@ -60,6 +60,13 @@ class XMLTypeDescription extends TypeDescription {
 		return TYPE_DESCRIPTION_NAMESPACE_XML;
 	}
 
+	/**
+	 * Set the validation strategy
+	 * @param $validationStrategy string XML_TYPE_DESCRIPTION_VALIDATE_...
+	 */
+	function setValidationStrategy($validationStrategy) {
+		$this->_validationStrategy = $validationStrategy;
+	}
 
 	//
 	// Implement abstract template methods from TypeDescription
--- a/plugins/importexport/native/filter/NativeExportFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeExportFilter.inc.php
@@ -16,6 +16,11 @@
 import('lib.pkp.plugins.importexport.native.filter.NativeImportExportFilter');
 
 class NativeExportFilter extends NativeImportExportFilter {
+
+	/** @var boolean If set to true no validation (e.g. XML validation) will be done */
+	var $_noValidation = null;
+
+
 	/**
 	 * Constructor
 	 * @param $filterGroup FilterGroup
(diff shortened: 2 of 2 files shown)


classes/xslt/XMLTypeDescription.inc.php
```diff
@@ -98,14 +141,18 @@ function parseTypeName($typeName) {
...
+	 * @return boolean
 	 */
-	function checkType(&$object) {
+	function checkType(&$object, $noValidation = null) {
 		// We only accept DOMDocument objects and source strings.
 		if (!is_a($object, 'DOMDocument') && !is_string($object)) return false;
 
 		// No validation...
-		if ($this->_validationStrategy == XML_TYPE_DESCRIPTION_VALIDATE_NONE) return true;
+		if ($this->_validationStrategy == XML_TYPE_DESCRIPTION_VALIDATE_NONE || $noValidation) return true;
```
asmecher: Is there some way to reuse the `_validationStrategy` member variable rather than adding a new validation flag?

bozana: maybe I will take a deeper look... But I thought that it is maybe good to differentiate the defined strategy (I suppose in the filter config) and exception i.e. per case option?

bozana: @asmecher, it would be possible to set the validationStrategy to `XML_TYPE_DESCRIPTION_VALIDATE_NONE` before calling the `isCompatible` function in the `NativeExportFilter`. This would reduce all the changes in pkp-lib -- only `setValidationStrategy` would be needed -- thus I suppose it is better to do it that way, right? I will then change it this way, OK?

asmecher: Yes, I think that's a little cleaner, thanks!

bozana: yup, yup, working... coming...

bozana:
@asmecher, changed, OK now? :-) THANKS!!!

plugins/importexport/native/filter/NativeExportFilter.inc.php
```diff
@@ -24,6 +29,46 @@ function __construct($filterGroup) {
...
+		return $this->_noValidation;
+	}
+
+	//
+	// Public methods
+	//
+	/**
+	 * @copydoc Filter::supports()
+	 */
+	function supports(&$input, &$output) {
```
asmecher: Is this function still needed?

bozana: Yes, I think so -- it sets the validation strategy to none, if the noValidation option is used...

PR #2293, opened 2017-02-16
pkp/pkp-lib#1709 consider query files and source attribute, and use the same file id for all revisions imports

bozana:
s. b1), b2) and b3) here: https://github.com/pkp/pkp-lib/issues/1709

### DIFF
--- a/classes/plugins/importexport/PKPImportExportDeployment.inc.php
+++ b/classes/plugins/importexport/PKPImportExportDeployment.inc.php
@@ -212,9 +212,20 @@ class PKPImportExportDeployment {
 	 * @param $revisionId integer
 	 * @return integer
 	 */
-	function getFileDBId($fileId, $revisionId) {
-		if (array_key_exists($fileId, $this->_fileDBIds) && array_key_exists($revisionId, $this->_fileDBIds[$fileId])) {
-			return $this->_fileDBIds[$fileId][$revisionId];
+	function getFileDBId($fileId, $revisionId = null) {
+		if (array_key_exists($fileId, $this->_fileDBIds)) {
+			// is there already the revisionId?
+			if ($revisionId) {
+				if (array_key_exists($revisionId, $this->_fileDBIds[$fileId])) {
+					return $this->_fileDBIds[$fileId][$revisionId];
+				} else {
+					return null;
+				}
+			} else {
+				// the revisionId is not important, but the fileId
+				// the DB Id is unique for a fileId
+				return current($this->_fileDBIds[$fileId]);
+			}
 		}
 		return null;
 	}
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
@@ -116,6 +116,12 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 
 		$revisionId = $node->getAttribute('number');
 
+		$source = $node->getAttribute('source');
+		$sourceFileAndRevision = null;
+		if ($source) {
+			$sourceFileAndRevision = explode('-', $source);
+		}
+
 		$genreId = null;
 		$genreName = $node->getAttribute('genre');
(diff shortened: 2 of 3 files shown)


classes/plugins/importexport/PKPImportExportDeployment.inc.php
```diff
@@ -212,9 +212,14 @@ function setFileDBIds($fileDBIds) {
...
-		if (array_key_exists($fileId, $this->_fileDBIds) && array_key_exists($revisionId, $this->_fileDBIds[$fileId])) {
-			return $this->_fileDBIds[$fileId][$revisionId];
+	function getFileDBId($fileId, $revisionId = null) {
+		if ($revisionId) {
+			if (array_key_exists($fileId, $this->_fileDBIds) && array_key_exists($revisionId, $this->_fileDBIds[$fileId])) {
+				return $this->_fileDBIds[$fileId][$revisionId];
+			}
+		} else {
+			// the DB Id is unique for a fileId
+			return current($this->_fileDBIds[$fileId]);
```
asmecher: It looks like "index not found" checking is present when revisions are specified, but not when only the file ID is specified. I'd suggest keeping consistent either way.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -189,11 +195,22 @@ function handleRevisionElement($node, $stageId, $fileId) {
...
+			}
+		}
+
 		if ($errorOccured) {
 			// if error occured, the file cannot be inserted into DB, becase
 			// genre, uploader and user group are required (e.g. at name generation).
 			$submissionFile = null;
 		} else {
+			// if the same file is already inserted, take its DB file ID
+			$DBId = $deployment->getFileDBId($fileId);
```
asmecher: It would be equivalent here if we included the `$revisionId` in this call, correct? If so, that would mean it's possible to remove the special case this PR introduces to `getFileDBId`.

bozana: Yes, you are right... :-)

bozana:
@asmecher, I made the suggested changes in the second commit above... OK so? THANKS!!!

classes/plugins/importexport/PKPImportExportDeployment.inc.php
```diff
@@ -213,8 +213,13 @@ function setFileDBIds($fileDBIds) {
 	 * @return integer
 	 */
 	function getFileDBId($fileId, $revisionId) {
```
asmecher: I think the changes to this function are no longer necessary, correct?

bozana: I think they are -- earlier, the result was returned only if there was an entry for a matching fileId and revisionId. Now, also a case where only matching fileId is found is considered -- that case is when there are more than one revision for a file to be imported -- earlier each revision was inserted as a new file i.e. with a new fileId.

bozana: I realized that we should consider revisionId parameter = null and return null in case the revisionId parameter is not null and DBId not found. This would check if that specific revision is inserted. If revisionId parameter is null and fileId exists the current DBId should be returned. Thus I will change it so...

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
bozana: I realized that this is not right -- we should be able to a) look if a revision is already inserted and b) to look only after the new file id. In case a) we have to provide revisionId parameter and the only matching value should be returned and if not found null. In case of b) we do not care about revisionID, we only want the new DB file ID. OK? The final changes consider all this...

bozana:
@asmecher, there are final changes, I think :-) See my last comments... THANKS!!! SORRY!!!

asmecher:
@bozana -- looks good, please go ahead (and sorry for the wild goose chase).

PR #2298, opened 2017-02-17
add additional instructions to roleBasedAccessDenied message for en_U…

ajnyga:
…S and fi_FI

Described here https://github.com/pkp/pkp-lib/issues/2297 and here http://forum.pkp.sfu.ca/t/ojs3-missing-author-roles-in-multi-journal-installations/28710

### DIFF
--- a/locale/en_US/user.xml
+++ b/locale/en_US/user.xml
@@ -22,7 +22,7 @@
 	<message key="user.authorization.loginRequired">Login required.</message>
 	<message key="user.authorization.privateOperation">The operation you tried to access is either private or doesn't exist.</message>
 	<message key="user.authorization.restrictedSiteAccess">You have been denied access. Access to this site is currently restricted.</message>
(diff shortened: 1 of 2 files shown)


locale/en_US/user.xml
```diff
@@ -22,7 +22,7 @@
 	<message key="user.authorization.loginRequired">Login required.</message>
 	<message key="user.authorization.privateOperation">The operation you tried to access is either private or doesn't exist.</message>
 	<message key="user.authorization.restrictedSiteAccess">You have been denied access. Access to this site is currently restricted.</message>
-	<message key="user.authorization.roleBasedAccessDenied">The current role does not have access to this operation.</message>
+	<message key="user.authorization.roleBasedAccessDenied"><![CDATA[The current role does not have access to this operation. <a href="profile/roles/">You can edit your roles here &raquo;</a>]]></message>
```
asmecher: This relative URL may or may not work (e.g. with installations that use the `path_info_disabled` config option) -- it's best to have all URL generation done on the PHP side.

PR #2307, opened 2017-02-22
pkp/pkp-lib#2072 Versioning for published articles

lilients:
See https://github.com/pkp/pkp-lib/issues/2072.


classes/submission/Submission.inc.php
```diff
@@ -248,6 +248,50 @@ function setHideAuthor($hideAuthor) {
...
+
+	/**
+	 * Get the current revision id for a submission
+	 * @param $contextId int
+	 * @return int
+	 */
+	function getCurrentVersionId($contextId = null) {
+		$submissionId = $this->getId();
+
+		if (!ctype_digit("$submissionId")) {
```
bozana: I think the $submissionId here (got with getId()) should always be a number

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+
+		return $submissionRevisions;
+	}
+
+	/**
+	 * Get all published revisions for a submission
+	 * @param $submissionId int
+	 * @return array
+	*/
+	function getPublishedSubmissionRevisions($submissionId, $contextId = null, $order = SORT_DIRECTION_DESC){
```
bozana: because this considers only published submissions, it should probably be places in the PublishedArticleDAO?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+		return array_pop($submissionRevisions);
+	}
+
+	/**
+	 * Get the current published revision id for a submission
+	 * @param $submissionId int
+	 * @param $contextId int
+	 * @return int
+	 */
+	function getLatestPublishedRevisionId($submissionId, $contextId = null) {
```
bozana: This function seem to be used only one time, in pages/article/ArticleHandler.inc.php, where the function getPublishedSubmissionRevisions is also called before this function... thus I think we do not need this function for now and can do array_pop in ArticleHandler

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+	 * Get all published revisions for a submission
+	 * @param $submissionId int
+	 * @return array
+	*/
+	function getPublishedSubmissionRevisions($submissionId, $contextId = null, $order = SORT_DIRECTION_DESC){
+		$submissionRevisions = $this->getSubmissionRevisionIds($submissionId, $contextId, $order);
+		$publishedSubmissionRevisions = array();
+
+		foreach($submissionRevisions as $revision){
+			$submission = $this->getById($submissionId, null, false, $revision);
```
bozana: submissionId can also be string i.e. public identifier, thus other function should be used instead of getById

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
 		$submission->setId($row['submission_id']);
+		$submission->setSubmissionRevision($submissionRevision);
 		$submission->setContextId($row['context_id']);
 		$submission->setLocale($row['locale']);
 		$submission->setStageId($row['stage_id']);
 		$submission->setStatus($row['status']);
 		$submission->setSubmissionProgress($row['submission_progress']);
 		$submission->setDateSubmitted($this->datetimeFromDB($row['date_submitted']));
 		$submission->setDateStatusModified($this->datetimeFromDB($row['date_status_modified']));
-		$submission->setDatePublished($this->datetimeFromDB($row['date_published']));
```
bozana: This is necessary, because the getDatePublished on a submission objects is often use. For all changes in pkp-lib we also have to keep OMP in mind. I would suggest keeping this and below in getById...

classes/submission/SubmissionDAO.inc.php
```diff
@@ -340,7 +422,7 @@ function getById($submissionId, $contextId = null, $useCache = false) {
 		if ($contextId) $params[] = (int) $contextId;
 
 		$result = $this->retrieve(
-			'SELECT	s.*, ps.date_published,
+			'SELECT	s.*,
```
bozana: it would be good to leave this here, in order to be able to set date published for submissions in _fromRow, else the function getDatePublished cannot be used any more.
Thus maybe to get the setting_name = 'datePublished' for the $submissionRevision, if $submissionRevision does not exist than for revision = 1.

classes/db/DAO.inc.php
```diff
@@ -563,16 +563,24 @@ function updateDataObjectSettings($tableName, $dataObject, $idArray) {
...
+		$params = array();
+		if (($idFieldName !== null) && ($revision == null)) {
+			$params = array_merge($params, array((int) $idFieldValue));
+		} elseif (($idFieldName !== null) && ($revision !== null)) {
+			$params = array_merge($params, array((int) $idFieldValue, (int) $revision));
 		}
+
+		$sql = "SELECT * FROM $tableName" .
+		((($idFieldName == null) && ($revision == null)) ? "" : " WHERE $idFieldName = ?" .
+		((($idFieldName !== null) && ($revision !== null)) ? " AND $revisionFieldName = ?" : ""));
```
bozana: Do we not need WHERE $idFieldName = ? here?

classes/submission/Submission.inc.php
```diff
@@ -248,6 +248,50 @@ function setHideAuthor($hideAuthor) {
 	}
 
 	/**
+	 * Set comments to editor.
+	 * @param $commentsToEditor string
+	 */
+	function setCommentsToEditor($commentsToEditor) {
```
bozana: Why this function?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+	 * @param $submissionId int
+	 * @return array
+	*/
+	function getPublishedSubmissionRevisions($submissionId, $contextId = null, $order = SORT_DIRECTION_DESC){
+		$submissionRevisions = $this->getSubmissionRevisionIds($submissionId, $contextId, $order);
+		$publishedSubmissionRevisions = array();
+
+		foreach($submissionRevisions as $revision){
+			$submission = $this->getById($submissionId, null, false, $revision);
+			if($submission->getDatePublished()){
```
bozana: getDatePublished does not contain the correct data, because it is not set within _fromRow...

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -250,12 +299,24 @@ function canAdminister($user) {
 	}
 
 	/**
+	 * Fetches the application-specific submission id from the request object.
+	 * Should be overridden by subclasses.
+	 * @param PKPRequest $request
+	 * @return int
+	 */
+	function getRequestedSubmissionId($request) {
```
bozana: this function is used only once, in this class, thus maybe not to create it?

classes/submission/Representation.inc.php
```diff
@@ -85,6 +85,33 @@ function getSubmissionId() {
...
+	function getSubmissionRevision() {
+		return $this->getData('submissionRevision');
+	}
+
+	/**
+	 * Get the current revision id for a galley from submission
+	 * @param $contextId int
+	 * @return int
+	 */
+	function getCurrentVersionId($contextId = null) {
```
asmecher: Hmm, I wonder whether relying on this to fetch the current revision is going to be too expensive in performance. I think it'll be possible to have a fairly standard fetch function that supports revisions e.g. like this:

    function getActiveSubmissions($contextId, ..., $revision = REVISION_LATEST) {
    }

...where `define('REVISION_LATEST', 0);`

Then I'd suggest adding a column e.g. to the `articles` table called `current_revision`. This would be really fast and easy to fetch and wouldn't impose much maintenance overhead.

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,204 @@
...
+		// save new submission version
+		$submissionDao->updateObject($submission);
+
+		// copy the authors from old version to new version
+		foreach($authors as $author) {
+			$authorId = (int)$author->getId();
+			$author->setVersion($submissionRevision);
+			$authorDao->insertObject($author, true);
+			$newAuthorId = (int)$authorDao->getInsertId();
+			$authorDao->update('UPDATE authors SET author_id = ? WHERE author_id = ?', array($authorId, $newAuthorId));
```
bozana: why is this necessary? should insert not do that? also, SQL statemets should always go to DAOs

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,204 @@
...
+			// copy file and link copy to new galley
+			if($galley->getFile()){
+				$context = $request->getContext();
+				$oldFile = $galley->getFile();
+				$fileStage = $oldFile->getFileStage();
+				$newFileId = $this->copyFile($context, $oldFile, $fileStage);
+				$galley->setFileId($newFileId);
+			}
+			// save new galley version
+			$articleGalleyDao->updateObject($galley);
```
bozana: hmmm... why do we update the whole object? could we not only add new files relations?

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,204 @@
...
+		// Add variables to the template.
+		$templateMgr = TemplateManager::getManager($request);
+		$templateMgr->assign('submission', $submission);
+		$templateMgr->assign('stageId', $stageId);
+		$templateMgr->assign('submissionRevision', $submissionRevision);
+
+		// Get publication date of submission revision 
+		$submissionDao = Application::getSubmissionDAO();
+		$submission = $submissionDao->getById($submission->getId(), null, false, $submissionRevision);
+		$isPublished = $submission->getDatePublished() ? false : true;
```
bozana: is this correct: if there is datePublished then false?

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -94,12 +95,18 @@ function initData($submission) {
 			$submissionAgencyDao = DAORegistry::getDAO('SubmissionAgencyDAO');
 			$submissionLanguageDao = DAORegistry::getDAO('SubmissionLanguageDAO');
 
+			// versioning
+			$contextId = Request::getContext()->getId();
```
bozana: Again, context can be retrieved from the submission

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -94,12 +95,18 @@ function initData($submission) {
...
+			$latestRevisionId = $submission->getCurrentVersionId($contextId);
+			if (!$revision) $revision = $latestRevisionId;
```
bozana: is this revision variable used in this function?

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
```diff
@@ -147,6 +154,18 @@ function execute($submission, $request) {
 		if (empty($supportedSubmissionLocales)) $supportedSubmissionLocales = array($context->getPrimaryLocale());
 		if (in_array($newLocale, $supportedSubmissionLocales)) $submission->setLocale($newLocale);
 
+		// Versioning
+		if ($request->getUserVar('submissionRevision')) {
+			$revision = (int)$request->getUserVar('submissionRevision');
+		} else {
+			$context = $request->getContext();
+			$contextId = $context->getId();
+			$revision = $submission->getCurrentVersionId($contextId);
```
bozana: submission function do not need context parameter, because every submission has it already

classes/submission/Submission.inc.php
```diff
@@ -248,6 +248,50 @@ function setHideAuthor($hideAuthor) {
...
+	function getSubmissionRevision() {
+		return $this->getData('submissionRevision');
+	}
+
+	/**
+	 * Get the current revision id for a submission
+	 * @param $contextId int
+	 * @return int
+	 */
+	function getCurrentVersionId($contextId = null) {
```
bozana: contextId parameter is not needed, submission object already has it -- always belong to a context

classes/submission/Submission.inc.php
```diff
@@ -248,6 +248,50 @@ function setHideAuthor($hideAuthor) {
...
+		$submissionId = $this->getId();
+
+		if (!ctype_digit("$submissionId")) {
+			$publishedArticleDao = DAORegistry::getDAO('PublishedArticleDAO');
+			$publishedArticle = $publishedArticleDao->getPublishedArticleByPubId('publisher-id', $submissionId, $contextId);
+			$submissionId = $publishedArticle->getId();
```
bozana: why submissionId again?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+		$sql = 'SELECT DISTINCT submission_revision FROM submission_settings ' .
+			   'INNER JOIN submissions ON submission_settings.submission_id = submissions.submission_id' .
+			   ' WHERE submission_settings.submission_id = ?'  .
+			   ($contextId ? ' AND submissions.context_id = ?' : '') .
+			   ' ORDER BY submission_revision ' . $this->getDirectionMapping($order);
+
+		$result = $this->retrieve($sql, $params);
+		$submissionRevisions = array();
+
+		foreach($result->getArray() as $submission) {
```
bozana: using array seems to need more memory, thus although these are not huge amount of data, maybe not to convert it to array. also in order to stay consistent...

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -66,6 +69,44 @@ function setReadOnly($readOnly) {
...
+	/**
+	 * Get the newest version
+	 * @param $request PKPRequest
+	 * @return int
+	 */
+	function getNewestVersion($request) {
+		$submission = $this->getSubmission();
+		$context = $request->getContext();
+		$contextId = $context->getId();
+		$submissionVersion = $submission->getCurrentVersionId($contextId);
```
bozana: contextId not needed...

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -196,8 +244,9 @@ function getDataElementSequence($gridDataElement) {
 	function setDataElementSequence($request, $rowId, $gridDataElement, $newSequence) {
 		$authorDao = DAORegistry::getDAO('AuthorDAO');
 		$submission = $this->getSubmission();
-		$author = $authorDao->getById($rowId, $submission->getId());
+		$author = $authorDao->getById($rowId, $submission->getId(), $this->_version);
 		$author->setSequence($newSequence);
+		$author->setVersion($this->_version);
```
bozana: is this here necessary?

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -280,11 +341,12 @@ function addAuthor($args, $request) {
 	 */
 	function editAuthor($args, $request) {
 		// Identify the author to be updated
-		$authorId = (int) $request->getUserVar('authorId');
+		$authorId = $request->getUserVar('authorId');
+		$version = $request->getUserVar('version');
```
bozana: maybe to keep the (int) cast
this i haven't concentrate on till now :-\

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -357,12 +419,15 @@ function updateAuthor($args, $request) {
 	 * @return JSONMessage JSON object
 	 */
 	function deleteAuthor($args, $request) {
-		if (!$request->checkCSRF()) return new JSONMessage(false);
+		// Identify the submission Id
+		$submissionId = $this->getRequestedSubmissionId($request);
+		// Identify the author to be deleted
+		$authorId = $request->getUserVar('authorId');
```
bozana: (int) cast

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -302,7 +364,7 @@ function editAuthor($args, $request) {
 	 */
 	function updateAuthor($args, $request) {
 		// Identify the author to be updated
-		$authorId = (int) $request->getUserVar('authorId');
+		$authorId = $request->getUserVar('authorId');
```
bozana: (int) cast

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -357,12 +419,15 @@ function updateAuthor($args, $request) {
 	 * @return JSONMessage JSON object
 	 */
 	function deleteAuthor($args, $request) {
-		if (!$request->checkCSRF()) return new JSONMessage(false);
+		// Identify the submission Id
+		$submissionId = $this->getRequestedSubmissionId($request);
+		// Identify the author to be deleted
+		$authorId = $request->getUserVar('authorId');
+		// Identify the version
+		$version = $request->getUserVar('version');
```
bozana: (int) cast

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -357,12 +419,15 @@ function updateAuthor($args, $request) {
 	 * @return JSONMessage JSON object
 	 */
 	function deleteAuthor($args, $request) {
-		if (!$request->checkCSRF()) return new JSONMessage(false);
```
bozana: why removing this?

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -374,7 +439,7 @@ function deleteAuthor($args, $request) {
 	 */
 	function addUser($args, $request) {
 		// Identify the author Id.
-		$authorId = (int) $request->getUserVar('authorId');
+		$authorId = $request->getUserVar('authorId');
```
bozana: (int) cast

controllers/grid/users/author/form/AuthorForm.inc.php
```diff
@@ -214,6 +215,9 @@ function execute() {
 		$author->setBiography($this->getData('biography'), null); // localized
 		$author->setPrimaryContact(($this->getData('primaryContact') ? true : false));
 		$author->setIncludeInBrowse(($this->getData('includeInBrowse') ? true : false));
+		$contextId = Request::getContext()->getId();
+		$newestVersion = $submission->getCurrentVersionId($contextId);
```
bozana: contextId not needed

controllers/tab/workflow/PKPWorkflowTabHandler.inc.php
```diff
@@ -131,13 +131,58 @@ function fetchTab($args, $request) {
...
 				$templateMgr->assign('representations', $representations->toAssociativeArray());
 
 				$templateMgr->assign('productionNotificationRequestOptions', $notificationRequestOptions);
+
+				// Versioning
+				$context = $request->getContext();
+				$templateMgr->assign('versioningEnabled', $context->getSetting('versioningEnabled'));
+
+				// get all revisions of this submission
+				$articleDao = DAORegistry::getDAO('ArticleDAO');
```
bozana: application specific classes (articleDao) should/can not be used in pkp-lib, but Application::getSubmissionDao() or so...

controllers/tab/workflow/PKPWorkflowTabHandler.inc.php
```diff
@@ -131,13 +131,58 @@ function fetchTab($args, $request) {
...
+				$submissionRevisions = $articleDao->getSubmissionRevisionIds($submission->getId(), $context->getId());
+				$templateMgr->assign('submissionRevisions', $submissionRevisions);
+
+				// get latest submission revision
+				$latestSubmissionRevision = $articleDao -> getLatestRevisionId($submission->getId(), $context->getId());
```
bozana: space around ->

locale/en_US/submission.xml
```diff
@@ -647,4 +647,22 @@
...
+	<message key="submission.versioning.version">Version {$version} ({$date})</message>
+	<message key="submission.versioning.linkToLatest">This is a previous version. See </message>
+	<message key="submission.versioning.latest">latest version</message>
+
+	<message key="submission.versioning.previousVersions">Previous Versions of this Article</message>
+
+	<message key="submission.production.version">Version {$submissionRevision}</message>
+	<message key="submission.production.newVersion">New Version</message>
+	<message key="submission.production.editMetadata">Edit Metadata of this Version</message>
+	<message key="submission.production.editMetadata.info">This is the metadata of version {$version}. </message>
```
bozana: is metadata singular or plural?

templates/submission/submissionMetadataFormTitleFields.tpl
```diff
@@ -8,6 +8,26 @@
...
+	{* There is only one supported submission locale; choose it invisibly *}
+	{foreach from=$supportedSubmissionLocaleNames item=localeName key=locale}
+		<input type="hidden" name="locale" value="{$locale|escape}" />
+	{/foreach}
+{else}
+	{* There are several submission locales available; allow choice *}
+	{fbvFormSection title="submission.submit.submissionLocale" for="locale"}
+		{fbvElement type="select" label="submission.submit.submissionLocaleDescription" id="locale" from=$supportedSubmissionLocaleNames selected=$locale translate=false readonly=$readOnly size=$fbvStyles.size.MEDIUM}
+	{/fbvFormSection}
+{/if}{* count($supportedSubmissionLocaleNames) == 1 *}
```
bozana: why is this part here, when it is already in core:submission/submissionLocale.tpl and included in controllers/modals/submissionMetadata/form/issueEntrySubmissionReviewForm.tpl just before this template?

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
+{**
+ * lib/pkp/templates/workflow/version.tpl
```
bozana: the path is relative to the repository root i.e. without lib/pkp/

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Copyright (c) 2014-2016 Simon Fraser University Library
+ * Copyright (c) 2003-2016 John Willinsky
```
bozana: The copyright year and name should be checked everywhere -- I didn't mentioned it everywhere...

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
...
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * Version view for workflow stage production
+ *}
+{if $isPublished}
```
bozana: ah, here is the reverse variable name :-)

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
...
+	<div class="pkp_notification">
+		<div class="notifyInfo">
+			This version has not been published yet.
```
bozana: This should be a locale key i.e. not hard coded...

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
...
+	<div class="pkp_context_sidebar">
+
+		{** edit metadata of this version **}
+		<div id="editMetadataDiv" class="pkp_tab_actions">
+			<ul class="pkp_workflow_decisions">
+				<li>{include file="linkAction/linkAction.tpl" action=$editMetadataLinkAction}</li>
+			</ul>
+		</div>
+		
+		{** schedule for publication **}
```
bozana: This part does not exist in all application, thus it should not be in the pkp-lib

templates/workflow/version.tpl
```diff
@@ -0,0 +1,54 @@
...
+	</div>
+
+	<div class="pkp_content_panel">
+
+		{** production ready files **}
+		{url|assign:productionReadyFilesGridUrl router=$smarty.const.ROUTE_COMPONENT component="grid.files.productionReady.ProductionReadyFilesGridHandler" op="fetchGrid" submissionId=$submission->getId() stageId=$stageId escape=false}
+		{load_url_in_div id="productionReadyFilesGridDiv-version_"|concat:$submissionRevision url=$productionReadyFilesGridUrl}
+
+		{** galleys **}
+		{url|assign:representationsGridUrl router=$smarty.const.ROUTE_COMPONENT component="grid.articleGalleys.ArticleGalleyGridHandler" op="fetchGrid" submissionId=$submission->getId() submissionRevision=$submissionRevision escape=false}
```
bozana: also ArticleGalleys should not be in the pkp-lib

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,204 @@
...
+		$submissionDao = Application::getSubmissionDAO();
+		$submissionRevision = $submissionDao->getLatestRevisionId($submissionId);
+		// get data of the old version
+		$submission = $submissionDao->getById($submissionId, null, false, $submissionRevision);
+		// authors
+		$authorDao = DAORegistry::getDAO('AuthorDAO');
+		$authors = $authorDao->getBySubmissionId($submissionId, true, false, $submissionRevision);
+		// galleys
+		import('classes.article.ArticleGalley');
+		$articleGalleyDao = DAORegistry::getDAO('ArticleGalleyDAO');
```
bozana: article galley should not be in pkp-lib. you could take a look at all the changes here and move everything that does not belong to OMP as well to OJS, but you can assume that the versioning will come for OMP soon too, thus leave the common things...

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,110 @@ function newDataObject() {
...
+	 * Get all revisions for a submission
+	 * @param $submissionId int
+	 * @param $contextId int optional
+	 * $param $order string optional default: ASC
+	 * @return array
+	 */
+	function getSubmissionRevisionIds($submissionId, $contextId = null, $order = SORT_DIRECTION_ASC) {
+
+		if (!ctype_digit("$submissionId")) {
+			$publishedArticleDao = DAORegistry::getDAO('PublishedArticleDAO');
```
bozana: again ojs specific code

classes/submission/Representation.inc.php
lilients: Yes, Carola and I discussed this topic as well. But if we create a new column for the latest revision - wouldn't the data be dublicate? But of course it would be much faster. So if you prefer this way, I am happy to implement the additional column with an update on the publication of the version.

asmecher: Yes, there's slight duplication of data, in that the current revision is implicitly the revision data with the highest revision number present in the settings table. I started writing my comment above to determine that using some clever joins, but that started to look like a nightmare for the query cardinalities.

Another option worth considering: what about just using subqueries? OTOH we now require MySQL and PostgreSQL versions that both support them. Including this in a join would look like...


    SELECT ... FROM ... WHERE xyz.revision = (SELECT MAX(revision) FROM submission_settings WHERE submission_id=1);

I do think it's worth a try if we can keep the SQL from getting insane.

classes/submission/SubmissionDAO.inc.php
lilients: true. done.

classes/submission/Representation.inc.php
lilients: I solved it like this:

`SELECT MAX(submission_revision) FROM submission_settings ... WHERE submission_id = ?`

classes/submission/Submission.inc.php
lilients: hmmm, seems to be a left over. Maybe a rebase didn't work here 🤔  ... Removed it.

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
lilients: nope. removed it. thanks!

classes/submission/PKPSubmissionMetadataFormImplementation.inc.php
lilients: yes, removed it. Thanks

controllers/grid/users/author/AuthorGridHandler.inc.php
lilients: No, removed it.

controllers/grid/users/author/AuthorGridHandler.inc.php
lilients: no reason. :) added it

locale/en_US/submission.xml
lilients: I think singular.... 🤔

templates/submission/submissionMetadataFormTitleFields.tpl
lilients: another leftover I guess. removed it.

templates/workflow/version.tpl
lilients: should I move the complete template to ojs or only these parts (schedule for publication and articleGalleys)?

templates/workflow/version.tpl
lilients: Should I move the whole tempalte to ojs or leave it in lib-pkp and add the ojs stuff via hook?

classes/db/DAO.inc.php
lilients: I think so ...

templates/workflow/version.tpl
bozana: I think the same template should exist in OJS and include/call the pkp-lib template (containing the common code)...

lilients: The production.tpl is completly in OJS, maybe it would be ok to leave version.tpl there as well? Also I do not really know how to reasonably separate the OJS and pkp-lib parts, because they are mingled. 😕

bozana: You can then move everything to OJS for now, if this solves it...

classes/submission/Submission.inc.php
```diff
@@ -248,6 +248,42 @@ function setHideAuthor($hideAuthor) {
...
+	 * Get the current revision id for a submission
+	 * @param $contextId int
+	 * @return int
+	 */
+	function getCurrentVersionId() {
+		$submissionId = $this->getId();
+		$contextId = $this->getContextId();
+
+		// handle publisher ids
+		if (!ctype_digit("$submissionId")) {
```
bozana: $submissionId = $this->getId(); should always be int, right? thus, this part of the code is not needed?

classes/submission/Submission.inc.php
```diff
@@ -337,14 +374,20 @@ function getAuthorEmails() {
...
-		else
-			return $authorDao->getBySubmissionId($this->getId(), false, true);
+		if ($version == null) {
+			$version = $this->getCurrentVersionId() ? $this->getCurrentVersionId() : 1; // if a new submission is submitted, version is 1
+		}
+
+		if (!$onlyIncludeInBrowse) {
+			return $authorDao->getBySubmissionId($this->getId(), false, false, $version);
+		} else {
+			return $authorDao->getBySubmissionId($this->getId(), false, true, $version);
```
bozana: These if... else... could maybe be reduces to only one call, giving $onlyIncludeInBrowse as parameter to the function getBySubmissionId, right?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -94,30 +95,85 @@ function newDataObject() {
...
+			FROM 	submission_settings ss
+				INNER JOIN submissions s ON ss.submission_id = s.submission_id
+			WHERE 	ss.submission_id = ?
+				'.($contextId ? ' AND s.context_id = ?' : '') .'
+			ORDER BY ss.submission_revision ' . $this->getDirectionMapping($order);
+
+		$result = $this->retrieve($sql, $params);
+		$submissionRevisions = array();
+
+		foreach($result as $submission) {
```
bozana: hmmm... $result is a ADORecordSet object and not array, so this does not work, I believe...
There would probably something like this be needed:
while (!$result->EOF) {
$row = $result->getRowAssoc(false);
$submissionRevisions[] = $row['submission_revision'];
$result->moveNext();
}
$result->Close();
return $submissionRevisions;

Is this function used at all?

classes/submission/SubmissionDAO.inc.php
```diff
@@ -296,7 +353,7 @@ function deleteAllPubIds($contextId, $pubIdType) {
 	 */
 	function updateLocaleFields($submission) {
 		$this->updateDataObjectSettings('submission_settings', $submission, array(
-			'submission_id' => $submission->getId()
+			'submission_id' => $submission->getId(), 'submission_revision' => ($submission->getData('submissionRevision') ? $submission->getData('submissionRevision') : 1)
```
bozana: maybe to use the submission's setter function getSubmissionRevision here?

controllers/grid/users/author/AuthorGridHandler.inc.php
```diff
@@ -358,11 +407,15 @@ function updateAuthor($args, $request) {
...
+		// Identify the submission Id
+		$submissionId = (int) $request->getUserVar('submissionId');
+		// Identify the author to be deleted
 		$authorId = (int) $request->getUserVar('authorId');
+		// Identify the version
+		$version = (int) $request->getUserVar('version');
+
 		$authorDao = DAORegistry::getDAO('AuthorDAO');
-		$authorDao->deleteById($authorId, $submission->getId());
+		$authorDao->deleteById($authorId, $submissionId, $version);
```
bozana: I think it is good to use `$submission = $this->getSubmission();` because of the security -- at least the authorization policies control who has access to that submission. This way it would be maybe possible to delete any author of any submission, buy just providing their IDs as request parameters i.e. manipulating the form?

controllers/tab/publicationEntry/PublicationEntryTabHandler.inc.php
```diff
@@ -106,10 +106,15 @@ function authorize($request, &$args, $roleAssignments) {
 	 * @return JSONMessage JSON object
 	 */
 	function submissionMetadata($args, $request) {
+		$revision = $request->getUserVar('revision') ? (int)$request->getUserVar('revision') : null;
 
-		$publicationEntrySubmissionReviewForm = $this->_getPublicationEntrySubmissionReviewForm();
+		if(isset($args['submissionRevision'])){
+			$revision = $args['submissionRevision'];
+		}
```
bozana: hmmm... Are those two things different: `$request->getUserVar('revision')` and `$args['submissionRevision']`?

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,153 @@
...
+		
+		// Add variables to the template.
+		$templateMgr = TemplateManager::getManager($request);
+		$templateMgr->assign('submission', $submission);
+		$templateMgr->assign('stageId', $stageId);
+		$templateMgr->assign('submissionRevision', $submissionRevision);
+
+		// Get publication date of submission revision 
+		$submissionDao = Application::getSubmissionDAO();
+		$submission = $submissionDao->getById($submission->getId(), null, false, $submissionRevision);
```
lilients: hmm, I thought we would need to get the object again with the submission revision to get the correct date. But it seems to work anyway.... 😣 👍

classes/controllers/tab/workflow/PKPVersioningTabHandler.inc.php
```diff
@@ -0,0 +1,153 @@
...
+
+		// Create edit metadata link action.
+		$dispatcher = $request->getDispatcher();
+		import('lib.pkp.classes.linkAction.request.AjaxModal');
+		$editMetadataLinkAction = new LinkAction(
+			'editMetadata',
+			new AjaxModal(
+				$dispatcher->url(
+					$request, ROUTE_COMPONENT, null,
+					'modals.submissionMetadata.IssueEntryHandler',
```
lilients: true, moved it to VersioningTabHandler.inc.php

classes/security/authorization/internal/VersioningRequiredPolicy.inc.php
```diff
@@ -0,0 +1,58 @@
...
+
+		$submission = $this->getAuthorizedContextObject(ASSOC_TYPE_SUBMISSION);
+
+		// Validate the submission revision id.
+		$submissionDao = Application::getSubmissionDAO();
+		$submissionRevision = $submissionDao->getById($submission->getId(), null, false, $submissionRevisionId);
+		if (!is_a($submissionRevision, 'Submission')) return AUTHORIZATION_DENY;
+
+		// Ensure that the submissionRevision exists.
+		if (!in_array($submissionRevisionId, $submissionDao->getSubmissionRevisionIds($submission->getId()))) AUTHORIZATION_DENY;
```
lilients: nope, removed it.

classes/submission/PKPAuthorDAO.inc.php
```diff
@@ -237,10 +249,11 @@ function insertObject($author) {
 						(int) $author->getPrimaryContact(),
 						(float) $author->getSequence(),
 						(int) $author->getIncludeInBrowse() ? 1 : 0,
+						(int) $author->getVersion(),
 				)
 		);
 
-		$author->setId($this->getInsertId());
+		if ($saveAsNewVersion == false) $author->setId($this->getInsertId());
```
lilients: good eye! :) thats a leftover of code I deleted. removed it

classes/submission/Submission.inc.php
lilients: it is needed if there is a custom id

classes/submission/Submission.inc.php
lilients: should I change that although it is like this in the master branch? https://github.com/pkp/pkp-lib/blob/master/classes/submission/Submission.inc.php#L344-L347 
I agree it would be easier to read, but I am not sure how much I am allowed to change here (because you do not really need that for versioning...)

classes/submission/Submission.inc.php
bozana: But getId() always returns the actual integer ID, so where does this custom id comes from here?

classes/submission/Submission.inc.php
bozana: I mean these lines here could look like this:
`return $authorDao->getBySubmissionId($this->getId(), false, $onlyIncludeInBrowse, $version);`

classes/submission/Submission.inc.php
lilients: hm, true the function says it should return an int, but there is no check 
https://github.com/pkp/pkp-lib/blob/master/classes/core/DataObject.inc.php#L173-L175
and I tested it, if I remove the if clause articles with custom ids are not displayed... so I would guess the function also returns the custom id if there is one

classes/submission/Submission.inc.php
lilients: yes I got that. ;) I was just wondering if I should "cosmetic" changes ... not sure if that will create problems with rebasing 🤔

classes/submission/SubmissionDAO.inc.php
lilients: yes the function is used here https://github.com/lilients/pkp-lib/blob/versioning/controllers/tab/workflow/PKPWorkflowTabHandler.inc.php#L152 and works fine 😉

classes/submission/Submission.inc.php
bozana: hmmm... maybe to debug? Because everywhere in the code it is so, for every getId()...

classes/submission/SubmissionDAO.inc.php
bozana: S. https://github.com/pkp/pkp-lib/blob/master/lib/adodb/adodb.inc.php#L3499. I suppose that your DB supports arrays but maybe some other does not, so maybe to stay with the usual methods/calls?

controllers/tab/publicationEntry/PublicationEntryTabHandler.inc.php
lilients: hmm, seems to be an old code snippet. Removed it. 👍

classes/submission/Submission.inc.php
lilients: hm, I don't know ho to debug this. The function getId() seems to be special for submissions, because of the custom ids ... any suggestions? 👼

bozana: I cannot believe that getId() is special for submissions -- we are using it everywhere :-)
You could print $submissionId in a file before this line, for example, and then do something that would call this function, then take a look at the file. For example:
$file = 'debug.txt';
$current = file_get_contents($file);
$current .= print_r($submissionId, true);
file_put_contents($file, $current);
Would print the submission id in the file debug.txt...

lilients: ahh, true. my bad. ok, I will remove the unnecessary code here. thanks!

Issue #1815, opened 2016-09-12
Move metadata (DC/GS) meta tags into HTML head


asmecher:

asmecher:
See http://forum.pkp.sfu.ca/t/ojs3-dc-highwire-metadata-not-included-in-the-html-header/19982. @NateWr, an easy solution would be to add something like a Smarty `$additionalHeaders` variable into `lib/pkp/templates/frontend/components/header.tpl`, then allow calling templates to populate it before including the header. Objections to that? Alternately, we could do something similar to what you've done with JS and CSS and allow PHP-side injection.

asmecher:
@bozana, could you have a look at https://github.com/pkp/ojs/pull/1024?

Vitaliy-1:
Many thanks for the update! All seems to work fine. All Metadata is displaying in the head now.

asmecher:
https://github.com/pkp/ojs/pull/1024 need sporting over to OMP master.

bozana:
Isn't this issue finished for OJS? -- because of the milestone...

asmecher:
Yes, it's finished for OJS (the milestone is currently OMP 3.1).

bozana:
Ah, sorry, I saw wrong... :-\

asmecher:
Pull requests:

- https://github.com/pkp/pkp-lib/pull/2318
- https://github.com/pkp/omp/pull/395

@bozana, could you have a look at this? The pkp-lib change adds the ability of hook registrants to specify priority in calling hooks -- this was necessary because we need to ensure that e.g. the DC and GS metadata tag plugins have a chance to inject the metadata before the PDF viewer plugin displays the file.

asmecher:
(This is a pretty big change to merge right before the release, so one option would be to release before merging and then merge it after. Still considering.)

bozana:
@asmecher, I went through everything roughly. If we would have some more time I could also test it...

asmecher:
Thanks, @bozana. On balance I think this is too risky to merge in so close to the release, and since we haven't been able to get confirmation from GS about whether OMP content is getting indexed anyway, there's little upside at the moment. I'll go through the rest of the review comments and fix, but leave unmerged until after OMP 3.1 is released.

asmecher:
Fixed and merged into OMP `master` -- thanks, @bozana. @NateWr, [this comment](https://github.com/pkp/omp/pull/395#discussion_r103530196) could still use a review -- not sure if it got lost in the noise.

PR #2318, opened 2017-02-28
pkp/pkp-lib#1815 Permit hooks to register with call priorities

asmecher:


classes/file/FileManager.inc.php
```diff
@@ -234,7 +234,6 @@ function readFileFromPath($filePath, $output = false) {
 	function downloadFile($filePath, $mediaType = null, $inline = false, $fileName = null) {
 		$result = null;
 		if (HookRegistry::call('FileManager::downloadFile', array(&$filePath, &$mediaType, &$inline, &$result, &$fileName))) return $result;
-		$postDownloadHookList = array('FileManager::downloadFileFinished', 'UsageEventPlugin::getUsageEvent');
```
bozana: I see the FileManager::downloadFileFinished is still there, at the end of the function, but what about UsageEventPlugin::getUsageEvent ?

asmecher: I removed the code that clears out the registry before downloading a file; the post download hook list was used to selectively restore hooks after downloading had completed. I don't think it's a good optimization so I removed it.

classes/plugins/HookRegistry.inc.php
```diff
@@ -61,13 +65,11 @@ static function clear($hookName) {
...
+	 * @param $hookSequence int Optional hook sequence specifier HOOK_SEQUENCE_...
 	 */
-	static function register($hookName, $callback) {
+	static function register($hookName, $callback, $hookSequence = HOOK_SEQUENCE_NORMAL) {
 		$hooks =& HookRegistry::getHooks();
-		if (!isset($hooks[$hookName])) {
-			$hooks[$hookName] = array();
-		}
-		$hooks[$hookName][] =& $callback;
+		$hooks[$hookName][$hookSequence][] =& $callback;
```
bozana: Does indexes $hookName and $hookSequence exist here for sure?

asmecher: As indexes in the array? No, but PHP doesn't complain in this case.

bozana: Ah yes, true, sorry... :-\

PR #2324, opened 2017-03-01
Check if reviewassignment is accepted before returning overdue notice

ajnyga:
see https://github.com/pkp/pkp-lib/issues/2323


ajnyga:
https://github.com/pkp/pkp-lib/issues/2323#issuecomment-283724254

PR #2332, opened 2017-03-02
pkp/pkp-lib#2114 multilingualism for file names and submission metadata

bozana:
s. https://github.com/pkp/pkp-lib/issues/2114

### DIFF
--- a/classes/submission/SubmissionFile.inc.php
+++ b/classes/submission/SubmissionFile.inc.php
@@ -41,6 +41,34 @@ class SubmissionFile extends PKPFile {
 		parent::__construct();
 	}
 
+	/**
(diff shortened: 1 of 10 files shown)


classes/submission/SubmissionFile.inc.php
```diff
@@ -66,6 +94,22 @@ function setFileId($fileId) {
 	}
 
 	/**
+	 * Get the locale of the submission.
+	 * @return string
```
asmecher: I would suggest noting in the getter/setters that this is not properly a property of the submission file (e.g. so devs know it won't be persisted to the DB with the usual update function).

PR #2351, opened 2017-03-10
pkp/pkp-lib#2306 Searching by SubmissionID enabled

defstat:


classes/submission/SubmissionDAO.inc.php
```diff
@@ -471,20 +471,22 @@ function getByUserId($userId, $contextId = null) {
 	 *  whose section will be included in the results (excluding others).
 	 * @param $includeDeclined boolean optional include submissions which have STATUS_DECLINED
 	 * @param $includePublished boolean optional include submissions which are published
+	 * @param $id int|null optional Filter by id.
```
asmecher: For consistency, use `$submissionId` rather than `$id`.

classes/submission/SubmissionDAO.inc.php
```diff
@@ -527,18 +530,23 @@ function getBySubEditorId($contextId, $subEditorId = null, $includeDeclined = tr
...
-	function getUnpublishedByUserId($userId, $contextId = null, $title = null, $stageId = null, $sectionId = null, $rangeInfo = null) {
+	function getUnpublishedByUserId($userId, $contextId = null, $id = null, $title = null, $stageId = null, $sectionId = null, $rangeInfo = null) {
 		$params = array_merge(
 			$this->getFetchParameters(),
 			array((int) ROLE_ID_AUTHOR, (int) $userId)
 		);
-		if ($title) $params[] = '%' . $title . '%';
+		if ($id) $params[] = (int) $id;
+		if ($title) {
+			$params[] = 'title';
```
asmecher: Good catch.

controllers/grid/submissions/assignedSubmissions/AssignedSubmissionsListGridHandler.inc.php
```diff
@@ -63,17 +63,19 @@ protected function loadData($request, $filter) {
 		$authorDao = DAORegistry::getDAO('AuthorDAO');
 
 		list($search, $column, $stageId, $sectionId) = $this->getFilterValues($filter);
-		$title = $author = null;
+		$title = $author = $id = null;
 		if ($column == 'title') {
```
asmecher: Could you swap these out for a `switch`/`case` block? I find these a lot more readable.

controllers/grid/submissions/exportableSubmissions/ExportableSubmissionsListGridHandler.inc.php
```diff
@@ -42,17 +42,20 @@ function loadData($request, $filter) {
 		$context = $request->getContext();
 
 		list($search, $column, $stageId, $sectionId) = $this->getFilterValues($filter);
-		$title = $author = null;
+		$title = $author = $id = null;
 		if ($column == 'title') {
```
asmecher: (`switch`/`case` here too)

controllers/grid/submissions/unassignedSubmissions/UnassignedSubmissionsListGridHandler.inc.php
```diff
@@ -69,11 +69,13 @@ protected function loadData($request, $filter) {
 		$context = $request->getContext();
 
 		list($search, $column, $stageId, $sectionId) = $this->getFilterValues($filter);
-		$title = $author = null;
+		$title = $author = $id = null;
 		if ($column == 'title') {
```
asmecher: (and here and probably elsewhere)

classes/submission/SubmissionDAO.inc.php
defstat: @asmecher the previous implementation broke the sql, and therefore the search at the submissions->My Queue -> My Authored grid. 
I haven't found any Issue for that, but if there is any, maybe we could close it.

classes/submission/SubmissionDAO.inc.php
defstat: Done

controllers/grid/submissions/assignedSubmissions/AssignedSubmissionsListGridHandler.inc.php
defstat: Done

controllers/grid/submissions/exportableSubmissions/ExportableSubmissionsListGridHandler.inc.php
defstat: Done

controllers/grid/submissions/unassignedSubmissions/UnassignedSubmissionsListGridHandler.inc.php
defstat: Done

classes/submission/SubmissionDAO.inc.php
```diff
@@ -471,20 +471,33 @@ function getByUserId($userId, $contextId = null) {
...
 	 */
-	function getBySubEditorId($contextId, $subEditorId = null, $includeDeclined = true, $includePublished = true, $title = null, $author = null, $stageId = null, $sectionId = null, $rangeInfo = null) {
+	function getBySubEditorId($contextId, $subEditorId = null, $includeDeclined = true, $includePublished = true, $submissionId = null, $title = null, $author = null, $stageId = null, $sectionId = null, $rangeInfo = null) {
 		$params = $this->getFetchParameters();
 		if ($subEditorId) $params[] = (int) $subEditorId;
 		$params[] = (int) $contextId;
 		$params[] = (int) ROLE_ID_MANAGER;
 		$params[] = (int) ROLE_ID_SUB_EDITOR;
 
+		if ($submissionId) {
```
asmecher: You could tighten this up a little using...

    if ($submissionId) $submissionId = filter_var($submissionId, FILTER_VALIDATE_INT);

...then when building the SQL...

    . ($submissionId!==null?' AND s.submission_id = ?':'')

Just trying to keep the code concise, since it's used in a few places!

defstat: We still have to put the `$params[] = (int) $submissionId;` line of code, right?
The `filter_var($submissionId, FILTER_VALIDATE_INT)` returns false if it does not pass the check. 
We could do something like
 `($submissionId!==null && !$submissionId?' AND s.submission_id = ?':'')` 
but that way, if the filter_var check fails, all the rows will be returned by the query (thus the 
`$params[] = -1;` line).
Am I missing something there?

asmecher: Sorry, yes...

    if ($submissionId) $submissionId = filter_var($submissionId, FILTER_VALIDATE_INT);

...should maybe be...

    if ($submissionId) $params[] = filter_var($submissionId, FILTER_VALIDATE_INT);

This should be OK if an invalid submission ID is entered, because `filter_var` returns `false` in that case, and the result of the `int` cast will be `0`. The check in the SQL side to add the condition will pass (`null !== false`) and the `0` condition will match no submissions, as appropriate.

defstat: Right! I should have seen that. Too much C# may do some harm after all! :)

fixed.

asmecher:
Looks good, thanks!

Issue #1911, opened 2016-10-21
enable longer locale names


mtub:
To be able to deal with locale names that differ from the standard xx_YY pattern we've used exclusively so far, the PKP applications should be able to deal with longer locale names. Review of OJS (higher priority) and OMP code is necessary to make sure introducing longer locale names will not break the applications.

asmecher:
There are still areas of the code that try to parse bits out of xx_YY (predominantly `PKPLocale` but also elsewhere). Those will need review at the same time.

asmecher:
@kaschioudi, do you think this is something you could look at?

kaschioudi:
Sure. Any available documentation for locale somewhere?

asmecher:
Not really, just that we assume throughout the code that it's `xx_YY` format. We need to relax that to support more diversity, and make sure that places assuming `xx_YY` will still behave nicely. @mtub, I think you were throwing around some proposals via email; would you mind putting the notes here for Kassim?

mtub:
Hi @kaschioudi, our first use case would be going for these two locale names:

sr_RS@latin
sr_RS@cyrillic 

(following the gettext example, see https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html)

kaschioudi:
I looked at the source code and the only place I found which actually tries to parse xx_YY locale is here (https://github.com/pkp/pkp-lib/blob/master/classes/i18n/PKPLocale.inc.php#L333). 

Considering that syntax for locales is `lang[_territory][.encoding][@variant]`, we currently have the following mapping (cf config.inc.php):
lang[_territory] => i18n.locale
encoding => i18n.client_charset

I suggest introducing a new configuration field under i18n category for the locale variant. Then update PKPLocale::initialize (https://github.com/pkp/pkp-lib/blob/master/classes/i18n/PKPLocale.inc.php#L123) to work with locale variant if defined.

Finally we could modify AppLocale::getLocale function to optionally return the locale variant if defined `static function getLocale($withVariant = false)`. 

@asmecher  what do you think?

asmecher:
[...]

PR #2356, opened 2017-03-13
add support for longer locale names #1911 

kaschioudi:
@asmecher


js/controllers/SiteHandler.js
```diff
@@ -184,7 +184,7 @@
 	$.pkp.controllers.SiteHandler.prototype.triggerTinyMCEInitialized =
 			function(tinyMCEObject) {
 
-		var $inputElement = $('#' + tinyMCEObject.id);
+		var $inputElement = $('#' + $.pkp.classes.Helper.escapeJQuerySelector(tinyMCEObject.id));
```
asmecher: I think `tinyMCEObject.id` is a proper ID, not a selector.

js/classes/Helper.js
```diff
@@ -350,6 +350,17 @@
 		}
 	};
 
+	/**
+	 * A function that takes care of escaping @ character which could be interpreted
+	 * as CSS notation. This is due to the fact that jQuery uses CSS syntax for selecting elements.
+	 * These characters must be escaped by placing two backslashes in front of them.
+	 * @param {string} selector jQuery element selector
```
asmecher: This function should turn an ID into a selector, correct? Wouldn't e.g. `elementId` be a better parameter name than `selector`?

templates/form/keywordInput.tpl
```diff
@@ -46,13 +46,13 @@
 		{rdelim});
 		</script>
 		<span id="{$FBV_id|escape}-localization-popover-container{$uniqId}" class="localization_popover_container pkpTagit">
-			<ul class="localization_popover_container localizable {if $formLocale != $currentLocale} flag flag_{$formLocale|escape}{/if}" id="{$formLocale|escape}-{$FBV_id|escape}{$uniqId}">
+			<ul class="localization_popover_container localizable {if $formLocale != $currentLocale} flag flag_{$formLocale|escape}{/if}" id="{$formLocale|escape|escape:jqselector}-{$FBV_id|escape}{$uniqId}">
```
asmecher: Is it necessary to escape IDs? I don't think e.g. `@` characters are invalid in IDs, and escaping them here might lead to double-escaping when selecting for them.

js/classes/Helper.js
kaschioudi: yes, I will rename it.

templates/form/keywordInput.tpl
kaschioudi: right. To be honest I can't remember why I escaped this line.. I just removed this extra escape and everything seems to be working well.

asmecher: There are other cases that look the same, where escaping is applied to markup element IDs.

templates/form/textInput.tpl
```diff
@@ -43,7 +43,7 @@
 					{if $FBV_readonly} readonly="readonly"{/if}
 					value="{$FBV_value[$thisFormLocale]|escape}"
 					name="{$FBV_name|escape}[{$thisFormLocale|escape}]"
-					id="{$FBV_id|escape}-{$thisFormLocale|escape}{$uniqId}"
+					id="{$FBV_id|escape}-{$thisFormLocale|escape|escape:jqselector}{$uniqId}"
```
asmecher: This double-escaping is messy -- since I think all applications of the `jqselector` escape mode are the same, I would suggest having that one call handle all escaping requirements. (Here and elsewhere.)

kaschioudi: so just $thisFormLocale|escape:jqselector instead of $thisFormLocale|escape|escape:jqselector ?

asmecher: Yes, though you'll have to check the escaping requirements of your use cases and possibly have the `jqselector` escape pass its resuls through to one of the `escape` types in the parent implementation.

kaschioudi: How would you handle escape parameters? For instance $variable|escape:javascript|escape:jqselector?

asmecher: I think the escaping requirements are the same for all use cases of `jqselector`, so it would just be `{$someId|escape:jqselector}`, and `jqselector` would take care of the rest.

kaschioudi: Okay.

kaschioudi: Here's a case where jqselector is applied to $selector:javascript: 
https://github.com/kaschioudi/pkp-lib/blob/1911-fix/templates/linkAction/buttonGenericLinkAction.tpl#L19

asmecher: Yes, that looks like a different use case. Maybe add two custom types of `escape`s -- `jqselector` for selectors, and `jsid` for IDs?

kaschioudi: alright.

templates/form/keywordInput.tpl
```diff
@@ -52,7 +52,7 @@
 			{if $FBV_label_content}<span>{$FBV_label_content}</span>{/if}
 			<div class="localization_popover">
 				{foreach from=$formLocales key=thisFormLocale item=thisFormLocaleName}{if $formLocale != $thisFormLocale}
-					<ul class="multilingual_extra flag flag_{$thisFormLocale|escape}" id="{$thisFormLocale|escape}-{$FBV_id|escape}{$uniqId}">
+					<ul class="multilingual_extra flag flag_{$thisFormLocale|escape}" id="{$thisFormLocale|escape:jqselector}-{$FBV_id|escape}{$uniqId}">
```
asmecher: I still think the escaping in `id="..."` attributes is not required, is it?

asmecher: (Also elsewhere in the pull request.)

classes/template/PKPTemplateManager.inc.php
```diff
@@ -1272,6 +1273,34 @@ function smartyAssign($value, $varName, $passThru = false) {
 	}
 
 	/**
+	 * Override the built-in smarty escape modifier to
+	 * add the jqselector escaping method.
+	 */
+	function smartyEscape($string, $esc_type = 'html', $char_set = 'ISO-8859-1') {
+		$pattern = "/(:|\.|\[|\]|,|=|@)/";
+		$replacement = "\\\\\\\\$1";
```
asmecher: I wonder whether this many backslashes are required because you're already escaping in the ID markup. That seems like too many.

asmecher:
Still a few things to work out, @kaschioudi. I haven't tested this yet -- have you tried with the two Serbian locales to see whether this works with that practical example?

classes/template/PKPTemplateManager.inc.php
kaschioudi: backslash needs to be doubled with double quotes.  moreover, with preg_replace, to use backslash in replacement, it must be doubled. So we need \\\\ for a single backslash.

kaschioudi:
@asmecher : I have tried with Serbian locales from https://github.com/vedranvucic/ojs/tree/master-serbian-updates & https://github.com/vedranvucic/pkp-lib/tree/master-serbian-updates

Issue #2274, opened 2017-02-09
[OJS] Automated Submission Review Reminders are sent with the url and signature of another journal


ajnyga:
Hi,

Yesterday two submission review reminders were sent with the url and the signature of another journal.

The situation was this:

In journal A there were two overdue review assignments. Automatic reminders were turned off.
In juornal B there are no review assignments yet, but the automatic reminders are turned on.

Yesterday, the automatic reminders in journal B were triggered and the system wrote two reminder letters concerning the overdue assignments in journal A and sent them to the reviewers of journal A.

The letters included the initials of journal B in the subject field, the URLs were leading to journal B but had the submission id from journal A. Also, the signature and the sender was the editor of Journal B. All the rest of the information, receiver, submission title, abstract etc. came from journal A.

There were two similar cases and both review assignments were made already in OJS2.
Now using 3.0.2.

ajnyga:
I guess that somehow the context used here https://github.com/pkp/pkp-lib/blob/ojs-stable-3_0_2/classes/task/ReviewReminder.inc.php#L55 is not the context of the review assignment but the current context that has triggered the Acron plugin?

Although the context should be set correctly here: https://github.com/pkp/pkp-lib/blob/ojs-stable-3_0_2/classes/task/ReviewReminder.inc.php#L156

I checked the database and there are no mistakes there. The submission context is what it should be and submission id in the review assignments is correct as well.

ajnyga:
I checked from a database dump that `numDaysBeforeSubmitReminder` for journal A was set to `0`.
For journal B it is set to `1`

ajnyga:
I can not figure out what has happened here. The only explanation I can think of is that for some reason this https://github.com/pkp/pkp-lib/blob/ojs-stable-3_0_2/classes/task/ReviewReminder.inc.php#L156 has returned the wrong context for some reason. But that just does not seem possible.

asmecher:
@ajnyga, are you using the acron plugin? If so, that might be worth investigating. That plugin piggy-backs onto normal user requests and performs scheduled tasks at the end of the request; since it's coming in on a user request, probably within a journal context, it's possible that the user's request is somehow getting used by the scheduled task e.g. to determine what journal to work with.

ajnyga:
Hi @asmecher 

[...]

PR #2357, opened 2017-03-13
Fix context in reminders sent using Acron

ajnyga:
There are two changes / issues here.

1. If you send automated reminders with Acron the current code in https://github.com/pkp/pkp-lib/blob/master/classes/mail/MailTemplate.inc.php#L160 may use the wrong $context in multijournal installations and show the wrong `contextUrl` value in the email

2. Another issue occurs in some reminders and changes several context values in the email. This happens also in multijournal installations and dependes on several circumtances. Details explained here https://github.com/pkp/pkp-lib/issues/2274#issuecomment-283079770 .

### DIFF
--- a/classes/mail/MailTemplate.inc.php
+++ b/classes/mail/MailTemplate.inc.php
@@ -152,12 +152,11 @@ class MailTemplate extends Mail {
 
 		if ($this->context) {
 			// Add context-specific variables
-			$router = $request->getRouter();
 			$dispatcher = $application->getDispatcher();
 			$params = array_merge(array(
 				'principalContactSignature' => $this->context->getSetting('contactName'),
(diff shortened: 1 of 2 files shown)


asmecher:
@bozana, would you mind taking a look at this? @ajnyga, thanks again for all your work!

ajnyga:
Thanks! @bozana I could not find a reason why the current code in https://github.com/pkp/pkp-lib/blob/master/classes/mail/MailTemplate.inc.php#L160 is using `$router->getRequestedContextPath($request)` but maybe there is a reason there that applies to some other mail template?

bozana:
Thanks a lot @ajnyga, this all looks well and logical! :-)))

PR #2361, opened 2017-03-16
port getPageArray and a PHP7 change

bozana:
This is a cherry-pick of some changes from ojs-dev-2_4 to ojs-stable-2_4_8, in order for them to be in the next release, e.g. to solve the problem of the roman numerals in Crossref


bozana:
@ctgraham, these are the submodule chages.
@asmecher, the last commit is the one you have mentioned that should be migrated too, correct?

PR #2363, opened 2017-03-16
Pkp/pkp lib#2064 standardize emailed url handling conflict1

defstat:


js/classes/TinyMCEHelper.js
```diff
@@ -10,8 +10,7 @@
  *
  * @brief TinyMCE helper methods
  */
-(function($) {
-
+(function ($) {
```
asmecher: Hmm, I'm surprised the JS linter didn't complain about this -- it's really, really picky about whitespacing.

js/classes/TinyMCEHelper.js
```diff
@@ -48,6 +48,26 @@
...
+
+				// If we found the data attribute, decode and return it.
+				if (variablesTypeEncoded !== undefined) {
+					return $.parseJSON(decodeURIComponent(
+							/** @type {string} */(variablesTypeEncoded)));
+				}
+
+				// If we could not find the data attribute, return an empty list.
+				return [];
+			};
```
asmecher: Yup, I'm pretty certain the JS linter would complain about this. I wonder whether it's not rebuilding all the JS like it should. I've added a change (https://github.com/pkp/ojs/commit/3f7a5121ca5a3f52d64fc68014b18d9e75862254) that disables date-based caching, which might be preventing everything from building as it should during testing. Can you rebase your branch off the master branch and force an update? That should cause it to run with this change in place. I suspect you'll find a bunch of JS complaints.

PR #2388, opened 2017-03-26
tr_TR locale: update

mtub:
Translation updates received from @snmzclk via https://github.com/snmzclk/OCS-Turkish-Language-Translation 

Haven't compared this to the master branch of pkp-lib. Since OCS is so far running based on the ocs-dev-2_3 branch, I think that's okay for now. During building a new and better OCS, we should look at differences between the master and the ocs-dev-2_3 branch of pkp-lib and talk to translators if we find issues.


mtub:
done :+1:

Issue #1307, opened 2016-03-24
Notification emails are generally Journal specific, but sent from the Site level.


ctgraham:
See http://forum.pkp.sfu.ca/t/notification-emails-reply-to-principal-contact-of-site-administrator-but-not-principal-contact-of-the-journal/9455

In OJS 2.4, It is confusing for readers (and for journal managers) to have the reply-to field set to the Site Admin, when the messaging is typically journal related and initiated by the journal manager.

Using the `$notifcation`'s `$context` to get the Journal Manager should be trivial.

asmecher:
@jnugent, this one too.

jnugent:
This looks great!  @asmecher please merge. Unless I should :)

asmecher:
All yours, @jnugent!

jnugent:
merged with great justice.

asmecher:
http://old.he-man.org/cartoon/cmotu-pop/powerpose1.jpg

PR #2402, opened 2017-03-30
Always use context specific reply-to address in notifications, if available

ajnyga:
Details here: http://forum.pkp.sfu.ca/t/ojs3-notifications-sender-and-reply-to-address/28703


asmecher:
@ctgraham, as someone who's tinkered with email policies more than most, do you have an opinion on this change?

ctgraham:
I have not tested the specific code, but I strongly affirm the general concept.  See also:
https://github.com/pkp/pkp-lib/issues/1307

asmecher:
Thanks for confirming, @ctgraham. And after a few more days pondering this I can't think of any reason not to go ahead. Merged -- thanks as always, @ajnyga!

Issue #2420, opened 2017-04-05
[OJS 3.x] No check on PHPMailer::send() status


ctgraham:
We catch exceptions from PHPMailer, but an error in sending doesn't necessarily generate an exception.
https://github.com/pkp/pkp-lib/blob/91da80b797e2f7dd98014134251c2884de6f30c5/classes/mail/Mail.inc.php#L516-L522

So, if `PHPMailer::send()` simply returns false, the email fails and there is no feedback in the UI or in the error log.

I replicated this by configuring a bad SMTP connection in config.inc.php and attempting to send a email to a user in OJS 3.0.2.

asmecher:
Thanks, @ctgraham!

PR #2421, opened 2017-04-05
pkp/pkp-lib#2420: Log email sending failures, even if PHPMailer doesn't raise an exception.

ctgraham:
Resolves pkp/pkp-lib#2420.

Why does the `catch (phpmailerException $e)` log `$mailer->ErrorInfo` and not `$e->errorMessage`?
https://github.com/pkp/pkp-lib/blob/91da80b797e2f7dd98014134251c2884de6f30c5/classes/mail/Mail.inc.php#L519

### DIFF
--- a/classes/mail/Mail.inc.php
+++ b/classes/mail/Mail.inc.php
@@ -473,6 +473,11 @@ class Mail extends DataObject {
 			$mailer->Host = Config::getVar('email', 'smtp_server');
 			$mailer->Username = Config::getVar('email', 'smtp_username');
 			$mailer->Password = Config::getVar('email', 'smtp_password');
+			if (Config::getVar('debug', 'show_stacktrace')) {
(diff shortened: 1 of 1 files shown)


classes/mail/Mail.inc.php
```diff
@@ -473,6 +473,11 @@ function send() {
 			$mailer->Host = Config::getVar('email', 'smtp_server');
 			$mailer->Username = Config::getVar('email', 'smtp_username');
 			$mailer->Password = Config::getVar('email', 'smtp_password');
+			if (Config::getVar('debug', 'show_stacktrace')) {
+				// debug level 3 represents client and server interaction, plus initial connection debugging
```
asmecher: Weird that they seem to have a mix of magic numbers and magic strings... I wish they'd use `define` like certain other projects ;)

PR #2439, opened 2017-04-13
Replace control characters with printable sequences

shoorick:
Raw control characters (ASCII 0x00-0x31) can cause issues with some software (subversion, terminals etc)


asmecher:
@shoorick, can you describe the high-level behavior this resolves? (FYI, this is a modification to a third-party library, and we try not to do that. That said, it's not used frequently in OJS so maybe there's another resolution.)

PR #2451, opened 2017-04-18
pkp/pkp-lib#2449: correct method documentation for GridRow and GridHandler

ctgraham:
Mostly making parameter ordering and typing consistent.


bozana:
Thanks!!!

Issue #2337, opened 2017-03-04
[OJS] Update Serbian locale


mtub:
A decision has been made regarding the locale names, see #1858 and #1911.

Opening this issue to track the renaming and updating of two Serbian locales.

mtub:
At the end of this process, the sr_SR folders (which are using the wrong country code and do not include the script variant) will have been removed/replaced.

mtub:
@asmecher , these pull requests are all of the translation work done so far. do you mind to have a look?

asmecher:
Many thanks! I've merged the fix for #1911 so these are finally getting a look. There's some work still required on the OJS PR before the XML will validate.

asmecher:
@mtub, all of these appear to be merged -- is this issue ready to be closed?

mtub:
Yes, closing…

PR #2459, opened 2017-04-20
#2337 remove outdated Serbian locale after merging the new Serbian lo…

mtub:
…cale variants


bozana:
Thanks a lot @mtub!!! :-)))

PR #2478, opened 2017-04-28
pkp/pkp-lib#2481 Fix step1 buttons for declined reviews

ajnyga:
See http://forum.pkp.sfu.ca/t/reviewer-has-access-to-manuscript-after-declining-review-request/30660


bozana:
This was merged with this PR: https://github.com/pkp/pkp-lib/pull/2482. So I will close it... THANKS!!!

Issue #2142, opened 2017-01-04
Suggestions for Native XML Import


pkel:
I am writing a script to automatically import articles/galleys via Native XML after copyediting has been done externally. The following suggestions for improvement came up:

  * Make `filesize` attribute on `<revision>` optional and derive the information from content of `<embed>`.
  * Make `user_group_ref` and `uploader` attributes optional and default to the importing user.
  * In OJS2, I had the possibility to set copyright holder and year via xml. I am missing this feature in OJS3.

My version is OJS 3.0.1

asmecher:
Thanks for the feedback, @pkel. Good suggestions.

ctgraham:
If the `user_group_ref` is optional, what should be the criteria for selecting one of potentially many user groups from a user?

ctgraham:
Specifically for @asmecher or @bozana, if a User and Group is specified in the import, but the user specified does not currently have that group, is that an error, or is that acceptable?

asmecher:
@ctgraham, I don't have a strong opinion on this either way -- potentially file under "enough rope to hang yourself".

ctgraham:
If the user group is optional, I'm preferring any of the user's current groups which has access to the specified stage.

PR #2484, opened 2017-05-02
pkp/pkp-lib#2142: Allow Native XML import to autodetect uploader and filesize for submission file revisions

ctgraham:
Resolves pkp/pkp-lib#2142 for the suggestions of optional elements being derived from known information.

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
@@ -102,8 +102,8 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 	/**
 	 * Handle a revision element
 	 * @param $node DOMElement
-	 * @param $fileId int File id
 	 * @param $stageId int SUBMISSION_FILE_...
+	 * @param $fileId int File id
 	 */
 	function handleRevisionElement($node, $stageId, $fileId) {
 		static $genresByContextId = array();
@@ -163,24 +163,13 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 
 		$uploaderUsername = $node->getAttribute('uploader');
 		$uploaderUserGroup = $node->getAttribute('user_group_ref');
-
-		// Determine the user group based on the user_group_ref element.
-		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
-		$userGroups = $userGroupDao->getByContextId($context->getId());
-		while ($userGroup = $userGroups->next()) {
-			if (in_array($uploaderUserGroup, $userGroup->getName(null))) {
-				$submissionFile->setUserGroupId($userGroup->getId());
-				break;
-			}
+		if (!$uploaderUsername) {
+			$user = $deployment->getUser();
+		} else {
+			// Determine the user based on the username
+			$userDao = DAORegistry::getDAO('UserDAO');
+			$user = $userDao->getByUsername($uploaderUsername);
 		}
-		if (!$submissionFile->getUserGroupId()) {
-			$deployment->addError(ASSOC_TYPE_SUBMISSION, $submission->getId(), __('plugins.importexport.common.error.unknownUserGroup', array('param' => $uploaderUserGroup)));
-			$errorOccured = true;
-		}
-
-		// Do the same for the user.
-		$userDao = DAORegistry::getDAO('UserDAO');
-		$user = $userDao->getByUsername($uploaderUsername);
 		if ($user) {
 			$submissionFile->setUploaderUserId($user->getId());
 		} else {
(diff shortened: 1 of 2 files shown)


plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -163,9 +163,42 @@ function handleRevisionElement($node, $stageId, $fileId) {
...
+			$userUserGroups = $userGroupDao->getByUserId($user->getId(), $context->getId());
+			while ($userGroup = $userUserGroups->next()) {
+				$defaultedUserGroup = $userGroup->getName(null);
+				// If the user's user group has access to this stage, select it
+				if (in_array($defaultedUserGroup, $stageUserGroupLookup)) {
+					break;
+				}
+				// if no current user group has access to this stage, fall through with the any of the user's current user groups
+			}
+			$uploaderUserGroup = $defaultedUserGroup;
```
bozana: $uploaderUserGroup is an array here, right? So would `in_array` below work? Maybe to also here assign the groupId to the file immediately and not search through groups again?

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -163,9 +163,42 @@ function handleRevisionElement($node, $stageId, $fileId) {
 
 		$uploaderUsername = $node->getAttribute('uploader');
 		$uploaderUserGroup = $node->getAttribute('user_group_ref');
+		if (!$uploaderUsername) {
+			$uploaderUsername = $deployment->getUser()->getUsername();
+		}
+
+		// Determine the user based on the username
+		$userDao = DAORegistry::getDAO('UserDAO');
+		$user = $userDao->getByUsername($uploaderUsername);
```
ctgraham: Agreed.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
ctgraham: Well, that's weird.  `UserGroup::getName()` declares itself to [return a string](https://github.com/ulsdevteam/pkp-lib/blob/616832c78d1f7df9fea713f2626aae21278c5895/classes/security/UserGroup.inc.php#L139), so I thought `$uploaderUserGroup` would ultimately be a string, but the original code sure does treat the results of `UserGroup::getName()` as an array on line 204.  To the debugMobile, Batman!

ctgraham: @asmecher, @bozana, `UserGroup::getName()` does not acknowledge that it can take a `null` as the `$locale`, nor that it can return an `array()` (or `null`, technically) as a result.
```
	/**
	 * Get user group name
	 * @param $locale string
	 * @return string
	 */
	function getName($locale) {
		return $this->getData('name', $locale);
	}
```
If the use of `in_array($uploaderUserGroup, $userGroup->getName(null))` is legitimate, I will modify the documentation for `UserGroup::getName()`; if the use is not, I will enforce the requirement of `$locale` to `getName()` (and create a new `getNames()` function?).  What say you, oh guardians of the API?

asmecher: That's a legit use -- the code self-docs should permit `$locale` to be null, and thus `@return` to be an array, when calling functions want the full list of data rather than a localized value.

ctgraham: Filed as https://github.com/pkp/pkp-lib/pull/2490.  Will fix this code shortly.

ctgraham:
@bozana, new push addressing your comments.  Thanks for the re-review.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -163,32 +163,64 @@ function handleRevisionElement($node, $stageId, $fileId) {
...
+			$userUserGroups = $userGroupDao->getByUserId($user->getId(), $context->getId());
+			$lastUserGroup = null;
+			while ($userGroup = $userUserGroups->next()) {
+				// If the user's user group has access to this stage, select it.
+				$lastUserGroup = $userGroup->getId();
+				if (in_array($lastUserGroup, $stageUserGroupLookup, true)) {
+					$submissionFile->setUserGroupId($userGroup->getId());
+					break;
+				}
+				// No matching user group has access to this stage, select the last one seen.
```
bozana: Should the while closing "}" be before this line?

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -163,32 +163,64 @@ function handleRevisionElement($node, $stageId, $fileId) {
...
+				if ($lastUserGroup) {
```
bozana: I think this if is not necessary -- there will always be $lastUserGroup here, right?

ctgraham: This should have been outside the while in case the user had no role in the context.

Issue #1791, opened 2016-09-02
Fix COUNTER plugin template references (and test plugin)


asmecher:
See CounterPlugin::logRequest near line 106:

    // TODO: consider the effect of LOCKSS on COUNTER recording
    assert(false); // Template names no longer apply (e.g. interstitial)

ctgraham:
Watching this issue.... there's a ton of changes in 2.4.x which need to be brought forward to 3.x.

bozana:
We don't have this plugin since 2.4.3, I believe -- since the Usage Statistics plugin came. Right? It seems that it has been there already before 2.4.3 and haven't been adapted since then i.e. after 2.4.3?

bozana:
@ctgraham, did you use this plugin?

ctgraham:
If we are all talking about reports/counter in 2.4.x (https://github.com/pkp/ojs/tree/ojs-dev-2_4/plugins/reports/counter), yes, this plugin is critical for us.  We updated the architecture of the plugin last year to better support multiple reports, updated the COUNTER release to 4.1, and added the AR1 report.  This plugin serves as the foundation for the reports available via SUSHI-Lite.

bozana:
@ctgraham, @asmecher: because the 3.0.1 should come in a few days, can we defer this to 3.0.2?

asmecher:
Agreed!

asmecher:
@bozana, gifting you this one!

bozana:
How generous... Deeply appreciate it... :-)

bozana:
@ctgraham, I would then just migrate the current Counter plugin from ojs-dev-2_4/ojs-stable-2_4_8 to master, correct? Or would you recommend any further changes?

bozana:
I will write my comments here, the things I realize during the forwardporting to OJS 3:
-- The old COUNTER reports (from OJS 2.4.x) were containing the remote galleys counts. This will not be the case in the OJS 3.

bozana:
From here: https://github.com/pkp/pkp-lib/issues/1314#issuecomment-254580745:

> @ctgraham, what implication does it have for COUNTER, if we do not distinguish full text from supplementary/other material? I think it is currently not distinguished in the UsageStatistics plugin, but I will have to double check.

@ctgraham, do you maybe know?

ctgraham:
[...]

PR #2489, opened 2017-05-03
pkp/pkp-lib#1791 introduce a new counter assoc type constant for othe…

bozana:
…r files

s.  https://github.com/pkp/pkp-lib/issues/1791

This PR is for the case we would like to keep the access counts on files other than full texts...


classes/statistics/PKPStatisticsHelper.inc.php
```diff
@@ -58,6 +58,9 @@
 define('STATISTICS_YESTERDAY', 'yesterday');
 define('STATISTICS_CURRENT_MONTH', 'currentMonth');
 
+// Constant used for submission files that are not full texts
+define('ASSOC_TYPE_SUBMISSION_FILE_COUNTER_OTHER', 0x0000213);
```
asmecher: I suggest moving this definition near the other `ASSOC_TYPE_`... defines. That'll make it less likely that someone will come along and think it's an unused number.

PR #2490, opened 2017-05-03
pkp/pkp-lib#2449: Fix parameter and return documentation in UserGroup (naming and typing)

ctgraham:
Filed against pkp-lib#2449, but pkp-lib#2449 can be left open for similar fixes.

### DIFF
--- a/classes/security/UserGroup.inc.php
+++ b/classes/security/UserGroup.inc.php
@@ -100,7 +100,7 @@ class UserGroup extends DataObject {
 
 	/**
 	 * Set the "show title" flag
-	 * @param $isDefault boolean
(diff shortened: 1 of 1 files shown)


classes/security/UserGroup.inc.php
```diff
@@ -136,7 +136,7 @@ function getLocalizedName() {
 	/**
 	 * Get user group name
 	 * @param $locale string
```
asmecher: Here and elsewhere:

    @param $locale string|null

ctgraham: Good call.  New push.

asmecher:
Merged, thanks!

PR #2491, opened 2017-05-04
pkp/pkp-lib#2142: correct fall-through logic in native import when specified user has no role in the context

ctgraham:
Should finally resolve pkp/pkp-lib#2142.

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
@@ -194,13 +194,12 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 				// If the user's user group has access to this stage, select it.
 				$lastUserGroup = $userGroup->getId();
 				if (in_array($lastUserGroup, $stageUserGroupLookup, true)) {
-					$submissionFile->setUserGroupId($userGroup->getId());
 					break;
 				}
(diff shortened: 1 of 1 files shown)


plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -197,10 +197,10 @@ function handleRevisionElement($node, $stageId, $fileId) {
 					$submissionFile->setUserGroupId($userGroup->getId());
```
bozana: Then this is maybe not needed, because the user group will be set after the while loop however, because the $lastuserGroup will exist, correct?

ctgraham: We're iterating an arbitrary user's groups per the context where the import is happening.  This could result in multiple groups which have access to this stage, or none.  If multiple, we'll break on the first one we find.  If none, we'll select any group (well, the last one we found) which the user is a member of, even though it doesn't have access to the requested stage.

bozana: but the user group will be set here and after the while loop again, right? it maybe does not matter, because they should be the same...

ctgraham: You are right.  This assignment is unnecessary.

bozana:
👍

Issue #2526, opened 2017-05-15
Native Import should discriminate between errors and warnings


ctgraham:
The `PKPNativeImportExportDeployment` exposes errors which were added to processed objects via an associative array of those processed objects:
https://github.com/pkp/pkp-lib/blob/06bb9cf6bb4b06dbaf17f5cd435321ccbb6d4de2/classes/plugins/importexport/PKPImportExportDeployment.inc.php#L150-L160

This leads to obfuscated code such as:
https://github.com/pkp/omp/blob/138786415a8314d294aab512364be517b662ffbf/plugins/importexport/native/NativeImportExportPlugin.inc.php#L141-L147
"Let's fetch the processed object ids and then perform a magic inline function on them!"

The `PKPNativeImportExportDeployment` should have dedicated methods `getProcessedObjectErrors()` and `getProcessedObjectWarnings()` to clarify the coding structure, and to allow for non-fatal import errors regardless of the association type.  For example, currently OJS considers errors fatal at submission level, and non-fatal at the issue level:
https://github.com/pkp/ojs/blob/7c40c0fc9ea80115eb9f05453ae8c3deccef1049/plugins/importexport/native/NativeImportExportPlugin.inc.php#L349-L391

bozana:
Thanks a lot @ctgraham! All PRs are merged... (and GitHub closed the issue)

ctgraham:
Thanks, @bozana; I'm about to import content into our first OJS3 journal!

bozana:
:thumbsup:
:four_leaf_clover: 
:pray:

ctgraham:
pkp/ojs#1385 introduces some unset variables.

PR #2529, opened 2017-05-15
pkp/pkp-lib#2526: Create dedicated error and warning functions for PKPNativeImportExportDeployment

ctgraham:
Foundation for resolving pkp/pkp-lib#2526; requires changes in each application because the `PKPNativeImportExportDeployment::getProcessedObjectsIds()` return changes.

### DIFF
--- a/classes/plugins/importexport/PKPImportExportDeployment.inc.php
+++ b/classes/plugins/importexport/PKPImportExportDeployment.inc.php
@@ -28,7 +28,13 @@ class PKPImportExportDeployment {
 	var $_submission;
 
 	/** @var array The processed import objects IDs */
-	var $_processedObjectsIds;
+	var $_processedObjectsIds = array();
+
+	/** @var array Warnings keyed by object IDs */
+	var $_processedObjectsErrors = array();
+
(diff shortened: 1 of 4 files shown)


classes/plugins/importexport/PKPImportExportDeployment.inc.php
```diff
@@ -28,7 +28,13 @@ class PKPImportExportDeployment {
 	var $_submission;
 
 	/** @var array The processed import objects IDs */
-	var $_processedObjectsIds;
+	var $_processedObjectsIdsi = array();
```
bozana: The "i" at the end is a mistake, right?

bozana:
Is there an OJS PR that corresponds to this changes? Is it https://github.com/pkp/ojs/pull/1385?

classes/plugins/importexport/PKPImportExportDeployment.inc.php
ctgraham: Yep.  Sloppy vim'ing.

ctgraham:
Yes, this will be incorporated into OJS via https://github.com/pkp/ojs/pull/1385

PR #2530, opened 2017-05-16
This class was importing it self

murilodbva:


asmecher:
Merged -- thanks, @murilodbva!

PR #2531, opened 2017-05-16
Corrected information and importing unnecessary

murilodbva:
The annotation 'file' was wrong.
The class 'handler.validation.HandlerValidator' is imported in 'handler.validation.HandlerValidatorCustom', so it's not necessary to import 2 times.


bozana:
Thanks @murilodbva!

PR #2533, opened 2017-05-16
fixed 'continue not in loop context' fatal error on PHP7

amtuannguyen:
This caused the statistic custom report generator form in OJS 3.0.2 stuck with the loading message.


asmecher:
Thanks -- good spotting!

Issue #2500, opened 2017-05-08
Native import mismanages file operations


ctgraham:
Multiple problems:

The `NativeXMLSubmissionFileFilter` presumes to delete the temporary file after ingest:
https://github.com/pkp/pkp-lib/blob/1fc4dae0c605bdef39987a1beb32f205ad4c6b4d/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php#L221-L223
This file is, however, only a temporary file if `NativeXMLSubmissionFileFilter::handleRevisionChildElement()` processed an `embed` tag.
https://github.com/pkp/pkp-lib/blob/1fc4dae0c605bdef39987a1beb32f205ad4c6b4d/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php#L233-L264
Otherwise, it is a remote URL or a local filesystem reference.

In the case of a [remote `href` in NativeXMLSubmissionsFileFilter::handleRevisionChildElement()`](https://github.com/pkp/pkp-lib/blob/1fc4dae0c605bdef39987a1beb32f205ad4c6b4d/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php#L246-L250), fetching this file should honor the proxy settings, if any.

When called from the CLI, `NativeImportExportPlugin` checks to see if the file specified on the command line and its containing directory are writable.
https://github.com/pkp/ojs/blob/00157d0e64d980e16edf76c41ace617dc4a089f6/plugins/importexport/native/NativeImportExportPlugin.inc.php#L322-L328
This is only needed in the case of import.

bozana:
Thanks a lot @ctgraham!!!

ctgraham:
This check of file operation success is prone to inaccurate results because `filesize()` is cached:
https://github.com/pkp/pkp-lib/pull/2502/files#diff-dcf23c77841b568585f1a1bed7522a63R163

ctgraham:
The `NativeXMLSubmissionFileFilter` will die if presented with a non-existent path within the `href`.

asmecher:
@bozana, it looks like you merged something to resolve this -- is it ready to be closed?

PR #2544, opened 2017-05-22
pkp/pkp-lib#2500: various Native Import filehandling fixes

ctgraham:
* clear file statcache before checking for success of SubmissionFile import
* allow handleRevisionChildElement to delete a failed temporary file
* warn if actual filesize differs from provided filesize
* error if encoded data is invalid

Addresses a problem with pkp/pkp-lib#2500


bozana:
Very nice, thanks a lot @ctgraham!!!

Issue #2572, opened 2017-06-08
Additional hooks


bozana:
s. http://forum.pkp.sfu.ca/t/hooking-to-publishissue/30863

Add a possibility to hook to a GridHandler operation/function, as well as in the publishIssue and unpublishIssue function.

bozana:
PRs:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/2573
ojs master: https://github.com/pkp/ojs/pull/1415

bozana:
@asmecher, OK so?

asmecher:
Never mind the other comment I just made (and deleted) -- I misread the `assert` condition. This looks fine, please go ahead!

PR #2573, opened 2017-06-08
pkp/pkp-lib#2572 add operation to the LoadComponentHandler hook

bozana:
s. https://github.com/pkp/pkp-lib/issues/2572


classes/core/PKPComponentRouter.inc.php
```diff
@@ -174,6 +174,10 @@ function &getRpcServiceEndpoint($request) {
 			// time.
 			$this->_rpcServiceEndpoint = $nullVar = null;
 
+			// Retrieve requested component operation
+			$op = $this->getRequestedOp($request);
```
asmecher: It's been a while since I looked at this, so forgive a potentially silly question -- but I think it's the `page` that component URLs lack, not the `op`, no?

PR #2588, opened 2017-06-15
pkp/pkp-lib#2189 Prevent localization popovers from covering tinymce …

NateWr:
…fullscreen mode


NateWr:
Simple CSS change. Self-merged.

PR #2590, opened 2017-06-16
pkp/pkp-lib#1270 Add a notice when no editorial actions are available

NateWr:


asmecher:
Please go ahead, @NateWr!

Issue #2503, opened 2017-05-08
Add plugin version installation tool


asmecher:
Per e.g. http://forum.pkp.sfu.ca/t/installing-a-plugin-via-git/30827, there should be an easy way to register a single plugin with the system (e.g. during development) without needing to run the entire upgrade script.

bozana:
@asmecher, this does not consider some other things a plugin could need, for example the filters installation...

bozana:
PRs:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/2595

bozana:
@asmecher, could you take a look at that PR?

asmecher:
Just a couple of quick comments -- otherwise, please go ahead!

bozana:
PR:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/2607

PR #2595, opened 2017-06-19
pkp/pkp-lib#2503 Add a tool to permit easy installation of plugins

bozana:
s. https://github.com/pkp/pkp-lib/issues/2503

This considers also other plugin installation functions...

### DIFF
--- a/tools/installPluginVersion.php
+++ b/tools/installPluginVersion.php
@@ -55,6 +55,43 @@ class InstallPluginVersionTool extends CommandLineTool {
 		$pluginVersion = $versionInfo['version'];
 		$versionDao = DAORegistry::getDAO('VersionDAO');
 		$versionDao->insertVersion($pluginVersion, true);
+
+		$pluginPath = dirname($this->_descriptor);
+		$pluginCategory = preg_replace('/^plugins\./', '', $pluginVersion->getProductType());
+		$pluginCategories = array('auth', 'blocks', 'citationFormats', 'citationLookup',
+			'citationOutput', 'citationParser', 'gateways', 'generic', 'importexport',
+			'metadata', 'oaiMetadataFormats', 'paymethod', 'pubIds', 'reports', 'themes',
+		);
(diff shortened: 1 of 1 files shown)


tools/installPluginVersion.php
```diff
@@ -55,6 +55,38 @@ function execute() {
 		$pluginVersion = $versionInfo['version'];
 		$versionDao = DAORegistry::getDAO('VersionDAO');
 		$versionDao->insertVersion($pluginVersion, true);
+
+		$pluginPath = str_replace('/version.xml', '', $this->_descriptor);
```
asmecher: Use `dirname($this->_descriptor);` -- or alternately maybe it would be better to provide the plugin path to the command line for this tool, rather than the `version.xml`.

tools/installPluginVersion.php
```diff
@@ -55,6 +55,38 @@ function execute() {
...
+		$pluginCategory = str_replace('plugins.', '', $pluginVersion->getProductType());
```
asmecher: Generally speaking I prefer using regular expression matches for this kind of thing -- they'll behave more predictably when the string doesn't match the pattern. It would probably be a good idea to check the result against a whitelist, since we're then including PHP from this name.

bozana:
@asmecher, so OK?

tools/installPluginVersion.php
```diff
@@ -55,6 +55,43 @@ function execute() {
 		$pluginVersion = $versionInfo['version'];
 		$versionDao = DAORegistry::getDAO('VersionDAO');
 		$versionDao->insertVersion($pluginVersion, true);
+
+		$pluginPath = dirname($this->_descriptor);
+		$pluginCategory = preg_replace('/^plugins\./', '', $pluginVersion->getProductType());
```
asmecher: I was thinking `preg_match`, rather than `preg_replace`, and there's already a whitelist available in `Application::getPluginCategories`.

PR #2607, opened 2017-06-21
pkp/pkp-lib#2503 fix install plugin tool

bozana:
s. #2503


bozana:
@asmecher, so OK?

asmecher:
Yes, that looks good, thanks!

Issue #2604, opened 2017-06-21
[OJS] Update pt_BR locale for OJS 3.1 release


mtub:
Issue tracking the updates to the pt_BR locale.

mtub:
Closing for the release…

PR #2626, opened 2017-06-28
Update locale user.xml pt_BR

wundo:
This PR adds a missing translation and fixes some incorrect translations for pt_BR locale.


wundo:
I think this is somewhat related with #2604

wundo:
@mtub I wasn't sure how to approach this, several translations translate `Simon Fraser University` to the local language, maybe we should standardize this across translations.

Let me know if you want that I submit a pull request for that as well.

wundo:
@mtub I've updated the PR based on your review.

mtub:
Thank you, I think it is better this way. Probably wouldn't mind adding another organization, but renaming an existing organization is not such a good way. Keeping the original name is my favorite, so thank you :)

Issue #2559, opened 2017-06-01
Consider introducing autoload behavior


asmecher:
Using an autoloader, we would be able to remove the `import` function entirely. I suspect there would be a considerable performance improvement and we wouldn't need to deal with pesky maintenance of all those LOCs.

asmecher:
@kaschioudi pointed out the Composer autoloader toolset, which is documented here: https://getcomposer.org/doc/01-basic-usage.md#autoloading

At a glance, it looks like this might require us to rename all classes to `.php` instead of `.inc.php` (!!!) -- not impossible, but certainly not ideal.

asmecher:
I like what you've proposed in the PR, @kaschioudi. All I suggest is moving the OJS stuff out of `pkp-lib`, and it should be ready to merge.

asmecher:
(Also see if you can resolve the code duplication between the two autoload functions.)

kaschioudi:
PR: https://github.com/pkp/pkp-lib/pull/2627

kaschioudi:
@asmecher :  this is ready for another look.

asmecher:
Thanks, @kaschioudi. Did you add a namespace for the application too, or decide against that?

asmecher:
Looks good, @kaschioudi. Can you also port the OJS change over to OMP? Feel free to self-merge those whenever you're ready.

kaschioudi:
https://github.com/pkp/omp/pull/425

kaschioudi:
@asmecher : just to be sure. By self merge, you mean accept the PR (into master) or merge this branch into api and submission list branch (into i2163_js_with_api) ?

asmecher:
Into master; not sure if you have privileges to do that, though, so let me know if not.

kaschioudi:
@asmecher OJS is ready to go. thanks.

asmecher:
Thanks, @kaschioudi. There was a merge conflict on the `lib/pkp` submodule, so I cherry-picked in the relevant commits:

- https://github.com/pkp/ojs/pull/1443/commits/9bce9d49427d49b9b72fda178cc03fb4e0814a70
- https://github.com/pkp/ojs/pull/1443/commits/ef663c1865b58776923424033c156e015cf69f1f

...and have already committed a submodule update.

The rest were merge commits and old submodule updates. Next time, see if you can get rid of these so that only the relevant commits are left -- an interactive rebase ought to do it. I can walk you through it if you like.

PR #2627, opened 2017-06-28
custom autoloaders for ojs and pkp-lib #2559

kaschioudi:
@asmecher


kaschioudi:
@asmecher 

FYI: Because I am using `Core::cleanFileVar()` to sanitize folder names, I couldn't include `autoloaders.inc.php` directly from `bootstrap.inc.php` because dependencies classes like `Core` and `PKPString` where not loaded yet.  That's the reason why `autoloaders.inc.php` is required at the end of `PKPApplication::__constructor()`

kaschioudi:
@asmecher : Another precision: 

All folder names in namespace are lowercase to be conform with OJS. For instance, `VerySpecialClass` under let say `PKP\Services\FooBar` namespace will be instantiate like `new PKP\Services\FooBar\VerySpecialClass()` but loaded from `lib/pkp/classes/services/foobar/VerySpecialClass.inc.php`

includes/autoloaders.inc.php
```diff
@@ -0,0 +1,73 @@
...
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * @ingroup index
+ *
+ * @brief Registers custom autoloaders 
+ */
+
+// registering autoloader function for OJS namespace
+spl_autoload_register(function($class) {
+	$prefix = 'OJS\\';
```
asmecher: OJS specifics shouldn't be going into pkp-lib. I'd be OK if we put everything for all apps into the PKP namespace.

asmecher: ...but perhaps better than jamming everything into the `PKP` namespace would be to move the `spl_autoload_register` call for the application into the application repo.

kaschioudi: in such case, we would need to try to load files from 2 locations: 
- classes/
- lib/pkp/classes 

is that okay?

Also should we search following a specific order?

asmecher: Not sure if you saw my second comment, but I do think it would be better to keep the two namespaces separate, just that the OJS-related part needs to be kept in that app's repository.

We should never have a class name that appears in both places, so I don't really care about the order.

kaschioudi: Sorry I didn't see the second comment.

includes/functions.inc.php
```diff
@@ -292,4 +292,39 @@ function cygwinConversion($path) {
...
+
+	// we expect at least one folder in the namespace
+	// there is no class defined directly under classes/ folder
+	if (count($parts) < 2) {
+		return;
+	}
+
+	$className = Core::cleanFileVar(array_pop($parts));
+	$parts = array_map(function($part) {
+		return strtolower(Core::cleanFileVar($part));
```
asmecher: I don't think we can universally `strtolower` these -- looking e.g. at `lib/pkp/classes/stageAssignment`. I think everything is `lowerCamelCase`.

kaschioudi: @asmecher `lcfirst()` should do it right?

asmecher: Yes, I think so.

kaschioudi: @asmecher Done.

PR #2656, opened 2017-07-18
pkp/pkp-lib#2655 Switch template message for review request on second…

ptrckb:
… or greater round


asmecher:
Thanks, @ptrckb. One small request: rather than concatenating together parts of the email template name, could you write this in a way that uses the template name in full? The team often finds where email templates are referred to by grepping the code for the template name, and it would be missed in this case.

ptrckb:
@asmecher, updated to use the full template names.

PR #2664, opened 2017-07-21
pkp/pkp-lib#2637 Replace gif notification with svg

RahulChakravarthy:


RahulChakravarthy:
@NateWr here are the image modification for lib/pkp

PR #2679, opened 2017-07-31
pkp/pkp-lib#2592 Clearing template cache clears compiled .css files

RahulChakravarthy:
Hi @NateWr, after adding this line to the clearTemplateCache() method and testing it frequently, I believe it successfully deletes any compiled .css files in the cache folder and reloads the page with the original .css file(s). Is this what you wanted or did I misinterpret your issue description? Thanks

### DIFF
--- a/classes/template/PKPTemplateManager.inc.php
+++ b/classes/template/PKPTemplateManager.inc.php
@@ -768,6 +768,7 @@ class PKPTemplateManager extends Smarty {
 	 * Clear template compile and cache directories.
 	 */
 	function clearTemplateCache() {
+		$this->clearCssCache();
 		$this->clear_compiled_tpl();
 		$this->clear_all_cache();
 	}


RahulChakravarthy:
Hi @NateWr sorry for the late response but do you mean adding a new option to clear the css cache in the admin index.tpl page which will call $templateMgr->clearCssCache when clicked? 

Something like this ->

selection_050

NateWr:
Actually I was thinking of just adding the `TemplateManager::clearCssCache()` call into the code executed by `AdminFunctionsHandler::clearTemplateCache()` here:

https://github.com/pkp/pkp-lib/blob/master/pages/admin/AdminFunctionsHandler.inc.php#L82-L86

So it would go from:

```
	function clearTemplateCache($args, $request) {
		$templateMgr->clearTemplateCache();
		$request->redirect(null, 'admin');
	}
```

To:

```
	function clearTemplateCache($args, $request) {
		$templateMgr->clearTemplateCache();
		$templateMgr->clearCssCache();
		$request->redirect(null, 'admin');
	}
```

PR #2686, opened 2017-08-03
add initial JWT support to API

samplet:
This is the JSON Web Tokens code written during the sprint. It doesn't do much, but it's a start for using patterns based on JWT for API security. @asmecher @kaschioudi

### DIFF
--- a/classes/core/PKPRequest.inc.php
+++ b/classes/core/PKPRequest.inc.php
@@ -565,8 +565,8 @@ class PKPRequest {
 
 		$user =& Registry::get('user', true, null);
 
-		// Reference required
-		if (is_a($_this, 'PKPRequest') && is_a($_this->getRouter(), 'APIRouter') && ($token = $_this->getUserVar('apiToken'))) {
+		$router = $_this->getRouter();
+		if (is_a($router, 'APIRouter') && !is_null($handler = $router->getHandler()) && !is_null($token = $handler->getApiToken())) {
 			if ($user === null) {
 				$userDao = DAORegistry::getDAO('UserDAO');
 				$user = $userDao->getBySetting('apiKey', $token);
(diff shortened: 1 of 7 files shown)


kaschioudi:
also on the profile page, under `API Key` tab, if `api_token_secret` is empty display a message notifying users that `api_token_secret` value is missing and the token could not be generated.

kaschioudi:
Other than that, everything else seems fine to me. Thanks @samplet for your excellent contribution.

asmecher:
> @asmecher : is it required to commit composer.lock file when new dependency is added to composer.json? If no, @samplet please drop composer.lock from the PR.

Composer recommends committing the lock file to version control. I read this somewhere on the Composer site, but can't find it a the moment. There's [this Stackoverflow thread](https://stackoverflow.com/questions/12896780/should-composer-lock-be-committed-to-version-control), however.

samplet:
@kaschioudi Sorry for the late reply! I forgot that I don’t have GitHub
set up to email me, so I was sitting around thinking “gee I wonder when
they’re gonna get back to me….” :anguished:

I updated the commit so that if the secret is not set (or empty),

  * it notifies the user in the form, and
  * the decoding middleware fails (which allows the authorization middleware to consider the request
unauthorized).

I have no idea what is the current preferred system for
notifications like this. I used the `createTrivialNotification` function, since it
doesn’t make sense to me to call it a form validation error.

I will add an OJS PR with the configuration option, and link to this PR
for reference.

Thanks for the feedback! Let me know if there’s anything else. I will
respond much faster! :zap: :smile:

kaschioudi:
@asmecher Since I am not so familiar with the OJS notification framework, could you please review this block https://github.com/samplet/pkp-lib/blob/9585950c889dc27f8fa9d0cea72fc416435bec61/classes/user/form/APIProfileForm.inc.php#L62-L66 ? thanks.

Other than that I believe this PR is ready for merge. Thank you @samplet .

asmecher:
@kaschioudi, that looks fine to me. It'll display a pop-over warning on the page.

kaschioudi:
Merged. thanks @samplet

samplet:
@kaschioudi Thanks for following through with this! :smile:

Issue #2570, opened 2017-06-06
[OJS 2.x] Instructions for References are wrong


ctgraham:
In OJS 2.x, the instructions for references state:
["Please separate individual references with a blank line."](https://github.com/pkp/pkp-lib/blob/8a03743be11ceeae51e53752a6396c2c52764136/locale/en_US/submission.xml#L16).
This strongly implies that you could paste formatted references with internal linebreaks, separated by blank lines.

The actual tokenizer for references, however, very explicitly assumes that [each line is one reference](https://github.com/pkp/pkp-lib/blob/8a03743be11ceeae51e53752a6396c2c52764136/classes/citation/CitationListTokenizerFilter.inc.php#L38-L48).

It would be trivial to correct the instruction to "Please enter each reference on its own line", but the option to paste formatted references separated by blank lines seems superior.

Note that the test case provides an unhelpful and convoluted test case:
https://github.com/pkp/pkp-lib/blob/8a03743be11ceeae51e53752a6396c2c52764136/tests/classes/citation/CitationListTokenizerFilterTest.php
This does not actually represent any form of a citation and mixes several styles of line separation in one test.

ctgraham:
I'm tempted to prefer a solution such as:
```
        /**
         * @see Filter::process()
         * @param $input string
         * @return mixed array
         */
        function &process(&$input) {
                // The default implementation assumes that raw citations are
                // separated with a blank line.

                // Normalize line endings.
                if (strpos($input, "\r") !== FALSE) {
                        // Windows formatting to *nix
                        $input = str_replace("\r\n", "\n", $input);
                        // Are returns begin used as line endings?
                        $input = str_replace("\r", "\n", $input);
                }

                // Remove trailing/leading whitespace per line
                $lines = explode("\n", $input);
                $input = '';
                foreach ($lines as $line) {
                        $input .= trim($line)."\n";
                }

                // Remove trailing/leading line breaks overall.
                $input = trim($input, "\n");

                // Normalize line seperation
                $input = String::regexp_replace('/\n{2,}/s', "\n\n", $input);
                if (strpos($input, "\n\n") === FALSE) {
                        $input = str_replace("\n", "\n\n", $input);
                }

                // Break up at line endings.
                if (empty($input)) {
                        $citations = array();
                } else {
                        $citations = explode("\n\n", $input);
                }

                // Remove numbers from the beginning of each citation.
                foreach($citations as $index => $citation) {
[...]

PR #2692, opened 2017-08-07
pkp/pkp-lib#2570: Enhance citation tokenizer to support multiline citations separated by a blank line

ctgraham:
Resolves pkp/pkp-lib#2570


classes/citation/CitationListTokenizerFilter.inc.php
```diff
@@ -34,25 +34,80 @@ function CitationListTokenizerFilter() {
...
+                $input = String::regexp_replace('/\n{2,}/s', "\n\n", $input);
+                if (strpos($input, "\n\n") === FALSE) {
+                        $input = str_replace("\n", "\n\n", $input);
+                }
+
+		// Check for multiline citations
+		$separation_exists = false;
+		$indentation_exists = false;
+		$unindented_exists = false;
+                $lines = explode("\n", $input);
```
asmecher: Space-based indents.

classes/citation/CitationListTokenizerFilter.inc.php
```diff
@@ -34,25 +34,80 @@ function CitationListTokenizerFilter() {
...
+		foreach ($lines as $line) {
+			if ($line == '') {
+				$separation_exists = true;
+			} else {
+				if (substr($line, 0, 1) === "\t" || substr($line, 0, 1) === ' ') {
+					$indentation_exists = true;
```
asmecher: By convention, use `$lowerCamelCase`

classes/citation/CitationListTokenizerFilter.inc.php
```diff
@@ -34,25 +34,80 @@ function CitationListTokenizerFilter() {
...
+                        // Windows formatting to *nix
+                        $input = str_replace("\r\n", "\n", $input);
+                        // Are returns begin used as line endings?
+                        $input = str_replace("\r", "\n", $input);
+                }
+
+                // Make blank lines truely blank
+                $lines = explode("\n", $input);
+                $input = '';
+                foreach ($lines as $line) {
```
asmecher: I would think a single multiline regexp would probably replace most of this.

ctgraham:
Addressed review concerns; improved tests.

classes/citation/CitationListTokenizerFilter.inc.php
```diff
@@ -36,18 +36,65 @@ function CitationListTokenizerFilter() {
...
 	function &process(&$input) {
 		// The default implementation assumes that raw citations are
-		// separated with line endings.
-		// 1) Remove empty lines and normalize line endings.
-		$input = String::regexp_replace('/[\r\n]+/s', "\n", $input);
-		// 2) Remove trailing/leading line breaks.
+		// separated with a blank line.
+
+		// Normalize line endings.
+		if (strpos($input, "\r") !== FALSE) {
```
asmecher: PHP's single-byte string operations could potentially [scramble UTF-8 data](https://stackoverflow.com/questions/3786003/str-replace-on-multibyte-strings-dangerous) -- especially when working with single bytes e.g. `\r`. Better to use `PKPString` wrappers -- that's also why we use `String::regexp_replace` rather than `preg_replace`.

ctgraham: Maybe with high ASCII and/or non-UTF-8 multibyte encodings, but [legit UTF-8 should be fine](https://stackoverflow.com/questions/2652193/can-str-replace-be-safely-used-on-a-utf-8-encoded-string-if-its-only-given-vali).

I can see the value in using the standard `PKPString` functions everywhere, though.

PR #2696, opened 2017-08-10
pkp/pkp-lib#2678 Improve API error message for invalid stage IDs

NateWr:


mtub:
Note to @ppv1979 - this will need updating in the Russian locale, too.

PR #2722, opened 2017-08-18
minor update to submission.xml

t4x0n:
some minor errors and I changed "Fichero" by "Archivo"  (in all OJS is used "archivo" but some cases need change this word)


mtub:
@marcbria, could you have a look?

marcbria:
Perfect. 
"Archivo" is the recommended term instead of "Fichero" in our glossary.

PR #2727, opened 2017-08-21
remove empty keys from locale files

mtub:


mtub:
These all stem from using the translator plugin and were not detected when reviewed... sorry.

Issue #2725, opened 2017-08-20
[OJS] Update es_ES locale for OJS 3.1 release


mtub:
Issue tracking the updates to the es_ES locale.

mtub:
Closing  (CC: @marcbria)

PR #2769, opened 2017-09-04
[Spanish translation] The word 'Preferencias' doesn't have diacritic

jorgesumle:


marcbria:
First look over those files. This week Jordi will make a second review.
@mtub we made changes with comments, as far as we don't have write permissions over the files.
It means more mechanic work for you and more chances to introduce errors. 
Let's talks about this by mail?

mtub:
@marcbria , can I merge this? Looks okay to me.

mtub:
#2725

marcbria:
Yes, you can. :-D

Issue #2602, opened 2017-06-21
[OJS] Update fr_CA locale for OJS 3.1 release


mtub:
Issue tracking the updates to the fr_CA locale.

PR #2780, opened 2017-09-13
Some modifications on locale fr_CA

pnault:
In french we use "Rédacteur" for editor. "Éditeur" is for the publisher.


mtub:
#2602

mtub:
Thanks, @pnault !

PR #2788, opened 2017-09-14
Update of it_IT in pkp-lib, version 2.4.8

ztajoli:
This is the update of pkp-lib locale for it_IT.
It is about version 2.4.8


mtub:
Thanks, merged and cherry-picked into `ojs-dev-2_4`

PR #2796, opened 2017-09-19
Move ADODB into a submodule

greisch:
Added adodb as submodule.

Added is maintained in GitHub
https://github.com/ADOdb/ADOdb.git


asmecher:
@greisch -- thanks, this has been on my list to examine for some time. I would prefer to go with a Composer dependency, though, rather than a submodule -- we're already using Composer for many dependencies.

Have you tested these changes? One of the reasons we're still using ADODB as-is is that several patches have been applied. I haven't assessed these to see how they behave in recent years;  my suspicion is that we aren't depending as much on those modifications as we used to, but they'll need to be checked e.g. across use cases and upgrades.

greisch:
Not check (yet...) ; Iam using a old version of OJS ; I try to migrate to a new apache2 webserver and OJS passed away (PHP5 to PHP7 killme) ....
Iam making changes in the code to migrate to last version and incompatibility arraise... (in DB access so change the library to last version in ADOdb git) ; After that I realized that I was using mysql instad of "mysqli" ...

Anyway...  Its a good idea to fully test the last version of ADOdb before acept the pull request or migrate to Composer... If migration to composer is the way to go.. then reject the pull request XD

Gratings!!!

asmecher:
Thanks, @greisch! Actually, our long term plan is likely to move away from ADODB entirely, using another more modern framework instead -- though I do see that ADODB is getting more maintenance lately, so that might be a false alarm. I do prefer a Composer dependency to a submodule, so I'll close this, but I do appreciate the contribution.

Issue #2064, opened 2016-12-05
Standardize emailed URL handling


asmecher:
Some emailed URLs rely upon the email variable including the `<a href="...">...</a>` markup being included in the variable. (See `EditorDecisionWithEmailForm::initData`.) Others assume the email template will provide the markup.

For the sake of allowing the submission URL to appear in TinyMCE fields as a "pretty" variable, the markup should be included in the variable.

This is currently not done consistently -- and in some cases the same variable is coded both ways, hence http://forum.pkp.sfu.ca/t/additional-tag-in-email-in-omp-1-2-0/26082.

Watch also the upgrade scripts, which edit existing email templates in converting from plaintext to HTML.

asmecher:
@defstat, mind taking a look at this?

defstat:
@asmecher The "Insert Tag" button of the editor adds the choices for the user, using the https://github.com/pkp/pkp-lib/blob/master/templates/form/textarea.tpl#L30 (among other lines of the same file) along with some javascript code that reads the "data-variable" attribute into a javascript object and displays it as a html list upon the "Insert Tag" button click.

I have 2 questions (lets say for the submissionUrl): 
1. how do we want the variables to be displayed in the popup (when we click the "Insert Tag" button)
2. How do we want the variable to be displayed in the editor when that variable is selected from the tags. (now the submissionURL url for example is being displayed as "URL" in the editor)

Regarding the emailTemplates, there is a number of user parameters and a number of used emailTemplates that use the <a href="$parameter">$parameter</a> pattern. Do we need the parameter to pass as <a href="$parameter">$parameter</a> through all the code and change the emailTemplates to add only the $parameter?
If that's the case, we want to do the same for the second question above? (so that if we click on a variable to add the URL as is to the editor?

asmecher:
Sorry for the delay on this, @defstat --

> how do we want the variables to be displayed in the popup (when we click the "Insert Tag" button)

Generally the `common.url` locale key should be sufficient. (This reads "URL" in English.) If we have a case where that's ambiguous, e.g. if there are a couple of URLs, then we could create something more specific -- but I think "URL" is enough for the moment.

> How do we want the variable to be displayed in the editor when that variable is selected from the tags. (now the submissionURL url for example is being displayed as "URL" in the editor)

I think it makes sense for both to be the same, e.g. "URL" in both cases. Unless I misunderstand the question, which is always possible :)

[...]

PR #2808, opened 2017-09-24
pkp/pkp-lib#2064 JS Lint fix

defstat:


asmecher:
Merged, thanks -- will see if that fixes it!

Issue #440, opened 2015-03-27
Grant implicit Manager access to Adminstrators


asmecher:
...as we did for OJS 2.4. This will help e.g. our hosting environment, where we want to be admins but not show up actively in the system. Otherwise we need to self-register as Managers, which implies editorial presence and will see us notified in situations where it's not appropriate.

bozana:
@asmecher when a press is created, admin gets the Press Manager role/group, s. 
https://github.com/pkp/omp/blob/master/controllers/grid/admin/press/form/PressSiteSettingsForm.inc.php#L91
https://github.com/pkp/pkp-lib/blob/master/controllers/grid/admin/context/form/ContextSiteSettingsForm.inc.php#l102-L115

Do you maybe think that admin should also get the other managerial roles/groups i.e. Press Editor and Production Editor?

asmecher:
I'm nervous about making assumptions about roles and user groups, because these can be customized and if there are several suitable roles we'll have to pick one.

If we made Manager access implicit for Admins, we would be able to remove the current user group assignment. That would decrease ambiguity but would probably lead users to continue operating as Administrator when they actually did want to be a Manager (e.g. meaning that they would be able to self-assign in the participants grid). So I'd suggest we remove the current automatic role assignment, and consider leading the user through self-assigning roles or user groups as part of the setup wizard.

bozana:
Hi @asmecher, 100 years later... :-) I do understand to remove the automatically assignment of a journal/press manager role to the admin user. But, where and how to allow/enable the role self-assignment for the admin? At the end of the settings wizard there is the users management page. Should there be just an info for the admin, that he/she can edit his/her user and assign himself/herself further roles? Or should there be, for example, one more tab "User Roles" where he/she would do that? Or something/somehow else? THANKS!!!

asmecher:
I suggest leaving the automatic assignment as it currently stands, for informational reasons (normally it's the way people want things). But we also need to adjust the policy set so that Site Administrators implicitly have access to anything the Editor does, without needing a role assignment. That'll be valuable e.g. for institutional hosts to explore the installation, without needing to assign a role that might end up in an informational spot by accident.

bozana:
Hmmm... I am not sure if I understand it correctly :-\ 
So, we will leave the automatic manager assignment to the admin user, as till now, right?
[...]

PR #2851, opened 2017-10-09
pkp/pkp-lib#440 users row action in the admin contexts grid

bozana:
s. https://github.com/pkp/pkp-lib/issues/440#issuecomment-332942106


asmecher:
Looks good, thanks!

PR #2858, opened 2017-10-11
#2861 Updating Arabic locale before OJS 3.1

vormia:
Last hour updates to the Arabic language files to match any changes in the English version before releasing OJS 3.1


locale/ar_IQ/api.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../dtd/locale.dtd">
+
+<!--
+  * locale/en_US/api.xml
```
mtub: This should be .../ar_IQ/..., but I will fix this later.

locale/ar_IQ/countries.xml
```diff
@@ -8,11 +8,7 @@
   * Copyright (c) 2000-2017 John Willinsky
   * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
   *
-  * Localized list of countries.
-  * Please don't edit or translate. This file is automatically generated using
-  * the ISO 3166 files of Debian's iso-codes package
-  * (https://packages.debian.org/sid/all/iso-codes) using the
-  * tools/poToCountries.php tool.
+  * Localized list of countries. Based on ISO 3166-1:2006.
```
mtub: We think it's best to not edit this file directly as we generate it automatically from the Debian country list. Could you revert the changes to this file?

locale/ar_IQ/api.xml
vormia: Please do. I have some problems using online editor of Github.

locale/ar_IQ/countries.xml
vormia: It is OK if the automatically generated list is in Arabic too, otherwise it will be solely in English.
By the way, this file had been translated since OJS 3.0.0.0
I think it's your call. You may do what's best for the system.

mtub:
This is for #2861

mtub:
superseded by #2865

vormia:
Thanks @mtub.

PR #2889, opened 2017-10-18
Update manager.xml in de_DE

zuphilip:
This is a mixture of straight-forward translations, comparing the context of its use in OJS, and some educated guesses.


locale/de_DE/manager.xml
```diff
@@ -529,4 +529,50 @@
 	<message key="plugins.importexport.common.error.filesizeMismatch">Die übermittelte Dateigröße "{$expected}" und die tatsächliche Dateigröße "{$actual}" stimmen nicht überein</message>
 	<!-- native importexport keys -->
 	<message key="plugins.importexport.native.error.submissionFileImportFailed">Die Einreichungsdatei konnte nicht importiert werden</message>
+	<message key="manager.navigationMenus.form.title">Titel</message>
+	<message key="manager.navigationMenus.form.path">Pfad</message>
+	<message key="manager.navigationMenus">Navigationsmenüs</message>
+	<message key="manager.navigationMenus.administration.conditionalWarning">Dieser Link wird nur für Administratoren angezeigt.</message>
+	<message key="manager.navigationMenus.logOut.description">Über diesen Link loggt sich die Benutzerin/der Benutzer aus.</message>
+	<message key="manager.navigationMenus.loggedOut.conditionalWarning">Dieser Link wird nur angezeigt, falls die Besucherin/der Besucher eingeloggt ist.</message>
+	<message key="manager.navigationMenus.loggedIn.conditionalWarning">Dieser Link wird nur angezeigt, falls die Besucherin/der Besucher nicht eingeloggt ist.</message>
```
stweil: I suggest to avoid `/` when it is not used to separate single words or parts of words. In this case `bzw.` might be a good alternative.

zuphilip: I was following the example https://github.com/zuphilip/pkp-lib/blob/dfc41d25b70e2b41a9e569cc5569460056430841/locale/de_DE/manager.xml#L125 , but fine for me to change. @mtub What do you think?

stweil: See also the [previous discussion](https://github.com/pkp/ojs/pull/1577) for another similar case.

mtub: Hi Philip and Stefan, thanks for the suggestions. The slash as the symbol between male/female versions is deeply integrated in the de_DE locale, it is used everywhere. If we were to tackle this, we couldn't do it before the release which is due now. Replacing it with "oder", "bzw." etc where appropriate is fine, but I'd like to think some more about "der Autor/die Autorin" cases after the release, so I'll follow Philips suggestion here.

mtub:
Thanks, @zuphilip !

Issue #2715, opened 2017-08-15
update countries.xml files for locales


mtub:
`countries.xml` is supposed to be automatically generated using the ISO 3166 files of Debian's iso-codes package (https://packages.debian.org/sid/all/iso-codes) using the tools/poToCountries.php tool.

The files should be re-created every once in a while, also they are missing from the following locales (in the `master` branch):

* ar_IQ
* fi_FI
* ru_RU (though in this case there was a discussion and the Pavel's update of the file was accepted)

and missing from

* sv_SE

mtub:
@asmecher, you've created the files so far, could you have a look at this?

asmecher:
Gob of code to extract the English country list out of the JSON file Debian has started using:

    php -r '$a = json_decode(file_get_contents("iso-codes/json/iso_3166-1.json"));foreach ($a as $b) foreach ($b as $c) {echo "<country name=\"" . $c->name . "\" code=\"" . $c->alpha_2 . "\"/>\n";}'

asmecher:
@mtub, could you have a look at https://github.com/pkp/pkp-lib/pull/2728?

Note: the standards are supplied as maps from English text to the translation in question. Since e.g. many of the translations still refer to Czech Republic rather than Czechia, the mappings don't match. I've left them as they are, i.e. I haven't manually corrected this, and I'd suggest we stick with that rather than getting into fine-tuning these. The fix should be taken from upstream.

asmecher:
Merged -- @mtub, you're off the hook :)

PR #2904, opened 2017-10-20
nobody uses long official names

t4x0n:
Hi, I'm South American and most of our countries have a name like "República de Chile", "República de Perú", "Estado plurinacional de Bolivia", etc. but really nobody uses this official names!  and "Sandwitch" is a mistake


asmecher:
@t4x0n, we don't maintain our own country lists (or modifications to them) -- currently we use the Debian lists as a source, and convert them to our own XML format via a script. See https://github.com/pkp/pkp-lib/issues/2715 for details. I'm hesitant to maintain any difference to these files, but see your point -- if you're able to help identify another source that's available for all the languages we support, I'd consider switching over.

Issue #2899, opened 2017-10-19
[OJS31] Error generating the path of Custom Page in Navigation Menu


crosfield:
After creating a Custom Page with a path for example 'latex' on the site instead of 'http://%OJSPATH%/latex' OJS gives the path 'http://%OJSPATH%/navigationMenu/view/latex'. Since such the path (if the user isn't logged in) doesn't exist we have the error "The current user is not permitted to preview."

asmecher:
@crosfield, what specific version of OJS are you using? Is this pre-release code on the `master` branch?

crosfield:
@asmecher Yes, from `master` branch just now

asmecher:
@defstat, could you take a look?

crosfield:
<img width="797" alt="screen shot 2017-10-20 at 01 32 55" src="https://user-images.githubusercontent.com/8432203/31797400-c359b544-b536-11e7-8445-723bbca30e7b.png">
<img width="719" alt="screen shot 2017-10-20 at 01 31 49" src="https://user-images.githubusercontent.com/8432203/31797405-c6bff464-b536-11e7-86b5-bf67ffc6f3f5.png">

defstat:
@asmecher PR: #2903

asmecher:
Thanks, @crosfield and @defstat!

defstat:
@asmecher I did change the path but didn't check the "The current user is not permitted to preview." issue.  I will check that now.

crosfield:
Unfortunately, these changes do not help in solving the problem. The problem occurs when the user is logged out. Here are screenshots from the
<img width="802" alt="screen shot 2017-10-20 at 19 21 06" src="https://user-images.githubusercontent.com/8432203/31831582-af511736-b5cc-11e7-92a1-1232c9737eca.png">
<img width="796" alt="screen shot 2017-10-20 at 19 21 44" src="https://user-images.githubusercontent.com/8432203/31831585-b327dafc-b5cc-11e7-96ae-0ea8c08d96fa.png">

 current branch right now.

asmecher:
The fix I just merged is likely all we'll have time to include in the OJS 3.1 release. I'm going to move this issue over to OJS 3.1.1 as it sounds like there's still some work to be done -- @crosfield and @defstat, can you summarize the remaining work for later?

crosfield:
[...]

PR #2905, opened 2017-10-20
[pkp-lib] #2899 Change authorization for custom path navigationMenuItem view op

defstat:


crosfield:
Unfortunately, these changes do not help in solving the problem. The problem occurs when the user is logged out. Here are screenshots from the current branch right now.

screen shot 2017-10-20 at 19 21 06
screen shot 2017-10-20 at 19 21 44

pages/navigationMenu/NavigationMenuItemHandler.inc.php
```diff
@@ -94,12 +94,6 @@ function view($args, $request) {
 			$contextId = $context->getId();
 		}
 
-		// Ensure that if we're previewing, the current user is a manager or admin.
-		$roles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
-		if (count(array_intersect(array(ROLE_ID_MANAGER, ROLE_ID_SITE_ADMIN), $roles))==0) {
-			fatalError('The current user is not permitted to preview.');
```
asmecher: It looks like this change just removes the check entirely, correct? Shouldn't previewing rightly be limited to admins and managers?

defstat: @asmecher the preview function has the check. The view function is for displaying the  custom page only.

asmecher: OK, thanks -- will merge.

PR #2912, opened 2017-10-21
Remove duplicate timezones (found by PhpStorm)

stweil:


asmecher:
@stweil, that's third-party code -- perhaps this needs to be contributed upstream, or we need to update our version of the library?

PR #2913, opened 2017-10-21
Add missing 'break' statement (found by PhpStorm)

stweil:


stweil:
Please note that this will change the value returned for that case. I did not test whether this has unwanted consequences.

PhpStorm also reports lots of other missing break statements where it is less obvious that the code is either wrong or superfluous like in the case handles here.

asmecher:
@stweil, thanks, but this is obsolete code that we're going to remove shortly. I believe one of our devs might already be working on it, so I'd rather not cause a potential merge conflict. However, you are correct, the `break` statement is missing!

PR #2914, opened 2017-10-21
Fix string concatenation (found by PhpStorm)

stweil:


stweil:
This is only a mechanical fix and completely untested as I don't know the context, so please review carefully.

Here is another code location also reported by PhpStorm:

    $router->url($request, null, null, 'viewPlugin', null, array('rowId' => $row->getId()+1)),

Maybe here it should be an arithmetic operation, not a string concatenation, and function getId should return an integer, not a string.

asmecher:
Merged, thanks. That `getRowId()+1` is actually a mathematical operation -- IIRC it's a hack to avoid a legitimate `0` index from being evaluated to `false` somewhere in the stack. There should definitely be a FIXME there.

PR #2941, opened 2017-10-25
Update Spanish manager.xml OJS3.1

t4x0n:
in OJS 3.1 are many phrases missed in Spanish and I could help to translate it, but I do not want to be annoying... I hope to understand GitHub for making a more significant input, in this case, I close the file early and I can't edit more, sorry.


t4x0n:
I made a mistake! These phrases are translated but in different positions, not follow the same order like English file, so I closed this request!

PR #2943, opened 2017-10-25
Fix text for required API key

stweil:
Only a system administrator can set api_key_secret in the
configuration file config.inc.php, so "you" is typically wrong.

Fix the English and German texts accordingly.

Add also the missing full stop.


stweil:
I wonder why "API key" is never translated to the German "API-Schlüssel". @mtub, should we fix that?

asmecher:
The English change is fine with me -- leaving the German to you, @mtub!

mtub:
Fine with me…

asmecher:
Thanks, @stweil!

Issue #2933, opened 2017-10-24
Introduce menu caching


asmecher:
I reverted the late addition of menu caching as it was breaking the tests (apparently it inadvertently cached e.g. the user's logged-in vs. logged-out status, i.e. logging out would not cause the menus to list the menu accordingly).

This may be part of a bigger piece of work to replace the file-based caching, but that's not a foregone conclusion.

asmecher:
The revert commits are listed above (once for `master` and once for `ojs-stable-3_1_0` -- the commits are identical).

defstat:
@asmecher 
PR:
pkp-lib: #2951

asmecher:
@defstat, did this pass the tests? (You'd need to create a PR with a submodule update as per [the wiki](https://pkp.sfu.ca/wiki/index.php?title=Github_Documentation_for_PKP_Contributors#PKP_library_submodule_changes) to get Travis to test changes that would otherwise only be in `pkp-lib`.)

asmecher:
@defstat, it looks like the tests are failing with PHP7.1 on both OMP and OJS. The logs show:

````
[02-Nov-2017 16:18:31 UTC] PHP Fatal error:  Uncaught Error: Call to a member function getMenuId() on null in /home/travis/build/pkp/omp/lib/pkp/classes/navigationMenu/NavigationMenuItemAssignmentDAO.inc.php:304
Stack trace:
#0 /home/travis/build/pkp/omp/lib/pkp/classes/navigationMenu/NavigationMenuItemAssignmentDAO.inc.php(217): NavigationMenuItemAssignmentDAO->unCacheRelatedNavigationMenus(0)
#1 /home/travis/build/pkp/omp/lib/pkp/classes/navigationMenu/NavigationMenuItemDAO.inc.php(376): NavigationMenuItemAssignmentDAO->insertObject(Object(NavigationMenuItemAssignment))
#2 /home/travis/build/pkp/omp/lib/pkp/classes/navigationMenu/NavigationMenuDAO.inc.php(302): NavigationMenuItemDAO->installNodeSettings(0, Object(XMLNode), 1, NULL, 0, true)
#3 /home/travis/build/pkp/omp/lib/pkp/classes/install/Installer.inc.php(770): NavigationMenuDAO->installSettings(0, 'registry/naviga...')
[...]

PR #2951, opened 2017-10-25
[pkp-lib] #2933 nav menus caching

defstat:


classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -359,9 +354,91 @@ public function getMenuTree(&$navigationMenu) {
...
+	}
+
+	/**
+	 * Helper function to transform the json_decoded cached NavigationMenu object (stdClass) to the actual NavigationMenu object
+	 * Some changes on the NavigationMenu objects must be reflected here
+	 * @param mixed $class
+	 * @param mixed $array
+	 * @return mixed
+	 */
+	function array_to_object($class, $array) {
```
asmecher: This should be `lowerCamelCase`.

classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -359,9 +354,91 @@ public function getMenuTree(&$navigationMenu) {
...
+
+
+
+	/**
+	 * Get a tree of NavigationMenuItems assigned to this menu
+	 * @param $navigationMenu \NavigationMenu
+	 *
+	 * @return array Hierarchical array of menu items
+	 */
+	public function getMenuTree(&$navigationMenu) {
```
asmecher: I don't think the reference is required.

classes/services/PKPNavigationMenuService.inc.php
defstat: Done

classes/services/PKPNavigationMenuService.inc.php
defstat: I think re reference is needed. The navigationMenu needs to "be changed" so that its menuTree array property is populated. (after the call the same navigationMenu is being used by the code)

asmecher: OK, thanks!

PR #3005, opened 2017-11-02
delete repeated key English submission.xml

t4x0n:
submission.event.participantAdded  was repeated in two lines 277 and 288

### DIFF
--- a/locale/en_US/submission.xml
+++ b/locale/en_US/submission.xml
@@ -274,6 +274,7 @@
 	<message key="submission.email.confirmDeleteLogEntry">Are you sure you want to delete this email log entry?</message>
 	<message key="submission.email.deleteLogEntry">Delete Log Entry</message>
 	<message key="submission.emailLog">Email Log</message>
+	<message key="submission.event.participantAdded">{$name} was assigned to this submission as a {$userGroupName}.</message>
 	<message key="submission.event.participantRemoved">"{$name}" ({$username}) is removed as a {$userGroupName}.</message>
 	<message key="submission.event.backToEventLog">Back to Event Log</message>
 	<message key="submission.event.confirmClearLog">Are you sure you want to clear the event log for this submission?</message>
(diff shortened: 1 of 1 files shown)


locale/en_US/submission.xml
```diff
@@ -274,7 +274,7 @@
 	<message key="submission.email.confirmDeleteLogEntry">Are you sure you want to delete this email log entry?</message>
 	<message key="submission.email.deleteLogEntry">Delete Log Entry</message>
 	<message key="submission.emailLog">Email Log</message>
-	<message key="submission.event.participantAdded">"{$name}" ({$username}) is added as a {$userGroupName}.</message>
+	<message key="submission.event.participantAdded">{$name} was assigned to this submission as a {$userGroupName}.</message>
```
mtub: Fully agree with deleting the duplicate key, but why did you remove the username var? It's also present in translations, so I'd rather stick with the more complete version and lose the other key...

t4x0n: Hi @mtub, I did delete it because of the order!  now we see in OJS the last form of the variable (like in line 288) and when I did delete it, I did copy this line and paste to replace the line 277, this way not made changes in that we see, otherwise if we would keep the line 277 with variable "({$username})" we will see some changes in this phrase and would show not only the name of the user but also the username. Of course, if you have a preference for that way I can add the username variable before merge this PR.

sorry, but was difficult explain in English I hope you understand.

mtub: @asmecher , thoughts?

asmecher: The variables available to that locale key are...
- `name`
- `username`
- `userGroupname`
...so I would lean towards the form that uses all of these, but it's a matter of preference, not technical correctness.

t4x0n: Yes, it's a matter of preference, I just was want to keep the current way (without username):
image
If you prefer to add (username) is not a problem. Indeed I was thinking about that, and if we use both name and username we are able to find users with similar names, but as the username is unique, this can help to make difference between similar names! So, in this case, we could maintain both name and username... what do you think @mtub?

mtub: I am definitely in favor of keeping the one with more information, including the username. Could you update your PR and keep that one?

t4x0n: of course! ;)

t4x0n:
@mtub is done!

mtub:
thanks!

Issue #2972, opened 2017-10-27
Untranslated locale keys in Roles select menu


mfelczak:
Using a new OJS 3.1 instance installed via git.

![roles](https://user-images.githubusercontent.com/200793/32112752-17509854-bb0c-11e7-9ae1-628ea2f6c57a.png)

asmecher:
From `ojs-stable-3_1_0` or `master`?

mfelczak:
ojs-stable_3_1_0

mfelczak:
for both the ojs and pkp-lib repos

asmecher:
@defstat, any ideas?

asmecher:
Committed to both `ojs-stable-3_1_0` and `master`. For OJS 3.1.0, apply https://github.com/pkp/pkp-lib/commit/92bba5fdec1d056634c986e21338c9b6091f896c in `lib/pkp` and https://github.com/pkp/ojs/commit/6fe1663de8eb21a59821f9bce79c66be01c6c473 in the base directory.

mfelczak:
Thanks @asmecher . It looks like 92bba5f includes 'Press Assistant' which sounds OMP-specific -- should this not be just 'Assistant' for OJS?

asmecher:
Oops -- corrected, thanks (also squinting at as many of the locales as I can).

PR #3019, opened 2017-11-06
Update German translation (issue #2972)

stweil:

### DIFF
--- a/locale/de_DE/common.xml
+++ b/locale/de_DE/common.xml
@@ -436,7 +436,7 @@
 	<message key="notification.editedAuthor">Autor/in bearbeitet.</message>
 	<message key="notification.confirmError">Es gab einen Fehler beim Bestätigen Ihrer Registrierung.</message>
 	<message key="notification.confirmSuccess">Sie haben sich erfolreich registriert.</message>
-	<message key="notification.email">Bitte eine E-Mail über diese Arten von Benachrichtigungen zuschicken.</message>
+	<message key="notification.email">Bitte keine E-Mail über diese Arten von Benachrichtigungen zuschicken.</message>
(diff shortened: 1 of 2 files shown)


dennmuel:
The English text was negated three years ago ([11c07ae](https://github.com/pkp/pkp-lib/commit/11c07aeacf5b814c1f3c6c429d62ab73d9eb7312)). Might well be other locales haven't been updated.
See the respective line in the English locale [here](https://github.com/pkp/pkp-lib/blob/2d3a99b488249361b254036ecfcc40fbd9f4b018/locale/en_US/common.xml#L443).

Maybe just "für diese Arten von Benachrichtigungen"? But honestly I think that users will get it one way or another as long as the checkbock says what it does and does what it says. :)

stweil:
Thanks for investigating that issue. I think the commit which negated the text without renaming the key was not a good idea. I strongly suggest to rename the key from "notification.email" to "notification.noEmail" and would add a commit which does that as soon as the project maintainers agree.

mtub:
I agree that the initial renaming was not a very good thing to do, and it seems as if almost every other locale still has this wrong. Changes like these should be done by introducing a new key.

Waiting for @asmecher to decide on where to go from here regarding the wrong locale keys for notification.Email. Okay with the other changes in this PR.

stweil:
@asmecher, I suggest to merge this PR first then before going on with renaming of the key.

asmecher:
Thanks, @stweil and @mtub. I agree that negating the locale key was a bad choice here -- in the future the key should be renamed, and existing translations should be adapted (when confident about the language) and removed when not. I've merged this PR, and removed incorrect locales, updating the ones that I'm confident about.

PR #3021, opened 2017-11-06
Switch URL for pkp.sfu.ca from HTTP to HTTPS

stweil:

### DIFF
--- a/classes/context/Context.inc.php
+++ b/classes/context/Context.inc.php
@@ -411,7 +411,7 @@ class Context extends DataObject {
 	/**
 	* Retrieve a statistics report pre-filtered on this context.
 	*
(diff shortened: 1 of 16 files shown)


stweil:
I did not change the URLs for plugins (`PLUGIN_GALLERY_XML_URL`), namespace and `base_url`. Should those URLs be updated, too?

mtub:
@asmecher , I am very much in favor of replacing the links. However, this sheds some light on dead links in our PHP code, I will open another issue to address these problems.

@stweil , I think http://pkp.sfu.ca/ojs/xml/plugins.xml can be updated as well. Wouldn't care too much about the base_url in the default config sample.

mtub:
see #3044 for bugzilla issue

Issue #3015, opened 2017-11-03
New article components have NULL entry_key


mfelczak:
When I add new article components in OJS 3.1 the entry_key is set to NULL in the genres table. It's not possible to then use GenreDAO::getByKey to retrieve these new entries.

bozana:
Is this then a problem somewhere?

mfelczak:
Hi @bozana, I'm writing a plugin where I'd like to retrieve a custom article component by key.

bozana:
Does the plugin then uses a controlled set of keys? -- just interested in requirements/cases...

mfelczak:
Hi @bozana, a new article component would be created manually and then the component would be fetched by key within the plugin. That being said, another acceptable approach would be to create the article component within the plugin and then set/save the key for future retrievals.

bozana:
@asmecher, do you think it would be a good solution to enable the editing/inserting of the custom keys for custom genres/components? I think they cannot be used in the core code -- because we do not know what they could be -- but maybe it will be helpful for other plugins that would need it?

asmecher:
Hmm, I don't generally want to rely on keys for much, but I agree that plugins are a valid case.

bozana:
PRs:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/3031
ojs master: https://github.com/pkp/ojs/pull/1673 (only submodule update)

bozana:
@mfelczak, would you like or could you test the PR?

bozana:
One question maybe: I allow all characters for genre/component keys. Do you maybe see any problem in there? -- e.g. when I check that they are unique. @asmecher, @mfelczak ?

bozana:
@asmecher, could you code review the PR above?

mfelczak:
Hi @bozana, given that it's been possible to add new components with NULL keys up until now, allowing empty keys would follow that pattern. That being said we could always check that non-NULL keys are still unique so that lookup by non-NULL key will return a single result.

bozana:
Yes, I see it the same way. Thanks @mfelczak!

asmecher:
I'd also suggest encouraging the same limitations on these keys that we use elsewhere for symbolic names, e.g. see the regexp in `ContextSiteSettingsForm.inc.php`.

asmecher:
Thanks, just a few comments!

bozana:
Thanks @asmecher! I made the changes you have suggested and will merge when the tests pass...

PR #3031, opened 2017-11-08
pkp/pkp-lib#3015 enable editing of the custom genre keys

bozana:
s. https://github.com/pkp/pkp-lib/issues/3015

### DIFF
--- a/controllers/grid/settings/genre/form/GenreForm.inc.php
+++ b/controllers/grid/settings/genre/form/GenreForm.inc.php
@@ -43,8 +43,15 @@ class GenreForm extends Form {
 		$this->setGenreId($genreId);
 		parent::__construct('controllers/grid/settings/genre/form/genreForm.tpl');
 
+		$request = Application::getRequest();
+		$context = $request->getContext();
+
+		$genreDao = DAORegistry::getDAO('GenreDAO');
+
 		// Validation checks for this form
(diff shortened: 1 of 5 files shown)


controllers/grid/settings/genre/form/GenreForm.inc.php
```diff
@@ -71,6 +77,8 @@ function initData($args, $request) {
 				'category' => $genre->getCategory(),
 				'dependent' => $genre->getDependent(),
 				'supplementary' => $genre->getSupplementary(),
+				'key' => $genre->getKey(),
+				'readOnly' => $genre->isDefault(),
```
asmecher: I'd suggest renaming this to `keyReadOnly` or something -- as is, it suggests that the entire form might be read-only.

templates/controllers/grid/settings/genre/form/genreForm.tpl
```diff
@@ -31,6 +31,9 @@
 {fbvFormSection title="manager.setup.groupType" for="category"}
 	{fbvElement type="select" id="category" from=$submissionFileCategories selected=$category translate=false}
 {/fbvFormSection}
+{fbvFormSection title="manager.setup.genres.key" for="key"}
+	{fbvElement type="text" id="key" value=$key maxlength="30" readonly=$readOnly size=$fbvStyles.size.SMALL}
```
asmecher: I think it might be necessary to add a note about this, since it's not at all evident what it's for. Something like: "An optional short symbolic identifer for this genre."

PR #3065, opened 2017-11-15
set error code on authorization failures

kaschioudi:
@NateWr @asmecher

### DIFF
--- a/classes/core/APIRouter.inc.php
+++ b/classes/core/APIRouter.inc.php
@@ -96,8 +96,30 @@ class APIRouter extends PKPRouter {
 
 		$sourceFile = sprintf('api/%s/%s/index.php', $this->getVersion(), $this->getEntity());
 
+		$dispatcher = $this->getDispatcher();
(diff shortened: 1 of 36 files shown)


classes/core/APIRouter.inc.php
```diff
@@ -96,8 +96,30 @@ function route($request) {
 
 		$sourceFile = sprintf('api/%s/%s/index.php', $this->getVersion(), $this->getEntity());
 
+		$dispatcher = $this->getDispatcher();
+		$dispatcher->setAuthorizationDenialCustomHandler(function($authorizationErrorCode) {
+			$msg = '';
```
asmecher: Since there's a `default:` clause, I think this is dead code.

kaschioudi: I believe If the variable does not get created before the switch statement, PHP will trigger a `Notice` under strict mode because the variable is being referenced from an outer scope (line 113).

Issue #3011, opened 2017-11-02
Pass FormValidatorURL error message to js url validation


MarHerUMR:
When adding a _FormValidatorURL_ check to a form field the input of the field will be validated on the onChange event.
When the validation fails a error label will be added: `<label for="foo" class="error">Please enter a valid URL.</label>`

The error message will always be the same even when you have a error message passed to the _FormValidatorURL_. I think it would be helpful to show the error message passed to the _FormValidatorURL_ instead of the more general error message.

MarHerUMR:
I had a quick look into it. jquery-validation according to the [documentation](https://jqueryvalidation.org/reference/#link-error-messages) can take custom error messages from the _data-msg-url_ attribute. So this could be achieved by passing the $message to the input field as _data-msg-url_.

asmecher:
Sorry, @MarHerUMR, I think I'm probably operating at lower-than-normal capacity :) but can you provide a motivating example?

MarHerUMR:
I have a form field where you can enter the url to your facebook profile / facebook page representing your journal. I've added two validators to the form field. One validating the url as a url and a regex validator to make shure it is a Facebook url. On either error I want to tell the user to use a valid Facebook url. With the data-msg-url it would be possible to present the user the hint about the invalid _Facebook_ url even before sending the form. And it would be clear right away that you have to enter a valid _Facebook_ url, not just any url.

asmecher:
OK, that makes sense too. As I noted at https://github.com/pkp/pkp-lib/issues/3008 I'm hesitant to invest too much time in refining the form builder vocabulary toolset, but I'd be happy to review and merge a contribution that accomplished this.

MarHerUMR:
OK, I will have a look into it over the weekend. Hopefully I can contribute for both issues. Do you have, of the top of your head, a suggestion what documentation (besides the fbv wiki page) I should read before getting into it?

asmecher:
No suggestions; I think the form builder vocab code is a little contorted but not especially deep.

asmecher:
Merged https://github.com/pkp/pkp-lib/pull/3079, thanks!

PR #3079, opened 2017-11-18
pkp/pkp-lib#3011 Adding url validation error message and localization…

MarHerUMR:
… for placeholder


asmecher:
Thanks, @MarHerUMR. No objections here -- merged.

Issue #3080, opened 2017-11-19
NavigationMenus - Custom Page Error on Site level


defstat:
See https://forum.pkp.sfu.ca/t/upgrade-ojs2-to-ojs31-errors-occurred-in-the-backend

defstat:
@asmecher 

pkp-lib: #3081
ojs: pkp/ojs#1711
OMP: pkp/omp#472

defstat:
@asmecher 

pkp-lib: #3082
ojs: pkp/ojs#1712

PR #3082, opened 2017-11-19
[pkp-lib] #3080 CurrentContext not available on Site level fix stable

defstat:


templates/controllers/grid/navigationMenus/form/navigationMenuItemsForm.tpl
```diff
@@ -48,7 +48,11 @@
 				{fbvFormSection title="manager.navigationMenus.form.path" for="path" required="true"}
 					{fbvElement type="text" id="path" value=$path required="true"}
 					<p>
-						{url|replace:"REPLACEME":"%PATH%"|assign:"exampleUrl" router=$smarty.const.ROUTE_PAGE context=$currentContext->getPath() page="navigationMenu" op="view" path="REPLACEME"}
+                        {if $currentContext}
```
asmecher: Space-based indents. Please config your editor to use tabs, and search/replace for these.

templates/controllers/grid/navigationMenus/form/navigationMenuItemsForm.tpl
```diff
@@ -48,7 +48,11 @@
...
+                            {url|replace:"REPLACEME":"%PATH%"|assign:"exampleUrl" router=$smarty.const.ROUTE_PAGE context=$currentContext->getPath() page="navigationMenu" op="view" path="REPLACEME"}
```
asmecher: Are you sure this is necessary? When `context` is not specified, the current context's path should be used already.

asmecher: (I think an equivalent but shorter fix would just be to remove the `context=...` parameter from the `{url ...}` call that was originally here.)

defstat: You are right @asmecher ... I changed it following your suggestion.

asmecher:
Merged and ported to `master`, thanks!

PR #3095, opened 2017-11-22
bug fix: emailTemplates from plugins are not getting installed when u…

riese-zbw:
…sing the webinterface of OJS 3.1 to install a plugin and only the default language is in installed.

Hi, 

there seems to be a small bug in pkp-lib:`/classes/plugins/PluginHelper.inc.php`.
The `$params` array seems to be missing the default/current locale.
This causes `Plugin->installEmailTemplateData()` to iterate over an empty array `$installer->installedLocales`, which results in no emailTemplate of the plugin getting installed.

Tested with OJS 3.1.0, PHP 5.6 and only one language in OJS installed (`en_US`), using the webinterface  
`Settings->Website->Plugins->Installed Plugins->Upload A New Plugin`

### DIFF
--- a/classes/plugins/PluginHelper.inc.php
+++ b/classes/plugins/PluginHelper.inc.php
@@ -122,6 +122,7 @@ class PluginHelper {
 			if(!is_file($installFile)) $installFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/xml/defaultPluginInstall.xml';
 			assert(is_file($installFile));
 			$params = $this->_getConnectionParams();
+			$params['locale'] = LOCALE_DEFAULT;
 			$installer = new Install($params, $installFile, true);
 			$installer->setCurrentVersion($pluginVersion);
 			if (!$installer->execute()) {
--- a/classes/plugins/PluginHelper.inc.php
+++ b/classes/plugins/PluginHelper.inc.php
@@ -121,11 +121,11 @@ class PluginHelper {
 			$installFile = $pluginDest . '/' . PLUGIN_INSTALL_FILE;
 			if(!is_file($installFile)) $installFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/xml/defaultPluginInstall.xml';
 			assert(is_file($installFile));
-                        $request = Application::getRequest();
-                        $site = $request->getSite();
-                        $params = $this->_getConnectionParams();
+			$siteDao = DAORegistry::getDAO('SiteDAO');
+			$site = $siteDao->getSite();
+			$params = $this->_getConnectionParams();
 			$params['locale'] = $site->getPrimaryLocale();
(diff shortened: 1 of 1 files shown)


asmecher:
At a glance I think that would work for `en_US`, but I think it would be necessary to set `additionalLocales` to the site's list of installed locales in order for other locales to get installed. Did you try this with locales other than `en_US`?

riese-zbw:
You are right, I overlooked `$pararms['additionalLocales']` is also not getting set.

Now I see two possible solutions to the Problem:
1. adding `$pararms['additionalLocales'] = array_keys(AppLocale::getAllLocales());` to `PluginHelper.inc.php`
2. forget about adding  `$pararms['locale']` and `$pararms['additionalLocales']` to `PluginHelper.inc.php` and also remove lines 59 to 66 from `PKPInstall->preInstall()` and fallback on lines 149 to 155 in `Installer->preInstall()`

I'm currently not sure which is the better solution.
If you have any suggestions, then I can update the request accordingly.

asmecher:
@riese-zbw, at a glance I would suggest using the current settings from the `Site` object -- `$site->getPrimaryLocale()` for `$params['locale']`, and `$site->getSupportedLocales()` for `additionalLocales`.

riese-zbw:
@asmecher: Hi, I updated the patch as you suggested and also added the code to the method `upgradePlugin` in `PluginHelper`.

classes/plugins/PluginHelper.inc.php
```diff
@@ -240,7 +244,11 @@ function upgradePlugin($category, $plugin, $path, &$errorMsg) {
 
 			$upgradeFile = $pluginDest . '/' . PLUGIN_UPGRADE_FILE;
 			if($fileManager->fileExists($upgradeFile)) {
+                                $request = Application::getRequest();
+                                $site = $request->getSite();
```
asmecher: Rather than using the `Request` object, which won't be available when invoked from the command line, I'd suggest using

    $siteDao = DAORegistry::getDAO('SiteDAO').
    $site = $siteDao->getSite();

classes/plugins/PluginHelper.inc.php
```diff
@@ -240,7 +244,11 @@ function upgradePlugin($category, $plugin, $path, &$errorMsg) {
...
 				$params = $this->_getConnectionParams();
+                                $params['locale'] = $site->getPrimaryLocale();
+                                $params['additionalLocales'] = $site->getSupportedLocales();
```
asmecher: It looks like your editor is using spaces for indents; could you change these to tab-based indents?

riese-zbw:
@asmecher, I updated the patch as you suggested and fixed the line indentation (tabs/spaces).

asmecher:
Excellent -- thanks for contributing, @riese-zbw!

Issue #2958, opened 2017-10-26
Improve author/user names


asmecher:
- [ ] Support multilingual names (important for users that work in e.g. both Cyrillic and Latin character sets)
- [ ] Consider whether to remove the Western-centric First, Middle, Last name divisions

asmecher:
@jmacgreg and @ctgraham, IIRC both of you worked a couple of sprints ago on a document that surveyed the third-party standards we support to see whether metadata mismatches would be imposed by e.g. having multilingual names, or not having first/middle/last. Does either of you have that handy?

crosfield:
I think the second problem is easily solved. In Russia and some other Eastern European countries instead of the Middle Name is a Patronymic Name (father's name). It is always specified in the articles, for example, in Russian journals. The First Name is always the proper Name, and Last Name - Surname (family name).

asmecher:
@crosfield, the second issue isn't really a problem in Western/Eurocentric cultures; it's definitely a problem elsewhere. See https://www.w3.org/International/questions/qa-personal-names, which recommends just going with a single field. However, as with multilingual names, the problem will be finding a model that plays well with all the various metadata standards we need to provide data to.

Ph-We:
Hi @asmecher, thank you again for addressing this issue, that is quite important for many of us!

As to Russian (and I believe, any Slavic) names, we are mostly happy with the ‘simplified’ one-field approach introduced in most bibliographic applications, like EndNote, Zotero, etc.
This approach allows (not requires) to split any name into 2 parts: 1) last name; 2) any other names. Usually the comma sign is used as a separator. 
This split is important, since different output styles have various requirements. Some require putting the second part first, other require abbreviating all the words in the second part, etc.

If the separator is optional, it might solve the problem with the ‘Corporate Author’ names entry. Now some of our users have problems in OJS, when trying to input such names, since any name require at least two parts (first name and last name). And those parts are displayed differently in different formats.

EndNote additionally allows a second separator for multiple-word last names. See
http://lib.tnnua.edu.tw/ezfiles/10/1010/img/427/EndNote.pdf
p. 57-58
![image](https://user-images.githubusercontent.com/22288161/32047901-20d0828e-ba50-11e7-906e-06fcf1753c10.png)

ctgraham:
A brief summary of the sprint outcomes is:

    Discriminate name localization based on script changes
[...]

PR #3128, opened 2017-11-28
Authors multilanguage

litvinovg:
Commit as a solution for bug  lib/pkp issue "Improve author/user names #2958 "

### DIFF
--- a/classes/identity/Identity.inc.php
+++ b/classes/identity/Identity.inc.php
@@ -29,9 +29,9 @@ class Identity extends DataObject {
 	 */
 	function getFullName($lastFirst = false) {
 		$salutation = $this->getData('salutation');
-		$firstName = $this->getData('firstName');
-		$middleName = $this->getData('middleName');
-		$lastName = $this->getData('lastName');
+		$firstName = $this->getLocalizedFirstName();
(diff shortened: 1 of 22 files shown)


classes/identity/Identity.inc.php
```diff
@@ -44,51 +44,57 @@ function getFullName($lastFirst = false) {
...
 
 	/**
 	 * Set first name.
 	 * @param $firstName string
+	 * @param $locale string
 	 */
-	function setFirstName($firstName) {
-		$this->setData('firstName', $firstName);
-	}
+        function setFirstName($firstName, $locale = null) {
```
asmecher: (There are some space-based indents here -- please set your editor to use tabs.)

classes/identity/Identity.inc.php
```diff
@@ -269,6 +275,30 @@ function getBiography($locale) {
 	function setBiography($biography, $locale) {
 		$this->setData('biography', $biography, $locale);
 	}
+	 
+	/**
+         * Get the localized firstName for this author
```
asmecher: Suggest using "for this identity", not "for this author". This is the parent class of both Author and PKPUser. (Also below)

PR #3131, opened 2017-11-28
Update submission.xml

josekarvalho:
Changing "below" to "above" on the language selection as the combo box is on top of the text.


asmecher:
Thanks! Merged and forward-ported to `master`.

PR #3155, opened 2017-12-04
Adding missing es_ES chains for API key feature (and example name change)

marcbria:
Adding missing es_ES chains for API key feature.

Secondary minor change to keep initials, but modify the example name to a latin, with a non existing name.
(Would be nice a tribute to "Aaron Hillel Swartz" here for all langs?).

@mtub in "pkp-lib", should I work against "ojs-stable-3.1.0" or "master"? I PRed to ojs-stable-3.1.0...
@alec what about a wikipage (or someplace) for devs and translators to know the right version to contrib?

### DIFF
--- a/locale/es_ES/user.xml
+++ b/locale/es_ES/user.xml
@@ -35,7 +35,7 @@
 	<message key="user.feminine">F</message>
 	<message key="user.firstName">Nombre</message>
 	<message key="user.gender">Sexo</message>
-	<message key="user.initialsExample">Luz Ruiz Plasencia = LRP</message>
+	<message key="user.initialsExample">Juan Anselmo Seguí = JAS</message>
 	<message key="user.initials">Iniciales</message>
 	<message key="user.interests">Intereses de revisión</message>
 	<message key="user.interests.description">(Para separar los intereses pulse la tecla Intro o de coma)</message>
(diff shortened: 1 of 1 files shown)


mtub:
Hi @marcbria, best to send those PRs to `master`, will (for a time) cherry-pick to the stable branch if feasible. Don't worry about this PR, will cherry-pick to master. I would always recommend working with the master branches (OCS being the exemption from the rule).

t4x0n:
hi @mtub, this PR is important for Spanish translation... please not forget merge with the master branch.

Thank you!

locale/es_ES/user.xml
```diff
@@ -35,7 +35,7 @@
 	<message key="user.feminine">F</message>
 	<message key="user.firstName">Nombre</message>
 	<message key="user.gender">Sexo</message>
-	<message key="user.initialsExample">Luz Ruiz Plasencia = LRP</message>
+	<message key="user.initialsExample">Juan Anselmo Seguí = JAS</message>
```
asmecher: Just curious, why does this PR include a name change?

t4x0n: I don't know, maybe @marcbria can explain that. The only difference I can see is Anselmo is a second name and Ruiz is a last name, in this commit the most important chains are for API key feature

marcbria: @asmecher you catch us. :-)

During the translation sometimes you get bored. Lots of files, sometimes mechanic tasks... you know.

When we took this file to add the missing API chains, we noticed the sample name was changed in a former commit, so thought was funny changing this name in every release.

It's not an important change. As "Luz Ruiz Placencia" our new proposal is also a non existing name as is "Alan Smithee" but in latin context. 

In short, if you prefer keeping changes to minimal, "Luz Ruiz..." is a good name to keep.

marcbria: Take a look to the comments around this file:
"Secondary minor change to keep initials, but modify default name to a latin, with a non existing name. (Would be nice a tribute to "Aaron Hillel Swartz" here for all langs?)."

But thinking it twice, including Aaron's name in every OJS will hide his fingerprint in google... and hide Aaron's publications it's not the best way to tribute. :-)

asmecher: Very happy to let the occasional wink sneak into the software -- for example, I always make sure to use `abbrev` instead of `abbreviation` in the code. Just making sure someone hadn't snuck their own name into it for fame and fortune.

asmecher:
Merged and cherry-picked forward to `master` (with a little conflict resolution), thanks!

PR #3186, opened 2017-12-14
pkp/pkp-lib#3188 Fix unexpected title on message for a declined submission

ptrckb:
The subject of the message an author receives for a declined submission ('Unsuitable submission') was not the subject of the template used in the 'Decline Submission' dialog ('Editor Decision') due to the use of different templates during message preparation (`EDITOR_DECISION_DECLINE`) and form execution (`SUBMISSION_UNSUITABLE`).


asmecher:
Thanks, @ptrckb, good catch. I've filed an issue for this at https://github.com/pkp/pkp-lib/issues/3188 -- this fixes the evident issue, but once this is merged, the "Unsuitable" email is not used and should be removed from the list of templates. I'll tackle that separately at the issue link.

PR #3191, opened 2017-12-14
smarty 3

Vitaliy-1:
Update of PKPTemplateManager class in order to adapt OJS for Smarty 3.1.31
Note in order to make it work on my local machine I had replaced depreciated Smarty methods not only here, but throughout the OJS code (just find & replace).
Still, some problems with templates persist in loading templates for admin dashboard.

### DIFF
--- a/classes/template/PKPTemplateManager.inc.php
+++ b/classes/template/PKPTemplateManager.inc.php
@@ -42,7 +42,7 @@ define('CSS_FILENAME_SUFFIX', 'css');
 
 import('lib.pkp.classes.template.PKPTemplateResource');
 
-class PKPTemplateManager extends Smarty {
+class PKPTemplateManager extends SmartyBC {
 	/** @var array of URLs to stylesheets */
 	private $_styleSheets = array();
 
@@ -61,6 +61,15 @@ class PKPTemplateManager extends Smarty {
 	/** @var PKPRequest */
 	private $_request;
 
+	/** @var string path to the active theme templates */
+	private $theme_template_dir;
+
+	/** @var string path to the old templates */
+	private $app_template_dir;
+
+	/** @var string path to the new templates */
+	private $core_template_dir;
+
 	/**
 	 * Constructor.
 	 * Initialize template engine and assign basic template variables.
@@ -76,12 +85,24 @@ class PKPTemplateManager extends Smarty {
 		$baseDir = Core::getBaseDir();
 		$cachePath = CacheManager::getFileCachePath();
 
+
+		// Set the theme template dir
+		$activeTheme = null;
+		$context = $request->getContext();
+		if (is_a($context, 'Context')) {
+			$activeTheme = $context->getSetting('themePluginPath');
+		} else {
+			$site = $request->getSite();
+			$activeTheme = $site->getSetting('themePluginPath');
+		}
+
+		$this->theme_template_dir = $baseDir . DIRECTORY_SEPARATOR . "plugins" . DIRECTORY_SEPARATOR . "themes" . DIRECTORY_SEPARATOR . $activeTheme . DIRECTORY_SEPARATOR . 'templates';
 		// Set the default template dir (app's template dir)
 		$this->app_template_dir = $baseDir . DIRECTORY_SEPARATOR . 'templates';
 		// Set fallback template dir (core's template dir)
 		$this->core_template_dir = $baseDir . DIRECTORY_SEPARATOR . 'lib' . DIRECTORY_SEPARATOR . 'pkp' . DIRECTORY_SEPARATOR . 'templates';
 
-		$this->template_dir = array($this->app_template_dir, $this->core_template_dir);
+		$this->template_dir = array($this->theme_template_dir, $this->app_template_dir, $this->core_template_dir);
 		$this->compile_dir = $cachePath . DIRECTORY_SEPARATOR . 't_compile';
 		$this->config_dir = $cachePath . DIRECTORY_SEPARATOR . 't_config';
 		$this->cache_dir = $cachePath . DIRECTORY_SEPARATOR . 't_cache';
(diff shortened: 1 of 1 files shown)


Vitaliy-1:
Hi @asmecher 

Forget to add the Smarty3  import statement in the commit:
require_once('./lib/pkp/lib/vendor/smarty/smarty/libs/SmartyBC.class.php');

Also, have checked once more, changes in the PKPTemplateManager class should be enough to run OJS.
Deprecated methods and their alternatives:
register_modifier => registerPlugin
register_block => registerPlugin
register_function => registerPlugin
get_template_vars => getTemplateVars
_smarty_include => https://www.smarty.net/forums/viewtopic.php?p=72534&sid=452c67adf3f59a4b02ab6adc6bd432a7
(This method was overloaded but didn't see you used it anywhere)

Think. problems on the admin front-end could be managed by reviewing assigned to the Smarty variables. Think one problem is with assigned URL variable (see the error message): 
https://github.com/Vitaliy-1/pkp-lib/blob/master/classes/template/PKPTemplateManager.inc.php#L1411
Also, I suppose upgrade process shouldn't lead to any significant changes in the code. At the worst, we will need to add custom depreciated methods in the Smarty child. 

Hope you are considering to make Smarty upgrade and my modest help would speed up the process :)

asmecher:
@Vitaliy-1, I've checked this out and have a few suggestions:
- The `import` statement shouldn't be necessary, as the current OJS3 release uses Composer autoloaders
- You'll need to update `lib/pkp/composer.json` to add the new Smarty version and use Composer to update the dependencies from this list
- `lib/pkp/classes/plugins/ThemePlugin.inc.php` needs the following line:

       array_unshift(
           $templateManager->template_dir,
           $this->_getBaseDir('templates')
       );

     ... changed to ...

      $templateManager->addTemplateDir($this->_getBaseDir('templates'));

With these in place, I'm still not having much luck getting Smarty3 to compile/display OJS templates -- is yours working? Are you working from something close to the `master` branch, or on something older?

Vitaliy-1:
Hmm, never actually used a composer. Maybe once. If it is similar to Maven, may take a look.

OJS latest stable release (3.1.0-1) worked in my case. There are depreciated Smarty functions in several plugins, they can cause troubles (but should work with them...). I was doing nothing more than looking at the error log and replacing depreciated functions with the working analogs to minimize PHP error log. 
Also, I have extended SmartyBC class (not Smarty) for leaving the support for {php} tags. 

Ok, will check once more.

Vitaliy-1:
@asmecher ,
Have checked with master branch and update through composer (the require statement for Smarty was still needed). Just updated PKPTemplateManager as per pull request. The same result, templates compile but with a number of notices and endless spinner on the admin dashboard side.
What errors do you see in the error log? Do they refer to Smarty depreciated functions?

asmecher:
@Vitaliy-1, I found two things so far:

1. Lots of "undefined property" notices were showing for me. These can be suppressed by putting the following at the end of the constructor in `classes/template/PKPTemplateManager.inc.php`:

        $this->error_reporting = E_ALL & ~E_NOTICE;

2. The `assign` modifier (implemented in PKPTemplateManager in `smartyAssign`) isn't working; this appears to be something to do with variable scoping, which is in Smarty 3 but wasn't in Smarty 2. This results in pretty much all of the AJAX requests breaking, as `|assign:"..."` is used pretty much throughout to get the URL for the subrequest.

Vitaliy-1:
@asmecher 
If it is something with variable scoping, could we just point that the scope is global when we need it?
From their documentation `{assign var=foo value="bar" scope="global"}`

Vitaliy-1:
Ah, I see that you use your own assign modifier.
In such cases:
https://github.com/pkp/pkp-lib/blob/master/templates/header/usernav.tpl#L67-L76
I get an error that `url` is not defined:
`PHP Warning:  Smarty error: url parameter is missing from load_url_in_el in C:\\xampp\\htdocs\\ojs-3.1.0-smarty3\\lib\\pkp\\lib\\vendor\\smarty\\smarty\\libs\\SmartyBC.class.php on line 453, referer: http://emedjournal.com/index.php/psp/submissions`

So, you are sure that it is because of `smartyAssign` and not `smartyUrl`. In this case, can we rewrite (override) Smarty `assign` method?

Vitaliy-1:
@asmecher 
Here is Smarty 3 assign method:
```
public function assign($tpl_var, $value = null, $nocache = false)
    {
        if (is_array($tpl_var)) {
            foreach ($tpl_var as $_key => $_val) {
                $this->assign($_key, $_val, $nocache);
            }
        } else {
            if ($tpl_var != '') {
                if ($this->_objType === 2) {
                    /** @var  Smarty_Internal_Template $this */
                    $this->_assignInScope($tpl_var, $value, $nocache);
                } else {
                    $this->tpl_vars[ $tpl_var ] = new Smarty_Variable($value, $nocache);
                }
            }
        }
        return $this;
    }
```
This line after if dtatement is new: ` $this->_assignInScope($tpl_var, $value, $nocache);`
Maybe will take closer look at it in the future.

asmecher:
Unfortunately it looks like the `assign` modifier might simply need to be removed. Fixing it for Smarty3 would involve locating the currently active template object from within the modifier, and it's not currently available [short of a gnarly hack](https://www.smarty.net/forums/viewtopic.php?p=85910).

I've converted all the `assign` modifier cases into `{capture}` on my local install, but I'm still encountering other breakages. I'll try to look into it further.

asmecher:
@Vitaliy-1, I've posted my own branches and moved work/discussion over to https://github.com/pkp/pkp-lib/issues/3242 for tracking. I'd suggest we continue working over there. (PRs on private branches tend to disappear e.g. with forced-pushes, so it's harder to maintain ongoing discussion.)

asmecher:
Pretty close to ready for merging over on the linked issue -- closing this one!

Issue #3156, opened 2017-12-04
[OJS] Submission Requirements translation are not available


forgive38:
Hi,
I checked this with a fresh install of OJS 3.1 : I do install with fr_FR as default language and en_US as additional language. When you post a new submission, the Submission Requirements are only  available in fr_FR, even interface is in en_US.
Same behavior in Settings -> Workflow -> Submission.

Thank you

bozana:
Hi @forgive38,

Hmmm... I am not sure I understand it correctly. 
How exactly do you try to add a new submission and what problem exactly occurs? What have you chosen as a submission language?
What exactly is the problem in Settings -> Workflow -> Submission?
Thanks!

crosfield:
@bozana He means that by default there is no translation of submission requirements into English. I confirm this. In Russian/English interface it is the same. By default, the fields with English choices are empty.
@forgive38 It is necessary to make the translation into English manually in Settings -> Workflow -> Submission on your language first and then they will appear on the English interface.

bozana:
Ah, OK, I see... :-)
When a journal is installed only the default settings for the currently used UI locale are installed, s. https://github.com/pkp/pkp-lib/blob/master/classes/db/SettingsDAO.inc.php#L171. 
All siteSupportedLocales are also set to be supported (as UI locales) by the new journal, s. https://github.com/pkp/ojs/blob/master/controllers/grid/admin/journal/form/JournalSiteSettingsForm.inc.php#L124.
The journal settings (and thus also those default localized settings) can only be edited in/for a locale if that locale is enabled for forms (under Settings > Website > Languages). Thus, maybe to install those default localized settings when a new locale is enabled for forms? -- If a locale is enabled for UI but not for forms, the author or reader will be able to use that UI locale, but will see the submission check list only in the existing, primary locale, which is OK, because the the other locale is not enabled for forms so the check list cannot be edited and adapted.
I'll provide a patch...

bozana:
PRs:
pkp-lib master: https://github.com/pkp/pkp-lib/pull/3163
ojs master: https://github.com/pkp/ojs/pull/1750

bozana:
@NateWr, could you take a look at these PRs?

bozana:
Thanks! I will then merge -- everything green :-) -- and will immediately add a new OMP commit/PR that removes that key...

forgive38:
hi @bozana 

Thanks a lot for this PR.
[...]

PR #3207, opened 2017-12-20
pkp/pkp-lib#3156 reload default localized context settings at locale …

bozana:
…reload

s. ttps://github.com/pkp/pkp-lib/issues/3156


bozana:
ah wrong branch

Issue #3204, opened 2017-12-20
COUNTER bot list is outdated


ctgraham:
For both OJS 2.4.x and OJS 3.1.x, the Project COUNTER bot list is outdated.  Reported here:
https://forum.pkp.sfu.ca/t/bot-list-for-the-counter-metric/29018

Project COUNTER [references](https://www.projectcounter.org/appendices/850-2/) a [GitHub repo for this bot list](https://github.com/atmire/COUNTER-Robots).

Clarification is needed regarding [how this repo's text export will be maintained](https://github.com/atmire/COUNTER-Robots/issues/3), but as-is, it can be a drop-in replacement.

asmecher:
Merged both -- thanks, @ctgraham.

ctgraham:
A fresh checkout of lib/pkp with a `git submodule update --init --recursive` doesn't seem to populate the `plugins/generic/usageStats/lib/counterBots`

PR #3213, opened 2017-12-21
Fix submodule for COUNTER bots.

ctgraham:
Add submodule and test to ensure file is present and readable.  Resolves pkp/pkp-lib#3204.

Adding the test involves refactoring the formerly private method.


asmecher:
Merged -- thanks, @ctgraham!

Issue #3235, opened 2018-01-03
Creative Commons license buttons should be protocol relative


ctgraham:
In OCS, the Creative Commons license buttons are referenced in the locale files by http, raising https errors on page display.  Reported here:
https://forum.pkp.sfu.ca/t/ocs-google-chrome-warning-on-https-page/36592/3

asmecher:
Merged, thanks!

PR #3236, opened 2018-01-03
pkp/pkp-lib#3235: Update CC license images to protocol relative URLs

ctgraham:
Resolves pkp/pkp-lib#3235, with new licensebuttons.net URL.


asmecher:
Also ported to `ocs-stable-2_3_6`.

Issue #3056, opened 2017-11-14
Navigation Config vs Static Pages


pkel:
I'm testing OJS 3.1 checked out from git tag `ojs-3_1_0-0`. The navigation configuration at *Navigation Menus > Navigation Menu Items >  Add Item* under `/management/settings/website` provides the option *Custom Page*. One can configure a path `/navigationMenu/view/%PATH%` there and then link it to a menu menu button.

On the other hand, I have another tab called *Static Pages* where I can configure pages with path `/%PATH%` directly. There is no direct option to link this pages in a menu. One can use *Remote URL* on menu item creation, but this enforces full url specification and thereby breaks tests on localhost.

All this seems very confusing to me. It would be much clearer if the menu item generation wouldn't allow page creation. An option for menu item creation with internal/relative links would be helpful.

Further, the first option for page creation seems broken:
![screenshot from 2017-11-14 16-24-14](https://user-images.githubusercontent.com/11337857/32787976-48262c96-c958-11e7-9b92-ed2a3c24e212.png)

defstat:
@asmecher PR for porting StaticPages as NMIs

It runs on update, if the static_pages table exists. It ports every staticPage that its path does not conflicts with a Custom Menu Item with the same path. If it does conflict, the ported staticPage doesn't get removed from the static_pages table. 

PRs:
pkp-lib: #3270 
OJS: pkp/ojs#1790

asmecher:
@defstat, just a few comments on the code, and otherwise it looks good.

The [most recent round of UI/UX testing](https://github.com/pkp/pkp-lib/projects/6) turned up some confusion about where to find the content creation tools (be they static pages or custom navigation menus), and I think once static pages are gone that confusion will remain because users will often want to create static pages that aren't necessarily menu items. For example, users will want to create pages and link them from custom blocks. They won't think go look for content creation tools in the navigation menus area.

I think it's worth reviewing the navigation menus toolset to see whether it would be possible to make the language more general, so that users wanting to create content (but not necessarily menu items) will look there.

asmecher:
(See also https://github.com/pkp/pkp-lib/issues/3287)

asmecher:
@defstat, does this still need to be ported to OMP as well?

defstat:
@asmecher 
OMP PR

OMP: pkp/omp#495

asmecher:
Just one comment -- please go ahead and merge when ready!

asmecher:
[...]

PR #3270, opened 2018-01-13
[pkp-lib] #3056 Port StaticPages as NMIs

defstat:


classes/navigationMenu/NavigationMenuItemDAO.inc.php
```diff
@@ -517,6 +517,35 @@ function unCacheRelatedNavigationMenus($id) {
...
+	 */
+	function portStaticPage($staticPage) {
+		$path = $staticPage->getPath();
+		$contextId = $staticPage->getContextId();
+
+		$existingNMIWithPath = $this->getByPath($contextId, $path);
+
+		$retNavigationMenuItemId = null;
+
+		if (!isset($existingNMIWithPath)) {
```
asmecher: In the case there *is* a conflicting NMI, I'd suggest using `error_log` to log a warning. (See `classes/install/Upgrade.inc.php` for other examples of this.)

defstat: @asmecher I added the following message
`WARNING: The StaticPage "tetst" uses a path (test) that conflicts with an existing Custom Navigation Menu Item path. Skipping this StaticPage.`

PR #3280, opened 2018-01-17
Associate reviews to round for author

forgive38:
This is PR related to https://forum.pkp.sfu.ca/t/associate-editor-decision-with-review-round-in-authordashboard/33741/5

Thank you for comments

### DIFF
--- a/classes/log/SubmissionEmailLogDAO.inc.php
+++ b/classes/log/SubmissionEmailLogDAO.inc.php
@@ -40,6 +40,40 @@ class SubmissionEmailLogDAO extends EmailLogDAO {
 		return parent::_getByEventType(ASSOC_TYPE_SUBMISSION, $submissionId, $eventType, $userId);
 	}
 
+        /**
+	 * Get submission email log entries by submission ID and event type
+	 * @param $submissionId int
+	 * @param $eventType SUBMISSION_EMAIL_...
+	 * @param $userId int optional Return only emails sent to this user.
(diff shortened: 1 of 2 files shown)


classes/log/SubmissionEmailLogDAO.inc.php
```diff
@@ -40,6 +40,39 @@ function getByEventType($submissionId, $eventType, $userId = null) {
...
+	 * @return DAOResultFactory
+	 */
+	function getByEventTypeAndDate($submissionId, $eventType, $userId = null, $date) {
+		$params = array(
+				(int) ASSOC_TYPE_SUBMISSION,
+				(int) $submissionId,
+				(int) $eventType,
+				$date
+			);
+		if ($userId) $params[] = $userId;
```
asmecher: Add a `(int)` cast here too -- it helps keep PostgreSQL happy when invalid data comes in from userland.

asmecher:
Hmm, matching on the date is going to be a little imprecise (if we cross a second boundary between recording the decision and sending the email, then the email won't match.

I have two ideas for how to resolve this:
- Add a formal way for the email to refer back to the round. The best way to do this OTOH would be to use the `event_log_settings` table. (Other options that aren't promising IMO: Change the `assoc_type`, or new a new column.)
- Present the emails in a chronological list with headings indicating when each new round was created.

I think that last option would be the easiest, and would play well with existing data.

PR #3284, opened 2018-01-18
[pkp-lib] #2899 Add support for custom NMI's paths to be displayed under context/path url

defstat:


classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -370,6 +368,8 @@ public function loadMenuTree(&$navigationMenu) {
 	 * @return array Hierarchical array of menu items
 	 */
 	public function getMenuTree(&$navigationMenu) {
+		//\HookRegistry::register('LoadHandler', array($this, '_callbackHandleCustomNavigationMenuItems'));
```
asmecher: Dead code -- I'd suggest removing this.

classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -516,4 +516,41 @@ private function _hasNMTreeNMIAssignmentWithChildOfNMIType($navigationMenu, $nav
...
+	public function _callbackHandleCustomNavigationMenuItems($hookName, $args) {
+		$request = \Application::getRequest();
+
+		$page =& $args[0];
+		$op =& $args[1];
+
+		// Construct a path to look for
+		$path = $page;
+		if ($op !== 'index') $path .= "/$op";
+		if ($ops = $request->getRequestedArgs()) $path .= '/' . implode('/', $ops);
```
asmecher: I'd suggest changing the terminology here -- "operation" / "op" is not the same thing as the requested arguments; use `$args` or `$arguments` for consistency.

classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -370,6 +368,8 @@ public function loadMenuTree(&$navigationMenu) {
 	 * @return array Hierarchical array of menu items
```
asmecher: This function doesn't appear to return anything -- either fix the code self-documentation, or better yet, use a normal `return` convention instead of the reference (unless there's a reason to do it this way).

classes/services/PKPNavigationMenuService.inc.php
```diff
@@ -516,4 +516,41 @@ private function _hasNMTreeNMIAssignmentWithChildOfNMIType($navigationMenu, $nav
 
 		return false;
 	}
+
+	public function _callbackHandleCustomNavigationMenuItems($hookName, $args) {
```
asmecher: Needs a function header comment.

defstat: Fixed

classes/services/PKPNavigationMenuService.inc.php
defstat: Fixed

classes/services/PKPNavigationMenuService.inc.php
defstat: Fixed

classes/services/PKPNavigationMenuService.inc.php
defstat: Fixed

Issue #2988, opened 2017-10-30
Installation page depends on navigation menu tables, which may not exist


crism:
On upgrading OJS 3.0.2-0 to 3.1.0-0, I hit the `/install` URL after upgrading all the software, but before upgrading the database. I got this in my error log:
```
[:error] ojs2: DB Error: ERROR:  relation "navigation_menus" does not exist\nLINE 1: SELECT * FROM navigation_menus WHERE context_id =  $1\n                      ^
```
and in the browser, the same text as a big, bold error. The install page for version _x_ shouldn’t depend on database features new to _x_ (or, really, any database features at all…).

defstat:
@asmecher As far as I can see there is a `SESSION_DISABLE_INIT` parameter that is assigned to `1` or `true` if the user is in the "context" of an installation page. But that variable is not assigned if the user decides to hit the "install" page.
Nevertheless, if the user follows the upgrade procedure described at the end of this document https://pkp.sfu.ca/ojs/UPGRADE, the install page gets no DB error. 

Is there another mechanism that checks whether the called page is an "install" page (a "tool" page in general), or if its an frontend page (and not administrative)?

asmecher:
@crism, just checking -- is `Installed` set to `Off` in `config.inc.php`?

crism:
@asmecher, yes, `Installed = Off`. When that’s the case, any visit to the site redirects to the install page. Under 3.1, until the database has been upgraded, that gives an error when it tries to load the nav menus.

asmecher:
Hmm, I can't reproduce this either. @crism, can you turn on `show_stacktrace` in `config.inc.php`?

crism:
@asmecher, it’s going to take a little while. I noticed this in our production system mid-upgrade, and certainly didn’t leave it that way. I’ll need to make a fresh 3.0.2 install somewhere…

asmecher:
See also https://forum.pkp.sfu.ca/t/unable-to-upgrade-to-3-1-0/34807.

asmecher:
(Cleared milestone, needs feedback)

defstat:
@asmecher It seems that this can happen when the installation page is being accessed through a url that includes the context. If this is the case, and in the config.inc.php the `installed = Off`, OJS is accessing https://github.com/pkp/pkp-lib/blob/master/classes/template/PKPTemplateManager.inc.php#L284 with a context defined, and it is trying to access the NavigationMenuDAO which may not exist. 

[...]

PR #3296, opened 2018-01-19
[pkp/pkp-lib] #2988  NavigationMenus not displayed if configuration installed=Off

defstat:


asmecher:
Where are the navigation menus loaded if this code is removed? (I don't see equivalent code being added somewhere else.)

defstat:
@asmecher The NMs are being loaded from within the templates like https://github.com/pkp/pkp-lib/blob/master/templates/frontend/components/header.tpl#L78.

There the https://github.com/pkp/pkp-lib/blob/master/classes/template/PKPTemplateManager.inc.php#L1559 is being called. Within there is the call to load the NM tree (for the requested area): https://github.com/pkp/pkp-lib/blob/master/classes/template/PKPTemplateManager.inc.php#L1595

asmecher:
Ah, thanks! That's a much better place to be loading them.

Issue #3318, opened 2018-01-25
Check minimum requirements before allowing to upgrade


ajnyga:
Hi,

Although I have been following the development of OJS fairly closely, I totally missed the fact that the minimum requirements were raised to PHP5.6. I have also seen a lot situations in the forum where people using older PHP versions have already upgraded and are now running into trouble. In some cases it is perhaps not that straightforward for a journal to update the server PHP.

So maybe the upgrade script could start with a PHP version check?

asmecher:
I didn't expect this was going to be a problem as PHP5.5 will have been unsupported for 2 full years this July -- but a version check on upgrade wouldn't hurt.

ajnyga:
One option would be to add the required versions as attributes to the upgrade xml file together with the version number: https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade.xml#L14
Or, you could add a call to a `minimumRequirements ` code that would executed right in the beginning. 

Comparing would probably happen with: http://php.net/manual/en/function.version-compare.php

asmecher:
Agreed -- easily enough done.

PR #3321, opened 2018-01-26
update ru_RU locale after #3318

ppv1979:


asmecher:
Fast as always! Thanks, @ppv1979.

PR #3388, opened 2018-02-15
Update ru_RU locale after #3364 & #3310, fix en_US typo after #3364

ppv1979:


asmecher:
Thanks as always, @ppv1979!

PR #3402, opened 2018-02-19
[pkp-lib] #2180 Prepared emails for reviewers display name variable

defstat:


forgive38:
Hi,
Thank for that, I tried it.
i have some issues, for example, when you select another email template and you reselect the precedent one, you lose some variables : they became {$variable}
I don't know if it's impact the email preparation...

PR #3414, opened 2018-02-22
Update ru_RU locale after #3399 and #3411

ppv1979:


asmecher:
Lightning-fast as always. Thanks, @ppv1979!

Issue #520, opened 2015-05-11
Improve Submission Library utility


asmecher:
See:
- http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8925 ("Enable selection of docs from Press Library when sending email to author at Accept Submission")
- http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8441 ("Grant optional public access to document library")

defstat:
@asmecher 

PRs
pkp-lib: #3424
OJS: pkp/ojs#1851
OMP: pkp/omp#505

defstat:
@asmecher @jmacgreg 
The PR adds the two improvements regarding the Library Documents. 
1. Upon an Editor Decission, a grid with the Library Documents has been added, under the email template, so that the Editor can select any file from there, so that it can be attached to the email.
2. Optionally, a Library Document can be indicated as "Public Access"ed, and a URL is provided so that any user (both logged-in and not) can download it from that URL. The URL is something like `[context]/publicLibrary/downloadLibraryFile/[fileId]`

defstat:
@asmecher I covered the review (with some notes)

asmecher:
@defstat, I added a few new comments. The necessary tweaks are minor, so please feel free to go ahead and merge once you've looked them over. Thanks!

defstat:
@asmecher Tests pasted. PRs merged.

asmecher:
Thanks, @defstat!

defstat:
@bozana 
Regarding your comment 

PRs:
pkp-lib: #3261 
OJS: pkp/ojs#1878

fgnievinski:
I wished PDFs would display in the browser instead of being downloaded.

(Should I open a new feature request or is this comment sufficient?)

Thanks.

fgnievinski:
Also, it'd be nice to be able to replace a file; currently it seems the only workaround is removing the old one and creating a new one, which unfortunately breaks links.

PR #3424, opened 2018-02-26
[pkp-lib] #520 Improve Submission Library Utility

defstat:


pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+		// Check the last string of the request.
+		// The code does not check the url format to be /publicLibrary/downloadLibraryFile/id
+		$publicFileId = $parts[count($parts) - 1];
+
+		$libraryFile = $libraryFileDao->getById($publicFileId);
+		if ($libraryFile && $libraryFile->getPublicAccess()) {
+			$filePath = $libraryFileManager->getBasePath() .  $libraryFile->getOriginalFileName();
+			$libraryFileManager->downloadFile($filePath);
+		} else {
+			fatalError('Unauthorized access. This file is not public.');
```
asmecher: I think a HTTP 403 would be more appropriate here.

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * @defgroup controllers_api_file File API controller
+ */
+
+/**
+ * @file controllers/api/file/FileApiHandler.inc.php
```
asmecher: The filename etc. needs updating here (if this stays unchanged).

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+
+// Import the base handler.
+import('classes.handler.Handler');
+
+class PublicLibraryFileHandler extends Handler {
+	/**
+	 * Constructor.
+	 */
+	function __construct() {
+		parent::__construct();
```
asmecher: Constructors like this can be removed.

templates/controllers/modals/editorDecision/form/promoteForm.tpl
```diff
@@ -78,6 +78,11 @@
 				{load_url_in_div id="reviewAttachmentsGridContainer" url=$reviewAttachmentsGridUrl}
 			</div>
 		{/if}
+
+		<div id="libraryFileAttachments">
```
asmecher: It might be worth making this div collapsible, as I suspect this will take up a lot of real estate relative to its frequency of use.

asmecher:
Thanks, @defstat! Just a few things to work on.

classes/template/PKPTemplateManager.inc.php
```diff
@@ -285,6 +285,9 @@ function initialize() {
 			$nmService = ServicesContainer::instance()->get('navigationMenu');
 
 			\HookRegistry::register('LoadHandler', array($nmService, '_callbackHandleCustomNavigationMenuItems'));
+
+			// Register Public library files handler
+			\HookRegistry::register('LoadHandler', array($this, 'setPublicLibraryFilesHandler'));
```
asmecher: I'm confused about this -- is it necessary?

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+ * @class PublicLibraryFileHandler
+ * @ingroup pages_publicLibraryFiles
+ *
+ * @brief Class defining a handler for public library file access
+ */
+
+// Import the base handler.
+import('classes.handler.Handler');
+
+class PublicLibraryFileHandler extends Handler {
```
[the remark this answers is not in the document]
asmecher: One option would be to move the authenticated download out of the FileApiHandler, into the new handler -- that way they would both be together.

classes/template/PKPTemplateManager.inc.php
defstat: @asmecher It may be my incomplete understanding on how the handlers are being registered on OJS, but shouldn't I register the handler like that in order for the new URL pattern (/publicLibrary/downloadLibraryFile/id) to became available?

templates/controllers/modals/editorDecision/form/promoteForm.tpl
defstat: Done

templates/controllers/grid/settings/library/form/newFileForm.tpl
```diff
@@ -47,6 +47,15 @@
 		{/fbvFormSection}
 	{/fbvFormArea}
 
+	{fbvFormSection list="true" translate=false}
+		{translate|assign:"enablePublicAccess" key="common.publicAccess"}
+		{fbvElement type="checkbox" id="publicAccess" value="1" checked=false label=$enablePublicAccess translate=false}
+		<p>
+			{url|assign:"downloadUrl" router=$smarty.const.ROUTE_PAGE page="publicLibrary" op="downloadLibraryFile" path="id"}
+			{translate key="settings.libraryFiles.public.viewInstructions" pagesPath=$downloadUrl}
```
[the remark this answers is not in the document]
defstat: Done

templates/controllers/grid/settings/library/form/editFileForm.tpl
```diff
@@ -51,6 +51,15 @@
 		{/fbvFormSection}
 	{/fbvFormArea}
 
+	{fbvFormSection list="true" translate=false}
+		{translate|assign:"enablePublicAccess" key="common.publicAccess"}
+		{fbvElement type="checkbox" id="publicAccess" value="1" checked=$publicAccess label=$enablePublicAccess translate=false}
+		<p>
+			{url|assign:"downloadUrl" router=$smarty.const.ROUTE_PAGE page="publicLibrary" op="downloadLibraryFile" path=$libraryFile->getId()}
+			{translate key="settings.libraryFiles.public.viewInstructions" pagesPath=$downloadUrl}
```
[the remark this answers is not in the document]
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+		if ($libraryFile && $libraryFile->getPublicAccess()) {
+			$filePath = $libraryFileManager->getBasePath() .  $libraryFile->getOriginalFileName();
+			$libraryFileManager->downloadFile($filePath);
+		} else {
+			fatalError('Unauthorized access. This file is not public.');
+		}
+	}
+}
+
+?>
```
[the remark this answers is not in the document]
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+	 * Constructor.
+	 */
+	function __construct() {
+		parent::__construct();
+	}
+
+	//
+	// Implement methods from PKPHandler
+	//
+	function authorize($request, &$args, $roleAssignments) {
```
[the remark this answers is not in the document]
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @param $request Request
+	 */
+	function downloadLibraryFile($args, $request) {
+		import('classes.file.LibraryFileManager');
+		$context = $request->getContext();
+		$libraryFileManager = new LibraryFileManager($context->getId());
+		$libraryFileDao = DAORegistry::getDAO('LibraryFileDAO');
+
+		$requestPath = $request->getRequestPath();
+		$parts = explode("/", $requestPath);
```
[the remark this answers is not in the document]
defstat: Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
```diff
@@ -0,0 +1,69 @@
...
+
+		// Check the last string of the request.
+		// The code does not check the url format to be /publicLibrary/downloadLibraryFile/id
+		$publicFileId = $parts[count($parts) - 1];
+
+		$libraryFile = $libraryFileDao->getById($publicFileId);
```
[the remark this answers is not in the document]
defstat: Right! Done

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
defstat: I think I fixed it - do we have a "template function" for that? (like TemplateManager->Handle403Redirect() or something)

controllers/grid/settings/library/form/EditLibraryFileForm.inc.php
```diff
@@ -44,7 +44,8 @@ function __construct($contextId, $fileId) {
 	function initData() {
 		$this->_data = array(
 			'libraryFileName' => $this->libraryFile->getName(null), // Localized
-			'libraryFile' => $this->libraryFile // For read-only info
+			'libraryFile' => $this->libraryFile, // For read-only info
+			'publicAccess' => $this->libraryFile->getPublicAccess() ? true : false
```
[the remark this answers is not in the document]
defstat: Changed it anyway.

classes/template/PKPTemplateManager.inc.php
asmecher: No, simply putting the handler class in the right spot and adding an `index.php` wrapper for it should be enough. You should be able to remove this.

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
asmecher: No; for now, what you've got is fine.

defstat: OK, thanks.

classes/template/PKPTemplateManager.inc.php
defstat: In that case I was missing something there...

pages/publicLibraryFiles/PublicLibraryFileHandler.inc.php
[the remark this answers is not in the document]
defstat: @asmecher I made a change for which I am not entirely sure it covered your recommendation. Both download functions are in one Handler, but I kept a function in the api that calls the new "common" handler for download. (so that it can be used here `https://github.com/pkp/pkp-lib/blob/master/controllers/api/file/linkAction/DownloadLibraryFileLinkAction.inc.php#L34` )

pages/libraryFiles/LibraryFileHandler.inc.php
```diff
@@ -0,0 +1,96 @@
...
+	 * Download a library file.
+	 * @param $args array
+	 * @param $request Request
+	 */
+	function downloadLibraryFile($args, $request) {
+		import('classes.file.LibraryFileManager');
+		$context = $request->getContext();
+		$libraryFileManager = new LibraryFileManager($context->getId());
+		$libraryFileDao = DAORegistry::getDAO('LibraryFileDAO');
+		$libraryFile = $libraryFileDao->getById($request->getUserVar('libraryFileId'));
```
asmecher: This fetch is also missing a check for the context ID.

Generally we have a pattern to follow for this: getter functions like `getById` should have an optional second parameter for the contextId. When specified, it adds an `AND context_id = ?` to the SQL. That way supplying the second parameter becomes an automatic check for the right context ID. I'd suggest adding that option to the DAO code then using it wherever we're calling `LibraryFileDAO::getById`.

xml/schema/libraryFiles.xml
```diff
@@ -49,6 +49,10 @@
 		<field name="submission_id" type="I8">
 			<NOTNULL/>
 		</field>
+		<field name="public_access" type="I1">
+				<NOTNULL/>
```
asmecher: Off the top of my head, ADODB doesn't always handle new columns with `NOT NULL` constraints, even though there's a `DEFAULT VALUE` here. I'd suggest removing the `NOT NULL`. It'll work just fine because `NULL` values will be the default, and they're equivalent to boolean `false`.

pages/libraryFiles/LibraryFileHandler.inc.php
defstat: @asmecher I made the change. I found that there was a documented parameter inconsistency, and I changed that too. (supposing that the other documented parameters don't serve any purpose for the current code version at least)

By making a quick search I think there are more such inconsistencies between function parameters and documented parameters, but I am not sure if that is on purpose.

For example 
https://github.com/pkp/pkp-lib/blob/master/classes/context/PKPSectionDAO.inc.php#L32
or 
https://github.com/pkp/pkp-lib/blob/master/classes/controlledVocab/ControlledVocabEntryDAO.inc.php#L35

asmecher: @defstat, thanks -- these self-documentation inconsistencies aren't intentional; feel free to quickly correct them when you see them.

PR #3452, opened 2018-03-06
Translation update for da_DK locale

tgc:
Updates to the danish translation from @neffe
This also takes care of the obsolete keys.
I requested merging with ojs-stable since this goes on top of the last update which was also merged to the stable branch.


asmecher:
Merged and cherry-picked forward to `master`. There was a merge conflict in cherry-picking, too -- if this happens too often, I'd suggest we start working with pulls directly to the `master` branch instead, but we haven't reached that threshold yet.

Thanks as always!

PR #3466, opened 2018-03-09
Prevent ErrorMsg to spread to other text inputs

MarHerUMR:
The `FBV_urlValidationErrorMessage` has to be cleared from the templateManager. Otherwise every following text input (without urlValidationErrorMsg) get's this error message applied.


asmecher:
Good catch -- thanks, @MarHerUMR!

Issue #3302, opened 2018-01-21
File upload gets stuck with certain language selections in multilingual journals (OJS 3.1.0.1)


ajnyga:
Hi,

This is related to: https://github.com/pkp/pkp-lib/issues/2114.

I first thought this was a bug, but it is actually related to journal language settings

Scenario:
- Journal primary language is English (`UI`, `Submissions`, `Forms` selected)
- Secondary language Finnish is selected (`UI`, `Submissions`)
- User tries to submit a new manuscript, which is written in Finnish.
- In the first stage user selects Finnish as the submission language and moves on to stage 2 to upload files.
- The firs phase (choosing the component and uploading the file) goes ok, but in the second phase of the file upload the **user gets stuck because the submission language is Finnish, but the system can not fill in the Finnish locale for the file name, because the Finnish language forms are not enabled in the settings**.

Suggested solution: in the language settings page, if `Submissions` is selected, the `Forms` selection is automatically enabled as well. Otherwise you will have malformed metadata.

bozana:
PRs:
https://github.com/pkp/pkp-lib/pull/3486
https://github.com/pkp/ojs/pull/1877
https://github.com/pkp/omp/pull/513

bozana:
@asmecher, could you review the PRs?

bozana:
Maybe a note: because there are lots of forms depending on the submission locale -- differently to the OJS 2.4.x -- this is the better way to solve it, and most obvious for the user, I believe: that the form locale has to be enabled when the submission locale is enabled, and that the submission locale is disabled when the form locale is disabled.
The existence of both locales is however justified for the case when only the form locale is enabled.

asmecher:
This solution is fine -- there's a bigger aspect needing clean-up, which is that I think we're not consistent about the use of primary locales with respect to submission files. I think sometimes we're assuming it's the submission locale, other times the jouranal's locale. But that's outside of this fix -- I just had a few comments; please go ahead and merge when you've gone through them. Thanks!

PR #3486, opened 2018-03-16
pkp/pkp-lib#3302 connect submission and form locales settings

bozana:
s. https://github.com/pkp/pkp-lib/issues/3302

I also changed the order of the columns in the language grid, so that form locale is before the submission locale...

### DIFF
--- a/controllers/grid/languages/LanguageGridCellProvider.inc.php
+++ b/controllers/grid/languages/LanguageGridCellProvider.inc.php
@@ -44,11 +44,11 @@ class LanguageGridCellProvider extends GridCellProvider {
 			case 'uiLocale';
 				return array('selected' => $element['supportedLocales'],
 					'disabled' => !$element['supported']);
-			case 'submissionLocale';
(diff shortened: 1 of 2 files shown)


controllers/grid/languages/LanguageGridCellProvider.inc.php
```diff
@@ -44,12 +44,12 @@ function getTemplateVarsFromRowColumn($row, $column) {
 			case 'uiLocale';
 				return array('selected' => $element['supportedLocales'],
 					'disabled' => !$element['supported']);
-			case 'submissionLocale';
```
asmecher: I don't think any of the changes to this file have any effect.

asmecher: (Maybe it's just to keep the code aligned with the order of presentation?)

bozana: yes, just for that :-)

PR #3494, opened 2018-03-19
[pkp-lib] #3462 Refine password element for user import (and possibly export)

defstat:


classes/security/Validation.inc.php
```diff
@@ -430,6 +430,32 @@ static function canAdminister($administeredUserId, $administratorUserId) {
...
 		return true;
 	}
+
+	/**
+	 * Validation process for imported passwords
+	 * @param $userToImport User ByRef. The user that is being imported. 
+	 * @param $encryption string null, sha1, md5 (or any other encryption algorithm defined)
+	 * @return string if a new password is generated, the function returns it.
+	 */
+	function importUserPasswordValidation(&$userToImport, $encryption) {
```
asmecher: The reference to `$userToImport` isn't needed.

classes/security/Validation.inc.php
```diff
@@ -430,6 +430,32 @@ static function canAdminister($administeredUserId, $administratorUserId) {
 		// There were no conflicting roles. Permit administration.
 		return true;
 	}
+
+	/**
+	 * Validation process for imported passwords
+	 * @param $userToImport User ByRef. The user that is being imported. 
+	 * @param $encryption string null, sha1, md5 (or any other encryption algorithm defined)
```
asmecher: This doesn't really use the `$encryption` setting, and there's no indication to the user if the given algorithm wasn't right for the journal -- it'll set a random password instead. I'd prefer if the error condition for a configuration mismatch was shown to the user.

plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php
```diff
@@ -105,24 +108,33 @@ function parseUser($node) {
...
 					$user->setPassword($password->textContent);
 				} else {
 					fatalError("User has no password.  Check your import XML format.");
 				}
+
 				break;
 		}
 
+		// Password Import Validation
+		$password = Validation::importUserPasswordValidation($user, $encryption);
```
asmecher: This function is only used here, and eventually I'd like to remove the `Validation` class (or at least keep it as small as possible). I'd suggest having the contents of that function live in this class instead.

defstat: Done.

classes/security/Validation.inc.php
defstat: Done.

classes/security/Validation.inc.php
defstat: The `encryption` parameter checks if an encryption method is being defined. If it does not, then the code manages the `password` as a plain password. If the encryption has a value, then the code checks the `password_needs_rehash` function. Not sure if there is a way to check if the has is sha1 or md5 (is it safe to check the length?)

The code has been changed a little so that the filtering function can return the errors produced from the filtering. (the `Filter` class has an _error array, but I can't see how the errors are being returned to the plugin caller)

plugins/importexport/users/PKPUserImportExportPlugin.inc.php
```diff
@@ -213,7 +216,8 @@ function importUsers($importXml, $context, $user) {
 		$importFilter = array_shift($userImportFilters);
 		$importFilter->setDeployment(new PKPUserImportExportDeployment($context, $user));
 
-		return $importFilter->execute($importXml);
+		$importFilter->execute($importXml);
+		return $importFilter;
```
asmecher: This change would require a change to the `@return` self-documentation. But rather than what you've done here, I'd suggest adding the filter as a reference parameter to the `importUsers` function, and leaving the return convention here as it used to be (which makes more sense to me).

PR #3498, opened 2018-03-20
Update ru_RU locale after #3424, #3463

ppv1979:


asmecher:
Thanks as always, @ppv1979!

PR #3538, opened 2018-03-28
Final update for sv_SE and 3.1.1

martinpub:
Hi @asmecher, here are the last changes also for the sv_SE locale. Still some remaining plugin translations which I will attend to, but those will have to be separate from the release, and maybe be cherry-picked to 3.1.1.


asmecher:
Thank you, @martinpub!

PR #3550, opened 2018-04-04
Move submission status check outside if clause

ajnyga:
@bozana @asmecher 

See https://forum.pkp.sfu.ca/t/automated-reminders-being-sent-for-archived-articles/39998/12 for details. Would be good to have in 3.1.1. because this seems to be a clear bug with an easy fix...


bozana:
Yes, this makes sense to me, but I will leave it to @asmecher to merge it when he thinks is good... i.e. if possible for 3.1.1...
Thanks @ajnyga!!!

asmecher:
Good catch, @ajnyga! These off-by-an-indentation problems can be hard to spot.

PR #3553, opened 2018-04-05
Move new reviews are in notification code outside default form else clause

ajnyga:
At the moment the notification code for new reviews only works for journals that use the default form, because the code is inside the else clause here: https://github.com/pkp/pkp-lib/blob/ojs-3_1_0-1/classes/submission/reviewer/form/ReviewerReviewStep3Form.inc.php#L199-L212

This change moves the code outside the else clause so that the journals using the custom forms also get these notifications.

see https://github.com/pkp/pkp-lib/issues/3423#issuecomment-378945708

### DIFF
--- a/classes/submission/reviewer/form/ReviewerReviewStep3Form.inc.php
+++ b/classes/submission/reviewer/form/ReviewerReviewStep3Form.inc.php
@@ -191,30 +191,32 @@ class ReviewerReviewStep3Form extends ReviewerReviewForm {
 			}
 			unset($comment);
 
-			$submissionDao = Application::getSubmissionDAO();
-			$submission = $submissionDao->getById($reviewAssignment->getSubmissionId());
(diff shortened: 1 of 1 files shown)


ajnyga:
@bozana would you have time to check this? I would fix this in our server because the journals using custom forms are now not getting the email notifications. Just want to have a second opinion that the fix is correct...

ajnyga:
for background see https://github.com/pkp/pkp-lib/issues/3423#issuecomment-378945708

bozana:
@ajnyga, yes, sorry for the delay -- last week I could not work much, but now...

ajnyga:
no problem @bozana sorry for pushing you!

ajnyga:
great, I will change this to our server and let the editor know it is fixed (and approved by PKP :D)

Issue #3407, opened 2018-02-20
[OJS] doubt about a translation


forgive38:
Hi all,
in lib/pkp/locale/en_US/common.xml, there is a message which is :
"<message key="notification.email">Do not send me an email for these types of notifications.</message>"
this message is translated in most of other languages as "Send me an email ..."

What is the good sens for this message ?
Thank you

mtub:
The English text is correct, the translations hadn't been updated. 

(The English text had been changed to its opposite meaning some time ago. This will be avoided in the future.)

PR #3584, opened 2018-04-11
Small correction on fr_CA

pnault:
Replacing a sentence that was doing exactly the contrary of what it
means according to the english version.


asmecher:
@pnault, I think this has already been corrected -- see https://github.com/pkp/pkp-lib/issues/3407 for details.

pnault:
@asmecher you are right. I guess my local instance is not up to date...

PR #3688, opened 2018-05-11
fix php notices when $args array is empty

paulo-graca:


asmecher:
Thanks, @paulo-graca, I've seen this warning pop up occasionally too.

PR #3689, opened 2018-05-11
verify if $this->settingsForm is defined or not null

paulo-graca:
fix PHP Fatal error:  Call to a member function validate() on null in /ojs/ojs-3.1.1/lib/pkp/controllers/tab/settings/paymentMethod/form/PaymentMethodForm.inc.php on line 110, referer: http://example.com/index.php/sarc-test/management/settings/distribution

### DIFF
--- a/controllers/tab/settings/paymentMethod/form/PaymentMethodForm.inc.php
+++ b/controllers/tab/settings/paymentMethod/form/PaymentMethodForm.inc.php
@@ -108,7 +108,7 @@ class PaymentMethodForm extends ContextSettingsForm {
 	 */
 	function validate() {
 		if (!$this->settingsForm->validate()) {
-			foreach ($this->settingsForm->getErrorsArray() as $field => $message) {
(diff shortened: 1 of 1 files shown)


asmecher:
@paulo-graca, are you working with a custom payment plugin or something similar? I'm not sure how this error would arise.

paulo-graca:
Hi!
No, we don't have payments. We are upgrading to 3.1.1.0. Our system was firstly installed based on 2.3.6.0 version. Perhaps there is something in our database that can trigger this error.

asmecher:
@paulo-graca, is it possible that you've mixed old OJS 2.x code in with OJS 3.x code, e.g. by unpacking the `.tar.gz` file for OJS 3.x over top of your OJS 2.x codebase?

paulo-graca:
We are currently working with 3.2.1-2 and we don't need the patch anymore. I will close this PR.

PR #3703, opened 2018-05-16
de_DE locate updates

mtub:
updated the de_DE locale according to the latest changes to the en_US locale


asmecher:
Thanks, @mtub! Also cherry-picked to `omp-stable-3_1_1` and `ojs-stable-3_1_1`, as I think the changes should apply there too.

PR #3730, opened 2018-05-23
de_DE locale update

mtub:
completing a file-by-file comparison of de_DE and en_US, nothing major happening, mainly introducing wiki links in the pkp-lib files' headers


bozana:
Thanks @mtub!

PR #3748, opened 2018-05-28
bad translation in context

t4x0n:
These words help to make clear what occurs. Before this was confused between Register and Login (login is Start a Session, not Register) in Spanish is similar but we have not a word like login, just an anglicism "loguearse" but is not common in all countries


asmecher:
Makes sense, though my Spanish is lumpy. Thanks, @t4x0n!

Issue #3812, opened 2018-06-19
possibility for plugins to add user and author metadata


bozana:
Currently the user and author form is implemented so that it is not possible to correctly add/influence the user/author metadata from a plugin.
Generally: 
-- the form construct, initData, readInputData, and execute function should use `parent::...` call at the appropriate place, in order for a plugin to be able to hook into
-- the object that is saved in the form executable function should either be the class attribute, or be provided to the parent::execute i.e. form::execute function, in order for a plugin to be able to access and change it.

bozana:
PRs:
pkp-lib ojs-stable-3_1_1: https://github.com/pkp/pkp-lib/pull/3813
ojs ojs-stable-3_1_1: https://github.com/pkp/ojs/pull/2019 (only submodule update)

pkp-lib omp-stable-3_1_1: 
omp omp-stable-3_1_1:  (only submodule update)

pkp-lib master:
ojs master: (only submodule update)
omp master: (only submodule update)

bozana:
@asmecher, could you take a look at the PR above? -- If OK, I will cherry-pick to other branches. If you think it should not go to stable branch, I will only provide it for master.
THANKS!!!

isgrim:
@bozana 
Thanks for implementing this. This will help alot with orcidProfile and other Plugins.

bozana:
@asmecher, may I merge this when the tests pass?

asmecher:
> I have a feeling that tests stopped totally... :-O

It looks to me like they're running now -- need me to take a look?

bozana:
Thanks a lot @asmecher! I merged the stable PRs and considered your comment in master. When the tests pass I will merge...

PR #3813, opened 2018-06-19
pkp/pkp-lib#3812 possibility for plugins to add user and author metadata

bozana:
s. https://github.com/pkp/pkp-lib/issues/3812


controllers/grid/users/author/form/AuthorForm.inc.php
```diff
@@ -193,7 +193,8 @@ function execute() {
 		$author = $this->getAuthor();
 		if (!$author) {
 			// this is a new submission contributor
-			$author = new Author();
+			$this->_author = new Author();
```
asmecher: This should use the `DAO::newDataObject` pattern instead of `new`.

PR #3829, opened 2018-06-26
Missing important chain (separator)

marcbria:
Reported here:
https://forum.pkp.sfu.ca/t/assign-evaluation-characters-appear-ojs-3-x/45721

@mtub this merge is important. Otherwise ojs will be defaced in Spanish.

### DIFF
--- a/locale/es_ES/common.xml
+++ b/locale/es_ES/common.xml
@@ -144,6 +144,7 @@
 	<message key="common.language">Idioma</message>
 	<message key="common.languages">Idiomas</message>
 	<message key="common.less">Menos</message>
+	<message key="common.commaListSeparator">, </message>
 	<message key="common.manage">Administrar</message>
 	<message key="common.mailingAddress">Dirección postal</message>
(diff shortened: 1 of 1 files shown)


bozana:
i allowed myself to merge it...
thanks @marcbria!

bozana:
@asmecher and @mtub, I am wondering if it would be better to add those two keys in every locale, then, if different in a language, it can be adjusted -- because earlier they were however hard coded. What do you think?

marcbria:
I'm wondering why we missed this chain. :-/
If it was introduced between translations, Bozana's proposal is the most reasonable.
Do we have more chains like this?

OT: @mtub in our "wish list" for the perfect translation tool, we need to add a feature to help us tracking the new and the removable chains between versions.

marcbria:
> @marcbria, may I maybe also add "common.semicolonListSeparator" as it is in en_US?

Yes please. Thanks Bozana.

asmecher:
@bozana, yes, your proposal makes sense to me.

PR #3850, opened 2018-06-29
Identify files by id and date

pkel:
Discussion: #3791

### DIFF
--- a/controllers/grid/files/FileNameGridColumn.inc.php
+++ b/controllers/grid/files/FileNameGridColumn.inc.php
@@ -58,7 +58,8 @@ class FileNameGridColumn extends GridColumn {
 		$submissionFileData = $row->getData();
 		$submissionFile = $submissionFileData['submissionFile'];
 		assert(is_a($submissionFile, 'SubmissionFile'));
-		$id = $submissionFile->getFileId() . '-' . $submissionFile->getRevision();
+		$timestamp = date('Y-m-d', strtotime($submissionFile->getDateUploaded()));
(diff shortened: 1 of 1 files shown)


NateWr:
Thanks @pkel! As I mentioned in the discussion on the issue, I'm in favor of adding the date to the file listing. I think it's a little bit too much to put into the label. Instead, let's add a column row to the file grid for the date uploaded. Do you need some guidance on how to go about that?

Also, instead of forcing `Y-m-d`, let's use the short date format set in the config file. You can do that with:

```
date(Config::getVar('general', 'date_format_short'), strtotime($x));
```

pkel:
Some guidance would indeed be appreciated.

NateWr:
No problem. So, in this PR, you've edited `FileNameGridColumn.inc.php`. You can see how this column is added to `SubmissionFilesGridHandler` here:

https://github.com/pkp/pkp-lib/blob/master/controllers/grid/files/SubmissionFilesGridHandler.inc.php#L148-L149

So, you'll need to create a new class for `FileDateGridColumn`, that is very similar to `FileNameGridColumn` except it returns the date. Then add that column to `SubmissionFilesGridHandler`.

retostauffer:
Had some time to look into this, thank you very much @NateWr for the hint (from July). Sorry for the late response on that. I have added a new column to the grid containing the file modification date which currently looks like this:
filedate

The changes imply some smaller changes/adjustments and one new file called `FileDateGridColumn.inc.php`, see pull request https://github.com/pkp/pkp-lib/pull/4019 . Would be nice if this might be considered in this or a similar way in one of the next releases.

Thanks for the support!

PR #3856, opened 2018-06-30
pkp/pkp-lib#3836 Privacy consent statement for registrationFormContexts

Vitaliy-1:
As I encountered with this issue while testing a new theme, some thoughts on managing of Privacy Statement for registration from the site-wide registration page. Also, should the term `Privacy Statement` be consistent for all cases? Because for default theme `Policy Statement` is used instead.


Vitaliy-1:
Hmm. Another thing, as I see it from the `registrationFormContexts` template, is that there is a possibility that registration for reader and reviewer groups can be disabled for all of the journals. But the form for entering registrational data would be still accessible and this can be quite confusing. Maybe there is a reason to add a check if any journal permits registration before redirecting to the register page from multi-journal site?

Also, just wondering, why user cannot register as an author from this page?

Vitaliy-1:
Sorry, the second commit should not be here.

Vitaliy-1:
Closing as it requires a rethinking of the concept

Issue #3868, opened 2018-07-03
lib/pkp/tools/installPluginVersion.php fails with SetCharSet() error


mfelczak:
Running into this issue with OJS 3.1.1-2 when trying to install a plugin, e.g.

```
$ php lib/pkp/tools/installPluginVersion.php plugins/generic/pln/version.xml

Dictionary driver '/usr/local/var/www/vhosts/ojs3_stable/lib/pkp/lib/adodb/datadict/datadict-.inc.php' not available

Fatal error: Call to a member function SetCharSet() on boolean in /usr/local/var/www/vhosts/ojs3_stable/lib/pkp/classes/plugins/Plugin.inc.php on line 708
```

```
$ php lib/pkp/tools/installPluginVersion.php plugins/generic/externalFeed/version.xml

Dictionary driver '/usr/local/var/www/vhosts/ojs3_stable/lib/pkp/lib/adodb/datadict/datadict-.inc.php' not available

Fatal error: Call to a member function SetCharSet() on boolean in /usr/local/var/www/vhosts/ojs3_stable/lib/pkp/classes/plugins/Plugin.inc.php on line 708
```

asmecher:
Hmm, I've never seen this before -- is this occurring in a properly-installed OJS that's otherwise functioning properly?

mfelczak:
Yes, initially ran into this with a hosted client install but also duplicated on my local install.

mfelczak:
Hey @bozana, just confirming that I've tested with the following plugins:
- pln
- static pages

bozana:
I can take a look...

bozana:
PRs:
pkp-lib ojs-stable-3_1_1: https://github.com/pkp/pkp-lib/pull/3877
ojs ojs-stable-3_1_1: (only submodule update)

pkp-lib omp-stable-3_1_1:
omp omp-stable-3_1_1: (only submodule update)

pkp-lib master:
ojs master: (only submodule update)

bozana:
@mfelczak, would it be possible for you to test the PR above?
I did get another error message, but I think the error source could be the same as yours...
When OK, I will ask @asmecher for the review...
Thanks a lot!!!

mfelczak:
Hi @bozana, that fixes it for me.

bozana:
@asmecher, could you also please take a look at those changes? Thanks!!!

asmecher:
[...]

PR #3877, opened 2018-07-05
pkp/pkp-lib#3868 fix installPluginVersion tool

bozana:
s. https://github.com/pkp/pkp-lib/issues/3868


tools/installPluginVersion.php
```diff
@@ -68,11 +68,23 @@ function execute() {
 		if ($plugin && is_object($plugin)) {
 			PluginRegistry::register($matches[1], $plugin, $pluginPath);
 		}
+		$plugin = PluginRegistry::getPlugin($matches[1], $plugin->getName());
```
asmecher: Is this line necessary?

bozana: For some strange reason without that line, the plugin object does not contain anythig, no pluginPath and no pluginCategory :-O

Issue #3578, opened 2018-04-10
Clean code for PHP 7


crism:
Summary of PRs: https://github.com/pkp/pkp-lib/issues/3578#issuecomment-404679076

## ojs
- [x] `PHP Warning:  Declaration of SubmissionFileDAO::fromRow($row) should be compatible with PKPSubmissionFileDAO::fromRow($row, $fileImplementation) in classes/article/SubmissionFileDAO.inc.php on line 23`
- [x] `PHP Warning:  Declaration of & UserSettingsDAO::getSetting($userId, $name, $journalId = NULL) should be compatible with & PKPUserSettingsDAO::getSetting($userId, $name, $assocType = NULL, $assocId = NULL) in classes/user/UserSettingsDAO.inc.php on line 19`
- [x] `PHP Warning:  Declaration of & UserSettingsDAO::getUsersBySetting($name, $value, $type = NULL, $journalId = NULL) should be compatible with & PKPUserSettingsDAO::getUsersBySetting($name, $value, $type = NULL, $assocType = NULL, $assocId = NULL) in classes/user/UserSettingsDAO.inc.php on line 19`
- [x] `PHP Warning:  Declaration of UserSettingsDAO::deleteSetting($userId, $name, $journalId = NULL) should be compatible with PKPUserSettingsDAO::deleteSetting($userId, $name, $assocType = NULL, $assocId = NULL) in classes/user/UserSettingsDAO.inc.php on line 19`
- [x] `PHP Warning:  Declaration of UserSettingsDAO::updateSetting($userId, $name, $value, $type = NULL, $journalId = NULL) should be compatible with PKPUserSettingsDAO::updateSetting($userId, $name, $value, $type = NULL, $assocType = NULL, $assocId = NULL) in classes/user/UserSettingsDAO.inc.php on line 19`
- [x] `PHP Warning:  Declaration of IssueGridRow::initialize($request) should be compatible with GridRow::initialize($request, $template = NULL) in controllers/grid/issues/IssueGridRow.inc.php on line 0`
- [x] `PHP Warning:  Declaration of SettingsPluginGridHandler::loadCategoryData($request, $categoryDataElement, $filter) should be compatible with PluginGridHandler::loadCategoryData($request, &$categoryDataElement, $filter) in controllers/grid/settings/plugins/SettingsPluginGridHandler.inc.php on line 18`
[...]

PR #3879, opened 2018-07-05
pkp/pkp-lib#3578: harmonizing Form::fetch() signatures

crism:
Fixes part of #3578. May break some plugins.

### DIFF
--- a/classes/controllers/tab/settings/form/ContextSettingsForm.inc.php
+++ b/classes/controllers/tab/settings/form/ContextSettingsForm.inc.php
@@ -99,10 +99,9 @@ class ContextSettingsForm extends Form {
 
 	/**
 	 * @see Form::fetch()
-	 * @param $request PKPRequest
 	 * @param $params array optional
 	 */
-	function fetch($request, $params = null) {
+	function fetch($request, $template = null, $display = false, $params = null) {
 		$templateMgr = TemplateManager::getManager($request);
 
 		// Insert the wizardMode parameter in params array to pass to template.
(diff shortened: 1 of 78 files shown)


classes/controllers/tab/settings/form/ContextSettingsForm.inc.php
```diff
@@ -99,10 +99,9 @@ function readInputData($request) {
 
 	/**
 	 * @see Form::fetch()
-	 * @param $request PKPRequest
 	 * @param $params array optional
 	 */
-	function fetch($request, $params = null) {
+	function fetch($request, $template = null, $display = false, $params = null) {
```
asmecher: @crism, did you check for cases where `$display` is `true`? I haven't seen any of these for a while, and would rather remove them if it doesn't tangle things up too much.

crism: I’m not aware of any cases where either the template or the display are actually passed explicitly… but this API is used by themes and plugins, so I didn’t want to disrupt it, just do the minimum to end the error messages.

asmecher: I think that parameter was more a quirk of the way `TemplateManager::display` used to delegate to `TemplateManager::fetch` with a similar `$display` parameter. That got removed with the move to Smarty3, IIRC, so I'd recommend removing it here too. I don't think it's used and I find it weird and confusing. (Which is a common reaction for me, so YMMV.)

asmecher: This looks ready to go per the testing, but I'm not sure whether you saw the above comment.

PR #3882, opened 2018-07-06
Control maxIdentifiers using a config setting

paulo-graca:
500 is a large number for a listing. In our case, without any caching mechanism, the page takes about 9sec to display 500 identifiers. We need to control the number of displayed identifiers.

### DIFF
--- a/classes/oai/OAIStruct.inc.php
+++ b/classes/oai/OAIStruct.inc.php
@@ -36,7 +36,7 @@ class OAIConfig {
 	var $tokenLifetime = 86400;
 
 	/** @var int maximum identifiers returned per request */
-	var $maxIdentifiers = 500;
+	var $maxIdentifiers;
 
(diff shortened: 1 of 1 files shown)


classes/oai/OAIStruct.inc.php
```diff
@@ -54,6 +54,9 @@ function __construct($baseUrl, $repositoryId) {
 
 		$this->maxRecords = Config::getVar('oai', 'oai_max_records');
 		if (!$this->maxRecords) $this->maxRecords = 100;
+		
+		$this->maxIdentifiers = Config::getVar('oai', 'oai_max_identifiers');
+		if (!$this->maxIdentifiers) $this->maxIdentifiers = 500;
```
asmecher: You can pass the `500` default in as a third parameter to `Config::getVar`.

PR #3883, opened 2018-07-06
pkp/pkp-lib#3578: harmonizing Form::initData() signatures

crism:
Fixes part of #3578; may break plugins.

### DIFF
--- a/classes/controllers/grid/GridHandler.inc.php
+++ b/classes/controllers/grid/GridHandler.inc.php
@@ -866,7 +866,7 @@ class GridHandler extends PKPHandler {
 	}
 
 	/**
-	 * Render the filter (a template or a Form).
+	 * Render the filter (a template).
 	 * @param $request PKPRequest
 	 * @param $filterData Array Data to be used by the filter template.
 	 * @return string
--- a/classes/controllers/grid/GridHandler.inc.php
+++ b/classes/controllers/grid/GridHandler.inc.php
@@ -876,16 +876,6 @@ class GridHandler extends PKPHandler {
 		switch(true) {
 			case $form === null: // No filter form.
 				return '';
-			case is_a($form, 'Form'): // Form object subclass
-				// Only read form data if the clientSubmit flag has been checked
-				$clientSubmit = (boolean) $request->getUserVar('clientSubmit');
-				if($clientSubmit) {
-					$form->readInputData();
-					$form->validate();
-				}
-
-				$form->initData($filterData, $request);
-				return $form->fetch($request);
 			case is_string($form): // HTML mark-up
 				$templateMgr = TemplateManager::getManager($request);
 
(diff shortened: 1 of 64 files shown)


classes/controllers/grid/GridHandler.inc.php
```diff
@@ -876,16 +876,6 @@ protected function renderFilter($request, $filterData = array()) {
 		switch(true) {
 			case $form === null: // No filter form.
 				return '';
-			case is_a($form, 'Form'): // Form object subclass
```
asmecher: I think this case is still used (grep around for subclasses of `ClientFormHandler`).

controllers/grid/settings/genre/form/GenreForm.inc.php
```diff
@@ -61,9 +61,9 @@ function __construct($genreId = null) {
 	/**
 	 * Initialize form data from current settings.
 	 * @param $args array
-	 * @param $request PKPRequest
 	 */
-	function initData($args, $request) {
+	function initData($args) {
```
asmecher: Was `$args` intentionally left here?

controllers/grid/settings/submissionChecklist/SubmissionChecklistGridHandler.inc.php
```diff
@@ -124,7 +124,7 @@ function editItem($args, $request) {
 		$submissionChecklistId = isset($args['rowId']) ? $args['rowId'] : null;
 		$submissionChecklistForm = new SubmissionChecklistForm($submissionChecklistId);
 
-		$submissionChecklistForm->initData($args, $request);
+		$submissionChecklistForm->initData($args);
```
asmecher: (As above -- not sure if `$args` is still needed)

classes/controllers/grid/GridHandler.inc.php
crism: @asmecher, pointers? I don’t see `ClientFormHandler` anywhere in the code base, except in JavaScript. I looked pretty thoroughly at usage before deleting that clause.

controllers/grid/settings/genre/form/GenreForm.inc.php
crism: Yes; it’s referenced at the end of the body of that method. There may be a better way to pass it in, but this satisfies PHP’s warnings.

controllers/grid/settings/submissionChecklist/SubmissionChecklistGridHandler.inc.php
crism: As above; `SubmissionChecklistForm` relies on a few things being passed in.

classes/controllers/grid/GridHandler.inc.php
asmecher: Just grepped a little further and it looks like you're right. I think the submission lists used to use this case but they've been adapted to vue.js. However, I think you still need to change the self-doc for GridHandler::getFilterForm.

PR #3907, opened 2018-07-12
pkp/pkp-lib#3578: harmonizing GridHandler signatures

crism:
Fixes part of #3578.


controllers/grid/files/review/#ReviewCategoryGridDataProvider.inc.php#
```diff
@@ -0,0 +1,80 @@
+<?php
```
asmecher: @crism, I think this was an accidental add.

PR #3916, opened 2018-07-15
Added uk_UA locale for generic plugins

algerds:


mtub:
Thanks, @algerds - the usageStats plugin is no longer part of pkp-lib, it has been moved to a separate repository. Please open a pull request there for that plugin.

PR #3934, opened 2018-07-17
Improve text when no download info is available

gonetil:
The text in Spanish was wrong. Here goes a better version ;-)


mtub:
@marcbria , thoughts?

marcbria:
@gonetil is right. Please accept the change.

Issue #3941, opened 2018-07-18
Permit dependent files in other areas of the workflow


asmecher:
Currently dependent files can only be added/administered on the galleys grid. Allow them elsewhere (e.g. for dependent files on XML documents).

asmecher:
PRs:
- https://github.com/pkp/pkp-lib/pull/3942
- https://github.com/pkp/ojs/pull/2070
- https://github.com/pkp/omp/pull/583

asmecher:
@bozana, when (if) the tests pass, could you take a look at this?

asmecher:
Merged!

PR #3942, opened 2018-07-18
pkp/pkp-lib#3941 Permit dependent files anywhere in the workflow

asmecher:


bozana:
@asmecher, everything looks good. I've tested it and I have one problem: when I upload an HTML Article Text file, then on the second tab, where I can also edit the file name, now I can see the dependent files grid. When I then try to upload a dependent file there clicking on "Upload file", the new modal opens, but it seems like the tabs handler is not functioning well there -- I only see a list of tabs as HTML list menu items. Is this also for you so?

asmecher:
Thanks, @bozana, good spotting -- there was a problem with non-unique markup elements clobbering each other. I've added another commit that fixes this: https://github.com/pkp/pkp-lib/pull/3942/commits/665e610a422a9418ce3c6e1e79d7926611e9abdc

Issue #3860, opened 2018-07-02
Allow to connect on custom ports / socket


j4rk0:
Following two lines need to be added there, so the query is going to be correctly build in the Illuminate Database component for connections on custom ports / sockets.
```
'port' => Config::getVar('database', 'port'),
'unix_socket'=> Config::getVar('database', 'unix_socket'),
```
These properties could be then set in config file…

asmecher:
@j4rk0, thanks -- would you mind opening a pull request to add this?

asmecher:
PR: https://github.com/pkp/pkp-lib/pull/3943

bozana:
PR:
https://github.com/pkp/pkp-lib/pull/4247
https://github.com/pkp/ojs/pull/2171

bozana:
@asmecher, OK so?

asmecher:
Yes -- please go ahead!

bozana:
Merged, thus closing...
Thanks @j4rk0 for contributing!!!

PR #3943, opened 2018-07-19
pkp/pkp-lib#3860 Allows to connect on custom ports

j4rk0:


bozana:
@asmecher, shall we then add the options for port and unix_socket to the config template? 
adodb and illuminate use different parameters for the connection: adodb uses "host:port" as the host parameter, and illuminate expects the extra port parameter. thus how to have the configuration settings optimal/the same for both? I see only two ways:
-- use additional settings port and unix_socket in the config, and then construct the host for adodb connection in the code like this: "host:port"
-- as till now use "host:port" in the config and parse the port for the illuminate.
I suppose the first solution is cleaner, right?
thanks!!!

bozana:
Merged within this PR https://github.com/pkp/pkp-lib/pull/4247, thus closing...

PR #3953, opened 2018-07-26
update locale fr_FR

forgive38:
some updates for locale fr_FR : I think these updates can be applied to master


mtub:
Thank you, @forgive38!

PR #3972, opened 2018-08-03
Typo on string #31 & #385

kant:


mtub:
Thanks, @kant. Seems to be in accordance with the Spanish translation standards (replacing "email" e.g.). Will merge now. Just mentioning @marcbria in case they want to have a closer look later.

PR #3978, opened 2018-08-10
Updated cs_CZ translation

Jirrka:
Translation to Czech language according to OJS 3.1.1.2


Jirrka:
Canceled, first I have to solve the conflict

PR #3987, opened 2018-08-15
Fix queryid call in getCanOpenClose

ajnyga:
Causes PHP Notice:  Undefined variable: queryId in /lib/pkp/controllers/grid/queries/QueriesAccessHelper.inc.php on line 62


asmecher:
Good spotting, @ajnyga! Porting around.

PR #4002, opened 2018-08-22
Update URL of Addthis.com Custom buttons

kant:


mtub:
@kant, thanks for spotting this. The string in question has been a duplicate entry, there is a newer translation in that file that is aligned with the English version. 

I've seen that the URL we have used so far is being redirected to the URL you have provided, so I will update all language files accordingly.

PR #4003, opened 2018-08-22
Update addthis support URLs

mtub:
as reported in #4002 by @kant 

@asmecher, could you have a quick look?


asmecher:
Merged and ported to `ojs-stable-3_1_1` and `omp-stable-3_1_1`, thanks!

PR #4069, opened 2018-09-17
 pkp/pkp-lib#4031 Submissions, implemented saving of submission requirements and privacy consent state

ViRuSTriNiTy:
see issue #4031


asmecher:
Looks good, @ViRuSTriNiTy -- one suggestion, though: hooks are messy to read and hard to maintain, and we try to reserve their use for cases like e.g. plugins where we don't want to add external concerns to the core code. Since this is all within the application/pkp-lib codebases anyway, I'd suggest adding the new fields to the DAO directly and dispensing with the hooks.

ViRuSTriNiTy:
@asmecher I removed the hooks and replaced them with appropriate method calls.

PR #4080, opened 2018-09-20
Make install of required package for tests explictly non-interactive.

ptrckb:


asmecher:
Merged -- thanks, @ptrckb!

PR #4093, opened 2018-09-26
Small changes to username instructions in Fi locale

ajnyga:
@mtub 

btw. why do we have both 
user.register.form.usernameAlphaNumeric and user.register.usernameRestriction ?


asmecher:
Merged & cherry-picked over to `stable-3_1_2`, thanks!

Issue #4014, opened 2018-08-27
Option to use a global sender address for emails (to fight spamfilters?)


ajnyga:
Currently OJS and OMP will send out emails so that the sender address is the actual address of for example the editor making a decision. 

In some cases this could lead to a situation where the email is marked as spam. Especially, if the sender and the receiver are from the same institution and use the same email server (which is probably quite common in many cases).

Suggestion:
- additional global_sender setting in config.inc.php
- if the setting is filled in with an email address, OJS will use that address for sending all emails. OJS installations can the create an actual email address (like no-reply@myinstitution.com) on their own email server and just make sure that the server keeps whitelisted
- the user email in OJS is used only in the reply-to field and the name of the sender
- additionally the sender name could be changed to something like "John Doe via sitename.com", because some users may be suspicious if they see a sender name mixed with an uknown email address

This is probably something that only requires small changes around here https://github.com/pkp/pkp-lib/blob/master/classes/mail/Mail.inc.php#L306 where the code checks for the setting in config.inc.php and if found, replaces the email address and sets the reply-to field.

Thoughts?

forgive38:
Hi all,

We need exactly this functionality
Thank you

asmecher:
@ctgraham, you've been fighting with DMARC etc. for quite some time and I think you're the best authority we've got on the subject -- do you think this would be a good long-term approach?

My sense is that it would, and might also facilitate some things we'd like to approach -- e.g. with a dedicated email account for the journal, we can start experimenting with getting OJS to consume email and not just send it. (Think responding directly to emails on discussions, for example.)

asmecher:
(FYI @mfelczak)

ctgraham:
@ajynga, the `default_envelope_sender` and `force_default_envelope_sender` in config.inc.php should already function this way, [courtesy of `Mail::getEnvelopeSender()`](https://github.com/pkp/pkp-lib/blob/22837fc09f5f95c21434095d842925d61612879e/classes/mail/Mail.inc.php#L489) and the [configuration check there](https://github.com/pkp/pkp-lib/blob/master/classes/mail/Mail.inc.php#L84-L86).

The `Reply-To` component is missing, but would readily be added [in the `Mail::Send` method](https://github.com/pkp/pkp-lib/blob/22837fc09f5f95c21434095d842925d61612879e/classes/mail/Mail.inc.php#L489).

If user addresses cannot be trusted implicitly, the best near-term and mid-term configuration is the use of:
```
[...]

PR #4119, opened 2018-10-03
pkp/pkp-lib#4014: Ensure envelope sender is not overwritten, enable multiple reply-tos and copy From to Reply-To when envelope sender is explicitly set

ctgraham:
This should improve SFP/DMARC compliance.

I'm pretty sure the envelope sender set at:
https://github.com/pkp/pkp-lib/blob/b7909a4f624828139675846b1b213d76d4f41a4d/classes/mail/Mail.inc.php#L489
is being overwritten by:
https://github.com/pkp/pkp-lib/blob/b7909a4f624828139675846b1b213d76d4f41a4d/classes/mail/Mail.inc.php#L491
because unset parameter `auto` defaults to `true`:
http://phpmailer.github.io/PHPMailer/classes/PHPMailer.PHPMailer.PHPMailer.html#method_setFrom

We also can/should copy the From address to a Reply-To when the Envelope Sender diverges from the From address, just for clarity.

Finally, this means we should treat Reply-To as an array of addresses, not just a single address.  Changing the semantics of `Mail::getReplyTo()` is a backwards-incompatible change, but the only usage instance in core (or the default plugins) is here in `Mail.inc.php`.

### DIFF
--- a/classes/mail/Mail.inc.php
+++ b/classes/mail/Mail.inc.php
@@ -322,8 +322,23 @@ class Mail extends DataObject {
 	 */
 	function setReplyTo($email, $name = '') {
 		if ($email === null) $this->setData('replyTo', null);
-		$this->setData('replyTo', array('name' => $name, 'email' => $email));
+		$this->setData('replyTo', array(array('name' => $name, 'email' => $email)));
 	}
+
+	/**
+	* Add a reply-to for the message.
+	* @param $email string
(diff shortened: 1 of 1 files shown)


asmecher:
This looks good, and I can't think of any downsides, though there's nothing quite like distributing FOSS to turn that up. Thanks, @ctgraham!

PR #3627, opened 2018-04-20
update es_ES translation in context

carzamora:
this line is not correct, @marcbria in this moment of the process the manuscript does not have reviewers, and the editors need to assign some reviewers.

If you prefer another sentence just tell me, and I can change it

PR #4127, opened 2018-10-08
update es_ES translation in context (new)

carzamora:
@mtub  this is for replacing #3627 because I change my nick from t4x0n to carzamora. 

I think this is the problem for sign my old PR.


mtub:
Thanks, @carzamora and @marcbria .

PR #4128, opened 2018-10-08
pkp/pkp-lib#4126 hide declined open reviews from authors

### LINKED ISSUE #4126: [OJS 3] Clarify reviewer state in ReviewerGrid
In our OJS 3.1.1.2 we had a misunderstanding the reviewers grid. The submission will be reviewed _open_. The grid contains two requests for review. One is in the state _Request Accepted_, the other in the state _Declined_.

The author of the submission asked if the submission itself was declined because of the state _Declined_, not knowing that only the request for review was declined. I think the misunderstanding could be avoided when it was clear _what_ was declined.

Right now the _Declined_ comes from [__('common.declined')](https://github.com/pkp/pkp-lib/blob/master/controllers/grid/users/reviewer/ReviewerGridCellProvider.inc.php#L166). I would suggest to create a new message _submission.review.status.declinedShort_ "Review declined". Maybe it would be also nice to have _submission.review.status.declined_ "The reviewer declined this review request." as a tooltip there.

So we would end up with:

`return '<span class="state declined" title="__('submission.review.status.declined')">'.__('submission.review.status.declinedShort').'</span>';`

What do you think?

ajnyga:
Fix #4126
Incomplete sql. 

Ping @asmecher, this is for 3.2 and should maybe be backported to 3.1. Hopefully the last of open review bugs saga.

### DIFF
--- a/classes/submission/reviewAssignment/ReviewAssignmentDAO.inc.php
+++ b/classes/submission/reviewAssignment/ReviewAssignmentDAO.inc.php
@@ -49,7 +49,7 @@ class ReviewAssignmentDAO extends DAO {
 	function getOpenReviewsByReviewRoundId($reviewRoundId) {
 		$params = array((int)$reviewRoundId, SUBMISSION_REVIEW_METHOD_OPEN);
 		$query = $this->_getSelectQuery() .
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @ajnyga! I've merged this and back-ported to both `ojs-stable-3_1_1` and `omp-stable-3_1_1`.

Issue #4133, opened 2018-10-11
Fix review assignment updated notification recipient


ajnyga:
Ping @asmecher, the Saga continues.

See https://forum.pkp.sfu.ca/t/review-assignment-updated/49082/10

pr in a moment

ajnyga:
edit: ok so this was actually already fixed in the master branch but committing a small addition to the code anyway (comparing the changed dates was not working because the other returned value included the time stamp 00:00:00)

asmecher:
(PR: https://github.com/pkp/pkp-lib/pull/4134)

asmecher:
Thanks, @ajnyga, I've merged that into `master`!

PR #4134, opened 2018-10-11
pkp/pkp-lib#4133 fix comparison of datetime

ajnyga:
ping @asmecher


asmecher:
Call me pedantic :) but I'd rather not do string operations on dates. I'd prefer something like this, even though it's longer... (needs testing!)

    $dueDate = strtotime($reviewAssignment->getDateDue());
    $reviewDueDate = strtotime($this->getData('reviewDueDate'));
    if (date('Y-m-d', $dueDate) != date('Y-m-d', $reviewDueDate)) {
        ...

One thing I'd **really** love to do after @NateWr has merged his form changes is standardize date handling -- the current mix of strings and numerics is maddening!

ajnyga:
You already have that in OJS in several places. I mean that substr used against those same values.

ajnyga:
This seems to work as well, would it be ok?

`strtotime($reviewAssignment->getDateDue()) != strtotime($this->getData('reviewDueDate'))`

`strtotime('2018-02-02 00:00:00')` and  `strtotime('2018-02-02')` return the same timestamp.

asmecher:
Yes, I think that would work. (Our date/time code is not particularly clean or consistent -- I'd like to get that cleaned up at some point!)

ajnyga:
changed

PR #4135, opened 2018-10-11
Section policies display

ViRuSTriNiTy:
see issue #2638

### DIFF
--- a/classes/context/PKPSectionDAO.inc.php
+++ b/classes/context/PKPSectionDAO.inc.php
@@ -83,23 +83,9 @@ abstract class PKPSectionDAO extends DAO {
 
 	/**
 	 * Retrieve all sections for a context.
-	 * @param $contextId int
-	 * @param $rangeInfo DBResultRange optional
-	 * @param $submittableOnly boolean ooptional. Whether to return only sections
-	 *  that can be submitted to by anyone.
 	 * @return DAOResultFactory containing Sections ordered by sequence
 	 */
-	public function getByContextId($contextId, $rangeInfo = null, $submittableOnly = false) {
-		$result = $this->retrieveRange(
-			'SELECT * FROM ' . $this->_getTableName() . ' WHERE ' .
-				$this->_getContextIdColumnName() . ' = ? ' .
-				($submittableOnly ? ' AND editor_restricted = 0' : '') .
-				' ORDER BY seq',
-			(int) $contextId, $rangeInfo
-		);
-
-		return new DAOResultFactory($result, $this, '_fromRow');
-	}
+	abstract function getByContextId($contextId, $rangeInfo = null);
 
 	/**
 	 * Retrieve the IDs and titles of the sections for a context in an associative array.
--- a/classes/context/PKPSectionDAO.inc.php
+++ b/classes/context/PKPSectionDAO.inc.php
@@ -51,7 +51,7 @@ abstract class PKPSectionDAO extends DAO {
 	 * @return array
 	 */
 	function getLocaleFieldNames() {
-		return array_merge(parent::getLocaleFieldNames(), array('title'));
+		return array_merge(parent::getLocaleFieldNames(), array('title', 'policy'));
 	}
 
 	/**
@@ -86,6 +86,22 @@ abstract class PKPSectionDAO extends DAO {
 	 * @return DAOResultFactory containing Sections ordered by sequence
 	 */
 	abstract function getByContextId($contextId, $rangeInfo = null);
+
+	/**
+	 * Retrieve the IDs and titles of the sections for a context in an associative array.
+	 * @param $contextId int context ID
+	 * @param $submittableOnly boolean optional
+	 * @return array
+	 */
+	function getTitles($contextId, $submittableOnly = false) {
+		$sections = array();
+		$sectionsIterator = $this->getByContextId($contextId);
+		while ($section = $sectionsIterator->next()) {
+			if ($submittableOnly && $section->getEditorRestricted()) continue;
+			$sections[$section->getId()] = $section->getLocalizedTitle();
+		}
+		return $sections;
+	}
 }
 
 
--- a/classes/context/PKPSectionDAO.inc.php
+++ b/classes/context/PKPSectionDAO.inc.php
@@ -83,9 +83,23 @@ abstract class PKPSectionDAO extends DAO {
 
 	/**
 	 * Retrieve all sections for a context.
+	 * @param $contextId int
+	 * @param $rangeInfo DBResultRange optional
+	 * @param $submittableOnly boolean ooptional. Whether to return only sections
+	 *  that can be submitted to by anyone.
 	 * @return DAOResultFactory containing Sections ordered by sequence
 	 */
-	abstract function getByContextId($contextId, $rangeInfo = null);
+	public function getByContextId($contextId, $rangeInfo = null, $submittableOnly = false) {
+		$result = $this->retrieveRange(
+			'SELECT * FROM ' . $this->_getTableName() . ' WHERE ' .
+				$this->_getContextIdColumnName() . ' = ? ' .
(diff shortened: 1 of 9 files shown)


pages/about/AboutContextHandler.inc.php
```diff
@@ -79,6 +79,36 @@ function submissions($args, $request) {
...
+
+				$section = $sectionDao->getById($sectionId);
+
+				$sectionPolicy = $section ? $section->getLocalizedPolicy() : null;
+				$sectionPolicyPlainText = trim(PKPString::html2text($sectionPolicy));
+				if (strlen($sectionPolicyPlainText) > 0)
+					$item['policy'] = $sectionPolicy;
+			});
+		} else {
+			AppLocale::requireComponents(LOCALE_COMPONENT_APP_AUTHOR); // for author.submit.notAccepting
```
asmecher: This locale key doesn't exist in OMP (which also uses this library). It would need to be added there too (with "press" instead of "journal").

pages/about/AboutContextHandler.inc.php
```diff
@@ -79,6 +79,36 @@ function submissions($args, $request) {
...
+
+		if (count($sections) > 0) {
+			array_walk($sections, function (&$item, $sectionId) use ($sectionDao) {
+				$item = array('title' => $item);
+
+				$section = $sectionDao->getById($sectionId);
+
+				$sectionPolicy = $section ? $section->getLocalizedPolicy() : null;
+				$sectionPolicyPlainText = trim(PKPString::html2text($sectionPolicy));
+				if (strlen($sectionPolicyPlainText) > 0)
```
asmecher: Is this to avoid section policies containing e.g. `<p></p>`? Maybe it would be better to address this on the storage side rather than the presentation side. (Especially since this same code also appears in https://github.com/pkp/ojs/pull/2132.)

pages/submission/PKPSubmissionHandler.inc.php
```diff
@@ -89,6 +89,8 @@ function wizard($args, $request) {
 		$step = isset($args[0]) ? (int) $args[0] : 1;
 		$templateMgr->assign('step', $step);
 
+		$templateMgr->assign('sectionId', $request->getUserVar('sectionId')); // to add a sectionId parameter to tab links in template
```
asmecher: I don't think it's a particular risk here, but best to force an `(int)` cast when reading in numeric IDs from the user. (Blanks/nonexistent parameters get cast to `0`, which evaluate to boolean `false`, so they are generally OK too.)

pages/submission/PKPSubmissionHandler.inc.php
```diff
@@ -222,6 +224,6 @@ function setupTemplate($request) {
 	 * @return int
 	 */
 	abstract function getStepCount();
-}
+		}
```
asmecher: Watch out for accidental whitespace changes.

templates/frontend/pages/submissions.tpl
```diff
@@ -15,20 +15,46 @@
...
+	<div class="submission_sections">
+		<h2>
+			{translate key="section.sections"}
+		</h2>
+		{if $sections|@count > 0}
+			<ul>
+				{foreach from=$sections key="sectionId" item="section"}
+					<li>
+						<h3>
+							{$section.title}
```
asmecher: Section titles should get the `escape` Smarty modifier. (HTML is not expected/permitted there.)

pages/about/AboutContextHandler.inc.php
ViRuSTriNiTy: This is basically the check for "is there a policy available". The code retrieves the localized policy, then converts it to plain text to determine whether the policy contains any readable text (as for example `<p></p>` isn't really a policy that needs to be displayed). So when a text is available the template variable to render the text in template accordingly. I hope this makes sense. Perhaps i should move this code to a method named like `doesSectionPolicyContainsAnyText()` to make the code more readable? What do you say?

pages/about/AboutContextHandler.inc.php
ViRuSTriNiTy: Should i provide a PR for OMP also? Or is there another possible solution?

pages/submission/PKPSubmissionHandler.inc.php
ViRuSTriNiTy: Makes sense, will do.

pages/submission/PKPSubmissionHandler.inc.php
ViRuSTriNiTy: Yep, dunno where this is coming from, i will fix it.

templates/frontend/pages/submissions.tpl
ViRuSTriNiTy: Fair enough.

pages/about/AboutContextHandler.inc.php
asmecher: Please! That's normally what we do. The codebases are pretty symmetrical (and getting more so as we further align them).

classes/context/PKPSectionDAO.inc.php
```diff
@@ -83,9 +83,39 @@ function deleteByContextId($contextId) {
...
 	/**
 	 * Retrieve all sections for a context.
+	 * @param $contextId int
+	 * @param $rangeInfo DBResultRange optional
+	 * @param $submittableOnly boolean ooptional. Whether to return only sections
+	 *  that can be submitted to by anyone.
 	 * @return DAOResultFactory containing Sections ordered by sequence
 	 */
-	abstract function getByContextId($contextId, $rangeInfo = null);
+	public function getByContextId($contextId, $rangeInfo = null, $submittableOnly = false) {
```
asmecher: This is currently abstract in `pkp-lib` because `ojs` (`SectionDAO`) and `omp` (`SeriesDAO`) offer different implementations of it; you'd need to consider the implications for OMP.

pages/about/AboutContextHandler.inc.php
asmecher: I think `SectionDAO` (OJS)/`SeriesDAO` (OMP)/`PKPSectionDAO` (pkp-lib) is the best candidate.

classes/context/PKPSectionDAO.inc.php
ViRuSTriNiTy: I suggest we leave the method abstract but add the `$submittableOnly` parameter. The implementation can then be extended in OMP to consider the editor restricted flag as it is done in OJS. If you agree I can provide a PR for OMP.

asmecher: Thanks, that sounds sensible to me.

ViRuSTriNiTy: Done, method remains abstract and `$submittableOnly` parameter added.

pages/submission/PKPSubmissionHandler.inc.php
ViRuSTriNiTy: Fixed.

templates/frontend/pages/submissions.tpl
ViRuSTriNiTy: Fixed in connection with template redesign in d7afda0 by @NateWr.

pages/about/AboutContextHandler.inc.php
ViRuSTriNiTy: Done via https://github.com/ViRuSTriNiTy/ojs/commit/ca35300a92d79c19b58b2947f430cee1f55de5ed

pages/about/AboutContextHandler.inc.php
ViRuSTriNiTy: Done via https://github.com/pkp/omp/pull/602/commits/1d8399c28001c2900918b1ce05c6f316c299fbd1

PR #4143, opened 2018-10-15
[pkp-lib] #3758 Support for managing submission metadata editing rights for authors

defstat:


asmecher:
(Addendum to https://github.com/pkp/pkp-lib/pull/4143#pullrequestreview-164847200: I remember speaking a long time ago with @lilients about granting/revoking permission to create new metadata revisions. I don't remember if any of that made it into the versioning changes but obviously it would make sense to align the two proposals!)

defstat:
Hi Alec! Thanks for the remarks. 

I have prepared the following: 

image

That is the form from the edit user_group form. At the database user_grup_settings there is a new setting_name='editPermits', with a value of an array [stage]=true or false. 

 That declares the global permits for each user group. 

Then at the submission's user assignment form:

image

we add a new checkbox. At the db table stage_assignments I have added a new column `permit_metadata_edits` with value 0 or 1. That can cover the "override" feature of the global assignment.

Now I have to take the user_group of the current user and make the following checks (do we do that anywhere?):
- if there are specific rules for the user assigned at the submission level, then we follow them 
- if there are global rules for the user_group then we follow them
- else we do the default (the code as-is now) 

Any thoughts on that?

asmecher:
A couple quick comments:

> At the database user_grup_settings there is a new setting_name='editPermits', with a value of an array [stage]=true or false.
>
> That declares the global permits for each user group.

Rather than serializing this (which means we can't do relational queries against it), why not add that as a new column to `user_group_stage`?

> Now I have to take the user_group of the current user and make the following checks (do we do that anywhere?):
>
>    if there are specific rules for the user assigned at the submission level, then we follow them
>    if there are global rules for the user_group then we follow them
>    else we do the default (the code as-is now)

I think when a user is assigned, the stage assignment form could be pre-filled from the new `user_group_stage` column (above). Then the new entry in `stage_assignments` gets created, and a new column (`can_change_metadata` or something) gets filled there. Then you would *only* need to check that column when seeing whether someone is allowed to make changes -- no need to consult the e.g. `user_group_stage` settings afterwards.

PR #4152, opened 2018-10-16
PT Translations

ambs:
Hi

Added some translations.
Can't undestand why I have the commit 2597025 here. But my git-fu is not good enough to solve it.
Can you cherry pick my other commit?

Thanks


mtub:
#4190

PR #4165, opened 2018-10-18
pkp/pkp-lib#4161 support generating DMARC compliant mails

tgc:
This goes on top of #4014 and makes it possible to munge the RFC5322.From based on new settings in config.inc.php.

Note that #4014 would always promote from to reply-to if the from address did not match the envelope sender. After this patch the from address will only be promoted to a reply-to if force_dmarc_compliant_from is enabled.


tgc:
You're not wrong. This patch will rewrite the RFC5322.From to avoid the problem you see. 
It also allows you to customize the display name in the rewritten RFC5322.From though currently it does not support extracting a 'sitename', only the 'John Doe' part via the %n modifier in the string.

ajnyga:
Hah great. What a coincidence that I was just today discussing this with a local university :)

I think it would be good the have something like "via Sitename" after the name so that people do not wonder about the mismatch between the sender and the reply-to. I saw that it was also in the FAQ you linked in the issue.

Do you have the code in production use?

tgc:
No I have not run this in production yet. I just finished this patch yesterday. I have done testing with the submission stage which sends out mail to all involved parties to verify that it does what I want.

ajnyga:
Fyi @tgc I created a plugin that replaces the current Mail::send and used some of the logic in your pull request. It works with the master branch version of the mail.inc.php (we use 3.1.1.2, but since it uses the hook that was fixed recently I updated that single file to master branch)

https://github.com/ajnyga/useGlobalSender

tgc:
A plugin is one way to solve it and it does avoid the need to modify pkp-lib in the current situation (well almost...), but I don't like the idea of having a slightly different version of core code in a plugin.
I still think this absolutely needs to be builtin functionality and maybe even the default.

ajnyga:
I absolutely agree that this is needed in the core. I like to avoid any local changes to the core code. We already have some that can not be solved with plugins, so anything that I can solve with a plugin is usually more preferable for me. Hopefully this will be in 3.2.

We have had the plugin in production at Journal.fi for a week now and seems good. We did some tests with a local university and everything is going past their spam filters.

tgc:
I added a commit that will allow use of the sitename in the From header. Thanks @ajnyga for showing me how to do that.

forgive38:
Hi,
Thank you, it seems to be exactly what we need to avoid problem with DMARC (especially with gmail addresses)

Regards

ajnyga:
@asmecher can we get this in 3.2?

asmecher:
@ajnyga, you mean the following two PRs...
- https://github.com/pkp/pkp-lib/pull/4165 (this one), and
- https://github.com/pkp/ojs/pull/2142
...correct?

If so, @ctgraham, would you mind taking a look at these two? I think we have confirmation from both @ajnyga and @tgc that they're useful/necessary for mail delivery.

ctgraham:
My only concern with this is that we are adding a new config.inc.php option "force_dmarc_compliant_from", which really just duplicates the original intent of the existing option "force_default_envelope_sender".  There is really no need for a distinction.

Let's refactor one or the other out of existence.  The "force_dmarc_compliant_from" has the benefit of being better named; "force_default_envelope_sender" may have historical value.

mfelczak:
Hi all, just wanted to add a +1 to include the latest mail updates here in the upcoming 3.1.2 release. It looks like we just need to resolve a few more questions.

> @ajnyga,you mean the following two PRs...
> 
> * #4165 (this one), and
> * [pkp/ojs#2142](https://github.com/pkp/ojs/pull/2142)
>   ...correct?

@ajnyga, are the above PRs still the relevant ones here? Are there any code changes needed to the above based on your production setup with your plugin?

@ctgraham, curious if you've had a chance to deploy the above in your hosting environment? Is the only outstanding issue for you the distinction between "force_default_envelope_sender" and "force_dmarc_compliant_from"?

tgc:
I have been running this in production on tidsskrift.dk for a few months now and have had no issues with it.
As far as I'm concerned these PRs are still the relevant ones.

I don't feel strongly about the "force_dmarc_compliant_from" setting. It might be better to just do what @ajnyga did with the GlobalSender plugin and let "force_default_envelope_sender" enable DMARC compliance. I don't think that would be too bad of a surprise for existing users.
I still think it would be a nice touch to have something similar to "dmarc_compliant_from_displayname" to make it possible to customize the name shown instead of having it hardcoded.

ajnyga:
Hi,

What Tom said above.

My plugin has had some problems with some users, but those are related to the use of the hook in mail.inc.php. The way the code works in the pull requests should be ok. I have used a similar code as a modification to the core since december.

I think using the envelope sender to trigger the functionality is ok. Not against adding a new setting though.

Important that the mail address has "via [site]" in the end! For me it is the same if we use the site name or have that as a configurable value like Tom suggests.

Just get this in the next release one way another 😄

ctgraham:
We don't yet have any 3.x journals in production which make use of email, so I haven't tried this patch in our environment.  From reading the patches, however, I have no concerns other than simplifying to a single configuration directive.

mfelczak:
Thanks @tgc, @ajnyga, and @ctgraham!

@asmecher, this looks like it's ready to go with the exception of the config.inc.php setting, which may need an executive decision.

asmecher:
Thanks, everyone! I've merged the PRs referenced in https://github.com/pkp/pkp-lib/pull/4165#issuecomment-456180619 and cherry-picked them over to OJS and OMP `stable-3_1_2` for the forthcoming 3.1.2 releases.

PR #4196, opened 2018-10-30
Adding a missing key to English locale

vormia:
This missing key is for the tool tip shown when hovering the action link (Users) under each record in the list of hosted journals.
It is under the permissions of site admin.


asmecher:
Thanks, @vormia!

PR #4199, opened 2018-10-30
Rivisions to the Arabic locale files

vormia:
These revisions were made after testing Arabic UI in a local installation


mtub:
Thanks, @vormia!

PR #4205, opened 2018-11-02
fix for a typo in PKPTemplateManager::smartyLoadStylesheet()

Vitaliy-1:
$context variable isn't used anywhere in this method.


asmecher:
Thanks, @Vitaliy-1!

PR #4230, opened 2018-11-12
Make user email lookup case insensitive.

mittinatten:
Make email-based user lookup case insensitive.

As discussed in https://forum.pkp.sfu.ca/t/forget-your-password-function-not-working-in-ojs-3-0-1/27063/24


asmecher:
Thanks, @mittinatten -- I've merged that for the 3.2 release!

Issue #4046, opened 2018-09-09
[OJS] Review form elements, proposal for a description field


ViRuSTriNiTy:
Hi there, 

i have added a description property to review form elements to display a description below the element title. The relevant discussion can be found in the [forums](https://forum.pkp.sfu.ca/t/description-for-review-form-elements/48070).

I will provide a PR.

So lonG
Daniel

ViRuSTriNiTy:
Sorry for the broken PR (3a12da3 and c32cad1), had a end of line sequence issue, created a new PR instead (#4047)

PR #4232, opened 2018-11-12
pkp/pkp-lib#4046 changed description tag from <span> to <div> 

ViRuSTriNiTy:
… to be HTML5 compliant as non-phrasing content is not allowed in <span>


asmecher:
Thanks, @ViRuSTriNiTy!

Issue #1619, opened 2016-07-18
Block adding discussion to oneself only


stranack:
When using the Add Discussion tool, it is possible for a user to only include themselves in the list of Participants. Testing showed people did this somewhat frequently.

Can we block saving unless the user has included someone other than themselves in the discussion?

jamshidhashimi:
PR:

pkp/pkp-lib#4303 

@asmecher please review. Thanks.

asmecher:
Merged. Thanks, @jamshidhashimi!

PR #4303, opened 2018-12-14
pkp/pkp-lib#1619 Blocks adding discussion to oneself only

jamshidhashimi:


controllers/grid/queries/form/QueryForm.inc.php
```diff
@@ -77,8 +77,11 @@ function __construct($request, $assocType, $assocId, $stageId, $queryId = null)
 
 		$this->setQuery($query);
 
+		$user = $request->getUser();
 		// Validation checks for this form
-		$this->addCheck(new FormValidator($this, 'users', 'required', 'stageParticipants.notify.warning'));
+		$this->addCheck(new FormValidatorCustom($this, 'users', 'required', 'stageParticipants.notify.warning', function($users) use ($user) {
+			return in_array($user->getId(), $users) && count($users) > 1;
```
asmecher: I'd suggest removing the first part of the check -- it might be valid for e.g. an Editor to create a conversation with an Author and a Copyeditor, but not want to participate in it themselves.

jamshidhashimi: Thanks and done.

asmecher:
Merged -- thanks, @jamshidhashimi!

PR #4346, opened 2019-01-07
pkp/pkp-lib#4340: limit available contexts by user

ctgraham:
Change logic of "available" contexts from:
Any of:
- an enabled context
- any context in which the user is a member of a user group
- any context, if the user is a site admin

To:
Either:
- if no user, an enabled context
- if a user, either:
  - any context, if the user is SITE_ADMIN
  - any context in which the user is a member of user group

### DIFF
--- a/classes/context/ContextDAO.inc.php
+++ b/classes/context/ContextDAO.inc.php
@@ -175,6 +175,9 @@ abstract class ContextDAO extends DAO {
 
 	/**
 	 * Retrieve available contexts.
+	 * If user-based contexts, retrieve all contexts assigned by user group
+	 *   or all contexts for site admin
+	 * If not user-based, retrieve all enabled contexts.
 	 * @param $userId int Optional user ID to find available contexts for
 	 * @param $rangeInfo Object optional
 	 * @return DAOResultFactory containing matching Contexts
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @ctgraham! @bozana, would you mind having a look at this?

ctgraham:
This is only part one of two for the parent issue.  The [second part](https://github.com/pkp/pkp-lib/issues/4340#issuecomment-451200517) is getting more messy by the day.  Can I get y'all's opinion on that as well?

bozana:
Yes, I think this is correct:
-- 'enabled' means publicly available. Thus, if a user is enrolled for such a journal and logged in, the user should have the access to the journal.
-- the only place where $contextDao->getAvailable() is called without userId is CounterReport and LegacyJR1, and here I think it is correct to consider only enabled journals.

@ctgraham, would you maybe like to commit into stable-3_1_2 -- it could be without conflicts -- and I can then cherry-pick it to ojs-stable-3_1_1 and adapt it to master?
Thanks a lot!

ctgraham:
In addition to the conflict, this PR will also require a fix for an outstanding problem: if the user is in the Site context and they only have access to one journal, the user navigation and context navigation only reflects the site context.

Issue #4398, opened 2019-01-25
User Import tries to assign user groups without a user id


ctgraham:
The assignment of user groups within the `UserXMLPKPUserFilter` depends on the persistence of a `User` via a `$userId`.
https://github.com/pkp/pkp-lib/blob/a15388c36aa6f2fff03ad34c107d07c8ed56aa89/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php#L242

This `$userId` was only assigned, however, if the user was created:
https://github.com/pkp/pkp-lib/blob/a15388c36aa6f2fff03ad34c107d07c8ed56aa89/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php#L213
or matched by username and email:
https://github.com/pkp/pkp-lib/blob/a15388c36aa6f2fff03ad34c107d07c8ed56aa89/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php#L159

The case of a mismatch of username and email is unhandled, and results in an unset `$userId` variable.

bozana:
merged, thus closing...

Issue #2135, opened 2016-12-31
Improve password reset to use one-use, time-limited hash


mtub:
**Describe the problem you would like to solve**
Password reset is a two-step process that emails a temporary password and asks the user to change it on the next login. This can be confusing for users who are accustomed to receiving a one-time link to set a new password when they forget a password.

**Describe the solution you'd like**
When the user has forgotten a password, send an email with a one-use, time-limited link with a hash. When clicking on the link, they can change their password and regain access to their account.

**Who is asking for this feature?**
System administrators and some users.

**Additional information**
This post was updated by @NateWr on 2021-11-15 to reflect the discussion below. The following was the original post.

> Passwords should not be emailed. We had a discussion (@ctgraham, @asmecher) about this, see #745. IIRC, it was decided to keep the support for the `password` variable in legacy versions of the applications, but remove it completely from the master/dev branches, see #750. That was 2015.
> 
> In 2016, we got this: #1469 (OJS) and this #1524 (OMP).
> 
> And now, OJS 3 still sends unencrypted emails with passwords. I cannot see why, but maybe I missed something. I'd really like to see this changed, what do you think?

TobiasWantzen:
What's the alternative? Spontaniously I'd send a temporary password by mail, valid for 2 hours, »User must change password on next log in.« activated.

WaDelma:
Usually password resets are handled by sending link to change the password that can be only used one time and is only valid for certain time.

asmecher:
As I understand it, we're emailing only temporary passwords when a change is required upon use, which is equivalent (in my opinion) to sending out a hash. There's at least one thing we could probably improve upon, which is a limited window for credentials (whether hash or password) sent in this way.

ctgraham:
I prefer eliminating any use of a temporary password in favor of consistent use of a password reset hash.  The hash is a dedicated-use tool and is inherently time-sensitive.

forgive38:
Hello,

Currently, with ojs 3.1.1, the process to reset a password is still in 2 steps:
1- receive a email with a one time link, then when you follow the link, you 
2 - receive a email with a temporary password, that you have to change the first time you use it.

I think the first step is enough.
One ambiguous thing for the second step, is the mail (email_text key="PASSWORD_RESET")
It is written "We have included your username and password in this email, which are needed for all work with this journal through its website." so I think the mail should be something like "the password included in this mail is temporary, you have to change it the next you log on the website"

thank you

[...]

PR #4399, opened 2019-01-25
pkp/pkp-lib#4398: Ensure user is persisted before trying assignUserToGroup()

ctgraham:
Resolves unset variable warning for pkp/pkp-lib#4398

### DIFF
--- a/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php
+++ b/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php
@@ -225,24 +225,27 @@ class UserXmlPKPUserFilter extends NativeImportFilter {
 			}
 		}
 
-		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
-		$userGroupsFactory = $userGroupDao->getByContextId($context->getId());
-		$userGroups = $userGroupsFactory->toArray();
+		// We can only assign a user to a user group if persisted to the database by $userId
+		if ($userId) {
+	  		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+  			$userGroupsFactory = $userGroupDao->getByContextId($context->getId());
+  			$userGroups = $userGroupsFactory->toArray();
 
-		// Extract user groups from the User XML and assign the user to those (existing) groups.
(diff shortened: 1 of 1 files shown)


bozana:
@ctgraham, yes, this is good, but I think we should also:
-- add an error, because it is a mismatch -- both username and e-mail are unique so it should be one and the same user
-- it seems to me that the USER_REGISTER e-mail should only be sent if a new user is inserted, or? -- If it is an existing user (with that username and e-mail) only the groups would be added to him/her, all other metadata are not changed, or? If so, maybe the e-mail should not be sent? What do you think?

If you agree I can add those...

bozana:
@asmecher, it seems to me that the USER_REGISTER e-mail should only be sent if a new user is being imported, or? -- Looking into the code, if the import XML contains an existing user (with that username and e-mail) only the groups would be added to him/her, all other metadata are not changed, or? If so, maybe the e-mail should not be sent in that case? What do you think?

asmecher:
It's common practice to send a "welcome" message after a user account is created (also to notify someone e.g. if their email has been misused by someone else). But per https://github.com/pkp/pkp-lib/issues/2135 we should definitely not be emailing passwords. In case it's a manager-created user we mitigate the risk by making it a one-time-use password (requiring reset upon the first login), but if it's a user who has registered their own account, a password is definitely inappropriate.

TL;DR: `USER_REGISTER` can be used for both self-registrations and manager-created accounts, but for self-registrations it should definitely not be sending passwords. So I see two approaches:
- Do not send a welcome email for self-registrations, or
- Create another email template for self-registrations that does not include a password.

I'd favour the second approach, on balance.

bozana:
Well... I actually thought something else:
When importing user, we look if that user already exists, s. https://github.com/pkp/pkp-lib/blob/stable-3_1_2/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php#L157-L159. In that case we also send the e-mail, as far as I can see: https://github.com/pkp/pkp-lib/blob/stable-3_1_2/plugins/importexport/users/filter/UserXmlPKPUserFilter.inc.php#L248-L255. So I was wondering why we would like to send that e-mail in that case i.e. if this is intended/correct behavior :-)
But, now you mentioned another thing/issue, that I will then take a look separately... :-)

PR #4476, opened 2019-02-13
pkp/pkp-lib#3535 Updating the Section settings heading

jamshidhashimi:
s. pkp/pkp-lib#3535


asmecher:
@jamshidhashimi, that's the right key, but I'd suggest renaming the locale key as well, not just its text. That'll mean changing any templates that refer to it.

Usually if there's just a minor change to the text of the locale key, we'll leave the other languages as they are -- but in this case it's a change of meaning, so it's probably best to remove that key from the other languages so that the translators know they need to re-translate it before the 3.2 release gets put together.

jamshidhashimi:
@asmecher thanks! I updated the pull request and sent a PR to OJS as well. Please see the [issue](https://github.com/pkp/pkp-lib/issues/3535) page for details.

PR #4509, opened 2019-02-17
make upgrade work again on postgres

helix84:
#3594 broke upgrade on postgres


asmecher:
Thanks, @helix84! Note that the `master` branch upgrade process is not yet supported/tested.

Issue #4514, opened 2019-02-19
[OJS] Recoginze https URL to Creative Commons licenses


zuphilip:
Currently, for the creative commons license the images are shown only if the URL is entered as a `http` URL and without an ending backslash, e.g. for
```
https://creativecommons.org/licenses/by-nc-nd/4.0/
```
you have to enter
```
http://creativecommons.org/licenses/by-nc-nd/4.0
```
I am speaking about the URL you can enter in the production step > schedule for publication:

![ojs-publication-license](https://user-images.githubusercontent.com/5199995/53031383-0adf5980-346d-11e9-87b4-ece99a13e8a6.png)

If you enter here another form of the CC-license, e.g. as a https-URL, then there will be no image shown. Is it possible to recognize in OJS different forms for the CC-licenses, especially the form suggested by CC itself?

jamshidhashimi:
@zuphilip, what version of OJS are you using?

Have you seen the following commits in the Master?
https://github.com/pkp/pkp-lib/commit/e45a1818f1d6dcda7539f226bdae737655d20993#diff-3fed0569f5f0ae40c985f62cfe134c27
https://github.com/pkp/pkp-lib/commit/cdbd845250bdd1f4c49dc69804ecace2bfa04e3a#diff-f69e5caa980791d6835909656e9eace7

cc: @asmecher

zuphilip:
We are using OJS 3.1.1.2. No, I haven't seen these changes. This might solve the `https` issue, but it seems then still that ending backslash will cause problems, when entered.

I would imagine, that several variants of entering a Creative Commons URL should be recognized. Maybe a simple cleaning step is enough to do first, i.e.
  1. Delete the protocol `http` or `https` at the beginning; delete a possible backslash at the end
  2. Check the cleaned string against a list of know licenses

asmecher:
@zuphilip, if you're able to test this change:
https://github.com/pkp/pkp-lib/pull/4539
...and confirm whether it works, I'll commit it for the soon-to-be-released OJS 3.1.2.

asmecher:
Thanks, @zuphilip!

PR #4539, opened 2019-02-25
pkp/pkp-lib#4514 Make CC license URL matching more flexible

asmecher:


classes/core/PKPApplication.inc.php
```diff
@@ -684,23 +684,25 @@ static function getCCLicenseOptions() {
...
-			'http://creativecommons.org/licenses/by-nd/4.0' => 'submission.license.cc.by-nd4.footer',
-			'http://creativecommons.org/licenses/by/4.0' => 'submission.license.cc.by4.footer',
-			'http://creativecommons.org/licenses/by-sa/4.0' => 'submission.license.cc.by-sa4.footer',
-			'http://creativecommons.org/licenses/by-nc-nd/3.0' => 'submission.license.cc.by-nc-nd3.footer',
-			'http://creativecommons.org/licenses/by-nc/3.0' => 'submission.license.cc.by-nc3.footer',
-			'http://creativecommons.org/licenses/by-nc-sa/3.0' => 'submission.license.cc.by-nc-sa3.footer',
-			'http://creativecommons.org/licenses/by-nd/3.0' => 'submission.license.cc.by-nd3.footer',
-			'http://creativecommons.org/licenses/by/3.0' => 'submission.license.cc.by3.footer',
-			'http://creativecommons.org/licenses/by-sa/3.0' => 'submission.license.cc.by-sa3.footer'
+			'|http[s]?://[www.]?creativecommons.org/licenses/by-nc-nd/4.0[/]?|' => 'submission.license.cc.by-nc-nd4.footer',
```
zuphilip: The dots `.` in regex should be escaped and the grouping of a single character is negligible, but for the `www` should use the normal brackets, i.e. I think you should instead use
```
https?://(www\.)?creativecommons\.org/licenses/by-nc-nd/4\.0/?
```
See also https://regex101.com/r/gbYJmU/1 . Similar for the other lines.

classes/core/PKPApplication.inc.php
```diff
@@ -684,23 +684,25 @@ static function getCCLicenseOptions() {
...
+			'|http[s]?://[www.]?creativecommons.org/licenses/by-sa/3.0[/]?|' => 'submission.license.cc.by-sa3.footer'
 		);
 
-		if (isset($licenseKeyMap[$ccLicenseURL])) {
-			PKPLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION);
-			return __($licenseKeyMap[$ccLicenseURL]);
+		foreach($licenseKeyMap as $pattern => $key) {
+			if (preg_match($pattern, $ccLicenseURL)) {
+				PKPLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION);
+				return __($licenseKeyMap[$ccLicenseURL]);
```
zuphilip: This line looks wrong for me. AFAIS `$licenseKeyMap[$ccLicenseURL]` should be replaced by `key`.

(I don't know what the `__(...)` part is doing or if we also have to change this.)

zuphilip:
I can't easily test, but I tried to look through the code and give you some comments.

classes/core/PKPApplication.inc.php
asmecher: Oops! Yes, my typo -- you're absolutely right.

classes/core/PKPApplication.inc.php
asmecher: Good spotting, @zuphilip, you're two for two. The `__` converts from a locale key to a translated string, but yes, it should be `$key` rather than `$licenseKeyMap[$ccLicenseURL]`. This is what happens when I try to code without testing :duck:

asmecher:
Thanks, @zuphilip -- I fixed the issues you spotted, tested the results, and merged manually. This will go into OJS 3.1.2.

classes/core/PKPApplication.inc.php
```diff
@@ -684,23 +684,25 @@ static function getCCLicenseOptions() {
...
-			'http://creativecommons.org/licenses/by-nd/4.0' => 'submission.license.cc.by-nd4.footer',
-			'http://creativecommons.org/licenses/by/4.0' => 'submission.license.cc.by4.footer',
-			'http://creativecommons.org/licenses/by-sa/4.0' => 'submission.license.cc.by-sa4.footer',
-			'http://creativecommons.org/licenses/by-nc-nd/3.0' => 'submission.license.cc.by-nc-nd3.footer',
-			'http://creativecommons.org/licenses/by-nc/3.0' => 'submission.license.cc.by-nc3.footer',
-			'http://creativecommons.org/licenses/by-nc-sa/3.0' => 'submission.license.cc.by-nc-sa3.footer',
-			'http://creativecommons.org/licenses/by-nd/3.0' => 'submission.license.cc.by-nd3.footer',
-			'http://creativecommons.org/licenses/by/3.0' => 'submission.license.cc.by3.footer',
-			'http://creativecommons.org/licenses/by-sa/3.0' => 'submission.license.cc.by-sa3.footer'
+			'|http[s]?://(www.)?creativecommons.org/licenses/by-nc-nd/4.0[/]?|' => 'submission.license.cc.by-nc-nd4.footer',
```
zuphilip: I would still escape the dots `\.` instead of `.` because then they only match a dot and not any character in a regex. Currently you are also matching e.g. `http://creativecommonsXorg/licenses/by-nc-nd/480`

zuphilip:
Thank you very much @asmecher for all the work! :bowing_man:

classes/core/PKPApplication.inc.php
asmecher: Right as usual. Tweaked that. Thanks!

Issue #4503, opened 2019-02-15
[OJS] Update nl_NL locale 


mtub:
* Updates to OJS, pkp-lib, and plugin translations in Dutch (nl_NL)

rvdb:
@mtub Ok, done, as far as I can see. I think I got it right during the plugin commits, which means that I've probably (mis?)filed a number of my earlier PR's to `master` branches instead of `stable-3_1_2`. If I should do anything to fix that, please tell me how. If I've overlooked anything, just shout!

mtub:
@rvdb , you can see the updates to all of your pull requests here. As soon as no PR is "open" anymore, we are done :-)

mtub:
that last one was a good thing to be able to do :)

rvdb:
> 
> 
> that last one was a good thing to be able to do :)

@mtub For sure, woohoo! If I'd get to tackle the `ojs-stable-3_1_1` branch, would I need to open a similar issue here, first? I hope I've learned from this bumpy trip, to a) first make sure my forked `ojs-stable-3_1_1` branch is up-to-date before branching it to a translation branch, and b) make sure to file PRs against this branch. Can't go worse than this, I'm convincing myself.

mtub:
Hi @rvdb, if you still want to address OJS 3.1.1, feel free to open another PR after completing your work.

rvdb:
plugin submodules (11) still have to come, hopefully tomorrow

rvdb:
@mtub done!

rvdb:
@mtub I realize there are probably more pressing issues, but it's been a month; have I filed my PRs at the wrong place?

mtub:
Everything is okay, @rvdb , sorry for the delay! Will review and merge.

rvdb:
Ok, thanks for merging, @mtub !

PR #4574, opened 2019-03-05
pkp/pkp-lib#4503 updated nl_NL locale for ojs-stable-3_1_1

rvdb:
updated nl_NL locale for pkp-lib


mtub:
Thank you so much, @rvdb!

PR #4585, opened 2019-03-11
Fix URIs for OAI namespaces

ptrckb:
Namespaces are not links and they do not have to be resolvable. Changing `http` to `https` define a completely different namespace that is often not standard.

This reverts commit 52edd3e1ac62c505661e0ee28d9bc168737aa2e8.


asmecher:
Oops! Yes, absolutely correct. I'll cherry-pick this forward to `master` as well. Thanks, @ptrckb!

Issue #4604, opened 2019-03-14
Undefined variable when retrieving submission by status


ptrckb:
With OJS 3.1.1, using `SubmissionDAO`'s `getByStatus()` triggers an error for an undefined variable in `getSearchWhere()`.

629e7ecc8b25429abe01ea51dc088ec465cdbef3 introduced the helper function but failed to rename one occurence of an old variable (`$search` -> `$phrase`).

This has already been fixed as part of 32d50e1a846d8c9493ec6a1f24661aa0b28371d7 in 'stable-3_1_2' and 'master'.

ptrckb:
pkp-lib: pkp/pkp-lib#4605
ojs: pkp/ojs#2330

asmecher:
Thanks, @ptrckb!

PR #4605, opened 2019-03-14
pkp/pkp-lib#4604 Fix undefined variable

ptrckb:


asmecher:
Merged, thanks -- this doesn't appear to apply to `stable-3_1_2` or `master`, so no need to forward-port.

PR #4620, opened 2019-03-21
fr_FR translations

mpbraendle:


asmecher:
@mpbraendle, I think this also came in against the wrong branch...

PR #4628, opened 2019-03-21
fr_FR translations added

mpbraendle:


asmecher:
That's got it -- thanks, @mpbraendle!

Issue #4533, opened 2019-02-23
[OJS] Duplicated content when submitting an invalid plugin's settings form


jonasraoni:
**Information**
- Browser: Chrome
- Branch: master
- PHP: 7

**Description**
When submitting a form with invalid values:
- Some plugins get the response filled with duplicated content (see attached image). Since the content is exactly what's not enclosed by the ```<form>``` tag, looks like it's a cheap fix
- The field with invalid content lost the value
- The client-side validation allows the form to be submitted with invalid fields

![image](https://user-images.githubusercontent.com/361921/53288513-06cb7880-379a-11e9-8896-08d897325f34.png)

asmecher:
Good spotting, @jonasraoni -- would you mind submitting a PR that moves content into the appropriate element? I remember hitting something like this before with our "required field" footer, but it looks like the same problem is appearing elsewhere.

jonasraoni:
@asmecher Yesterday I've updated some files that use the Ajax handler and had some content out of the form tag. Follow some notes:
About the "required field" that you've mentioned, it gets duplicated in "Ajax forms" when the json returns status = false.

In the profile tab it's possible to send an email to the users. So, I've forced the validation to fail there and as feedback I got an alert with the HTML content of the page (I've replaced it by an error message) + fading error notification... So, my lazy question is: is this alert being used elsewhere? Or everything should be using the notification?!

There are other cases that I need to investigate that will require me to use the software properly or input some fake data =]

asmecher:
The duplication of the "required field" message can be fixed by moving that element into the `<form>` element.

For the profile tab, can you send a draft PR or a gist that shows what you're proposing?

jonasraoni:
Hmm I just saw that the file Handler.js has a method handlejson which shows the alert when the json.status is false. As in this case it isn't needed, I suggest to just break the error handling in a method, then it will be overridable. I'll send a draft PR later =]

Issue #4532, opened 2019-02-23
[OJS] Minor mistake in notification message when removing a plugin


jonasraoni:
Hi,

I tried to find a similar issue, but failed. So, whenever I remove a plugin I get the following notification:
> Plugin "##plugins.XXX.YYY##" successfully deleted

Branch = master
PHP = 7

jonasraoni:
@asmecher I'll add this one to the batch.

PR #4634, opened 2019-03-23
Updated templates + minor corrections.

jonasraoni:
- Templates using the AjaxFormHandler had their content moved into the `<form>` to avoid displaying duplicated content in case there's a submit error (pkp/pkp-lib#4533), also removed the type attribute from the `<script>` tag
- Fixed the notification message when removing a plugin (pkp/pkp-lib#4532)
- PluginHelper.inc.php: replaced an invalid translation key when uploading an invalid file type
- UserGridHandler.inc.php: replaced an invalid error message (the HTML content of the form was being displayed in an alert) by the text "validator.filled"


jonasraoni:
@asmecher

asmecher:
Looks good -- thanks, @jonasraoni!

PR #4643, opened 2019-03-28
pkp/pkp-lib#4642 Remove obsolete breakpoint for workflow sidebar

NateWr:


NateWr:
Style change only. Self-merging.

PR #4651, opened 2019-04-03
pkp/pkp-lib#4618 reset permissions using context default values

ajnyga:
See https://github.com/pkp/pkp-lib/issues/4618

### DIFF
--- a/classes/submission/SubmissionDAO.inc.php
+++ b/classes/submission/SubmissionDAO.inc.php
@@ -517,7 +517,7 @@ abstract class SubmissionDAO extends DAO implements PKPPubIdPluginDAO {
 
 	/**
 	 * Delete the attached licenses of all submissions in a context.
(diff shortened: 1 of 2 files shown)


classes/submission/SubmissionDAO.inc.php
```diff
@@ -517,7 +517,7 @@ function deleteByContextId($contextId) {
 
 	/**
 	 * Delete the attached licenses of all submissions in a context.
-	 * @param $submissionId int
+	 * @param $contextId int
 	 */
 	function deletePermissions($contextId) {
```
NateWr: Looks like this method is no longer used anywhere. Can you remove it?

ajnyga: ok, was unsure if it would be still needed. Removed now.

PR #4669, opened 2019-04-10
Replace call to preg_replace with preg_replace callback - OJS 2 on PHP 7

Ianvdl:
This fixes the call to preg_replace() with a deprecated /e flag on PHP 7 by changing it to a preg_replace_callback() instead.

Please see this forum thread for more details: https://forum.pkp.sfu.ca/t/import-peer-reviews-button-template-field-names-missing/52383

The fix turned out to be a bit simpler than I originally suggested in that thread. It does not appear to have any negative side-effects elsewhere, and the "Import Peer Reviews" button works again.


asmecher:
Thanks, @Ianvdl -- merged and cherry-picked also to `ojs-dev-2_4`.

PR #4677, opened 2019-04-13
add addPolicies method to the GatewayPlugin abstract class

quoideneuf:
Change required to support:
See: https://github.com/pkp/ojs/pull/2360

### DIFF
--- a/classes/plugins/GatewayPlugin.inc.php
+++ b/classes/plugins/GatewayPlugin.inc.php
@@ -64,6 +64,15 @@ abstract class GatewayPlugin extends Plugin {
 		$context = Application::getRequest()->getContext();
 		return is_null($context) ? 0 : $context->getId();
 	}
+
+	/**
+	 * Get policies to the authorization process
+	 * @param $request PKPRequest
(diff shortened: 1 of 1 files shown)


classes/plugins/GatewayPlugin.inc.php
```diff
@@ -64,6 +64,11 @@ function getCurrentContextId() {
 		$context = Application::getRequest()->getContext();
 		return is_null($context) ? 0 : $context->getId();
 	}
-}
-
 
+	/**
+	 * Add Policies to the authorization process
+	 */
+	function addPolicies() {
```
asmecher: If this just returns the policies, it should be called `getPolicies` rather than `addPolicies`, no? Also I'd suggest adding a `@return` self-doc, e.g.:
```
@return array Set of authorization policies
```
Then it would make sense to have the constructor in `GatewayPlugin` apply the policies, rather than leaving it to the subclass. That would reduce the risk that a plugin implements `addPolicies`/`getPolicies` expecting it to take care of the rest.

asmecher:
Thanks, @quoideneuf!

PR #4739, opened 2019-05-08
Problem on Add Reviewer => Enroll Existing User

eniocarboni:
Jlabel plugin need to convert function .size() with the equivalent .length property.
.size() was a jquery removed function from 3.0 (see https://api.jquery.com/size/)
See issue #4737


eniocarboni:
I close the pull request because I realized I had the wrong starting branch where there were already commits.
I try to do it again

PR #4753, opened 2019-05-13
[pkp-lib] #3758 Selectively permit author metadata changes after submission

defstat:


classes/security/UserGroupDAO.inc.php
```diff
@@ -994,6 +1007,42 @@ function getRecommendOnlyGroupIds($contextId, $roleId = null) {
...
+		$result->Close();
+		return $userGroupIds;
+	}
+
+	/**
+	 * Get a list of roles not able to change submissionMetadataEdit permission option.
+	 * @return array
+	 */
+	static function getNotChangeMetadataEditPermissionRoles() {
+		return array(ROLE_ID_MANAGER);
```
asmecher: I'm confused -- shouldn't the editor/manager role be able to change this flag?

classes/stageAssignment/StageAssignmentDAO.inc.php
```diff
@@ -176,16 +210,17 @@ function insertObject($stageAssignment) {
...
+					(?, ?, ?, %s, ?, ?)',
 				$this->datetimeToDB(Core::getCurrentDate())
 			),
 			array(
 				$stageAssignment->getSubmissionId(),
 				$this->nullOrInt($stageAssignment->getUserGroupId()),
 				$this->nullOrInt($stageAssignment->getUserId()),
-				$stageAssignment->getRecommendOnly()?$stageAssignment->getRecommendOnly():0
+				$stageAssignment->getRecommendOnly()?$stageAssignment->getRecommendOnly():0,
+				$stageAssignment->getCanChangeMetadata()?$stageAssignment->getCanChangeMetadata():0
```
asmecher: These should both be `(int)` casts -- not sure why the convoluted syntax :)

classes/security/UserGroupDAO.inc.php
asmecher: ...or does this mean that others cannot disable the flag for editors/managers?

controllers/grid/users/stageParticipant/form/AddParticipantForm.inc.php
```diff
@@ -47,10 +70,34 @@ function getSubmission() {
...
+			$this->_isChangePermitMetadataAllowed = !in_array($stageAssignment->getUserGroupId(), $this->_managerGroupIds);
+			$this->_isChangeRecommentOnlyAllowed = in_array($stageAssignment->getUserGroupId(), $this->_possibleRecommendOnlyUserGroupIds);
+		}
+	}
+
 	/**
 	 * @copydoc Form::fetch()
 	 */
 	function fetch($request, $template = null, $display = false) {
+		$this->initialize($request);
```
asmecher: This is already called in the constructor; is this second call a mistake?

controllers/grid/users/stageParticipant/form/AddParticipantForm.inc.php
```diff
@@ -71,17 +118,47 @@ function fetch($request, $template = null, $display = false) {
...
+
+			$userDao = DAORegistry::getDAO('UserDAO');
+			/** @var $currentUser User */
+			$currentUser = $userDao->getById($stageAssignment->getUserId());
+
+			$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+			/** @var $userGroup UserGroup */
+			$userGroup = $userGroupDao->getById($stageAssignment->getUserGroupId());
+
+			$templateMgr->assign('assignmentId', $this->_assignmentId);
```
asmecher: Use this syntax for long sets of assignments:
```
$templateMgr->assign(array(
    ...
));
```

templates/controllers/grid/settings/roles/form/userGroupForm.tpl
```diff
@@ -16,6 +16,7 @@
 			selfRegistrationRoleIds: {$selfRegistrationRoleIds|@json_encode},
 			recommendOnlyRoleIds: {$recommendOnlyRoleIds|@json_encode},
 			roleForbiddenStagesJSON: {$roleForbiddenStagesJSON},
+			notChangeMetadataEditPermissionRoles: {$notChangeMetadataEditPermissionRoles|@json_encode},
```
asmecher: `@` in this case is subtle :)

xml/schema/rolesAndUserGroups.xml
```diff
@@ -40,6 +40,10 @@
 			<NOTNULL/>
 			<DEFAULT VALUE="0"/>
 		</field>
+		<field name="permit_metadata_edit" type="I1">
+			<NOTNULL/>
+			<DEFAULT VALUE="0"/>
```
asmecher: Have you tested this for upgrades? I think ADODB might fail to add the default, making the `NOT NULL` addition cause an error.

defstat: @asmecher I did test it running both the installation process and the upgrade process, without any issues. Do you think that I should check something else regarding that?

classes/security/UserGroupDAO.inc.php
defstat: Yes that is the idea - that in fact no one can change this setting for the managers - they should always be able to make changes.

controllers/grid/users/stageParticipant/form/AddParticipantForm.inc.php
defstat: Done

xml/schema/rolesAndUserGroups.xml
asmecher: I suspect it'll crop up with PostgreSQL if anywhere; leave it for now and I'll test that before release.

classes/stageAssignment/StageAssignmentDAO.inc.php
defstat: Probably have pasted the syntax from somewhere else.

Changed!

controllers/grid/users/stageParticipant/form/AddParticipantForm.inc.php
defstat: It seems that it's not needed here........

templates/controllers/grid/settings/roles/form/userGroupForm.tpl
defstat: I think I followed exactly the same approach as the `selfRegistrationRoleIds` variable here - (and I am not sure what "subtle" means here :) )

asmecher: It is correct this way -- and needed because of a nasty quirk of Smarty syntax.

xml/schema/rolesAndUserGroups.xml
defstat: I am relying on the Travis tests for PostgreSQL. Don't have it installed on my dev machine. Could install it if necessary.

js/controllers/SiteHandler.js
```diff
@@ -39,6 +39,10 @@
 
 		$('.go').button();
 
+		this.bindGlobal('reloadPage', function () {
```
asmecher: Is this used anywhere? I don't see it.

defstat: You are right @asmecher. This code was used at a previous version of the feature, in order to reload the submission page upon the change of the permitMetadataEdit option for a participant. 

Removed.

PR #4757, opened 2019-05-13
pkp/pkp-lib#4675 Use getter method to retrieve localized data

NateWr:


NateWr:
Wrong base to the branch. Will redo.

PR #4767, opened 2019-05-15
pkp/pkp-lib#4756: Create NMI custom page URL via page/op/path instead of handler

ctgraham:
Resolves pkp/pkp-lib#4756 for master.


ctgraham:
Already cherry picked as https://github.com/pkp/pkp-lib/commit/7f978ddbbd359f142dd77b1ade0cbd39709ee6b5.

PR #4825, opened 2019-06-11
Rename submission and submissiondao classes

ajnyga:
Rename Submission to PKPSubmission & SubmissionDAO to PKPSubmissionDAO


asmecher:
@ajnyga, this would need a PR for OMP as well, or the master branch will be broken; is that something you can work on, or would you like support from the dev team on that?

PR #4843, opened 2019-06-25
Stable 3 1 2

withanage:
@asmecher 

This PR supports https://github.com/pkp/omp/pull/687 and relaxes the restrictions on the length of  the license.


asmecher:
Do you expect the locale keys introduced here to be used in OJS as well? If not, they should be added to the OMP locale files, not the pkp-lib ones.

withanage:
yes,  with the new modifications in the https://github.com/pkp/omp/pull/687 all the keys can be put in  only OMP.
I will  delete this pull request.

Issue #4896, opened 2019-07-04
Ordering of sidebar blocks changed upon visiting the settings form


asmecher:
**Describe the bug**
The order of the sidebar blocks is not preserved when visiting the Appearance settings form. Saving the form without fixing the block order persists the change inadvertently.

**To Reproduce**
1. Create a mix of custom and built-in sidebar blocks.
2. In the appearance settings, intermix the two types of blocks and save the settings. (The saved order should be reflected in the public view.)
3. Revisit the appearance settings form. The block order will not be the same as the public interface reflects.
4. Saving the settings form will cause the new (unwanted) order to be persisted.

**What application are you using?**
 OJS or OMP version 3.1.2

**Additional information**
Reported at https://forum.pkp.sfu.ca/t/sidebar-blocks-re-shuffled-re-ordered-on-their-own-when-accessing-settings-websites-appearance-in-ojs-3-1-2/54479

asmecher:
PR for `stable-3_1_2`: https://github.com/pkp/pkp-lib/pull/4898

asmecher:
Forward-ported to `master`. Closing.

PR #4898, opened 2019-07-04
pkp/pkp-lib#4896 Return loaded plugins in sequence order

asmecher:


mpbraendle:
In some cases I get the following warning in my error log:
PHP Warning:  uasort() expects parameter 1 to be array, null given in lib/pkp/classes/plugins/PluginRegistry.inc.php on line 154
So probably a check on null should be added.

asmecher:
I think I resolved this in `master` but not `stable-3_1_2`. https://github.com/pkp/pkp-lib/commit/a7e05da200f29824fcbc9ec511ad024e63c5d74f should do it.

PR #4928, opened 2019-07-16
PKPRequest::getServerHost() enchanced to deal with multiple host names in a host header.

zozlak:
Depending on the reverse proxy configuration every reverse proxy may append its host name to the HTTP_X_FORWARDED_HOST instead of replacing it (this is e.g. a default Apache behaviour).

In such a case the final header value is a (comma-delimited) list of server names with the original request server name in front, e.g. `clarin.oeaw.ac.at, tokeneditor.apollo.arz.oeaw.ac.at` and the right way to process it is to split it into single host names and take the first value.

### DIFF
--- a/classes/core/PKPRequest.inc.php
+++ b/classes/core/PKPRequest.inc.php
@@ -331,6 +331,8 @@ class PKPRequest {
 				: (isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST']
 				: (isset($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME']
 				: $default));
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @zozlak, that seems sensible -- but can you convert the space-based indents to tabs for consistency with the rest of the codebase?

zozlak:
Done

asmecher:
Thanks -- merged and back-ported to `stable-3_1_2` for the next release from that branch!

asmecher:
This caused test failures -- I adjusted the implementation to rely less on array operations: https://github.com/pkp/pkp-lib/commit/778b738fce3d5dcc50d915db7656ecd62f8f6a94

Issue #4936, opened 2019-07-19
Send database debug messages to the log rather than the browser


asmecher:
The `debug` option in the `[database]` section of `config.inc.php` currently dumps database queries (and some other ADODB debugging information) to the browser. However, this has two big problems:
1) Many of the server interactions are expected to be JSON-encoded, and dumping SQL queries corrupts that; and
2) When the `debug` option is enabled, all queries are dumped to the browser for all users, potentially disrupting usage.

Proposal: direct messages to the server log instead.

asmecher:
Merged!

PR #4937, opened 2019-07-19
pkp/pkp-lib#4936 Log ADODB debug messages (rather than displaying)

asmecher:


asmecher:
@defstat, would you mind reviewing this? It changes the `debug` setting in `config.inc.php` so that messages are sent to the PHP log rather than the browser. (See the issue description at https://github.com/pkp/pkp-lib/issues/4936.)

classes/db/DBConnection.inc.php
```diff
@@ -56,6 +57,16 @@ function __construct() {
...
+	/**
+	 * Log message handler for ADODB.
+	 * @param $msg string
+	 * @param $newline boolean
+	 * @see ADOConnection::outp
+	 */
+	static function logAdodbMessage($msg, $newline=true) {
+		error_log(Registry::get('system.debug.startTime') . ': ' . PKPString::html2text($msg));
+	}
+
```
defstat: @asmecher It works perfectly fine. Could I suggest 2 things:

- Perhaps adding some prefix for those logs, in order to distinguish from other PHP warnings/errors - Something like: `error_log('PKP-Database-Logger ' . Registry::get('system.debug.startTime') . ': ' . PKPString::html2text($msg));` 
- Perhaps we could add some data regarding the caller of the sql. Something like `
$backtrace = debug_backtrace(2);
error_log(print_r($backtrace, true) . '-> ' . Registry::get('system.debug.startTime') . ': ' . PKPString::html2text($msg));` (even though that could result to a rather extensive log. Maybe we could traverse on the `$backtrace` in order to display less, and more useful info)

asmecher: Thanks, @defstat! I agree with the first; let's come back to the second if there's trouble locating the SQL. I find grepping for a piece of the SQL works pretty well.

asmecher:
Merged to `stable-3_1_2` and `master`!

Issue #4953, opened 2019-07-29
lib/pkp/tools/installPluginVersion.php install data issue when running from command line


defstat:
**Describe the bug**
When trying to execute `lib/pkp/tools/installPluginVersion.php` on a plugin that overrides the `getInstallDataFile()` function, a fatal exception is being thrown because of the `$installer->dataXMLParser` being null (the exception occurs at `https://github.com/pkp/pkp-lib/blob/master/classes/plugins/Plugin.inc.php#L543`).

When the tool script is executed, the `preInstall()` function (`https://github.com/pkp/pkp-lib/blob/master/classes/install/Installer.inc.php#L130`) is not executed, causing the `dataXMLParser` not being initialized. 

**Workaround**
The web tool to initiate the upgrade process for OJS is working as expected.

**What application are you using?**
 That may correlate to all 3.x.x versions

**Additional information**
`PHP Fatal error:  Uncaught Error: Call to a member function parseData() on null in /lib/pkp/classes/plugins/Plugin.inc.php:515
Stack trace:
#0 /lib/pkp/tools/installPluginVersion.php(104): Plugin->installData('Installer::post...', Array)
#1 /lib/pkp/tools/installPluginVersion.php(112): InstallPluginVersionTool->execute()
#2 {main}
  thrown in /lib/pkp/classes/plugins/Plugin.inc.php on line 515`

asmecher:
@defstat, try this PR: https://github.com/pkp/pkp-lib/pull/4968

asmecher:
Merged and cherry-picked forward to `master` -- thanks for the review, @defstat!

PR #4968, opened 2019-08-01
pkp/pkp-lib#4953 Ensure an XML parser object is available for data installation

asmecher:


classes/plugins/Plugin.inc.php
```diff
@@ -512,7 +512,9 @@ function installData($hookName, $args) {
 
 		// Install all data files.
 		foreach($dataFiles as $dataFile) {
-			$sql = $installer->dataXMLParser->parseData($dataFile);
+			$dataXMLParser = new DBDataXMLParser();
```
defstat: I would have proposed this line and the one after that [515-516] to go above the `foreach`, I don't think that it is needed to be instantiated for each `dataFile`

asmecher: Yes, thanks! But let me review your other comment before I do that.

asmecher:
> ...in order to follow the same norm as...

Yes, thanks, that's a better solution. I don't like intervening this way with member objects but we're already doing it, so better to stay consistent. I've force-pushed up the change to use what you've suggested instead -- can you take a quick look at it?

classes/plugins/Plugin.inc.php
asmecher: The other comment obsoleted this one.

defstat:
I think we could merge it. Thanks!

Issue #4984, opened 2019-08-11
[OJS] Notifications/tasks popover being displayed unexpectedly


jonasraoni:
**Describe the bug**
The list of notifications is displayed unexpectedly after being opened and closed.

**To Reproduce**
1. Click on the `Tasks` link to display the notifications.
2. Click again to close it.
3. Click anywhere in the page and the notifications will be displayed again.

**What application are you using?**
 OJS: /master
Browser: Chrome

asmecher:
This got merged -- thanks, @jonasraoni!

PR #4986, opened 2019-08-11
pkp/pkp-lib#4984 Fixes the tasks popover

jonasraoni:
Added an option to clear the "callWhenClickOutside" callbacks and fixed the tasks popover.

Should I open a PR against OJS to update the /js/pkp.min.js?

### DIFF
--- a/js/pages/header/TasksHandler.js
+++ b/js/pages/header/TasksHandler.js
@@ -78,6 +78,12 @@
 			setTimeout(this.callbackWrapper(this.setPopoverSize_), 500);
 		} else {
 			$popover.css('height', '').css('overflow-y', '');
+			if (event) {
+				this.trigger('callWhenClickOutside', [{
+					container: $header,
+					clear: true
+				}]);
+			}
 		}
 	};
 
--- a/js/controllers/SiteHandler.js
+++ b/js/controllers/SiteHandler.js
@@ -472,12 +472,13 @@
 			return;
 		}
 
-		if (eventParams.callback === undefined) {
-			return;
-		}
-
 		var id = eventParams.container.attr('id');
-		this.outsideClickChecks_[id] = eventParams;
+
+		if (eventParams.clear) {
+			delete this.outsideClickChecks_[id];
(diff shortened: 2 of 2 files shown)


js/pages/header/TasksHandler.js
```diff
@@ -62,22 +62,27 @@
 	 */
 	$.pkp.pages.header.TasksHandler.prototype.appendToggleIndicator_ =
 			function(callingElement, event) {
-
 		var $header = this.getHtmlElement(),
-				$popover = $header.find('#notificationsPopover'),
-				$toggle = $header.find('#notificationsToggle');
+			$popover = $header.find('#notificationsPopover'),
```
asmecher: The JS linter will complain about this -- it expects a double-tab indent when continuing a previous line. Try `bash lib/pkp/tools/buildjs.sh -n` to run the linter locally.

asmecher: (Or you can create a PR that only updates the submodule with the specially-formulated commit name referring to your branch, i.e.: `pkp/pkp-lib#4984 Submodule update ##jonasraoni/i4984-fix-tasks-popover##` -- that'll cause Travis to run tests including a JS linting round. But that's a pretty painful way to work through JS linting errors.)

asmecher:
Don't worry about recompiling the JS -- that's part of the build process, so we'll make sure we get to it before packaging up the next release.

js/pages/header/TasksHandler.js
jonasraoni: @asmecher, I've updated the PR.
Just to confirm, it's expected that the `buildjs.sh` is emitting `1 error(s), 247 warning(s)`, right?

jonasraoni: I replaced the compiler.jar by the version 20120917 and looks like it fixed the problem =]

jonasraoni:
@asmecher If there's nothing more missing, feel free to merge =]

asmecher:
Thanks, @jonasraoni!

PR #5001, opened 2019-08-16
Added possibility to inject environment variables into config.inc.php for Docker

alvinchevolleaux:
When creating containerised applications using a tool such as Docker, it is desirable to inject configuration using environment variables due to the transient nature of containers and images.

This additional check will look for uppercased strings allowing underscores also and check if an environment variable exists with the same name. This occurs after the `defined` check so it will not affect the existing constants check.

### DIFF
--- a/classes/config/ConfigParser.inc.php
+++ b/classes/config/ConfigParser.inc.php
@@ -103,6 +103,8 @@ class ConfigParser {
 					} else if (defined($value)) {
 						// The value matches a named constant
 						$value = constant($value);
+					} else if (preg_match('/^[A-Z_]+$/', $value) && getenv($value) !== false) {
+						$value = getenv($value);
 					}
 				}
 


alvinchevolleaux:
Adding a usage example to clarify:

```
;;;;;;;;;;;;;;;;;;;;;
; Database Settings ;
;;;;;;;;;;;;;;;;;;;;;

[database]

driver = mysqli
host = DB_HOSTNAME
username = DB_USER
password = DB_PASSWORD
name = DB_NAME
; Set the non-standard port and/or socket, if used
; port = 3306
; unix_socket = /var/run/mysqld/mysqld.sock

; Enable persistent connections
persistent = Off

; Enable database debug output (very verbose!)
debug = Off
```
When environment variables DB_HOSTNAME, DB_USER, etc are present, the above snippet of config.inc.php will use these environment variables. It has worked in my own deployment of OJS on Amazon's ECS service.

Currently, I need to initialise the pkp-lib submodule and overwrite the ConfigParser.inc.php class with my custom version. This is fine for now but native environment variable support would be great. Saving a static text file of configuration has security implications and goes against the way containers are designed.

asmecher:
Thanks, @alvinchevolleaux! @defstat, could you have a look at this one?

classes/config/ConfigParser.inc.php
```diff
@@ -103,6 +103,8 @@ static function &readConfig($file) {
 					} else if (defined($value)) {
 						// The value matches a named constant
 						$value = constant($value);
+					} else if (preg_match('/^[A-Z_]+$/', $value) && getenv($value) !== false) {
+						$value = getenv($value);
 					}
 				}
```
defstat: @alvinchevolleaux I believe that the environment variables are strings only, so the `$value` will get a string value if, for example we want to give it a boolean value. Have you tried that scenario?

defstat: Another thing that we may consider here could be a way to easily distinguish the environment variables that we are using inside the `config.inc.php`. We could do that by adding a pattern for those like, for example, `environment(DB_USER)`

PR #5026, opened 2019-08-29
Complete manager.xml es_ES translations

gonetil:
Translations based on manager.xml file that comes with OJS 3.1.2.1 . There seem to be some texts removed from master branch, not sure if that's correct


asmecher:
Thanks, @gonetil! Yes, there will be some removals -- there was a settings form rewrite that reorganized things somewhat.

PR #5050, opened 2019-09-09
make HookRegistry ignore non-callable callbacks

prechelt:
If display_errors=on, a broken plugin (such as OpenAIREPlugin
in my case) that registers something that cannot be called,
will provoke a message that breaks OJS's AJAX pages (because
many JSON responses will become ill-formed).
This patch ignores hook callbacks that are !is_callable
and so makes OJS robust against such cases.

### DIFF
--- a/classes/plugins/HookRegistry.inc.php
+++ b/classes/plugins/HookRegistry.inc.php
@@ -104,6 +104,8 @@ class HookRegistry {
 			ksort($hooks[$hookName], SORT_NUMERIC);
 			foreach ($hooks[$hookName] as $priority => $hookList) {
 				foreach ($hookList as $hook) {
+					if (!is_callable($hook))
+						continue;  // else broken plugins break OJS if display_errors = On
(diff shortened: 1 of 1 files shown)


asmecher:
@prechelt, the behaviour you describe will affect pretty much every aspect of OJS -- `display_errors` will break AJAX requests. Rather than merging point fixes for this, it might be worth considering a broader fix -- or even removing `display_errors` so users are less tempted to lean on it. One side-effect of this fix will be that users will no longer get a warning if an uncallable hook is registered, and that warning might lead them to fix a code problem that would be otherwise very hard to find.

prechelt:
I see. Removing `diplay_errors` is a bit of a pity, but the much sounder solution overall. I guess that's the way to go. I'll consider this PR done.

asmecher:
Thanks, @prechelt! Thinking about this, other options are available -- e.g. a custom error handler that delivers warnings/errors in JSON format for requests that specify a JSON response. However, I think using the logs will be more reliable.

PR #5142, opened 2019-10-10
pkp/pkp-lib#4273: Move h1 out of breadcrumbs nav

ctgraham:
Moves the page-level `h1` out of the breadcrumbs `nav` for pkp/pkp-lib#4273.

### DIFF
--- a/templates/frontend/components/breadcrumbs.tpl
+++ b/templates/frontend/components/breadcrumbs.tpl
@@ -23,6 +23,7 @@
 			<span class="separator">{translate key="navigation.breadcrumbSeparator"}</span>
 		</li>
 		<li class="current">
+			<a href="{$currentUrl}" aria-current="page">
 			{capture name="currentTitleH1"}
 				{if $currentTitleKey}
 					{translate key=$currentTitleKey}
@@ -31,6 +32,7 @@
 				{/if}
 			{/capture}
 			{$smarty.capture.currentTitleH1}
+			</a>
 		</li>
 	</ol>
 </nav>
--- a/templates/frontend/components/breadcrumbs.tpl
+++ b/templates/frontend/components/breadcrumbs.tpl
@@ -23,13 +23,16 @@
 			<span class="separator">{translate key="navigation.breadcrumbSeparator"}</span>
 		</li>
 		<li class="current">
-			<h1>
+			{capture name="currentTitleH1"}
 				{if $currentTitleKey}
 					{translate key=$currentTitleKey}
 				{else}
 					{$currentTitle|escape}
 				{/if}
-			</h1>
+			{/capture}
+			{$smarty.capture.currentTitleH1}
 		</li>
 	</ol>
 </nav>
+<h1 class="pageCurrentTitle">{$smarty.capture.currentTitleH1}</h1>
+


asmecher:
@israelcefrin, would you mind reviewing this? (1/2)

templates/frontend/components/breadcrumbs.tpl
```diff
@@ -23,13 +23,16 @@
...
 		</li>
 		<li class="current">
-			<h1>
+			{capture name="currentTitleH1"}
 				{if $currentTitleKey}
 					{translate key=$currentTitleKey}
 				{else}
 					{$currentTitle|escape}
 				{/if}
-			</h1>
```
israelcefrin: Removing heading from breadcrumb trail is perfect. Hoewever, given we are fixing this issue for headings on breadcrumbs and our accessibility report has noted it, could be possible also add a link to the current page on this element with parameter `aria-current="page"`? 
This would improve accessibility while screen readers should announce this link is the current page in the navigation.

Reference: https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html#ex_label

ctgraham:
I've added this to the PR.

templates/frontend/components/breadcrumbs.tpl
```diff
@@ -23,13 +23,18 @@
 			<span class="separator">{translate key="navigation.breadcrumbSeparator"}</span>
 		</li>
 		<li class="current">
-			<h1>
+			<a href="{url}" aria-current="page">
```
asmecher: I don't think off the top of my head that `{url}` will be reliable; I think `{$currentUrl}` will, though. (Sorry, I don't remember the details of `{url}`.)

asmecher: Anything coming from a `POST` request (or AJAX request) will cause this to behave badly.

ctgraham: Swapped `{url}` for `{$currentUrl}`.

asmecher: Thanks! But this still needs consideration for `POST` and AJAX requests -- I'd recommend perhaps making this just a `span` rather than an anchor in cases like that.

ctgraham: What would be an example of an AJAX request that would call the breadcrumbs template?

I tested the POST by checking out the `$currentUrl` in the login process.  It points to the POSTed URL, which is accurate.  Is this problematic for something like `login/signIn`?  The more awkward bit might be that it includes the CSRF token in a GET search result, which presumably wouldn't be reusable.

asmecher: I can't think of any AJAX examples off the top of my head; perhaps that one will be OK, but I'll see if anything comes to mind and follow up here if not. For POST URLs, you can use the template variable `requiresFormRequest` to check.

ctgraham: @asmecher, I'm still not seeing why having a hyperlink to the POST'd URL is a bad thing within the breadcrumbs.  Both use-cases I can think of (/user/register and /login/signIn) work as GET links.  These and other use-cases will improve the more RESTful we become (e.g. GET /login and POST /login instead of a dedicated POST URI of "signIn").

That said, @israelcefrin, what would it mean to screen readers if for consistency we moved the `aria-current` from the link to the `li`, where we already have a parallel classing of "current"?  We could then easily drop the `a` tag if undesirable, maintaining the semantics of the "current" breadcrumb regardless of whether there is a link.

israelcefrin: @ctgraham the paralel class for current in LI element is only a visual resource that can be formated via CSS. It doesn't affect screen readers.
However,  if we move the aria-current to a non-link element it will not be clear, for user reconition, what it is the real behaviour and/or state of that element. For instance:
```
<ul>
  <li><a href="...">Home</a></li>
  <li><a href="...">Issue 1</a></li>
  <li><a href="..." aria-current="true">My article</a></li>
</ul>
```
A screen reader will announce:
Link Home. Lin Issue 1. Link My Article. Current link.

If we move the `aria-current` to the `LI `element (dropping off the `A HREF` element), screen readers   won't announce the current link state, because navigating with `TAB `in a keyboard will not be possible to recognize the `LI `element neither it's `aria-current` state. 

Just to illustrate, I've been playing around some quick tests for the following code:
```
<ul>
  <li><a href="...">Home</a></li>
  <li><a href="...">Issue 1</a></li>
  <li aria-current="page">My article</li>
</ul>
```
Narrator screen reader on windows10 didn't announce `My article` item neither the current page state. Because `aria-current` works with a set of elements to mark one as the current. `UL > LI` is a  different element from `UL >LI > A`, what makes not comparable to check and set a single `UL > LI` fully announceable.
Maybe a solution would be include a visually hidden text in `LI` text element,but it would add a text var to translate also.i.e.:
```
<ul>
  <li><a href="...">Home</a></li>
  <li><a href="...">Issue 1</a></li>
  <li><span class="hidden_class_name"> Current page:</span>My article</li>
</ul>
```
I am not a fan of this solution, but it is an valid accessibility technique , semantically speaking. However, as I said it will affect even locale/translation files, i.e., more files to edit.

asmecher: @ctgraham:
> These and other use-cases will improve the more RESTful we become (e.g. GET /login and POST /login instead of a dedicated POST URI of "signIn").

OK, that's a good motivation, I think. We've already added `isPost` checks on the form submission code, so the worst of bad behaviour is presented. Please go ahead!

israelcefrin: @asmecher may I merge then?

asmecher: Yes, and forward-port to `master`, if you're satisfied with the rest of it.

PR #5143, opened 2019-10-11
fr_CA translation

mhvezina:
fr_CA translation reviewed and updated

These are the main changes:
I have :
- reordered message keys based on the en_US equivalent files
- added feminisation of titles/roles (e.g. : "utilisateur-trice", "auteur-e") (but I have kept common terms ("droit d'auteur", "compte utilisateur", "données utilisateur"))
- standardized the use of "téléversement" vs "téléchargement" (and related verb "téléverser" vs "télécharger") for "upload" and "download"
- standardized the use of « some text » instead of "some text"
- standardized the spacing (that is one blank before "?", ":", "!" (better yet would be non-breaking space (&nbsp;) but the translator plugin seems to systematically remove them...)
- removed a few tupples
- removed obsolete message keys (i.e. those keys that didn't have any equivalent in the en_US files)
- fixed a few typos, fixed what seemed to be like wrong translations (sometimes difficult to get the context though...)
- completed what was missing (either new message keys since last translation was done and/or modified message keys)


asmecher:
Thanks, @mhvezina, merged and forward-ported to `master`: ceef9fdb49470eb056e88d0930777897faeadddc

Issue #5146, opened 2019-10-11
Missing hook call for submissionfilesuploadform::execute


ctgraham:
**Describe the bug**
`SubmissionFilesUploadForm::execute()` never calls its parent, so it can never be called from a plugin hook.
https://github.com/pkp/pkp-lib/blob/d0d3534fecfdedbf1c4a1dabf46d6bd840d317e7/controllers/wizard/fileUpload/form/SubmissionFilesUploadForm.inc.php#L125-L182

**To Reproduce**
1. Add a hook for `submissionfilesuploadform::execute` into a plugin
2. Break on the supplied function
3. Upload a submission file
4. Never enter the function

**What application are you using?**
 OJS 3.1.2

**Additional information**
For this hook to be truly useful, the caller needs to be able to inspect and manipulate the submission file which was created within the `execute()` method, so we need to expose that with an accesssor.

asmecher:
Thanks, @ctgraham! Committed and forward-ported to `master`.

PR #5147, opened 2019-10-11
pkp/pkp-lib#5146: Enable execute hook for SubmissionsFilesUploadForm

ctgraham:
Resolves pkp/pkp-lib#5146

### DIFF
--- a/classes/form/Form.inc.php
+++ b/classes/form/Form.inc.php
@@ -303,15 +303,17 @@ class Form {
 	/**
 	 * Execute the form's action.
 	 * (Note that it is assumed that the form has already been validated.)
+	 * @param mixed $functionArgs,... Arguments from the caller to be passed to the hook consumer
+	 * @return mixed Result from the consumer to be passed to the caller.  Send a true-ish result if you want the caller to do something with the return value.
 	 */
-	function execute() {
+	function execute(...$functionArgs) {
 		// Call hooks based on the calling entity, assuming
 		// this method is only called by a subclass. Results
 		// in hook calls named e.g. "papergalleyform::execute"
 		// Note that class and function names are always lower
 		// case.
 		$returner = null;
-		HookRegistry::call(strtolower_codesafe(get_class($this) . '::execute'), array_merge(array($this), func_get_args(), array(&$returner)));
+		HookRegistry::call(strtolower_codesafe(get_class($this) . '::execute'), array_merge(array($this), $functionArgs, array(&$returner)));
 		return $returner;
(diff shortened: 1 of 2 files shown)


controllers/wizard/fileUpload/form/SubmissionFilesUploadForm.inc.php
```diff
@@ -177,10 +179,23 @@ function execute() {
 				'username' => $user->getUsername()
 			)
 		);
+		
+		$this->_submissionFile = $submissionFile;
```
asmecher: I think these changes might be a little less convoluted if you passed `$submissionFile` as a parameter to `parent::execute`; where the hook is called, `func_get_args` is used to pass any incoming parameters into the hook.

ctgraham:
I support being less convoluted.

classes/form/Form.inc.php
```diff
@@ -303,8 +303,10 @@ function validate($callHooks = true) {
 	/**
 	 * Execute the form's action.
 	 * (Note that it is assumed that the form has already been validated.)
+	 * @param mixed $functionArgs,... Arguments from the caller to be passed to the hook consumer
+	 * @return mixed Result from the consumer to be passed to the caller.  Send a true-ish result if you want the caller to do something with the return value.
 	 */
-	function execute() {
+	function execute(...$functionArgs) {
```
asmecher: Points for the splat operator! But I'd suggest also using that in the `HookRegistry::call` function, for consistency.

ctgraham: Now variable is used for consistency.

controllers/wizard/fileUpload/form/SubmissionFilesUploadForm.inc.php
ctgraham: Deconvoluted.

controllers/wizard/fileUpload/form/SubmissionFilesUploadForm.inc.php
```diff
@@ -177,6 +177,11 @@ function execute() {
 				'username' => $user->getUsername()
 			)
 		);
+		
+		$hookResult = parent::execute($submissionFile);
```
asmecher: I would love to replace this pattern with something else -- but doing some experimentation with variadic parameter lists and references, it's a little inflexible (either all variadic parameters must be passed by reference using `&...$listName` or all must be passed by value without the `&`). I'll give this some thought but for the moment this will stick around.

PR #5201, opened 2019-10-19
pkp/pkp-lib#4919-fix Rewrite Slim routing using PATH_INFO

asmecher:


asmecher:
Not merging, as this is too big a change for a build. Will merge instead into `master` for the OJS 3.2 release.

Issue #5156, opened 2019-10-14
Missing gzip and egrep cli options in config.TEMPLATE.inc.php


ctgraham:
**Describe the bug**
Options were added in 2.4.x to config.inc.php for "egrep" and "gzip" to support access file processing for Usage Statistics.

https://github.com/pkp/ojs/blame/ojs-stable-2_4_8/config.TEMPLATE.inc.php#L462-L466

The [`CopyAccessLogFileTool`](https://github.com/pkp/pkp-lib/blob/d0d3534fecfdedbf1c4a1dabf46d6bd840d317e7/tools/copyAccessLogFileTool.php#L53-L58) and [`FileLoader`](https://github.com/pkp/pkp-lib/blob/d0d3534fecfdedbf1c4a1dabf46d6bd840d317e7/classes/task/FileLoader.inc.php#L305) exist in 3.x, as does [`FileManager::compressFile` and `FileManager::decompressFile`](https://github.com/pkp/pkp-lib/blob/d0d3534fecfdedbf1c4a1dabf46d6bd840d317e7/classes/file/FileManager.inc.php#L565-L625).  All of these depend on the "egrep" or "gzip" options.

**To Reproduce**
1. Install OJS based on config.TEMPLATE.inc.php
2. Call `FileManager::decompressFile()`
3. Fails with `admin.error.executingUtil`

**What application are you using?**
 OJS 3.1.2

**Additional information**
Instead of specifying these CLI tools in config.inc.php we should prefer a built-in, like `PharData`, `ZipArchive`, and `preg_match`.

PR #5211, opened 2019-10-22
pkp/pkp-lib#5156: enable compression/decompression routines in FileManger without system call

ctgraham:
pkp/pkp-lib#5156 add extension options for compression/decompression in favor of gzip system calls.

### DIFF
--- a/classes/file/FileManager.inc.php
+++ b/classes/file/FileManager.inc.php
@@ -543,21 +543,97 @@ class FileManager {
 
 	/**
 	 * Decompress passed gziped file.
-	 * @param $filePath string
-	 * @param $errorMsg string
-	 * @return boolean|string
+	 * @param $filePath string The compressed file's filename
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean|string False on failure, or the decompressed filename on success.
 	 */
 	function decompressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
+			$newFilePath = preg_replace('/[.]tgz$/i', '.tar.gz', $filePath);
+			$newFilePath = preg_replace('/[.]gz$/i', '', $newFilePath);
+			if ($newFilePath !== $filePath) {
+				$gz = gzopen($filePath, 'rb');
+				if ($gz) {
+					$file = fopen($newFilePath, 'wb');
+					if ($file) {
+						while (!gzeof($gz)) {
+							$data = gzread($gz, $chunk);
+							fwrite($file, $data);
+						}
+						fclose($file);
+						gzclose($gz);
+						unlink($filePath);
+						return $newFilePath;
+					}
+					fclose($file);
+				}
+			}
+		}
 		return $this->_executeGzip($filePath, true, $errorMsg);
+	}
+
+	
+	/**
+	 * Extract files from tgz or zip
+	 * @param $filePath string The compressed archive filename
+	 * @param $destination string The location to which to extract the contents
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean Success
+	 */
+	function extractFiles($filePath, $destination, &$errorMsg) {
+		$success = false;
+		$errors = array();
+		if (extension_loaded('Phar')) {
+			try {
+				$zip = new PharData($filePath);
+				$zip->extractTo($destination);
+				$success = true;
+			} catch (Exception $e) {
+				$errors[] = $e->getMessage();
+			}
+		}
+		if (!$success && extension_loaded('zip')) {
+			$zip = new ZipArchive();
+			if ($zip->open($filePath) === true && $zip->extractTo($destination) === true) {
+				$zip->close();
+				$success = true;
+			} else {
+				$errors[] = 'ZipArchive extraction failed.';
+			}
+		}
+		if ($errors) {
+			$errorMsg = implode("\n", $errors);
+		}
+		return $success;
 	}
 
 	/**
 	 * Compress passed file.
 	 * @param $filePath string The file to be compressed.
-	 * @param $errorMsg string
-	 * @return boolean|string
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean|string False on failure, or the compressed filename on success.
 	 */
 	function compressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
+			$gzFilePath = $filePath.'.gz';
+			$file = fopen($filePath, 'rb');
+			if ($file) {
+				$gz = gzopen($gzFilePath, 'wb');
+				if ($gz) {
+					while (!feof($file)) {
+						$data = fread($file, $chunk);
+						gzwrite($gz, $data);
+					}
+					gzclose($gz);
+					fclose($file);
+					unlink($filePath);
+					return $gzFilePath;
+				}
+				fclose($file);
+			}
+		}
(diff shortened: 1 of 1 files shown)


classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
 	 */
 	function decompressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
+			$newFilePath = preg_replace('/[.]tgz$/i', '.tar.gz', $filePath);
+			$newFilePath = preg_replace('/[.]gz$/i', '', $newFilePath);
+			if ($newFilePath !== $filePath) {
+				$gz = gzopen($filePath, 'rb');
+				if ($gz) {
+					$file = fopen($newFilePath, 'wb');
```
asmecher: Feel free to declare this out-of-scope, but now that we're generating our own new filename, it's making me antsy that we're overwriting without warning. Generally I would rather us either specify a target filename as a parameter to this function, or use a function like https://www.php.net/manual/en/function.tempnam.php to safely generate a new file (returning its name as the return value, as we currently do).

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
+	 * Extract files from tgz or zip
+	 * @param $filePath string The compressed archive filename
+	 * @param $destination string The location to which to extract the contents
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean Success
+	 */
+	function extractFiles($filePath, $destination, &$errorMsg) {
+		$success = false;
+		$errors = array();
+		if (extension_loaded('Phar')) {
```
asmecher: Do we have any `Phar` use cases?

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
 	 * @param $filePath string The file to be compressed.
-	 * @param $errorMsg string
-	 * @return boolean|string
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean|string False on failure, or the compressed filename on success.
 	 */
 	function compressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
+			$gzFilePath = $filePath.'.gz';
```
asmecher: (As above, re: potentially clobbering existing files without warning)

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
-	 * @param $filePath string
-	 * @param $errorMsg string
-	 * @return boolean|string
+	 * @param $filePath string The compressed file's filename
+	 * @param $errorMsg string Any fatal error encountered (modified by reference)
+	 * @return boolean|string False on failure, or the decompressed filename on success.
 	 */
 	function decompressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
```
asmecher: A defined constant would be better at the top of the file, rather than hard-coding a chunk size at a couple points in the code.

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
+			if ($file) {
+				$gz = gzopen($gzFilePath, 'wb');
+				if ($gz) {
+					while (!feof($file)) {
+						$data = fread($file, $chunk);
+						gzwrite($gz, $data);
+					}
+					gzclose($gz);
+					fclose($file);
+					unlink($filePath);
```
asmecher: I know this is command-line `gzip`'s default behavior, but this makes me nervy the same way as the overwrite does; I'd rather see us separate the creation of a compressed file from the deletion of the original. Open to other perspectives, of course!

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
 	 */
 	function compressFile($filePath, &$errorMsg) {
+		if (extension_loaded('Zlib')) {
+			$chunk = 1024*1024;
+			$gzFilePath = $filePath.'.gz';
+			$file = fopen($filePath, 'rb');
+			if ($file) {
+				$gz = gzopen($gzFilePath, 'wb');
+				if ($gz) {
+					while (!feof($file)) {
```
asmecher: `feof` will return true on an error condition; accordingly, this code will currently nuke the original file without warning if an error occurs partway through.

classes/file/FileManager.inc.php
```diff
@@ -543,21 +543,97 @@ function parseFileExtension($fileName) {
...
+						while (!gzeof($gz)) {
+							$data = gzread($gz, $chunk);
+							fwrite($file, $data);
+						}
+						fclose($file);
+						gzclose($gz);
+						unlink($filePath);
+						return $newFilePath;
+					}
+					fclose($file);
```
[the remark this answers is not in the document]
ctgraham: Oops.  Will fix.

classes/file/FileManager.inc.php
ctgraham: Agreed, this is terrible, but it is also the legacy function.  But the legacy function is also localized enough that we can fix it.  So, I guess will fix as well.

classes/file/FileManager.inc.php
ctgraham: This is the only one I actually care about in all of this mess.  I'm not actually using `Phar` except for `PharData` which has convenient functions to decompress either Zip or TGZ archives in one handler.

classes/file/FileManager.inc.php
ctgraham: *mutters about legacy function*

classes/file/FileManager.inc.php
ctgraham: True.  Will fix.

classes/file/FileManager.inc.php
ctgraham: *continues muttering*

classes/file/FileManager.inc.php
ctgraham: True.  Will fix.

classes/file/FileManager.inc.php
asmecher: Ah, good, that makes sense. What do you think about a 3rd-party polyfill e.g. https://packagist.org/packages/alchemy/zippy (at a glance)?

asmecher: [appreciates the muttering sympathetically]

asmecher: Check the Packagist listing for each -- I often use the popularity numbers as a proxy for "how much outcry will there be if this thing goes away without an easy path forward".

classes/file/FileManager.inc.php
ctgraham: So, if I fix this, I am either:
* Changing the method signature to something outrageous `...pressFile(string|array $filepathOrOldNewPathArray, array &$errorMsg)` or `...pressFile(string $oldFilePath, array &$errorMsg, string $newFilePath = null)`
* Or, changing to an incompatible but sane method signature in a minor or patch release.

Both of these are terrible.

*Muttering intensifies to grumbling.*

asmecher: Maybe cleaning this up would be best left to the `master` branch.

ctgraham: I will happily pay you tomorrow for my hamburger today.

I'll fix my own goofs in this PR and write up a new PR for master (hopefully).

PR #5223, opened 2019-10-24
fr_CA translation updated

mhvezina:
Mainly consist of a few fixes + translation of new keys + removing of obsolete ones (comparison made to the en_US corresponding files)

I wasn't sure if I should have changed the PO-Revision-Dates or not....

There are still a few keys left to translate in submission.po and manager.po (pkp-lib) but I understand it is OMP stuff (?)... I will leave that to do and might put effort in the translation of the plugins locale instead...


asmecher:
Thanks, @mhvezina! I believe you have the honour of being the first to work directly with the new PO format.

PR #5224, opened 2019-10-24
fr_CA translation updated

mhvezina:
A few fixes (same fixes that have been also changed to the corresponding `*.po` files in the `master `branch)


asmecher:
Thanks, @mhvezina!

PR #5238, opened 2019-10-31
defstat/pln#44 append unique id to deposit grid id

jnugent:
Appends a unique id to the div element, fixing the double-bind problem, and solves defstat/pln#44


asmecher:
Thanks, @jnugent, that's perfect!

Issue #5250, opened 2019-11-05
CLI TOOL - Importing users is crashing for existing usernames


israelcefrin:
**Describe the bug**
It is not possible import users via CLI `tools/importExport.php` if the username already exists in Database.

**To Reproduce**
1. Create a XML sample of users from a journal and keep some original usernames.
2. In CLI run `php tools/importExport.php UserImportExportPlugin import exported-file-of-users.xml [journal name] 
`
3. It will stop the process on the first existing username on DB. 
4. See the error, i.e.:
`DB Error: Duplicate entry 'johndoe' for key 'users_username'</h1>ojs2: DB Error: Duplicate entry 'johndoe' for key 'users_username'`

**What application are you using?**
 OJS 3.1.2

**Additional information**
It is a multi-journal import in different stages. Some users are present on journals that are beind currently imported. And we can filter it out easily.

asmecher:
(Just documenting the diagnosis here that we arrived at in Slack -- the users that encounter problems are disabled in the import XML. The `getByUsername` and `getUserByEmail` fetches in the XML import code are currently specified not to look for disabled user accounts; this should be changed.)

PR #5259, opened 2019-11-06
Fixing username checking to import users via CLI

israelcefrin:
This patch fixes the flag checking for duplicated usernames for  disallowed users described in this issue #5250


asmecher:
Thanks, @israelcefrin! Will also back-port to `stable-3_1_2`.

Issue #5285, opened 2019-11-19
[OJS] Stop collecting user IP to email and event logs


ajnyga:
OJS/OMP will store users IP to the email and event logs. This is not required and we should make sure that we do not collect personal data that we do not use.

https://github.com/pkp/pkp-lib/blob/master/xml/schema/log.xml
https://github.com/pkp/pkp-lib/blob/master/classes/log/EmailLogDAO.inc.php
https://github.com/pkp/pkp-lib/blob/master/classes/log/EventLogDAO.inc.php
https://github.com/pkp/pkp-lib/blob/master/classes/log/EmailLogEntry.inc.php
https://github.com/pkp/pkp-lib/blob/master/classes/log/EventLogEntry.inc.php

asmecher:
Merged, thanks!

PR #5286, opened 2019-11-19
pkp/pkp-lib#5285 stop collecting ip in email and event logs

ajnyga:
https://github.com/pkp/pkp-lib/issues/5285


asmecher:
Thanks, @ajnyga! I think we may have some trouble with ADODB not knowing how to drop a `NOTNULL` constraint, but we'll catch that during upgrade testing.

Issue #5305, opened 2019-11-28
Undefined variable contextId when searching for a reviewer


ptrckb:
6bacfd7422601aef382e521dfb34a345b8b22182 restricted the reviewer search to a context but forgot the actual selection of the context.

OJS 3.1.2 only, `master` is ok (a095fdd81ee6ff75f7700ed2e5ad859f92cac99c).

ptrckb:
OJS: pkp/ojs#2541 (submodule update)
pkp-lib: pkp/pkp-lib#5306

PR #5306, opened 2019-11-28
pkp/pkp-lib#5305 Fix undefined context id when searching reviewer.

ptrckb:


api/v1/users/PKPUserHandler.inc.php
```diff
@@ -263,6 +265,7 @@ private function _buildListRequestParams($slimRequest) {
 	private function _buildReviewerListRequestParams($slimRequest) {
 
 		$returnParams = $this->_buildListRequestParams($slimRequest);
+		$contextId = $returnParams['contextId'];
```
asmecher: I think this warning could instead be resolved by removing https://github.com/pkp/pkp-lib/pull/5306/files#diff-22f1b14d85eac6760b01a79df3eef5ffR294 (and this new line)... I don't see any way `$returnParams['contextId']` can be overridden (per https://github.com/pkp/pkp-lib/pull/5306/files#diff-22f1b14d85eac6760b01a79df3eef5ffR293) and `$contextId` isn't otherwise used.

api/v1/users/PKPUserHandler.inc.php
```diff
@@ -249,6 +249,8 @@ private function _buildListRequestParams($slimRequest) {
 			}
 		}
 
+		$returnParams['contextId'] = $context->getId();
```
asmecher: I think this API endpoint can be called site-wide, which would cause a fatal error on non-object here (as `$context` would be `null`).

PR #5309, opened 2019-11-28
Correction to editor.po pt_BR

carolinatanigushi:
Corrected string that said "reviews" instead of "revisions".


asmecher:
Thanks, @carolinatanigushi!

PR #5311, opened 2019-12-01
Allow apps to extend the workflow settings

ajnyga:


asmecher:
Smarty should try to find the template first in the app, then in the shared library, so just putting a `workflow.tpl` in the app's template directory should do the same thing without requiring changes to the PHP code...

ajnyga:
Sure, but if in PPS I want to add new forms the workflow settings here https://github.com/pkp/pkp-lib/blob/master/pages/management/ManagementHandler.inc.php#L185-L245 how can you extend that now from within an application?

asmecher:
Yep, absolutely right... a hook is an option but this is less spaghetti. I'd love to have a better mechanism for this but I'm not sure what it is.

Issue #5322, opened 2019-12-03
Fatal error when viewing editorial report


asmecher:
From https://forum.pkp.sfu.ca/t/upgrade-to-ojs-3-1-2-3-statistics-editorial-generate-php-fatal-error/57332/3:
```
[Tue Dec 03 16:20:47.554305 2019] [php7:error] [pid 15283] [client 10.25.8.2:54480] PHP Fatal error: Uncaught TypeError: Argument 1 passed to PKP\Statistics\SubmissionStatistics::__construct() must be an instance of PKP\Statistics\object, instance of stdClass given, called in /var/www/ojs/lib/pkp/classes/services/EditorialStatisticsService.inc.php on line 75 and defined in /var/www/ojs/lib/pkp/classes/statistics/SubmissionStatistics.inc.php:27\nStack trace:\n#0 /var/www/ojs/lib/pkp/classes/services/EditorialStatisticsService.inc.php(75): PKP\Statistics\SubmissionStatistics->__construct(Object(stdClass))\n#1 /var/www/ojs/lib/pkp/pages/stats/PKPStatsHandler.inc.php(225): PKP\Services\EditorialStatisticsService->getSubmissionStatistics(1)\n#2 /var/www/ojs/lib/pkp/classes/core/PKPRouter.inc.php(390): PKPStatsHandler->editorialReport(Array, Object(Request))\n#3 /var/www/ojs/lib/pkp/classes/core/PKPPageRouter.inc.php(231): PKPRouter->_authorizeInitializeAndCallRequest(Array, Object(Request), Arr /var/www/ojs/lib/pkp/classes/statistics/SubmissionStatistics.inc.php on line 27, referer: http://xxx.xxx.ca/index.php/cjur/management/tools
```
It looks like the type annotation needs to be namespaced properly (and that the user's version of PHP treats this mismatch as a Fatal Error).

PR #5325, opened 2019-12-04
pkp/pkp-lib#5322 Fixed fatal error under PHP < 7.2.

jonasraoni:


asmecher:
The `@param` self-doc needs updating too...

Issue #5319, opened 2019-12-03
PHP undefined variable warning


asmecher:
When using the editorial stats report:
```
PHP Notice:  Undefined variable: dateFormat in /home/asmecher/git/ojs-stable-3_1_2/lib/pkp/classes/services/queryBuilders/SubmissionStatisticsQueryBuilder.inc.php on line 165
```
The variable appears never to be defined before it's used at https://github.com/pkp/pkp-lib/blob/ef4343313446746a6fa8b8ee4c6a397403d35ea3/classes/services/queryBuilders/SubmissionStatisticsQueryBuilder.inc.php#L165.

asmecher:
@jonasraoni, could you take a look at this one?

PR #5330, opened 2019-12-05
pkp/pkp-lib#5319 Fixed missing date filter.

jonasraoni:
That's how this `$dateFormat` piece was expected to work, which should change the value of published submissions =]

My overall opinion about the Illuminate\Database is that it works great for simple queries, but since it's missing too many constructions (I'm not sure it exists something better in PHP), we end up using too much "raw" and the bound parameters are lost in the process, which requires the use of [merge bindings](https://laravel.com/api/5.7/Illuminate/Database/Query/Builder.html#method_mergeBindings) (a shortcut for `array_merge_recursive` that end up producing unexpected results)/reorder the parameters manually... If only they had added named parameters, everything could be cleaner 🤔

### DIFF
--- a/classes/services/queryBuilders/SubmissionStatisticsQueryBuilder.inc.php
+++ b/classes/services/queryBuilders/SubmissionStatisticsQueryBuilder.inc.php
@@ -78,6 +78,15 @@ class SubmissionStatisticsQueryBuilder extends BaseQueryBuilder {
 
 		$capsule = $this->capsule;
 
+		$publishedSubmissionsQuery = $capsule
+			->table('published_submissions AS ps')
+			->where('ps.submission_id', Capsule::raw('ranged_s.submission_id'))
+			->whereRaw(
+				'ps.date_published BETWEEN COALESCE(?, ps.date_published) AND COALESCE(?, ps.date_published)',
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @jonasraoni! Named bindings are referenced here: https://laravel.com/docs/5.2/database#running-queries Not sure if this is passed down to the underlying database, or abstracted in Illuminate\Database?

jonasraoni:
@asmecher I've seen this before, but it's limited. You can't use the binding more than once (`:name > 1 and :name  < 2` will fail), you can't set/update the value through the name later,  and in the end they are converted to positional bindings, which brings the pain of keeping the parameters in the right order.

I'll check/confirm again later... Maybe I can come up with a simple patch/workaround =]

jonasraoni:
@asmecher, just a follow-up up here... It's possible to enable `PDO::ATTR_EMULATE_PREPARES` through the Illuminate's configuration, then the shared named bindings will work... BUT it has breaking counter effects (just tested a bit and found one), so it's better to keep it disabled...
I'd like to dig deeper and try to fix it in the upstream, but I have other things to do 😁

Issue #5132, opened 2019-10-08
[OJS] Update a internal id with native xml importer


ChristianeE:
We want to use the native xml importer of OJS 3.1.2 to finish a submission for publishing.
For this case it would be fine if the already existing submission would be pushed forward to this step. So we want to update the internal id.
In the source code the option to upgrade the internal id is not supported yet. 
Is there a possibility to add this?

retostauffer:
Hy all

Unfortunately no answer so far. We spent quite some work in automatizing monitoring/publishing our submissions, the xml importer is the one crucial thing missing :).

As @ChristianeE wrote we would like to **publish** new articles using the native xml plugin. We are using a custom OJS2 version for this since years to automatize this step which makes everyone's life easier and would love to have this feature in OJS3 as well.

Before starting to customize OJS3 - does anyone have a hint where to start? Would help us a lot.

Thanks in advance!
Reto

asmecher:
Hi @retostauffer! There are stubs in place for this work -- the XSD has support for an `advice="update"` attribute on the `id` element, so the XML itself supports e.g. `<submission id="123" advice="update">`, but the PHP code doesn't support it yet. The stub in the code is in `lib/pkp/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php` in `parseIdentifier`:
```
$advice = $element->getAttribute('advice');
switch ($element->getAttribute('type')) {
    case 'internal':
        // "update" advice not supported yet.
        assert(!$advice || $advice == 'ignore');
        break;
    ...
}
```
You'll see in the `handleElement` function in the same class where the `$submissionDao->insertObject` function is called. You'll need to juggle the handling of the `id` element so it can be parsed before the `insertObject` call, and if it's an already-existing submission, you'll need to fetch the existing one rather than calling `insertObject`. Hope that helps!

retostauffer:
Hy @asmecher 
Thank you very much for your fast response. Supports our first idea - should not be too difficult, hope we can come back with custom solution, soon :).
All best and thank you for your support,
Reto & Christiane

PR #5331, opened 2019-12-06
Enable native XML importer to update submissions

retostauffer:
Dear @asmecher 

Thanks again for your advice regarding issue/suggestion https://github.com/pkp/pkp-lib/issues/5132. We were able to implement a "draft" to enable the NativeXML Plugin to update (`advice = "update"`) existing submissions using the internal submission ID. The pull request contains attempt to get this feature running.

However, on the way we encountered some additional problems. E.g., **updating** an **already published** article results in a php **error** (rather than having a useful error message on the UX).

And, much more critical, **on the way** (must be somewhere in either `insertObject()`, `populateObject()`, or `updateObject()` on the `SubmissionDAO` object(s)) the library **deletes the entry** from the `submissions` **database** table. Later on, it tries to update/insert "supplementary files" (from the XML file). If a specific **file type** (genre) does **not exist** php runs into an error - which stops execution - and the **submission is lost forever** :).

It would be great if you could have a quick look on the current state - and maybe give some advice/hints how/where we could try to proceed (to save some time).

Thanks a lot and all best from Austria,
@ChristianeE and Reto

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
@@ -66,9 +66,30 @@ class NativeXmlSubmissionFilter extends NativeImportFilter {
 		$deployment = $this->getDeployment();
 		$context = $deployment->getContext();
 
-		// Create and insert the submission (ID needed for other entities)
+		// Searching for ID
+		$internalSubmissionId = NULL;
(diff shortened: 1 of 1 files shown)


asmecher:
@retostauffer, I don't think there's any code in the import/export toolset that would delete a submission, so I wonder if it's not e.g. updating its ID to something else so that you're not able to find it?

PR #5443, opened 2020-01-27
Fix turkish language selection

gabriele-h:
As explained in https://forum.pkp.sfu.ca/t/12764

We came across this issue a few days ago. The whole ojs instance stopped working when activating tr_TR and clicking on the selection "Türkçe". Since this is a rather grave outcome, we think this should simply be fixed with a oneliner for everybody.

### DIFF
--- a/classes/i18n/PKPLocale.inc.php
+++ b/classes/i18n/PKPLocale.inc.php
@@ -174,6 +174,10 @@ class PKPLocale {
 				default: assert(false);
 			}
 		}
+		// Fix known php issue for turkish language selection, see also:
+		// https://www.sobstel.org/blog/php-call-to-undefined-method-on-tr-tr-locale/
+		if ($locale == 'tr_TR') setlocale(LC_CTYPE, 'en_US');
+
 	}
 
 	/**


asmecher:
@gabriele-h, what version of OJS (or OMP) are you using, and what version of PHP?

gabriele-h:
@asmecher thanks for the quick response!
Versions: OJS 3.1.2.1, PHP 7.2.25

I do feel a bit weird about not opening an issue before doing the pull request. Please be patient with me as I am new to OJS.

asmecher:
@gabriele-h, I'm just trying to determine whether this is an issue with our code, or something in the underlying PHP release. Do you get the same error message in your log as the other user reported (`Call to undefined method User::getId()`)?

gabriele-h:
It didn't look to me as if the issue in PHP was fixed or was going to be fixed at any point. This quick fix might be a dirty hack, but digging into how to handle PHP's strtolower() and strtoupper() and the according locale madness seems a bit over the top?

Regarding the error message I would have to repeat the tests. In our current setup we have no access to the PHP logs, as it is a shared environment. With debug info enabled we had this in the output on the page: "Class 'Smarty_Internal_Undefined'" - which sounds like some string conversion went awry in a plugin we have activated (the PHP bug report did mention class names breaking because of the strtolower/strtoupper issues).

asmecher:
@gabriele-h, the reference to `Smraty_Internal_Undefined` helps -- this appears to be a problem in the Smarty library we use for template management. See https://github.com/smarty-php/smarty/issues/155 for details. I'll add a comment over there.

gabriele-h:
Okay. So let's abandon this merge request? Thank you so much for your patience and for referencing the smarty issue!

asmecher:
@gabriele-h, please keep it as a local work-around until a Smarty update gets released including a fix for it, and thanks for the heads-up!

PR #5480, opened 2020-02-06
Fix reassign DOIs

ajnyga:
https://github.com/pkp/pkp-lib/issues/5473


asmecher:
Sorry, @ajnyga, I missed that you'd sent in a PR on this issue. Did you see the solution I posted over on #5473? I'm interested in the renamed settings here; did you find a glitch around those?

ajnyga:
`pub-id::` is the right variation and should be used, the other one is just wrong. Using the other version will cause a problem, do not remember in detail what that was.

asmecher:
Ah, yes, I corrected that without even noticing when reinstating the previous ADODB::replace-based call: b8b81cad2f787e2f5ac2d2fd98c21097398cf66a
Thanks, @ajnyga!

PR #5485, opened 2020-02-07
Tests correction

withanage:
@asmecher  and @NateWr 

Hi Alec and Nate,

As I was running the cypress tests, I found in the master branch, these elements have changed and corrected.
Please ignore this PR, if this was corrected in other place.


cypress/support/commands.js
```diff
@@ -36,7 +36,7 @@ Cypress.Commands.add('install', function() {
 	cy.get('input[id=additionalLocales-fr_CA').check();
 
 	// Complete the installation
-	cy.get('button[id^=submitFormButton-]').click();
+	cy.get('button[id^=submitFormButton-]', {timeout: 90000}).click();
```
asmecher: @withanage, this won't extend the timeout on the `click()` call, it only extends the timeout on the `cy.get` call. The default timeout should be long enough for the tests to pass (and Travis gets through it OK). Are tests breaking somewhere on Travis, or on your local setup?

cypress/support/commands.js
```diff
@@ -329,8 +329,8 @@ Cypress.Commands.add('createUser', user => {
 	if (!('roles' in user)) user.roles = [];
 	cy.get('div[id=userGridContainer] a:contains("Add User")').click();
 	cy.wait(2000); // Avoid occasional glitches with given name field
-	cy.get('input[id^="givenName-en_US"]').type(user.givenName, {delay: 0});
-	cy.get('input[id^="familyName-en_US"]').type(user.familyName, {delay: 0});
+	cy.get('input[name="givenName[en_US]"').type(user.givenName, {delay: 0});
+	cy.get('input[name="familyName[en_US]"').type(user.familyName, {delay: 0});
```
asmecher: These lines are equivalent (the elements can be matched either way), so I don't think this changes anything...

cypress/support/commands.js
```diff
@@ -341,7 +341,7 @@ Cypress.Commands.add('createUser', user => {
 	cy.get('select[name=country]').select(user.country);
 	cy.contains('More User Details').click();
 	cy.get('span:contains("Less User Details"):visible');
-	cy.get('input[id^="affiliation-en_US"]').type(user.affiliation, {delay: 0});
+	cy.get('input[name="affiliation[en_US]"').type(user.affiliation, {delay: 0});
```
asmecher: Likewise, I don't think this change will cause different behaviour...

cypress/support/commands.js
withanage: @asmecher  sorry, I  have overiddden it and corrected it again. this can be ignored.

cypress/support/commands.js
withanage: in the ids, I did not see the language attribute, therfore added it to the name.

 
```html
<input type="text" maxlength="255" required="true" validation="required" class="field text required" name="givenName[en_US]" value="" id="givenName-5e3d86cd81092" aria-required="true">
```

asmecher: Very strange -- that's different than I'm seeing. Do you have two languages installed and enabled (i.e. was the installation performed using the installation test, or did you perform it yourself)?

withanage: I was  using the  cypress GUI and  I think, I may have missed some updates somehow, when merging from the pkp/pkp-lib.
Just did a fresh installation and a worked fine.

withanage:
Closing,  cause this is PR is not needed.

asmecher:
Not a problem, it's a new toolset for us; feel free to catch me on Slack if you have any questions about this!

Issue #5493, opened 2020-02-09
[OHS] Installation error


jonasraoni:
Describe the bug
When trying to install (or maybe upgrade) on PHP 7 + MySQL 5.7 it was failing with an error.

What application are you using?
OHS 2.3
PHP 7.1

jonasraoni:
PR: https://github.com/pkp/pkp-lib/pull/5494

jonasraoni:
@asmecher is the harvester going to be archived? If yes, I'll close the issues I've opened against it.

Also, maybe the OPS should be pinned [here](https://github.com/pkp).

PR #5494, opened 2020-02-09
pkp/pkp-lib#5493 Fixed installation error

jonasraoni:
Once the reconnection is completed, some objects are left abandoned with the old and disconnected connection.

### DIFF
--- a/classes/install/PKPInstall.inc.php
+++ b/classes/install/PKPInstall.inc.php
@@ -174,24 +174,7 @@ class PKPInstall extends Installer {
 			return false;
 		}
 
-		// Re-connect to the created database
-		$this->dbconn->disconnect();
-
-		$conn = new DBConnection(
-			$this->getParam('databaseDriver'),
-			$this->getParam('databaseHost'),
-			$this->getParam('databaseUsername'),
-			$this->getParam('databasePassword'),
-			$this->getParam('databaseName'),
-			true,
-			$this->getParam('connectionCharset') == '' ? false : $this->getParam('connectionCharset')
-		);
-
-		DBConnection::getInstance($conn);
-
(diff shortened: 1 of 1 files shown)


asmecher:
@jonasraoni, that same code is distributed with OJS and OMP and I don't recall seeing a problem with it; can you describe the behaviour in more detail?

jonasraoni:
Hi @asmecher! FYI I'm just resurrecting some issues that I got merged into another repository.

Well, I'll have to dig this one up in order to get more details. I just remember that some classes kept a copy of the disconnected database  (so perhaps just a local problem in OHS, that could be fixed by removing the local copies) and failed.

asmecher:
@jonasraoni, I suspect it's the default setting of `persistent` to `On` in `config.TEMPLATE.inc.php` (and thus new installations of OHS). For OJS and OMP, we've turned that to `Off` and I suspect the same thing in OJS would solve the problem you encountered.

jonasraoni:
@asmecher, I disabled the persistent option, but got the same output:
`Errors occurred during installation:
A database error has occurred: Duplicate entry '' for key 'email_templates_default_email_key'`

Without investigating much, I just see two options, the fix that I proposed or to locate all classes which may keep an internal copy of the connection and force them to use the singleton.

I think that in general is better to avoid side-effects (in this case, breaking the connection).

asmecher:
@jonasraoni, I guess I'm still confused about why this is appearing in OHS in this case when that code has been widely distributed e.g. with OJS without changes for a very long time... do you know what's behaving differently in your setup?

jonasraoni:
No idea why it didn't fail before, I just can investigate... Here I'm using PHP 7.1 + MySQL 8 + Apache running inside Docker :)

PR #5524, opened 2020-02-19
option to reference files in export xml

quoideneuf:


plugins/importexport/native/filter/NativeExportFilter.inc.php
```diff
@@ -107,6 +107,14 @@ function createOptionalNode($doc, $parentNode, $name, $value) {
 		$parentNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), $name, htmlspecialchars($value, ENT_COMPAT, 'UTF-8')));
 		return $node;
 	}
+
+	/**
+	 * Set xml filtering opts
+	 * @param $opts array
+	 */
+	function setOpts($opts) {
+		$this->opts = $opts;
```
asmecher: This should have a property declaration in the class.

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.inc.php
```diff
@@ -121,12 +120,33 @@ function createSubmissionFileNode($doc, $submissionFile) {
...
 
 		$submissionFileNode->appendChild($revisionNode);
-
 		// Embed the file contents
-		$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($submissionFile->getFilePath())));
-		$embedNode->setAttribute('encoding', 'base64');
-		$revisionNode->appendChild($embedNode);
 
+		if (array_key_exists('no-embed', $this->opts)) {
+			$hrefNode = $doc->createElementNS($deployment->getNamespace(), 'href', );
```
asmecher: I think that trailing comma is a syntax error...

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.inc.php
```diff
@@ -121,12 +120,33 @@ function createSubmissionFileNode($doc, $submissionFile) {
...
+			if (array_key_exists('use-file-urls', $this->opts)) {
+				import('lib.pkp.classes.core.PKPComponentRouter');
+				$router = new PKPComponentRouter();
+				$router->_contextList = [1 => $context];
```
asmecher: `_context` is a private member (if not declared that way, by the `_` prefix as a convention); it shouldn't be accessed directly, and the hard-coded context ID of 1 will probably cause trouble. Is this to facilitate command-line access where there is no request?

quoideneuf: Yes, trying to use the PKPRouter::url function to generate the download url, but it seems like it isn't intended to be used outside of a web server context. Is there a better way to do this? I was basically trying to reverse engineer the same link that appears on the workflow tab in the application...

asmecher: Have a look for calls to `$dispatcher->url` e.g. in `lib/pkp/classes/task/ReviewReminder.inc.php`.

quoideneuf: Made this change and it seems to work:
```
    $router = new PKPComponentRouter();
-    $router->_contextList = [1 => $context];
+    $app = Application::getApplication();
+    $router->setApplication($app);
```

asmecher: What about:
```
$dispatcher = $application->getDispatcher();
$submissionReviewUrl = $dispatcher->url(Application::get()->getRequest(), ROUTE_COMPONENT, $context->getPath(), ...);
```

quoideneuf: Yes, that seems to work too:
https://github.com/pkp/pkp-lib/pull/5524/files#diff-8c95f7fe60b5739a2bed1f1c4443a005R128-R134

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.inc.php
```diff
@@ -121,12 +120,29 @@ function createSubmissionFileNode($doc, $submissionFile) {
...
 
+		if (array_key_exists('no-embed', $this->opts)) {
+			$hrefNode = $doc->createElementNS($deployment->getNamespace(), 'href');
+			if (array_key_exists('use-file-urls', $this->opts)) {
+				$app = Application::getApplication();
+				$dispatcher = $app->getDispatcher();
+				$params = ["fileId" => $submissionFile->getFileId(),
+						   "revision" => $submissionFile->getRevision(),
+						   "submissionId" => $submissionFile->getSubmissionId(),
+						   "stageId" => 1];
```
asmecher: `stageId` is the workflow stage ID. You can get this using `SubmissionFileDAO::getWorkflowStageId`.

quoideneuf: Ok, I think this and the other issues are all resolved now.

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.inc.php
quoideneuf: fixed now: https://github.com/pkp/pkp-lib/pull/5524/files#diff-8c95f7fe60b5739a2bed1f1c4443a005R126

PR #5677, opened 2020-03-27
pkp/pkp-lib#4042 record new rounds with own decision type

ajnyga:
https://github.com/pkp/pkp-lib/issues/4042


locale/en_US/submission.po
```diff
@@ -1475,6 +1478,9 @@ msgstr "Revision required."
 msgid "notification.type.editorDecisionResubmit"
 msgstr "Resubmit for review."
 
+msgid "notification.type.editorDecisionNewRound"
+msgstr "New review round."
```
NateWr: > A new round of reviews was initiated.

ajnyga: fixed

ajnyga: Probably no need to restarts tests for this?

NateWr: Nope.

PR #5779, opened 2020-04-23
pkp/pkp-lib#5177 Added locale string for issue title

thinkbulecount2:
Addressing @israelcefrin's comments in here: https://github.com/pkp/pkp-lib/issues/5177


thinkbulecount2:
Not needed, change is done on OJS side.

Issue #5781, opened 2020-04-23
Suppress SMTP authentication in PHPMailer


HongPong:
**Describe the problem you would like to solve**
The client SMTP mailer had a certificate that could not be tuned to the expected host.

**Describe the solution you'd like**
An optional switch in the configuration file to adjust the PHPMailer parameters to skip checking the certificate of the contacted SMTP server.

**Who is asking for this feature?**
Server administrator/developer

**Additional information**
I was able to get this working and will attach the PR.

HongPong:
Both PR's are required to get the expected behavior here. Thank you for considering the issue!

asmecher:
PRs:
- https://github.com/pkp/ojs/pull/2720
- https://github.com/pkp/pkp-lib/pull/5782

Minor comments left on each; thanks, @HongPong. Please drop a note here when you've had a chance to go over them!

HongPong:
PRs updated thanks for the review :)

asmecher:
Merged for release in OJS 3.2.1 -- thanks for contributing!

PR #5782, opened 2020-04-23
suppress the SMTP check by adding SMTPOptions parameter to PHPMailer instance #5781

HongPong:
This alters the PHPMailer parameters if the new configuration variable is set "On". The configuration variable is added to the OJS config file in the other PR.

Issue description:
https://github.com/pkp/pkp-lib/issues/5781


asmecher:
@HongPong, could you convert your tabs from spaces to tab characters?

HongPong:
done, let me know if anything else should be adjusted. appreciate the quick feedback.

Issue #5843, opened 2020-05-08
User object in payment records for past users null


ewhanson:
Payments tab would not load when pulling up payment record linked to a nonexistent user. 

`$user` can be null for payments if those users were deleted. Payment records for an old user were not properly being updated when their account was merged with a new user.

Fixes:
- [x] Catch nonexistent users in `PaymentGridCellProvider` and label them for JMs via new locale key.
- [x] Properly transfer over payment records to new user

ewhanson:
Hi @asmecher, could you review these pull requests that address this issue? Thanks.

PRs:
#5845
#5846 (stable)
pkp/ojs#2737
pkp/ojs#2738 (stable)

asmecher:
Thanks, @ewhanson! I've committed the `master` branch parts of this and will leave it for the 3.3 release, unless it's important to the hosting group to have it in `stable-3_2_0` or `stable-3_2_1`.

PR #5845, opened 2020-05-08
pkp/pkp-lib#5843 User object in payment records for past users null

ewhanson:


asmecher:
Manually merged by cherry-picking.

Issue #5856, opened 2020-05-11
Review Form not getting the correct title and description


defstat:
**Describe the bug**
Check https://forum.pkp.sfu.ca/t/ojs-3-2-0-3-review-forms-not-displaying-title-and-description-instructions/60424 for a bug description

defstat:
@asmecher 

**PRs**
[PKP-LIB]: #5857
[OJS]: pkp/ojs#2740
[OMP]: pkp/omp#810
[OPS]: pkp/ops#21

asmecher:
Merged for 3.2.1 -- thanks, @defstat!

PR #5857, opened 2020-05-11
[pkp-lib] #5856 Fix for missing title and description when reviewform is displayed

defstat:


controllers/grid/settings/reviewForms/form/PreviewReviewForm.inc.php
```diff
@@ -56,7 +56,7 @@ function initData() {
 			$request = Application::get()->getRequest();
 			$context = $request->getContext();
 			$reviewFormDao = DAORegistry::getDAO('ReviewFormDAO'); /* @var $reviewFormDao ReviewFormDAO */
-			$reviewForm = $reviewFormDao->getById($this->reviewFormId, Application::getContextAssocType(), $context->getId());
+			$reviewForm = $reviewFormDao->getById($this->reviewFormId, Application::getContextAssocType(), $context->getId()); /** @var ReviewForm $reviewForm  */
```
asmecher, line 59: Just curious, is this annotation necessary for your IDE, or is for PHP Scrutinizer or something else? I've found that the `DAORegistry::getDAO` annotation (for example) is necessary because the function only specifies `@return DAO` but the calling code wants to execute functions defined only in subclasses. In this case, `ReviewFormDAO::getById` declares `@return ReviewForm`, so I wouldn't expect it to be a problem.

defstat: @asmecher you are right. 

The problem with that line was the line 58. My editor (Visual Studio Code) supports PHPDoc annotation for those thinks (I may be able to extend it somehow but have not found the time to do that yet). The annotation `/* @var $reviewFormDao ReviewFormDAO */` is not getting parsed as a `variable annotation` but just as another comment. (it needs the `/**` in front instead of `/*`). So the `getById` couldn't get resolved as `returns ReviewForm`. 

In any case, the `/** @var ReviewForm $reviewForm  */` is unnecessary and should be removed.

asmecher: Ah! Great, let's standardize on `/**` -- everything seems to accept that. I have some work to do on annotations for Scrutinizer and hopefully it'll improve things for Visual Studio Code as well.

defstat: If you like I can "fix" things if I encounter them as well. Or batch-fix them through the code. 
Right now I have found 2 things: 
- the `/*` annotation.
- And a warning about the classes usage in Services: if you try to use `Publication` when annotating, it warns that it could be `\Publication` that I need to consider instead. It's clever enough to work with both, but it warns anyway

asmecher: Either way is fine, and these improvements are good to make to the `master` branch! I tend to do this kind of work when I'm not able to focus on bigger things, but that's just me.

Issue #5061, opened 2019-09-11
[OJS 3.1.2-1] Crossref module: Language of all metadata EXCEPT author names depends on submission language (author names depend on locale chosen)


Ph-We:
**Describe the bug**
This one was introduced with the 'multilingual names' feature. Language of all metadata exported always depended on submission language. However, author names being part of metadata depend on locale chosen (by the journal manager at the moment). So one part of metadata may be exported in one language, while the other in the other one. Which is incorrect.

**To Reproduce**
1. In a multilingual journal go to Tools --> Import/Export --> CrossRef XML Export Plugin
2. Choose any article with multilingual names, then choose any locale different from the submission language of this article.
3. Click 'Download XML'
4. The language of all metadata except the author names will be the same as the submission language. The author names will be exported in the language of the locale chosen.

**What application are you using?**
 OJS 3.1.2-1

asmecher:
This needs a little more attention from someone who knows CrossRef's name model -- deferring for 3.1.2-3.

defstat:
It seems that the appropriate addition to the deposited XML for such cases is the http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#alt-name tag. 

I will propose a PR adding that to the authors tag, and try to validate that against Crossref's latest schema.

defstat:
The tags that can be used here are
- http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#contributors
- http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#person_name
- http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#language.atts
- http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#alt-name
- http://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#name

@asmecher 

**PRs**
[pkp-lib]: #5874
[OJS]: pkp/ojs#2750
[OMP]: pkp/omp#814
[OPS]: pkp/ops#25

asmecher:
Thanks, @defstat, I just added one minor comment but otherwise these are ready to go. Please go ahead when you're ready.

defstat:
@asmecher I have merged [pkp-lib], [OJS] and [OMP]. 
Don't have permission to merge on OPS (tagging @NateWr @ajnyga)

asmecher:
[...]

PR #5874, opened 2020-05-14
[pkp-lib] #5061 Fix default locale priority for getGivenName and getFamilyName

defstat:


classes/identity/Identity.inc.php
```diff
@@ -116,14 +116,14 @@ function setGivenName($givenName, $locale) {
...
-		return $this->getLocalizedData(IDENTITY_SETTING_GIVENNAME);
+	function getLocalizedGivenName($defaultLocale = null) {
+		return $this->getLocalizedData(IDENTITY_SETTING_GIVENNAME, $defaultLocale);
 	}
 
 	/**
 	 * Get family name.
 	 * @param $locale string
-	 * @return string
+	 * @return string|array
```
asmecher, line 126: This will always return an `array`, won't it?

defstat: @asmecher, not if you provide a `locale` other than `null`, I think.

asmecher: Ah, of course :)

PR #5947, opened 2020-06-01
Obfuscating email addresses

arturbonnett:
Very important to reduce the number of spam by removing the exposure.
Native "Smarty" feature: https://www.smarty.net/docs/en/tips.obfuscating.email.tpl


asmecher:
Merged -- thanks, @arturbonnett!

PR #5965, opened 2020-06-04
Updating some Arabic locale files before June, 15

vormia:


asmecher:
Merging and cherry-picking to stable-3_2_1 -- thanks, @vormia!

Issue #5964, opened 2020-06-03
Notification when author creates a version


diegoabadan:
**Describe the problem you would like to solve**
When an author creates a version of your preprint, there is no notification in "Tasks"

**Describe the solution you'd like**
A notification would be created, allowing the Editor / Moderator to click and view a new version of submission.

**Who is asking for this feature?**
Server Admins

**Additional information**
Scielo wants this feature on its preprint server. We thought that it could ideally be part of OPS (and OJS), and not part of a plugin.

If they agree, we can send a PR.

asmecher:
Yes, I think a PR would be welcome on this, thanks!

asmecher:
https://github.com/pkp/pkp-lib/pull/6020 is a patch for the `stable-3_2_1` branch -- I'll leave that open for the moment.

@JhonathanLepidus, reviewing this again, I think it might need tweaking around who receives the notifications. I would expect the users assigned to the submission to receive notifications, but it looks like all managerial users are currently coded to be notified. For large installations that receive a lot of submissions, it might cause too much noise for managing editors.

JhonathanLepidus:
@asmecher , we added another commit that modifies the notification to be sent only to the users assigned to the submission. You could check it in the PR #6020

asmecher:
Thanks, @JhonathanLepidus, could you send that change in as a PR against `master`?

JhonathanLepidus:
I just sent it: #6025 

You could merge only the last commit (that modifies the notification) because the other commits were already merged in a previous PR (#5983 )

asmecher:
Thanks, I merged the last commit to `master`: https://github.com/pkp/pkp-lib/commit/df5d56b870b56e6d0ce0c8d88a6f413a67eba16b

asmecher:
The `master` branch should now have all the relevant changes in, but `stable-3_2_1` does not (I'm just finishing the release of `3.2.1-0`). I'll leave https://github.com/pkp/pkp-lib/pull/6020 for `3.2.1-1` (and have moved this issue to that milestone). Thanks, @JhonathanLepidus!

asmecher:
Merged https://github.com/pkp/pkp-lib/pull/6020 to `stable-3_2_1` for release in 3.2.1-2 -- thanks, all!

PR #6020, opened 2020-06-18
pkp/pkp-lib#5964 Adds a new notification for when a new version of a submission is created

JhonathanLepidus:
We are sending this PR so it can be present in 3.2.1-x and other 3.2.x future releases. This can be helpful for people until a 3.3 release arrives.


asmecher:
@JhonathanLepidus, this looks very clean! Could you fill out a matching [github issue](https://github.com/pkp/pkp-lib/issues/new?assignees=&labels=&template=feature_request.md&title=Create+a+notification+when+a+new+version+is+created) (if there isn't one already)? I'd like to discuss delivery there too, specifically...

- Who should receive the notification? Currently (from this PR) it looks like all managerial roles assigned to the journal/press/server, but I think it might better be sent to editorial users assigned to the current submission. Plus, we might consider other people, e.g. authors, if that's desirable.

Link to this PR from the new issue entry so everything is kept together.

We're likely going to release OJS 3.2.1 tomorrow, so this probably won't make it in, but as you've noted it'll be ready for users who need to patch it while we're working on the next release.

Thanks!

diegoabadan:
Alec, is related to this issue:
https://github.com/pkp/pkp-lib/issues/5964

Our wish is not to need to apply the patch now, then 3.2.1, then 3.2.1-1 etc. So, if we can have it before 3.3 it would be nice, even if I can't get into tomorrow's launch specifically.

asmecher:
Thanks, @diegoabadan, it looks like I closed that one accidentally.

diegoabadan:
Thanks, @asmecher !

asmecher:
Cherry-picked the last commit to `master` -- thanks, @JhonathanLepidus!

PR #6108, opened 2020-07-21
pkp/pkp-lib#6085 Forwarded the context's locale where needed to avoid…

jonasraoni:
… delivering the report with the wrong translation.


asmecher:
Thanks, @jonasraoni! Merged.

Issue #2993, opened 2017-10-31
reCaptcha not working from China


ajnyga:
Hi,

I do not have first hand knowledge of this, but one of our journals is reporting that an author from China is unable to register to OJS because the reCaptcha is not visible.

This is probably due to local internet policies and noticed elsewhere as well: https://github.com/stellar/stellar-client/issues/1162

Should OJS have a secondary system for preventing spam that would not have the limitations of reCaptcha?

mfelczak:
Confirming that some of our hosted clients have also run into this issue.

asmecher:
@mfelczak, do you have any preference from that client? I suspect we'd need to look at a built-in service again, though I really don't like maintaining our own. I suspect there's a third-party library we could use.

mfelczak:
Hi @asmecher, I've checked with the client and they couldn't provide any alternatives. Digging around a bit, BotDetect Captcha looks promising for a possible 3rd-party integration: https://captcha.com/php-captcha.html

asmecher:
https://github.com/gregwar/captcha also appears to be heavily used, and it's in Composer with very few dependencies...

jmacgreg:
Hi all, see also https://pad.foebud.org/google-alternatives for some alternatives. The Recaptcha-specific section is quoted below: 

>     http://textcaptcha.com/
> 
>     https://www.scorchsoft.com/blog/recaptcha-alternative-honeypot-spam-prevention/ honeypot forms
> 
>     checking IP addresses against RBLs
> 
>     https://captcha.com/ - BotDetect, a CAPTCHA implementation
> 
>     https://akismet.com/ Akismet, known spammer API -> https://de.wikipedia.org/wiki/Akismet#Datenschutzprobleme_in_Deutschland,_%C3%96sterreich_und_der_Schweiz
> 
>     https://www.drupal.org/project/botcha Botcha "anti-captcha" (the technique used is very easy to implement elsewhere but also very effective)

GrazingScientist:
Although not coming from China, in my public institute in Germany, we have a need of a reCaptcha alternative and hence would be very thankful for having a out-of-the-box-plugin.

t4x0n:
I want to propose a very simple idea, and this can be implemented even (I think) with reCaptcha enabled, a Honeypot like this: http://tidyrepo.com/registration-honeypot/ 

taken from the link above:
[...]

PR #6114, opened 2020-07-23
pkp/pkp-lib#2993 switch google.com URL to recaptcha.net

jnugent:
This switches the URL used for loading the reCaptcha javascript.  To fully fix problems for users from China, `enable_cdn` will also need to be turned off in the installation's `config.inc.php` file but I suspect that this will have already been done for Chinese users since a large portion of interface depends on jQuery anyway.


asmecher:
Thanks, @jnugent!

PR #6124, opened 2020-07-27
Update user.po

p-heckler:
I added a few missing strings to FR_Fr, mostly taken from Fr_CA.


asmecher:
@p-heckler, would you be willing to try entering these via Weblate rather than Github? (Among other things, this will get things merged into the right branch -- currently `stable-3_2_1`.) See https://docs.pkp.sfu.ca/translating-guide/en/ for details.

p-heckler:
@asmecher Yes, sorry about that, I should have checked the doc before committing. I entered this as well as some other edits to OJS FR_Fr on Weblate already. I will try to finish bringing this locale up to date by the end of the week if I can find the time (OJS first, then Web library).

asmecher:
Excellent, thanks!

PR #6127, opened 2020-07-28
PreviewReviewForm - Remove unused and incorrect variable

ecancer:
The $count variable is incorrectly set, receiving the value 1 instead of the number of review form elements. It is also unused and so can be removed.


asmecher:
Merging and forward-porting to `master` -- thanks, @ecancer!

PR #6199, opened 2020-08-24
Remove redundancy

Mandroide:
Remove redundancy of statements


asmecher:
Always happy to tidy things up like this -- thanks, @Mandroide!

PR #6245, opened 2020-09-21
indexing pdf files fix

sioulasdimitris:
there was an issue with indexing the pdf files.

After creating the txt file
$exec = sprintf($prog, escapeshellarg($this->getFilePath()));
$this->fp = @Popen($exec, 'rb');

this->fp was still pointing at the pdf file so the parsing could'nt work.

Changed the filepath
$this->setFilePath(substr($this->filePath, 0, strrpos($this->filePath, ".")).".txt");
$this->fp = @fopen($this->filePath, 'rb');
So the fp to point at the new txt file.

### DIFF
--- a/classes/search/SearchHelperParser.inc.php
+++ b/classes/search/SearchHelperParser.inc.php
@@ -31,7 +31,10 @@ class SearchHelperParser extends SearchFileParser {
 
 		if (isset($prog)) {
 			$exec = sprintf($prog, escapeshellarg($this->getFilePath()));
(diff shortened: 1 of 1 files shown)


asmecher:
@sioulasdimitris, I think you're encountering this problem because your PDF to text conversion tool is configured to generate a `.txt` file with the same name as the PDF, when OJS expects it to send the text to standard output. (That's the purpose of the `-` in the default command line.)

PR #6247, opened 2020-09-22
pkp/pkp-lib#5772 adapt DAO for the JSON serialization

Vitaliy-1:


classes/db/DAO.inc.php
```diff
@@ -354,7 +354,13 @@ function convertFromDB($value, $type) {
 				break;
 			case 'object':
 			case 'array':
-				$value = unserialize($value);
+				// Check needed only for upgrade to 3.3, where values are retrieved before being encoded to JSON
+				$decodedValue = json_decode($value, true);
+				if (!is_null($decodedValue)) {
```
asmecher, line 359: Once we have everything moved to migrations we won't have to worry about this kind of stuff! But it'll help to note it in a git issue so this kind of thing can be cleared out when it's no longer needed. I've filed https://github.com/pkp/pkp-lib/issues/6250; please add a note here along the lines of:
```
// FIXME: pkp/pkp-lib#6250 Remove after 3.3.x upgrade code is removed (see also pkp/pkp-lib#6247)
```
...and also add a note on #6250 about this.

classes/site/SiteDAO.inc.php
```diff
@@ -18,6 +18,12 @@
 
 class SiteDAO extends DAO {
 
+	/** @var string One of the SCHEMA_... constants */
+	var $schemaName = 'site';
```
asmecher, line 22: This should still be `= SCHEMA_SITE` rather than re-introducing the string constant, I think.

classes/db/DAO.inc.php
```diff
@@ -406,7 +412,7 @@ function convertToDB($value, &$type) {
 		switch ($type) {
 			case 'object':
 			case 'array':
-				$value = serialize($value);
+				$value = json_encode($value);
```
asmecher, line 415: Does `JSON_UNESCAPED_UNICODE` need to be specified here too?

classes/site/SiteDAO.inc.php
Vitaliy-1: There was a problem associated with it, I don't remember exactly. Does PHP allow that?

asmecher: Yes, you can use a constant to initialize a class member (if that's the problem you're thinking of) -- all I can think is that maybe you were using the constant before it was defined. But anyway, I think this takes precedence: https://github.com/pkp/ojs/pull/2873#discussion_r493912372

classes/db/DAO.inc.php
Vitaliy-1: Yes, you are right. Do you think other `JSON_UNESCAPED_...` options could be useful here? It all shouldn't affect decoding but I'm not experienced here.

asmecher: No, I'm not aware of others -- just spotted this one due to inconsistency with other `json_decode` calls in the PR.

Vitaliy-1: Yes, I've used it in the upgrade script according to the recommendations I found on the web; just forgot to put it here. Looking at the other options, I don't think that turning off escaping of line terminators or slashes will make any difference.

PR #6305, opened 2020-10-29
pkp/pkp-lib#4595 Consider sessions for journals using alternative domains

ajnyga:
Consider journals using alternative domains and allow session domain overrides from config.inc.php

With these changes, if no config value is given, each journal using alternative domain will create a new session for the journal.

Scenario 1: an OJS installation with:
site.com
site.com/journal1
journal2.com (own domain)

When the user goes journal2.com a new session and cookie is created. If that user logs in and moves to the site level (site.com) she will have to log in again. This is the ideal functionality with environments where the journals need to act independent from the site level.

Scenario 2:  an OJS installation with:
site.com
site.com/journal1
journal2.site.com (subdomain)

The ideal solution is to use the config setting to override the cookie domain so that it will always be site.com.
This means that only one cookie and session is ever created even when the user visits journal2.site.com. Also, the user only has to log in once.

*Question*

What would happen, if in scenario 1 the site would override the cookie domain and just use "site.com". Would that break the journal2.com site?

@asmecher @ctgraham

### DIFF
--- a/classes/session/SessionManager.inc.php
+++ b/classes/session/SessionManager.inc.php
@@ -38,11 +38,17 @@ class SessionManager {
 		ini_set('session.name', Config::getVar('general', 'session_cookie_name')); // Cookie name
 		ini_set('session.cookie_lifetime', 0);
 		ini_set('session.cookie_path', Config::getVar('general', 'session_cookie_path', $request->getBasePath() . '/'));
-		ini_set('session.cookie_domain', $request->getServerHost(null, false));
 		ini_set('session.gc_probability', 1);
 		ini_set('session.gc_maxlifetime', 60 * 60);
 		ini_set('session.auto_start', 1);
 		ini_set('session.cache_limiter', 'none');
+
+		// Allow cookie_domain override from config.inc.php
+		if (Config::getVar('general', 'session_cookie_domain')){
+			ini_set('session.cookie_domain', Config::getVar('general', 'session_cookie_domain'));
+		} else {
+			ini_set('session.cookie_domain', $request->getServerHost(null, false));
+		}
 
 		session_set_save_handler(
 			array($this, 'open'),
@@ -53,6 +59,14 @@ class SessionManager {
 			array($this, 'gc')
 		);
 
+		// Check if the session is tied to the parent domain
+		if (isset($this->userSession) && $this->userSession->getDomain() && $this->userSession->getDomain() != $request->getServerHost(null, false)) {
+			// if current host contains . and the session domain (is a subdomain of the session domain), adjust the session's domain parameter to the parent
+			if (strtolower(substr($request->getServerHost(null, false), -1 - strlen($this->userSession->getDomain()))) == '.'.strtolower($this->userSession->getDomain())) {
+				ini_set('session.cookie_domain', $this->userSession->getDomain());
+			}
+		}
+
 		// Initialize the session. This calls SessionManager::read() and
 		// sets $this->userSession if a session is present.
 		session_start();
@@ -61,14 +75,6 @@ class SessionManager {
 		$ip = $request->getRemoteAddr();
 		$userAgent = $request->getUserAgent();
 		$now = time();
-
-		// Check if the session is tied to the parent domain
-		if (isset($this->userSession) && $this->userSession->getDomain() && $this->userSession->getDomain() != $request->getServerHost(null, false)) {
-			// if current host contains . and the session domain (is a subdomain of the session domain), adjust the session's domain parameter to the parent
-			if (strtolower(substr($request->getServerHost(null, false), -1 - strlen($this->userSession->getDomain()))) == '.'.strtolower($this->userSession->getDomain())) {
-				ini_set('session.cookie_domain', $this->userSession->getDomain());
-			}
-		}
 
 		if (!isset($this->userSession) || (Config::getVar('security', 'session_check_ip') && $this->userSession->getIpAddress() != $ip) || $this->userSession->getUserAgent() != substr($userAgent, 0, 255)) {
 			if (isset($this->userSession)) {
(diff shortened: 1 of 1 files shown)


ctgraham:
In scenario 1, what do you mean by "if the user first logs in to site.com and then moves to journal2.com, the same cookie domain (site.com) and the same cookie is still used." ?

If a user connects to `journal2.com`, no cookies for `site.com` will be offered by the user agent in the request.  If `journal2.com` tries to set a cookie for `site.com` in the response, the user agent should not allow it.

ajnyga:
Yeah, I was thinking about that (see the question below).

Scenario 1 basically should talk about subdomains. That is the enviroment we are using and that is the use case where the original problem has risen.

Having completely separate domains probably need more than just this. I mean for example the shared user database is not a good idea there.

ajnyga:
The difficulty here is that I do not have the necessary setup for testing this with multiples domains.

ctgraham:
Have you tried hostfile overrides here?  If you can spin up a development instance or a local VM with the webserver configured to response to different hostnames, you can override those hostnames in your local `hosts` file on you workstation.  That will allow you to interact with the install as if you had registered the domain names.

ajnyga:
hmm, I have to see if I would get that working with XAMPP. The installation I am working with is actually online.

The pr above imo covers most cases. The question is, as you noticed, what happens when someone moves from site.com to journal2.com.

classes/session/SessionManager.inc.php
```diff
@@ -38,12 +38,18 @@ function __construct($sessionDao, $request) {
...
 		ini_set('session.cookie_lifetime', 0);
 		ini_set('session.cookie_path', Config::getVar('general', 'session_cookie_path', $request->getBasePath() . '/'));
-		ini_set('session.cookie_domain', $request->getServerHost(null, false));
 		ini_set('session.gc_probability', 1);
 		ini_set('session.gc_maxlifetime', 60 * 60);
 		ini_set('session.auto_start', 1);
 		ini_set('session.cache_limiter', 'none');
 
+		// Allow cookie_domain override from config.inc.php
+		if (Config::getVar('general', 'session_cookie_domain')){
```
asmecher, line 47: You can provide a default in the optional 3rd parameter of `Config::getVar`, i.e. I think you could replace this entire block with:
```
ini_set(
  'session.cookie_domain',
  Config::getVar('general', 'session_cookie_domain', $request->getServerHost(null, false))
);
```

classes/session/SessionManager.inc.php
```diff
@@ -53,6 +59,14 @@ function __construct($sessionDao, $request) {
 			array($this, 'gc')
 		);
 
+		// Check if the session is tied to the parent domain
```
asmecher, line 62: I would expect the `config.inc.php` override to apply to this as well when it's provided, no?

classes/session/SessionManager.inc.php
```diff
@@ -215,7 +221,14 @@ function gc($maxlifetime) {
 	 * @return boolean
 	 */
 	function updateSessionCookie($sessionId = false, $expireTime = 0) {
-		$domain = ini_get('session.cookie_domain');
```
asmecher, line 218: Could you describe this change in a little more detail? I would expect, as long as we set the domain correctly upon session creation, and use `ini_set` to store it, then `ini_get` should always be reliable.

classes/session/SessionManager.inc.php
ctgraham: Wait.  How would `$this->userSession` ever be already set in the constructor?  Can `session_set_save_handler()` somehow intervene to call `$this->read()` even before the object is instantiated?  This seems off.

ajnyga: @ctgraham I was debugging this and it seems you are correct yesterday and you are right, I could not see any situation where $this->userSession was set and the check would actually work

ajnyga: Because as far as I can see, this is the exact code that should take care of subdomains, right?

ajnyga: Oh, and if you move that piece of code to it's original place, after session_start, you get:
`PHP Warning: ini_set(): A session is active. You cannot change the session module’s ini settings at this time in lib/pkp/classes/session/SessionManager.inc.php on line 76,`

ajnyga: What if we do this in the beginning of construct. Seems to worki

```
		$cookieDomain = $request->getServerHost(null, false);
		$siteBaseUrlHost = parse_url(Config::getVar('general', 'base_url'), PHP_URL_HOST);
		if (strtolower(substr($cookieDomain, -1 - strlen($siteBaseUrlHost))) == '.'.strtolower($siteBaseUrlHost)) {
			$cookieDomain = $siteBaseUrlHost;
		}
		ini_set('session.cookie_domain', $cookieDomain);
```

ajnyga: That would be the only required change. No new config setting needed.

Basically we are comparing the base_url setting and the getServerHost. If getServerHost returns a subdomain, the code will compare it with the base_url setting and use the base_url instead. 

If there would be an independent journal domain, getServerHost would return that and would use it like it does with the current code, creating a whole new session.

If the base_url setting is not configured correctly (OJS works even if you leave the pkp.sfu.ca url there) it should cause a problem

Do you see a problem @ctgraham ?

classes/session/SessionManager.inc.php
ajnyga: yes, I have a potential solution in the discussion above

classes/session/SessionManager.inc.php
ajnyga: I have a potential solution below that would not require any new settings and should work with any combination of domains.

classes/session/SessionManager.inc.php
ajnyga: See 
https://github.com/pkp/pkp-lib/pull/6315
https://github.com/pkp/ojs/pull/2895

Issue #6284, opened 2020-10-15
During upgrade to 3.3 email templates' stage_id is populated before being created


Vitaliy-1:
Upgrades of OMP/OPS/OJS from versions below 3.2.1 to the 3.3 (current master branch) throw a fatal error:
```
[code: Installer Installer::migrateStaticPagesToNavigationMenuItems]
[data: dbscripts/xml/upgrade/3.2.1_update.xml]
[code: Installer Installer::installEmailTemplate]
PHP Fatal error:  Uncaught Exception: DB Error: Unknown column 'stage_id' in 'field list' Query: INSERT INTO email_templates_default
                                (email_key, can_disable, can_edit, from_role_id, to_role_id, stage_id)
                                VALUES
                                ('ANNOUNCEMENT', 0, 1, 16, 1048576, null) in /home/doc/OJS/development/omp/lib/pkp/classes/db/DAO.inc.php:709
Stack trace:
#0 /home/doc/OJS/development/omp/lib/pkp/classes/db/DAO.inc.php(231): DAO->handleError()
#1 /home/doc/OJS/development/omp/lib/pkp/classes/mail/EmailTemplateDAO.inc.php(215): DAO->update()
#2 /home/doc/OJS/development/omp/lib/pkp/classes/install/Installer.inc.php(651): EmailTemplateDAO->installEmailTemplates()
#3 /home/doc/OJS/development/omp/lib/pkp/classes/install/Installer.inc.php(445): Installer->installEmailTemplate()
#4 /home/doc/OJS/development/omp/lib/pkp/classes/install/Installer.inc.php(265): Installer->executeAction()
#5 /home/doc/OJS/development/omp/lib/pkp/classes/install/Installer.inc.php(188): Installer->executeInstaller()
#6 /home/doc/OJS/development/omp/lib/pkp/classes/cl in /home/doc/OJS/development/omp/lib/pkp/classes/db/DAO.inc.php on line 709
``` 
This happens because email templates are being installed before the new upgrade scripts creates a new column `stage_id`: https://github.com/pkp/pkp-lib/blob/be20999f5ebee89d970ef42f6092a56b4f275d14/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php#L49-L54

Possible solutions:
- Run `Installer::installEmailTemplate` after`PKPv3_3_0UpgradeMigration.inc.php` script execution
-  Create `stage_id` column before email templates are installed if it doesn't exists

[...]

PR #6311, opened 2020-11-03
pkp/pkp-lib#6284 Add stage_id column if doesn't exist

Vitaliy-1:


asmecher:
@Vitaliy-1, can this not go in the migration script?

Vitaliy-1:
@asmecher, I didn't have any notifications or adverse events when running this. When called statically, SchemaBuilder already has data regarding the connection. Or you mean something different?

asmecher:
That's not what I meant, but no, you're right, it needs to live in Installer for now and the `FIXME` will help flag it for removal later. Please go ahead!

Issue #6285, opened 2020-10-15
[OMP] During upgrade to 3.3 script tries to drop submissions.section_id which doesn't exist


Vitaliy-1:
OMP 3.3 (current master branch).

Upgrade from 3.2.0 throws error:
```
ERROR: Upgrade failed: DB: SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'section_id'; check that column/key exists (SQL: alter table `submissions` drop `section_id`)
```
I don't think OMP has sections, at least I don't see the correspondent table and row in `submissions` and `publications` on my test instance. Think, [this statement](https://github.com/pkp/pkp-lib/blob/be20999f5ebee89d970ef42f6092a56b4f275d14/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php#L29)  should be moved from the shared library.

asmecher:
@Vitaliy-1, a 3.2.0 database should have a `section_id` column in the `submissions` table; see https://github.com/pkp/pkp-lib/blob/3_2_0-0/xml/schema/submissions.xml#L164...L166. Was this database already partially upgraded from 3.2.x to 3.3.x? If so, maybe the column was already dropped?

Vitaliy-1:
Yes, table `section_editors` contains a column named `section_id` but I don't see the same in the `submissions` table.
Just made a test installation of 3.2.0-3 downloaded from the site to confirm that.
```
SHOW COLUMNS FROM submissions

submission_id	bigint(20)	NO	PRI	
    NULL
	auto_increment	
context_id	bigint(20)	NO	MUL	
    NULL
		
locale	varchar(14)	YES		
    NULL
		
current_publication_id	bigint(20)	YES		
    NULL
		
date_last_activity	datetime	YES		
    NULL
		
date_submitted	datetime	YES		
    NULL
		
last_modified	datetime	YES		
    NULL
		
stage_id	bigint(20)	NO		1		
status	tinyint(4)	NO		1		
submission_progress	tinyint(4)	NO		1		
work_type	tinyint(4)	NO		0		

```

asmecher:
> Yes, table section_editors contains a column named section_id but I don't see the same in the submissions table.

[...]

PR #6312, opened 2020-11-03
pkp/pkp-lib#6285 Drop submissions.section_id only if exists

Vitaliy-1:


asmecher (approved):
Looks good, @Vitaliy-1 -- please go ahead!

PR #6315, opened 2020-11-04
pkp/pkp-lib#4595 consider subdomains in session_domain 

ajnyga:


ctgraham:
Interesting.  Feels a bit like dark magic, but if we add some commentary in config.inc.php to describe what is happening, this is a fairly reasonable and elegant solution.  It will not cover the cases of some service providers, if they use single installs, but these are probably edge cases.  For example:
```
base_url = https://myOJShost.com
base_url[customerA1] = https://one.customer.com
base_url[customerA2] = https://two.customer.com
base_url[customerA3] = https://three.customer.com
base_url[customerB1] = https://one.another.com
base_url[customerB2] = https://two.another.com
base_url[customerB3] = https://three.another.com
base_url[cusomterC] = https://see.sea.cc
```

ajnyga:
Hi,

Yes true, the bug described in the original issue would probably still be there in a situation where you have separate site level domain and separate domain name for journal subdomains. Do you have a possible solution in mind that would support this?

In a way the current pr does what the original code was supposed to do. Of course the difference is that the `base_url` setting is used for comparison because the ini_set has to happen before session_start.

ctgraham:
When the removed (and currently broken) code was originally implemented in OJS 2.4, apparently the call to `session_start()` really did modify `$this->userSession` via `SessionManger::read()`.  (#$%#, *that* was ugly)
https://github.com/pkp/pkp-lib/blob/f825d80a9ddd210c3ac0bec8d2b14041d3b40993/classes/session/SessionManager.inc.php#L58-L59

I think the existing code to escalate the session to the parent *should* almost work, if we were to move most of the logic currently in the constructor into the SessionManager::open()` and `SessionManager::close()` methods.  The example above is yet another edge case which is not covered in the older code (since the domain `https://customer.com` or `https://another.com` is never referenced in config.inc.php).

Is it worth the refactoring effort to restore this old functionality? The primary common use case your code doesn't handle, which the original code was designed to, is that of `www` subdomain prefixes to any of the named `base_url[]`s:
```
base_url = https://myOJShost.com
base_url[customerA] = https://customer.com
base_url[customerB] = https://another.com
base_url[cusomterC] = https://see.sea.cc
```

Arguably, access starting at `https://www.customer.com` and `https://customer.com` should be treated equivalently, if routed to the same vhost.

PR #6316, opened 2020-11-04
Fixing mistranslation of Production Editor

luizbgomide:


asmecher:
@luizbgomide, could you make this change via translate.pkp.sfu.ca rather than Github? That way it'll be less likely to collide with changes made by other translators (and will get merged to the appropriate branches). Thanks!

luizbgomide:
Sure, I did not know that this plataform existed. I will do it right away.

PR #6333, opened 2020-11-13
pkp/pkp-lib#6331 Delimit single citations with a line break before import

Vitaliy-1:


eocarragain:
I tested this patch against OJS 3.2.1-1 and can confirm it fixes the issue & behaves as expected. Thanks @Vitaliy-1 !

Issue #6286, opened 2020-10-15
[OPS] During upgrade to 3.3 script refers to nonexistent table review_forms


Vitaliy-1:
OPS upgrade from 3.2.1 to 3.3 (current master branch).

During upgrade receiving the error message:
```
[pre-install]
[load: upgrade.xml]
[version: 3.2.1.0]
[code: Installer Installer::checkPhpVersion]
[code: Installer Installer::changeSubmissionStageToProduction]
[code: Installer Installer::installEmailTemplate]
[migration: PKPv3_3_0UpgradeMigration]
ERROR: Upgrade failed: DB: There is no column with name 'assoc_type' on table 'review_forms'.
```
Looks like the table `review_forms` doesn't exist in OPS and this part of the upgrade script should be moved from the shared library to OJS and OMP: https://github.com/pkp/pkp-lib/blob/be20999f5ebee89d970ef42f6092a56b4f275d14/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php#L83-L85

asmecher:
@Vitaliy-1, do you want to take a crack at this one too? Your proposal makes sense to me.

Vitaliy-1:
@asmecher, looking at this code once again, I don't see that this script changes the table in any way. Am I understand this correctly that here and in other places of the Migration script, it's expected the column to not be nullable? Although I can't find this on Laravel's site, `...->nullable(false)->change()` does the trick, at least on MySQL.

asmecher:
@Vitaliy-1, I'm still learning the ins and outs of the Laravel schema toolset -- it appears that nullable is [presumed false](https://laravel.com/docs/7.x/migrations#column-modifiers) during schema creation, but it's possible that for already-existing columns you need to manually reset it using `nullable(false)` or no change will be made. Does your testing confirm this?

Vitaliy-1:
@asmecher, yes, my test shows that nullable default value already exists, it should be explicitly changed. 

And for the JSON result of the statement on `authors` table: `$table->bigInteger('publication_id')->change();`:
```
{"type":"bigInteger","name":"publication_id","autoIncrement":false,"unsigned":false,"change":true}
```
And for `nullable(false)`:
```
{"type":"bigInteger","name":"publication_id","autoIncrement":false,"unsigned":false,"nullable":false,"change":true}
```
[...]

Issue #6093, opened 2020-07-15
Setup foreign key constraints


jonasraoni:
(Note: Remaining special cases have been filed here: https://github.com/pkp/pkp-lib/issues/8333)

**Describe the problem you would like to solve**
The tables are missing foreign key constraints, and that's a source of orphan data/unexpected statistics.

**Describe the solution you'd like**
Add constraints everywhere, considering carefully if deletes in cascade are the best option (easier to deal with and also to have accidents).

**Migrations to review/add constraints to:**

`pkp-lib`:
- [x] `lib/pkp/classes/migration/install/AnnouncementsMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/CategoriesMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/CommonMigration.inc.php` **except**:
  - [ ]  `plugin_settings.context_id` (uses `0` for site-wide)
  - [ ] `notifications.context_id` (uses `0` for site-wide)
  - [ ] `notifications.user_id` (uses `0` for "anyone" IIRC)
- [x] `lib/pkp/classes/migration/install/ControlledVocabMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/DoiMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/FailedJobsMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/FilesMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/GenresMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/JobsMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/LibraryFilesMigration.inc.php`
- [x] `lib/pkp/classes/migration/install/LogMigration.inc.php` **except**:
  - [ ] `email_log.sender_id` (uses `0` for "nobody in particular")
- [x] `lib/pkp/classes/migration/install/MetadataMigration.inc.php` **except**:
  - [ ] `filters.context_id` (uses `0` for site-wide)
  - [ ] `filters.parent_filter_id` (uses `0` for none)
- [x] `lib/pkp/classes/migration/install/MetricsMigration.inc.php` [moved to apps, FKs added already)
- [x] `lib/pkp/classes/migration/install/NavigationMenusMigration.inc.php`, **except**:
  - [ ] `navigation_menus.context_id` (`0` used for site-wide)
  - [ ] `navigation_menu_items.context_id` (`0` used for site-wide)
[...]

PR #6347, opened 2020-11-18
pkp/pkp-lib#6286 review_forms upgrade script, don't allow null values

Vitaliy-1:


Vitaliy-1:
I've noticed that this line in the script: https://github.com/Vitaliy-1/pkp-lib/blob/8271ff67399386cb6a9a69c17e8ab90933c16f80/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php#L388
Is duplicated below: https://github.com/Vitaliy-1/pkp-lib/blob/8271ff67399386cb6a9a69c17e8ab90933c16f80/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php#L427, where the sequence is regenerated.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -72,43 +72,39 @@ public function up() {
 
 		// pkp/pkp-lib#6093 Don't allow nulls (previously an upgrade workaround)
 		Capsule::schema()->table('user_settings', function (Blueprint $table) {
-			$table->bigInteger('assoc_type')->default(0)->change();
-			$table->bigInteger('assoc_id')->default(0)->change();
+			$table->bigInteger('assoc_type')->default(0)->nullable(false)->change();
+			$table->bigInteger('assoc_id')->default(0)->nullable(false)->change();
```
asmecher, line 76: When we start defining `FOREIGN KEY` relationships properly (https://github.com/pkp/pkp-lib/issues/6093) we won't be able to use `0` IDs as de-facto `NULL` values as they'll cause DB errors. So I'd recommend changing any case where a `0` ID is used so that it's updated to `NULL`, and the schema/migration code is adapted to allow `NULL` entries. That's definitely the case with `user_settings`.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -72,43 +72,39 @@ public function up() {
...
-			$table->bigInteger('review_round_id')->change();
+			$table->bigInteger('review_round_id')->nullable(false)->change();
 		});
 		Capsule::schema()->table('authors', function (Blueprint $table) {
-			$table->bigInteger('publication_id')->change();
+			$table->bigInteger('publication_id')->nullable(false)->change();
 		});
 		Capsule::schema()->table('edit_decisions', function (Blueprint $table) {
-			$table->bigInteger('review_round_id')->change();
+			$table->bigInteger('review_round_id')->nullable(false)->change();
```
asmecher, line 107: I believe this should still be `NULL`able for decisions that aren't related to the review stages.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
asmecher: (I've commented on the cases I could spot; the ones I haven't mentioned should be fine.)

asmecher (changes requested):
Just two comments to work on, thanks!

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
Vitaliy-1: @asmecher, are you sure on this one? On my test instances 3.2.0 and 3.2.1, those (that aren't associated with the review stage) have a `0` value. On the master branch, `accept and skip review` records the following:
```
edit_decision_id	submission_id	review_round_id	stage_id	round	editor_id	decision	date_decided		
566	254	0	1	0	1	1	2020-11-24 11:20:06	
```
Or this is going to be something new? In this case all those values should be converted from `0` to `null` as per previous comment.

asmecher: You're right, currently it defaults to `0`. In the future we'll need to convert that to `NULL`, but for now, it can stay as `0`.

Vitaliy-1: If it's planned to make null values there, I've removed those lines completely from the script (the last commit).

asmecher: OK, but you'd need to permit `NULL` values here: https://github.com/pkp/pkp-lib/pull/6347/files#diff-f4951b8d90b205bf60a1111a761b62eb75f598c0c536738da01d027561a33f1cR101
...and, for consistency (so we don't have both `NULL` and `0` values), remove the `(int)` casts where review assignments are inserted and updated. (Use `DAO::nullOrInt` instead.)

Vitaliy-1: @asmecher, I've updated PRs but I don't get it. How it's possible that review assignments may have null or int values for review round ID? Those can be inserted or updated outside of the review process, like editor decisions?

asmecher: Sorry, @Vitaliy-1, I'm yet again reading too quickly and leading you astray. I had meant the `edit_decisions` table but you've already removed that line and I was pointing at the wrong table.

Vitaliy-1: No problem. I've removed the last commit.

PR #6360, opened 2020-11-21
Update ru_RU locale after PR #6332

ppv1979:


asmecher:
Thanks for these, @ppv1979!

Issue #6452, opened 2020-12-07
error when removing a user group


bozana:
**Describe the bug**
I created a new user group and then try to remove it.

**To Reproduce**
1. Go to User & Roles > Roles > a User Group
2. Click on "Remove" action
3. See error in the php error log file

**What application are you using?**
 OMP version 3.3

**Additional information**
`PHP Fatal error: Uncaught Error: Cannot use object of type stdClass as array in /.../omp-master/lib/pkp/classes/security/UserGroupDAO.inc.php:408\nStack trace:\n#0 /.../omp-master/lib/pkp/controllers/grid/settings/roles/UserGroupGridHandler.inc.php(319): UserGroupDAO->getContextUsersCount()\n#1 /.../omp-master/lib/pkp/classes/core/PKPRouter.inc.php(391): UserGroupGridHandler->removeUserGroup()\n#2 /.../omp-master/lib/pkp/classes/core/PKPComponentRouter.inc.php(257): PKPRouter->_authorizeInitializeAndCallRequest()\n#3 /.../omp-master/lib/pkp/classes/core/Dispatcher.inc.php(143): PKPComponentRouter->route()\n#4 /.../omp-master/lib/pkp/classes/core/PKPApplication.inc.php(353): Dispatcher->dispatch()\n#5 /...omp-master/index.php(64): PKPApplication->execute()\n#6 {main}\n thrown in /.../omp-master/lib/pkp/classes/security/UserGroupDAO.inc.php on line 408, referer: http://omp-master.bb/index.php/press1/management/settings/access `

after the fix below the following error appears:
[...]

PR #6453, opened 2020-12-07
pkp/pkp-lib#6452 error when removing a user group

bozana:
s. https://github.com/pkp/pkp-lib/issues/6452

### DIFF
--- a/classes/security/UserGroupDAO.inc.php
+++ b/classes/security/UserGroupDAO.inc.php
@@ -399,12 +399,12 @@ class UserGroupDAO extends DAO {
 			'SELECT	COUNT(DISTINCT(uug.user_id)) AS row_count
 			FROM	user_groups ug
 				JOIN user_user_groups uug ON ug.user_group_id = uug.user_group_id
-			WHERE	context_id = ?' .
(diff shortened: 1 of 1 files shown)


classes/security/UserGroupDAO.inc.php
```diff
@@ -399,12 +399,12 @@ function getContextUsersCount($contextId, $userGroupId = null, $roleId = null) {
...
 			FROM	user_groups ug
 				JOIN user_user_groups uug ON ug.user_group_id = uug.user_group_id
-			WHERE	context_id = ?' .
+			WHERE	ug.context_id = ?' .
 				($userGroupId?' AND ug.user_group_id = ?':'') .
 				($roleId?' AND ug.role_id = ?':''),
 			$params
 		);
-		$row = $result->current();
+		$row = (array) $result->current();
```
asmecher, line 407: This is fine (and I'll merge it as is), but I'm starting to prefer the object syntax to the array syntax. In this case, instead of:
```
$row = (array) $result->current();
return $row ? $row['row_count'] : 0;
```
...you would write...
```
$row = $result->current();
return $row ? $row->row_count : 0;
```

PR #6497, opened 2020-12-18
Backport viewPublic library files action to 3.1.2

ctgraham:
C.f. #6473

Issue #6473, opened 2020-12-11
Allow for access to public Library Files without forcing a download


ctgraham:
**Describe the problem you would like to solve**
Example: Designers and Managers want a place to store publicly available files which can be rendered directly to the browser, rather than requiring the `Content-Disposition` of "attachment".

**Describe the solution you'd like**
Add an endpoint of `{context}`/libraryFiles/viewPublic/`{id}`

**Who is asking for this feature?**
Site designers looking to upload CSS, JavaScript, Images, or other files.

ctgraham:
These referenced PRs are designed to add the functionality without requiring a translation round, but a subsequent clarification of the message `settings.libraryFiles.public.viewInstructions` should probably trigger a translation round.

asmecher:
A long time ago, we used to have a MIME type list for files that should be presented inline vs. downloaded; does it make sense to present both download and view options for all file types? Has browser behaviour improved? https://github.com/pkp/pkp-lib/blob/ojs-2_4_8-2/classes/file/PKPFileDAO.inc.php

The immediate need for this is gone since we're presenting "inline"-style galleys (HTML, XML) using their own plugins, which take care of the needed behaviour -- one example of the current inline-vs-download logic applies to CSS attached to HTML galleys:
https://github.com/pkp/ojs/blob/e44d405dbb1bd10f7ac98452db020d51120ed56f/plugins/generic/htmlArticleGalley/HtmlArticleGalleyPlugin.inc.php#L137

ctgraham:
My opinion is that the `Content-Disposition: attachment` header is unnecessary at this point and generally shouldn't be used.  Let the browser deal with the file.

I would have just changed the link to inline everything, but, I don't want to introduce a backwards-incompatible change to stable, and I do want to backport to stable.

Thus, an overly verbose compromise.

ctgraham:
Preferences on next steps here, @asmecher ?

asmecher:
Sorry about the wait. I think the library is going to be used for type of files that don't have ambiguity over inline presentation vs. downloading (as .txt or .css files might), so I'm OK with just letting the browser decide.

ctgraham:
Refactored as a bugfix, rather than an enhancement.

asmecher:
Merged those 4 PRs -- thanks, @ctgraham!

PR #6499, opened 2020-12-18
Provide handler to view public library files

ctgraham:
C.f. #6473 for HEAD; requested backports of #6497 and #6498.


ctgraham:
Recoding per https://github.com/pkp/pkp-lib/issues/6473#issuecomment-759764955

PR #6715, opened 2021-02-03
pkp/pkp-lib#6710 Unserialize journal settings depending on a value data type

Vitaliy-1:


Vitaliy-1:
(Note: not ready yet, described details in the issue)

PR #6769, opened 2021-02-18
pkp/pkp-lib#5844 Changing site-wide about field to rich textarea. And creating a a migration to handle cases when old data has html special characters

henriqueramos:
Evidence it's attached on issue #5844.


classes/components/forms/site/PKPSiteInformationForm.inc.php
```diff
@@ -37,10 +36,12 @@ public function __construct($action, $locales, $site) {
 		$this->action = $action;
 		$this->locales = $locales;
 
-		$this->addField(new FieldTextarea('about', [
+		$this->addField(new FieldRichTextarea('about', [
 				'label' => __('admin.settings.about'),
 				'isMultilingual' => true,
 				'value' => $site->getData('about'),
+				'toolbar' => 'bold italic | link | blockquote bullist numlist',
```
asmecher, line 43: I think it's best to leave the toolbar and plugis at their defaults so the field is consistent with the other rich text fields.

classes/migration/ConvertSiteWideAboutToRichTextArea.inc.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+/**
+ * @file classes/migration/ConvertSiteWideAboutToRichTextArea.inc.php
```
asmecher, line 4: We don't have a pattern for this yet, but I think issue-specific migration classes should identify the issue numbers in both class names and class filenames, e.g. `I5844ConvertSiteWideAboutToRichTextArea`.

asmecher (changes requested):

classes/migration/ConvertSiteWideAboutToRichTextArea.inc.php
henriqueramos: Indeed. I've add as a prefix on it. I.e "Issue5844".

PR #6816, opened 2021-03-03
pkp/pkp-lib#6801 replace name with original_file_name for dependent files

Vitaliy-1:


classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -474,6 +474,18 @@ private function _migrateSubmissionFiles() {
...
+		$dependentFiles = Capsule::table('submission_files')
+			->select('file_id', 'original_file_name')
+			->where('file_stage' , '=', SUBMISSION_FILE_DEPENDENT)
+			->get();
+		foreach ($dependentFiles as $dependentFile) {
+			Capsule::table('submission_file_settings')
+				->where('file_id', '=', $dependentFile->file_id)
+				->where('setting_name', '=', 'name')
+				->update(['setting_value' => $dependentFile->original_file_name]);
+		}
```
NateWr, line 487: This might cause an issue with a large array as Alec mentioned. @asmecher is there a more performant way to do this?

Vitaliy-1: Something like https://laravel.com/docs/8.x/queries#chunking-results?

Vitaliy-1: I mean instead of getting all records in `$dependentFiles`, process by 100 items at a time.  I think, recalling the conversation, it should be enough.

asmecher: This wouldn't be a problem with array size, as `foreach(Capsule::table(...)->...->get())` will work through the result set via iteration rather than materializing to a PHP array. (This is where `get()` is better than `pluck()` -- the latter materializes the results into memory.)

It would be quicker if written as a multi-table `UPDATE`, but that would need to be written once each for PostgreSQL and MySQL and that's best avoided unless the optimization is necessary.

The large installations I'm aware of don't publish often with HTML, so I'd be happy to leave this as it is until we know it needs optimizing.

asmecher: Oh! Doing a bit of searching, it looks like `get()` may not actually iterate, but instead materializes into memory just like `pluck`. (See e.g.: https://stackoverflow.com/questions/28082279/how-do-i-use-laravels-chunk-to-avoid-running-out-of-memory) So yes, `chunk` is what I'd recommend using in large result lists -- it's easy enough to add here and I'll look at adding it elsewhere.

PR #6856, opened 2021-03-11
Closes pkp/pkp-lib#3193 Added user email templates to the discussion …

jonasraoni:
…and grouped user templates in the assign participant.

### DIFF
--- a/js/controllers/grid/queries/QueryFormHandler.js
+++ b/js/controllers/grid/queries/QueryFormHandler.js
@@ -44,8 +44,6 @@
 			this.templateUrl_ = options.templateUrl;
 		}
 
-		this.textSubject_ = null;
-
 		// Attach form elements events.
 		$form.find('#template').change(
 			this.callbackWrapper(this.selectTemplateHandler_));
@@ -83,7 +81,6 @@
 		var $form = this.getHtmlElement();
 		var template = $form.find('[name="template"]');
 		$form.find('[name="subject"]').val(template.find('option:selected').text());
-		this.textSubject_
 		$.post(this.templateUrl_, template.serialize(),
 				this.callbackWrapper(this.updateTemplate), 'json');
 	};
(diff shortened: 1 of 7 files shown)


js/controllers/grid/queries/QueryFormHandler.js
```diff
@@ -0,0 +1,123 @@
...
+	 * @param {Event} event The triggering event.
+	 * @private
+	 */
+	$.pkp.controllers.grid.queries.
+			QueryFormHandler.prototype.selectTemplateHandler_ =
+					function(sourceElement, event) {
+		var $form = this.getHtmlElement();
+		var template = $form.find('[name="template"]');
+		$form.find('[name="subject"]').val(template.find('option:selected').text());
+		this.textSubject_
```
NateWr, line 86: Is this intended to not have a `;` at the end? I think that syntax is only compatible with some browsers.

NateWr (changes requested):
This looks great, @jonasraoni, thanks. I just ran into one problem while testing. The templates don't load into the message body when I use the "notify" feature.

**Steps to Reproduce**
1. Open submission workflow.
2. Find "Participants" list on the right and expand the options for one user.
3. Click "Notify"
4. In the message window, select one of the custom templates.
5. Nothing appears in the message area.

It looks like the request gets sent off to the stage participants grid, but the response value is empty for the body:

```json
{
	"status": true,
	"content": {
		"body": "",
		"variables": null
	},
	"elementId": "0",
	"events": null
}```

js/controllers/grid/queries/QueryFormHandler.js
jonasraoni: Hmm, if I remember well, I was going to save the texts (to recover the last typed texts while switching through the templates), but then I changed my mind, as it would require a button to reset...
So, I don't know why the `textSubject_` is still there :D

jonasraoni:
Thanks @NateWr, I'll take a look at this issue.

Issue #6970, opened 2021-04-21
Allow app-specific event registration


Vitaliy-1:
**Describe the problem you would like to solve**
The current implementation of Laravel events allows registering events and their listeners only in shared library, see: https://github.com/pkp/pkp-lib/blob/c5de60d01d3ac197df3f17a2bc58f3b0cf97e009/classes/core/PKPEventServiceProvider.inc.php, in boot method or listen property. Allow each application to add its own events and listeners, modify existing ones. 

**Describe the solution you'd like**
Boot additional event data from app-specific EventServiceProvider class

Vitaliy-1:
This simplifies events and listeners registration and implements Laravel's event discovery: https://laravel.com/docs/7.x/events#event-discovery

Published Event as a first example is added to the newly created `observers/events` directory, listeners are expected to be added to `observers/listeners`, see: https://github.com/Vitaliy-1/pkp-lib/blob/a8b4fcda4a094ccbef0d50fe4ec07c2720ac3d8e/classes/core/PKPEventServiceProvider.inc.php#L156
This allows avoiding the creation of the app-specific `EventServiceProvider` class.

Vitaliy-1:
Think it's ready, the last commit should fix tests for PHP 7.3, although I'm not sure why other tests are failing on Crossref scripts.
@asmecher, can you take a look at the implementation?

Vitaliy-1:
Adding a custom listener to the already defined event from a plugin is quite straightforward: https://github.com/Vitaliy-1/listener/blob/13cb133cdf81140969231c9619468e4a2b0d56c6/ListenerPlugin.inc.php#L57 and should work alongside with event discovery

PR #7005, opened 2021-04-29
pkp/pkp-lib#6970 Laravel event discovery implementation

Vitaliy-1:


classes/core/PKPEventServiceProvider.inc.php
```diff
@@ -78,6 +84,82 @@ public function boot()
...
+        {
+            /**
+             * @param SplFileInfo $file
+             * @param string $basePath base path of the application
+             * @return string listener's full class name
+             */
+            protected static function classFromFile(SplFileInfo $file, $basePath): string
+            {
+                $pathFromBase = trim(Str::replaceFirst($basePath, '', $file->getRealPath()), DIRECTORY_SEPARATOR);
+                $libPath = 'lib' . DIRECTORY_SEPARATOR . 'pkp' . DIRECTORY_SEPARATOR;
```
asmecher, line 129: I think it's safe and way more readable to just use `/` in place of `DIRECTORY_SEPARATOR.

classes/core/PKPEventServiceProvider.inc.php
```diff
@@ -78,6 +84,82 @@ public function boot()
...
+
+    /**
+     * Get the listener directories that should be used to discover events.
+     *
+     * @return array
+     */
+    protected function discoverEventsWithin()
+    {
+        return [
+            $this->app->basePath('lib/pkp/classes/observers/listeners'),
```
asmecher, line 159: It's arguable whether we should have this inside the `classes` directories or not, but until the autoloader supports that I can see why you've put them here :)

I'll continue to think about that and hope to have a proposal next week.

classes/core/PKPEventServiceProvider.inc.php
```diff
@@ -78,6 +84,82 @@ public function boot()
...
+                );
+            }, []);
+    }
+
+    /**
+     * Get the listener directories that should be used to discover events.
+     *
+     * @return array
+     */
+    protected function discoverEventsWithin()
```
asmecher, line 156: Does this scan happen on a per-request basis or does this use the Laravel event cache? (https://laravel.com/docs/8.x/events#event-discovery-in-production)

classes/core/PKPEventServiceProvider.inc.php
Vitaliy-1: I'm not sure here what `SplFileInfo::getRealPath()` returns on the Windows machine. `DIRECTORY_SEPARATOR` here is for correct matching when doing replacements. According to the comment here it returns backslashes: https://www.php.net/manual/en/splfileinfo.getpathname.php

classes/core/PKPEventServiceProvider.inc.php
Vitaliy-1: No, it requires registration of at least Laravel's `FileServiceProvider`. Haven't dive deep into the code, but maybe `CacheServiceProvider` also. I'm not sure how slow the iteration [through listener's methods](https://github.com/laravel/framework/blob/cada3d21c7f127ba01bb949b2fa045b7faaedbca/src/Illuminate/Foundation/Events/DiscoverEvents.php#L55) with `ReflectionClass` is. Do you think event caching is something that should be implemented right away?

Vitaliy-1: On the second look, I can override Laravel's methods to use our file cache. Not necessary to integrate additional Laravel services. @asmecher, do you think it's worth an effort to do it now, before merging this PR?

asmecher: I'm content to start as it is, but we'll have to watch as we adopt the event toolset to see whether performance becomes an issue. Thanks!

PR #7011, opened 2021-05-03
pkp/pkp-lib#5299 Move Submission Preview to pkp-lib

ajnyga:


classes/services/PKPSubmissionService.inc.php
```diff
@@ -941,4 +941,66 @@ public function updateStatus($submission)
...
+     *
+     * @param $user User
+     * @param $submission Submission
+     *
+     * @return bool
+     */
+    public function canPreview($user, $submission)
+    {
+        // Don't grant access until submission reaches Copyediting stage
+        if ($submission->getData('stageId') < WORKFLOW_STAGE_ID_EDITING) {
```
asmecher, line 957: I don't recommend using `<>` comparisons on defined constants -- the numerical values of these aren't intended to have significance beyond being unique. Better to use `in_array` and list the options, I think.

asmecher (changes requested):

classes/services/PKPSubmissionService.inc.php
```diff
@@ -941,4 +941,66 @@ public function updateStatus($submission)
...
+    /**
+     * Checks if this user is granted access to preview
+     * based on their roles in the context (i.e. Manager, Editor, etc).
+     *
+     * @param $user User
+     * @param $submission Submission
+     *
+     * @return bool
+     */
+    public function _roleCanPreview($user, $submission)
```
NateWr, line 985: Make this a `protected` function. We don't want to expose it as part of the submission service API.

ajnyga:
Running tests but the last ones ended up with just OPS failing and that should be fixed now (just a typo in a new Cypress test).

pkp-lib: https://github.com/pkp/pkp-lib/pull/7011
OJS: https://github.com/pkp/ojs/pull/3125
OMP: https://github.com/pkp/omp/pull/970
OPS: https://github.com/pkp/ops/pull/397

This is for both https://github.com/pkp/pkp-lib/issues/5299 and https://github.com/pkp/pkp-lib/issues/8090
"two flies with one strike" (I know @asmecher  likes to collect these!)

classes/submission/Repository.php
```diff
@@ -576,4 +608,34 @@ protected function getCurrentPublicationIdByPublications(Submission $submission)
...
+     */
+    protected function _roleCanPreview(User $user, Submission $submission): bool
+    {
+        $roleDao = DAORegistry::getDAO('RoleDAO'); /* @var $roleDao RoleDAO */
+        if ($user && $submission) {
+            $subscriptionAssumedRoles = [
+                ROLE_ID_MANAGER,
+                ROLE_ID_SUB_EDITOR,
+                ROLE_ID_ASSISTANT,
+                ROLE_ID_SUBSCRIPTION_MANAGER
```
asmecher, line 629: This role ID is only defined in pkp-lib, so I suppose this would probably cause a fatal error (in recent versions of PHP) in OMP and OPS.

ajnyga: So `Role::ROLE_ID_MANAGER` instead, right?

ajnyga: added `Role::`

asmecher: No, `ROLE_ID_SUBSCRIPTION_MANAGER` is an OJS-specific role (and not the same thing as `ROLE_ID_MANAGER`). This would need to be included only in OJS, so I suppose you'll want to use the OJS subclass to augment the pkp-lib list with `ROLE_ID_SUBSCRIPTION_MANAGER`.

Issue #7034, opened 2021-05-07
Import/Export plugin - remove get_class dependency for filter retrieval 


defstat:
**Describe the bug**
There are some cases where the `get_class()` function is being used for retrieving class specific filter.  Namespaces introduction may result to unspecified filter when the get_class is being used

**What application are you using?**
 OJS, OMP or OPS main branch

defstat:
@asmecher @NateWr 

**PR**
[pkp-lib]: #7035

asmecher:
No, closing!

PR #7035, opened 2021-05-07
[pkp-lib] #7034 Remove get_class dependency for retrieving File filter

defstat:


asmecher:
Looks good, please merge. There will be some like this that are hard to find and will break when we comprehensively add namespaces!

PR #7049, opened 2021-05-10
pkp/pkp-lib#4622 Enabling Laravel's queue worker at register_shutdown_function

henriqueramos:


classes/core/PKPQueueProvider.inc.php
```diff
@@ -0,0 +1,75 @@
...
+            ::isavailable()
+                ->notexcedeedattempts()
+                ->limit(1)
+                ->first();
+
+        if (!$job) {
+            return;
+        }
+
+        // $artisan = app()->make('Illuminate\Support\Facades\Artisan');
```
henriqueramos, line 45: One of my first ideas, because at this point `app()` was already instantiated. But didn't work due to some notInstantiable errors.

I'm digging the rabbit hole further more.

classes/migration/FailedJobsMigration.inc.php
```diff
@@ -0,0 +1,42 @@
...
+use Illuminate\Support\Facades\Schema;
+
+class FailedJobsMigration extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up()
+    {
+        Schema::create('failed_jobs', function (Blueprint $table) {
```
henriqueramos, line 25: Created a failed_jobs, thus we could do a better tracking of it.

classes/core/PKPContainer.inc.php
```diff
@@ -41,8 +46,8 @@ protected function registerBaseBindings()
         static::setInstance($this);
         $this->instance('app', $this);
         $this->instance(Container::class, $this);
-        $this->singleton(Illuminate\Contracts\Debug\ExceptionHandler::class, function () {
```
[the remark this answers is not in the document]
Vitaliy-1: Thanks. You may see other similar namespacing issues that originate from recent Alec's work. Sorry, haven't reviewed it yet.

classes/core/PKPQueueProvider.inc.php
Vitaliy-1: Likely also a namespacing issue. See https://github.com/Vitaliy-1/pkp-lib/commit/a8b4fcda4a094ccbef0d50fe4ec07c2720ac3d8e, haven't merged this yet

classes/core/PKPContainer.inc.php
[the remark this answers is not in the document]
asmecher: Yes, it just needs a `\` prefix now that the `PKPContainer` class is in a namespace.

composer.json
```diff
@@ -62,5 +63,11 @@
 				"./tools/pre-commit.sh"
 			]
 		}
-	}
+    },
+    "autoload": {
+        "psr-4": {
```
asmecher, line 67: The *only* reason we can't adopt PSR4 autoloading is that we're using `.inc.php` extensions rather than `.php`. I'd definitely like to move to PSR4 but this might involve a mass rename and some tricky scheduling. Let's talk about this during the dev call.

henriqueramos: Indeed. But for the new things we could opt for using the PSR-4 directly.

asmecher: Yes -- but better than that, we might be able to just rename everything in core code (since we're using classloading) and leave the deprecated `import` function to handle legacy `.inc.php` imports. Needs thought/experimentation.

classes/services/PKPPublicationService.inc.php
```diff
@@ -624,22 +652,39 @@ public function publish($publication)
...
+        );
         event(new PublishedEvent($newPublication, $publication, $submission));
 
         // Update the search index.
-        if ($newPublication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
-            Application::getSubmissionSearchIndex()->submissionMetadataChanged($submission);
-            Application::getSubmissionSearchIndex()->submissionFilesChanged($submission);
-            Application::getSubmissionSearchDAO()->flushCache();
-            Application::getSubmissionSearchIndex()->submissionChangesFinished();
+        if (itsPublished) {
```
asmecher, line 686: Missing `$`

classes/services/PKPPublicationService.inc.php
```diff
@@ -624,22 +652,39 @@ public function publish($publication)
 
         // Log an event when publication is published. Adjust the message depending
         // on whether this is the first publication or a subsequent version
-        if (count($submission->getData('publications')) > 1) {
-            $msg = $newPublication->getData('status') === PKPSubmission::STATUS_SCHEDULED ? 'publication.event.versionScheduled' : 'publication.event.versionPublished';
-        } else {
-            $msg = $newPublication->getData('status') === PKPSubmission::STATUS_SCHEDULED ? 'publication.event.scheduled' : 'publication.event.published';
-        }
-        SubmissionLog::logEvent(Application::get()->getRequest(), $submission, SubmissionEventLogEntry::SUBMISSION_LOG_METADATA_PUBLISH, $msg);
+        $availableMessages = [
```
asmecher, line 655: It's tough to review code when there are new concepts mixed in with general clean-up; can you keep clean-up to separate PRs when working on something like this? I see the long lines and ternaries you're working to clean up, but the array lookup syntax in `$availableMessages` isn't clearer to me.

classes/services/PKPPublicationService.inc.php
```diff
@@ -624,22 +652,39 @@ public function publish($publication)
...
         event(new PublishedEvent($newPublication, $publication, $submission));
 
         // Update the search index.
-        if ($newPublication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
-            Application::getSubmissionSearchIndex()->submissionMetadataChanged($submission);
-            Application::getSubmissionSearchIndex()->submissionFilesChanged($submission);
-            Application::getSubmissionSearchDAO()->flushCache();
-            Application::getSubmissionSearchIndex()->submissionChangesFinished();
+        if (itsPublished) {
+            dispatch(new UpdateSubmissionSearchJob($submission->getId()));
```
asmecher, line 687: I think this should tie in more closely with @Vitaliy-1's work (https://github.com/pkp/pkp-lib/issues/6970 and the `PublishEvent` above) -- i.e. the job should be processed as part of the `PublishEvent`.

asmecher: (There are also other calls to re-index full text that should get the same treatment -- e.g. in `lib/pkp/controllers/api/file/PKPManageFileApiHandler.inc.php`.)

asmecher:
Thanks, @henriqueramos, this is coming along nicely!

To keep things linking between repos nicely on Github, rather than using `[4622]` as the prefix in your commits, please use `pkp/pkp-lib#4622`.

I think you've introduced some concepts here that'll mean some reconciliation with the existing code, for example...
- By this proposal, I think you intend to remove the acron plugin entirely and instead use a new `config.inc.php` option? The alternative -- which would be more consistent for existing users, who won't expect to have to rearrange their configuration -- would be to keep the existing acron plugin but change it to use the new `register_shutdown_function` instead.
- Where new `config.inc.php` parameters are introduced, they should be documented in `docs/RELEASE`. (See `docs/release-notes` for other examples.)

Vitaliy-1:
Wouldn't it be better to make those jobs listen to an event rather than being initialized in an Event constructor? 
(Sorry if I missed a discussion on this earlier)

henriqueramos:
> Wouldn't it be better to make those jobs listen to an event rather than being initialized in an Event constructor?
> (Sorry if I missed a discussion on this earlier)

The idea behind the PR it's save the serialized job as a database record to be used in the "queuing" system.

Vitaliy-1:
Is the idea those jobs to be queued when an event fires, like [queue event listeners](https://laravel.com/docs/7.x/events#queued-event-listeners) or something different?

henriqueramos:
> Is the idea those jobs to be queued when an event fires, like [queue event listeners](https://laravel.com/docs/7.x/events#queued-event-listeners) or something different?

Yes, I've used both concepts in the end.

When some events are called, a listener will be called, and inside this listener, a job will be inserted into database.

tests/config/config.TEMPLATE.pgsql.inc.php
```diff
@@ -100,3 +100,8 @@
 
 [debug]
 show_stacktrace = On
+
+[queues]
+
+default_queue = null
```
asmecher, line 106: Is there a practical case for this to have another value?

Domains/Jobs/Job.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        return $this->maxAttempts;
+    }
+
+    /**
+     * Add a local scope for not excedeed attempts
+     *
+     *
+     */
+    public function scopeNotExcedeedAttempts(Builder $query): Builder
```
asmecher, line 168: Typo: should be "scopeNotExceededAttempts", I think.

Domains/Jobs/Traits/Worker.php
```diff
@@ -0,0 +1,189 @@
...
+     * Get Job's timeout value
+     *
+     */
+    public function getTimeout(): int
+    {
+        return $this->timeout;
+    }
+
+    /**
+     * Set Job's sleep value
```
asmecher, line 128: This is a number of seconds, I think? Could you add that to the self-documentation? (Units in some of the others might be helpful too, like `allowedMemory` being megabytes)

Domains/Jobs/Traits/Worker.php
```diff
@@ -0,0 +1,189 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file Domains/Jobs/Traits/Worker.php
```
asmecher, line 6: This class mirrors the defaults in the constructor to [`WorkerOptions`](https://laravel.com/api/5.8/Illuminate/Queue/WorkerOptions.html#method___construct), right? Are we overriding these anywhere? If not, and this is for future potential use but not a current use case yet, I'd favour not adding it.

henriqueramos: This trait enables our Job Model (Domains/Jobs/Job.php) to customized the parameters for the WorkerOptions inside the PKPQueueProvider (classes/core/PKPQueueProvider.inc.php).

tests/config/config.TEMPLATE.pgsql.inc.php
henriqueramos: Yes, if you're using a Message broker, you could put add the default queue here.

henriqueramos: I've changed it to use as value the string "test_default_queue".

Domains/Jobs/Traits/Worker.php
asmecher: OK, but are there cases yet where this is done?

asmecher: (^ @henriqueramos)

classes/core/PKPQueueProvider.inc.php
```diff
@@ -0,0 +1,67 @@
...
+    public function runJobsAtShutdown(): void
+    {
+        $disableRun = Config::getVar('queues', 'disable_jobs_run_at_shutdown', false);
+
+        if ($disableRun) {
+            return;
+        }
+
+        $job = PKPJobModel
+            ::isavailable()
```
asmecher, line 36: Is this `alllowercase` a Laravel convention? Generally we use `lowerCamelCase` for function naming.

classes/observers/events/UnpublishedEvent.inc.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @var PKPPublication $newPublication The publication being published */
+    public $newPublication;
+
+    /** @var PKPPublication $publication Old publication, before processing */
+    public $publication;
+
+    /** @var PKPSubmission $submission Submission associated with the publication */
+    public $submission;
+
+    public function __construct(PKPPublication $newPublication, PKPPublication $publication, PKPSubmission $submission)
```
asmecher, line 38: Needs code header with descriptions of parameters.

tests/classes/queues/QueueTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @see PKPTestCase::setUp()
+     */
+    protected function setUp(): void
+    {
+        $this->configData = Config::getData();
+
+        if ($this->configData['queues']['disable_jobs_run_at_shutdown']) {
+            $this->markTestSkipped('Config [\'queues\'][\'disable_jobs_run_at_shutdown\'] isnt disabled.');
```
asmecher, line 37: *isn't

classes/migration/install/FailedJobsMigration.inc.php
```diff
@@ -0,0 +1,44 @@
...
+
+class FailedJobsMigration extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up()
+    {
+        Schema::create('failed_jobs', function (Blueprint $table) {
+            $table->bigIncrements('id');
```
asmecher, line 28: For consistency with our database design elsewhere in the system, could you call this `failed_job_id`?

Domains/Jobs/Exceptions/JobException.php
```diff
@@ -0,0 +1,35 @@
...
+ * @brief Exception for Job domain
+ */
+
+namespace PKP\Domains\Jobs\Exceptions;
+
+use Exception;
+
+class JobException extends Exception
+{
+    public const INVALID_PAYLOAD = 'invalid.payload';
```
asmecher, line 24: This is just to avoid coding the same string when creating the exception, right? I don't mind if the same string appears in a few places, but either way, this would need to be a little more helpful to someone who ran into it.

tests/config/config.TEMPLATE.mysql.inc.php
```diff
@@ -100,3 +100,8 @@
 
 [debug]
 show_stacktrace = On
+
+[queues]
```
asmecher, line 104: Don't forget this comment:
https://github.com/pkp/pkp-lib/pull/7049#issuecomment-848828840

asmecher (changes requested):
Just a few things to look over, then I'd suggest having @Vitaliy-1 look over it.

plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
```diff
@@ -205,15 +205,11 @@ public function &process(&$document)
...
         // Index imported content
-        $submissionSearchIndex = Application::getSubmissionSearchIndex();
         foreach ($importedObjects as $submission) {
             assert($submission instanceof Submission);
-            $submissionSearchIndex->submissionMetadataChanged($submission);
-            $submissionSearchIndex->submissionFilesChanged($submission);
+            event(new MetadataChanged($submission));
         }
 
-        $submissionSearchIndex->submissionChangesFinished();
```
asmecher, line 215: The `submissionChangesFinished` function should be called after *all* submission files have been updated (in the case of a batch update); this is currently used by the Lucene plugin (and possibly others) to trigger re-indexing of content after a batch operation. I see you've moved it into the jobs that handle individual index changes (e.g. https://github.com/pkp/pkp-lib/pull/7049/files#diff-b954038ae0729b3fba218bfa9a9b05ad358a4e9b5104fdf01392be335cfbb24bR67). This will work fine for cases where only a single submission is changing, but in cases like this where a batch is being processed, it'll slow the indexing down considerably.

henriqueramos: Thank you for the explanation.

And I think, the unique job only works on Laravel 8.x. We are yet on the 7.x. I will check.

classes/migration/install/FailedJobsMigration.inc.php
henriqueramos: This is a copy of the [Laravel's default failed_jobs table](https://github.com/illuminate/queue/blob/7.x/Console/stubs/failed_jobs.stub). I'm afraid about changing it and crashing the framework internals.

asmecher: OK, let's leave it as `id` then, but could you note that source in the file so we know (and credit) where it came from?

Domains/Jobs/Exceptions/JobException.php
henriqueramos: Indeed. Also, I'm using this kind of logic for the past 5 years, it's easier to find where is called (`JobException::INVALID_PAYLOAD`), instead the pure string (invalid.payload). Became a unique placeholder.

asmecher: OK, but if end users encounter this message they won't have any idea what's happening (unless they search the code). And a more helpful message can be searched the same way as this unique placeholder. Could you make this a human-readable string? If we adopt this kind of placeholder, we should consider it as a general approach, rather than just adding it here. Otherwise the codebase can get really inconsistent with lots of new ideas that are never adopted beyond experiments.

Jobs/Metadata/BatchMetadataChangedJob.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Execute the job.
+     *
+     */
+    public function handle(): void
+    {
+        $successful = 0;
+
+        $articleSearchIndex = Application::getSubmissionSearchIndex();
```
asmecher, line 50: Could you rename this to `submissionSearchIndex` instead of `articleSearchIndex`? Best to use application-agnostic language in pkp-lib.

asmecher: (Also elsewhere)

classes/core/PKPEventServiceProvider.inc.php
```diff
@@ -28,7 +41,26 @@ class PKPEventServiceProvider extends EventServiceProvider
      * @var array $listen $event => $listeners[]
      * @brief Registering events & listeners, see Illuminate\Events\EventServiceProvider
      */
-    protected $listen = [];
+    protected $listen = [
```
Vitaliy-1, line 44: I recommend using Event Discovery. One of the recent PRs adds file caching to resolve possible performance issues. The only thing is when testing newly added listener to not forget to clear the cache. Sometimes I use Nate's plugin to clear the cache on every request: https://github.com/NateWr/cacheBuster

classes/observers/events/BasePublicationEvent.inc.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+namespace PKP\observers\events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+
+use PKP\publication\PKPPublication;
+use PKP\submission\PKPSubmission;
+
+class BasePublicationEvent
```
Vitaliy-1, line 25: Is `BasePublicationEvent` used as an actual event somewhere? If not, I'd recommend moving it away from the `events` folder and making abstract

classes/observers/listeners/SubmissionUpdatedListener.inc.php
```diff
@@ -0,0 +1,34 @@
...
+use PKP\observers\events\BasePublicationEvent;
+
+class SubmissionUpdatedListener
+{
+    /**
+     * Handle the listener call
+     *
+     *
+     */
+    public function handle(BasePublicationEvent $event)
```
Vitaliy-1, line 30: Interesting approach, it allows to avoid code duplication; I'd probably use an event subscriber here to listen to Publish/Unpublish events separately to avoid ambiguities.

classes/core/PKPEventServiceProvider.inc.php
Vitaliy-1: (listeners use the same type-hinting for event discovery as per Laravel's documentation)

classes/observers/listeners/BatchMetadataChangedListener.inc.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @brief Listener fired when submission metadata's changed on batch
+ */
+
+namespace PKP\observers\listeners;
+
+use PKP\Jobs\Metadata\BatchMetadataChangedJob;
+use PKP\observers\events\BatchMetadataChanged;
+
+class BatchMetadataChangedListener
```
Vitaliy-1, line 23: I'd be happy to discuss naming patterns for event and listener classes for consistency. I recommend avoiding the last part `Listener` in the class name as the namespace already includes that information. The same for events,  `PublishedEvent` could be renamed to something like `PublicationPublished` (that was created by me before taking namespaces into consideration). I also believe that listener should perform only one task and its action should be reflected in the class name, so one event could have many listeners but we should avoid creating a universal `multitasked`  listener; but I don't have a strong opinion here. What do you think?

Vitaliy-1 (changes requested):
Looks good! I left a couple of comments.

classes/core/PKPEventServiceProvider.inc.php
henriqueramos: The only downside with the Event Discovery it's the extensive Reflection usage. An operation like that multiple times it's slowly.

Also, hard coding the most common events and listeners help other developers to understand the flows.

classes/observers/events/BasePublicationEvent.inc.php
henriqueramos: It's extended at `PublishedEvent` and `UnpublishedEvent`.

classes/observers/listeners/BatchMetadataChangedListener.inc.php
henriqueramos: We could create a forum to discuss this kind of naming and come with some standard rules. Because, as Uncle Bob mentioned on 'Clean Code', naming things are hard.

I usually opt to use `Listener` as part of the class name to avoid problems during autocomplete and `use` dependency injections.

The `PublishedEvent` was named thus to become more generic across OMP/OJS/OPS.

And I agree with the rule of "One Handle Action per Listener".

classes/observers/listeners/SubmissionUpdatedListener.inc.php
henriqueramos: Yes! My main idea behind here it's to avoid code duplication.

In fact, `PublishedEvent` and `UnpublishedEvent` are called in different flows, but internally they do a similar job, update the search engine. So I've opt to hook both to call `SubmissionUpdatedListener` on `PKPEventServiceProvider::$listen`, but keeping their own "personality".

classes/core/PKPEventServiceProvider.inc.php
Vitaliy-1: This is documented in Laravel docs and those are grouped in designated places in the code so, I think, should ok from a perspective for others to understand the concept. 

The results of calls to Reflection API, in the case of Event Discovery, are cached and it wouldn't be called on every request.

classes/observers/events/BasePublicationEvent.inc.php
Vitaliy-1: Yeah, I lean towards keeping in `observers/events` only those that are used in the system and, in this case, moving `BasePublicationEvent` from there with declaring this class as abstract.

classes/observers/listeners/SubmissionUpdatedListener.inc.php
Vitaliy-1: My concern here is that looking at the `SubmissionUpdatedListener::handle()` it's not clear to which event it listens. The ambiguity for me here is that it does not listen to `BasePublicationEvent` but to one of its ancestors.

classes/core/PKPEventServiceProvider.inc.php
henriqueramos: I came across a doubt about this approach of caching the discovered events without [extra commands](https://laravel.com/docs/7.x/events#event-discovery) like `event:list`, `event:clear` and `event:cache`.

We should add those ideas on our CLI tools.

Also, add a flag to enable or not the caching development.

tests/config/config.TEMPLATE.mysql.inc.php
henriqueramos: @asmecher I had a brief conversation with @NateWr. He told me about some script do create the `docs/RELEASE` correctly.

NateWr: Sorry @henriqueramos! That was my misunderstanding. I thought you were talking about the list of issues under each release. I think for configuration parameters you still need to manually add this information.

Alec, should he be creating a new document under docs/release-notes/README-3.4.0 or replacing docs/RELEASE with something new?

asmecher: There is already a `docs/release-notes/README-3.4.0`; I've synced it with `docs/RELEASE`. Please add configuration change documentation to both.

classes/observers/listeners/SubmissionUpdatedListener.inc.php
henriqueramos: I've added a `subscribe` method on this Listener. This probably will let users know better which events are associated with it.

classes/observers/events/BasePublicationEvent.inc.php
asmecher: @henriqueramos and @Vitaliy-1, is this resolved?

tests/phpunit-env1.xml
```diff
@@ -1,10 +1,40 @@
...
+    convertWarningsToExceptions="true"
+    processIsolation="false"
+    stopOnFailure="true"
+    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
+    <testsuites>
+        <testsuite name="Common Test Suite">
+            <directory suffix="Test.php">classes</directory>
+        </testsuite>
+    </testsuites>
+    <coverage>
```
asmecher, line 21: Does the code coverage stuff work? We used to have a test suite for this using Openstack but it's long since removed.

tools/events.php
```diff
@@ -0,0 +1,227 @@
...
+                array_keys(self::AVAILABLE_OPTIONS)
+            );
+        }
+
+        $this->{$this->option}();
+    }
+}
+
+try {
+    $tool = new commandEvents($argv ?? []);
```
henriqueramos, line 197: @asmecher After seeing the #7171, I thought the command belongs to 4622, because he depends more on the EventServiceProvider caching.

tests/phpunit-env1.xml
henriqueramos: Yes. I ran the following command on OMP and generated the following report.

```
export XDEBUG_MODE=coverage && php lib/pkp/lib/vendor/phpunit/phpunit/phpunit --configuration lib/pkp/tests/phpunit-env1.xml  --filter testRoute --coverage-html "lib/pkp/tests/results/coverage-html"
```
Screenshot from 2021-12-13 22-47-06

classes/observers/events/BasePublicationEvent.inc.php
Vitaliy-1: @henriqueramos, sorry for not being clear during our meeting.
To avoid code duplication in the case of Event classes, I recommend using traits rather than extending a base class. E.g., something like:
```php
namespace PKP\observers; // outside events or listeners dir

trait Publicationable
{
  public PKPPublication $newPublication;
  public PKPPublication $publication;
  public PKPSubmission $submission;

  public function __construct(PKPPublication $newPublication, PKPPublication $publication, PKPSubmission $submission)
  {
     ...
  }
}
```
and then we can use:
```php
class PublishedEvent
{
	use Dispatchable;
	use Publicationable;
}
```
and for the listener part:
```php
class SubmissionUpdated
{
   /**
   * @param \PKP\observers\events\PublishedEvent $event
   */
  public function handle(PublishedEvent $event)
  {
     dispatch(new UpdateSubmissionSearchJob($event->submission->getId()));
   }
}
```
or if it listening to several events, use subscriber
```php
class SubmissionUpdated
{
   public function subscribe(Dispatcher $events): void
  {
     $events->listen(
         PublishedEvent::class,
         self::class . '@handlePublished'
     );

     $events->listen(
         UnpublishedEvent::class,
         self::class . '@handleUnpublished'
     );
   }

  public function handlePublished(PublishedEvent $event)...

  public function handleUnpublished(UnpublishedEvent $event)... 
}
```
This makes it clear which event listener actually listens.
Let me know if you'd need some additional info here

Domains/Jobs/Exceptions/JobException.php
henriqueramos: Changed the output message.

classes/observers/events/BasePublicationEvent.inc.php
henriqueramos: Thank you for the idea of using Trait on this @Vitaliy-1! :tada:

Jobs/Submissions/RemoveSubmissionFromSearchIndexJob.php
```diff
@@ -0,0 +1,62 @@
...
+        $submission = Repo::submission()->get($this->submissionId);
+
+        if (!$submission) {
+            $this->failed(new JobException(JobException::INVALID_PAYLOAD));
+
+            return;
+        }
+
+        $submissionSearchIndex = Application::getSubmissionSearchIndex();
+        $submissionSearchIndex->articleDeleted($submission->getId());
```
asmecher, line 59: This function is OJS-only and will fail if called in OMP or OPS; one resolution could be to move this part to the OJS repo only, but a better solution would be to add a common function in `SubmissionSearchIndex` that works for all 3 apps.

classes/migration/install/FailedJobsMigration.inc.php
```diff
@@ -0,0 +1,47 @@
...
+use Illuminate\Support\Facades\Schema;
+
+class FailedJobsMigration extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up()
+    {
+        if (!Schema::hasTable('failed_jobs')) {
```
asmecher, line 27: Is there a reason this condition is present? The other `install` migrations don't check for table existence.

classes/observers/traits/Publicationable.inc.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file classes/observers/events/BasePublicationEvent.inc.php
```
asmecher, line 6: (Self documentation needs updating)

classes/observers/traits/Publicationable.inc.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @brief Base event for publications being published or unpublished
+ */
+
+namespace PKP\observers\traits;
+
+use PKP\publication\PKPPublication;
+use PKP\submission\PKPSubmission;
+
+trait Publicationable
```
asmecher, line 23: Should this be `Publishable` rather than `Publicationable`?

tests/classes/core/PKPComponentRouterTest.php
```diff
@@ -260,6 +260,8 @@ public function testRoute()
         // Simulate context DAOs
         $this->_setUpMockDAOs('context1', 'context2');
 
+        $this->expectOutputRegex('/"status":false/');
```
asmecher, line 263: Is this an intentional change? I wouldn't expect this PR to affect the router code.

classes/migration/install/FailedJobsMigration.inc.php
henriqueramos: This migration file it's used for `install` and `upgrade` flows. I added it to handle scenarios where this table exists already.

asmecher: If the schema in this file ever changes, that'll lead to silent failures where the `hasTable` check skips creating the already-existing table but the SQL expects a different table structure. I'd rather leave out the condition and have it fail than have to debug a missing schema change later.

classes/observers/traits/Publicationable.inc.php
Vitaliy-1: Sorry for my limited English. That was in my example, meaning trait which suits for different actions with publication rather than only publish/unpublish. But if it's only for these 2 events, `Publishable` is definitely better.

classes/migration/install/FailedJobsMigration.inc.php
henriqueramos: Okay!

Without versioning the migrations ([like default Laravel and his timestamp named migrations did](https://laravel.com/docs/8.x/migrations#generating-migrations)), removing this `hasTable` is the best approach, so far.

tests/classes/core/PKPComponentRouterTest.php
henriqueramos: This unit test was failing.

asmecher: Can you write a little more about that? I'd worry that the PR added an unexpected behaviour, in which case rather than changing the tests, it would be better to check the behaviour.

henriqueramos: While running the following command:
```
export XDEBUG_MODE=coverage && php lib/pkp/lib/vendor/phpunit/phpunit/phpunit --configuration lib/pkp/tests/phpunit-env1.xml  --filter testRoute --coverage-html "lib/pkp/tests/results/coverage-html" --debug --verbose
```

Without this `$this->expectOutputRegex('/"status":false/');` lead us to an error (attached below):
Screenshot from 2021-12-18 20-29-21

With it, the test passes.
Screenshot from 2021-12-18 20-31-15

asmecher: That suggests to me that your changes are introducing output where there was none before. Have you looked into where that's happening?

henriqueramos: I'm tracking down. Until Monday I will got more info.

Domains/Jobs/Traits/Worker.php
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @class Worker
+ * @ingroup domains
+ *
+ * @brief Worker trait for Jobs model
+ */
+
+namespace PKP\Domains\Jobs\Traits;
+
+trait Worker
```
asmecher, line 20: Could you document the parameters here to include e.g. units? (Seconds, megabytes, etc.)

henriqueramos: Done.

Issue #6963, opened 2021-04-19
Improve OAI performance


asmecher:
There are FIXMEs in PKPOAIDAO regarding inefficiency in seeking through a result set to the correct subset. For large repositories these will perform poorly.

asmecher:
PRs:
- https://github.com/pkp/pkp-lib/pull/7073
- https://github.com/pkp/ojs/pull/3134

@bozana, would you mind taking a look at this (before I adapt them for OPS and OMP)?  Essentially they convert the raw SQL to Laravel's querybuilder which permits better augmentation of the queries by calling code for counts etc. This doesn't improve the performance of the DRIVER plugin, though -- that'll still need attention, but I'm kind of waiting for someone to tell me it's a problem first.

jonasraoni:
@asmecher, one thing that I've noticed in the wild: The "Identity" verb of https://raco.cat (OJS 3.1.1.4) takes a long time to return, but other "limited" verbs (`ListSets`, `ListRecords` with a `set` argument) are ok.

asmecher:
> The "Identity" verb of https://raco.cat (OJS 3.1.1.4) takes a long time to return

That's likely due to the query determining the date of the first record. To make that query performant we'll either need to use functional indexes (not yet supported in our MySQL baseline), or move to using soft deletes (Illuminate supports these nicely but we don't use them yet). Or cache that value somewhere.

asmecher:
@bozana, I've addressed your review comments; could you take a second look? Thanks!

asmecher:
Thanks, @bozana! I've addressed all your comments; in addition to the earlier PRs...
- https://github.com/pkp/pkp-lib/pull/7073
- https://github.com/pkp/ojs/pull/3134

...I've ported this to OPS and OMP:
- https://github.com/pkp/ops/pull/162
- https://github.com/pkp/omp/pull/983

I don't think these need separate review, so I'll merge them.

PR #7073, opened 2021-05-21
pkp/pkp-lib#6963 Improve OAI performance

asmecher:


classes/oai/PKPOAIDAO.inc.php
```diff
@@ -291,10 +273,12 @@ public function _doCommonOAIFromRowOperations($record, $row)
...
      * @param $orderBy string UNFILTERED
+     * @param $offset int? Query offset
+     * @param $limit int? Query limit
      *
-     * @return Iterable
+     * @return \Illuminate\Database\Query\Builder
      */
-    abstract public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId = null, $orderBy = 'journal_id, submission_id');
+    abstract public function _getRecordsRecordSetQuery($setIds, $from, $until, $set, $submissionId = null, $orderBy = 'journal_id, submission_id', $offset = null, $limit = null);
 }
```
bozana, line 282: I think we do not need `$offset` and `$limit` parameters for this function -- as you did above the total/count, offset and limit are called on the results of the function?

bozana (commented):
@asmecher, just a comment/question

classes/oai/PKPOAIDAO.inc.php
```diff
@@ -143,22 +143,13 @@ public function getRecord($dataObjectId, $setIds = [])
      */
     public function getRecords($setIds, $from, $until, $set, $offset, $limit, &$total)
     {
-        $result = $this->_getRecordsRecordSet($setIds, $from, $until, $set);
+        $query = $this->_getRecordsRecordSetQuery($setIds, $from, $until, $set);
+        $total = $query->count();
+        $results = $query->offset($offset)->limit($limit)->get();
```
jonasraoni, line 148: Considering the OAI doesn't have a "pageNumber" parameter and that the result set is already sorted by IDs, it's possible to achieve a better performance by replacing the `offset` (the more you skip, the slower it gets) by a `where` clause (that [Laravel's method](https://laravel.com/api/7.x/Illuminate/Database/Query/Builder.html#method_chunkById) we discussed in a Beacon issue might cover it).

asmecher: That's a good idea, but it'll require us to change the way OAI resumption tokens are stored so that we know the first ID to fetch on subsequent calls. For now the changes here and in https://github.com/pkp/ojs/pull/3134 at least move the iteration out of PHP, which will resolve the bottleneck noted in the FIXME. We can come back to this when it shows up as a new bottleneck.

jonasraoni: I think the offset/limit will be enough, unless there are installations with millions of records out there 🤔
In case we have access to such installation it would be good to run a query just to check the performance of retrieving the last page.

bozana: How would we could then get the total numbers?

PR #7106, opened 2021-06-07
pkp/pkp-lib#7105 Allow queued jobs to be processed by a cron job

henriqueramos:
This is a `sub pull request` for the pkp/pkp-lib#4622 [PR 3132](https://github.com/pkp/ojs/pull/3132).


henriqueramos:
Keeping in mind this should only be merged after [PR 3132](https://github.com/pkp/ojs/pull/3132)

asmecher:
(https://github.com/pkp/pkp-lib/pull/7049 is the pkp-lib counterpart to https://github.com/pkp/ojs/pull/3132.)

Jobs/Metadata/MetadataChangedJob.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    protected $submissionId;
+
+    /**
+     * Create a new job instance.
+     *
+     */
+    public function __construct(int $submissionId)
+    {
+        $this->connection = Config::getVar('queues', 'default_connection', 'sync');
```
asmecher, line 47: I think you're trying to provide for use of both `sync` and `database` queues, but the `database` queue should always be available, and I think we'll be dependent enough on it that we should focus on making it work in all deployments (e.g. shutdown function as well as cron and daemon-based approaches) rather than planning to fall back on `sync`.

Jobs/Submissions/DeleteSubmissionFileJob.php
```diff
@@ -0,0 +1,82 @@
...
+ * @file Jobs/Submissions/DeleteSubmissionFileJob.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class DeleteSubmissionFileJob
+ * @ingroup jobs
+ *
+ * @brief Classe to handle the Submission File deletion as a Job
```
asmecher, line 15: *Class

Jobs/Submissions/DeleteSubmissionFileJob.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file Jobs/Submissions/DeleteSubmissionFileJob.php
```
asmecher, line 6: I see a bit of inconsistency with job naming ("Delete" vs. "Deleted" in `DeletedSubmissionSearchJob` vs. `DeleteSubmissionFileJob`). Likewise below for events (`DeleteSubmissionFile` vs. `DeletedSubmission`).

Since the code that triggers the event will handle doing the named action (e.g. deleting a submission file) outside of the event and job, we should be careful not to suggest that it's the event and job that do the task. So maybe this makes sense:

- Event names: Name these for what has happened and use past tense (e.g. `SubmissionDeleted`)
- Job names: Name these for what the job itself is doing (e.g. `RemoveSubmissionFileIndex`)

I'm not so sure about a convention for the listener. For example, currently there's one called `DeleteSubmissionFileListener` and it associates the event and the job. If there are several jobs, the same listener will be used to associate all of them, correct?

classes/core/PKPEventServiceProvider.inc.php
```diff
@@ -27,7 +38,23 @@ class PKPEventServiceProvider extends EventServiceProvider
      * @var array $listen $event => $listeners[]
      * @brief Registering events & listeners, see Illuminate\Events\EventServiceProvider
      */
-    protected $listen = [];
+    protected $listen = [
```
asmecher, line 41: I can imagine that this declaration will grow really long if we adopt this practice throughout the software. I'm guessing there are other ways to associate events with listeners?

henriqueramos: We could move to use [`Event Subscribers`](https://laravel.com/docs/7.x/events#event-subscribers) entirely.

Jobs/Submissions/DeleteSubmissionFileJob.php
henriqueramos: > I'm not so sure about a convention for the listener. For example, currently there's one called DeleteSubmissionFileListener and it associates the event and the job. If there are several jobs, the same listener will be used to associate all of them, correct?

Yes.

asmecher: OK, then it makes sense to me that the listener is named for the thing that happened (e.g. `SubmissionDeleted`), rather than what needs to happen as a result.

classes/core/PKPEventServiceProvider.inc.php
asmecher: Those subscriptions would need to be scanned and cached, correct?

Domains/Jobs/Job.php
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Get queue's size
+     *
+     *
+     */
+    public function size(?string $queue = null): int
+    {
+        return $this->queuedat($this->getQueue($queue))
```
asmecher, line 203: (Should this be `queuedAt`? Not sure where this is implemented, suspecting a `Model` convention.)

Domains/Jobs/Traits/Attributes.php
```diff
@@ -0,0 +1,126 @@
...
+ * @brief Attributes trait for Jobs model
+ */
+
+namespace PKP\Domains\Jobs\Traits;
+
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Date;
+
+/**
+ * Those attributes become from payload array
```
asmecher, line 24: Typo: "come from", I think

Domains/Jobs/Traits/Attributes.php
```diff
@@ -0,0 +1,126 @@
...
+            return null;
+        }
+
+        return $this->payload['timeout'];
+    }
+
+    /**
+     * Return the job's timeout at value
+     *
+     * @return mixed
```
asmecher, line 87: ?Date

Domains/Jobs/Traits/Attributes.php
```diff
@@ -0,0 +1,126 @@
...
+            return null;
+        }
+
+        return $this->payload['delay'];
+    }
+
+    /**
+     * Return the job's timeout value
+     *
+     * @return mixed
```
asmecher, line 73: This `@return` doesn't match the function (`?string`) and I think when formally declaring return types we can just skip the `@return` self-doc.

asmecher: (Also elsewhere)

Domains/Jobs/Traits/Attributes.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file Domains/Jobs/Traits/Attributes.php
```
asmecher, line 6: I'm still finding this trait a little extraneous; unless I'm reading it wrong, the defaults are always used, no? I'm also aware that you're introducing some new concepts here (e.g. `Eloquentable`) that should be considered for adoption elsewhere. To step back from this specific implementation, could you write a little about the new patterns here you'd like to propose for wider adoption and their relationship to existing concepts (e.g. `DataObject` vs. `Eloquentable`)?

PR #7166, opened 2021-07-01
pkp/pkp-lib#7125 Converting submission file to new Repository pattern

henriqueramos:


classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,466 @@
...
+            ])->count();
+        return (bool) $result > 0;
+    }
+
+    /**
+     * @copydoc PKPPubIdPluginDAO::changePubId()
+     */
+    public function changePubId($pubObjectId, $pubIdType, $pubId)
+    {
+        DB::table($this->settingsTable)
```
asmecher, line 365: I think this query is broken; it appears to only change `locale` and possibly `type`, but I think it was intended to change e.g. the setting value as well. See `PKPSubmissionFileDAO::changePubId` and the `$keyColumn` parameter to [ADODB's "replace" function](https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:replace).

classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,466 @@
...
+        if (
+            in_array(
+                $submissionFile->getData('assocType'),
+                [
+                    Application::ASSOC_TYPE_REVIEW_ROUND,
+                    Application::ASSOC_TYPE_REVIEW_ASSIGNMENT
+                ]
+            )
+        ) {
+            return;
```
asmecher, line 430: Did the original logic for this accidentally get removed? See `PKPSubmissionFileDAO::insertObject`.

classes/submissionFile/SubmissionFile.inc.php
```diff
@@ -13,16 +13,16 @@
...
 use APP\i18n\AppLocale;
-
-use PKP\db\DAORegistry;
+use PKP\core\PKPApplication;
 
 // Define the file stage identifiers.
 
 class SubmissionFile extends \PKP\core\DataObject
 {
+    public const SUBMISSION_FILE_PUBLIC = 1;
```
asmecher, line 25: I think this is deprecated and shouldn't be used any more.

classes/submissionFile/SubmissionFile.inc.php
```diff
@@ -374,10 +362,40 @@ public function setChapterId($chapterId)
     {
         $this->setData('chapterId', $chapterId);
     }
+
+    public function itsOnReviewFileStage(): bool
```
asmecher, line 366: This is stylistic, but I'd rather just do the comparison in the calling code rather than code up one-liners for each type of file. If the file type set changes, it's easy to forget to adjust these too, and it's harder to identify affected code when searching for e.g. a constant if you have to remember to also search for this function by name.

controllers/wizard/fileUpload/form/PKPSubmissionFilesUploadBaseForm.inc.php
```diff
@@ -194,25 +194,32 @@ public function getSubmissionFiles()
...
                 } elseif ($reviewRound) {
                     // Retrieve the submission files for the given review round.
-                    $submissionFilesIterator = Services::get('submissionFile')->getMany([
-                        'reviewRoundIds' => [(int) $reviewRound->getId()],
-                        'submissionIds' => [(int) $this->getData('submissionId')],
-                    ]);
+                    $collector = Repo::submissionFiles()
+                        ->getCollector()
+                        ->filterByReviewRoundIds([(int) $reviewRound->getId()])
+                        ->filterBySubmissionIds([(int) $this->getData('submissionId')]);
```
asmecher, line 200: I know this was a preexisting problem, but whenever we're filtering by a user-specified submission ID, it's always worth making sure it belongs to the current context (journal) as well.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -221,22 +235,37 @@ public function handleElement($node)
         // Add and edit the submission file revisions one-by-one so that a useful activity
         // log is built and past revisions can be accessed
         if (count($allRevisionIds) < 2) {
-            $submissionFile = Services::get('submissionFile')->add($submissionFile, $request);
+            $id = Repo::submissionFiles()->add($submissionFile);
```
asmecher, line 238: For consistency, use `$submissionFileId` rather than `$id`

asmecher (changes requested):
Thanks, @henriqueramos, looking good overall! Here's my initial feedback.

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,247 @@
...
+use PKP\core\interfaces\CollectorInterface;
+use PKP\plugins\HookRegistry;
+
+class Collector implements CollectorInterface
+{
+    /** @var DAO */
+    public $dao;
+
+    /** @var array get submission files for one or more file stages */
+    protected $fileStages = [];
```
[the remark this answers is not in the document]
henriqueramos: An empty array won't be part of the query itself as implemented at line 193.

With strict type checking or `isset` plus null values, the result will be the same.

With nullable params:
```bash
php > $value = null;
php > var_dump($value);
NULL
php > var_dump(isset($value));
bool(false)
```

With empty array:
```bash
php > $value = [];
php > var_dump($value);
array(0) {
}
php > var_dump($value !== []);
bool(false)
```

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,247 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @copydoc CollectorInterface::getQueryBuilder()
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $qb = DB::table($this->dao->table . ' as sf');
```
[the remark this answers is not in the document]
henriqueramos: The [table alias](https://www.mysqltutorial.org/mysql-alias/) here it's to help the SQL statements constructions, also we are returning a Builder statement here, not a query itself.

The `select` it's used after this call.

classes/submissionFile/SubmissionFile.inc.php
henriqueramos: I've found it being used on classes/install/Upgrade.inc.php:541, so, named it as a constant, just for tracking.

henriqueramos: https://github.com/pkp/ojs/pull/3160/files#diff-590dda691fda77f1799f0b82b40112f4b5890c543ba9fb61de436af7a9557220R541

asmecher: It used to be defined as a constant, but it was removed when it became historical only. Since it's only used in the legacy upgrade code, and eventually will be removed from there, it's best to keep it there only.

classes/submissionFile/Collector.inc.php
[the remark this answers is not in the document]
asmecher: See: https://teampkp.slack.com/archives/C5KFSJF6Y/p1626716034046200?thread_ts=1626716034.046200&cid=C5KFSJF6Y

classes/submissionFile/SubmissionFile.inc.php
henriqueramos: There's three ideas behind this implementation:

1) The DRY principle, to avoid the knowledge duplication (more [here](https://thevaluable.dev/dry-principle-cost-benefit-example/)) across the codebase;
2) The PRS-12 "rules" about lines ([2.3 Lines section](https://www.php-fig.org/psr/psr-12/)), where lines shouldn't be longer than 80 characters (as I have found at some places with "inline validation/comparison"); and
3) Avoid an Anemic Domain Model (more (here)[https://martinfowler.com/bliki/AnemicDomainModel.html]).

classes/submissionFile/Collector.inc.php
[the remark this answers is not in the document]
asmecher: I don't see a `select`; isn't this just falling back on the default (`SELECT *` when not otherwise specified)?

classes/submissionFile/DAO.inc.php
henriqueramos: Changed it to uses the Builder updateOrInsert method

classes/submissionFile/DAO.inc.php
henriqueramos: Small typo, forgot to add the `!` into the `in_array` in the previous commit.

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,254 @@
...
+        return $this;
+    }
+
+    /**
+     * @copydoc CollectorInterface::getQueryBuilder()
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $qb = DB::table($this->dao->table . ' as sf')
+            ->select('u.*');
```
asmecher, line 191: Should be `sf.*`, not `u.*`.

classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Get a collection of announcements matching the configured query
+     */
+    public function getMany(Collector $query): LazyCollection
+    {
+        $rows = $query
+            ->getQueryBuilder()
+            ->leftJoin('submissions as s', 's.submission_id', '=', 'sf.submission_id')
+            ->leftJoin('files as f', 'f.file_id', '=', 'sf.file_id')
+            ->select(['sf.*', 'f.*', 's.locale as locale'])
```
asmecher, line 119: The `Collector` already has a `select()` clause specifying `sf.*` (or it should be once that is corrected -- see other comment), so I would suggest using `addSelect(['f.*', 's.locale as locale'])` here. But since these additional joins are required for `fromRow` to work, why not move them into the `getQueryBuilder` function entirely?

classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,465 @@
...
+    /**
+     * @copydoc EntityDAO::get()
+     */
+    public function get(int $id): ?SubmissionFile
+    {
+        $row = DB::table($this->table . ' as sf')
+            ->join('submissions as s', 's.submission_id', '=', 'sf.submission_id')
+            ->join('files as f', 'f.file_id', '=', 'sf.file_id')
+            ->where($this->primaryKeyColumn, '=', $id)
+            ->addSelect(['sf.*', 'f.*', 's.locale as locale'])
```
NateWr, line 84: I want to revisit @asmecher [comment](https://github.com/pkp/pkp-lib/pull/7166/files/9532058ec064abe544b625c50821da55e36f41ae#diff-1ab22d4d9db06416915a2c05a7209dc7107d2d7249313a383f2d7fdcec7ffc42):

> The `Collector` already has a `select()` clause specifying `sf.*` (or it should be once that is corrected -- see other comment), so I would suggest using `addSelect(['f.*', 's.locale as locale'])` here. But since these additional joins are required for `fromRow` to work, why not move them into the `getQueryBuilder` function entirely? 

The way I did it was to include the additional required joins in the `getQueryBuilder` function. In hindsight, however, it is probably more performant not to do this. That way a call to `getIds` or `getCount` will not include unnecessary join/select statements.

To do this in a nice way, we'd have to extend the `Illuminate\Database\Query` class for every entity, as well as the `DB` facade, which seems like more abstraction than is beneficial in this case. The other workaround is a little bit ugly but works:

```php
public function selectObject(): Builder
{
    return $this
        ->join('submissions as s', 's.submission_id', '=', 'sf.submission_id')
        ->join('files as f', 'f.file_id', '=', 'sf.file_id')
        ->select(['sf.*', 'f.*', 's.locale as locale']);
}

public function get(int $id): ?SubmissionFile
{
    $q = DB::table($this->table . ' as sf')
        ->where($this->primaryKeyColumn, '=', $id);
    $row = $this->selectObject($q)
        ->first();
    return $row ? $this->fromRow($row) : null;
}
```

This would provide us a way to prevent additional DB calls in the `fromRow` method, which are pretty common and not very performant.

classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,465 @@
...
+
+        if (empty($submissionFileId)) {
+            return null;
+        }
+
+        $submissionFile = $this->get($submissionFileId);
+
+        if (!$submissionFile->itsOnFileProofStage()) {
+            return $submissionFile;
+        }
```
NateWr, line 262: Is the logic reversed here? I think it should return the file if it's in the proof file stage.

classes/submissionFile/DAO.inc.php
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    /**
+     * Remove a specific file assignment from a review round.
+     */
+    public function deleteReviewRoundAssignment(int $submissionFileId): void
+    {
+        // Remove currently assigned review files.
+        DB::table('review_round_files')->where([
+            'submission_id' => (int) $submissionFileId
```
NateWr, line 330: I think the table colulmn here should be `submission_file_id`.

classes/submissionFile/Repository.inc.php
```diff
@@ -0,0 +1,701 @@
...
+    protected $request;
+
+    /** @var PKPSchemaService $schemaService */
+    protected $schemaService;
+
+    public function __construct(Request $request)
+    {
+        $this->schemaService = new PKPSchemaService();
+        $this->dao = new DAO($this->schemaService);
+        $this->request = $request;
```
NateWr, line 61: The schema service and the dao can be automatically injected:

```
public function __construct (DAO $dao, Request $request, PKPSchemaService $schemaService)
    {
        $this->dao = $dao;
        $this->request = $request;
        $this->schemaService = $schemaService;
```

NateWr (commented):
Great work @henriqueramos! I've left a few comments, mostly about style / syntax. I'll move on to the other PRs now.

classes/core/PKPApplication.inc.php
```diff
@@ -284,6 +284,7 @@ public function getHttpClient()
             'headers' => [
                 'User-Agent' => $userAgent,
             ],
+            'verify' => false,
```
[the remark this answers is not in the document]
henriqueramos: Was a temporary fix to allow Travis run the tests without SSL cert expired bug being throwing.

api/v1/submissions/PKPSubmissionFileHandler.inc.php
```diff
@@ -196,23 +197,25 @@ public function getMany($slimRequest, $response, $args)
...
-        \HookRegistry::call('API::submissions::files::params', [&$params, $slimRequest]);
-
-        $params['submissionIds'] = [$this->getAuthorizedContextObject(ASSOC_TYPE_SUBMISSION)->getId()];
-
+        $collector = Repo::submissionFiles()
+            ->getCollector()
+            ->filterBySubmissionIds(
+                [$submission->getId()]
+            )
+            ->filterByReviewRoundIds($allowedReviewRoundIds)
```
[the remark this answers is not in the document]
henriqueramos: Or at [Collector, line 201 ](https://github.com/pkp/pkp-lib/pull/7166/files#diff-1fbcec3f507f0b4e1ea4afe98da72aec2b9a57511c64cee696767b31f5e98f48R210), we could implement another filtering for `$this->reviewRoundIds !== []`.

There's any scenario where is valid to do a `IN` with empty values? I don't think so.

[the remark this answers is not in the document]
NateWr: > There's any scenario where is valid to do a IN with empty values? I don't think so.

No, there's not, and I believe Laravel will throw an exception if it tries to execute a `whereIn` with an empty array. But I'd prefer to keep the `Collector::filterBy...()` syntaxes consistent across the board. `null` = do not apply this filter, `array` = apply this filter. Let the collector throw an exception and the code that's calling it can be fixed at source, instead of trying to build an exception into the `Collector`.

classes/core/PKPContainer.inc.php
```diff
@@ -75,6 +79,13 @@ public function renderForConsole($output, Throwable $e)
             };
         });
 
+        // This singleton is necessary to keep user selected language across the application
+        $this->singleton(IsoCodesFactory::class, function () {
+            $driver = new GettextExtensionDriver();
+            $driver->setLocale(PKPLocale::getLocale());
+            return new IsoCodesFactory(null, $driver);
+        });
```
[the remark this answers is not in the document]
henriqueramos: Those lines deserve a few comments from my side.

sokil/php-isocodes uses the `LANGUAGE` env variable to handle the i18n inside GettextExtensionDriver.

On my Apache installation, this env `LANGUAGE` was marked as `pt-br`. And even we changed the PKPLocale to `en`, data coming from php-isocodes stays as pt-br version.

With this change, I've forced the php-isocodes to use the data from PKPLocale.

classes/services/queryBuilders/PKPSubmissionFileQueryBuilder.inc.php
```diff
@@ -17,7 +17,7 @@
 use Illuminate\Support\Facades\DB;
 
 use PKP\services\queryBuilders\interfaces\EntityQueryBuilderInterface;
-use PKP\submission\SubmissionFile;
+use PKP\submissionFile\SubmissionFile;
```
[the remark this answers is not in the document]
henriqueramos: For some reason, during the rebases, this file came back from graveyard.

classes/services/PKPSubmissionFileService.inc.php
```diff
@@ -154,13 +153,16 @@ public function getProperties($submissionFile, $props, $args = null)
...
                     break;
                 case 'dependentFiles':
-                    $dependentFilesIterator = Services::get('submissionFile')->getMany([
-                        'assocTypes' => [ASSOC_TYPE_SUBMISSION_FILE],
-                        'assocIds' => [$submissionFile->getId()],
-                        'submissionIds' => [$submission->getId()],
-                        'fileStages' => [SubmissionFile::SUBMISSION_FILE_DEPENDENT],
-                        'includeDependentFiles' => true,
-                    ]);
+                    $collector = Repo::submissionFiles()
```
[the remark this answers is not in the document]
henriqueramos: > This file shouldn't exist anymore. Once we've migrated to the repo, the service class can be removed.
> 
> It looks like maybe the `getProperties` method is the only thing left here? If so, that should go into the schema map's `mapByProperties` method.

For some reason, during the rebases, this file came back from graveyard.

api/v1/submissions/PKPSubmissionFileHandler.inc.php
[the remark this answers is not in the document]
henriqueramos: Okay!

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,253 @@
...
+                    ->whereIn('review_round_id', $this->reviewRoundIds);
+            });
+        }
+
+        if ($this->reviewIds !== null) {
+            $qb->join('review_files as rf', 'rf.submission_file_id', '=', 'sf.submission_file_id')
+                ->whereIn('rf.review_id', $this->reviewIds);
+        }
+
+        if ($this->assocTypes !== null) {
```
asmecher, line 223: This still needs fixing: https://github.com/pkp/pkp-lib/pull/7166#discussion_r676841825

classes/submissionFile/SubmissionFile.inc.php
asmecher: This still needs fixing.

classes/submissionFile/SubmissionFile.inc.php
asmecher: I think I'd still rather remove these, at least for now, and consider their addition more broadly outside of this PR. It's easy to partially introduce new conventions and never make them universal, which causes a lot of confusion.

controllers/wizard/fileUpload/form/PKPSubmissionFilesUploadBaseForm.inc.php
asmecher: This still needs fixing.

classes/core/PKPApplication.inc.php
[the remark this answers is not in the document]
asmecher: Most servers seem to have rolled out certificate updates behind the scenes to fix the Let's Encrypt issue that started this. I suspect you'll find this problem has gone away (and if it's caused by your local machine make sure to update your packages).

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,253 @@
...
+use Illuminate\Support\Facades\DB;
+use PKP\core\interfaces\CollectorInterface;
+use PKP\plugins\HookRegistry;
+
+class Collector implements CollectorInterface
+{
+    /** @var DAO */
+    public $dao;
+
+    /** @var array get submission files for one or more file stages */
```
asmecher, line 26: The self-docs for these need to permit nulls (e.g. `?array` here).

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,253 @@
...
+        if ($this->fileStages !== null) {
+            $qb->whereIn('sf.file_stage', $this->fileStages);
+        }
+
+        if ($this->genreIds !== null) {
+            $qb->whereIn('sf.genre_id', $this->genreIds);
+        }
+
+        if ($this->fileIds !== null) {
+            $qb->join('submission_file_revisions as sfr', 'sfr.submission_file_id', '=', 'sf.submission_file_id')
```
asmecher, line 206: This could potentially match several results, I think? If that's the case, then you'd need to do a subquery here in order to avoid multiplying the results.

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,253 @@
...
+        $this->count = $count;
+
+        return $this;
+    }
+
+    /**
+     * Offset the number of objects retrieved, for example to
+     * retrieve the second page of contents
+     */
+    public function offset(int $offset): self
```
asmecher, line 178: Should be `?offset`

classes/core/PKPApplication.inc.php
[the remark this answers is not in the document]
henriqueramos: Happened only to Travis servers. I will remove this config flag and see if works.

classes/submission/form/PKPSubmissionSubmitStep2Form.inc.php
```diff
@@ -65,15 +64,17 @@ public function fetch($request, $template = null, $display = false)
...
+            $submissionFilesIterator = Repo::submissionFiles()
+                ->getMany($submissionFilesCollector);
             foreach ($submissionFilesIterator as $submissionFile) {
-                $submissionFiles[] = Services::get('submissionFile')->getSummaryProperties($submissionFile, [
-                    'request' => $request,
-                    'submission' => $this->submission,
-                ]);
+                $submissionFiles[] = Repo::submissionFiles()
+                    ->getSchemaMap()
+                    ->map($submissionFile);
```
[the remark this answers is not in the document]
henriqueramos: Or event `summarizeMany` and remove the `foreach`.

classes/submissionFile/Collector.inc.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        $this->includeDependentFiles = $includeDependentFiles;
+
+        return $this;
+    }
+
+    /**
+     * Limit the number of objects retrieved
+     */
+    public function limit(int $count): self
```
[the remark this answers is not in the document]
henriqueramos: No, this should return the Collector instance itself, to be chained with other methods.

classes/submissionFile/Collector.inc.php
henriqueramos: No, this should return the Collector instance itself, to be chained with other methods if necessary.

classes/submissionFile/Collector.inc.php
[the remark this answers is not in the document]
asmecher: I mean `public function limit(?int $count): self` -- meaning that `$count can be supplied as a `null` to reset a limit that was previously specified.

classes/submissionFile/Collector.inc.php
asmecher: I mean `?int $offset` (as above: https://github.com/pkp/pkp-lib/pull/7166#discussion_r721708330)

henriqueramos: On lines 241 and 245 I'm counting if those values are greater than 0 to apply the query restraints.

classes/submissionFile/Collector.inc.php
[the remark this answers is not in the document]
henriqueramos: On lines 241 and 245 I'm counting if those values are greater than 0 to apply the query restraints.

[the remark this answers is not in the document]
asmecher: I'd like to keep this consistent with the other filter properties, which use `null` for "no limit" (proposed/documented [here](https://github.com/pkp/pkp-docs/issues/812#issuecomment-900639385)). The other collectors use `null` for limit and offset; see e.g.: https://github.com/pkp/ojs/blob/main/pages/issue/IssueHandler.inc.php#L157

`0` and `null` are both a little arbitrary in this case, but I think `limit(null)` is more intuitive than `limit(0)` (which suggests 0 results to me).

classes/submissionFile/DAO.inc.php
henriqueramos: Currently this is the way we did it. Even on [main branch](https://github.com/pkp/pkp-lib/blob/bf272b821ce4d5d13cc2645426a02722663bcc59/classes/submission/PKPSubmissionFileDAO.inc.php#L228).

classes/services/PKPSubmissionFileService.inc.php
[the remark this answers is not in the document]
henriqueramos: Has been fixed. ;)

classes/submissionFile/DAO.inc.php
henriqueramos: Of course!

NateWr: I think only `SubmissionFile::SUBMISSION_FILE_PROOF` for this one. (dependent files can't have pub ids)

classes/core/PKPApplication.inc.php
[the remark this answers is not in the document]
henriqueramos: OPS tests still failing due to SSL certs expired :( They don't update the certs on the servers yet.

classes/submissionFile/SubmissionFile.inc.php
henriqueramos: I will revert the changes where I used them and create an issue to enable this in a near future.

classes/core/PKPContainer.inc.php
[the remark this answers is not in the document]
henriqueramos: Pull Request has been created. https://github.com/pkp/pkp-lib/pull/7382

api/v1/submissions/PKPSubmissionFileHandler.inc.php
```diff
@@ -196,23 +198,22 @@ public function getMany($slimRequest, $response, $args)
...
-        foreach ($filesIterator as $file) {
-            $items[] = Services::get('submissionFile')->getSummaryProperties($file, $propertyArgs);
-        }
+        $items = Repo::submissionFiles()
+            ->getSchemaMap()
+            ->summarizeMany($files);
 
         $data = [
-            'itemsMax' => Services::get('submissionFile')->getCount($params),
+            'itemsMax' => Repo::submissionFiles()->getCount($collector),
```
asmecher, line 216: I think `getCount` will re-run the query; I believe you should be able to just use `files->count()`, since it doesn't have a limit applied.

classes/components/forms/context/PKPPaymentSettingsForm.inc.php
```diff
@@ -41,7 +43,7 @@ public function __construct($action, $locales, $context)
         $this->locales = $locales;
 
         $currencies = [];
-        $isoCodes = new \Sokil\IsoCodes\IsoCodesFactory();
+        $isoCodes = app(IsoCodesFactory::class);
```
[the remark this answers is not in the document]
henriqueramos: Moved to #7282.

classes/components/forms/context/PKPMastheadForm.inc.php
```diff
@@ -43,7 +43,7 @@ public function __construct($action, $locales, $context, $imageUploadUrl)
         $this->action = $action;
         $this->locales = $locales;
 
-        $isoCodes = new IsoCodesFactory();
+        $isoCodes = app(IsoCodesFactory::class);
```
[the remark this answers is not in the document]
henriqueramos: Moved to #7282.

asmecher (approved):

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -15,6 +15,7 @@
 
 use APP\core\Application;
 use APP\core\Services;
+use APP\facades\Repo;
```
asmecher, line 18: This is a duplicate; it breaks the upgrade tests: https://app.travis-ci.com/github/pkp/ojs/builds/241578343

henriqueramos: Removed the `PKP\facades\Repo`;

PR #7170, opened 2021-07-02
pkp/pkp-lib#7127 WIP Convert UserDAO/UserService to new Respository/DAO/Map pattern

asmecher:


classes/user/DAO.inc.php
```diff
@@ -0,0 +1,344 @@
...
+     * Retrieve a user by API key.
+     *
+     * @param $allowDisabled boolean
+     *
+     * @return User?
+     */
+    public function getByApiKey(string $apiKey, bool $allowDisabled = true): ?User
+    {
+        return $this->getBySetting('apiKey', $apiKey, $allowDisabled);
+    }
```
NateWr, line 165: Seems to me like if we're going to add an `$allowedDisable` argument to the method it should be `false` by default.

classes/user/DAO.inc.php
```diff
@@ -0,0 +1,344 @@
...
+            ->join('user_settings AS us', 'u.user_id', '=', 'us.user_id')
+            ->where('us.setting_name', '=', $settingName)
+            ->where('us.setting_value', '=', $settingValue)
+            ->when(!$allowDisabled, function ($query) {
+                return $query->where('u.disabled', '=', false);
+            })
+            ->get('u.user_id AS user_id')
+            ->first();
+        return $row ? $this->get($row->user_id) : null;
+    }
```
NateWr, line 207: This seems like it duplicates the `filterBySetting` method of the collector. Can we reuse that and just take the `first()` result?

classes/user/DAO.inc.php
```diff
@@ -0,0 +1,344 @@
...
+    /**
+     * Retrieve a user by setting.
+     *
+     * @param $settingName string
+     * @param $settingValue string
+     * @param $allowDisabled boolean
+     *
+     * @return User?
+     */
+    public function getBySetting($settingName, $settingValue, $allowDisabled = true)
```
NateWr, line 195: Missing return type hint

NateWr: A few more methods below missing it too.

classes/user/PKPUserAction.inc.php
```diff
@@ -117,8 +118,7 @@ public function mergeUsers($oldUserId, $newUserId)
             }
         }
 
-        $userDao = DAORegistry::getDAO('UserDAO'); /** @var UserDAO $userDao */
-        $userDao->deleteUserById($oldUserId);
+        Repo::user()->delete(Repo::user()->get($oldUserId));
 
         return true;
     }
```
NateWr, line 124: I'd like to see thee `PKPUserAction` methods merged into the Repository and then removed.

classes/user/Repository.inc.php
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * @file classes/user/Repository.inc.php
+ *
+ * Copyright (c) 2014-2020 Simon Fraser University
+ * Copyright (c) 2000-2020 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class announcement
+ *
+ * @brief A repository to find and manage announcements.
```
NateWr, line 11: Docs refere to anouncements in two places. Also, update copyright date.

classes/user/Repository.inc.php
```diff
@@ -0,0 +1,333 @@
...
+        );
+        $report = new Report($dataSource);
+
+        HookRegistry::call('User::getReport', $report);
+
+        return $report;
+    }
+
+    public function getRolesOverview($args = [])
+    {
```
NateWr, line 297: This method should be passed a `Collector` with the required arguments rather than receiving them as an assoc array.

classes/user/UserSettingsDAO.inc.php
```diff
@@ -18,9 +18,6 @@
 namespace PKP\user;
 
 use APP\core\Application;
-use PKP\db\DAORegistry;
-
-use PKP\db\DAOResultFactory;
 
 class UserSettingsDAO extends \PKP\db\DAO
```
NateWr, line 22: Maybe not in this PR, but it would be nice to do away with this class, since it breaks convention with other DAOs. I think some of the methods are covered in the regular DAO and some could be moved there.

classes/user/maps/Schema.inc.php
```diff
@@ -0,0 +1,262 @@
...
+                    break;
+                case 'signature':
+                    $output[$prop] = $user->getSignature(null);
+                    break;
+                case 'authId':
+                    $output[$prop] = $user->getAuthId();
+                    break;
+                case 'authString':
+                    $output[$prop] = $user->getAuthStr();
+                    break;
```
NateWr, line 145: I know you didn't change anything here, but is it safe to have the `authId` and `authString` exposed in the API? Eventually non-managers will be able to access it. We might need some protections in the map (you can see how this is handled in the submission/publication map when data is passed to a reviewer).

schemas/user.json
```diff
@@ -43,6 +47,21 @@
...
+		},
+		"dateLastLogin": {
+			"type": "string",
+			"validation": [
+				"date:Y-m-d H:i:s"
+			]
+		},
+		"inlineHelp": {
+			"type": "boolean"
+		},
```
NateWr, line 64: It's up to you how you handle this, but one quirk of the API/validation stuff when editing an object is that you need to permit something to be `nullable` in order to support resetting it. For example, if someone had a setting of `inlineHelp: true` and you wanted to remove this (without setting it to false), you have to pass a `null` value to the API. That will throw a validation error unless you add this validation rule:

```
"validation": [
    "nullable"
]
```

schemas/user.json
```diff
@@ -20,7 +22,9 @@
 			"type": "string"
 		},
 		"biography": {
-			"$ref": "#/definitions/LocaleObject"
+			"type": "string",
+			"multilingual": true,
+                        "apiSummary": true
```
NateWr, line 27: Indendtation is off.

NateWr: Same for other lines with "apiSummary" below.

classes/user/maps/Schema.inc.php
```diff
@@ -0,0 +1,262 @@
...
+        $output = [];
+
+        foreach ($props as $prop) {
+            switch ($prop) {
+                case 'id':
+                    $output[$prop] = (int) $user->getId();
+                    break;
+                case 'userName':
+                    $output[$prop] = $user->getUserName();
+                    break;
```
NateWr, line 109: I've been avoiding mapping every property manually by having a default condition call getData:

```php
   default:
        $output[$prop] = $user->getData($prop);
        break;
```

That way we only have to write code when mapping something that isn't straightforward.

NateWr (commented):
I haven't done a full top-to-bottom review, but I had a look at the Repo/DAO/Collector/Schema classes and the JSON schema. It looks good! Just a few comments for discussion mostly.

classes/user/maps/Schema.inc.php
```diff
@@ -0,0 +1,262 @@
+<?php
+/**
+ * @file classes/user/maps/Schema.inc.php
+ *
+ * Copyright (c) 2014-2020 Simon Fraser University
+ * Copyright (c) 2000-2020 John Willinsky
```
NateWr, line 6: Copyright date

classes/user/maps/Schema.inc.php
```diff
@@ -0,0 +1,262 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class publication
+ *
+ * @brief Map publications to the properties defined in the publication schema
```
NateWr, line 11: Update to refer to users

controllers/grid/settings/category/form/CategoryForm.inc.php
```diff
@@ -211,21 +211,23 @@ public function fetch($request, $template = null, $display = false)
...
-                'contextId' => $context->getId(),
-                'roleIds' => Role::ROLE_ID_SUB_EDITOR,
-                'assignedToCategory' => (int) $this->getCategoryId(),
-            ]);
+            $assignedToCategory = iterator_to_array(Repo::user()->getIds(
+                Repo::user()->getCollector()
+                    ->filterByContextIds([$context->getId()])
+                    ->filterByRoleIds([Role::ROLE_ID_SUB_EDITOR])
+                    ->filterByAssignedCategoryIds([$this->getCategoryId()])
+            ));
```
NateWr, line 230: This probably works but since we're no longer returning the DAOResultIterator it may be better to use the Collection's `->toArray()` method.

controllers/grid/settings/user/UserGridRow.inc.php
```diff
@@ -70,8 +71,7 @@ public function initialize($request, $template = null)
                 $actionArgs['newUserId'] = $rowId;
 
                 // Verify that the old user exists
-                $userDao = DAORegistry::getDAO('UserDAO'); /** @var UserDAO $userDao */
-                $oldUser = $userDao->getById($this->getOldUserId());
+                $oldUser = Repo::user()->get($this->getOldUserId());
```
NateWr, line 74: You've probably tested this but I'm always a little bit wary of assuming an ID is an `int` when working with the grids.

classes/user/form/ContactForm.inc.php
```diff
@@ -35,7 +35,10 @@ public function __construct($user)
         // Validation checks for this form
         $this->addCheck(new \PKP\form\validation\FormValidatorEmail($this, 'email', 'required', 'user.profile.form.emailRequired'));
         $this->addCheck(new \PKP\form\validation\FormValidator($this, 'country', 'required', 'user.profile.form.countryRequired'));
-        $this->addCheck(new \PKP\form\validation\FormValidatorCustom($this, 'email', 'required', 'user.register.form.emailExists', [DAORegistry::getDAO('UserDAO'), 'userExistsByEmail'], [$user->getId(), true], true));
+        $this->addCheck(new \PKP\form\validation\FormValidatorCustom($this, 'email', 'required', 'user.register.form.emailExists', function ($email, $userId) {
+            $user = Repo::user()->dao->getByEmail($email);
+            return !$user || $user->getId() != $userId;
+        }, [$user->getId()], true));
```
NateWr, line 41: Although we don't yet have a API for adding/editing users, this is the sort of thing that should probably be captured in the Repository's `validate` method. Or, if that's not in use yet, we could remove the `validate` method from the Repository so that it's not presumed to be complete.

NateWr: (In general, the old forms don't use the validate methods -- they're only used in the REST API.)

cypress/support/commands.js
```diff
@@ -391,7 +391,7 @@ Cypress.Commands.add('createUser', user => {
 		cy.get('form[id=userRoleForm]').contains(role).click();
 	});
 	cy.get('form[id=userRoleForm] button[id^=submitFormButton]').click();
-	cy.get('span[id$="-username"]:contains("' + Cypress.$.escapeSelector(user.username) + '")');
+        cy.waitJQuery();
```
NateWr, line 394: It looks like this actually changes the check, so that it's no longer waiting to see if the user was actually created and appears in the list. Was that intentional?

tests/classes/notification/PKPNotificationManagerTest.php
```diff
@@ -354,13 +354,15 @@ private function getFixtureCreateNotificationSendEmail($expectedNotification)
...
-        $userDaoStub->expects($this->any())
-            ->method('getById')
-            ->will($this->returnValue($testUser));
-        DAORegistry::registerDAO('UserDAO', $userDaoStub);
+        /*        $userDaoStub = $this->getMockBuilder(\PKP\user\DAO::class)
+                    ->setMethods(['getById'])
+                    ->getMock();
+                $userDaoStub->expects($this->any())
+                    ->method('getById')
+                    ->will($this->returnValue($testUser));*/
```
NateWr, line 362: Did you intend to leave this commented block in?

NateWr (commented):
:+1: I've had a look through the rest of it. Just a few minor comments here and there. A ton of work on this!

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+     */
+    public function filterExcludeSubmissionStage(int $submissionId, int $stageId, int $userGroupId): self
+    {
+        $this->excludeSubmissionStage = [
+            'submission_id' => $submissionId,
+            'stage_id' => $stageId,
+            'user_group_id' => $userGroupId,
+        ];
+        return $this;
+    }
```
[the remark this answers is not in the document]
asmecher: :+1:

classes/user/UserSettingsDAO.inc.php
asmecher: Yes, I thought the same but needed to control scope on this one. I think we'll have to do some future work on things like relationship tables in particular.

tests/classes/notification/PKPNotificationManagerTest.php
asmecher: Fixed, thanks!

api/v1/_email/PKPEmailHandler.inc.php
```diff
@@ -137,11 +137,13 @@ public function create(ServerRequestInterface $slimRequest, APIResponse $respons
...
 
-        // Only permit emails to be sent to active users in this context
-        $params['status'] = 'active';
-        $params['contextId'] = $contextId;
 
-        $userIds = Services::get('user')->getIds($params);
+        $userIds = iterator_to_array(Repo::user()->getIds(
+            Repo::user()->getCollector()
+                ->filterByContextIds([$contextId])
+                ->filterByDisabled(false) // Only permit emails to be sent to active users in this context
```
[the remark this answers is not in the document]
asmecher: See https://github.com/asmecher/pkp-lib/commit/f134a8bd2b0c830bc0a01b8a69f02822d1e77aff -- I changed this to `filterByStatus` with the parameter being `Collector::STATUS_ENABLED`, `Collector::STATUS_DISABLED`, or `Collector::STATUS_ALL`:
- `STATUS_ALL` is `null` for consistency with our "null means don't care" habit. Not that we should be comparing with non-constants when they are defined for the purpose.
- The default is `STATUS_ENABLED`. There are lots of places where we didn't check the enabled status and thus returned disabled users in user lists; this tightens it up to *only* the management area. IMO we should add a filter for this over there, but I'll leave that for now.
- For now the constants are strings and null. However, once our PHP baseline supports it, I'd like to use PHP enums. This will help with autocompletion, prevent coding errors, and make error checking a built-in. Enums are due for PHP8.1: https://wiki.php.net/rfc/enumerations

I didn't go with your suggested approach of two functions because they would interact strangely with an "only enabled users" default. Presumably the default would go on the `includeDisabled` property -- meaning that calling code would need to call both `includeDisabled` and `isDisabled` if they wanted to use `isDisabled`, when you might expect just `isDisabled` to be enough.

Does that look OK? I think this will get a lot more intuitive and logical once enums are in, but this sets the stage for it.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+            $this->submissionAssignment = null;
+            if ($stageId !== null || $userGroupId !== null) {
+                throw new \InvalidArgumentException('If a stage or user group ID is specified, a submission ID must be specified as well.');
+            }
+        } else {
+            $this->submissionAssignment = [
+                'submission_id' => $submissionId,
+                'stage_id' => $stageId,
+                'user_group_id' => $userGroupId,
+            ];
```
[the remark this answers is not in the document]
asmecher: I kept these three together because they are intertwined and refer to a submission assignment. We already have a `filterByUserGroupId` (and `userGroupIds` property) separate from the one here, used to check user group enrollment (as opposed to assignment to a submission with a particular user group). Likewise, we have another join on `stage_assignments` for excluded submission stages. I suppose I could use `assignedSubmissionId`, `assigedSubmissionStageId`, and `assignedSubmssionUserGroupId`, but I'm not sure that's clearer.

classes/user/maps/Schema.inc.php
asmecher: I think this old auth stuff should be removed in favour of a standard OpenID/OAuth implementation with whatever that entails. I recommend leaving this for now and cleaning it up either when we expose the user API more broadly, or when we integrate OpenID/OAuth support, whichever comes first.

schemas/user.json
asmecher: This is one of those areas I'm a little less confident about. Based on this, I would guess most non-required properties should be nullable, e.g. `affiliation`, `billingAddress`, `familyName`, maybe even `interests`... -- do I have that right?

schemas/user.json
asmecher: Fixed (tabs in place of spaces), thanks.

classes/user/maps/Schema.inc.php
asmecher: Done: https://github.com/asmecher/pkp-lib/commit/a9434fa2cc7a48e7db99e50ecf963b203388010a

api/v1/submissions/PKPSubmissionHandler.inc.php
```diff
@@ -503,21 +503,18 @@ public function getParticipants($slimRequest, $response, $args)
...
-            }
+        $usersIterator = Repo::user()->getMany(
+            Repo::user()->getCollector()
+                ->filterByContextIds([$context->getId()])
+                ->filterSubmissionAssignment($submission->getId(), $stageId)
+        );
+        $args = [
+            'request' => $request,
+            'slimRequest' => $slimRequest,
+        ];
```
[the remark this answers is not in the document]
asmecher: Removed, thanks!

api/v1/submissions/PKPSubmissionHandler.inc.php
```diff
@@ -674,11 +671,11 @@ public function versionPublication($slimRequest, $response, $args)
...
-        $userService = Services::get('user');
-        $usersIterator = $userService->getMany([
-            'contextId' => $submission->getContextId(),
-            'assignedToSubmission' => $submission->getId(),
-        ]);
+        $usersIterator = Repo::user()->getMany(
+            Repo::user()->getCollector()
+                ->filterByContextIds([$submission->getContextId()])
+                ->filterSubmissionAssignment($submission->getId())
+        );
```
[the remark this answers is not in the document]
asmecher: Done, thanks!

api/v1/users/PKPUserHandler.inc.php
```diff
@@ -99,20 +99,50 @@ public function getMany($slimRequest, $response, $args)
...
         $propertyArgs = [
             'request' => $request,
             'slimRequest' => $slimRequest,
         ];
-        foreach ($usersItereator as $user) {
-            $items[] = Services::get('user')->getSummaryProperties($user, $propertyArgs);
+        $map = Repo::user()->getSchemaMap();
+        $items = [];
+        foreach ($usersIterator as $user) {
+            $items[] = $map->summarize($user);
```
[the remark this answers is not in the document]
asmecher: Done!

classes/cliTool/MergeUsersTool.inc.php
```diff
@@ -115,11 +116,14 @@ public function execute()
      */
     protected function _getUserBySpecifier($specifier)
     {
-        $userDao = DAORegistry::getDAO('UserDAO'); /** @var UserDAO $userDao */
         if (substr($specifier, 0, 3) == 'id=') {
-            return $userDao->getById(substr($specifier, 3));
+            $userId = substr($specifier, 3);
+            if (!ctype_digit($userId)) {
+                break;
```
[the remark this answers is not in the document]
asmecher: That should be a `return null`; fixed.

controllers/grid/settings/category/form/CategoryForm.inc.php
asmecher: Done!

controllers/grid/settings/user/UserGridRow.inc.php
asmecher: We'll probably have a lot of cases like this to work out as we adopt strict typehinting. I've added an int cast for now.

classes/user/form/ContactForm.inc.php
asmecher: Whoops, I think I brought that `Repository::validate` in wholesale from the Announcement code and forgot to check through it. I've removed it for now.

cypress/support/commands.js
asmecher: I think the problem was that this check assumed that the user was going to appear on the current page, and that may not be the case. It showed up when I temporarily removed the ordering code -- the ordering has been reinstituted, but I still think we can't assume the new user will be on the current page (think e.g. of adding an 11th user on a 10-user list).

classes/user/maps/Schema.inc.php
```diff
@@ -0,0 +1,262 @@
...
+                        $route = $slimRequest->getAttribute('route');
+                        $arguments = $route->getArguments();
+                        $dispatcher = $this->request->getDispatcher();
+                        $output[$prop] = $dispatcher->url(
+                            $this->request,
+                            \PKPApplication::ROUTE_API,
+                            $arguments['contextPath'],
+                            'users/' . $user->getId()
+                        );
+                    }
```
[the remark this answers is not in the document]
asmecher: Done, thanks!

classes/components/forms/context/PKPNotifyUsersForm.inc.php
```diff
@@ -53,11 +53,13 @@ public function __construct($action, $context, $userGroups)
                 'value' => $userGroup->getId(),
                 'label' => $userGroup->getLocalizedData('name'),
             ];
-            $this->userGroupCounts[$userGroup->getId()] = Services::get('user')->getCount([
-                'contextId' => $userGroup->getData('contextId'),
-                'userGroupIds' => [$userGroup->getId()],
-                'status' => 'active',
-            ]);
+            $dao = Repo::user()->dao;
+            $this->userGroupCounts[$userGroup->getId()] = $dao->getCount(
```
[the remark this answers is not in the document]
asmecher: Done, thanks: https://github.com/asmecher/pkp-lib/commit/f981145a7e8a36700a4e063e7930ffa19aadb19c
(Also for OJS PR)

classes/core/EntityDAO.inc.php
```diff
@@ -84,12 +84,20 @@ public function exists(int $id): bool
...
+        return DB::table($this->table);
+    }
+
     /**
      * Get an object by its ID
      */
     public function get(int $id): ?DataObject
     {
-        $row = DB::table($this->table)
+        $row = $this->_getFetchQuery()
```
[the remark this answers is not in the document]
asmecher: I think this got mooted by some other code (e.g. removal of `columns`) -- reverted.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+    public const ORDER_DIR_DESC = 'DESC';
+
+    /** @var DAO */
+    public $dao;
+
+    public int $orderBy = self::ORDERBY_ID;
+    public string $orderDirection = 'ASC';
+
+    /** @var array|null */
+    public $userGroupIds = null;
```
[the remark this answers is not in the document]
asmecher: Done throughout, thanks!

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    /**
+     * Limit results to users registered after a specified date.
+     */
+    public function filterRegisteredAfter(?string $registeredAfter): self
+    {
+        $this->registeredAfter = $registeredAfter;
+        return $this;
+    }
```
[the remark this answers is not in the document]
asmecher: Meh, having a couple of functions with obvious usage trumps more defined constants in my books. Are you OK with how things are already here?

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+
+    /**
+     * Filter by assigned subeditor section IDs
+     */
+    public function filterByAssignedCategoryIds(?array $categoryIds): self
+    {
+        $this->assignedCategoryIds = $categoryIds;
+        return $this;
+    }
```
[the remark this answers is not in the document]
asmecher: :+1: I like this. Fixed.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+                            ->join('user_settings AS us', function ($join) {
+                                $join->on('u.user_id', '=', 'us.user_id')
+                                    ->whereIn('us.setting_name', [Identity::IDENTITY_SETTING_GIVENNAME, Identity::IDENTITY_SETTING_FAMILYNAME]);
+                            })
+                            ->where(DB::raw('LOWER(us.setting_value)'), 'LIKE', $likePattern)
+                            ->orWhere(DB::raw('LOWER(email)'), 'LIKE', $likePattern)
+                            ->orWhere(DB::raw('LOWER(username)'), 'LIKE', $likePattern);
+                    });
+                }
+            })
```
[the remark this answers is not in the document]
asmecher: :+1: fixed, good spotting.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+                        $dateTime = new \DateTime();
+                        $dateTime->sub(new \DateInterval('P' . $daysSinceMin . 'D'));
+                        $query->where('ra_latest.date_assigned', '>', $dateTime->format('Y-m-d'));
+                        if (is_array($this->daysSinceLastAssignment) && !empty($this->daysSinceLastAssignment[1])) {
+                            $daysSinceMax = (int) $this->daysSinceLastAssignment[1] + 1; // Add one to include upper bound
+                            $dateTime = new \DateTime();
+                            $dateTime->add(new \DateInterval('P' . $daysSinceMax . 'D'));
+                            $query->where('ra_latest.date_assigned', '<', $dateTime->format('Y-m-d'));
+                        }
+                    });
```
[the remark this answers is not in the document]
asmecher: I'll admit to finding the run-on sentences of fluent interfaces a little awkward when formatting them as I would procedural code -- but done.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+        // Limit and offset results for pagination
+        if (!is_null($this->count)) {
+            $q->limit($this->count);
+        }
+        if (!is_null($this->offset)) {
+            $q->offset($this->offset);
+        }
+
+        $siteDao = DAORegistry::getDAO('SiteDAO'); /** @var SiteDAO $siteDao */
+        $site = $siteDao->getSite();
```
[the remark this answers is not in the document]
asmecher: I'd rather we left the request object out of things unless we're sure the code is being executed from within a request. Eventually we'll need to clear up some spaghetti related to this for command-line tools.

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,589 @@
...
+        $siteLocale = $site->getPrimaryLocale();
```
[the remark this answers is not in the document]
asmecher: That's true of e.g. submissions, but users are site-level data and locale precedences (e.g. required fields) are defined by the site locale.

classes/user/DAO.inc.php
```diff
@@ -0,0 +1,344 @@
...
+     * @param $allowDisabled boolean If true, allow fetching a disabled user.
+     */
+    public function get($id, $allowDisabled = true): ?User
+    {
+        $user = parent::get($id);
+        if (!$allowDisabled && $user->getDisabled()) {
+            return null;
+        }
+        return $user;
+    }
```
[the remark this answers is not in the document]
asmecher: Hmm, this leads to a discrepancy: `getMany` now defaults to excluding disabled users unless a collector function is called, but `get` would force calling code to handle the disabled case explicitly. I was actually going to propose flipping the default here to `false` to match the collector default.

classes/user/DAO.inc.php
asmecher: (See: https://github.com/pkp/pkp-lib/pull/7170/files#r692547579)

classes/user/DAO.inc.php
asmecher: Agreed.

classes/user/PKPUserAction.inc.php
asmecher: This is split between PKPUserAction and UserAction; what's your opinion on extending PKP\user\Repository into APP\user\Repository and injecting that? Still the go-to for cases like this?

classes/user/Repository.inc.php
asmecher: :+1: Fixed!

classes/user/Collector.inc.php
```diff
@@ -0,0 +1,572 @@
...
+                        ->where('assoc_type', '=', ASSOC_TYPE_CATEGORY)
+                        ->whereIn('assoc_id', $this->assignedCategoryIds);
+                });
+            })
+
+            ->when($this->searchPhrase !== null, function ($query) {
+                $words = explode(' ', $this->searchPhrase);
+                foreach ($words as $word) {
+                    $query->whereIn('u.user_id', function ($query) use ($word) {
+                        $likePattern = '%' . addcslashes(PKPString::strtolower($word), '%_') . '%';
```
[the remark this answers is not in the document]
asmecher: Yes, I'll fix that!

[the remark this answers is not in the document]
asmecher: Commit: https://github.com/asmecher/pkp-lib/commit/5157f73de4c9181ec90ba463dcca4ac75b1494ce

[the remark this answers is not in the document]
bozana: :+1:  Then I'll do the same :-)

classes/user/PKPUserAction.inc.php
asmecher: Done, thanks -- OJS and OMP required an extended Repository class, OPS didn't.

classes/user/DAO.inc.php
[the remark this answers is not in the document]
asmecher: Done!

classes/user/Collector.inc.php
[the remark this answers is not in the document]
NateWr: Have you looked at the [Service Container](https://laravel.com/docs/8.x/container) for Laravel? I think this is how they deal with app-wide classes that need to be overridden for unit testing. It interacts well with their automatic dependency injection (eg - `__construct(Locales $locales)` can be made to auto-inject a different class during unit tests).

PR #7177, opened 2021-07-09
Fix `getIntMaxFileMBs`

Commifreak:
I think we might have a problem here:

The function name says, we want MBs as output unit, right?

I have `1G` set in my php.ini.

The function gets `1` in `$num` and `G` in `$scale`. So far so good.
If the scale is g, the function divides by 1024?? Why? If we want MB, we had to multiply. And case k vice versa.

Second change: `// no break` to `break;` - we use PHP 7.4 and with no break, all cases are actaully executed.

```php
<?php
define('UPLOAD_MAX_FILESIZE', ini_get('upload_max_filesize'));
echo UPLOAD_MAX_FILESIZE;
echo "<br />";

		$num = substr(UPLOAD_MAX_FILESIZE, 0, (strlen(UPLOAD_MAX_FILESIZE) - 1));
		$scale = strtolower(substr(UPLOAD_MAX_FILESIZE, -1));
		

		switch ($scale) {
			case 'g':
			echo "is g<br />";
				$num = $num * 1024;
			case 'k':
			echo "is k<br />";
				$num = $num / 1024;
		}
		echo floor($num);
```

returns: 

```
1G
is g
is k
1
```

which is wrong, in my opinion.

With breaks it results in:

```
1G
is g
1024
```

which seems right to me. Where does the comment `// no break` comes from and why is it noted explicitly?

If I got something wrong, let me know.

### DIFF
--- a/classes/core/PKPApplication.inc.php
+++ b/classes/core/PKPApplication.inc.php
@@ -910,7 +910,8 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
                 break;
             default:
                 // No suffix or nothing matched, so this is "b" (Byte)
-                $num = $num / 1024 / 1024
+                // Reset $num to the limit without cut the last digit
+                $num = UPLOAD_MAX_FILESIZE / 1024 / 1024;
                 break;
         }
         return floor($num);
--- a/classes/core/PKPApplication.inc.php
+++ b/classes/core/PKPApplication.inc.php
@@ -908,6 +908,10 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
             case 'k':
                 $num = $num / 1024;
                 break;
+            default:
+                // No suffix or nothing matched, so this is "b" (Byte)
+                $num = $num / 1024 / 1024
+                break;
         }
         return floor($num);
     }
--- a/classes/core/PKPApplication.inc.php
+++ b/classes/core/PKPApplication.inc.php
@@ -903,10 +903,11 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
         $scale = strtolower(substr(UPLOAD_MAX_FILESIZE, -1));
         switch ($scale) {
             case 'g':
-                $num = $num / 1024;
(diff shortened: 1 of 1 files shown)


asmecher:
You're right, that function is a mess! I suspect the GB case simply hasn't been used (it's pretty uncommon to express maximum upload size in GB, I think). But it looks like another case is missing too -- expressing it in just bytes, without a suffix. (See: https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes)

Commifreak:
You are right. I added that.

classes/core/PKPApplication.inc.php
```diff
@@ -903,10 +903,15 @@ public static function getIntMaxFileMBs()
...
-                // no break
-            case 'k':
                 $num = $num * 1024;
+                break;
+            case 'k':
+                $num = $num / 1024;
+                break;
+            default:
+                // No suffix or nothing matched, so this is "b" (Byte)
+                $num = $num / 1024 / 1024
```
asmecher, line 913: I think this will shear off the last digit of the byte count, no?

asmecher: (Also missing `;`)

asmecher:
(It's a shame that PHP doesn't provide something internally to convert this shorthand -- there's a note and a PHP code sample regarding this here: https://www.php.net/manual/en/function.ini-get.php)

classes/core/PKPApplication.inc.php
Commifreak: It will shear off the last digit? What do you mean?

Ill fix the typo asap - dint ran the code, just edited it here.

Commifreak: Ah, that shearing off... Got it!

asmecher:
I think you need to add a `m` suffix case, or else the `default:` will apply.

asmecher:
Thanks, @Commifreak, that's merged!

PR #7185, opened 2021-07-12
pkp/pkp-lib#7181 Provide application info to plugin gallery for fetch of more specialized XML

asmecher:


withanage:
Hi @asmecher ,
I have tested  the PR   for ojs 3.3.0 and confirm the  query string is built as it should be  (e.g. ojs2/3..3.0.1)  and  currently the full xml is fetched  in the current status and the plugin gallery is loaded into the gallery grid without errors.

PR #7201, opened 2021-07-24
pkp/pkp-lib#7190 - General fixes for the import

jonasraoni:
- Updated the "direct_sales_price" to work with a zeroed value
- Updated the user group to be matched in a case-insensitive way
- Fixed the storage of the cover image
- Updated the base64 decoding to be more permissive
- Fixed a bug due to a non-declared variable

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
@@ -127,7 +127,7 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 		if ($credit = $node->getAttribute('credit')) {
 			$submissionFile->setData('credit', $credit);
 		}
-		if ($directSalesPrice = $node->getAttribute('direct_sales_price')) {
+		if (strlen($directSalesPrice = $node->getAttribute('direct_sales_price'))) {
 			$submissionFile->setData('directSalesPrice', $directSalesPrice);
 		}
 		if ($genreId) {
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
@@ -302,6 +302,7 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 			}
 		}
 
+		$newFileId = null;
 		if ($temporaryFilename) {
 			$fileSizeOnDisk = filesize($temporaryFilename);
 			$expectedFileSize = $node->getAttribute('filesize');
@@ -320,10 +321,7 @@ class NativeXmlSubmissionFileFilter extends NativeImportFilter {
 			$deployment->setFileDBId($node->getAttribute('id'), $newFileId);
 		}
 
-		if (isset($newFileId)) {
-			return $newFileId;
-		}
-		return null;
+		return $newFileId;
 	}
 
 	/**
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
@@ -70,6 +70,7 @@ class NativeXmlSubmissionFilter extends NativeImportFilter {
 		$submissionDao = DAORegistry::getDAO('SubmissionDAO'); /* @var $submissionDao SubmissionDAO */
 		$submission = $submissionDao->newDataObject();
 
+		$submission->setData('locale', $node->getAttribute('locale') ?: $context->getPrimaryLocale());
 		$submission->setData('contextId', $context->getId());
 		$submission->stampLastActivity();
 		$submission->setData('status', $node->getAttribute('status'));
--- a/plugins/importexport/native/filter/NativeXmlPKPPublicationFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlPKPPublicationFilter.inc.php
@@ -62,7 +62,6 @@ class NativeXmlPKPPublicationFilter extends NativeImportFilter {
 	 */
 	function handleElement($node) {
 		$deployment = $this->getDeployment();
-		$context = $deployment->getContext();
 
 		$submission = $deployment->getSubmission();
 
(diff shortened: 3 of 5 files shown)


plugins/importexport/native/filter/NativeXmlPKPAuthorFilter.inc.php
```diff
@@ -77,11 +79,11 @@ function handleElement($node) {
 		if ($node->getAttribute('seq')) $author->setSequence($node->getAttribute('seq'));
 
 		// Identify the user group by name
-		$userGroupName = $node->getAttribute('user_group_ref');
+		$userGroupName = Stringy::create($node->getAttribute('user_group_ref'));
 		$userGroupDao = DAORegistry::getDAO('UserGroupDAO'); /** @var $userGroupDao UserGroupDAO */
 		$userGroups = $userGroupDao->getByContextId($context->getId());
 		while ($userGroup = $userGroups->next()) {
-			if (in_array($userGroupName, $userGroup->getName(null))) {
+			if ($userGroupName->containsAny($userGroup->getName(null) ?? [], false)) {
```
jonasraoni, line 86: I just applied this "fix" here, perhaps we should adopt case-insensitive comparisons on any generic string key 🤔

Another option, that would also address the value validation implicitly, would be to build the DTD on run-time, so we would be able to add value requirements to the attributes.

jnugent:
Hi @jonasraoni  I know this PR has been sitting a while now, I wanted to let you know that it still goes cleanly against stable-3_3_0 as I have just applied it to our test drive site.  I also fixes the problem with being unable to save the publication metadata.

One small problem is that the change to pkp-native.xsd, where the locale attribute has been removed, can we leave that in there?  There's still a locale field in the publications table currently and I don't think it hurts anything.  I added it back in my test because it was preventing me from importing XML I had (it was failing validation because the locale attribute was present).

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -127,7 +127,7 @@ function handleElement($node) {
 		if ($credit = $node->getAttribute('credit')) {
 			$submissionFile->setData('credit', $credit);
 		}
-		if ($directSalesPrice = $node->getAttribute('direct_sales_price')) {
+		if (strlen($directSalesPrice = $node->getAttribute('direct_sales_price'))) {
```
defstat, line 130: in `main` branch the fix for that is 

```
 $directSalesPrice = $node->getAttribute('direct_sales_price');
        if ($directSalesPrice != null) {
            $submissionFile->setData('directSalesPrice', $directSalesPrice);
        }
```
I suppose here you want to have the same affect?

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
```diff
@@ -282,7 +282,7 @@ function handleRevisionElement($node) {
 						if (($e = $childNode->getAttribute('encoding')) != 'base64') {
 							$deployment->addError(ASSOC_TYPE_SUBMISSION, $submission->getId(), __('plugins.importexport.common.error.unknownEncoding', array('param' => $e)));
 						} else {
-							$content = base64_decode($childNode->textContent, true);
```
defstat, line 285: we don't need the "strict" flag here? if no, should we remove the check below also?

asmecher: I would favour leaving the strict flag on, rather than potentially importing partial content without a failure.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.inc.php
jonasraoni: I can't remember very well what was happening, I just remember there was an issue with `direct_sales_price="0"`.

As the `getAttribute()` always returns a string, I think the `strlen` has a clearer intention, so I think I'll leave it here and perhaps update the /main (if there's something here that needs to be ported there) 🤔

plugins/importexport/native/filter/NativeXmlPKPAuthorFilter.inc.php
jonasraoni: Considering we're looking forward to replacing the plugin by the API, I'll ignore my suggestion here =]

plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
```diff
@@ -70,6 +70,7 @@ function handleElement($node) {
 		$submissionDao = DAORegistry::getDAO('SubmissionDAO'); /* @var $submissionDao SubmissionDAO */
 		$submission = $submissionDao->newDataObject();
 
+		$submission->setData('locale', $node->getAttribute('locale') ?: $context->getPrimaryLocale());
```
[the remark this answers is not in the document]
defstat: @jonasraoni is this visible to you?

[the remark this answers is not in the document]
jonasraoni: I did this modification because the `publications.locale` field isn't being used in `stable-3_3_0`, [its value is taken](https://github.com/pkp/pkp-lib/blob/stable-3_3_0/classes/publication/PKPPublicationDAO.inc.php#L55-L58) from the `submissions.locale` (which remains `null` without this modification and creates some issues).

The reasoning was kept in `main`, the `publications.locale` has been wiped off and now there's just the `submissions.locale`, so this behavior needs to be forwarded.

classes/filter/Filter.inc.php
```diff
@@ -455,6 +455,10 @@ function &execute(&$input, $returnErrors = false) {
 			$this->_output =& $preliminaryOutput;
 		}
 
+		if (!$this->_output) {
```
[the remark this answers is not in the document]
jonasraoni: Now this code just exists in the PR for `pkp-lib/main` =]

I avoid internationalizing error messages which are not displayed to the end user (this case), this is something that I agree: https://stackoverflow.com/a/474468

plugins/importexport/native/filter/PKPPublicationNativeXmlFilter.inc.php
```diff
@@ -84,11 +83,6 @@ function createEntityNode($doc, $entity) {
 		$isPublished = $entity->getData('status') === STATUS_PUBLISHED;
 		$isPublished ? $entityNode->setAttribute('seq', (int) $entity->getData('seq')) : $entityNode->setAttribute('seq', '0');
 
-		$entityLanguages = $entity->getData('languages');
```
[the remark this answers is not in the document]
jonasraoni: I guess the languages are handled here: https://github.com/pkp/pkp-lib/blob/main/plugins/importexport/native/filter/PKPPublicationNativeXmlFilter.inc.php#L316

So this one looks like forgotten old to me:
- The `Publication::getData('languages')` will probably return an array, then it's strange to use it in the next line
- There reverse filter, `NativeXmlPKPPublicationFilter`, doesn't have the inverse process to read the languages back

plugins/importexport/native/filter/NativeXmlSubmissionFilter.inc.php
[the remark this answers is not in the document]
jonasraoni: Here's a related issue: https://github.com/pkp/pkp-lib/issues/7958

asmecher:
Manually merged.

PR #7279, opened 2021-09-13
pkp/pkp-lib#6093 Add foreign key constraints

asmecher:


classes/install/Installer.inc.php
```diff
@@ -461,8 +461,10 @@ public function executeAction($action)
                     // Back out already-executed migrations.
                     while ($previousMigration = array_pop($this->migrations)) {
                         try {
-                            $previousMigration->down();
-                        } catch (PKP\install\DowngradeNotSupportedException $e) {
+                            if ($previousMigration->down() === true) {
```
Vitaliy-1, line 464: According to the method's signature, it returns boolean or null, but I don't see that any of those implementations return anything. Am I missing something?

classes/migration/upgrade/PKPv3_2_1SubeditorCategoriesMigration.inc.php
```diff
@@ -52,7 +51,7 @@ public function up()
     /**
      * Reverse the downgrades
      */
-    public function down()
+    public function down(): ?bool
     {
         throw new PKP\install\DowngradeNotSupportedException();
```
Vitaliy-1, line 56: Just noticed that needs a backslash (but is not actually related to the commit)

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -223,7 +222,7 @@ public function up()
     /**
      * Reverse the downgrades
      */
-    public function down()
+    public function down(): ?bool
     {
         throw new PKP\install\DowngradeNotSupportedException();
```
Vitaliy-1, line 227: The same as above

classes/install/Installer.inc.php
asmecher: The only one that returns anything is the PreflightCheckMigration. None of the other migrations need to break out of a downgrade at the moment.

classes/migration/upgrade/PKPv3_2_1SubeditorCategoriesMigration.inc.php
asmecher: Fixed, thanks!

classes/install/Installer.inc.php
asmecher: But I had been assuming that a `?bool` return typehint would allow missing return statements without errors/warnings (taking a `void` as a `null`) but trying with another PHP release it looks like I was wrong :( I will re-work this.

asmecher: Done -- I didn't want to force all migrations to `return false` for a single migration's convenience, so I went instead with emptying out the installer's migration list in the preflight check when a fallback version is set and further downgrades aren't wanted. It's a much cleaner solution.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
asmecher: Got rid of these!

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.inc.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function up(): void
+    {
+        try {
+            // pkp/pkp-lib#6903 Prepare to add foreign key relationships
+            // Clean orphaned assoc_type/assoc_id data in announcement_types
+            $contextDao = \APP\core\Application::getContextDAO();
+            $orphanedIds = DB::table('announcement_types AS at')->leftJoin($contextDao->tableName . ' AS c', 'at.assoc_id', '=', 'c.' . $contextDao->primaryKeyColumn)->whereNull('c.' . $contextDao->primaryKeyColumn)->orWhere('at.assoc_type', '<>', \Application::get()->getContextAssocType())->distinct()->pluck('at.type_id');
+            foreach ($orphanedIds as $typeId) {
+                error_log("Removing orphaned announcement type ID ${typeId} with no matching context ID.");
```
jonasraoni, line 33: I'm not sure if it's useful to have these atomic logs with IDs. Then I think it's better to just [remove without listing](https://github.com/pkp/pkp-lib/pull/7279/files#diff-00cbf4c41a927a521d0edbb75ce310a7d68fab5c72f6a14d5328841478095c5eR45).

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.inc.php
```diff
@@ -0,0 +1,85 @@
...
+            $orphanedIds = DB::table('announcement_settings AS a_s')->leftJoin('announcements AS a', 'a_s.announcement_id', '=', 'a.announcement_id')->whereNull('a.announcement_id')->distinct()->pluck('a_s.announcement_id');
+            foreach ($orphanedIds as $announcementId) {
+                error_log("Removing orphaned settings for missing announcement ID ${announcementId}");
+                DB::table('announcement_settings')->where('announcement_id', '=', $announcementId)->delete();
+            }
+        } catch (\Exception $e) {
+            if ($fallbackVersion = $this->setFallbackVersion()) {
+                $this->_installer->log("A pre-flight check failed. The software was successfully upgraded to ${fallbackVersion} but could not be upgraded further (to " . $this->_installer->newVersion->getVersionString() . '). Check and correct the error, then try again.');
+            }
+            throw ($e);
```
jonasraoni, line 58: I think it's useful for the user to know exactly where the error happened, so in this part https://github.com/pkp/pkp-lib/blob/main/tools/migration.php#L72 (and maybe others) I think it should be outputting the `Exception::toString()` instead of just the message 🤔

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.inc.php
asmecher: At least until we're more confident about supporting end users through data correction, I'd like to err on the side of being more verbose rather than less. (But I did move these to `installer->log` rather than `error_log`.)

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.inc.php
asmecher: Good point -- fixed. (This is already the case in the `Installer` when the stack trace is enabled in `config.inc.php`.)

Issue #6408, opened 2020-12-02
User-facing category page is inaccessible 


Vitaliy-1:
**Describe the bug**
A category page cannot be viewed. Receiving a fatal error while trying to access:
```
127.0.0.1:36424 [500]: GET /index.php/psp/catalog/category/first-category - Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'publication_tlpl.locale' in 'where clause' in /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:78
Stack trace:
#0 /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php(78): PDO->prepare()
#1 /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(361): Doctrine\DBAL\Driver\PDOConnection->prepare()
#2 /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(664): Illuminate\Database\Connection->Illuminate\Database\{closure}()
#3 /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(631): Illuminate\Database\Connection->runQueryCallback()
#4 /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(373): Illuminate\Database\Connect in /home/doc/OJS/development/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php on line 671
``` 

**To Reproduce**
1. Create a category page, e.g., accessible by URL `../first-category`
2. Check the page available at: `.../[journal_path]/catalog/category/first-category`

**What application are you using?**
 OJS 3.3

asmecher:
[...]

PR #7309, opened 2021-09-20
 Fixes problem with retrieving publications in alphabetically ordered category when changing language

AlineLepidus:
Change in the SQL query by title and order of presentation of publications, as it was only retrieving publications with titles that have translations in the primary and current language.
When choosing the language and the publications did not have a translated title, it was not presented in the category.

Signed-off-by: AlineLepidus <aline@lepidus.com.br>

### DIFF
--- a/classes/services/queryBuilders/PKPSubmissionQueryBuilder.inc.php
+++ b/classes/services/queryBuilders/PKPSubmissionQueryBuilder.inc.php
@@ -90,7 +90,7 @@ abstract class PKPSubmissionQueryBuilder extends BaseQueryBuilder implements Ent
 		} elseif ($column === 'dateLastActivity') {
 			$this->orderColumn = 's.date_last_activity';
 		} elseif ($column === 'title') {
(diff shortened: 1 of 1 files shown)


asmecher:
@AlineLepidus, could you describe the cause in a little more detail? We have a widespread practice of considering both a "preferred" (current) and "fallback" (primary, guaranteed to have data) locale when fetching localized content. In this case it's the current locale, falling back on the submission locale. Do you have cases where there is no title in the submission's locale?

(See related: https://github.com/pkp/pkp-lib/issues/6408)

AlineLepidus:
@asmecher
No, the case is when don't have title in current locale.
In shorts words, when don't have title in current locale and choosing category alphabetically ordered, the publication don't appear same having title in primary locale.

It's better explained in [forum](https://forum.pkp.sfu.ca/t/category-does-not-list-all-articles-when-sorting-by-title-is-selected/69168),  the version is different, but the problem is the same.

Issue #6091, opened 2020-07-15
Enable all classes to be autoloaded


jonasraoni:
**Describe the problem you would like to solve**
Not knowing which classes/constants exist at a given scope, or worse, having to import all of them manually is a possible cause of problems.
If a call located at the beginning of the call stack stops loading a given class/constant, it will create a breaking change, as the deeper calls may not be loading everything they need.

**Describe the solution you'd like**
I would like to have a generic auto_loader, which is able to discover all the existing classes.
To achieve this, I guess all the classes should follow a strict namespace pattern (obeying a naming/folder structure).

TODO:
- [x] Permit autoloading of everything in `classes`
- [x] Permit autoloading of controllers
- [x] Permit autoloading of pages
- [x] Permit autoloading of plugins
- [x] Rename all `.inc.php` to `.php`
  - [x] Controllers
  - [x] Classes
  - [x] Included plugins
  - [x] Major external plugins
  - [x] Pages
  - [x] API Handlers
  - [x] Others
- [x] Scan for and fix unnecessary calls to import
- [x] Survey other potential areas for autoloading
- [x] Fix scanning for mailable class names (see https://github.com/pkp/pkp-lib/pull/7444#discussion_r747413300)

asmecher:
There is already a custom [autoloader in `lib/pkp/include/functions.inc.php`](https://github.com/pkp/pkp-lib/blob/3_2_1-0/includes/functions.inc.php#L321) called `customAutoload`, and it's [installed in `lib/pkp/classes/core/PKPApplication.inc.php`](https://github.com/pkp/pkp-lib/blob/3_2_1-0/classes/core/PKPApplication.inc.php#L203..L212).

It's not used everywhere, though; there are lots of places where `import` statements could be replaced with full namespaces to promote autoloading. Here's [one example](https://github.com/pkp/ojs/blob/3_2_1-1/pages/workflow/WorkflowHandler.inc.php#L73) where the autoloader is used instead of an `import` statement. This will be a matter of ongoing cleanup (and hands are welcome).

I believe there will be a few cases where an autoloader by itself won't be enough, e.g. when unserializing an object when the class definition is not loaded. That's used in some of the `_settings` tables, and will probably be necessary in some of the caching code as well when we reimplement the object caching as part of https://github.com/pkp/pkp-lib/issues/3304 (not yet merged). It's probably the case that when the class specifies a namespace, it's included in the serialized form...

jonasraoni:
[...]

Issue #6080, opened 2020-07-09
Ability to rename locales (removing/changing country names)


ctgraham:
**Describe the problem you would like to solve**
Journal Managers would like to remove or change the country designations from the locales as named in `registry/locales.xml`, but this is not configurable by the Journal Manager or Site Administrator.  Currently it can only be accomplished by a local code change.

**Describe the solution you'd like**
First preference: a plugin which can modify the locale names; Second preference: a setting to point to an alternate locales.xml file.

**Who is asking for this feature?**
Journal Managers, Site Administrators

**Additional information**
The PKP plugin approach is implausible, as the locales are loaded as the generic plugins are loaded.
```
  00 : 'lib/pkp/classes/cache/GenericCache.inc.php:63' called 'PKPLocale::_allLocalesCacheMiss',
  01 : 'lib/pkp/classes/cache/FileCache.inc.php:115' called 'GenericCache->get',
  02 : 'lib/pkp/classes/i18n/PKPLocale.inc.php:671' called 'FileCache->getContents',
  03 : 'lib/pkp/classes/i18n/PKPLocale.inc.php:381' called 'PKPLocale::_getAllLocalesCacheContent',
  04 : 'lib/pkp/classes/context/Context.inc.php:265' called 'PKPLocale::getAllLocales',
  05 : 'classes/template/TemplateManager.inc.php:76' called 'Context->getSupportedLocaleNames',
  06 : 'lib/pkp/classes/template/PKPTemplateManager.inc.php:885' called 'TemplateManager->initialize',
  07 : 'lib/pkp/classes/plugins/Plugin.inc.php:352' called 'PKPTemplateManager::getManager',
  08 : 'lib/pkp/classes/plugins/Plugin.inc.php:118' called 'Plugin->_registerTemplateResource',
  09 : 'lib/pkp/classes/plugins/LazyLoadPlugin.inc.php:28' called 'Plugin->register',
  10 : 'plugins/generic/lensGalleyBits/LensGalleyBitsPlugin.inc.php:23' called 'LazyLoadPlugin->register',
  11 : 'lib/pkp/classes/plugins/PluginRegistry.inc.php:69' called 'LensGalleyBitsPlugin->register',
  12 : 'lib/pkp/classes/plugins/PluginRegistry.inc.php:143' called 'PluginRegistry::register',
  13 : 'lib/pkp/classes/core/Dispatcher.inc.php:132' called 'PluginRegistry::loadCategory',
[...]

PR #7336, opened 2021-09-25
Feature/main/i6328 improved localization

jonasraoni:
**Description**
- Instead of registering a specific file (e.g. `plugin/locale/en/common.po`), the code is now expecting a folder (e.g. `plugin/locale`), which should follow a single rule: its direct sub-folders should refer to locale names (e.g. `./en_US`, `./pt_BR`, etc.).
- The application keeps track of the paths, but doesn't load anything until a translation is requested.
- When a translation for a specific locale is requested, it tries to find a `LocaleBundle` (a **merged** collection with all the translation files found for the given locale), if not found, it will build one.

**LocaleBundle**
- When a `LocaleBundle` is requested, the `Locale` class will pass through the registered folders (it's also possible to register a callback/dynamic file loader), access the sub-folder related to the requested locale (e.g. `./en_US`), and retrieve the available `.po` files recursively. These files will be sent to the newly created `LocaleBundle`.

**Translation**
- When a translation is requested, the `LocaleBundle` will attempt to find the translations (merged .po files) in the disk/cache. The cache key is built using the path + modified date of its locale files + expiration interval. This way, any modification in the expiration, order of files, paths or modification dates will invalidate the cache.
- Since a translation might be requested before the application and all the plugins registered their folders, every different combination of available folders, at the time of the request, will issue a new cache (in my OJS installation, after installing/running some things, I've ended up with 8 cached files for the `en_US` bundle, each one with around ~450KB).
- If there's no cache, every .po file of the bundle will have to be parsed, transformed into an array, and have its translations merged sequentially into the bundle (there's an optional priority/order argument). This process also has an individual cache for each locale file (invalidated by the modification date).

Merging physically the locale files into a single one would also give a little performance bump.


jonasraoni:
I separated the commit in four pieces: the feature is in the first, the next two are mostly editor replacements, and the last has generic updates I did in the way (this one I could move into another issue, but didn't).

**Description**
- Instead of registering a specific file (e.g. `plugin/locale/en/common.po`), the code is now expecting a folder (e.g. `plugin/locale`), which should follow a single rule: its direct sub-folders should be mappable to locale names (e.g. `./en_US`, `./pt_BR`, etc.).
- The application keeps track of the paths, but doesn't load anything until a translation is requested.
- When a translation for a specific locale is requested, it tries to find a `LocaleBundle` (a collection with all the locale files for the given locale), if not found, it will build one.

**LocaleBundle**
- When a `LocaleBundle` is requested, the `Locale` class will pass through the registered folders (it's also possible to register a callback/dynamic file loader), access the sub-folder related to the requested locale (e.g. `./en_US`), and retrieve the available `.po` files recursively. These files will be sent to the newly created `LocaleBundle`.

**Translation**
- When a translation is requested, the `LocaleBundle` will attempt to find the translations in the cache. The cache key is built using the path + modified date of its locale files. This way, any modification in the order of files, paths or modification dates will invalidate the cache.
- As the cache key is dynamic (this is intentional, since a translation might be requested before all the plugins registered their folders), every different combination will create a new cache (in my OJS installation, after installing/running some things, I've got 8 cached files for the en_US locale, each one with ~450KB).
- If there's no cache, every locale file of the bundle will have to be parsed, transformed into an array, and have its translations merged sequentially into the bundle (there's an optional priority order). This process has an individual cache for each locale file (invalidated by the modification date).

ps: If we're ready, I could already merge physically the locale files into a single one.

classes/core/PKPContainer.inc.php
```diff
@@ -184,7 +196,19 @@ protected function loadConfiguration()
             'auth_mode' => null,
         ];
 
-        $this->instance('config', new Repository($items)); // create instance and bind to use globally
+        // Cache configuration
+        $items['cache'] = [
```
jonasraoni, line 200: Later on, this should be configurable, to allow the user to use a distributed caching solution.

classes/i18n/LocaleConversion.inc.php
```diff
@@ -0,0 +1,229 @@
...
+ * @ingroup i18n
+ *
+ * @brief Provides methods to convert locales from one format to another
+ */
+
+namespace PKP\i18n;
+
+use PKP\facades\Locale;
+
+class LocaleConversion
```
jonasraoni, line 26: I found a package to address these conversions, but as we have some "business logic" here, I just had it moved to a separate class.

classes/i18n/Locale.inc.php
```diff
@@ -0,0 +1,428 @@
...
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use RecursiveRegexIterator;
+use RegexIterator;
+use SimpleXMLElement;
+
+class Locale implements LocaleInterface
+{
+    /** @var string Keeps the locales available in the system */
+    protected const LOCALE_REGISTRY_FILE = 'registry/locales.xml';
```
jonasraoni, line 49: I'd like to remove this XML file and add the locales through code, is there something against?

asmecher: Can you describe what you have in mind? See related: https://github.com/pkp/pkp-lib/issues/6080

jonasraoni: Given that this file is used only by the `getLocales()` method.

- I was going to build the list of available locales based on the existing locale files.
- And about the "metadata", I thought about using a package, but I didn't find something that address all the needs in a fast search, just:
  - https://github.com/leodido/langcode-conv: has iso639-1, iso639-2/t, iso639-2/b and iso639-3, but not the direction (rtl/ltr)
  - https://punic.github.io: has a lot of locale information, inclusive the direction, but not the other representations

If there isn't a suitable package, I'd just transform the file into code:
```php
return [
  new LocaleMetadata(...),
  ...
];
```

---

About the issue, it's possible to replace the `Locale` in the Laravel container and provide a customized implementation.

asmecher: We already have ISO639-3 included in OJS/OMP/OPS as part of https://github.com/sokil/php-isocodes (which is already used for countries and currencies).

I think the features that'll be hard to keep are:
- LRT vs RTL
- Completeness indication (I suppose this could be gleaned when generating the cache; currently we go by a heuristic where around 90% of the strings in the English language must be covered by the translation to be considered complete)
- Country-specific stylesheet (this can be optionally specified with the `stylesheet` attribute, and is useful for RTL languages)

classes/core/PKPApplication.inc.php
```diff
@@ -989,6 +989,28 @@ public static function getMetadataFields()
...
+    public static function isInstalled(): bool
+    {
+        return !!Config::getVar('general', 'installed');
+    }
+
+    /**
+     * Retrieves if the application is running an upgrade
+     * @param bool $upgrade If a value is passed, it will be used to set if the application is upgrading
+     */
+    public static function isUpgrading(?bool $upgrade = null): bool
```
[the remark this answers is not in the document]
asmecher: We're not using this pattern elsewhere in PHP; I'd suggest moving the setter to its own function.

asmecher:
One more request: could you document the changes that developers using older versions of OJS will need to know in adapting their code to this? It'll give me something to think about while reviewing, plus @NateWr can collect this for his release notebook. See https://github.com/pkp/pkp-lib/issues/6091#issuecomment-796314189 for an example.

jonasraoni:
Ok! I think I'll create new issues for these extra modifications.

classes/i18n/Locale.inc.php
jonasraoni: - LTR/RTL: I'll check if there's a small package providing it, I saw that "punic" package has, but I didn't test.
- Completeness: Are we using this somewhere? I think it would be useful to have it displayed on the locale selection (90% complete. [Contribute with translations here])
- Country-specific stylesheet: I removed this just to create a discussion. Do you have any idea about the usage? I think it should be implemented as a file upload in the administrative page 🤔

asmecher: Completeness is currently displayed on the installation form and will let users know what to expect, including helping them choose between competing variants (e.g. `pt_PT` and `pt_BR`; `fr_CA` and `fr_FR`).

About the stylesheet: It's a language-specific stylesheet, not a country-specific one. This can be very useful with non-Latin scripts and layouts, e.g. Hebrew, Arabic, etc.

classes/core/PKPApplication.inc.php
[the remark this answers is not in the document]
jonasraoni: I updated the code to use: `Application::upgrade()` and `SessionManager::disable()`.

classes/i18n/Locale.inc.php
jonasraoni: @asmecher Anything against moving the locale stylesheet into the administration page (file upload/textarea)?
If that's ok, I can create an issue `remove the locales.xml` to address this and the other items.

asmecher: I think the idea was that we would get translators/user communities to contribute stylesheets that would be shipped as part of the application -- e.g. an `ar_IQ`-specific stylesheet that fixes display issues for that language. That hasn't happened yet, but I think it's an important feature.

jonasraoni: Great =]
@asmecher did you see the other comments that I left in this PR? If you've already checked them and don't have any specific observation/opinion, then just let me know... I'll be working on them next days.

classes/facades/Locale.inc.php
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Locale
+ *
+ * @brief This facade provides access to the locale
+ */
+
+namespace PKP\facades;
```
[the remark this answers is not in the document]
jonasraoni: The interface which I implemented to handle the translations (`\Illuminate\Contracts\Translation\Translator`) has a method with this name, so it's not possible, unless I change the code structure.

My biggest motivation was getting rid of the "Facade" (personal pre-concept), but given that we're already heavily using them (e.g. `DB`) and it kind of works faster than using the `app()`, I'll keep it this way...

classes/i18n/Locale.inc.php
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * @copy \Illuminate\Contracts\Translation\Translator::getLocale()
+     */
+    public function getLocale(): string
+    {
+        if (!$this->locale) {
+            $request = $this->_getRequest();
+            $locale = $request->getUserVar('uiLocale')
```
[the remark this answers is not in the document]
jonasraoni: Looks like it's something that doesn't exist anymore, I'll remove...

classes/facades/Locale.inc.php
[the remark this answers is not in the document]
asmecher: @henriqueramos has also [proposed changing](https://github.com/pkp/pkp-lib/pull/7166/files#diff-2a88d2b16f09aaae268c624e13a2605dce904385c8cb3cc4ecdd20e44bbf942b) `App::make` to `app()->make`; I don't have enough depth of knowledge of the Laravel idiom to know which is better for us, but would like us to stay consistent. Could the two of you chat a bit and make a recommendation?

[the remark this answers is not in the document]
jonasraoni: There was a small discussion about it here: https://github.com/pkp/pkp-lib/pull/7166/files#r737492347

I would normally opt-in for the `App::make()` because it's less likely to cause conflicts (namespaced), but as the `app()` works slightly faster, I ended up replacing the `App::make(xxx)` by `app(xxx)`.

classes/i18n/translation/IsoCodesTranslationDriver.inc.php
```diff
@@ -0,0 +1,80 @@
...
+
+namespace PKP\i18n\translation;
+
+use DomainException;
+use Gettext\Loader\MoLoader;
+use PKP\i18n\interfaces\LocaleInterface;
+use PKP\i18n\translation\Translator;
+use Sokil\IsoCodes\TranslationDriver\TranslationDriverInterface;
+
+class IsoCodesTranslationDriver implements TranslationDriverInterface
```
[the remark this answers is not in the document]
asmecher: Could you drop a link to the upstream `sokil/php-isocodes` issue here?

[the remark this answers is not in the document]
jonasraoni: Sure, here it is: https://github.com/sokil/php-isocodes/issues/51

I've also created a PR here, which was merged already: https://github.com/elcobvg/laravel-opcache/issues/17

classes/user/InterestManager.inc.php
```diff
@@ -63,14 +59,13 @@ public function getInterestsForUser($user)
...
         foreach ($interestDao->getUserInterestIds($user->getId()) as $interestEntryId) {
-            if (!isset($interestsCache[$interestEntryId])) {
-                $interestsCache[$interestEntryId] = $interestEntryDao->getById(
-                    $interestEntryId,
-                    $controlledVocab->getId()
-                );
-            }
-            if (isset($interestsCache[$interestEntryId])) {
-                $interests[] = $interestsCache[$interestEntryId]->getInterest();
+            /** @var InterestEntry */
```
jonasraoni, line 62: My misunderstanding, the interest entry isn't based on the user.

classes/core/PKPContainer.inc.php
```diff
@@ -215,7 +210,11 @@ protected function loadConfiguration()
             'channel' => 'errorlog',
         ];
 
-        $items['mail']['default'] = static::getDefaultMailer();
+        $items['mail']['default'] = static::getDefaultMailer() === 'disabled' ? 'sendmail' : static::getDefaultMailer();
+        // Replace it by a mocked instance, just to keep things working
+        if (static::getDefaultMailer() === 'disabled') {
+            Mail::fake();
+        }
```
jonasraoni, line 217: Mostly to hide error messages in the build, but perhaps it's useful for an user that doesn't have an email server o_O

classes/task/FileLoader.inc.php
```diff
@@ -387,27 +387,6 @@ private function _stageFile()
     {
         $this->moveFile($this->_processingPath, $this->_stagePath, $this->_claimedFilename);
     }
-
-    /**
-     * Send the passed message to the administrator by email.
-     *
-     * @param string $message
-     */
-    private function _notify($message, $messageType)
```
jonasraoni, line 401: Unused method.

Besides the `new Mail()`, there are just other two places using the old Mail class. I was going to update, but I think Vitaliy is working on it.

locale/en_US/common.po
```diff
@@ -585,6 +585,9 @@ msgstr "Override"
 msgid "common.pagination"
 msgstr "{$start}-{$end} of {$total}"
 
+msgid "common.listSeparator"
```
jonasraoni, line 588: I found these missing locales in the error log

classes/core/PKPString.inc.php
```diff
@@ -31,18 +30,16 @@ class PKPString
...
-
-            // Set up required ini settings for mbstring
-            // FIXME Do any other mbstring settings need to be set?
-            mb_internal_encoding($clientCharset);
+        static $isInitialized;
+        if (!$isInitialized) {
+            if (self::hasMBString()) {
+                // Set up default encoding
+                mb_internal_encoding('utf-8');
+                ini_set('default_charset', 'utf-8');
```
jonasraoni, line 40: Just centralized the charset settings in one place.

classes/install/form/InstallForm.inc.php
```diff
@@ -66,43 +65,18 @@ public function __construct($request)
...
-            '' => __('common.notApplicable'),
-            'utf8' => 'Unicode (UTF-8)'
-        ];
-
-        $this->supportedDatabaseCharsets = [
-            '' => __('common.notApplicable'),
-            'utf8' => 'Unicode (UTF-8)'
-        ];
-
-        $this->supportedDatabaseDrivers = [
```
jonasraoni, line 82: Removed fixed settings and databases which we're not supporting.

classes/template/PKPTemplateManager.inc.php
```diff
@@ -284,7 +285,7 @@ public function initialize($request)
         }
 
         // Register custom functions
-        $this->registerPlugin('modifier', 'translate', '\__');
+        $this->registerPlugin('modifier', 'translate', [$this, 'smartyTranslateModifier']);
```
jonasraoni, line 288: TODO: Document the new syntaxes to use pluralized locale keys in the templates, I've extended the translation modifier and the translate tag.

templates/install/install.tpl
```diff
@@ -22,8 +22,8 @@
 	<br />
 
 	<div class="app__contentPanel">
-		{fbvFormSection label="installer.locale" for="locale" style="position: absolute;"}
-			{fbvElement type="select" name="locale" id="installLanguage" from=$localeOptions selected=$locale translate=false size=$fbvStyles.size.SMALL subLabelTranslate=true}
+		{fbvFormSection label="common.language" for="installLanguage" style="position: absolute;"}
```
jonasraoni, line 25: I've added this field to change the language during the installation, but it needs some styling, I can do it, but I decided to ensure it's not going to be removed :)

Another generic improvement would be to choose a default locale considering the locales from the `Accept-Language` header.

classes/site/VersionCheck.inc.php
```diff
@@ -72,82 +72,55 @@ public static function getCurrentDBVersion()
...
+                $version['patch'] = [];
+                foreach ($xml->patch as $patch) {
+                    $version['patch'][$patch['from']] = (string) $patch;
+                }
+            }
+            $cache = [
+                'createdAt' => new DateTime(),
+                'data' => $version
+            ];
+            Cache::put($key, $cache);
```
jonasraoni, line 108: Hmm, there should be an extra check here (`FileManager::isVirtualPath($path)`)

templates/install/install.tpl
asmecher: :+1: agreed, an `Accept-Language` check would be great -- this should be done in the session manager where sessions are initiated.

asmecher: (But no need to tackle that here -- in fact better file a new issue for it.)

tests/registry/locales.xml
```diff
@@ -1,23 +0,0 @@
...
-  *
-  * Locales registry file.
-  *
-  * This file lists all locales available to the system.
-  * (i.e., existing under the locales/ directory)
-  -->
-
-<!DOCTYPE locales SYSTEM "../../dtd/locales.dtd">
-
-<locales>
```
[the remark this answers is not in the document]
asmecher: The `bic21qualifiers` and `bic21subjects` files (XML and DTD) are definitely used in OMP. The rest can be removed (some dependent on your locale-related changes).

classes/task/FileLoader.inc.php
asmecher: Yes, tagging @Vitaliy-1.

Vitaliy-1: Thanks, I'll replace those with correspondent Mailable classes

locale/en_US/common.po
Vitaliy-1: I suspect it was replaced with `common.commaListSeparator` and `common.semicolonListSeparator`, at least I used those.

jonasraoni: Thanks! So I'll remove this locale and update the places which are still using the old one.

classes/i18n/translation/IsoCodesTranslationDriver.inc.php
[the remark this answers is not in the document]
asmecher: Thanks!

templates/install/install.tpl
```diff
@@ -136,9 +136,6 @@
 				{fbvFormSection label="installer.databaseDriver" description="installer.databaseDriverInstructions"}
 					{fbvElement type="select" id="databaseDriver" from=$databaseDriverOptions selected=$databaseDriver translate=false size=$fbvStyles.size.SMALL}
 				{/fbvFormSection}
-				{fbvFormSection label="installer.connectionCharset"}
-					{fbvElement type="select" id="connectionCharset" from=$connectionCharsetOptions selected=$connectionCharset translate=false size=$fbvStyles.size.SMALL}
-				{/fbvFormSection}
```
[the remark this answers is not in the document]
asmecher: Not quite true -- leaving it blank allows the DB to default!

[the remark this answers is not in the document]
jonasraoni: I took this decision partially based on this:
https://github.com/pkp/pkp-lib/blob/da5fe5542c79d0e8b5cdff2fd8a88c41a8896bab/classes/install/PKPInstall.inc.php#L108-L109

I'll double check later. By navigating in GitHub, looks like it will store "[Off](https://github.com/pkp/pkp-lib/blob/da5fe5542c79d0e8b5cdff2fd8a88c41a8896bab/classes/install/PKPInstall.inc.php#L218)" in the config, which will be restored as [false](https://github.com/pkp/pkp-lib/blob/52ff51c8e481dcbf6d83fc75103812dd25070b12/classes/config/ConfigParser.inc.php#L100), then I'm curious what happens when it's passed to [Laravel](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Connectors/MySqlConnector.php#L69).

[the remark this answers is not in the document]
asmecher: Please do let me know what you turn up -- don't forget that we still have ADODB to consider, as long as we're still carrying the upgrade steps prior to 3.3.0; there will probably be considerations for ADODB as well as Laravel.

classes/core/PKPApplication.inc.php
```diff
@@ -931,6 +965,41 @@ public static function getMetadataFields()
...
+     */
+    public static function isUpgrading(): bool
+    {
+        return defined('RUNNING_UPGRADE');
+    }
+
+    /**
+     * Retrieves whether the application is ready (installed and not being upgraded)
+     */
+    public static function isReady(): bool
```
NateWr, line 988: I just noticed this helper function and I like it! Can I make a naming suggestion? I feel like `isReady()` could mean a lot of things (ie - "is ready to be installed"). Can we call it something like `isRunning()` or `isActive()` or `isLive()`?

Or we could go the opposite route and prepare for a maintenance mode by renaming it to `isUnderMaintenance()` and returning the opposite `bool`.

jonasraoni: Sure! `isUnderMaintenance()` sound better for me =]

asmecher: Don't forget this change, @jonasraoni!

classes/i18n/Locale.inc.php
```diff
@@ -0,0 +1,428 @@
...
+            $locales = array_map('utf8_decode', $locales);
+        }
+
+        return $locales;
+    }
+
+    /**
+     * @copy LocaleInterface::installLocale()
+     */
+    public function installLocale(string $locale): void
```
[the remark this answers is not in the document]
asmecher: @Vitaliy-1's branch might have already modified aspects of this; I'd suggest checking with him.

classes/i18n/LocaleMetadata.inc.php
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Whether the locale expects text on the right-to-left format
+     */
+    public function isRightToLeft(): bool
+    {
+        $locale = $this->_parse();
+        $language = strtolower($locale->language ?? '');
+        $script = strtolower($locale->script ?? '');
+        $rightToLeftLanguages = array_fill_keys(['ar', 'dv', 'fa', 'he', 'ku', 'nqo', 'prs', 'ps', 'sd', 'syr', 'ug', 'ur', 'yi'], true);
```
asmecher, line 93: Hmm, there were several benefits of the old `locales.xml` descriptor that we're losing here...
- Easy customization of the language display names,
- Easy determination of which languages were "complete enough" to even include in the installer UI,
- RTL vs. RTL status,
- Language-specific CSS file support

I'm not a fan of hard-coding lists like this, though admittedly this is one that is not likely to change often.

On the other hand, it's nice to have the completeness etc. automatically determined.

@NateWr, if I could ask a second opinion on this -- do you prefer descriptor or no descriptor?

classes/site/VersionCheck.inc.php
asmecher: (Don't forget this comment!)

classes/template/PKPTemplateManager.inc.php
asmecher: (this one also needs attention)

composer.json
```diff
@@ -50,7 +52,6 @@
 				"Apply PKP ADODB patches": "lib/adodb.diff"
 			},
 			"laravel/framework": {
-				"Inhibit __ Laravel helper": "lib/laravel-helper-4017.diff",
```
asmecher, line 53: Just curious about this -- the `gettext/translator` package appears to provide its own `__` function, as does `laravel/framework`. How do these two avoid collision?

jonasraoni: Sure, the `__()` function from `gettext/translator` isn't included by default.

classes/template/PKPTemplateManager.inc.php
jonasraoni: I've documented the usage in the code, but perhaps it should be done in the PKP Docs as well (in case we have a section explaining the internal tags/filters).

classes/i18n/Locale.inc.php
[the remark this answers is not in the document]
jonasraoni: Ok!

classes/i18n/Locale.inc.php
```diff
@@ -0,0 +1,428 @@
...
+        // Delete locale-specific data
+        $emailTemplateDao = DAORegistry::getDAO('EmailTemplateDAO'); /** @var EmailTemplateDAO $emailTemplateDao */
+        $emailTemplateDao->deleteEmailTemplatesByLocale($locale);
+        $emailTemplateDao->deleteDefaultEmailTemplatesByLocale($locale);
+    }
+
+    /**
+     * @copy LocaleInterface::getSupportedFormLocales()
+     */
+    public function getSupportedFormLocales(): array
```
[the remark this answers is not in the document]
jonasraoni: For now I'll just leave a comment to avoid using them if possible 😁

templates/install/install.tpl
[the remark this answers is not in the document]
jonasraoni: Just to confirm, for now this setting value is really breaking things:
```php
PHP Fatal error: Uncaught PDOException: SQLSTATE[42000]:
Syntax error or access violation: 1115 Unknown character set: '' in /var/www/html/lib/pkp/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:79
```

classes/i18n/Locale.inc.php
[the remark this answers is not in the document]
Vitaliy-1: (That branch was merged if it's still relevant)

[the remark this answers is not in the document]
jonasraoni: Thanks @Vitaliy-1.

classes/i18n/LocaleMetadata.inc.php
NateWr: I think the RTL/CSS stuff is ok without a descriptor. That's not something that changes, really.

I do like the idea of allowing an installation to configure which locales it supports with a descriptor of some kind. We don't really have a pattern for this, but I've sometimes thought about having a directory in the `files_dir` where an install could place some files to override some of our registry files. (For example, if I as a host want a different set of default user groups, I could add a `files_dir/config/userGroups.xml` file with my own.) That's a use-case for keeping some of these registry files around where it is easy to do so, but I don't have a strong preference in this case.

PR #7351, opened 2021-09-29
return value of getTitles()

ohilbig01:
the function getTitles() should return the locales as keys, like in getFullTitles()

### DIFF
--- a/classes/publication/PKPPublication.inc.php
+++ b/classes/publication/PKPPublication.inc.php
@@ -134,7 +134,7 @@ class PKPPublication extends \PKP\core\DataObject
             if (!$title) {
                 continue;
             }
-            $return[] = $this->getLocalizedTitle($locale);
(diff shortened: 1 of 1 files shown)


asmecher:
@ohilbig01, is there calling code e.g. in OJS or in a plugin that depends on having the locale as the key? Or is this a feature request for external (3rd-party) code?

ohilbig01:
Hi @asmecher, I don't  think OJS or OMP are calling it, but I use it for a plugin I'm writing. I am using `getTitles()` in the way `getFullTitles()` is used in `Dc11SchemaPublicationFormatAdapter.inc.php`, so I thought, it should return the array with the locales as key. 
If this return value is intentional, I can do it differently.

asmecher:
Thanks, @ohilbig01 -- no objection here! I'll also cherry-pick this to `stable-3_3_0` for the next release.

PR #7414, opened 2021-10-31
pkp/pkp-lib#6991 Improve user search query

jonasraoni:


classes/security/UserGroupDAO.inc.php
```diff
@@ -779,46 +776,76 @@ function deleteSettingsByLocale($locale) {
...
-				$searchSql .= ' AND (  ' . join(' AND ', $searchFieldMap) . '  ) ';
+				$searchSql .= 'AND (
+					' . implode(' AND ', $usersFilter) . '
+					OR EXISTS(
+						SELECT 0
+						FROM user_settings us
+						WHERE
+							u.user_id = us.user_id
+							AND (' . implode(' AND ', $settingsFilter) . ')
+					)
```
jonasraoni, line 827: There were many joins on the `user_settings` just to check if there's a match, so I reduced to a single sub-query.

But the code isn't the same, here I'm checking for any user setting, while the previous code is searching just in the given name, family name and affiliation.

classes/security/UserGroupDAO.inc.php
```diff
@@ -779,46 +776,79 @@ function deleteSettingsByLocale($locale) {
...
+				// Search the user settings
+				$filters[] = '
+					EXISTS(
+						SELECT 0
+						FROM user_settings us
+						WHERE
+							u.user_id = us.user_id
+							' . str_repeat(' AND LOWER(us.setting_value) LIKE LOWER(?)', count($terms)) . '
+					)';
+				array_push($params, ...$terms);
```
jonasraoni, line 825: Hmmm, I'll probably submit an update to move this into the line 813 and use a `GROUP_CONCAT`/`STRING_AGGR`, otherwise the search won't be very effective.

jonasraoni: Updated... The performance is still great in my use-case.

asmecher:
This is likely something we'll need to offer as a patch for affected users but not merge/release with 3.3.0-x for compatibility's sake. We list MySQL 4.1 and PostgreSQL 9.5 as base requirements for 3.3.0 (in `docs/README.md`) and while I haven't checked specifically I suspect this is too fancy for MySQL 4.1. (Verification is welcome on that!)

We have more freedom to bump that requirement with 3.4.0 and have used subqueries much more in the existing rewrite of these queries.

jonasraoni:
@asmecher I had checked the [documentation of MySQL 4.1](https://downloads.mysql.com/docs/refman-4.1-en.pdf) before adding the `GROUP_CONCAT`, luckily this is the exactly version where it was introduced `GROUP_CONCAT() [875] was added in MySQL 4.1.`.

I just did a small update to the PR, since `$groupConcat(setting_value, '')` would produce a different output in MySQL (`a,b`) and PostgreSQL (`ab`).

classes/security/UserGroupDAO.inc.php
```diff
@@ -778,47 +779,82 @@ function deleteSettingsByLocale($locale) {
...
+				switch (get_class(Capsule::connection())) {
+					case MySqlConnection::class:
+						$concatSettingValue = "GROUP_CONCAT(setting_value SEPARATOR '')";
+						break;
+					case PostgresConnection::class:
+						$concatSettingValue = "STRING_AGG(setting_value, '')";
+						break;
+					default:
+						throw new DomainException('Unrecognized database');
 				}
```
jonasraoni, line 815: This is the second time I've added such customization, perhaps there's a library providing such extensions out there (I doubt)... If not, perhaps we should polyfill the Laravel Database or centralize these customizations somewhere.

classes/security/UserGroupDAO.inc.php
```diff
@@ -778,47 +779,82 @@ function deleteSettingsByLocale($locale) {
...
+
+				switch (get_class(Capsule::connection())) {
+					case MySqlConnection::class:
+						$concatSettingValue = "GROUP_CONCAT(setting_value SEPARATOR '')";
+						break;
+					case PostgresConnection::class:
+						$concatSettingValue = "STRING_AGG(setting_value, '')";
+						break;
+					default:
+						throw new DomainException('Unrecognized database');
```
jonasraoni, line 814: This is the second time I've added such customization, perhaps there's a library providing such extensions out there (I doubt)... If not, perhaps we should polyfill the Laravel Database or centralize these customizations somewhere.

luizbgomide:
I tried those changes on 3.3.0.3 with MySQL 8.0.25, it works and the search got REALLY FAST, orders of magnitude better. But I also got a minor issue, the number of results per page is not being followed, some searches shows only a few itens per page (even though there are more) instead of the set number (10, 25, etc).

Even the starting page showing all users display this behavior of smaller number of record counts.

BTW, I'm also still affected by duplicated user settings with NULL values (I still following up on the subject to try to decide how to better handle this issue), and since I've seem those causing issues with reviewer assignment searches, maybe this could be the culprit here...

jonasraoni:
Thanks for testing @luizbgomide :)

I just checked some installations where I've applied this fix and I couldn't simulate this issue... Also, the updates I did shouldn't affect the initial page (which is skipping the filters that I updated).

About the duplicated records, this is the fix I wrote for MySQL (it's a bit more verbose than it should because I had to support an older MySQL version), if you remove the first line, it should also work for PostgreSQL:

```sql
SET optimizer_switch = 'derived_merge=off';

DELETE s
FROM user_settings s
INNER JOIN (
	SELECT best_duplicated.*
    FROM (
		SELECT best.*
		-- Find the duplicated entries
		FROM (
			SELECT s.setting_name, s.user_id, s.locale
			FROM user_settings s
			GROUP BY
				s.setting_name, s.user_id, s.locale
			HAVING COUNT(0) > 1
		) AS duplicated
		-- Find the best matching records for each entry
		INNER JOIN user_settings best
			ON best.setting_name = duplicated.setting_name
			AND best.user_id = duplicated.user_id
			AND best.locale = duplicated.locale
			AND CONCAT(COALESCE(best.assoc_id, ''), '-', COALESCE(best.assoc_type, '')) = (
				SELECT CONCAT(COALESCE(s.assoc_id, ''), '-', COALESCE(s.assoc_type, ''))
				FROM user_settings s
				WHERE s.setting_name = duplicated.setting_name
				AND s.user_id = duplicated.user_id
				AND s.locale = duplicated.locale
				ORDER BY
					CASE s.assoc_id
						WHEN 0 THEN 0
						WHEN NULL THEN 1
						ELSE 2
					END,
					CASE s.assoc_type
						WHEN 0 THEN 0
						WHEN NULL THEN 1
						ELSE 2
					END, s.assoc_id, s.assoc_type		
				LIMIT 1
			)
	) best_duplicated
) best_duplicated
	-- Record matches the key fields, which means it's part of the duplicated set
	ON s.setting_name = best_duplicated.setting_name
	AND s.user_id = best_duplicated.user_id
	AND s.locale = best_duplicated.locale
	-- But it's not the best, so it should be removed
	AND (
		COALESCE(s.assoc_id, '') <> COALESCE(best_duplicated.assoc_id, '')
		OR COALESCE(s.assoc_type, '') <> COALESCE(best_duplicated.assoc_type, '')
	);

-- These two settings are currently being saved with assoc_type = 0 and assoc_id = 0, so it's better to update now before they get duplicated
UPDATE user_settings SET assoc_type = 0 WHERE assoc_type = 256 AND setting_name IN ('apiKeyEnabled', 'apiKey');

-- Update the older entries, which were not duplicated
UPDATE user_settings SET assoc_type = COALESCE(assoc_type, 0), assoc_id = COALESCE(assoc_id, 0) WHERE assoc_type IS NULL OR assoc_id IS NULL;
```

jonasraoni:
@asmecher Given the presence of duplicated `user_settings`, I've updated the PR to handle them, so I think this should be ok to get merged into 3.3.0.

luizbgomide:
Tried the query and I think something odd happened, the `filterSection` and `filterEditor` also coalesced to a single setting, I don't think that is supposed to happen...

jonasraoni:
I've seen these two settings in an old installation, which has been migrated to OJS 3.3.0.

Looks like they were used by OJS 2.X, the [interface was removed in 2014](https://github.com/pkp/ojs/commit/39b8086a61243ecd4c84730f3c20ba93d168fd3d), so I think you can safely remove these settings.

luizbgomide:
Just applied the SQL to fix the user settings and the load time for the user list of a journal get from 3.6 minutes to 18 SECONDS!!! And the search time when looking at all journals went from almost 7 minutes to 1.4 minutes.

After that, improved user search query, the load time drop further to 5.6 seconds, and a search in all journals dropped to 16 seconds!!!

That is fantastic!!! The SQL DELETE/UPDATE also fixed the issue I had with the search showing fewer records than selected in "items per page".

luizbgomide:
> Can you please tell me what SQL did you apply? I have the same issue where searching users "include users with no roles in the journal" it goes very slow 7 minutes approximately.

The one provided by @jonasraoni above: https://github.com/pkp/pkp-lib/pull/7414#issuecomment-961857377

Issue #6112, opened 2020-07-23
[OJS] 3.2.1 - Slow or not loading search in "Add Reviewer" form


KRONWALLED1134:
Hi

We recently upgraded our OJS installation to 3.2.1.
Everything works fine except for the Add reviewer form.

The users reported that it loads performs searches very slow or not at all.

I checked the database then and saw very long running queries (about 14-15 minutes) regarding the searches in the Add Reviewer form.

Example (Search for 'he' in the Add reviewer form)
`SHOW FULL PROCESSLIST;` 
```
+-----+------+-----------+------+---------+------+----------+-----------------------+| Id  | User | Host      | db   | Command | Time | State    | Info                  |
+-----+------+-----------+------+---------+------+----------+-----------------------+
[...]

PR #7435, opened 2021-11-06
pkp/pkp-lib#6675 Improved query performance and updated auto complete to abort previous requests.

jonasraoni:
I've noticed that we have different types of search (e.g. might return different results for the same query) for the same type of data, it would be good to normalize, but I'll leave it for another task.


classes/security/UserGroupDAO.inc.php
```diff
@@ -430,30 +430,42 @@ function getUsersByContextId($contextId, $searchType = null, $search = null, $se
...
-					AND ug.user_group_id = uug.user_group_id
-					AND ug.role_id = ?' .
-				($contextId ? ' AND ug.context_id = ?' : '') .
-				')' .
-			(isset($search) ? ' AND (usgs.setting_value LIKE ? OR usfs.setting_value LIKE ? OR u.email LIKE ? OR u.username LIKE ?)' : ''),
-			$params,
-			$rangeInfo
-		);
+		$result = $this->retrieveRange($sql, $params, $rangeInfo);
 		return new DAOResultFactory($result, $this->userDao, '_returnUserFromRowWithData');
```
asmecher, line 469: This will discard `$rangeInfo`; is that intentional?

jonasraoni: @asmecher I didn't inspect what's happening here, I just updated the query =]
I'll check! Perhaps I can also get rid of this UserGroupDAO call and use the User service/repo...

jonasraoni: Self-reminder: move the `getUsersNotInRole()` (it's currently being used only by this auto-complete) into the user collector once the https://github.com/pkp/pkp-lib/issues/6112 gets merged.

PR #7438, opened 2021-11-06
pkp/pkp-lib#6112 Optimized user query builder

jonasraoni:


classes/services/queryBuilders/PKPUserQueryBuilder.inc.php
```diff
@@ -441,267 +443,227 @@ public function getIds() {
...
-								$q->where('us.setting_name', 'biography');
-								$q->where(Capsule::raw('lower(us.setting_value)'), 'LIKE', "%{$word}%");
-							})
-							->orWhere(function($q) use ($word) {
-								$q->where('us.setting_name', 'orcid');
-								$q->where(Capsule::raw('lower(us.setting_value)'), 'LIKE', "%{$word}%");
-							})
-							->orWhere(Capsule::raw('lower(cves.setting_value)'), 'LIKE', "%{$word}%");
+							->orWhereExists(function (Builder $query) use ($word): void {
+								$query->from('user_interests', 'ui')
```
jonasraoni, line 552: The `/main` branch has removed search over this table. Should I remove here as well? I just left to avoid changing things too much.

classes/services/queryBuilders/PKPUserQueryBuilder.inc.php
```diff
@@ -441,267 +443,227 @@ public function getIds() {
...
+					// Days since last review assignment
+					->when(!empty($this->daysSinceLastAssignment), function (Builder $query) {
+						$daysSinceLastAssignment = is_array($this->daysSinceLastAssignment) ? $this->daysSinceLastAssignment : [$this->daysSinceLastAssignment];
+						$dbTimeMin = (string) Carbon::now()->subDays((int) reset($daysSinceLastAssignment))->toDateString();
+						$query->where('ra_stats.last_assigned', '<=', $dbTimeMin);
+						if (count($daysSinceLastAssignment) > 1) {
+							// Subtract an extra day so that our outer bound rounds "up". This accounts for the UI rounding "down" in the string "X days ago".
+							$dbTimeMax = (string) Carbon::now()->subDays((int) end($daysSinceLastAssignment) + 1)->toDateString();
+							$query->where('ra_stats.last_assigned', '>=', $dbTimeMax);
+						}
```
jonasraoni, line 622: This whole snippet sounded strange for me when comparing against the `/main` version, so I kept the previous behavior here 🤔

If you don't know which one is right, I'll find where it's used and play with the interface to check whether it's a bug or not.

- Different filter
```php
$query->where('ra_stats.last_assigned', '>', $dbTimeMin); // main
$query->where('ra_stats.last_assigned', '<=', $dbTimeMin);
```

- Main `subs` days, while this branch `adds`
```php
$dbTimeMax = (string) Carbon::now()->subDays((int) end($daysSinceLastAssignment) + 1)->toDateString(); // main
$dbTimeMax = (string) Carbon::now()->addDays((int) end($daysSinceLastAssignment) + 1)->toDateString();
```

Another difference on the filter
```php
$query->where('ra_stats.last_assigned', '<', $dbTimeMax); //main
$query->where('ra_stats.last_assigned', '>=', $dbTimeMax);
```

classes/services/queryBuilders/PKPUserQueryBuilder.inc.php
asmecher: Actually that was an oversight in `main` -- could you re-add it there too?

classes/services/queryBuilders/PKPUserQueryBuilder.inc.php
asmecher: Please run a few tests on it to see if it's behaving right, thanks!

jonasraoni: Ok, I'll take a look. Just wanted to get a fast answer :)

jonasraoni: FYI I concluded the `/main` was incorrect.

PR #7456, opened 2021-11-14
Feature/main/i6112 Improved add reviewer

jonasraoni:


api/v1/users/PKPUserHandler.inc.php
```diff
@@ -125,9 +125,11 @@ public function getMany($slimRequest, $response, $args)
             ->searchPhrase($params['searchPhrase'] ?? null)
             ->orderBy($orderBy, $orderDirection, [AppLocale::getLocale(), $request->getSite()->getPrimaryLocale()])
             ->limit($params['count'] ?? null)
-            ->offset($params['offset'] ?? null)
-            ->filterByStatus($params['status'] ?? $collector::STATUS_ALL);
-
+            ->offset($params['offset'] ?? null);
```
jonasraoni: Just updated this to create a discussion about the mapping (there are other similar cases in this file).
Given the `active` is the same as the `Collector::STATUS_ACTIVE`, I think we can just let it pass through and validate it in a single place.

classes/components/forms/context/PKPNotifyUsersForm.inc.php
```diff
@@ -38,11 +39,14 @@ class PKPNotifyUsersForm extends FormComponent
...
-     * @param DAOResultFactory $userGroups Allowed user groups
      */
-    public function __construct($action, $context, $userGroups)
+    public function __construct($action, $context)
     {
         $this->action = $action;
+        /** @var \PKP\security\UserGroupDAO */
+        $userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+        $userGroups = $userGroupDao->getByContextId($context->getId());
+        $userCountByGroupId = $userGroupDao->getUserCountByContextId($context->getId());
```
jonasraoni, line 49: Just wrote this to avoid issuing database requests in a loop below.

pages/management/ManagementHandler.inc.php
```diff
@@ -401,25 +405,25 @@ public function access($args, $request)
         $apiUrl = $dispatcher->url($request, PKPApplication::ROUTE_API, $context->getPath(), 'contexts/' . $context->getId());
         $notifyUrl = $dispatcher->url($request, PKPApplication::ROUTE_API, $context->getPath(), '_email');
         $progressUrl = $dispatcher->url($request, PKPApplication::ROUTE_API, $context->getPath(), '_email/{queueId}');
-        $userGroups = DAORegistry::getDAO('UserGroupDAO')->getByContextId($context->getId());
 
         $userAccessForm = new \APP\components\forms\context\UserAccessForm($apiUrl, $context);
-        $notifyUsersForm = new \PKP\components\forms\context\PKPNotifyUsersForm($notifyUrl, $context, $userGroups);
+        $enableBulkEmails = in_array($context->getId(), (array) $request->getSite()->getData('enableBulkEmails'));
+        $notifyUsersForm = $enableBulkEmails ? new PKPNotifyUsersForm($notifyUrl, $context) : null;
```
jonasraoni, line 411: In Smarty there's an "{if}" checking the `$enableBulkEmails`, if it's false, the `PKPNotifyUsersForm` isn't used, so I also moved the `$userGroups` inside it, to avoid issuing a query.

jonasraoni: ps: I didn't add this improvement to the stable-3_3_0, I think I'll do it (in case these forms exist there).

classes/user/Collector.inc.php
```diff
@@ -13,18 +13,24 @@
...
-use PKP\db\DAORegistry;
+use PKP\core\PKPString;
 use PKP\identity\Identity;
 use PKP\plugins\HookRegistry;
 
 class Collector implements CollectorInterface
 {
     public const ORDERBY_ID = 'id';
     public const ORDERBY_GIVENNAME = 'givenName';
-    public const ORDERBY_FAMILYNAME = 'famlyName';
```
asmecher, line 27: Whoops, good catch!

classes/user/Collector.inc.php
```diff
@@ -394,183 +408,159 @@ public function getQueryBuilder(): Builder
...
+                            return $bindings;
+                        }, []);
+                        $bindinds[] = ++$lastIndex;
+                        
+                        $query->from('user_settings', 'us')
+                            ->whereColumn('us.user_id', '=', 'u.user_id')
+                            ->where('us.setting_name', '=', $settingMap[$this->orderBy])
+                            ->whereIn('us.locale', $locales)
+                            ->orderByRaw("COALESCE(us.setting_value, '') = ''") // Prefers non-null/non-empty values
+                            ->orderByRaw('CASE us.locale ' . str_repeat('WHEN ? THEN ?', count($locales)) . ' ELSE ? END', [$bindinds]) // Then follow the locale order
```
asmecher, line 551: Should `$bindinds` be supplied as wrapped in an array? It looks to me like it should be passed in as-is. But I'm finding this segment of code a little hard to read, so I could be wrong.

classes/security/UserGroupDAO.inc.php
```diff
@@ -458,6 +459,33 @@ public function getByContextId($contextId = null, $dbResultRange = null)
...
+            ->where('u.disabled', '=', 0)
+            ->groupBy('ug.user_group_id')
+            ->select('ug.user_group_id')
+            ->selectRaw('COUNT(0) AS count');
+
+        $sql = $query->toSql();
+        $params = $query->getBindings();
+
+        $result = [];
+        foreach ($this->retrieveRange($sql, $params) as $row) {
```
asmecher, line 483: Is it necessary to use `retrieveRange`? It looks to me like you can just use a `->select(...)->get()` and get rid of `$sql`, `$params`, etc.

api/v1/users/PKPUserHandler.inc.php
asmecher: I haven't come to a firm idea on this, but you're right that the use of `$params` and the collector functions/constants currently duplicate some things. I would lean towards letting the collectors implement the model, then draw on that in the handler. Personally I don't like the use of arrays (as we're using `$params` here, and previously used them in `EntityDAO` subclasses) -- their contents can't be typed/controlled it's too easy to make a mistake.

classes/user/Collector.inc.php
jonasraoni: It works this way as well (there's a `array_merge_recursive` in the internals), so I didn't notice =]

classes/security/UserGroupDAO.inc.php
jonasraoni: Copy/paste failure 😑

api/v1/users/PKPUserHandler.inc.php
jonasraoni: I'm also against passing static information through dictionaries... In a next iteration, I think we should get rid of them.
But for now, I'll just revert this change =]

classes/user/Collector.inc.php
```diff
@@ -33,8 +39,7 @@ class Collector implements CollectorInterface
     public const STATUS_DISABLED = 'disabled';
     public const STATUS_ALL = null;
 
-    /** @var DAO */
-    public $dao;
+    public DAO $dao;
```
jonasraoni, line 42: I've just noticed that we're using PHP 7.4 on `/main`, so I'll make more updates here.

jonasraoni:
@asmecher I've addressed the comments that I had left and broke the query into pieces (if you prefer to leave it the way it was, I can revert).

classes/user/Collector.inc.php
```diff
@@ -330,254 +342,260 @@ public function offset(?int $offset): self
...
+     */
+    protected function buildOrderBy(Builder $query): self
+    {
+        $orderByFields = [self::ORDERBY_ID => 'u.user_id'];
+        if ($orderByField = $orderByFields[$this->orderBy] ?? null) {
+            $query->orderBy($orderByField, $this->orderDirection);
+            return $this;
+        }
+
+        $settings = [self::ORDERBY_GIVENNAME => Identity::IDENTITY_SETTING_GIVENNAME, self::ORDERBY_FAMILYNAME => Identity::IDENTITY_SETTING_FAMILYNAME];
```
jonasraoni, line 577: When sorting by given name, isn't it interesting to add the last name to the sorting (and vice-versa)?

asmecher: Sure, if it's easy to add.

jonasraoni: @asmecher Ok, I've updated.

classes/security/UserGroupDAO.inc.php
```diff
@@ -458,6 +460,24 @@ public function getByContextId($contextId = null, $dbResultRange = null)
...
 
+    /**
+     * Retrieves a keyed Collection (key = user_group_id, value = count) with the amount of active users for each user group
+     */
+    public function getUserCountByContextId(int $contextId = null): Collection
+    {
+        return DB::table('user_groups', 'ug')
+            ->join('user_user_groups AS uug', 'uug.user_group_id', '=', 'ug.user_group_id')
+            ->join('users AS u', 'u.user_id', '=', 'uug.user_id')
+            ->when($contextId, function (Builder $query) use ($contextId): void {
```
asmecher, line 471: I'd like us to use strict comparisons on filters, rather than treating falsy values equivalently. The DAOs are quite sloppy about this but the Collectors are much cleaner and there's no harm in being stricter when introducing/reworking code in DAOs too. So:
```PHP
->when($contextId !== null, function(Builder $query) use ($contextId) {
```

classes/security/UserGroupDAO.inc.php
```diff
@@ -458,6 +460,24 @@ public function getByContextId($contextId = null, $dbResultRange = null)
         return new DAOResultFactory($result, $this, '_returnFromRow', [], $sql, $params, $dbResultRange);
     }
 
+    /**
+     * Retrieves a keyed Collection (key = user_group_id, value = count) with the amount of active users for each user group
+     */
+    public function getUserCountByContextId(int $contextId = null): Collection
```
asmecher, line 466: `int` and `null` is a contradiction.

asmecher (commented):
Just a couple clean-ups.

classes/security/UserGroupDAO.inc.php
jonasraoni: Ok!

Issue #7167, opened 2021-07-02
Null values in assoc columns in user_settings prevent correct data reading after update


Vitaliy-1:
**Describe the bug**
The bug described on the forum: https://forum.pkp.sfu.ca/t/null-values-in-user-settings-safe-to-remove/68272
Old installations may have NULL values in `assoc_type` and `assoc_id` columns in `user_settings` table. This leads to duplicated rows when this data is updated, e.g. through the user profile form:
```
user_id	locale	setting_name	assoc_type	assoc_id	setting_value	setting_type	
4	en_US	givenName	NULL            NULL            Doe	        string	
4	en_US	givenName	0	        0	        NewDoe	        string	
```
According to the report, this leads to problems in reading user data, the one that I can reproduce appears in Add Reviewer form: searching for a reviewer gives duplicated entries of this particular user in example:
![user-settings](https://user-images.githubusercontent.com/20665587/124279709-7d3dc980-db50-11eb-8d8f-b95c2ad0dd8e.png)

Also, during data retrieval, old setting value, e.g. `Doe` may be displayed instead of `NewDoe`. This one I wasn't able to reproduce 

**To Reproduce**
At glance, this problem affects old installs and I'm unable to reproduce it on a clean instance. I can confirm that my old test instance upgraded from 3.0 is affected.

**What application are you using?**
 OJS 3.3

**Additional information**
I think to solve the issue the best would be to add an upgrade script that removes duplicates (those with `0` are newer and should remain intact while rows with `NULL` should be removed) or replaces `NULL` values with the default `0`s if the row isn't duplicated. 
Associated issue: https://github.com/pkp/pkp-lib/issues/7146

jonasraoni:
Just adding notes from my other issue:

To identify if you have duplicated `user_settings`
```sql
SELECT s.user_id, s.setting_name, s.locale
FROM user_settings s
GROUP BY user_id, setting_name, locale
HAVING COUNT(0) > 1
```

The `apiKey` and `apiKeyEnabled` are also duplicated, but they follow a slightly different pattern, the older records have `assoc_type` = 256 and `assoc_id` = `null`

Perhaps the problem was introduced here: https://github.com/pkp/pkp-lib/pull/6347/files#diff-f4951b8d90b205bf60a1111a761b62eb75f598c0c536738da01d027561a33f1cR74-R75

- Yeah, an upgrade script will be needed. And as far as I could see from my observations, there's no risk to discard the nulled data once it's duplicated.
[...]

PR #7178, opened 2021-07-09
pkp/pkp-lib#6724 Upgrade optimizations: submission locales, submissio…

ewhanson:
…n event logs, delete old revisions

PR #7464, opened 2021-11-20
pkp/pkp-lib#7167 Fix duplicated user settings and remove unused fields

jonasraoni:


classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
```diff
@@ -94,10 +94,6 @@ public function up(): void
             });
         }
 
-        // Use nulls instead of 0 for assoc_type/id in user_settings
-        DB::table('user_settings')->where('assoc_type', 0)->update(['assoc_type' => null]);
-        DB::table('user_settings')->where('assoc_id', 0)->update(['assoc_id' => null]);
```
jonasraoni, line 99: I've removed this as it's probably the origin of the duplicated records.

classes/migration/install/CommonMigration.inc.php
```diff
@@ -95,14 +95,13 @@ public function up(): void
 
         // Locale-specific user data
         Schema::create('user_settings', function (Blueprint $table) {
+            $table->bigInteger('user_settings_id')->autoIncrement();
```
jonasraoni, line 98: Added for the sake of normalization.

classes/install/Installer.inc.php
```diff
@@ -466,7 +466,11 @@ public function executeAction($action)
                         try {
                             $this->log(sprintf('revert migration: %s', get_class($previousMigration)));
                             $previousMigration->down();
-                        } catch (\PKP\install\DowngradeNotSupportedException $e) {
+                        } catch (DowngradeNotSupportedException $e) {
+                            $this->log(sprintf('downgrade for "%s" unsupported: %s', get_class($previousMigration), $e->getMessage()));
+                            break;
+                        } catch (Exception $e) {
+                            $this->log(sprintf('error while downgrading "%s": %s', get_class($previousMigration), Config::getVar('debug', 'show_stacktrace') ? (string) $e : $e->getMessage()));
```
jonasraoni, line 473: Generic exceptions in the migration `down()` were leaking.

classes/plugins/PluginRegistry.inc.php
```diff
@@ -258,9 +249,12 @@ public static function getCategories()
      */
     public static function loadAllPlugins($enabledOnly = false)
     {
-        // Retrieve and register categories (order is significant).
-        foreach (self::getCategories() as $category) {
-            self::loadCategory($category, $enabledOnly);
+        static $isLoaded;
```
jonasraoni, line 252: I doubt this is going to be called many times, but in case it does, it should run faster.

classes/plugins/PluginRegistry.inc.php
```diff
@@ -89,21 +89,12 @@ public static function register($category, $plugin, $path, $mainContextId = null
         $pluginName = $plugin->getName();
         $plugins = & self::getPlugins();
 
-        // Allow the plugin to register.
-        if (!$plugin->register($category, $path, $mainContextId)) {
+        // If the plugin is already loaded or failed/refused to register
+        if (isset($plugins[$category][$pluginName]) || !$plugin->register($category, $path, $mainContextId)) {
```
jonasraoni, line 93: I arrived here after the `StaticPagesPlugin` tried to run its migration twice. I checked the hooks and there were many duplicated entries registered.

Another potencial side-fix would be to block the `HookRegistry` from registering the same hook+callback, but as a closure can be passed, I decided to not bother with it for now.

jonasraoni:
@asmecher I've tried to decrease the running time of Travis with some settings, but didn't change that much (I just checked OPS and the total time decreased from 1h45 to 1h39).

Also, what about removing the stale branches in pkp-lib, ojs, etc?
image

classes/migration/install/CommonMigration.inc.php
asmecher: There are a lot of `_settings` tables and none of them currently have IDs; if we're going to start introducing them I'd like to consider it in its own issue, and if it seems promising, make the change all at once.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
asmecher: @jonasraoni, I think we should standardize on `null` instead of `0` *everywhere* we're currently using `0` as a placeholder ID. As we start declaring `FOREIGN KEY` relationships formally, we won't be able to get away with this looseness any more. (That change will come to `assoc_type` / `assoc_id` relationships last, because DBMSs won't understand them at the DDL level, but I think it's important to be consistent anyway as we roll out code that's pickier about `0` vs. `null` in PHP.)

FWIW, this change was added [here](https://github.com/pkp/pkp-lib/pull/6347/files) by @Vitaliy-1.

classes/migration/install/CommonMigration.inc.php
jonasraoni: Ok! I've opened an issue about normalizing the database, so I'll create a PR there.

classes/migration/upgrade/PKPv3_3_0UpgradeMigration.inc.php
jonasraoni: That's right, but the problem here is that the table was updated, together with all its records, but the code stills using zeroes, then whenever a user updates its data, things get duplicated (older record with null + new record with 0).

But in fact, I just removed this to speed up the migration, given that these fields will be removed by my PR later on.

jonasraoni: Sooo, I think these two lines should be also removed from the `stable-3_3_0` branch. If you agree, I can create an issue + PR for it.

asmecher: That makes sense -- please go ahead!

classes/migration/upgrade/v3_4_0/I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields.inc.php
```diff
@@ -0,0 +1,122 @@
...
+        if (!Schema::hasColumn('user_settings', 'assoc_id') && !Schema::hasColumn('user_settings', 'assoc_type')) {
+            return;
+        }
+
+        $isMysql = DB::connection() instanceof MySqlConnection;
+        $deleteFrom = 'DELETE FROM user_settings s USING';
+        $joinFilter = 'WHERE';
+        if ($isMysql) {
+            // The optimizer might cut-off the sub-queries/closures below and make MySQL fail to delete from the table which is selecting from
+            DB::unprepared("SET optimizer_switch = 'derived_merge=off'");
```
asmecher, line 40: @jonasraoni, can you document this a little more in the comment? e.g. is there a link to information about it, and conditions to replicate? We may need to consider this problem elsewhere (e.g. https://github.com/pkp/pkp-lib/pull/7178/files#r755415915) and I'd like to make sure we're treating joins with updates/deletes systematically.

jonasraoni: Ah, this problem is old, it happens in MySQL when you try to `DELETE`/`UPDATE` and there's a sub-query using the same target table.
It was possible to fix it by enclosing the sub-query with another `(SELECT * FROM (original subquery) AS fix`, but it doesn't work in newer versions of MySQL, which detect the useless sub-query and removes it :D

Then, disabling this optimizer setting, avoids this behavior...

In case you'd like to see something less hacky, but slower, I could create a temporary table to store the IDs which should be removed.

jonasraoni: Now that I'm re-watching the query, maybe it's possible to solve the problem without a sub-query, I'll check 🤔

jonasraoni: Yeah, it worked, I compared the outputs and got the same result, I'll update the PR tomorrow with this:
```sql
DELETE s
FROM user_settings s
INNER JOIN user_settings duplicated
	ON s.setting_name = duplicated.setting_name
    AND s.user_id = duplicated.user_id
    AND s.locale = duplicated.locale
    AND (
		COALESCE(s.assoc_type, -9999) <> COALESCE(duplicated.assoc_type, -9999)
        OR COALESCE(s.assoc_id, -9999) <> COALESCE(duplicated.assoc_id, -9999)
	)
LEFT JOIN user_settings best
	ON best.setting_name = duplicated.setting_name
	AND best.user_id = duplicated.user_id
	AND best.locale = duplicated.locale
    AND COALESCE(best.assoc_id, 9999) < COALESCE(s.assoc_id, 9999)
    AND COALESCE(best.assoc_type, 9999) < COALESCE(s.assoc_id, 9999)
WHERE best.user_id IS NOT NULL
```

asmecher: I worry that this may be one of those things that pass Travis and local testing but result in errors on other versions of MySQL/PostgreSQL that we're not testing -- is the original issue documented somewhere?

jonasraoni: Yeah, it's documented here:
https://dev.mysql.com/doc/refman/8.0/en/delete.html#idm46019841632784:~:text=You%20cannot%20delete%20from%20a%20table%20and%20select%20from%20the%20same%20table%20in%20a%20subquery.

jonasraoni: Ok! I've updated the query to avoid using a sub-query.

classes/migration/upgrade/v3_4_0/I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields.inc.php
```diff
@@ -0,0 +1,114 @@
...
+                    -- Ensures a better record was found (if not found, it means the current duplicated record is the best and shouldn't be removed)
+                    AND best.user_id IS NOT NULL"
+            );
+        }
+
+        // Here we should be free of duplicates, so it's safe to remove the columns without creating duplicated entries.
+        Schema::table(
+            'user_settings',
+            function (Blueprint $table): void {
+                // Drop deprecated fields (this will implicitly drop the unique key "user_settings_pkey")
```
asmecher, line 99: From https://dev.mysql.com/doc/refman/8.0/en/alter-table.html:
> If columns are dropped from a table, the columns are also removed from any index of which they are a part. If all columns that make up an index are dropped, the index is dropped as well. If you use CHANGE or MODIFY to shorten a column for which an index exists on the column, and the resulting column length is less than the index length, MySQL shortens the index automatically. 

...so I think this will drop those columns from the `user_settings_pkey` index, but not delete the whole index.

As for PostgreSQL (https://www.postgresql.org/docs/7.4/sql-altertable.html):
> Indexes and table constraints involving the column will be automatically dropped as well.

This is awfully brief but seems to suggest that the index will be fully dropped.

jonasraoni: That's true! I was deceived by my MySQL editor (it's adding an extra clause to remove the index) 😁

jonasraoni:
@asmecher I've updated the PR to address the index, the tests are passing yet, so feel free to take a look again 😁

Issue #7506, opened 2021-11-30
Unable to Upload/Select Files from Copyediting stage (fix ineffective)


p6steve:
This is a re-occurence of https://github.com/pkp/pkp-lib/issues/4754 since that fix is not effective in our cases.

The related PKP Community post is https://forum.pkp.sfu.ca/t/unable-to-upload-select-files-from-copyediting-stage-fix-ineffective/70932

If, in a submission, a file is attached to a discussion then in Copyediting it is not possible to use “Upload/Select Files” from Copyediting stage.

It happens that the window opens but does not load the part for uploading or selecting files.
![](https://user-images.githubusercontent.com/27674815/57630400-36726b80-75a6-11e9-90e8-e676ef1c60d7.png)
)
 OJS version 3.3.0.6

p6steve:
I have proposed an extended fix in https://github.com/pkp/pkp-lib/pull/7507

PR #7507, opened 2021-11-30
pkp/pkp-lib#7506 Unable to Upload/Select Files from Copyediting stage (fix ineffective)

p6steve:
Unable to Upload/Select Files from Copyediting stage (fix ineffective) - this change extends the previous fix [https://github.com/pkp/ojs/pull/2396] to also guard the $note->GetAssocId() method call from the case where $note == null - ie. when there is an orphan note in the db

PS. p6steve is my GitHub ID - Steve Roe - steve@henleycloudconsulting.co.uk is my ID on PKP Community


asmecher:
Thanks, @p6steve -- if I could request one change: you'll see an `assert` statement that checks the note and its `getAssocType`; could you turn that into an `if` including both of its checks rather than leaving it in place and then re-testing half of its conditions? Thanks!

p6steve:
Ah yes - good point - this now done...

asmecher:
Merged -- thanks, @p6steve!

PR #7508, opened 2021-11-30
pkp/pkp-lib#5998 Implementing caching and timeout on plugins.xml request

henriqueramos:


classes/plugins/PluginGalleryDAO.inc.php
```diff
@@ -54,23 +67,79 @@ public function getNewestCompatible($application, $category = null, $search = nu
...
+            Application::CONTEXT_SITE,
+            function (FileCache $cache) {
+                $cache->setEntireCache($this->getExternalDocument());
+            }
+        );
+
+        $cacheTime = $cache->getCacheTime();
+
+        // Checking if the cache is older than 1 day, or its null
+        if ($cacheTime === null || (time() - $cacheTime > 24 * 3600)) {
```
asmecher, line 125: `24 * 3600` could probably stand to be a class constant too.

classes/plugins/PluginGalleryDAO.inc.php
```diff
@@ -54,23 +67,79 @@ public function getNewestCompatible($application, $category = null, $search = nu
...
+            // This cache is out of date; flush it.
+            $cache->setEntireCache($this->getExternalDocument());
```
asmecher, line 127: If the PKP site goes down for a little while, rather than relying on the cached version, this will wipe out the cache, I think.

henriqueramos: Yes. Perhaps I should verify if the content wasn't an Exception before flush the cache.

classes/plugins/PluginGalleryDAO.inc.php
henriqueramos: Indeed.

classes/plugins/PluginGalleryDAO.inc.php
```diff
@@ -54,23 +72,84 @@ public function getNewestCompatible($application, $category = null, $search = nu
...
+
+        $cacheTime = $cache->getCacheTime();
+
+        // Checking if the cache is older than 1 day, or its null
+        if ($cacheTime === null || (time() - $cacheTime > self::TTL_CACHE_SECONDS)) {
+            // This cache is out of date; so, lets request a new version.
+            $response = $this->getExternalDocument();
+
+            // The plugins.xml request wasnt empty, so lets replace it
+            if ($response !== '') {
```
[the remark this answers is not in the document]
asmecher: Won't the `getRemoteDocument` function return `null` (not an empty string) when the request fails?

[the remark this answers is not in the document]
henriqueramos: In case of request failure, the `catch` statement inside it will log data with `error_log` and return an empty string.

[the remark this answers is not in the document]
asmecher: I'm talking about this `null`: https://github.com/pkp/pkp-lib/pull/7508/files/d73a1eaba50ffc4035dc576f3b739bd709ac65f2#diff-72f7228e4162f97f27cd0f7c8f85b06af2d2a1c6298e5c268830ec45cb47f8a1R107

cypress/tests/integration/Plugins.spec.js
```diff
@@ -0,0 +1,26 @@
...
+        cy.server();
+        cy.route('GET', '**/grid/plugins/plugin-gallery-grid/fetch-grid*').as('getPluginList');
+        cy.login('dbarnes', null, 'publicknowledge');
+
+        cy.get('a:contains("Website")').click();
+        cy.waitJQuery();
+        cy.get('button#plugins-button').click();
+        cy.get('button#pluginGallery-button').click();
+        cy.wait('@getPluginList', {
+            responseTimeout: 30000
```
henriqueramos, line 21: This `responseTimeout` acts a trigger to avoid a too long response time on `plugin-gallery-grid/fetch-grid` response.

cypress/tests/integration/Plugins.spec.js
```diff
@@ -0,0 +1,26 @@
...
+        }).then((evt) => {
+            cy.readFile('cache/fc-loadPluginsXML-0.php');
```
henriqueramos, line 23: This guarantees the file exists.

classes/plugins/PluginGalleryDAO.inc.php
```diff
@@ -54,23 +72,84 @@ public function getNewestCompatible($application, $category = null, $search = nu
...
+
+        $cacheTime = $cache->getCacheTime();
+
+        // Checking if the cache is older than 1 day, or its null
+        if ($cacheTime === null || (time() - $cacheTime > self::TTL_CACHE_SECONDS)) {
+            // This cache is out of date; so, lets request a new version.
+            $response = $this->getExternalDocument();
+
+            // The plugins.xml request wasnt empty, so lets replace it
+            if (!$response || $response !== '') {
```
asmecher, line 135: Purely syntactical, but this still looks off to me. I think it should be...
```
if ($response !== null) {
```

henriqueramos: Done.

PR #7553, opened 2021-12-17
pkp/pkp-lib#7171 Add an interface for viewing processing queue status

henriqueramos:
# Evidence on OJS

Screenshot from 2021-12-16 22-32-30


pages/management/PKPToolsHandler.inc.php
```diff
@@ -160,4 +166,101 @@ public function resetPermissions($args, $request)
...
+                'attempts' => $currentItem->attempts,
+                'created_at' => __('manager.jobs.createdAt', ['createdAt' => strftime($dateFormatShort, $currentItem->created_at)]),
+            ];
+        }
+
+        $queuedJobs = new PKPStatsJobsTable(
+            'queuedJobsTable',
+            [
+                'label' => __('manager.jobs.viewQueuedJobs'),
+                'description' => __('manager.jobs.totalCount', ['total' => '<strong>' . $totalQueuedJobs . '</strong>']),
```
asmecher, line 234: The formatting should be moved into the locale key, I think.

classes/components/PKPStatsJobsTable.inc.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function __construct(string $id, array $args = [])
+    {
+        AppLocale::requireComponents([
+            LOCALE_COMPONENT_PKP_MANAGER,
+            LOCALE_COMPONENT_APP_MANAGER
+        ]);
+
+        $this->id = $id;
+        $this->init($args);
```
asmecher, line 49: If this is the only use of `init`, why not just move that code into the constructor?

classes/components/PKPStatsJobsTable.inc.php
```diff
@@ -0,0 +1,89 @@
...
+    /** @var null|string Table description */
+    public $description = null;
+
+    /** @var null|string Table label */
+    public $label = null;
+
+    /**
+     * Constructor
+     *
+     * @param $args array Optional arguments
```
asmecher, line 39: Needs to be updated to match the number of parameters.

classes/components/PKPStatsJobsTable.inc.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->id = $id;
+        $this->init($args);
+    }
+
+    /**
+     * Initialize the handler with config parameters
+     *
+     * @param $args array Configuration params
+     */
+    public function init($args = [])
```
asmecher, line 57: It'll be nice to use named parameters to replace this kind of pattern, once we have a PHP8.0 baseline! A fluent interface would work here too, but I'm not sure it's worth the extra verbiage.

locale/en_US/manager.po
```diff
@@ -2172,3 +2172,23 @@ msgstr "Validate Email (Site)"
...
+msgid "manager.jobs.list.id"
+msgstr "ID"
+
+msgid "manager.jobs.list.displayName"
+msgstr "Job"
+
+msgid "manager.jobs.list.attempts"
+msgstr "Attempts"
+
+msgid "manager.jobs.list.created_at"
```
asmecher, line 2190: Let's stick with lowerCamelCase for locale key naming.

pages/management/PKPToolsHandler.inc.php
```diff
@@ -160,4 +166,101 @@ public function resetPermissions($args, $request)
...
+        $templateMgr = TemplateManager::getManager($request);
+
+        $queuedJobs = $this->buildQueuedJobsTable();
+
+        $totalQueuedJobs = $queuedJobs['total'];
+        $queuedData = $queuedJobs['data'];
+
+        $state = [
+            'components' => [
+                $queuedData->id => $queuedData->getConfig(),
```
asmecher, line 183: This is more @NateWr's area of expertise -- but isn't `id` a string, and `getConfig` returns an array?

pages/management/PKPToolsHandler.inc.php
```diff
@@ -160,4 +166,101 @@ public function resetPermissions($args, $request)
...
+     * Build the Queued Jobs Table
+     */
+    protected function buildQueuedJobsTable(): array
+    {
+        $dateFormatShort = Application::get()
+            ->getRequest()
+            ->getContext()
+            ->getLocalizedDateTimeFormatShort();
+
+        $jobsQuery = DB::table('jobs')
```
asmecher, line 208: The handler class shouldn't be building SQL queries; this should be in a DAO or Collector.

pages/management/PKPToolsHandler.inc.php
```diff
@@ -160,4 +166,101 @@ public function resetPermissions($args, $request)
...
+            ->whereNotNull('queue')
+            ->whereNull('reserved_at');
+
+        $totalQueuedJobs = $jobsQuery->count();
+
+        $queuedJobsItems = $jobsQuery->get()
+            ->toArray();
```
asmecher, line 215: Since you're just doing a `foreach`, the `toArray` shouldn't be necessary.

pages/management/PKPToolsHandler.inc.php
asmecher: (Likewise, the handler class shouldn't be iterating through raw database records.)

classes/components/PKPStatsJobsTable.inc.php
henriqueramos: I agree with the named parameters usage, but I don't know about PHP 8.0 adoption across our community base. I think this could be added in a near future.

classes/components/PKPStatsJobsTable.inc.php
henriqueramos: I'd like to keep those things separated, allowing us to do a better unit testing. Currently we don't have a good code coverage on unit testing, and I'm thinking in work at the next months in cover at least the main classes.

locale/en_US/manager.po
henriqueramos: Okay!

pages/management/PKPToolsHandler.inc.php
henriqueramos: There's a few other examples using this methodology.

https://github.com/pkp/pkp-lib/blob/main/pages/dashboard/DashboardHandler.inc.php#L188

pages/management/PKPToolsHandler.inc.php
henriqueramos: Unfortunately, at this point of code history, we doesn't have a Domain/DAO/Repository/Collector for the `jobs` table.

I've created a Eloquent Model for `jobs` at #4622. It was two months before the starting of Repository refactoring.

Right now I don't know if I should wait for the #4622 being merged, with his Eloquent Model stuff, or create a Repository here for the `jobs`. Both patterns are conflicting, in my view.

pages/management/PKPToolsHandler.inc.php
henriqueramos: I've added it for debugging purposes, and forgot to remove. :facepalm:

classes/core/EntityDAO.inc.php
```diff
@@ -113,22 +113,24 @@ public function fromRow(stdClass $row): DataObject
...
-                    $row->setting_name,
-                    $this->convertFromDB(
-                        $row->setting_value,
-                        $schema->properties->{$row->setting_name}->type
-                    ),
-                    empty($row->locale) ? null : $row->locale
-                );
-            }
-        });
+        if ($this->settingsTable) {
```
henriqueramos, line 116: `jobs` doesnt have a `settingsTable`, so I've added this conditional here to avoid SQL errors.

pages/management/PKPToolsHandler.inc.php
asmecher: (reminder about this comment!)

tools/jobs.php
```diff
@@ -0,0 +1,420 @@
...
+            $messages,
+            $title,
+            'fg=white;bg=red',
+            ' ',
+            true
+        );
+    }
+}
+
+class commandJobs extends CommandLineTool
```
asmecher, line 71: This class needs to be localized; I'd suggest putting the relevant locale keys in `lib/pkp/locale/en_US/admin.po` where you can't find already-existing ones.

tools/jobs.php
```diff
@@ -0,0 +1,420 @@
...
+                        'Reserved At',
+                        'Available At',
+                        'Created At'
+                    ]
+                ],
+                $data->all(),
+            );
+
+        $pagination = [
+            'pagination' => [
```
asmecher, line 264: The `0`, `1`, and `2` magic numbers correspond to last page, current page, and next page (where available), right? Could you use class constants for these?

cypress/tests/integration/Jobs.spec.js
```diff
@@ -0,0 +1,25 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ */
+
+ describe('Jobs tests', function() {
+    it('Check if Jobs page is alive and with contents', function() {
+        cy.login('dbarnes', null, 'publicknowledge');
+
+        // Add two test jobs on queue
+        cy.exec('php lib/pkp/tools/jobs.php test');
```
asmecher, line 15: What happens if there is a queue runner that handles these? I presume they'll disappear as soon as the queue runner hits them?

tools/jobs.php
```diff
@@ -0,0 +1,420 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file tools/jobs.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
```
asmecher, line 8: For all new code, copyright end date should be 2022.

cypress/tests/integration/Jobs.spec.js
henriqueramos: They're being added into a queue called `queuedTestJob`. I think the probability of this happen is extreme low.

tools/jobs.php
henriqueramos: I don't think it's possible to i18n the tools. I've tried and the result was unsuccessful.

henriqueramos: Screenshot from 2022-01-19 11-50-42

pages/management/PKPToolsHandler.inc.php
henriqueramos: He's getting the result from `id` parameter and using as the key value.

Something like this:
Screenshot from 2022-01-19 12-06-50

tools/jobs.php
henriqueramos: Thank you @NateWr!

cypress/tests/integration/Jobs.spec.js
henriqueramos: Also, the test job always throw an exception, so, he's going back to the queue every time.

asmecher: Would it be possible to assert something based on the specific jobs that were added, rather than assuming that the queue was empty beforehand (and thus the 2 entries must be the test entries)? Then make the `purge` command only clear the `queuedTestJob` queue? That would be a little more error-proof, I think.

locale/en_US/admin.po
```diff
@@ -326,3 +326,96 @@ msgstr "Error: can't execute \"{$utilPath}\".  Check your {$utilVar} setting in
...
+msgid "admin.cli.tool.jobs.purge.successful.all"
+msgstr "Deleted all jobs!"
+
+msgid "admin.cli.tool.jobs.total.jobs"
+msgstr "We have {$total} queued jobs"
+
+msgid "admin.cli.tool.jobs.option.doesnt.exists"
+msgstr "Option {$option} does not exist."
+
+msgid "admin.cli.tool.jobs.mean.this"
```
asmecher, line 417: I think these last two are test locale keys.

locale/en_US/common.po
```diff
@@ -1990,3 +1990,9 @@ msgstr "Trying to run a transformation that is not supported in your installatio
 
 msgid "filter.import.error.couldNotImportNode"
 msgstr "Could not import XML Node"
+
+msgid "navigation.tools.jobs"
```
asmecher, line 1994: These are only viewable by the manager, right? I'd suggest keeping them out of `common.po` if that's the case. The `manager.po` and `admin.po` files are more appropriate.

pages/management/PKPToolsHandler.inc.php
asmecher: I kept misreading...
```php
$queuedData->id => $queuedData->getConfig()
```
...as...
```php
$queuedData->id = $queuedData->getConfig()
```
My mistake!

tools/jobs.php
```diff
@@ -0,0 +1,427 @@
...
+                        new TableCell(
+                            __('admin.cli.tool.jobs.pagination'),
+                            [
+                                'colspan' => 3,
+                                'style' => new TableCellStyle(['align' => 'center'])
+                            ]
+                        )
+                    ],
+                    [
+                        __('admin.cli.tool.jobs.pagination.current'),
```
asmecher, line 295: I think there are implicit keys here (`self::CURRENT_PAGE` et al) -- should those be made explicit?

locale/en_US/admin.po
henriqueramos: They're used if someone doesn't pass a valid option for the command.

tools/jobs.php
henriqueramos: I don't think so, the `self::CURRENT_PAGE` is used on the array constructions, the `i18n` here it's just to display the values according to those.
Screenshot from 2022-01-20 03-13-48


And if someone want's to traverse over the results, they need to pass the `--page=` parameter.

cypress/tests/integration/Jobs.spec.js
henriqueramos: I've updated the test to handle this checking if the there's two rows with queue as 'QueuedTestJob' on the page. Also, cleaning only the `queuedTestJob` queue after tests.

locale/en_US/admin.po
henriqueramos: Screenshot from 2022-01-20 03-24-29

asmecher: Ah, I see. It's OK, but things like the plural check (`this` vs. `those`) really lead to the proliferation of a lot of locale keys (worse for translators) without any real improvement in the user's life. If you like you could use neutral language like "Did you mean one of the following?"

tools/jobs.php
asmecher: OK, thanks!

cypress/tests/integration/Jobs.spec.js
asmecher: Thanks!

Support/Interfaces/Core/Repository.php
```diff
@@ -0,0 +1,65 @@
...
+ * @brief Interface for Repository Classes
+ */
+
+namespace PKP\Support\Interfaces\Core;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\Model;
+
+interface Repository
```
asmecher, line 24: As I understand it, Illuminate doesn't normally have a Repository pattern, but it can be useful to borrow some of those conventions from e.g. Symfony (https://www.twilio.com/blog/repository-pattern-in-laravel-application). Since this class doesn't extend a base class, and since the particular conventions we use are arbitrary, I wonder whether we shouldn't make the functions here align a little more closely with our own `Repository` pattern. For example, `get` instead of `find`, `add` instead of `create`, `edit` instead of `update`, etc. Otherwise it's a whole set of synonyms to remember.

locale/en_US/admin.po
asmecher: (don't forget this one)

Domains/Jobs/Interfaces/JobRepositoryInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ * @ingroup support
+ *
+ * @brief Interface for Domain's Jobs Repository
+ */
+
+namespace PKP\Domains\Jobs\Interfaces;
+
+use PKP\Support\Interfaces\Core\Repository;
+
+interface JobRepositoryInterface extends Repository
```
asmecher, line 22: Is this interface still necessary now that it's empty?

henriqueramos: This is necessary to enable the dependency injection of repository.

https://github.com/pkp/pkp-lib/pull/7553/files#diff-fbd6403657ede20e86a0f309dcf1a064e0da1d8748e4e65c2635ccfd8bc8d1deR29

asmecher: Is there a benefit to using the `JobRepositoryInterface` rather than `PKP\Domains\Jobs\Repositories\Job::class` in registering the dependency injection?

henriqueramos: Binding interface instead concrete class increase the malleability of system. Allowing us to implement more services with less effort (e.g moving the local database jobs table to SQS or RabbitMQ). Some comments about this subject, [here](https://stackoverflow.com/a/64589363/1890634).

The only 'mea culpa' I should mentioned is I've called the `PKP\Domains\Jobs\Repositories\Job::class` directly on the `Repo.inc.php` instead the Interface. :man_facepalming:

asmecher: Facilitating testing is a good argument for interfaces rather than classes in DI, so in principle I'm sold. However, rolling this out consistently would involve creating a *lot* of empty interfaces without an existing use case (e.g. tests) to justify them. Codebase bloat and promising but incompletely applied patterns are constant challenges. So I think it's OK to go ahead with this as is, but generally I'd prefer to just stick with `Job::class` until there's a specific need for an interface.

henriqueramos: Implement a better code coverage using unit tests is a wish I have. Those changes here are just the initial implementation on this subject. ;)

PR #7574, opened 2021-12-25
pkp/pkp-lib#7568 Added option to disable the mail delivery

jonasraoni:


classes/submission/action/EditorAction.inc.php
```diff
@@ -198,7 +198,7 @@ public function addReviewer($request, $submission, $reviewerId, &$reviewRound, $
                         PKPNotification::NOTIFICATION_TYPE_ERROR,
                         ['contents' => __('email.compose.error')]
                     );
-                    trigger_error($e->getMessage(), E_USER_WARNING);
+                    trigger_error('Failed to send email: ' . $e->getMessage(), E_USER_WARNING);
```
jonasraoni, line 201: This is just to address a strange/unhelpful error message in the error_log while running the tests.

classes/task/FileLoader.inc.php
```diff
@@ -392,27 +392,6 @@ private function _stageFile()
     {
         $this->moveFile($this->_processingPath, $this->_stagePath, $this->_claimedFilename);
     }
-
-    /**
-     * Send the passed message to the administrator by email.
-     *
-     * @param $message string
-     */
-    private function _notify($message, $messageType)
```
jonasraoni, line 401: This method isn't being used, but there's another place using the old Mail class, which needs to be wiped off.

tools/travis/prepare-tests.sh
```diff
@@ -49,6 +49,9 @@ if [[ "$DISABLE_PATH_INFO" == "1" ]]; then
 	sed -i -e "s/disable_path_info = Off/disable_path_info = On/" config.inc.php
 fi
 
+# Replaces the mailer by a mocked instance
+sed -i -e "s/default = sendmail/default = disabled/" config.inc.php
```
jonasraoni, line 53: Update the tests to disable the mail delivery, perhaps it will bring some performance improvement, but my motivation was to avoid error messages.

As I wrote in the issue, setting this to "log", will pollute the error_log with emails.

classes/core/PKPContainer.inc.php
```diff
@@ -207,7 +208,11 @@ protected function loadConfiguration()
             'channel' => 'errorlog',
         ];
 
-        $items['mail']['default'] = static::getDefaultMailer();
+        $items['mail']['default'] = static::getDefaultMailer() === 'disabled' ? 'sendmail' : static::getDefaultMailer();
```
henriqueramos, line 211: What do you think in move this logic to the `getDefaultMailer` method instead of using a ternary here?

jonasraoni: I did it, but for some reason, I moved back, I'll check later, perhaps this part will disappear in a next iteration.

classes/core/PKPContainer.inc.php
```diff
@@ -333,6 +350,32 @@ protected function settingProxyForStreamContext(): void
...
+        return $this;
+    }
+
+    /**
+     * Get the path to the storage directory.
+     */
+    public function storagePath($path = ''): string
+    {
+        return ($this->storagePath ?: (new PrivateFileManager())->getBasePath()) . ($path != '' ? "/$path" : '');
+    }
```
jonasraoni, line 378: Some of the Laravel's facades assumes that the container is implementing the `Illuminate\Contracts\Foundation\Application`, which has a lot of extra methods, as we're not doing it yet, I've just added these methods to satisfy the `Log` facade.

tools/travis/prepare-tests.sh
```diff
@@ -49,6 +49,9 @@ if [[ "$DISABLE_PATH_INFO" == "1" ]]; then
 	sed -i -e "s/disable_path_info = Off/disable_path_info = On/" config.inc.php
 fi
 
+# Setup the mailer to use a logger
+sed -i -e "s/default = sendmail/default = log/;s/log_path/mail.log/" config.inc.php
```
jonasraoni, line 53: I'm disabling the email delivery to decrease the noise in the `error.log`.

classes/services/PKPSchemaService.inc.php
```diff
@@ -377,7 +377,7 @@ public function addPropValidationRules($rules, $ruleKey, $propSchema)
                 }
                 break;
             case 'object':
-                foreach ($propSchema->properties as $subPropName => $subPropSchema) {
+                foreach ($propSchema->properties ?? [] as $subPropName => $subPropSchema) {
```
jonasraoni, line 380: There's a lot of warnings related to this.

This object is built from a `json_decode()` over the JSON schema, and [some "objects" in the schema are missing the `properties`](https://github.com/pkp/pkp-lib/blob/main/schemas/decision.json#L23-L25), I didn't dig deeper than that (for me looks like a way to declare a generic object without fixed properties).

Another way to fix this would be to force declaring the properties in the schema file, so I could add a `"properties": []`.

tests/classes/doi/DoiGeneratorTest.php
```diff
@@ -19,7 +19,7 @@
 
 import('lib.pkp.tests.PKPTestCase');
 
-class DoiUtilsTest extends PKPTestCase
+class DoiGeneratorTest extends PKPTestCase
```
jonasraoni, line 22: There was a warning stating the filename doesn't fit the class name.

classes/services/PKPSchemaService.inc.php
jonasraoni: Moved to https://github.com/pkp/pkp-lib/issues/7995

Issue #7575, opened 2021-12-25
Unauthenticated Open Redirection Vulnerability


agussetyar:
---
name: Bug report
about: Broken PATH/URL Validation on Set Locale Action.
title: 'Open Redirection Vulnerability'
labels: 'Bug'
assignees: 'Maintainer'

---

**Describe the bug**
This bug allow user to redirect to arbitrary URL and can lead to Scam or Phising.

**To Reproduce**
1. Go to 'http://example.com/index.php/index/user/setLocale/NEW_LOCALE?source='
2. fill in parameter value with '@domain.com'
3. Got redirected to evil URL 'http://example.com/index.php/index/user/setLocale/NEW_LOCALE?source=@evil.com'

**What application are you using?**
I think all version of OJS

**Additional information**
HTTP Response Header
```
> GET /index.php/index/user/setLocale/NEW_LOCALE?source=@evil.com HTTP/1.1
> Host: example.com
> User-Agent: curl/7.77.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 302 Moved Temporarily
< Date: Sat, 25 Dec 2021 16:12:16 GMT
< Server: Apache
< Set-Cookie: OJSSID=3dvkl5sttelrnvcdmebksnv601; path=/; domain=example.com
< Location: http://example.com@evil.com
< Content-Length: 0
< Content-Type: text/html; charset=UTF-8
```

Because browser with automatically redirected the URL after @ symbol.

henriqueramos:
@asmecher could you please check those pull requests?

[pkp-lib](https://github.com/pkp/pkp-lib/pull/7578)
[ojs](https://github.com/pkp/ojs/pull/3263)
[omp](https://github.com/pkp/omp/pull/1046)
[ops](https://github.com/pkp/ops/pull/212)

asmecher:
@agussetyar, can you confirm which version of OJS you were able to reproduce this on? Looking at https://github.com/pkp/pkp-lib/blob/38b0064f8f8164dff06ef085cadeeb25936c4d64/pages/user/PKPUserHandler.inc.php#L54 it should be necessary for the `source` parameter to start with a `/` in modern releases. Your example (starting with a `@`) should not behave as you describe in a modern release.

@henriqueramos, were you able to reproduce the initial report?

agussetyar:
@asmecher Sorry. After I verified, it turns out that this vulnerability is in OJS version 2.4.x.
[...]

PR #7578, opened 2021-12-26
pkp/pkp-lib#7575 Fixing the Open Redirect Vulnerability on user/setLocale endpoint

henriqueramos:


pages/user/PKPUserHandler.inc.php
```diff
@@ -50,16 +51,26 @@ public function setLocale($args, $request)
             $session->setSessionVar('currentLocale', $setLocale);
         }
 
-        $source = $request->getUserVar('source');
-        if (preg_match('#^/\w#', $source) === 1) {
-            $request->redirectUrl($source);
+        if (!isset($_SERVER['HTTP_REFERER'])) {
```
jonasraoni, line 54: I didn't watch the PR (I mean the problem vs solution), so this is just a generic comment about the `HTTP_REFERER` 😁

Given this header might be removed by enterprise proxies and I remember Opera (or Firefox) also had an option to disable it, I think it would be better to use the `Origin` header, which should be safer (if proxies attempt to break it, they will also break CORS requests), except for really old browsers.

PR #7622, opened 2022-01-13
pkp/pkp-lib#7016 Restore check for author include in browse list

asmecher:


asmecher:
Needs updating after changes to https://github.com/pkp/pkp-lib/pull/7619/files!

PR #7628, opened 2022-01-17
Fix for full text indexing not working.

bkawula:
I'm not sure why this is happening, but the text that's extracted from the PDF files doesn't get parsed by PKPString. The output is always blank, but converting it to UTF-8 using mb_convert_encoding solves the problem for me. It appears that other have had this problem too: https://forum.pkp.sfu.ca/t/ojs-3-1-2-4-problem-with-rebuildsearchindex/57666


bkawula:
Any change that this can be merged? I keep having to reapply this to all of my instances in order to have full text searching work. I can see that full text searching is broken on almost all OJS journals I see out in the world. For instance: https://journaloftibetanliterature.org/index.php/jtl/search/search?query=metaperformativity&dateFromYear=&dateFromMonth=&dateFromDay=&dateToYear=&dateToMonth=&dateToDay=&authors=

jonasraoni:
@bkawula looks like a harmless update. Before I get it merged, could you please provide a file/text sample that fails to be imported?
I've also found some issues with the search https://github.com/pkp/pkp-lib/issues/8920, this part of the application will start receiving some attention now, looks like there's also a working Lucene plugin, which will probably be released with the version 3.4.

bkawula:
@jonasraoni , it's every PDF file in our systems, but this was a case that was brought up yesterday (all the PDF files here): https://ojs.lib.uwo.ca/index.php/westernumirror/issue/view/1398

It's not an issue that's noticeable because the metadata still gets indexed, but if you copy and paste a unique phrase from a PDF file into the search you can see there's a problem. It's especially evident for publications that have really skeletal metadata like newspaper back catalogues: https://ojs.lib.uwo.ca/index.php/westernnews

It's been sometime since I worked on this, but I remember outputting the value of $text and discovering that it was blank after it went through the various PKPString methods so something is going on there with PKPString.

jonasraoni:
@jnugent as you took part on the [forum discussion](https://forum.pkp.sfu.ca/t/ojs-3-1-2-4-problem-with-rebuildsearchindex/57666) regarding this PR...

Do you still remember where this problem was happening? I'd like to check if it's still happening on OJS +3.3, and also to check if these PRs are fixing the same issue by accident: https://github.com/pkp/pkp-lib/issues/8915.

jnugent:
@jonasraoni apologies for the delay.  I have a new 3.3 install migrated that is experiencing issues with words not appearing in the full text search index. Originally it was due to these calls:

https://github.com/pkp/pkp-lib/blob/stable-3_3_0/classes/search/SubmissionSearchIndex.inc.php#L39

the PKPString:: calls expect UTF8 content and occasionally these methods return empty strings, where $text becomes empty.  This method is fed content from the server side command line tools that extract text from the galleys.

jonasraoni:
I've created an issue here: https://github.com/pkp/pkp-lib/issues/9665
I'll use this PR as a base, just to keep some credits, and add extra fixes.

Issue #3391, opened 2018-02-16
pkp-lib/classes/xslt/XMLTypeDescription.inc.php Doesn't work when behind HTTP proxy


imranh2:
> PHP Warning: DOMDocument::schemaValidate(https://www.crossref.org/schemas/crossref4.3.6.xsd): failed to open stream: Connection refused in /var/www/html/lib/pkp/classes/xslt/XMLTypeDescription.inc.php on line 133, referer: https://ijpds.org/management/importexport/plugin/CrossRefExportPlugin

`if (!$xmlDom->schemaValidate($this->_validationSource)) return false;` Is the problem child.

I've fixed it by replacing that line with:
```

                                $curlCh = curl_init();
                                if ($httpProxyHost = Config::getVar('proxy', 'http_host')) {
                                        curl_setopt($curlCh, CURLOPT_PROXY, $httpProxyHost);
                                        curl_setopt($curlCh, CURLOPT_PROXYPORT, Config::getVar('proxy', 'http_port', '80'));
                                        if ($username = Config::getVar('proxy', 'username')) {
                                                curl_setopt($curlCh, CURLOPT_PROXYUSERPWD, $username . ':' . Config::getVar('proxy', 'password'));
                                        }
                                }
                                curl_setopt($curlCh, CURLOPT_URL, $this->_validationSource);
                                curl_setopt($curlCh, CURLOPT_RETURNTRANSFER, 1);
                                $output = curl_exec($curlCh);

                                if (!$xmlDom->schemaValidate($output)) return false;
```

asmecher:
This would solve the problem for XSD-based schemata, but not e.g. DTDs. Perhaps it's possible to solve this more generally using [stream_context_set_default](http://php.net/manual/en/function.stream-context-set-default.php)?

bozana:
S. also discussion in this issue: https://github.com/pkp/pkp-lib/issues/1955

NateWr:
Is this issue still relevant or can it be closed?

henriqueramos:
@asmecher Could you please check those PRs? Thank you!

https://github.com/pkp/omp/pull/1058
https://github.com/pkp/ojs/pull/3286
https://github.com/pkp/ops/pull/226
https://github.com/pkp/pkp-lib/pull/7638

asmecher:
[...]

PR #7638, opened 2022-01-20
pkp/pkp-lib#3391 Handling XMLTypeDescription validation behind Proxy

henriqueramos:


classes/core/PKPContainer.inc.php
```diff
@@ -272,6 +273,43 @@ protected static function getDefaultMailer(): string
...
+        if (!$proxyUri) {
+            return;
+        }
+
+        /**
+         * `Connection close` here its to avoid slowness. More info at https://www.php.net/manual/en/context.http.php#114867
+         * `request_fulluri` its related to avoid proxy errors. More info at https://www.php.net/manual/en/context.http.php#110449
+         */
+        $opts = [
+            'http' => [
```
asmecher, line 301: I would expect a `http` proxy to be configured in an `http` section, and an `https` proxy to be configured in a separate `https` section, rather than an `http` proxy taking precedence over `https` and only configuring the `http` section of the options. It does appear there may be an open underlying PHP bug related to this (https://bugs.php.net/bug.php?id=74796) that might affect testing, though.

henriqueramos: According to those comments [[1]](https://www.php.net/manual/en/function.stream-context-create.php#74795) [[2]](https://www.php.net/manual/en/function.stream-context-create.php#74795) the key should be `http` even if we pass a HTTPS proxy.

This is mentioned on a comment at your link too.

```
 [2017-06-22 11:42 UTC] anrdaemon at freemail dot ru
And I just found out that settings "https" context has no effect.
Both "https://" and "http://" URL's fall into "http" context.
WHY?…
```

henriqueramos: @asmecher /\

asmecher: What form did you write your `http_proxy` / `https_proxy` configurations in `config.inc.php` to get this working?

henriqueramos: I've filled just only one of them.

asmecher: OK, but what form was the URL?

henriqueramos: `http_proxy = "http://127.0.0.1:8888"`

asmecher: Hmm, what do you get for `stream_get_transports()`? Is this tested working with Tinyproxy?

henriqueramos: I've removed the `http://` schema from `http_proxy` URI config.

```
http_prox="127.0.0.1:8888"
```

Used this code on `index.php` to test the xml loading over proxy:

```
function getPluginsXML() {
    $url = 'https://pkp.sfu.ca/ojs/xml/plugins.xml';

    dd(simplexml_load_file($url));
    return true;
}

register_shutdown_function('getPluginsXML');
```


Those are the logs from `/var/log/tinyproxy/tinyproxy.log` when I hit the OJS url.

```
CONNECT   Feb 02 20:01:47 [1496706]: Connect (file descriptor 7): omp [127.0.0.1]
CONNECT   Feb 02 20:01:47 [1496706]: Request (file descriptor 7): CONNECT pkp.sfu.ca:443 HTTP/1.0
INFO      Feb 02 20:01:47 [1496706]: No upstream proxy for pkp.sfu.ca
INFO      Feb 02 20:01:47 [1496706]: opensock: opening connection to pkp.sfu.ca:443
INFO      Feb 02 20:01:48 [1496706]: opensock: getaddrinfo returned for pkp.sfu.ca:443
CONNECT   Feb 02 20:01:48 [1496706]: Established connection to host "pkp.sfu.ca" using file descriptor 8.
INFO      Feb 02 20:01:48 [1496706]: Not sending client headers to remote machine
INFO      Feb 02 20:01:52 [1496706]: Closed connection between local client (fd:7) and remote client (fd:8)
```

And the printed response on front-end:

Screenshot from 2022-02-02 20-02-52

asmecher: ***Testing***

Have you tested with the original use case? Copying my instructions from Slack before they disappear:

First step: Configure the DOI plugin.
- Go to Settings > Website > Plugins
  - Enable the DOI plugin
  - Go to the DOI plugin's Settings
    - Tick "Articles"
    - Enter e.g. "10.1234" in DOI prefix
    - Save the form
Second step: Assign a DOI to a submission.
- Go to Submissions > Mwandenga
  - Go to Publication
  - Go to Identifiers
  - Assign a DOI (you may have to unpublish first)
  - Publish the submission.
Third step: Export the submission using CrossRef XML.
- Go to the Tools > Import/Export > CrossRef
  - Fill in a name, email, username, and password into the settings form
  - Hit "Save"
  - You should now see the "Articles" tab (reload the page if not)
  - The "Articles" tab should have the published article from the earlier step listed.
  - Select the checkbox for XML validation.
  - Select the checkbox beside the article and hit "Export" to generate the XML.

You should see Tinyproxy log entries showing requests to CrossRef as it fetches the schema files.

***URL formats***

We have two different scenarios in which `http_proxy` / `https_proxy` will be used, and we need to make sure that both uses expect the same format. The right approach needs to be documented in `config.TEMPLATE.inc.php`. If possible we should avoid breaking existing configurations, so if `http` URLs were our previously recommended configuration, then it would be best to stay with it.

The two use cases are:

1. Guzzle (the form that was supported in the past)
    The format is documented here but not in great detail: https://docs.guzzlephp.org/en/stable/request-options.html#proxy
    As a transport, Guzzle uses a few options behind the scenes, e.g. CURL (if available), or PHP's stream functions (if it's not). So the format needs to work for all of Guzzle's transports and we should stay as standard as possible in our recommendations.

2. PHP's default stream context (this is what you're adding).

I checked Guzzle's source and found that [it transforms](https://github.com/guzzle/guzzle/blob/c1fd316f0a0f3325ed1e7cdbe61030418b868f9f/src/Handler/StreamHandler.php#L445) `http://` URLs into `tcp://`. This is in line with what [stream_context_set_default](https://www.php.net/manual/en/function.stream-context-set-default.php) recommends. So I suggest adopting the same pattern rather than dropping the scheme part of the URL.

henriqueramos: @asmecher went thru all testing steps, and seems everything worked well. Following it's the `tinyproxy.log` excerpt.

Screenshot from 2022-02-08 00-45-42

asmecher: OK, looks good. Is it ready for a code review?

henriqueramos: It's ready to another code review round.

Issue #7613, opened 2022-01-12
Keyword auto complete not working while submitting a preprint/book/submission


jonasraoni:
**Describe the bug**
The keywords auto-complete at the 3rd submission step of OPS (perhaps others) isn't working.

**To Reproduce**
1. Ensure the installation have some available keywords
2. Register a new submission
3. At the 3rd step (Enter Metadata), type some letters and wait for the auto complete
4. Nothing happens, no request is made and a div "No search results." appear at the bottom of the page

**What application are you using?**
 OPS 3.3.0

**PRs**
**stable-3_3_0**
https://github.com/pkp/pkp-lib/pull/7625
https://github.com/pkp/ojs/pull/3276
https://github.com/pkp/omp/pull/1054
https://github.com/pkp/ops/pull/222

**main**
https://github.com/pkp/pkp-lib/pull/7626
https://github.com/pkp/ojs/pull/3277
https://github.com/pkp/omp/pull/1053
https://github.com/pkp/ops/pull/220

**Additional information**
![image](https://user-images.githubusercontent.com/361921/149127190-d44ae2a7-c74d-42fa-9242-2e1c08ce7762.png)

jonasraoni:
@asmecher Could you please take a look at the [comment I've left in the OMP's PR](https://github.com/pkp/omp/pull/1053#pullrequestreview-853697392)?

p.s.: I was going to replace the `tag-it` component (looks like it's the only place using it) by our autosuggest, but then I decided to leave it for the next refactoring (replace the current wizard by Vue components).

asmecher:
@jonasraoni, are you waiting for something from me on this?

jonasraoni:
@asmecher I've modified the OMP's VocabHandler handler for  to retrieve ONIX countries:
https://github.com/pkp/omp/pull/1053/commits/241f612a415fa7bef5c63e74e018884d496ce288

If that's fine for you, then I can merge everything.

asmecher:
Thanks, @jonasraoni, please go ahead!

Issue #7639, opened 2022-01-20
Add support for the native import/export plugin to OPS 3.3.0-x


jonasraoni:
**Describe the problem you would like to solve**
The current release of OPS doesn't support the XML tooling for importing/exporting data.

**Describe the solution you'd like**
Merge some updates from the /main branch to the `stable-3_3_0` branch to make this functionality available.

**Who is asking for this feature?**
System administrators looking forward to migrate data from other systems.

**PRs**
https://github.com/pkp/ops/pull/223
https://github.com/pkp/pkp-lib/pull/7640

jonasraoni:
Hi @asmecher!

I've created draft [PRs](https://github.com/pkp/pkp-lib/issues/7639#issue-1109232484), later I'll recheck if the basic features (import/export through the CLI and web interface) are still working fine after rebasing.

I've been using these updates to import data locally, and everything seems to be working fine... Most of the changes are additions. The modified parts are inside classes related to the native import/export plugin.

asmecher:
@defstat, could you review these?

@jonasraoni, I'm assuming you went back to the commits where this was first added to `main` and cherry-picked them to `stable-3_3_0`? Or did you bring in the files from `main` and back-port them to compatibility with `stable-3_3_0` again? Just to know what to look for.

jonasraoni:
If I remember well:
https://github.com/pkp/pkp-lib/pull/7640: Cherry picked the first commit + adjusts.
https://github.com/pkp/ops/pull/223: Grabbed the needed files from `/main` + adjusts.

asmecher:
@jonasraoni, that's a lot more code being added to `stable-3_3_0` `lib/pkp` than I expected naively; is there really no impact on the OJS and OMP stable branch?

jonasraoni:
@asmecher I just checked and I fully agree, I think these PRs were a kind of draft.
I'll check if it's possible to decrease the amount of changes.

Anyway, let me know if you're ok to get this amount of code merged in 3.3.0... Then, given that most of changes are additions, I can just test the updates against OJS/OMP.

asmecher:
Unless the potential impact can be brought down on OJS/OMP 3.3.0-x, I'd suggest just leaving these changes to the 3.4.0 release. If someone has a firm requirement for this support in 3.3.0-x, please speak up!

asmecher:
[...]

PR #7640, opened 2022-01-20
pkp/pkp-lib#7613 Add support for the native import/export plugin

jonasraoni:


asmecher:
See: https://github.com/pkp/pkp-lib/issues/7639#issuecomment-1133090531

Issue #7649, opened 2022-01-25
Add support for limiting allowed hosts


asmecher:
**Description of problem**

OJS/OMP/OPS currently uses the `HTTP_HOST`, `SERVER_NAME`, and `HTTP_X_FORWARDED_HOST` headers to detect the current hostname for formulating absolute URLs.

Per https://portswigger.net/web-security/host-header, these headers may be user-controlled and thus not trustworthy. This could be used to e.g. send password reset emails with poisoned links that direct the user to a 3rd-party site, where the reset hash can be captured. (We are not currently aware of another possible abuse via this mechanism.)

**Solution**

The changes described here add support for a list of allowed host names to be provided in the config.inc.php configuration file. User-supplied values are checked against this list and only one of the listed values will be allowed.

**Affected versions**

This issue affects *all releases* of OJS, OMP, and OPS versions 3.3.0-8 and older.

**How to resolve the issue**
1. Update the code to support the new `allowed_hosts` setting. This can be done by any of the following...
   1. Upgrade OJS, OMP, or OPS to version 3.3.0-9 or newer, *or*
   2. If using git, update to the latest `stable-3_3_0` or `stable-3_2_1` branches, *or*
   3. Apply the appropriate patch for your installation:
     - 3.3.0-9 and newer: Not affected.
     - 3.3.0-1 through 3.3.0-8: [OJS](https://github.com/pkp/pkp-lib/files/8157778/ojs-3.3.0-8.diff.txt) [OMP](https://github.com/pkp/pkp-lib/files/8157780/omp-3.3.0-8.diff.txt) [OPS](https://github.com/pkp/pkp-lib/files/8157781/ops-3.3.0-8.diff.txt)
     - 3.2.0 or 3.2.1: [OJS](https://github.com/pkp/pkp-lib/files/8157784/ojs-3.2.1-4.diff.txt) [OMP](https://github.com/pkp/pkp-lib/files/8157785/omp-3.2.1-4.diff.txt) [OPS](https://github.com/pkp/pkp-lib/files/8157786/ops-3.2.1-4.diff.txt)
2. Add an `allowed_hosts` setting to the `general` section of your `config.inc.php` configuration file. Here is the description/example from `config.TEMPLATE.inc.php`:
  ```
; Restrict the list of allowed hosts to prevent HOST header injection.
; See docs/README.md for more details. The list should be JSON-formatted.
; An empty string indicates that all hosts should be trusted (not recommended!)
; Example:
[...]

PR #7656, opened 2022-01-27
pkp/pkp-lib#7649 Add support for allowed_hosts limits to avoid HOST header injection

asmecher:


bozana:
@asmecher, where/how do we implicitly trust `base_url` hosts?

bozana:
I would have two more questions, not closely connected with this issue -- they work, so you can ignore them:

1) In the function `getBaseUrl`: 
Why do we use `false` in `getServerHost` when the first parameter and the return value should be a string:
```
 $serverHost = $this->getServerHost(false);
        if ($serverHost !== false) {
        ...

```
2) In getServerHost:
Why don't we have `getServerHost($default = 'localhost'...` instead of:
```
getServerHost($default = null, $includePort = true)
    {
        if ($default === null) {
            $default = 'localhost';
        }
```

asmecher:
> @asmecher, where/how do we implicitly trust base_url hosts?

What I mean is that when the `base_url[...]` configuration entries are being used, we use them in place of `getIndexUrl` (which in turn relies on the `HTTP_HOST` header) in order to construct URLs; see `PKPRouter::_urlGetBaseAndContext`. So using `base_url[...]` means we don't rely on `HTTP_HOST` when generating URLs.

> In the function `getBaseUrl`, why do we use false in `getServerHost` when the first parameter and the return value should be a string?

This is old code; it's checking whether a hostname could be automatically detected or not. By passing in a `false` as a default, then checking if that `false` is passed back out of the function, we're determining whether the default had to be used (i.e. nothing better could be identified for a hostname).

> In getServerHost, why don't we have `getServerHost($default = 'localhost'...` instead of...

This is also old code, but using
```php
function myFunction($param = null, $someOtherParameter = null) {
    if ($param === null) $param = 'someStringValue';
    ...
}
```
instead of
```php
function myFunction($param = 'someStringValue', $someOtherParameter = null) {
    ...
}
```
...can be useful to allow callers to avoid re-coding the `someStringValue` when they really just want to provide something for `$someOtherParameter` and leave `$param` at its default. This kind of hack can go away when we're allowed to use named parameters.

bozana:
Thanks @asmecher, I would then say it can be merged :)

jonasraoni:
@asmecher What about failing with a 400 + a helpful message when the host is invalid? Just to avoid someone pointing a fake domain to the installation.

Perhaps it would be useful to also allow the rule to be based on a regular expression. Then the user would be able to define the allowed domains using something like `/^[\w]+\.pkp\.sfu\.ca$/`.

asmecher:
Thanks, @bozana and @jonasraoni!

> What about failing with a 400 + a helpful message when the host is invalid? Just to avoid someone pointing a fake domain to the installation.

This led me to a bit of a rewrite. The earlier implementation had operated when a URL was generated (e.g. when generating a response), which is a bad place to suddenly jump in and respond with a `400`. So instead I wrote a new authorization policy, which is a much more logical place to check and intervene. I've updated all the PRs to suit.

@bozana and @jonasraoni, could you take another look? Much appreciated!

classes/security/authorization/AllowedHostsPolicy.inc.php
```diff
@@ -0,0 +1,72 @@
...
+            AuthorizationPolicy::AUTHORIZATION_PERMIT : AuthorizationPolicy::AUTHORIZATION_DENY;
+    }
+
+    /**
+     * Handle a mismatch in the allowed hosts expectation.
+     */
+    public function callOnDeny()
+    {
+        http_response_code(400);
+        fatalError('400 Bad Request');
```
jonasraoni, line 70: I think it's useful to log a helpful error message here, just to help the admin figure out why he's getting a 400.

jonasraoni (approved):

bozana:
@asmecher, sorry for my 'silly' question 'again': What if the host is one of the base_url hosts now that the check is at the beginning? Would PKPHandler authorization work? :flushed:
Else, yes, you can merge... :-)

asmecher:
> What if the host is one of the base_url hosts now that the check is at the beginning? Would PKPHandler authorization work?

@bozana, not a silly question at all! Now that this is implemented as a policy/request validation concern rather than a URL generation concern, it'll be necessary for either...
- all domains to be explicitly listed in `allowed_hosts`, or
- `allowed_hosts` checking to be disabled by leaving the setting empty. This can actually be done safely when virtual hosting, since the web server will already be using the `Host` header to map the request to the site, essentially filtering the header for allowed values (**if that's the way virtual hosting is configured**). In that scenario there would be no need to use `allowed_hosts`.

classes/security/authorization/AllowedHostsPolicy.inc.php
asmecher: Good idea, thanks -- I will add this.

PR #7740, opened 2022-03-05
pkp/pkp-lib#7698 Implementing Laravel Routing

henriqueramos:


api/v1/users/PKPUserHandler.inc.php
```diff
@@ -15,92 +15,50 @@
...
-            'orderBy',
-            'orderDirection',
-            'roleIds',
-            'searchPhrase',
-            'status',
-        ]);
-
-        $params['contextId'] = $context->getId();
-
-        HookRegistry::call('API::users::params', [&$params, $slimRequest]);
```
asmecher, line 103: Rather than removing the hook entirely, I would suggest:
- Providing an equivalent set of parameters to it (`$params, $request`)
- Documenting the change in conventions at the top of the PR, so Nate can be flagged about developer documentation updates when this is merged.

(Same with others below.)

If the hook is removed, it'll make it really tough for developers who use the hook to debug why -- their code will stop getting called and they won't know why. A breaking change (the slim request being swapped out for a Laravel request object) will give them an error message they will be able to work with.

api/v1/users/PKPUserHandler.inc.php
```diff
@@ -15,92 +15,50 @@
...
-    public function getMany($slimRequest, $response, $args)
+    public function getMany(Request $request): JsonResponse
     {
-        $request = $this->getRequest();
-        $context = $request->getContext();
-
-        if (!$context) {
-            return $response->withStatus(404)->withJsonError('api.404.resourceNotFound');
-        }
+        $querystrings = $request->query(null);
```
asmecher, line 42: Use lowerCamelCase: `$queryStrings`. (Here and below.)

classes/core/Controllers/Middlewares/FillContextBasedOnUri.php
```diff
@@ -0,0 +1,53 @@
...
+use Illuminate\Http\Request;
+use PKP\core\Core;
+use PKP\db\DAORegistry;
+
+class FillContextBasedOnUri
+{
+    /**
+     * Handle an incoming request.
+     */
+    public function handle(Request $request, Closure $next)
```
asmecher, line 18: This duplicates code from PKPRouter::getContext, I think; it would be better to call that directly if possible. (A lot of that code is old and ready for removal since we're no longer supporting apps with different context depths -- OCS and OHS used to do this -- so spreading it out further into the codebase is code rot.)

classes/core/Controllers/Middlewares/DecodingApiToken.php
```diff
@@ -0,0 +1,101 @@
...
+            $apiUser = Repo::user()->getByApiKey($apiToken);
+
+            if ($apiUser === null || !$apiUser->getData('apiKeyEnabled')) {
+                return new JsonResponse(
+                    ['error' => 'api.403.unauthorized'],
+                    Response::HTTP_UNAUTHORIZED
+                );
+            }
+
+            $contextId = $request->attributes->get('pkpContext')->getData('id');
```
asmecher, line 68: These next two lines are dead code, I think?

classes/core/Controllers/Middlewares/Permissions/MatchRoles.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param string $roles Comma-separated list with roles
+     */
+    public function handle($request, Closure $next, string ...$roles)
+    {
+        try {
+            $userRoles = $this->getUserRoles($request);
+
+            if (array_intersect($userRoles, $roles) === []) {
```
asmecher, line 47: At a glance, I'm not sure this is working... `getUserRoles` will return an array of `roleId => true`, and this function will be provided with a string containing a list of numeric constants, if I'm reading it right.

api/v1/users/PKPUserHandler.inc.php
```diff
@@ -318,54 +333,47 @@ private function _processAllowedparams($params, $allowedKeys)
...
+    public function __construct()
     {
-        $request = $this->getRequest();
+        $this->_handlerPath = 'users';
 
-        $context = $request->getContext();
-        if (!$context) {
-            return $response->withStatus(404)->withJsonError('api.404.resourceNotFound');
-        }
+        $roles = implode(',', [Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR]);
```
asmecher, line 347: Is it possible to continue to use the role IDs symbolically rather than collapsing them into a string?

asmecher (changes requested):
Hi Henrique! I've made a few specific requests, but here are a few general ones too:

- Unless we roll out this change comprehensively and remove Slim for 3.4.0 -- which is probably too ambitious -- this PR will will serve as a reference to developers refactoring similar handlers later. So to keep it as simple as possible it's best to keep miscellaneous tweaks out. The shorter the PR, the better.
- The grafting of Laravel directory names into our longstanding patterns is awkward, e.g. classes/core/Controllers/Middlewares/ConfigureBaseRequest, which mixes upper/lower cases and Laravel vs. PKP concerns. If you're following a purely Laravel pattern that you think is going to be rolled out more broadly, then I'd stick with purely Laravel directory naming. (I think `Middlewares` is not pluralized, though? If that's not a Laravel convention, then staying with PKP conventions e.g. `classes/core/middleware` would be fine with me.)

api/v1/users/PKPUserHandler.inc.php
```diff
@@ -220,13 +146,102 @@ public function getReviewers($slimRequest, $response, $args)
...
+            )
+        );
+
+        return $response;
+    }
+
+    /**
+     * Get a single user
+     */
+    public function getUser(Request $request): JsonResponse
```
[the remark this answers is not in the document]
henriqueramos: Won't be possible to move this under the class `PKPUserHandler`, who extends the `APIHandler`. We needed a class extending the `Illuminate\Routing\Controller`. :/

classes/core/Controllers/Middlewares/DecodingApiToken.php
henriqueramos: No, the `$apiUser->getRoles($contextId, true);` is used here to setting the user's roles, based on the given context.

See https://github.com/pkp/pkp-lib/blob/main/classes/user/User.inc.php#L471

classes/core/Controllers/Middlewares/FillContextBasedOnUri.php
henriqueramos: I tried to keep the logic on this middleware simple. I got the logic to retrieve context based on path data, did the database requests and inject it as an attribute on Request itself.

The `PKPRouter::getContext` isn't a static method, so I couldn't use it easier here. I will always need to inject and construct a new PKPRouter instance, to call the `getContext`. Thus, keeping it extremely coupled. :(

api/v1/users/PKPUserHandler.inc.php
henriqueramos: Unfortunately no, because middleware params on Laravel accept only single string or CSVs.

As [mentioned here](https://laravel.com/docs/7.x/middleware#middleware-parameters):

>Middleware parameters may be specified when defining the route by separating the middleware name and parameters with a :. Multiple parameters should be delimited by commas:

classes/core/Controllers/Middlewares/Permissions/MatchRoles.php
henriqueramos: Fixed the return of `getUserRoles`. Right now, he's returning an array with the roles Ids, thus enabling us to use `array_intersect` correctly here.

classes/core/PKPContainer.inc.php
```diff
@@ -263,7 +272,7 @@ protected function loadConfiguration()
      */
     public function basePath($path = '')
     {
-        return $this->basePath . ($path ? "/$path" : $path);
+        return $this->basePath . ($path ? "/${path}" : $path);
```
asmecher, line 275: This form of string interpolation will be deprecated: https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation

PR #7755, opened 2022-03-10
pkp/pkp-lib#7716: Hide DB password in the administration "system info" page 

### LINKED ISSUE #7716: Hide DB password in the administration "system info" page
**Describe the bug**
Every user with rights to see the Administration pages can see the DB pwd.
As far as admins know this info (or can check it in the config.inc.php) may be is a good idea not showing this info in this pages... or at least, transform it to asterisks.

**To Reproduce**
1. Go to 'Administration' 
2. Click on 'System Information' (url: https://foojournal.org/admin/systemInfo)
3. Scroll down to 'database section'
4. See the DB pwd.

 OJS, 3.x
 
**Additional information**

imagen

Tyl13:
Resolves #7716 . It explicitly checks the category for database before checking the name. This way, redacting any new passwords added to OJS Configuration can be done on a case by case basis.

### DIFF
--- a/templates/admin/systemInfo.tpl
+++ b/templates/admin/systemInfo.tpl
@@ -96,24 +96,15 @@
 						<td colspan="2" class="app--admin__systemInfoGroup">{$category}</td>
 					</tr>
 					{foreach from=$settings item="value" key="name"}
-						{if $category == "database"}
-							{if $name != "password"}
+					
 								<tr>
 									<td>{$name|escape}</td>
-									<td>{$value|escape}</td>
+									{if $value == "password"}
+										<td>**************</td>
+									{else}
+										<td>{$value|escape}</td>
+									{/if}
 								</tr>
-							{else}
-								<tr>
-									<td>{$name|escape}</td>
-									<td>**************</td>
-								</tr>
-							{/if}
-						{else}
-							 <tr>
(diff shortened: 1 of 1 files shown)


templates/admin/systemInfo.tpl
```diff
@@ -96,10 +96,24 @@
 						<td colspan="2" class="app--admin__systemInfoGroup">{$category}</td>
 					</tr>
 					{foreach from=$settings item="value" key="name"}
-						<tr>
-							<td>{$name|escape}</td>
-							<td>{$value|escape}</td>
-						</tr>
+						{if $category == "database"}
```
asmecher, line 99: Hi @Tyl13! I think this could be done a lot more succinctly with
```
{if $name == "password"}
	<td>********</td>
{else}
	<td>{$value|escape}</td>
{/if}
```
(Any setting named `password` should not be displayed.)

Tyl13: I agree, I originally had it on all passwords that could be displayed. Also, that is a lot more succinct. I can't believe I didn't to limit it to the value td instead of the whole row. Thanks, I will fix and quickly throw back up.

asmecher:
Thanks, @Tyl13! Merged and cherry-picked forward to `main`.

PR #7758, opened 2022-03-14
pkp/pkp-lib#6778 Whitespace around author names should be trimmed 

Tyl13:
This uses array map to trim white space from each given name. This is a simple fix to #6778 . Tested on English only, and on English and German with an empty German name and a name with whites paces for both English and German. 

I am unsure if it should be a bug to be able to choose to not put a name for each different language, or if it should automatically assign the default name to each locale that is in use?


asmecher:
Thanks, @Tyl13! Committed and forward-ported to `main`. I can confirm that it's intended behaviour for names not to be forced for entry in *all* enabled languages; that would lead to users just copy/pasting the same data into all fields. We have a series of fallbacks during display that should lead to the "best available" language being chosen.

PR #7762, opened 2022-03-15
pkp/pkp-lib#7761 Update psysh and other dependencies

NateWr:


asmecher:
See https://github.com/pkp/pkp-lib/issues/7761#issuecomment-1133384354

PR #7778, opened 2022-03-21
pkp/pkp-lib#7768 Fix tests after id removed from element 

NateWr:


asmecher:
See: https://github.com/pkp/pkp-lib/issues/7768#issuecomment-1074509370

PR #7788, opened 2022-03-23
pkp/pkp-lib#7787 Fix User::getContactSignature() return type

Vitaliy-1:


classes/user/User.inc.php
```diff
@@ -415,11 +415,14 @@ public function setInlineHelp($inlineHelp)
         $this->setData('inlineHelp', $inlineHelp);
     }
 
-    public function getContactSignature(?string $locale = null)
+    public function getContactSignature(?string $locale = null): string
```
asmecher, line 418: @Vitaliy-1, could you add self-documentation for this function?

PR #7803, opened 2022-03-29
pkp/pkp-lib#6222 Embed discussion content into the notification message

forgive38:
this is a proposition, with adding a new email template DISCUSSION_NOTIFICATION

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -66,26 +66,26 @@
         },
         {
             "name": "brick/math",
-            "version": "0.9.3",
+            "version": "0.10.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/brick/math.git",
-                "reference": "ca57d18f028f84f777b2168cd1911b0dee2343ae"
+                "reference": "459f2781e1a08d52ee56b0b1444086e038561e3f"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/brick/math/zipball/ca57d18f028f84f777b2168cd1911b0dee2343ae",
-                "reference": "ca57d18f028f84f777b2168cd1911b0dee2343ae",
+                "url": "https://api.github.com/repos/brick/math/zipball/459f2781e1a08d52ee56b0b1444086e038561e3f",
+                "reference": "459f2781e1a08d52ee56b0b1444086e038561e3f",
                 "shasum": ""
             },
             "require": {
                 "ext-json": "*",
-                "php": "^7.1 || ^8.0"
+                "php": "^7.4 || ^8.0"
             },
             "require-dev": {
                 "php-coveralls/php-coveralls": "^2.2",
-                "phpunit/phpunit": "^7.5.15 || ^8.5 || ^9.0",
-                "vimeo/psalm": "4.9.2"
+                "phpunit/phpunit": "^9.0",
+                "vimeo/psalm": "4.25.0"
             },
             "type": "library",
             "autoload": {
@@ -110,19 +110,15 @@
             ],
             "support": {
                 "issues": "https://github.com/brick/math/issues",
-                "source": "https://github.com/brick/math/tree/0.9.3"
+                "source": "https://github.com/brick/math/tree/0.10.2"
             },
             "funding": [
                 {
                     "url": "https://github.com/BenMorel",
                     "type": "github"
-                },
-                {
-                    "url": "https://tidelift.com/funding/github/packagist/brick/math",
-                    "type": "tidelift"
                 }
             ],
-            "time": "2021-08-15T20:50:18+00:00"
+            "time": "2022-08-10T22:54:19+00:00"
         },
         {
             "name": "components/jquery",
@@ -915,34 +911,31 @@
         },
         {
             "name": "doctrine/event-manager",
-            "version": "1.1.1",
+            "version": "1.1.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/event-manager.git",
-                "reference": "41370af6a30faa9dc0368c4a6814d596e81aba7f"
+                "reference": "eb2ecf80e3093e8f3c2769ac838e27d8ede8e683"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/event-manager/zipball/41370af6a30faa9dc0368c4a6814d596e81aba7f",
-                "reference": "41370af6a30faa9dc0368c4a6814d596e81aba7f",
+                "url": "https://api.github.com/repos/doctrine/event-manager/zipball/eb2ecf80e3093e8f3c2769ac838e27d8ede8e683",
+                "reference": "eb2ecf80e3093e8f3c2769ac838e27d8ede8e683",
                 "shasum": ""
             },
             "require": {
                 "php": "^7.1 || ^8.0"
             },
             "conflict": {
-                "doctrine/common": "<2.9@dev"
+                "doctrine/common": "<2.9"
             },
             "require-dev": {
-                "doctrine/coding-standard": "^6.0",
-                "phpunit/phpunit": "^7.0"
+                "doctrine/coding-standard": "^9",
+                "phpstan/phpstan": "~1.4.10 || ^1.5.4",
+                "phpunit/phpunit": "^7.5 || ^8.5 || ^9.5",
+                "vimeo/psalm": "^4.22"
             },
             "type": "library",
-            "extra": {
-                "branch-alias": {
-                    "dev-master": "1.0.x-dev"
-                }
-            },
             "autoload": {
                 "psr-4": {
                     "Doctrine\\Common\\": "lib/Doctrine/Common"
@@ -989,7 +982,7 @@
             ],
             "support": {
                 "issues": "https://github.com/doctrine/event-manager/issues",
-                "source": "https://github.com/doctrine/event-manager/tree/1.1.x"
+                "source": "https://github.com/doctrine/event-manager/tree/1.1.2"
             },
             "funding": [
                 {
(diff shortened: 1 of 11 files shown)


NateWr:
Thanks @forgive38! Some quick, initial feedback before I dive in deeper: have a read of the [Variables](https://docs.pkp.sfu.ca/dev/documentation/en/utilities-email#variables) section in our new documentation about the email system in 3.4. In particular, look at the information about passing an object (`Query $query`) into a mailable's constructor, and creating variables _and descriptions_ from that.

An example you can follow is [DecisionEmailVariable](https://github.com/pkp/pkp-lib/blob/main/classes/mail/variables/DecisionEmailVariable.inc.php). You'll want to create a similar `QueryEmailVariable`. This is how we are ensuring consistency of email variables across all mailables.

So instead of doing `$mailable->addData(...)`, you'll pass in the query to the mailable's constructor, (`$mailable = new MyMailable($context, $submission, $query)`) and that will generate the email  variables for you (after you've set up `QueryEmailVariable`). The descriptions are important because we will show those to the editor when they are editing email templates.

Otherwise at first glance this looks good. I will dive in deeper once you've updated with the email variable class.

NateWr:
@Vitaliy-1 may be able to provide some insight here. He's been working on refactoring some of the notification stuff and he may be able to recommend a strategy to convert this to the new mailable system.

Vitaliy-1:
If there is no aim to refactor notifications in this PR, I'd just leave it for now and send email directly with the new Mailable system. Notification Manager allows suppressing emails by passing an additional parameter to `createNotification()` method.

forgive38:
@NateWr : I update PRs

controllers/grid/queries/form/QueryNoteForm.inc.php
```diff
@@ -141,24 +145,39 @@ public function execute(...$functionArgs)
...
+        if (!in_array($currentUser->getId(), $queryDao->getParticipantIds($query->getId()))) {
+            $queryDao->insertParticipant($query->getId(), $currentUser->getId());
         }
 
         $notificationManager = new NotificationManager();
+        /** @var NotificationSubscriptionSettingsDAO $notificationSubscriptionSettingsDAO */
+        $notificationSubscriptionSettingsDao = DAORegistry::getDAO('NotificationSubscriptionSettingsDAO');
+
         foreach ($queryDao->getParticipantIds($query->getId()) as $userId) {
+            $user = Repo::user()->get((int) $userId);
```
NateWr, line 179: It looks like `$user` isn't used until the end of the loop, in `$mailable->recipients([$user])`. If that's the case, you can move this down there so that it's closer to the code that uses it.

controllers/grid/queries/form/QueryNoteForm.inc.php
```diff
@@ -141,24 +145,39 @@ public function execute(...$functionArgs)
...
                 $queryDao->updateObject($query);
             }
         }
 
+        $submissionId = $query->getAssocId();
+        $submission = Repo::submission()->get($submissionId);
+
+        $mailable = new MailDiscussionMessage($request->getContext(), $submission, $query);
+        $emailTemplate = Repo::emailTemplate()->getByKey($request->getContext()->getId(), $mailable::getEmailTemplateKey());
+        $context = $request->getContext();
```
NateWr, line 161: Throughout this function you mix use of `$request->getContext()` and `$context`. I don't mind whichever you prefer, but I'd recommend keeping it consistent throughout the function.

controllers/grid/queries/form/QueryNoteForm.inc.php
```diff
@@ -141,24 +145,39 @@ public function execute(...$functionArgs)
...
+        $mailable->body($emailTemplate->getLocalizedData('body'))
+                ->subject('[' . $context->getLocalizedAcronym() . '] '."Re: ".$emailTemplate->getLocalizedData('subject'))
```
NateWr, line 163: We're leaving out journal acronyms from email subjects. The subject itself will be customizable using any of the email template variables, so if someone wants to include a journal prefix on all emails, they can do so by editing the subject to be `[{$acronym}] {$subject}`. But the default should just be `{$subject}`.

(Also, we can't have `Re:` in the code here. If you want to put that in, it should go into a locale key. But also it should only appear on replies, not the head note.)

controllers/grid/queries/form/QueryNoteForm.inc.php
```diff
@@ -169,22 +188,47 @@ public function execute(...$functionArgs)
...
+
+                $mailable->addData([
+                    'notificationUrl' => $notificationManager->getNotificationUrl($request, $notification),
+                    'unsubscribeLink' =>
+                        '<br /><a href=\'' .
+                        $notificationManager->getUnsubscribeNotificationUrl($request, $notification) .
+                        '\'>' .
+                        __('notification.unsubscribeNotifications') .
+                        '</a>',
+                    'siteTitle' => $request->getContext()->getLocalizedName()
```
NateWr, line 226: You can remove `siteTitle` here (see comment above).

controllers/grid/queries/form/QueryNoteForm.inc.php
```diff
@@ -169,22 +188,47 @@ public function execute(...$functionArgs)
...
                 $query->getId(),
-                Notification::NOTIFICATION_LEVEL_TASK
+                Notification::NOTIFICATION_LEVEL_TASK,
+                null,
+                true// don"t send email
             );
-        }
 
+            // Check if the user is unsubscribed
+                $notificationSubscriptionSettings = $notificationSubscriptionSettingsDao->getNotificationSubscriptionSettings(
```
NateWr, line 209: Looks like extra indentation starts here.

NateWr (changes requested):
Thanks @forgive38! One kind of tricky thing to figure out regarding how we get the subject/message -- I may have mis-directed you. Otherwise, mostly some minor style/comment changes.

classes/mail/variables/QueryEmailVariable.inc.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $notes = $this->query->getReplies(null, \PKP\note\NoteDAO::NOTE_ORDER_DATE_CREATED, \PKP\db\DAO::SORT_DIRECTION_DESC, true);
+        $lastNote = $notes->next();
+
+        return
+        [
+            self::SUBJECT => $this->query->getHeadNote()->getTitle(),
+            self::LAST_NOTE_CONTENT => $lastNote->getContents(),
+        ];
+    }
```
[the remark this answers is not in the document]
Vitaliy-1: I'd prefer to add notes as something like read-only property to the `Query` object and set it before the latter is passed to the Mailable's constructor.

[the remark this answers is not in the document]
Vitaliy-1: Or keep it simple, use `Query::setData()`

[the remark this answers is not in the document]
NateWr: So you mean something like this?

```php
$query->setData();
$mailable = new ExampleMailable($query);
```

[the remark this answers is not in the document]
Vitaliy-1: Yes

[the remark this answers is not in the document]
NateWr: I'm worried that it will create confusion with the type hinting. The `ExampleMailable` can type hint that it wants a `Query` object, but it has no way of saying that it expects `$query->setData()` to be run before it is passed in.

I'm leaning towards something like this in the mailable:

```php
public const SUBJECT = 'subject';
public const MESSAGE = 'message';

/**
 * @param array $notes All notes in the query, ordered by most recent first
 * @param string $subject The query's subject
 */
public function __construct(Context $context, PKPSubmission $submission, array $notes, string $subject)
{
    parent::__construct(array_slice(func_get_args(), 0, -2));

    $lastNote = end($notes);
    
    $this->addData([
        self::SUBJECT => $this->subject,
        self::MESSAGE => $lastNote->getContents(),
    ]);
}

public static function getDataDescriptions(): array
{
    $variables = parent::getDataDescriptions();
    $variables[self::SUBJECT] = __('...');
    $variables[self::MESSAGE] = __('...');
    return $variables;
}
```

What do you think?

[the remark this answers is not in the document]
forgive38: Just for information, you can't use variable named 'message' in template because Laravel populate this variable
https://laravel.com/docs/9.x/mail#inline-attachments
So I named it 'content'

forgive38:
@NateWr : sorry for the delay ...

So I push an update proposition on top on work made by @Vitaliy-1 

Maybe we can factorize classes/mail/mailables/DiscussionSubmission.php , DiscussionReview.php  DiscussionCopyediting.php  DiscussionProduction.php ?

composer.lock
```diff
@@ -66,26 +66,26 @@
         },
         {
             "name": "brick/math",
-            "version": "0.9.3",
+            "version": "0.10.2",
```
NateWr, line 69: Don't commit any changes to the `composer.lock` file. We only commit changes when we update the `composer.json` file.

controllers/grid/queries/traits/StageMailable.php
```diff
@@ -30,7 +31,7 @@ abstract public function getStageId();
     /**
      * @return Mailable which corresponds to the given workflow stage
      */
-    protected function getStageMailable(Context $context, Submission $submission): Mailable
+    protected function getStageMailable(Context $context, Submission $submission, string $subject = "", string $message = ""): Mailable
```
NateWr, line 34: Use single quotes, `'`, instead of double quotes, `"`, to define the strings.

NateWr (changes requested):
This looks great, thanks @forgive38! 

> Maybe we can factorize classes/mail/mailables/DiscussionSubmission.php , DiscussionReview.php DiscussionCopyediting.php DiscussionProduction.php ?

Let's use a `trait` that can hold the variable constants and some helper methods that the mailable classes can use. You can look at other traits, like `ReviewerComments`  in `lib/pkp/classes/mail/traits` to see how we're using this.

@Vitaliy-1 will you pick up this PR and do whatever needs to be done to incorporate https://github.com/pkp/pkp-lib/issues/8127 and https://github.com/pkp/pkp-lib/issues/7874? You can merge it separately or pull it into your own work. Whatever is easiest for you.

controllers/grid/queries/traits/StageMailable.php
forgive38: ok

forgive38:
I try to make a `trait`but I'm not sure that it is efficient: please feel free to modify the code or remove the trait
HTH

Vitaliy-1:
Thanks, @forgive38. I've added a template and merged it. PRs with these commits: https://github.com/pkp/pkp-lib/issues/6222#issuecomment-1265067079.

Issue #7793, opened 2022-03-24
Allow for plugins to describe support for fuzzy or wildcarded versions


ctgraham:
**Describe the problem you would like to solve**
In the Plugin Gallery, plugin developers are currently required to explicitly name every application version which the plugin version supports.  This means that new assertions must be added even after patch releases.  The version contract should, however, allow us to forecast version support, promising (for example) for a plugin to work in 3.3.0-_anything_, or even 3.3._anything_-_anything_.

**Describe the solution you'd like**
Allow for plugin version support to be specified in plugins.xml as wildcards, or as partial versions.  See also: https://github.com/pkp/pkp-lib/issues/7531 for parallel discussion.

**Who is asking for this feature?**
Plugin developers

**Additional information**
The limiter seems to be here: https://github.com/pkp/pkp-lib/blob/087819073c4d323dc4d7c4d06eba8a793c4d3ad0/classes/plugins/PluginGalleryDAO.inc.php#L308-L342

Tyl13:
While I have been doing work on this, I was wondering. Should the user be explicit about the version that he is wild carding? For example, I consider `3.3.*-*` as explicit but that is technically the same as `3.3.*`  when we stop looking at the wild cards and only comparing the version specified. 

If the second version is allowed, then should I ignore any characters after the wild. For example, if a person placed `3.3*-10`. It doesn't make sense for that to be like that but it could have been a mistake and they forgot to specify the build version. In which case, we wouldn't want it to succeed because it could be not functional for the version it is being used on.

ctgraham:
> There could be an issue where it works on 3.3 until 3.3.0-10. Then the plugin has to be changed to be explicit for all versions of 3.3.0 except for 3.3.0-10.

Yes, but according to semantic versioning, `-10` should not introduce a backwards incompatible change.  If the plugin introduces an incompatible change, it is up to the author to either redeclare the compatibility, or test for and address the version incompatibility in the plugin code.

Also, note that in the Plugin Gallery any more recent release overrides former releases.  So, if I promise compatibility with "3.3" in my plugin's "1.2", but "3.3.8" introduces backwards incompatible changes, I can release "1.3" and declare compatibility with "3.3" again.

asmecher:
Hi all -- I'm bungee-jumping in without reading the whole thread, just to say that if it's possible to adopt patterns or even libraries/tools to make the version specifications compatible with e.g. what Composer uses, I'd recommend it.

[...]

PR #7809, opened 2022-03-30
pkp/pkp-lib#7793 Changed to use Composer's Semver

Tyl13:
A new function, isCompatibile, will now create a versionParser class and semver class from the composer/semver package. Then it parses the version passed to compare, this parsing includes any constraints. It should allow for any of these as a constraint: https://getcomposer.org/doc/articles/versions.md . My testing did not bring up any issues with the constraints.  

I tried to think of a way to remove the second foreach function, but you can end up with multiple multiConstraints after parsing the constraints. For example, `>1.1 <1.2 || 2.2 || >=3.3 <3.5"` will end up with a multiConstraint that has the first and last contraints as multiConstraints and the second as a single Constraint. It is improved by breaking out of the second foreach if it fails at anypoint since the multiConstraint is viewed as a logical AND. It is also able to succeed as long one of the groupings succeeds since they are viewed as logical ORs.

This fixes: https://github.com/pkp/pkp-lib/issues/7793

PR for main: https://github.com/pkp/pkp-lib/pull/7876

### DIFF
--- a/classes/site/Version.inc.php
+++ b/classes/site/Version.inc.php
@@ -51,8 +51,8 @@ class Version extends DataObject {
 		if (is_object($version)) {
 			return $this->compare($version->getVersionString());
 		}
-		$breaks = substr_count($version, ".");
-		return version_compare($this->getVersionString(true, $breaks), $version);
+		$seperators = substr_count($version, ".");
+		return version_compare($this->getVersionString(true, $seperators), $version);
 	}
 
 	/**
@@ -274,14 +274,15 @@ class Version extends DataObject {
 	/**
 	 * Return complete version string.
 	 * @numeric boolean True (default) iff a numeric (comparable) version is to be returned.
+	 * @seperators	integer	4	(default) The number of seperators (.) in the version string, used to truncate to the same number of seperators.
 	 * @return string
 	 */
-	function getVersionString($numeric = true, $breaks = 3) {
-		if ($breaks == 0) {
+	function getVersionString($numeric = true, $seperators = 3) {
+		if ($seperators == 0) {
 			$numericVersion = sprintf('%d', $this->getMajor());
-		} elseif ($breaks == 1) {
+		} elseif ($seperators == 1) {
 			$numericVersion = sprintf('%d.%d', $this->getMajor(), $this->getMinor());
-		} elseif ($breaks == 2) {
+		} elseif ($seperators == 2) {
 			$numericVersion = sprintf('%d.%d.%d', $this->getMajor(), $this->getMinor(), $this->getRevision());
 		} else {
 			$numericVersion = sprintf('%d.%d.%d.%d', $this->getMajor(), $this->getMinor(), $this->getRevision(), $this->getBuild());
--- a/classes/site/Version.inc.php
+++ b/classes/site/Version.inc.php
@@ -51,7 +51,8 @@ class Version extends DataObject {
 		if (is_object($version)) {
 			return $this->compare($version->getVersionString());
 		}
-		return version_compare($this->getVersionString(), $version);
+		$wilds = $this->findWildCards($version);
+		return version_compare($this->getVersionString($wilds), $version);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


classes/site/Version.inc.php
```diff
@@ -13,7 +13,8 @@
  *
  * @brief Describes system version history.
  */
-
+use Composer\Semver\VersionParser;
+use Composer\Semver\Semver as Sem;
```
ctgraham, line 17: Was this aliased because there was another `Semver` reference already in scope, or for another reason?

classes/site/Version.inc.php
```diff
@@ -284,6 +285,47 @@ function getVersionString($numeric = true) {
 
 		return $numericVersion;
 	}
+
+	/**
+	 * Checks if this version is compatible with given version
```
ctgraham, line 290: We are overloading the meaning of "version" here: both to mean a specific version (or the `Version` object) and to mean a version constraint string.  This should be clarified.

classes/site/Version.inc.php
```diff
@@ -284,6 +285,47 @@ function getVersionString($numeric = true) {
...
+	 * 		false iff the version given is incompatible with this version
+	 * @param $version string/Version the version to compare against
+	 * @return boolean
+	 */
+	function isCompatible($version) {
+		$versionParser = new VersionParser();
+		$semver = new Sem();
+		$parsed = $versionParser->parseConstraints($version);
+		$systemVersion = $this->getVersionString();
+		if ($parsed instanceof Composer\Semver\Constraint\MultiConstraint) {
```
ctgraham, line 302: Why do multiple constraints need to be examined separately?   Will a direct `return $semver->satisfies($systemVersion, $parsed);` not suffice?

classes/site/Version.inc.php
```diff
@@ -284,6 +285,47 @@ function getVersionString($numeric = true) {
...
+			$success = false;
+			$localFailure = false;
+			$multiConstraints = $parsed->getConstraints();
+			foreach ($multiConstraints as $constraints) {
+				$localFailure = false;
+				if ($constraints instanceof Composer\Semver\Constraint\MultiConstraint) {
+					$constraintsLV2 = $constraints->getConstraints();
```
ctgraham, line 309: Hopefully superseded, but `$constraintsLV2` is not camelCase compliant.   To describe `$constraintsLevelTwo`, spell it out for future readers.

composer.json
```diff
@@ -34,6 +35,9 @@
 		"vendor-dir": "lib/vendor",
 		"platform": {
 			"php": "7.3.0"
+		},
```
ctgraham, line 38: This addition seems spurious to this feature.  I think it was probably backported from `main` inadvertently.

composer.lock
```diff
@@ -240,20 +255,105 @@
                 }
             ],
             "description": "jQuery UI is a curated set of user interface interactions, effects, widgets, and themes built on top of the jQuery JavaScript Library. Whether you're building highly interactive web applications or you just need to add a date picker to a form control, jQuery UI is the perfect choice.",
+            "support": {
+                "issues": "https://github.com/components/jqueryui/issues",
+                "source": "https://github.com/components/jqueryui/tree/master"
+            },
             "time": "2016-09-16T05:47:55+00:00"
         },
+        {
```
ctgraham, line 264: Unless all of these other `composer.lock` changes are introduced by adding the `composer/semver` dependency, this is the only appropriate change to this file.

ctgraham (changes requested):
Some questions and some requested changes.

classes/site/Version.inc.php
```diff
@@ -284,6 +285,47 @@ function getVersionString($numeric = true) {
 
 		return $numericVersion;
 	}
+
+	/**
+	 * Checks if this version is compatible with given version
+	 * Returns:
+	 * 		true iff the version given is compatible with this version
+	 * 		false iff the version given is incompatible with this version
+	 * @param $version string/Version the version to compare against
```
ctgraham, line 294: Is this intended as `string|Version`?  `VersionParser::parseConstraints()` expects a string, so specifying `Version` here seems suspect.

classes/site/Version.inc.php
Tyl13: It was aliased because it wasn't working without the alias originally. It now works without the alias, so I removed the alias.

classes/site/Version.inc.php
Tyl13: Superseded by the above.

composer.json
Tyl13: The allow-plugins is because when running `composer install`, it complains that `cweagans/composer-patches contains a Composer plugin which is currently not in your allow-plugins config. ` and I must have hit y.

classes/site/Version.inc.php
Tyl13: No, this should just be string. The versionParser is removed.

composer.json
Tyl13: Actually, if it isn't allowed then it won't build properly and will throw an error: 
`PHP Fatal error:  Cannot redeclare __() (previously
 declared in /var/www/ojs-stable-3_3_0/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php:893) in /var/www/ojs-
stable-3_3_0/lib/pkp/classes/i18n/PKPLocale.inc.php on line 921`

composer.lock
Tyl13: If the require is done before even installing it, then it does all the same updates and leaves my composer.lock the same as this version.

ctgraham: @asmecher , what is your opinion on PRs and dependency lockfile management?  Obviously, we want to insert a new dependency here, but I don't think we want to generate a new lockfile baseline with a bunch of coincidental updates.

classes/site/Version.inc.php
Tyl13: `Checks if the Version is compatible with the given version constraint string
formatted per composer/semver specifications; 
c.f. https://getcomposer.org/doc/articles/versions.md#writing-version-constraints` 
I liked the included link to the specifications.

classes/site/Version.inc.php
ctgraham: `@param $version string/Version` is still malformed.

composer.json
ctgraham: This was added in `main` here: https://github.com/pkp/pkp-lib/commit/8f9d32a8ed5157ebc8c467f03c9f6ad8940f14f2

And was discussed and affirmed in Slack, though not with respect to 3.3.

I'll allow it, though pending the broader conversation on lockfile dependency management.

classes/site/Version.inc.php
```diff
@@ -284,6 +284,23 @@ function getVersionString($numeric = true) {
...
+	 * Checks if the Version is compatible with the given string of constraints for the version, 
+	 * formatted per composer/semver specifications; 
+	 * c.f. https://getcomposer.org/doc/articles/versions.md#writing-version-constraints
+	 * Returns:
+	 * 		true iff the version given is compatible with this version
+	 * 		false iff the version given is incompatible with this version
+	 * @param $version string the string of constraints for the version to be checked against
+	 * @return boolean
+	 */
+	function isCompatible($version) {
```
ctgraham, line 298: Ok, this is pedantic, but can you rename the parameter to `$constraints`, so the current version can just be `$version` rather than `$systemVersion`, which is a bit of a misnomer?

ctgraham (changes requested):
Pedantry, sorry.

ctgraham:
One more question here, @Tyl13 , how will older application versions interact with the new version constraint strings when they are added to the plugin-gallery.xml?  If there is a risk of `version_compare` behaving badly with the new strings formatted against `composer/semver` standards, then we should also change the schema of the plugin-gallery.xml to look for requirements as either the existing `version` tag, or a new `semver` tag.  If there is no risk of overloading the `version` tag for prior application versions, then we can overload it as proposed.

Tyl13:
@ctgraham, are you talking about lines like this: https://github.com/pkp/pkp-lib/blob/acbcc6c51081510f0b6728afca6ba9a7cdb3b0a6/classes/plugins/PluginGalleryDAO.inc.php#L294? 
The version being specified here is the version given in the release on this line:
https://github.com/pkp/plugin-gallery/blob/811decce6e4fe1093d80868ba61431c395d7659f/plugins.xml#L13

If you are referring to something else, please let me know.

Tyl13:
Based on my testing, any of the additional characters in the constraints causes the version_compare to return a 1, meaning the second version is being viewed as lower than the first. So they just fail their comparison which seems like intended behavior for a this change. Changing it from version to <semver> will mean older versions don't look at it at all. So either option is usable.

ctgraham:
I think the key is here:
> If a part contains special version strings these are handled in the following order: `any string not found in this list < dev < alpha = a < beta = b < RC = rc < # < pl = p`.

https://www.php.net/manual/en/function.version-compare.php

This statement guarantees us a `!== 0`

ctgraham (approved):
@asmecher , I think we've covered all the bases here.  I support a squash-and-merge.  Are we still in the practice of the maintainer performing the submodule update in the applications, or should @Tyl13 submit application PRs to update the submodule ref?  @Tyl13 , when merged, please practice a cherry-pick PR to `main`.

PR #7832, opened 2022-04-06
pkp/pkp-lib#7651 Fatal error when trying to load PKPXMLParser

touhidurabir:
fix to issue pkp#7651 through non `PSR-4` standard `import` function .


touhidurabir:
closing this pull request and will open a new one which will have check and fix for all possible classes that has missed this `PKPXMLParser ` class import .

Issue #1234, opened 2016-03-07
[OJS] silly code left from refactoring


crism:
```php
if (($publishedArticle) && $publishedArticle ...
```

asmecher:
I can't help but read 1234 as a sample issue number.

crism:
@asmecher, I know—I did a double-take after submitting. Then I figured I’d won some kind of prize. `(-:`

PR #7833, opened 2022-04-06
pkp#7651 PKPXMLParser class missing namespace import

touhidurabir:
Issue pkp#7651 fixing through `PSR-4` standard class namespace import .

### DIFF
--- a/classes/navigationMenu/NavigationMenuItemDAO.inc.php
+++ b/classes/navigationMenu/NavigationMenuItemDAO.inc.php
@@ -20,6 +20,7 @@ namespace PKP\navigationMenu;
 use PKP\db\DAORegistry;
 use PKP\db\DAOResultFactory;
 use PKP\xml\XMLNode;
+use PKP\xml\PKPXMLParser;
 
 class NavigationMenuItemDAO extends \PKP\db\DAO
 {


asmecher:
Thanks, @touhidurabir! A few quick notes:
- We have a convention of titling pull requests with `pkp/pkp-lib#1234 Issue Title Here`; the `pkp/pkp-lib#1234` part is necessary because a pull request might go to a repository (e.g. `ojs`) that's not where the issue is tracked (always `pkp-lib`) and this helps keep everything interlinked.
- As you'll see, Travis tests don't get started here because the PR is to `pkp-lib` but tests run on the individual application repositories (`ojs`, `omp`, `ops`). To get Travis tests to start, also send in a PR to the OJS repository with a submodule update; see https://docs.pkp.sfu.ca/dev/testing/en/continuous-integration#prepare-pull-requests-for-testing for instructions.
- I think there are some other cases where `PKPXMLParser` is used in a class but not imported. For example, the issue mentions `lib/pkp/plugins/generic/acron/PKPAcronPlugin.inc.php:308` (from the error message). I'd suggest searching for any class that makes use of PKPXMLParser and doesn't have an `import` statement at the top of the class.
- The `main` branch already has a lot better adoption of namespacing and autoloading and I will merge some more big improvements there before 3.4.0 is released. But for the moment, just focus on `stable-3_3_0` and we'll forward-port anything necessary afterwards.

touhidurabir:
@asmecher Thanks for the notes and details . in this case I think it is better for me to close the pull request for the main branch . 

About the `lib/pkp/plugins/generic/acron/PKPAcronPlugin.inc.php` , it seems that someone has already added the namespace `PKP\xml\PKPXMLParser` for that file in `main` branch. That is way it's not included in my pull request . 

As you have mentioned that there is a big improvement to merge, I am going to close this pull request for `main` branch .

asmecher:
> I am going to close this pull request for `main` branch.

OK! And I see that the initial report talks about `main` too, so that's OK. But could you check if the reproduce steps cause an error on your `stable-3_3_0` branch? If so, it might be necessary to fix this in both places.

asmecher:
OK! I think the best approach is going to be to fix it in different ways in each branch. I think you've probably already figured this out but in case it's useful here's how I'd approach it...

- For `stable-3_3_0`:
  - Grep the codebase for uses of `PKPXMLParser`. I use:
    ```sh
    git ls-files | grep -e "\.php$" | xargs fgrep -l PKPXMLParser
    ```
    ...which turns up the following in the `pkp-lib` repo...
    ```
    classes/cliTool/ScheduledTaskTool.inc.php
    classes/controlledVocab/ControlledVocabDAO.inc.php
    classes/db/DBDataXMLParser.inc.php
    classes/db/SettingsDAO.inc.php
    classes/db/XMLDAO.inc.php
    classes/install/Installer.inc.php
    classes/navigationMenu/NavigationMenuDAO.inc.php
    classes/navigationMenu/NavigationMenuItemDAO.inc.php
    classes/plugins/PluginSettingsDAO.inc.php
    classes/security/UserGroupDAO.inc.php
    classes/xml/PKPXMLParser.inc.php
    classes/xml/XMLNode.inc.php
    classes/xml/XMLParserDOMHandler.inc.php
    plugins/generic/acron/PKPAcronPlugin.inc.php
    ```
    Many of these do already include `import('lib.pkp.classes.xml.PKPXMLParser');` at the top, but where they don't, it's safe to add it. Because almost nothing is namespaced in `stable-3_3_0`, if another piece of code that's already been executed does the `import` of a class it'll already be present -- which makes it hard to predict whether or not a class will be present sometimes. This will be a lot clearer once we've fully adopted namespaces (`main` adds a lot of this already).
- For `main`:
  - Do the same as above, but instead of the `import` statement, you should be able to add this where it doesn't already exist:
    ```php
    use PKP\xml\PKPXMLParser;
    ```

touhidurabir:
Ok. closing this pull request and will open a new one .

asmecher:
PRs for `stable-3_3_0`:
- https://github.com/pkp/ojs/pull/3367
- https://github.com/pkp/pkp-lib/pull/7843

asmecher:
All merged -- thanks, @touhidurabir! Closing.

Issue #7846, opened 2022-04-08
removeLocaleKey.php always uses app and pkp-lib locale directories


ewhanson:
**Describe the bug**
When a single path is given as a parameter to the `removeLocaleKey` script, the script always checks both the app directory and the pkp-lib directory and removes locale keys from both.

**To Reproduce**
1. Run the `removeLocaleKey` tool with a single path specified, e.g. `php lib/pkp/tools/removeLocaleKey test.locale.key locale`
2. See that locale keys will be removed from both the app and pkp-lib

**What application are you using?**
pkp-lib, main branch

**Additional information**
N/A

ewhanson:
@asmecher, could you have a look at this? It fixes a bug where locale keys would always be removed from both the app and pkp-lib. Thanks!

PR:

- #7847

PR #7847, opened 2022-04-08
pkp/pkp-lib#7846 Update removeLocaleKey to use provided directory path

ewhanson:


tools/removeLocaleKey.php
```diff
@@ -39,8 +39,8 @@ public function __construct($argv = [])
 
         $this->localeKey = array_shift($argv);
 
-        if (sizeof($this->argv) > 2) {
-            $this->dirs = $argv;
+        if (sizeof($argv) == 1) {
+            $this->dirs = [array_shift($argv)];
```
asmecher, line 43: I think this doesn't work if more than one path is specified. I think the only fix required is changing the `if` condition from
```php
if (sizeof($this->argv) > 2) {
    $this->dirs = $argv;
}
```
...to...
```php
if (!empty($this->argv)) {
    $this->dirs = $argv;
}
```

ewhanson: I see, that makes sense. Looking at it again, it looks like setting it to
```PHP
if (sizeof($this->argv) > 1) {
    $this->dirs = $argv;
}
```
would take care of things as the file path is removed in `$this-argv`, right?

asmecher: Sure, that's equivalent.

asmecher: Sorry, I misspoke -- `> 1` is still wrong. The `array_shift` already popped the key off `$argv` so all that should be left is the directory names if any were specified.

ewhanson: Right! Sorry to make this more confusing than it needs to be. Was getting a bit hung up on the class property `$argv` vs what was passed into the constructor. I've updated the PR.

asmecher: Not a problem! Merged, thanks.

PR #7853, opened 2022-04-11
pkp/pkp-lib#6209 Multiple use of id="setup-button" in website settings

touhidurabir:
fix to issue pkp/pkp-lib#6209 which eliminate the multiple use of duplicate `id="setup-button"` in website settings .

### DIFF
--- a/templates/management/website.tpl
+++ b/templates/management/website.tpl
@@ -30,7 +30,7 @@
 						@set="set"
 					/>
 				</tab>
-				<tab id="setup" label="{translate key="navigation.setup"}">
+				<tab id="appearance-setup" label="{translate key="navigation.setup"}">
 					<pkp-form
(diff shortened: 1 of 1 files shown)


templates/management/website.tpl
```diff
@@ -45,7 +45,7 @@
 				{call_hook name="Template::Settings::website::appearance"}
 			</tabs>
 		</tab>
-		<tab id="setup" label="{translate key="navigation.setup"}">
+		<tab id="configuration-setup" label="{translate key="navigation.setup"}">
```
asmecher, line 48: The label for this is "Setup", so I'd rather not swap that out for a synonym ("Configuration").

There are two "setup" tabs -- one a top-level tab:
image

...and another nested inside "appearances":
image

I'd suggest renaming the one inside "appearances" to `appearances-setup`.

PR #7875, opened 2022-04-18
[pkp-lib] #6999 Fixing directSalesPrice fixes OpenAccess selection for SubmissionFile

defstat:


asmecher:
Do this PR's changes actually affect behaviour, or are they just formatting? (Wanting to make sure I don't miss something subtle.)

Issue #6983, opened 2021-04-22
On password change, invalidate other sessions


asmecher:
When a user changes their password, it should invalidate any other open sessions from the user.

touhidurabir:
@asmecher 2 PR added targeting the `pkp-lib` and `ojs` `stable-3_3_0` branch . please review and if all ok and test passes properly , I will make PR then for other repos . 

[pkp-lib](https://github.com/pkp/pkp-lib) `stable-3_3_0` --> [pkp/pkp-lib#6983 On password change, invalidate other sessions](https://github.com/pkp/pkp-lib/pull/7877) 
[ojs](https://github.com/pkp/ojs) `stable-3_3_0` --> [pkp/pkp-lib#6983 Submodule Update ##touhidurabir/i6983_fix_stable_3_3_0##](https://github.com/pkp/ojs/pull/3378)

Also as I have noticed , this change can be pushed to `main` branch also . So in that case should I prepare the PRs also for the main branch for all the repos(`pkp-lib`, `ojs`, `ops`, `omp`) for this issue ?

asmecher:
Thanks, @touhidurabir! I've added comments here: https://github.com/pkp/pkp-lib/pull/7877

Once you've addressed those, I'd say it's ready for submodules for OMP and OPS, and forward-porting to `main` as well.

touhidurabir:
@asmecher I have made new pull request as I made a mess of the last one during the git rebase twice mistakenly and force push . 

New Push list 

[pkp-lib](https://github.com/pkp/pkp-lib) stable-3_3_0 --> [pkp/pkp-lib#6983 On password change, invalidate other sessions](https://github.com/pkp/pkp-lib/pull/7883)
[ojs](https://github.com/pkp/ojs) stable-3_3_0 --> [pkp/pkp-lib#6983 Submodule Update ##touhidurabir/i6983_fix_stable_3_3_0##](https://github.com/pkp/ojs/pull/3380)

This new pull request takes into the all password update case considerations as 

1. User password update .
2. User password update by admin/journal manager from journal settings
3. user password reset
4. user password force update on login as password updated or reset

Here is the last pull request reference https://github.com/pkp/pkp-lib/pull/7877 as that has some discussion going on . 

Once the test passes , please review and if all ok , then I will move forward with these changes to push to `ops` and `omp` and port to `main` branch .

asmecher:
[...]

PR #7877, opened 2022-04-19
pkp/pkp-lib#6983 On password change, invalidate other sessions

touhidurabir:
This RP aims to fix issue pkp/pkp-lib#6983 . By this enhancement , users will be force logged out from any other devices except the current one when update the password successfully .

### DIFF
--- a/classes/session/SessionDAO.inc.php
+++ b/classes/session/SessionDAO.inc.php
@@ -16,6 +16,7 @@
 
 
 import('lib.pkp.classes.session.Session');
+use Illuminate\Support\Facades\DB;
 
 class SessionDAO extends DAO {
 
@@ -169,6 +170,22 @@ class SessionDAO extends DAO {
 		$row = $result->current();
 		return $row ? (boolean) $row->row_count : false;
 	}
+
+	/**
+     * Delete given user's all sessions or except for the given session id
+     *
+     * @param int       			$userId
+     * @param mixed<string|null>    $sessionId
+     *
+     * @return void
+     */
+    public function deleteUserSessions(int $userId, string $excludableSessionId = null) {
+
+        DB::table('sessions')
+            ->where('user_id', $userId)
+            ->when($excludableSessionId, fn($query) => $query->where('session_id', '<>', $excludableSessionId))
+            ->delete();
+    }
 }
 
 
--- a/controllers/tab/user/ProfileTabHandler.inc.php
+++ b/controllers/tab/user/ProfileTabHandler.inc.php
@@ -20,6 +20,7 @@
 
 import('classes.handler.Handler');
 import('lib.pkp.classes.core.JSONMessage');
+import('lib.pkp.classes.session.SessionManager');
 
 class ProfileTabHandler extends Handler {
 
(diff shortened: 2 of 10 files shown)


classes/session/SessionDAO.inc.php
```diff
@@ -169,6 +169,22 @@ function sessionExistsById($sessionId) {
...
+     *
+     * @param int       $userId
+     * @param string    $sessionId
+     *
+     * @return void
+     */
+    public function deleteAllUserSessionsExceptGivenId($userId, $sessionId) {
+
+        $this->update(
+            'DELETE FROM sessions WHERE user_id = ? AND session_id != ?',
```
asmecher, line 184: I see you've just followed existing patterns here, but this is an aspect where we're adopting Laravel's query builder toolset as we write/rewrite SQL queries. Eventually we'd like to remove any old-style `DAO::update` calls, which used to be executed using ADODB and are now around for backwards compatibility. (See the deprecations in `lib/pkp/classes/db/DAO.inc.php`.)

(For the `main` branch, we're also adding type hinting to functions where possible.)

Note that `!=` in SQL is probably not going to work with PostgreSQL (thought I haven't checked). The "most" standard SQL operator is `<>`.

controllers/tab/user/ProfileTabHandler.inc.php
```diff
@@ -262,18 +263,26 @@ function changePassword($args, $request) {
...
+
 		import('lib.pkp.classes.user.form.ChangePasswordForm');
 		$passwordForm = new ChangePasswordForm($request->getUser(), $request->getSite());
 		$passwordForm->readInputData();
 
 		if ($passwordForm->validate()) {
+
 			$passwordForm->execute();
+
+			SessionManager::logoutFromOtherDevices();
```
asmecher, line 277: This is good for the user profile tab's password edit tool, but I think you'll also have to add calls to this function everywhere else the password can be reset -- in the Journal Manager's administration interface where they can edit a password, and also in the password reset process, off the top of my head. I'd suggest looking for uses of:
- lib/pkp/classes/user/form/LoginChangePasswordForm.inc.php
- lib/pkp/controllers/grid/settings/user/form/UserDetailsForm.inc.php

asmecher (changes requested):
Thanks, @touhidurabir! A couple of comments...

classes/session/SessionDAO.inc.php
touhidurabir: Yes, I was trying to follow the same code pattern so that it does not get polluted with different code style and adaptations . But I agree using the Laravel style is much better option as in that case we do not have to worry about the SQL compatibility for different DB drivers . I will update it to Laravel style `Query Builder` format as I can see `Eloquent` has not yet adopted into the system .

touhidurabir: about the `!=` and `<>`, as I have checked, up to `PostgreSQL 14` (which is the latest one) has support for both (https://www.postgresql.org/docs/14/functions-comparison.html) .

controllers/tab/user/ProfileTabHandler.inc.php
touhidurabir: @asmecher I yes you are right and I missed few more section . But as I dig deeper into this issue based on you feedback , I think we need a way not only to invalidate other device sessions but also a method to invalidate all device sessions . Such use cases are

1. when from the journal management , an admin update another user's password , that user need to be logged out from all devices including the one that he is using right now . 

2. Another case is when user reset password from a device as that regenerate a password from the reset link comparing to the more modern and adaptive approach where user can set the password as user like . 

so I guess we need method like that can force logout other user . 

```php
SessionManager::logoutFromAllDevices(USER_ID); 
```

asmecher: I would suggest a single method to achieve both `logoutFromAllDevices` and `logoutFromOtherDevices`, with an optional session ID to exclude if it's provided. Something like e.g. `invalidateSessions(int $userId, string $excludeSessionId = null)`.

touhidurabir: @asmecher that's also an option and probably the better one . I thought about that but wanted to go with more of a Laravel like implementation where the method name and implementation itself is self explanatory . But having one single method to handle it better for future refactoring and enhancement . Will go with single method approach .

asmecher: I would happily merge either approach; this is stylistic. But personally I prefer to avoid multiplying out the number of methods, especially when both cases can be accommodated easily with a conditional clause (untested):
```php
DB::table('sessions')
   ->where('user_id', '=', $userId)
   ->when($excludeSessionId !== null, fn($query) => $query->where('session_id', '<>', $excludeSessionId))
   ->delete();
```
A lot of my stylistic preferences are shaped by trying to keep the codebase as small as possible over the years.

touhidurabir:
@asmecher I mistaken did git rebase fixup twice and as I have forced pushed , it got merge with other commit . I am closing this one will open a new pull request .

PR #7886, opened 2022-04-22
pkp/pkp-lib#6983 On password change, invalidate other sessions

touhidurabir:
This RP aims to fix issue https://github.com/pkp/pkp-lib/issues/6983 . By this enhancement , users will be force logged out from any other devices except the current one when update the password successfully . This takes into the consideration of cases as

1. User password update .
2. User password update by admin/journal manager from journal settings
3. user password reset
4. user password force update on login as password updated or reset

### DIFF
--- a/classes/session/SessionDAO.inc.php
+++ b/classes/session/SessionDAO.inc.php
@@ -203,7 +203,7 @@ class SessionDAO extends DAO
      * @return void
      */
     public function deleteUserSessions(int $userId, string $excludableSessionId = null) 
-	{
+    {
         DB::table('sessions')
             ->where('user_id', $userId)
             ->when($excludableSessionId, fn($query) => $query->where('session_id', '<>', $excludableSessionId))
--- a/classes/session/SessionDAO.inc.php
+++ b/classes/session/SessionDAO.inc.php
@@ -18,6 +18,7 @@
 namespace PKP\session;
 
 use PKP\db\DAO;
+use Illuminate\Support\Facades\DB;
 
 class SessionDAO extends DAO
 {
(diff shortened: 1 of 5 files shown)


classes/session/SessionDAO.inc.php
```diff
@@ -192,6 +193,22 @@ public function sessionExistsById($sessionId)
         $row = $result->current();
         return $row ? (bool) $row->row_count : false;
     }
+
+	/**
+     * Delete given user's all sessions or except for the given session id
+     *
+     * @param int                   $userId
```
asmecher, line 200: If the annotations are only reiterating the type hints, better to leave them out. If you want to include parameter descriptions (and I recommend it), unfortunately it's currently necessary to restate the type. (Ref for `@param` PHPDoc tag: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#510-param)

classes/session/SessionManager.inc.php
```diff
@@ -140,6 +142,31 @@ public static function getManager(): SessionManager
...
+     */
+    public function invalidateSessions(int $userId, string $excludableSessionId = null): bool 
+	{
+        $this->getSessionDao()->deleteUserSessions($userId, $excludableSessionId);
+
+        return true;
+    }
+
+	/**
+     * Get the Session DAO instance assocaited with the current request
```
asmecher, line 161: *associated

controllers/grid/settings/user/form/UserDetailsForm.inc.php
```diff
@@ -319,6 +320,14 @@ public function execute(...$functionParams)
                 } else {
                     $this->user->setPassword(Validation::encryptCredentials($this->user->getUsername(), $this->getData('password')));
                 }
+
+				$sessionManager = SessionManager::getManager();
```
asmecher, line 324: Looks like you've got a mix of tabs and spaces here; for `main`, indenting should be 4 spaces. (The auto-formatter should be taking care of this; maybe Captain Hook is not working to set up your git hooks?)

asmecher (changes requested):
Just a couple of small comments, @touhidurabir, thanks!

controllers/tab/user/ProfileTabHandler.inc.php
```diff
@@ -300,15 +301,25 @@ public function changePassword($args, $request)
...
+
+			$sessionManager = SessionManager::getManager();
+            $sessionManager->invalidateSessions($user->getId(), $sessionManager->getUserSession()->getId());
+
             $notificationMgr = new NotificationManager();
-            $notificationMgr->createTrivialNotification($request->getUser()->getId());
-            return new JSONMessage(true);
+            $notificationMgr->createTrivialNotification($user->getId());
+            
+			return new JSONMessage(true);
```
asmecher, line 320: There are still some tab-based indents here; there should be pre-commit git hooks that run the code formatter over the code (using http://captainhook.info). Could you check into whether that's working?

touhidurabir: fixed it for this file and other files . should be fine now .

Issue #7909, opened 2022-05-08
Can not accept a submission if associated review has any review form attached to it.


touhidurabir:
**Describe the bug**
At `Accept Submission`, if the associated review has any review form attached to it , it throws internal exception with 500 error code.

**To Reproduce**
1. Create a Submission.
2. Assign a reviewer and make sure it has a `Review Form` associated with it.
3. Complete the review process.
4. Try to `Accept Submission` and that will present a 500 error page . 

**What application are you using?**
 OJS installation of `main` branch .

touhidurabir:
### Exception Tracing 

Exception originated from `PKP\mail\traits\ReviewerComments` as defined 

```php
if ($reviewFormElements->wasEmpty()) {
     return '';
}
```

Which delegate the call to `PKP\reviewForm\ReviewFormElementDAO` in the method `getByReviewFormId` as defined 

```php
...
return new DAOResultFactory($result, $this, '_fromRow');
```

which delegate to `PKP\db\DAOResultFactory` at method `getCount` where as no $sql is passed , so this line get executed as

```php
if ($this->sql === null) {
    throw new \Exception('DAOResultFactory instances cannot be counted unless supplied in constructor (DAO ' . get_class($this->dao) . ')!');
}
```

touhidurabir:
@asmecher PRs. once reviewed, I will see if I need to forward the changes to `OPS` and `OMP`.

[pkp-lib](https://github.com/pkp/pkp-lib) --> [pkp/pkp-lib#7909 Can not accept a submission if associated review has any review form attached to it](https://github.com/pkp/pkp-lib/pull/7913)
[ojs](https://github.com/pkp/ojs) --> [pkp/pkp-lib#7909 Submodule Update ##touhidurabir/i7909_fix_main##](https://github.com/pkp/ojs/pull/3394)

asmecher:
Thanks, @touhidurabir! See this comment: https://github.com/pkp/pkp-lib/pull/7913#discussion_r868192045

touhidurabir:
@asmecher changes are pushed . please review now .

asmecher:
Merged -- thanks, @touhidurabir!

PR #7913, opened 2022-05-08
pkp/pkp-lib#7909 Can not accept a submission if associated review has any review form attached to it

touhidurabir:
This PR aims to solve pkp/pkp-lib#7909 .

### DIFF
--- a/classes/reviewForm/ReviewFormElementDAO.inc.php
+++ b/classes/reviewForm/ReviewFormElementDAO.inc.php
@@ -242,7 +242,7 @@ class ReviewFormElementDAO extends \PKP\db\DAO
             $rangeInfo
         );
 
-        return new DAOResultFactory($result, $this, '_fromRow');
+        return new DAOResultFactory($result, $this, '_fromRow', [], $this->getSql(), [(int) $reviewFormId]);
     }
 
     /**
--- a/classes/reviewForm/ReviewFormElementDAO.inc.php
+++ b/classes/reviewForm/ReviewFormElementDAO.inc.php
@@ -232,17 +232,17 @@ class ReviewFormElementDAO extends \PKP\db\DAO
     public function getByReviewFormId($reviewFormId, $rangeInfo = null, $included = null)
     {
         $result = $this->retrieveRange(
-            'SELECT *
-			FROM review_form_elements
-			WHERE review_form_id = ?
-			' . ($included === true ? ' AND included = 1' : '') . '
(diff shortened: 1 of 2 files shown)


classes/reviewForm/ReviewFormElementDAO.inc.php
```diff
@@ -242,7 +242,7 @@ public function getByReviewFormId($reviewFormId, $rangeInfo = null, $included =
             $rangeInfo
         );
 
-        return new DAOResultFactory($result, $this, '_fromRow');
+        return new DAOResultFactory($result, $this, '_fromRow', [], $this->getSql(), [(int) $reviewFormId]);
```
asmecher, line 245: The `$sql` and `$params` parameters to `DAOResultFactory::__construct` are sunsetted hacks for places where the code previously relied on ADODB's result counting tools and we haven't been able to remove/rewrite reliant code yet. I think you can keep this deprecated aspect localized to just this class by doing this instead:
```sql
diff --git a/classes/reviewForm/ReviewFormElementDAO.inc.php b/classes/reviewForm/ReviewFormElementDAO.inc.php
index 9343d0a73..50314140a 100644
--- a/classes/reviewForm/ReviewFormElementDAO.inc.php
+++ b/classes/reviewForm/ReviewFormElementDAO.inc.php
@@ -232,17 +232,17 @@ class ReviewFormElementDAO extends \PKP\db\DAO
     public function getByReviewFormId($reviewFormId, $rangeInfo = null, $included = null)
     {
         $result = $this->retrieveRange(
-            'SELECT *
+            $sql = 'SELECT *
                        FROM review_form_elements
                        WHERE review_form_id = ?
                        ' . ($included === true ? ' AND included = 1' : '') . '
                        ' . ($included === false ? ' AND included = 0' : '') . '
                        ORDER BY seq',
-            [(int) $reviewFormId],
+            $params = [(int) $reviewFormId],
             $rangeInfo
         );
 
-        return new DAOResultFactory($result, $this, '_fromRow');
+        return new DAOResultFactory($result, $this, '_fromRow', [], $sql, $params);
     }
 
     /**
```

touhidurabir: @asmecher yeah thats another way to do this but I thought may be we need to have a way get the generated sql for this functionality in other places also as even if this is deprecated , we can not get rid of it completely right now . But as the same time right now I don't know where we may need this approach again that I have implemented so changing the implementation as you have mentioned .

PR #7915, opened 2022-05-09
greek

sioulasdimitris:
greek locale files


asmecher:
@sioulasdimitris, I have 3 large PRs for content in the `stable-3_3_0` branches:
- This one
- https://github.com/pkp/ojs/pull/3396/files
- https://github.com/pkp/usageStats/pull/53

Currently we are accepting upstream translations to the `main` branch, which will be included in the 3.4.0 release in Q3 this year. However, there have been significant changes between `main` and `stable-3_3_0`, so we can't easily apply these changes to `main`.

I think there are 3 options:
1. Merge these changes to `stable-3_3_0` and do not forward-port them to `main`. This will mean the content will have to be re-translated at some point for 3.4.0 and newer.
2. Merge these changes to `stable-3_3_0` and force them forward into `main`. This will probably destroy some already-completed translation work in `main` (there are lots [like this](https://github.com/pkp/pkp-lib/commit/e350e57b9cbafd86a9606a20848693494f32b6f9#diff-50d79502581831532767fa629918cf158b1bd27b92d33938e668f9161ce3968e), and will result in changes between 3.3.x and 3.4.x that were already applied being lost (there are lots like [this](https://github.com/pkp/pkp-lib/commit/ed9eb6c0a873931cf2caf6dfe3a691185e6ade05))
3. Manually bring these translations forward into `main` e.g. using Weblate.

To avoid these kinds of problems I recommend using Weblate (https://translate.pkp.sfu.ca) -- we always have those point to the branch that we're actively managing, which allows us to plan for translation management for the longer term.

sioulasdimitris:
@asmecher  I am going with option number 2, I am trying to merge these changes to stable-3_3_0.
is it ok? can you apply the merge? thank you

asmecher:
Hi @sioulasdimitris! I have merged the 3 PRs into `stable-3-3_0`, but I think it would be a mistake to force them forward to `main`. Unfortunately this means that the work will need to be manually forward-ported to `main` or else users will experience a decrease in translation completeness/quality when the upgrade to 3.4 (after that is released).

We do need to figure out some policy for how to manage cases like these and reduce the amount of repeat translation work that is required. I'll see if we can tackle that as a subject during our upcoming Colombia sprint.

Issue #7930, opened 2022-05-17
Ensure umask is respected in file creation


asmecher:
**Describe the bug**
From https://forum.pkp.sfu.ca/t/ojs-v3-3-0-10-some-submitted-files-in-files-dir-journals-are-world-writable/72969:
> Our nightly system check discovered world-writable files in $files_dir/journals, namely:
>
>    4/articles/337
>    4/articles/339
>    4/articles/336
>    4/articles/342
>    4/articles/338

> I checked the umask setting in config.inc.php, but it is “0022”.

**What application are you using?**
 OJS, OMP or OPS version 3.3.0-x (reproduced on 3.3.0-10)

asmecher:
@bozana, could you review this?

bozana:
@asmecher, just one comment/question, else all good... :+1:

asmecher:
Thanks -- merged!

PR #7931, opened 2022-05-17
pkp/pkp-lib#7930 Apply umask to file/directory creation (stable-3_3_0)

asmecher:


classes/services/PKPFileService.inc.php
```diff
@@ -31,19 +31,19 @@ class PKPFileService {
 	 * Initialize and configure flysystem
 	 */
 	public function __construct() {
-
+		$umask = Config::getVar('files', 'umask');
```
bozana, line 34: what if this var is commented out in the config file? do we have to consider that case at all?

asmecher: Good point -- I made the default 0022 apply if the umask isn't specified.

Issue #6610, opened 2021-01-15
[OMP] Make ONIX export problems easier to understand/solve


asmecher:
**Describe the bug**
When there's a problem with the ONIX export, i.e. not enough data to properly populate an XML document that will pass the validation rules, the user gets a `Could not convert submissions.` fatal error and nothing more.

**To Reproduce**
1. Go to the ONIX export plugin
2. Try to export an incomplete submission (one that does not have a publication format, for example)
3. See error

The error can be determined by dumping `$errors` in `lib/pkp/classes/xslt/XMLTypeDescription.inc.php` around line 135, but isn't available via the UI.

**What application are you using?**
OMP pre-3.3

asmecher:
@defstat, I'm assigning this to you because you're working on some related error handling cleanup for import/export tools, but this isn't a priority for 3.3.0. (It's possible you're already working on fixing this -- I don't recall if the ONIX plugin made an appearance in the OMP port of your error handling cleanup.)

defstat:
@asmecher I have extended the #6490 functionality in order for other plugins to have access to the native import/export changes. I also added the functionality to ONIX plugin, which, when reviewed, could resolve this issue.

defstat:
@asmecher 

I added support for displaying the export results. 

All the changes are contained within the ONIX3.0 plugin. The code for main branch comes next. Cherry picking is not possible between the two right now.

**PR**
OMP-stable3.3.0: pkp/omp#1126

defstat:
@asmecher 

New PRs added.

**PRs**
pkp-lib-stable3.3.0: pkp/pkp-lib#7942
OMP-stable3.3.0: pkp/omp#1126

defstat:
@asmecher

Changed the code to pass only the date part of the file in order to determine the file that will be downloaded.

**PRs**
pkp-lib-stable3.3.0: https://github.com/pkp/pkp-lib/pull/7942
OMP-stable3.3.0: https://github.com/pkp/omp/pull/1126

defstat:
@asmecher I resolved the PR review comments. I added the current PRs in the [top ](https://github.com/pkp/pkp-lib/issues/6610#issue-786485868) comment of the current issue.

asmecher:
Looks good, thanks, @defstat. Can you forward-port this to `main`?

defstat:
[...]

PR #7942, opened 2022-05-24
pkp/pkp-lib#6610 Add support for checking a plugin's exported file

defstat:


classes/plugins/ImportExportPlugin.inc.php
```diff
@@ -144,10 +144,20 @@ function getExportPath() {
 	 * @param $objectsFileNamePart string Part different for each object type.
 	 * @param $context Context
 	 * @param $extension string
+	 * @param $dateFilenamePart string The datetime part of the exported filename
 	 * @return string
 	 */
-	function getExportFileName($basePath, $objectsFileNamePart, $context, $extension = '.xml') {
-		return $basePath . $this->getPluginSettingsPrefix() . '-' . date('Ymd-His') .'-' . $objectsFileNamePart .'-' . $context->getId() . $extension;
+	function getExportFileName($basePath, $objectsFileNamePart, $context, $extension = '.xml', $dateFilenamePart = null, $dateFilenamePartFormat = 'Ymd-His') {
```
asmecher, line 150: @defstat, I think this is going to break several functions that make use of `getExportFileName` in different import/export plugins; can you try making these improvements without modifying the function signature?

defstat: @asmecher I can easily remove the second additional parameter but I need to know the `$dateFilenamePart` of the filename, so that I can pass it to the form. I could not think a better way to do that, than creating it outside the `getExportFileName`  and passing it as a parameter. All the functions that are using it already, will use the default value, which is null, so the code will again create the same date part as before. 

I could, probably, extract the part of the filename that corresponds to the datetime (for example take the string between the first `-` character and the third `-` character) - but then I have the problem of recreating the filename path (without copying the `getExportFileName` code to the plugin)

Adding the name in the parameters list serves both purposes.

If you have another idea for that I can try it.

Issue #8007, opened 2022-06-16
The ACron plugin isn't reloading the scheduled tasks when a plugin is enabled/disabled


jonasraoni:
**Describe the bug**
After enabling/disabling a plugin, it should have the chance to install its scheduled tasks, but its `AcronPlugin::parseCronTab` hook won't be called.

This feature depends on a hook that doesn't exist anymore, restoring it/replacing with an event should be enough to fix.
The hook was removed at this commit: https://github.com/pkp/pkp-lib/commit/e678078cf1717bd9785b446f84037eeda969fa9d#diff-f5e64b38361922e91f9ae224817e9e76e30f352086149a0e433d278ac28a8c15L246

**To Reproduce**
1. Enable the acron plugin.
2. Enable any other plugin which has scheduled tasks.
3. Check if the scheduled task was installed.

**PRs**
- `stable-3_3_0`
- https://github.com/pkp/pkp-lib/pull/8008
- https://github.com/pkp/ojs/pull/3433

- `main`
- https://github.com/pkp/pkp-lib/pull/8009
- https://github.com/pkp/ojs/pull/3434

**What application are you using?**
OJS 3.3

jonasraoni:
@asmecher I found this issue while inspecting the PKP PN plugin.

asmecher:
@jonasraoni, sorry for the wait on this. I've merged the `stable-3_3_0` changes, but had a comment/question on the `main` one; was this a manual syntax change or something the autoformatter did? Also, this will unfortunately need a rebase!

jonasraoni:
@asmecher I've left a comment and rebased.

asmecher:
Merged, thanks, @jonasraoni. Tagging @touhidurabir -- this might add a few more conflicting lines on #7105 due to reformatting.

PR #8008, opened 2022-06-16
pkp/pkp-lib#8007 Ensured the AcronPlugin is notified when another plugin is enabled/disabled.

jonasraoni:


classes/plugins/LazyLoadPlugin.inc.php
```diff
@@ -78,6 +78,7 @@ function setEnabled($enabled) {
 			$contextId = 0;
 		}
 		$this->updateSetting($contextId, 'enabled', $enabled, 'bool');
+		HookRegistry::call('Plugin::setEnabled', [$this, $enabled]);
```
asmecher, line 81: There are some important plugins that don't call `parent::setEnabled` in their own implementations of `setEnabled` -- the ORCID profile plugin and anything extending `BlockPlugin` or `GatewayPlugin`, for example. There are also other uses of `setEnabled` that you might not expect (https://github.com/pkp/customBlockManager/blob/a758bdc116df01e3ac5c2202b4d19bbca3f72e10/controllers/grid/form/CustomBlockForm.inc.php#L112). So I think especially for `stable-3_3_0` it would be beneficial to rely on the `enable` / `disable` grid actions rather than the `Plugin::setEnabled` call.

jonasraoni: Hmm, really a surprise 😁

jonasraoni: @asmecher Ok, I've updated this PR to recover the hook.

PR #8009, opened 2022-06-16
Bugfix/main/8007 fix acron plugin

jonasraoni:


classes/observers/events/PluginEnabledChanged.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @brief Event fired when a plugin is enabled/disabled
+ */
+
+namespace PKP\observers\events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use PKP\plugins\LazyLoadPlugin;
+
+class PluginEnabledChanged
```
jonasraoni, line 23: I've created a hook for `stable-3_3_0`, but here I moved to an event and created this discussion: https://github.com/pkp/pkp-lib/discussions/8010

plugins/generic/acron/PKPAcronPlugin.inc.php
```diff
@@ -31,38 +36,34 @@
...
      * @copydoc Plugin::register()
      *
      * @param null|mixed $mainContextId
      */
-    public function register($category, $path, $mainContextId = null)
+    public function register($category, $path, $mainContextId = null): bool
     {
         $success = parent::register($category, $path, $mainContextId);
-        HookRegistry::register('Installer::postInstall', [&$this, 'callbackPostInstall']);
+        HookRegistry::register('Installer::postInstall', fn () => $this->_callbackPostInstall(...func_get_args()));
```
asmecher, line 49: Hmm, this syntax feels less friendly to me than the array callback function, and it wraps the actual callback in an extra function call. Are there advantages to this?

jonasraoni: Ah, it's mostly personal bias against using strings for things that should be typed. PHP should allows us to reference first-class functions without using strings.

Benefits: a code editor is able to show invalid method calls (e.g. inexistent method, bad arguments, non-public), it must also make the life of code analyzers easier.

asmecher: We'll get better options for [first class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) when we require a baseline PHP 8.1:
```php
HookRegistry::register('Installer::postInstall', $this->callbackPostInstall(...));
```
I'd suggest we wait for that; I agree that having code completion is much better than the old array construction but the semantics of another wrapper function and not-totally-obvious `...func_get_args()` invocation muddy the waters.

jonasraoni: Agreed, let's wait for 8.1 😁
I just updated the PR again.

Issue #6895, opened 2021-03-30
IP location and institution service


bozana:
- [x] Implement IP location solution
- [x] Implement IP institution solution

This issue discusses and defines the solutions that can be used to detect user location and institution, so that those can be used in usage statistics.

**IP location solution:**
We will use the free DB-IP lite city database (e.g. https://download.db-ip.com/free/dbip-city-lite-2021-05.mmdb.gz) and the MaxMind library https://github.com/maxmind/GeoIP2-php per default. 
The library will be provided via composer, i.e. together with our code.
The Geo data and granularity (country and city+region) option for usage stats would need to be set up on the site level, and can be opted-out on the journal level.
If Geo data is selected on the site level, the DB-IP lite city database will be downloaded (in the files/usageStats/ folder) and a monthly scheduled task provided for the DB update.
The IP to Geo data processing happens before the usage event is logged. Then, the anonymized IP address and the Geo data will be logged (and cached).
To anonymize the IP, the IP is encoded using a randomly created salt that changes every day. When salt changes, the usage event log file as well as IP-Geo data cache change.
The country and region iso code and city name will be saved. S. https://github.com/pkp/pkp-lib/issues/6782#issuecomment-853760292.

**IP institution solution:**
We will add new DB tables `institutions` and `institution_settings` containing the id, ROR, and localized names.
The DB table `institutional_subscription_ip` will be renamed into `institution_ip` table so it can be used more widely. Instead of having `subscription_id` it will contain the `institution_id` column. The table `institutional_subscriptions` will contain a new column `institution_id`.
The `institution_id` will be logged.

_Research:_
**IP location data:**
It seems that finding out the Geo data is not an easy task at all, so mostly the services are not for free. Thus, if someone would like to have a 'perfect' (more accurate and extensive) service, then they could pay for it.

API solutions (GDPR problem): 
Most of the solutions provide an API, that a registered user can use for free up until ca. 10.000 - 50.000 requests per month, or for a price. But we cannot use/consider those APIs because we would forward the user's IP address to them, which would require a user consent.
Databases:
MaxMind data (that we were using too) are pretty much used (e.g. there seems to exist plugins for WordPress, Drupal,...) and apparently well. There are also others, e.g. IP2Location, DB-IP... 
For registered users MaxMind and DB-IP provide a free download of their lite database (that is not so accurate), or a full database download for a price. IP2Location provides a commercial download.
Examples:
[...]

Issue #6781, opened 2021-02-22
COUNTER Release 5


bozana:
Implement the COUNTER Release 5 for OJS/OMP/OPS usage statistics.
Here we can collect everything we decide is necessary. We can have a discussion below and every time we decide something we can summarize it here.

It seems the Release 5 with lots of changes is out there. 
Here a guide for journals: https://www.projectcounter.org/wp-content/uploads/2020/08/Module_2_Journal_Usage_20200811.pdf.

1. Processing rules for COUNTER 5 reports, s. https://www.projectcounter.org/code-of-practice-five-sections/7-processing-rules-underlying-counter-reporting-data/:
a) Double click filtering (s. section 7.2):
This is implemented here: https://github.com/pkp/usageStats/blob/master/UsageStatsLoader.inc.php#L194-L224. Till know the differentiation was between the access of HTML, PDF and other. This seems not to be needed any more -- We can change it to consider 30 seconds for any link i.e. file.
b) Unique Items (s. section 7.3):
In our case Item is an article. The matching report is AR1. And the rule is: "If multiple transactions qualifying for the Metric_Type in question represent the same item and occur in the same user-sessions, only one unique activity MUST be counted for that item." Where user-session seems to be defined for an hour, as far as I understand it.
The question if the article versions do belong to the same Item is still open. Due to the way we represent them internally I would say they do belong to the same Item.
c) Unique Titles (s. section 7.4):
In the case of a journal Title = a journal and the report = Title Master Report. Similar to the rule for the unique item above, the rule here is: "If multiple transactions qualifying for the Metric_Type in question represent the same title and occur in the same user-session only one unique activity MUST be counted for that title.". Where the user-session seems to be defined for an hour. I.e. here, if a user accesses one article and then another in the same session, it would only count once.
This rule i.e. report seems not to be used for single journals -- introduced mostly for books. Do we need it (e.g. for libraries and multi-journal installations)?
d) Internet Robots and Crawlers (s. section 7.8):
Same as for Release 4.
COUNTER maintains the current list of internet robots and crawlers at https://github.com/atmire/COUNTER-Robots.
[...]

PR #8050, opened 2022-06-29
6782: usage stats improvements

bozana:
This PR considers the following issues:
https://github.com/pkp/pkp-lib/issues/6782
https://github.com/pkp/pkp-lib/issues/6895
https://github.com/pkp/pkp-lib/issues/6781
https://github.com/pkp/pkp-lib/issues/7318
https://github.com/pkp/pkp-lib/issues/5543

It relates to the appropriate applications (OJS, OMP, OPS) PRs.
The PRs and commits explanations will only be provided here below and not in the applications PRs.

This PR considers:

- institutions management (used in subscriptions and usage stats)
- institutional usage stats as defined in COUNTER R5, i.e.: considering institutions in usage event log files, saving institutional usage on submission level in a db table, providing them in COUNTER R5 SUSHI API/reports
- Geo usage, i. e.: using DB-IP city light database per default + geo2Ip API to get the Geo data from IP, considering Geo data in log files, saving Geo usage on submission level in a db table, providing it via Geo stats API
- usage event log file format change
- Laravel events and listeners for usage events and logging
- log file and usage stats data processing based on COUNTER R5, using a combination of a scheduled task and Laravel jobs
- new metrics tables that separate the usage for assoc types, containing the same information as the old db table metrics
- stats API (JSON and relevant CSV responses) for different assoc types (not only submissions, as till now, but also contexts, issues, and files)
- considers supp files in submissions stats API
- COUNTER R5 SUSHI API and relevant reports
- removal of usageStats plugin (also usage stats report plugin) , report generator, views report plugin, and MetricsDAO (the way the stats data was retrieved)
- adaptation of all the code in main repositories that uses stats
- new usage stats settings (site + context level)
- usage stats display settings ThemePlugin and in default theme
- all install and upgrade scripts
- review comments from the first review round (s. https://github.com/pkp/pkp-lib/issues/6782#issuecomment-1168557059)
 
This PR does not consider, and still to come:

- tool for the migration of the old log file format to the new one, considering also the apache log files, reprocessing of those old log files + the possibility to consider them for COUNTER R5 reports (3.4 release)
- usage stats display settings in all PKP themes (3.4 release)
- UI for the new stats API functions (contexts, issues, Geo, files) and CSV reports (not necessarily 3.4)
- UI and CSV COUNTER R5 reports (not necessarily 3.4)
- Tests for statistics (as soon as possible)


asmecher (commented):
@bozana, just a few initial comments while I figure out how this all fits together!

classes/task/FileLoader.inc.php
```diff
@@ -274,29 +235,73 @@ public function checkFolderStructure($install = false)
     //
     // Protected methods.
     //
+    /**
+     * @copydoc ScheduledTask::executeActions()
+     */
+    protected function executeActions(): bool
```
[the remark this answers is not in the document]
bozana: The plugins/generic/usageStats should not exist any more -- it is deleted...
Hmmm... 
How did you get my code/branch -- it seems your folder is still there...
Hmmm...
Shell I then however consider this: s. https://github.com/pkp/ojs/pull/3379/commits/d16663cf0f1ce92855c0f94eea91c8266cc7de31#r885089057 ?

classes/observers/listeners/UsageEventLog.inc.php
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        $cachedGeoData = $this->geoDataCache->getContents();
+        if (array_key_exists($hashedIp, $cachedGeoData)) {
+            return $cachedGeoData[$hashedIp];
+        }
+
+        $reader = $countryIsoCode = $regionIsoCode = $cityName = null;
+        try {
+            $reader = new Reader($this->getGeoDBPath());
```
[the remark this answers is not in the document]
bozana: I added it to the composer, but have not changed the lock file yet -- I wanted to do that at the end. It should be required by default.
So if you remove the lock file and do `composer --working-dir=lib/pkp install` it should work fine.
Sorry that I have forgotten to tell you that! :-(

classes/task/FileLoader.inc.php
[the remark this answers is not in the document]
asmecher: I suspected as much :) I just checked out your branches, but that doesn't remove the files. I'll get rid of it manually. For end users, this won't be an issue.

classes/observers/listeners/UsageEventLog.inc.php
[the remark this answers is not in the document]
asmecher: Not a problem, thanks!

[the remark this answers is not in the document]
asmecher: (Off the top of my head, `composer install` won't be enough -- it depends on the lockfile, so wouldn't install a new dependency that didn't appear there.)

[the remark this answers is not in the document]
bozana: yes, you would need to remove the lock file -- before merging I will change the lock file too..

classes/context/Context.inc.php
```diff
@@ -514,99 +515,36 @@ public function updateSetting($name, $value, $type = null, $isLocalized = false)
...
-    * @return null|array The selected data as a simple tabular
-    *  result set or null if metrics are not supported by this context.
-    */
-    public function getMetrics($metricType = null, $columns = [], $filter = [], $orderBy = [], $range = null)
-    {
-        // Add a context filter and run the report.
-        $filter[PKPStatisticsHelper::STATISTICS_DIMENSION_CONTEXT_ID] = $this->getId();
-        $application = Application::get();
-        return $application->getMetrics($metricType, $columns, $filter, $orderBy, $range);
+        return $enableInstitutionUsageStats;
```
asmecher, line 547: This will need a `(bool)` cast -- installations that upgraded from previous versions will not have this setting present, which will cause:
```
PHP Fatal error:  Uncaught TypeError: PKP\context\Context::isInstitutionStatsEnabled(): Return value must be of type bool, null returned in /home/asmecher/git/ojs-main/lib/pkp/classes/context/Context.inc.php:547
```

classes/observers/listeners/UsageEventLog.inc.php
asmecher, line 395: If the `IPGeoDB.mmdb` file hasn't been downloaded by the `UpdateIPGeoDB` task before this runs, it'll result in...
```
PHP Fatal error:  Uncaught InvalidArgumentException: The file "/home/asmecher/git/ojs-main/files/usageStats/IPGeoDB.mmdb" does not exist or is not readable. in /home/asmecher/git/ojs-main/lib/pkp/lib/vendor/maxmind-db/reader/src/MaxMind/Db/Reader.php:83
Stack trace:
#0 /home/asmecher/git/ojs-main/lib/pkp/lib/vendor/geoip2/geoip2/src/Database/Reader.php(66): MaxMind\Db\Reader->__construct()
#1 /home/asmecher/git/ojs-main/lib/pkp/classes/observers/listeners/UsageEventLog.inc.php(395): GeoIp2\Database\Reader->__construct()
#2 /home/asmecher/git/ojs-main/lib/pkp/classes/observers/listeners/UsageEventLog.inc.php(346): PKP\observers\listeners\UsageEventLog->getLocation()
#3 /home/asmecher/git/ojs-main/lib/pkp/classes/observers/listeners/UsageEventLog.inc.php(146): PKP\observers\listeners\UsageEventLog->getGeoData()
#4 /home/asmecher/git/ojs-main/lib/pkp/classes/observers/listeners/UsageEventLog.inc.php(58): PKP\observers\listeners\UsageEventLog->prepareUsageEvent()
#5 /home/asmecher/git/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(441): PKP\observers\listeners\UsageEventLog->handle()
#6 /home/asmecher/git/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(249): Illuminate\Events\Dispatcher->Illuminate\Events\{closure}()
#7 /home/asmecher/git/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php(450): Illuminate\Events\Dispatcher->dispatch()
```
The `catch` should be expanded to handle this kind of problem too.

classes/core/PKPRequest.inc.php
```diff
@@ -542,6 +542,14 @@ public function isBot()
         return $this->_isBot;
     }
 
+    /**
+     * Check if the HTTP_DNT (Do Not Track) is set
+     */
+    public function isDNTSet(): bool
+    {
+        return (array_key_exists('HTTP_DNT', $_SERVER) && ((int) $_SERVER['HTTP_DNT'] === 1));
```
asmecher, line 550: I would guess that `HTTP_DNT` is probbaly in `$_SERVER` as a `string`. I would suggest using...
```php
return (bool) $_SERVER['HTTP_DNT'] ?? false;
```

classes/core/PKPRequest.inc.php
```diff
@@ -542,6 +542,14 @@ public function isBot()
         return $this->_isBot;
     }
 
+    /**
+     * Check if the HTTP_DNT (Do Not Track) is set
+     */
+    public function isDNTSet(): bool
```
asmecher, line 548: I would suggest renaming this function to `getDoNotTrack` -- I didn't know at a glance what DNT was until I came here.

schemas/institution.json
```diff
@@ -0,0 +1,56 @@
...
+				"nullable"
+			]
+		},
+		"ror": {
+			"type": "string",
+			"description": "The ROR of this institution. See: https://ror.org/",
+			"apiSummary": true,
+			"validation": [
+				"nullable",
+				"regex:/^0[^ILOU]{6}\\d{2}$/"
```
asmecher, line 39: Users [should interact with RORs as URLs](https://ror.org/facts/):
> image

Currently the ROR field requires the user to strip off the numeric ID for the ROR to be accepted, which they will not know how to do (and the error message doesn't clarify):

image

I would recommend we follow the lead of ORCiDs and store RORs as full URLs in the database. I've only skimmed the COUNTER code of practice, and it seems unclear about it, but I'll bet it expects full URLs instead of just the ID part.

classes/sushi/CounterR5Report.inc.php
```diff
@@ -0,0 +1,537 @@
...
+            $missingRequiredParams[] = 'customer_id';
+        }
+        if (!isset($params['begin_date'])) {
+            $missingRequiredParams[] = 'begin_date';
+        }
+        if (!isset($params['end_date'])) {
+            $missingRequiredParams[] = 'end_date';
+        }
+        if (!empty($missingRequiredParams)) {
+            throw new SuhiException(
```
asmecher, line 243: Typo: missing "s"

classes/sushi/CounterR5Report.inc.php
```diff
@@ -0,0 +1,537 @@
...
+    protected function checkCustomerId($params): void
+    {
+        $institutionName = $institutionId = null;
+        $customerId = $params['customer_id'];
+        if (is_numeric($customerId)) {
+            if ($customerId == 0) {
+                $institutionName = 'The World';
+            } else {
+                $institution = Repo::institution()->get($customerId);
+                if (isset($institution) && $institution->getContextId() == $this->context) {
```
asmecher, line 267: I think `$this->context` will be an object, not the ID.

classes/sushi/CounterR5Report.inc.php
```diff
@@ -0,0 +1,537 @@
...
+     *
+     * @throws SushiException
+     */
+    public function processReportParams($request, $params): void
+    {
+        $this->context = $request->getContext();
+
+        $this->checkRequiredParams($params);
+
+        $this->checkCustomerId($params);
```
asmecher, line 195: This relies on `$this->platformId`, but that doesn't get set until `$this->setPlatform` is called (just below). I get:
```
[20-Jul-2022 00:34:28 UTC] Slim Application Error:
Type: Error
Message: Typed property PKP\sushi\CounterR5Report::$platformId must not be accessed before initialization
File: /home/asmecher/git/ojs-main/lib/pkp/classes/sushi/CounterR5Report.inc.php
Line: 274
```

classes/sushi/CounterR5Report.inc.php
asmecher: (Adding typehints throughout will help catch these at the source!)

classes/sushi/CounterR5Report.inc.php
```diff
@@ -0,0 +1,537 @@
...
+use DateTime;
+use PKP\context\Context;
+
+abstract class CounterR5Report
+{
+    /** The access method */
+    public const ACCESS_METHOD = 'Regular';
+
+    /** Access type */
+    public const ACCESS_TYPE = 'OA_Gold';
```
asmecher, line 30: Is this accurate for all users?

classes/sushi/CounterR5Report.inc.php
```diff
@@ -0,0 +1,537 @@
+<?php
+
+/**
+* @file classes/sushi/CounterR5Report.inc.php
+*
+* Copyright (c) 2013-2021 Simon Fraser University
```
asmecher, line 6: (Copyright dates should be bumped for all new code)

api/v1/stats/sushi/PKPStatsSushiHandler.inc.php
```diff
@@ -0,0 +1,230 @@
...
+                'handler' => [$this, 'getReportsPR1'],
+                'roles' => $roles
+            ],
+        ];
+    }
+
+    /**
+     * @copydoc PKPHandler::authorize()
+     */
+    public function authorize($request, &$args, $roleAssignments)
```
asmecher, line 79: I think this leaves SUSHI reports open to anyone who knows the URL and has a login (e.g. any author). I would expect these endpoints to be available to managers, admins, and subscription managers. For cases where journals want to grant external access, the journal will want to create a Subscription Manager account with an API key and provide this to the consumer. (The OJS/OMP/OPS API already handles API keys via `ApiTokenDecodingMiddleware`.)

classes/migration/install/InstitutionsMigration.inc.php
```diff
@@ -0,0 +1,81 @@
...
+            $table->foreign('context_id')->references($contextDao->primaryKeyColumn)->on($contextDao->tableName);
+        });
+
+        // Locale-specific institution data
+        Schema::create('institution_settings', function (Blueprint $table) {
+            $table->bigInteger('institution_id');
+            $table->string('locale', 14)->default('');
+            $table->string('setting_name', 255);
+            $table->text('setting_value')->nullable();
+            $table->foreign('institution_id')->references('institution_id')->on('institutions');
```
asmecher, line 44: Should have `->onDelete('cascade')`

classes/institution/Collector.inc.php
```diff
@@ -0,0 +1,174 @@
...
+                        })
+                        ->orWhere(function ($qb) use ($ip) {
+                            $qb->whereNull('ip.ip_end')
+                                ->where('ip.ip_start', '=', $ip);
+                        });
+                });
+            }
+        }
+
+        //$qb->select('i.*')->get();
```
asmecher, line 162: (cleanup)

classes/institution/Collector.inc.php
```diff
@@ -0,0 +1,174 @@
...
+        $this->includeSoftDeletes = $include;
+        return $this;
+    }
+
+    /**
+     * @copydoc CollectorInterface::getQueryBuilder()
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $qb = DB::table($this->dao->table . ' as i');
```
asmecher, line 95: Just a preference, but I prefer to add `->select('i.*')` here. Then anything that needs to augment this select (e.g. for a plugin or something) can use `addSelect` to add columns. If other code needs to override it (e.g. the calls that just get the IDs) they can replace the column list by just calling `->select(...)` again.

classes/institution/DAO.inc.php
```diff
@@ -0,0 +1,277 @@
...
+     */
+    public function newDataObject(): Institution
+    {
+        return App::make(Institution::class);
+    }
+
+    /**
+     * Check if an institution exists with this ID and context ID
+     */
+    public function existsInContext(int $id, int $contextId): bool
```
asmecher, line 63: (Again just a preference, but rather than `$id` I'd suggest using `$institutionId`. That way it's clear at a glance what kind of ID we're talking about.)

classes/institution/Institution.inc.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function getLocalizedName(string $preferredLocale = null): string
+    {
+        return $this->getLocalizedData('name', $preferredLocale);
+    }
+
+    /**
+     * Get the name of the institution
+     */
+    public function getName(string $locale = null): string|array
```
asmecher, line 73: Should be `?string` here and elsewhere, I think?

classes/migration/upgrade/v3_4_0/I6782_OrphanedMetrics.inc.php
```diff
@@ -0,0 +1,123 @@
...
+        // Clean orphaned metrics submission IDs
+        // as submission_id
+        $orphanedIds = DB::table('metrics AS m')->leftJoin('submissions AS s', 'm.submission_id', '=', 's.submission_id')->whereNotNull('m.submission_id')->whereNull('s.submission_id')->distinct()->pluck('m.submission_id');
+        $orphandedSubmissions = DB::table('metrics')->select('*')->whereIn('submission_id', $orphanedIds);
+        DB::table('metrics_tmp')->insertUsing($metricsColumns, $orphandedSubmissions);
+        DB::table('metrics')->whereIn('submission_id', $orphanedIds)->delete();
+
+        // as assoc_id
+        $orphanedIds = DB::table('metrics AS m')->leftJoin('submissions AS s', 'm.assoc_id', '=', 's.submission_id')->where('m.assoc_type', '=', Application::ASSOC_TYPE_SUBMISSION)->whereNull('s.submission_id')->distinct()->pluck('m.assoc_id');
+        $orphandedSubmissionsAssocId = DB::table('metrics')->select('*')->where('assoc_type', '=', Application::ASSOC_TYPE_SUBMISSION)->whereIn('assoc_id', $orphanedIds);
```
asmecher, line 64: Lately we've been duplicating constants in the migration classes -- see e.g.:

https://github.com/pkp/pkp-lib/blob/a4d500a74589f2a008dc29c11938d3b7fe0fa1cb/classes/migration/upgrade/PKPv3_2_1SubeditorCategoriesMigration.inc.php#L22

This ensures that the migrations use the values for constants that are appropriate for the version of the software that the upgrade is intended for. If the constant later changes (e.g. if we harmonize some of the `ASSOC_TYPE_...` values across different applications), those constants can change without breaking migrations that might otherwise have relied on them.

classes/observers/listeners/UsageEventLog.inc.php
```diff
@@ -0,0 +1,472 @@
...
+            if (!$success) {
+                // Files directory wrong configuration?
+                error_log("UsageEventLog: Couldn't create {$usageEventLogsDir}.");
+            }
+        }
+        return $usageEventLogsDir . '/usage_events_' . date('Ymd') . '.log';
+    }
+
+    /**
+     * Get salt
```
asmecher, line 217: Could you document this a little more? If I'm not sure what salt is or how it's used in event logging, this is confusing.

classes/observers/listeners/UsageEventLog.inc.php
```diff
@@ -0,0 +1,472 @@
...
+     *    'country' => string Country ISO code,
+     *    'region' => string Region ISO code
+     *    'city' => string City name
+     *   ]
+     *  ]
+     *
+     */
+    protected function getLocation(string $ip, string $hashedIp, bool $flush): array
+    {
+        $ip = '142.58.100.60';
```
asmecher, line 377: Looks like some leftover debugging code :)

classes/observers/listeners/UsageEventLog.inc.php
```diff
@@ -0,0 +1,472 @@
...
+
+    /**
+    * Get document type based on the mimetype
+    * The mimetypes considered here are subset of those used in PKPFileService::getDocumentType()
+    *
+    * @return int One of the StatisticsHelper::STATISTICS_FILE_TYPE_ constants
+    */
+    protected function getDocumentType(string $mimetype): int
+    {
+        switch ($mimetype) {
```
asmecher, line 456: The MIME type detection tools often don't work well on certain content types (e.g. HTML), so `PKPString::mime_content_type` adds some additional mechanisms to use the file extension in cases where the MIME type is frequently misdetected. Both OpenOffice/LibreOffice and HTML are frequently misdetected types. I suspect Flysystem (which we're using lately for filesystem abstraction) improves on that -- I'd suggest checking over https://flysystem.thephpleague.com/docs/advanced/mime-type-detection/#article to see.

classes/observers/traits/UsageEvent.inc.php
```diff
@@ -0,0 +1,164 @@
...
+
+    public ?SubmissionFile $submissionFile;
+
+    /** Application's complete version string */
+    public string $version;
+
+    /**
+     * Create a new usage event instance.
+     */
+    protected function constructUsageEvent(int $assocType, Context $context, Submission $submission = null, Representation $representation = null, SubmissionFile $submissionFile = null)
```
asmecher, line 60: These are all type-hinted and can't be swapped accidentally without causing an error, so no need to change it, but this is a case where you could also consider using named arguments.

classes/observers/traits/UsageEvent.inc.php
```diff
@@ -0,0 +1,164 @@
...
+                        $canonicalUrlParams[] = 'version';
+                        $canonicalUrlParams[] = $publicationId;
+                    }
+                }
+                break;
+            case Application::getContextAssocType():
+                $canonicalUrlOp = '';
+                $canonicalUrlPage = 'index';
+                break;
+        }
```
asmecher, line 107: This should throw an exception in the default case.

classes/observers/traits/UsageEvent.inc.php
```diff
@@ -0,0 +1,164 @@
...
+            null,
+            $canonicalUrlPage,
+            $canonicalUrlOp,
+            $canonicalUrlParams
+        );
+
+        // Make sure we log the server name and not aliases.
+        $configBaseUrl = Config::getVar('general', 'base_url');
+        $requestBaseUrl = $request->getBaseUrl();
+        if ($requestBaseUrl !== $configBaseUrl) {
```
asmecher, line 149: This is adapted from elsewhere in the codebase, right? This kind of URL logic gives me the heebie-jeebies and I'll be much happier when we leave to a framework!

classes/plugins/ThemePlugin.inc.php
```diff
@@ -877,6 +881,146 @@ public function isColourDark($color, $limit = 130)
...
+     */
+    protected function loadChartJavascript(): void
+    {
+        $request = Application::get()->getRequest();
+        $templateMgr = TemplateManager::getManager($request);
+
+        // Register Chart.js on the frontend article view
+        $templateMgr->addJavaScript(
+            'chartJS',
+            'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.0.1/Chart.js',
```
asmecher, line 934: We used to have an `enable_cdn` setting in `config.inc.php` that selected whether the software used a CDN or a local copy of the JS. That was removed with https://github.com/pkp/pkp-lib/issues/6478 and we're now using local copies of things (as far as I know). This effectively hides a JS dependency where it's hard to find in the codebase; I'd recommend adding it to our npm package set and linking to it locally.

classes/plugins/ThemePlugin.inc.php
```diff
@@ -877,6 +881,146 @@ public function isColourDark($color, $limit = 130)
...
+        foreach ($data as $monthlyDownloadStats) {
+            [$year, $month] = explode('-', $monthlyDownloadStats['date']);
+            $month = ltrim($month, '0');
+            $statsByMonth[$year][$month] = $monthlyDownloadStats['value'];
+            $totalDownloads += $monthlyDownloadStats['value'];
+        }
+        return [
+            'data' => $statsByMonth,
+            'label' => __('common.allDownloads'),
+            'color' => $this->getUsageStatsDispalyColor(REALLY_BIG_NUMBER),
```
asmecher, line 989: Typo: `Display`

asmecher: (Also, why is `REALLY_BIG_NUMBER` being used here?)

classes/sushi/CounterR5Report.inc.php
bozana: COUNTER R5 currently considers two access types, controlled and oa_gold. OA_Gold is only if the content has **always** been open. Currently, controlled is also content that is OA after some/embargo period. Thus we would need to track that at the submission level. S. https://github.com/pkp/pkp-lib/issues/6782#issuecomment-858610664.
Thus, we decided to 'ignore' this filter ACCESS_TYPE at the moment, s.  this https://github.com/pkp/pkp-lib/issues/6782#issuecomment-858621390. I.e. if subscription journals would need to have a better differentiation they would need to implement that tracking on submission level. For now we only support OA.

classes/sushi/CounterR5Report.inc.php
bozana: What does this mean?

api/v1/stats/sushi/PKPStatsSushiHandler.inc.php
bozana: Hmmm...
We decided:

> Yes, this is the minimal implementation for per-customer reports. We can make per-customer reports public, or, at our option, require authentication. I am not opposed to making them public by default; if a service provider want to add in authentication, this could be a contribution or sponsored development.

here: https://github.com/pkp/pkp-lib/issues/6781#issuecomment-945802002.
And to implement the opt-out for institutions.

But I could also change it to restrict the access to those roles.
Just to double check it: @ctgraham, what do you think?

classes/observers/listeners/UsageEventLog.inc.php
bozana: I am not sure but: this function already gets the calculated/detected mimetype -- it only then returns the COUNTER file type depending on the mime type. So do you think I should do something else here?

classes/institution/Institution.inc.php
bozana: It is not needed, because we have the default value and the user would use the function without parameters: `->getName()`
If we would like to use it like this `->getName(null)` and without the default value `public function getName(?string $locale) `then we would need it. Do we have a proffered way for these cases?

classes/sushi/CounterR5Report.inc.php
bozana: > (Adding typehints throughout will help catch these at the source!)

Here we can not add extra/additional typehints, right?
institution->getContextId() does return int (typehinted)... :thinking:

classes/migration/upgrade/v3_4_0/I6782_OrphanedMetrics.inc.php
bozana: I also added the protected function getContextAssocType()...

classes/observers/listeners/UsageEventLog.inc.php
bozana: Does this sound OK:
  ```
/**
  * Get salt (used for IP hashing) from the file.
  * For privacy reasons, the salt will change every day.
  * When salt changes, the IP hash will change too, so the cach containing the hashed IPs (for Geo and institutions data) will be removed.
  * If the salt file does not exist, it will be created.
  * If the salt file exists and it was not modified at the current day the salt will be changed.
*/
```

File: classes/observers/traits/UsageEvent.inc.php (hunk 23 above)
Comment (bozana) [reply classes/observers/traits/UsageEvent.inc.php:107]: I needed to reorganize the code a little bit for this...

File: classes/plugins/ThemePlugin.inc.php (hunk 26 above)
Comment (bozana) [reply classes/plugins/ThemePlugin.inc.php:989]: To be hones, I do not know -- I've kept it from UsageStatsPlugin, s. https://github.com/pkp/usageStats/blob/main/UsageStatsPlugin.inc.php#L870.

File: classes/sushi/CounterR5Report.inc.php (hunk 10 above)
Comment (asmecher) [reply classes/sushi/CounterR5Report.inc.php:30]: Could the report maybe then be disabled when access is configured not to be open access? I'm on the fence about whether this is worth the trouble, but it seems like it should be impossible to generate a report when the access mode is unsupported, and we do have that setting ready to hand.

File: classes/sushi/CounterR5Report.inc.php (hunk 11 above)
Comment (asmecher) [reply classes/sushi/CounterR5Report.inc.php:6]: Change the copyright end dates to 2022. (Actually for new code it should just be (c) 2022, but if the start date is older, it's probably because the header got copied from another file -- which is I suppose derivative :))

File: classes/observers/listeners/UsageEventLog.inc.php (hunk 21 above)
Comment (asmecher) [reply classes/observers/listeners/UsageEventLog.inc.php:456]: As coded, it will frequently detect `text/xml` instead of `text/html`, and `application/zip` instead of the `application/vnd.openxmlformats-officedocument.*` MIME types.

File: classes/institution/Institution.inc.php (hunk 17 above)
Comment (asmecher) [reply classes/institution/Institution.inc.php:73]: I think this is a quirk of type hinting that I've looked up before. No change needed!

File: classes/observers/listeners/UsageEventLog.inc.php (hunk 21 above)
Comment (asmecher) [reply classes/observers/listeners/UsageEventLog.inc.php:456]: Ah! OK, then you should be fine, thanks.

File: classes/plugins/ThemePlugin.inc.php (hunk 26 above)
Comment (asmecher) [reply classes/plugins/ThemePlugin.inc.php:989]: OK, let's leave it as is for now.

File: api/v1/stats/sushi/PKPStatsSushiHandler.inc.php [hunk 27]
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Get the current status of the reporting service
+     */
+    public function getStatus(SlimHttpRequest $slimRequest, APIResponse $response, array $args): APIResponse
+    {
+        $request = $this->getRequest();
+        $context = $request->getContext();
+        // use only the name in the contxt primary locale to be consistent
```
asmecher, line 92: *context

classes/context/Context.inc.php
asmecher: (I suspect the other new boolean settings will be the same, in case you didn't catch them too!)

bozana:
Closing this PR, which was the second round of review...

PR #8056, opened 2022-06-30
[pkp-lib][stable_3.3.0] #7893 Change AuthorForm supportedLocales from Form Locales(AppLocale::getSupportedFormLocales) to context's Submission Locales (getSupportedSubmissionLocaleNames())

defstat:


controllers/grid/users/author/form/PKPAuthorForm.inc.php
```diff
@@ -26,13 +26,19 @@ class PKPAuthorForm extends Form {
 	 * Constructor.
 	 */
 	function __construct($publication, $author) {
-		parent::__construct('controllers/grid/users/author/form/authorForm.tpl');
+		$submission = Services::get('submission')->get($publication->getData('submissionId'));
+		$context = Services::get('context')->get($submission->getData('contextId'));
```
NateWr, line 30: Both of these are expensive operations, because the objects are so large. We should try to pass them into `new AuthorForm()` from the objects that already exist in `AuthorGridHandler`:

```
$form = new AuthorForm(
    $this->getPublication(),
    $author,
    $this->getSubmission(),
    $request->getContext()
);
```

There are two places that `AuthorForm` is instantiated in the grid.

controllers/grid/users/author/form/PKPAuthorForm.inc.php
```diff
@@ -26,13 +26,19 @@ class PKPAuthorForm extends Form {
...
+
+		parent::__construct(
+			'controllers/grid/users/author/form/authorForm.tpl', 
+			true, 
+			$publication->getData('locale'), 
+			$context->getSupportedSubmissionLocaleNames()
```
NateWr, line 36: Are you sure this is the right method? I think you might want `$context->getSupportedSubmissionLocales()` but I'm not sure.

NateWr (changes requested):

controllers/grid/users/author/form/PKPAuthorForm.inc.php
defstat: Here I am the same function as in https://github.com/pkp/pkp-lib/blob/stable-3_3_0/classes/submission/form/SubmissionSubmitForm.inc.php#L44. 

By using the proposed by the PR function, the result is the following
image

By using the `$context->getSupportedSubmissionLocales()` the result is the following:
image

The first (and proposed by the PR) seems the correct to me.

NateWr: :+1: You're right, thanks for looking into it.

controllers/grid/users/author/form/PKPAuthorForm.inc.php
NateWr: Looks good!

PR #8057, opened 2022-06-30
Feature/main/8040 update tests

jonasraoni:


tests/phpunit.xml
```diff
@@ -1,22 +1,32 @@
...
             <directory suffix="Test.php">classes</directory>
         </testsuite>
+        <testsuite name="ApplicationClasses">
+            <directory suffix="Test.php">../../../tests/classes</directory>
+        </testsuite>
+        <testsuite name="LibraryPlugins">
+            <directory suffix="Test.php">../plugins/*/*/tests</directory>
+        </testsuite>
+        <testsuite name="ApplicationPlugins">
+            <directory suffix="Test.php">../../../plugins/*/*/tests</directory>
```
jonasraoni, line 28: Moved path configuration to this file

tools/runAllTests.sh
```diff
@@ -65,22 +57,37 @@ while getopts "bCPcpfdH" opt; do
 			;;
 		d)	DEBUG="--debug"
 			;;
+		R)	DO_COVERAGE=1
+			;;
 	esac
 done
-phpunit='php lib/pkp/lib/vendor/phpunit/phpunit/phpunit'
+PHPUNIT='php lib/pkp/lib/vendor/phpunit/phpunit/phpunit --configuration lib/pkp/tests/phpunit.xml --testdox --no-interaction'
```
jonasraoni, line 64: The `testdox` will just prettify the output in the build

tests/classes/notification/PKPNotificationManagerTest.php
```diff
@@ -464,12 +478,4 @@ private function getTrivialNotification()
 
         return $notification;
     }
-
-    public function tearDown(): void
-    {
-        parent::tearDown();
-
-        // See: http://docs.mockery.io/en/latest/reference/phpunit_integration.html
```
asmecher, line 472: Are you sure this is no longer necessary? I think it's recommended when mixing PHPUnit and Mockery.

classes/install/form/InstallForm.inc.php
```diff
@@ -43,7 +43,7 @@ class InstallForm extends MaintenanceForm
     public $supportedDatabaseDrivers = [
         // <driver> => array(<php-module>, <name>)
         'mysqli' => ['mysqli', 'MySQLi'],
-        'postgres9' => ['pgsql', 'PostgreSQL'],
+        'postgres' => ['pgsql', 'PostgreSQL'],
```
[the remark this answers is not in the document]
jonasraoni: I'll double check. I wrote this `postgres9`, and at first I thought it was just a typo 😁

tests/classes/notification/PKPNotificationManagerTest.php
jonasraoni: I moved the `Mockery::close()` to the `PKPTestCase::tearDown()`.

classes/install/form/InstallForm.inc.php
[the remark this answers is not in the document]
jonasraoni: I've just checked and it makes sense to keep the postgres9 :)

Should I create an issue to replace these ADOdb identifiers by Laravel? Also, the `mysql` driver can be removed, the extension has been gone since PHP 7.0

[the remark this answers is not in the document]
asmecher: This is a really good reason to avoid wandering out of scope when making changes -- it's a very subtle change unrelated to the issue that could've gone unnoticed and been very tricky to debug!

We'll remove the ADODB driver identifiers and generally clean up database connection stuff when we're able to remove ADODB fully (after the upgrade to 3.3.x gets removed). But yes, feel free to remove the `mysql` driver from the installation form.

tests/classes/core/PKPComponentRouterTest.php
```diff
@@ -267,10 +272,13 @@ public function testRoute()
...
         Registry::set('user', $user);
+        $serviceEndpoint = $this->router->getRpcServiceEndpoint($this->request);
+        $handler = $serviceEndpoint[0];
+        $handler->addPolicy(new UserRolesRequiredPolicy($this->request), true);
         $this->router->route($this->request);
 
-        self::assertNotNull($serviceEndpoint = & $this->router->getRpcServiceEndpoint($this->request));
-        self::assertInstanceOf('\PKP\controllers\grid\notifications\NotificationsGridHandler', $handler = & $serviceEndpoint[0]);
+        self::assertNotNull($serviceEndpoint);
+        self::assertInstanceOf('\PKP\controllers\grid\notifications\NotificationsGridHandler', $handler);
```
asmecher, line 281: BTW, you can use `self::assertInstanceOf(\PKP\controllers\grid\notifications\NotificationsGridHandler::class, $handler);` -- I've been converting string-based forms (especially `is_a` into `instanceof`) when I find them.

tests/phpunit.xml
```diff
@@ -1,22 +1,32 @@
...
     processIsolation="false"
-    stopOnFailure="true"
-    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
+    stopOnFailure="true">
     <testsuites>
-        <testsuite name="Common Test Suite">
+        <testsuite name="LibraryClasses">
             <directory suffix="Test.php">classes</directory>
         </testsuite>
+        <testsuite name="ApplicationClasses">
```
asmecher, line 21: :+1:, much better to be using the PHPUnit suite management rather than reinventing it in shell scripts!

tests/classes/core/PKPComponentRouterTest.php
jonasraoni: Yeah, it's also possible to use it on instances (e.g. `$instance::class`) =]

PR #8088, opened 2022-07-07
pkp/pkp-lib#8060 User locales column default updated to empty array

touhidurabir:
This PR aims to provide a solution for pkp/pkp-lib#8060.

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I8060_UpdateUserLocalesDefaultToEmptyArrayFromNull.inc.php
+++ b/classes/migration/upgrade/v3_4_0/I8060_UpdateUserLocalesDefaultToEmptyArrayFromNull.inc.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I8060_UpdateUserLocalesDefaultToEmptyArrayFromNull.php
+ *
+ * Copyright (c) 2014-2022 Simon Fraser University
+ * Copyright (c) 2000-2022 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I8060_UpdateUserLocalesDefaultToEmptyArray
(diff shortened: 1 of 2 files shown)


asmecher:
I'd recommend combining the two migrations into one, and removing the `nullable` flag from `user.locales` once the default has been set. That way it'll be impossible for future `null` values to get into the system.

classes/migration/upgrade/v3_4_0/I8060_UpdateUserLocalesDefaultToEmptyArrayFromNull.inc.php
```diff
@@ -0,0 +1,36 @@
...
+        DB::table('users')->whereNull('locales')->update(['locales' => '[]']);
+
+        Schema::table('users', function ($table) {
+            $table->string('locales', 255)->nullable(false)->default('[]')->change();
+        });
+    }
+
+    public function down(): void
+    {
+        throw new DowngradeNotSupportedException('Downgrade unsupported due to updated data');
```
Vitaliy-1, line 34: Basically, we can support downgrades in these cases by saving data as a property of the current migration object before we modify it.

asmecher: In this case, I would recommend restoring the `locales` column to `nullable` but leaving the data as it was after the `up` function runs. The difference between `null` and `'[]'` is negligible, so not worth the pain of implementing a 100% restore to the previous state.

But yes, @Vitaliy-1, in theory that's possible -- I hadn't considered it and will probably use it in other migrations!

touhidurabir: @asmecher I have update the migration `down` so that it can be reverted back to `nullable` but I think it' best to keep the `default` to `[]` .

PR #8110, opened 2022-07-24
pkp/pkp-lib#8027 OPS preprints should support DOI versioning

ewhanson:


api/v1/dois/PKPDoiHandler.inc.php
```diff
@@ -260,11 +260,36 @@ public function edit(SlimRequest $slimRequest, APIResponse $response, array $arg
...
+		}
+
+		// Copy DOI object data
+		$newDoi = clone $doi;
+		$newDoi->unsetData('id');
+		$newDoi->setAllData(array_merge($newDoi->getAllData(), $params));
+		$newDoiId = Repo::doi()->add($newDoi);
+
+		Repo::doi()->updatePubObjectDoiFromType($pubObject ,$pubObjectType, $newDoiId);
+		Repo::doi()->removeDanglingObjects($doi->getId());
```
NateWr, line 290: Since `getPubObjectFromType()` and `updatePubObjectDoiFromType()` are all part of the funny logic of the API, let's make these protected methods of the handler class. I know you'll need a new `DoiHandler()` for each application, but it should be pretty small. You can also then define some constants for the types there which match the API specifications, without needing to roll this out to the rest of the application.

It might also free you to do a little bit more in the `edit` method and less in the helper methods. I would imagine something like a more general-purpose use function in the repository:

```php
class Repository
{
    public function isAssigned(int $doiId): bool
    {
        $publicationCount = ...;
        $galleyCount = ...;

        return $publicationCount || $galleyCount;
    }
}
```

And then the rest of the logic is in the API handler, and the app-specific API handler only needs to be responsible for returning the right variable for each object type. It doesn't have to take the action:

```php
class PKPDoiHandler extends APIHandler
{
    public const OBJECT_TYPE_SUBMISSION = 'submission';

    protected method getPubObjectRepo(string $type)
    {
        return match($type) {
            self::OBJECT_TYPE_SUBMISSION => Repo::submission(),
        };
    }

    public function edit(...)
    {
        // ...

        $repo = $this->getPubObjectRepo($type);
        if (!$repo) {
            return $response->withStatus(403)->withJsonError('pub.type.not.recognized');
        }
        
        // ...

        $repo->edit(...);

        if (!Repo::doi()->isAssigned($doi->getId()) {
            Repo::doi()->delete($doi);
        }
    }
}
```

classes/components/listPanels/PKPDoiListPanel.inc.php
```diff
@@ -163,7 +163,13 @@ public function getConfig()
...
             'list.expandAll',
             'list.collapseAll',
-            'manager.dois.update.failedCreation'
+            'manager.dois.update.failedCreation',
+			'publication.version',
+			'doi.manager.versions.countStatement',
+			'doi.manager.versions.view',
+			'doi.manager.saveDois',
+			'doi.manager.editDois',
+			'doi.manager.versions.modalTitle',
```
NateWr, line 172: showing tabs not spaces here

NateWr: I'm noticing that throughout the changes in this PR...

locale/en_US/manager.po
```diff
@@ -835,6 +835,21 @@ msgstr "None"
...
+msgstr "View all"
+
+msgid "doi.manager.versions.modalTitle"
+msgstr "DOIs for all versions"
+
+msgid "doi.manager.saveDois"
+msgstr "Save changes"
+
+msgid "doi.manager.editDois"
+msgstr "Edit DOI(s)"
```
NateWr, line 851: Can we get away with re-using existing locale keys, like "Save" and "Edit"?

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+        $publication = $event->newPublication;
+
+        /** @var ContextDAO $contextDao */
+        $contextDao = Application::getContextDAO();
+        /** @var Context $context */
+        $context = $contextDao->getById($submission->getData('contextId'));
+
+        $shouldVersionPublication = $context->getData(Context::SETTING_DOI_VERSIONING);
+        $notFirstPublication = $publication->getData('version') != 1;
+        if ($shouldVersionPublication && $notFirstPublication) {
```
NateWr, line 60: Reverse the logic here to return early:

```php
if (!$shouldVersionPublication || !$notFirstPublication) {
    return;
}
```

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public function handlePublishedEvent(PublishedEvent $event): void
+    {
+        $submission = $event->submission;
+        $publication = $event->newPublication;
+
+        /** @var ContextDAO $contextDao */
+        $contextDao = Application::getContextDAO();
+        /** @var Context $context */
+        $context = $contextDao->getById($submission->getData('contextId'));
```
NateWr, line 56: The context is also in `PublishedEvent`:

```php
$context = $event->context;
```

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $shouldVersionPublication = $context->getData(Context::SETTING_DOI_VERSIONING);
+        $notFirstPublication = $publication->getData('version') != 1;
```
NateWr, line 59: I'm not sure if this is certain to be true. I think it is possible to end up with a different version number even if it is the earliest publication that is currently published. It may be possible through publishing/unpublishing things.

You may need to rely on logic like you use below, where you check if a DOI exists that matches a DOI of a previous publication or something.

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+     * NB: The existing doiId should be removed from the latest publication when it matches that of the previous publication,
+     * i.e. it is a copy of the previous publication. This prevents a legitimate DOI from being removed,
+     * e.g. if a versioned publication is unpublished then republished.
+     *
+     */
+    public function handlePublishedEvent(PublishedEvent $event): void
+    {
+        $submission = $event->submission;
+        $publication = $event->newPublication;
+
```
NateWr, line 52: I think you should maybe check if DOIs are enabled on the context-level before doing anything. When we have some settings (like should dois be versioned) which are dependent on other settings (dois enabled), there's always a risk that someone will have dois disabled but the old setting to version DOIs will still be in the system.

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+            $previousPublication = $publicationsSorted->get(1);
+
+            // The existing doiId should be removed from the latest publication when it matches that of the previous publication,
+            // i.e. it is a copy of the previous publication. This prevents a legitimate DOI from being removed,
+            // e.g. if a versioned publication is unpublished then republished.
+            $hasPreviousPublicationDoiId = $latestPublication->getData('doiId') === $previousPublication->getData('doiId') && $latestPublication->getData('doiId') !== null;
+            if ($context->isDoiTypeEnabled(Repo::doi()::TYPE_PUBLICATION) && $hasPreviousPublicationDoiId) {
+                $publication->setData('doiId', null);
+                Repo::publication()->dao->update($publication);
+            }
```
NateWr, line 81: Lines 62-81 would be good to move into a protected function. I'm not sure I understand all of the inline comments and one clear, succinct docblock above a function might make it easier to understand the general intent of what's happening.

classes/observers/listeners/VersionPublicationDois.php
```diff
@@ -0,0 +1,107 @@
...
+            self::class . '@handlePublishedEvent'
+        );
+    }
+
+    /**
+     * Checks if new publication version should carry over existing DOIs or mint new ones and creates new DOIs as necessary
+     *
+     * NB: The existing doiId should be removed from the latest publication when it matches that of the previous publication,
+     * i.e. it is a copy of the previous publication. This prevents a legitimate DOI from being removed,
+     * e.g. if a versioned publication is unpublished then republished.
```
NateWr, line 45: You use NB, i.e. and e.g. here. I'd prefer to just use words in these cases like "Please note" and "for example" or "such as".

classes/observers/listeners/VersionPublicationDois.php
NateWr: Similar with 83-101 below.

locale/en_US/manager.po
```diff
@@ -907,9 +907,31 @@ msgstr ""
 msgid "doi.manager.settings.registrationAgency.none"
 msgstr "None"
 
+msgid "doi.manager.settings.doiVersioning"
+msgstr "DOI Versioning"
+
+msgid "doi.manager.settings.doiVersioning.description"
+msgstr "Assign a new DOI to each publication version? "
+"Most preprint servers will want to use the default option (yes) to ensure preprint versions are correctly assigned DOIs."
```
NateWr, line 915: > Most preprint servers should use the ... to ensure that each version of a preprint is assigned a unique DOI.

NateWr: This needs to go into the OPS repository.

NateWr (changes requested):

bozana (commented):
Also a few general comments/questions:

- I think the `doiVersioning` setting is not considered at the installation?
- In the function `unpublish()`  (publication repository), the `getDoisForSubmission()` is called: There only the current publication is considered (to mark stale). I think one could unpublish another, not the current publication as well, correct? (Maybe then it should have the unpublished publication as parameter and get those DOIs?)
- The documentation of the function `getDoisForSubmission()` in the code seems to be wrong
- Why do we not consider DOI status `unassigned` (instead of using `unregistered` for that too)? -- This way the user could maybe select an unregistered DOI and try to register it, although there is no DOI...

classes/publication/Repository.php
```diff
@@ -293,14 +295,18 @@ public function add(Publication $publication): int
      * Makes a copy of an existing publication, without the datePublished,
      * and makes copies of all associated objects.
      */
-    public function version(Publication $publication): int
+    public function version(Publication $publication, Submission $submission, Context $context, User $editor): int
     {
```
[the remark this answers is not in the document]
ewhanson: I'm not using it, but @NateWr suggested it be added now since we were already changing the function signature as it would likely be helpful to have the associated user for hooks/events/etc.

classes/components/listPanels/PKPDoiListPanel.php
```diff
@@ -172,6 +176,13 @@ public function getConfig()
             'manager.dois.update.failedCreation',
             'plugins.importexport.common.action.export',
             'manager.dois.actions.description',
+            'manager.dois.update.failedCreation',
+            'publication.version',
```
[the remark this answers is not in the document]
ewhanson: Looks like this is only in OPS, so I'll move it over to pkp-lib. I'll probably wait until right before merge to avoid translation conflicts. @asmecher, could I give you a heads up then for you to pull in the latest translations?

classes/publication/Repository.php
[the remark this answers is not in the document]
NateWr: @asmecher has pushed back on this idea in our discussion about events here: https://github.com/pkp/pkp-lib/discussions/8388#discussioncomment-4040677

Maybe until we settle on an approach, it's better to load the dependencies you need inside the event listener. It won't be as performant but it is easier to add arguments to a function later than it is to remove them.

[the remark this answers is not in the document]
bozana: This is also a place in the code where there is no event... So yes, I would remove it...

api/v1/dois/PKPDoiHandler.php
```diff
@@ -269,28 +275,59 @@ public function edit(SlimRequest $slimRequest, APIResponse $response, array $arg
...
             return $response->withStatus(400)->withJson($errors);
         }
 
-        Repo::doi()->edit($doi, $params);
+        $pubObjectType = $slimRequest->getParsedBodyParam('pubObjectType');
+        $pubObjectId = $slimRequest->getParsedBodyParam('pubObjectId');
 
-        $doi = Repo::doi()->get($doi->getId());
+        // Default behaviour, only edits DOI
+        if (empty($pubObjectType) && empty($pubObjectId)) {
```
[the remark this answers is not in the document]
ewhanson: This one was a bit messy. It's not always included because this `edit()` function covers two (theoretical) use cases.

1. Being used as a "pure" API endpoint to only edit the DOI with no side effects. (This is the theoretical case, and I'm not sure if this should be allowed as it could lead to unintended consequences). This is the case without including `$pubObjectType` and `$pubObjectId`.
2. Being used to edit the DOIs in the context of their publication object (how it's used in the UI). The code below are the side effects that need to happen when editing a DOI in context. 

```PHP
// Copy DOI object data
$newDoi = clone $doi;
$newDoi->unsetData('id');
$newDoi->setAllData(array_merge($newDoi->getAllData(), ['doi' => $params['doi']]));
$newDoiId = Repo::doi()->add($newDoi);

// Update pubObject with new DOI and remove elsewhere if no longer in use
$pubObjectRepo->edit($pubObject, ['doiId' => $newDoiId]);
if (!Repo::doi()->isAssigned($doi->getId())) {
    Repo::doi()->delete($doi);
}
```

Given this, do you a request should always be required to include a `$pubObjectType` and `$pubObjectId` or is it okay to keep as is?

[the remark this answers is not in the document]
bozana: So the 1. case is maybe when editing a DOI manually via the UI/DOI management page?
And the 2. case is e.g. when publishing a new version?
If so maybe to add a comment in the function documentation or here in the code...
And, yes, we can then keep it as is...

ewhanson:
Hi @bozana, thanks for reviewing this! To your specific questions:

> I think the `doiVersioning` setting is not considered at the installation?

- `doiVersioning` is part of the `context.json` schema file, so it should be setup correctly as part of the installation. Is there anywhere else it was missing?

> In the function `unpublish()` (publication repository), the `getDoisForSubmission()` is called: There only the current publication is considered (to mark stale). I think one could unpublish another, not the current publication as well, correct? (Maybe then it should have the unpublished publication as parameter and get those DOIs?)

- I've updated `getDoisForSubmission()` to account for all publications.

> The documentation of the function `getDoisForSubmission()` in the code seems to be wrong

- Thanks! Updated this as well.

> Why do we not consider DOI status `unassigned` (instead of using `unregistered` for that too)? -- This way the user could maybe select an unregistered DOI and try to register it, although there is no DOI...

- In terms of the possible statuses a DOI can have, there is no `unassigned` status because a DOI object won't exist until it has been assigned. In terms of the DOI management UI, I see how this could be confusing. We have a "Needs DOI" filter that partially addresses this use case, but maybe it would be helpful to have the badge that says "Unregistered" say "Unassigned" when there are no DOIs assigned. If that sounds like a good strategy to you, I can create a new issue for it.

bozana:
Hi @ewhanson, thanks a lot!
If there is no reason not to do it I would prefer to have a badge that says "Needs DOI" (or "Unassigned")...

classes/doi/Repository.php
```diff
@@ -310,6 +309,26 @@ public function depositAll(Context $context)
         }
     }
 
+    /**
+     * Checks whether a DOI object is referenced by ID on any pub objects
+     */
+    public function isAssigned(int $doiId): bool
```
bozana, line 315: Do we need to implement this function in apps, e.g. for issues (ojs), files (omp)?

api/v1/dois/PKPDoiHandler.php
```diff
@@ -269,28 +278,59 @@ public function edit(SlimRequest $slimRequest, APIResponse $response, array $arg
...
+
+        // Copy DOI object data
+        $newDoi = clone $doi;
+        $newDoi->unsetData('id');
+        $newDoi->setAllData(array_merge($newDoi->getAllData(), ['doi' => $params['doi']]));
+        $newDoiId = Repo::doi()->add($newDoi);
+
+        // Update pubObject with new DOI and remove elsewhere if no longer in use
+        $pubObjectRepo->edit($pubObject, ['doiId' => $newDoiId]);
+        if (!Repo::doi()->isAssigned($doi->getId())) {
```
bozana, line 318: We could theoretically provide pubObjectType to the isAssigned function -- the same DOI can only be assigned to an object from the same type -- or?

classes/components/listPanels/PKPDoiListPanel.php
```diff
@@ -172,6 +177,12 @@ public function getConfig()
             'manager.dois.update.failedCreation',
             'plugins.importexport.common.action.export',
             'manager.dois.actions.description',
+            'publication.version',
```
bozana, line 180: We need to remove this key from here, right?

api/v1/dois/PKPDoiHandler.php
```diff
@@ -269,28 +278,59 @@ public function edit(SlimRequest $slimRequest, APIResponse $response, array $arg
...
             return $response->withStatus(400)->withJson($errors);
         }
 
-        Repo::doi()->edit($doi, $params);
+        $pubObjectType = $slimRequest->getParsedBodyParam('pubObjectType');
+        $pubObjectId = $slimRequest->getParsedBodyParam('pubObjectId');
 
-        $doi = Repo::doi()->get($doi->getId());
+        // Default behaviour, only edits DOI
+        if (empty($pubObjectType) && empty($pubObjectId)) {
```
bozana, line 292: I think we need to consider issue type here -- it always fits into this default behavior because there is no versioning for issues -- correct? Else, further down the type not found will be returned when an issue DOI is edited.

api/v1/dois/PKPDoiHandler.php
```diff
@@ -33,12 +33,16 @@
...
 use PKP\security\Role;
+use PKP\services\PKPSchemaService;
 
 use Slim\Http\Request as SlimRequest;
 use Slim\Http\Response;
 
 class PKPDoiHandler extends APIHandler
 {
+    public const OBJECT_TYPE_PUBLICATION = 'publication';
+    public const OBJECT_TYPE_REPRESENTATION = 'representation';
```
bozana, line 44: I think we need to consider chapter and submission file type in OMP

classes/doi/Repository.php
ewhanson: Added!

classes/components/listPanels/PKPDoiListPanel.php
ewhanson: No, this is the locale key that needed to be transferred from OPS to pkp-lib. This is the `"Version 1"`, `"Version 2"`, etc string used in the version modal.

api/v1/dois/PKPDoiHandler.php
ewhanson: That's a good idea. It would reduce the number of database queries for each check if it's only checking the one pub object type.

api/v1/dois/PKPDoiHandler.php
ewhanson: Yes, I've opted to use the preexisting constants in the DOI repository, e.g. `Repo::doi()::TYPE_PUBLICATION`.

api/v1/dois/PKPDoiHandler.php
ewhanson: Yes, I've rewritten it to account for issues in OJS so the issue type is now accounted for.

PR #8122, opened 2022-07-29
pkp/pkp-lib#6091 Rename .inc.php controllers to .php; fix namespacing

asmecher:


jonasraoni (approved):

jonasraoni:
Just two reminders:
- Namespaced classes without the `class_alias()`: serialized objects might fail to be revived, hopefully there are few instances of `serialize()` in the codebase to analyze.
The cache should be cleared whenever something is updated, but there might be something saved in the database.
- The removal of `import()` might raise issues with non-namespaced constants used down the call stack (in other files).

asmecher:
These are just the controller classes, which never get serialized, and are not generally named in e.g. plugins, so I was comfortable skipping the `class_alias` stuff that was applied elsewhere.

We're not currently using non-namespaced constants; legacy code uses `define`, which does not assume the namespace of the class that it's invoked from. (We're gradually moving these into class constants, at the same time as we're adding namespaces, but constants aren't defined in handlers except for one or two outlier cases.)

asmecher:
Manually merged.

PR #8133, opened 2022-08-04
pkp/pkp-lib#7105 Job process via cron and job command enhancement

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#7105


classes/core/PKPRequest.php
```diff
@@ -103,6 +103,13 @@ public function setDispatcher($dispatcher)
      */
     public function &getDispatcher()
     {
+        if ( ! $this->_dispatcher ) {
```
asmecher, line 106: This potentially an invasive change. Instead, I think we should stop depending on `PKPRequest` in areas of the code that may be called e.g. via command line. A common problem is URL generation; the usual pattern, `$request->url`, depends on rather than calling `$request->url`, for example, you can use instead:
```php
$request = $application->getRequest();
$dispatcher = $application->getDispatcher();
$url = $dispatcher->url($request, PKPApplication::ROUTE_PAGE, ...);
```
This still depends on `$request` being fetched and passed in -- we have more work to do around disentangling this -- but it'll work when called in a command-line environment. This is done e.g. in `lib/pkp/classes/task/ReviewReminder.php`.

classes/task/ProcessQueueJobs.php
```diff
@@ -0,0 +1,54 @@
...
+        while ($jobBuilder->count()) {
+            $jobQueue->runJobsInQueue();
+        }
+
+        return true;
+    }
+}
+
+if (!PKP_STRICT_MODE) {
+    class_alias('\PKP\task\ProcessQueueJobs', '\ProcessQueueJobs');
```
asmecher, line 53: New classes don't need aliases; this is just for backwards compatibility.

tools/jobs.php
```diff
@@ -292,6 +296,52 @@ protected function list(): void
...
+                __(
+                    'admin.cli.tool.jobs.available.options.run.empty.description', 
+                    ['queueName' => $queue,]
+                )
+            );
+
+            return;
+        }
+
+        while($jobBuilder->count()) {
```
asmecher, line 332: Just checking -- is this resilient against failed jobs? (i.e. if a job remains in the queue because it failed, will this hang?)

asmecher (changes requested):
Just a few comments, @touhidurabir -- I think this is just about ready to be merged separately before moving on to the next steps with queues. Thanks!

classes/core/PKPRequest.php
touhidurabir: I agree 100% and this is not an ideal implementation but more like a patch up or hot fix . But the problem lies with our excessive use of code such as 
```php
Application::get()->getRequest()->getDispatcher()
```

if we have have even used it like 
```php
Application::get()->getDispatcher()
```

it would have been ok or the way you have mentioned . Now if this does not get included , then we have to refactor all the occurrence of such code blocks to make it compatible with queue jobs or any command line utility that may run on separate process outside of main app process . I can see one potential place for such refactoring is in the `mailable` section . Also others also need to be careful about code implementation so that it would not break if that implementation get associated with a queue jobs through worker. I feel like most of us continue to use approach like `Application::get()->getRequest()->getDispatcher()` and get stuck with queue jobs to figure out why it's not working or throwing internal exception . Also our job ui is not sophisticated yet to debug failed jobs from there( that is the next part I want to work on related to queue jobs) . 

One example of such case already is job `PKP\Jobs\Email\EditorialReminder` failed when running it through the job command line tool as the core `mailable` and `email variable classes` have call such as `Application::get()->getRequest()->getDispatcher()`. 

@asmecher  In fact I did not like that implementation of mine either and have nothing against removing that but doing so  we will be forcing others to do a lot of refactoring probably. what do you think about it ? if we want to avoid such refactoring now , how about keep it for now, let others know about such cases, after a certain time we remove it when we are sure that such calls have been removed from our codebase , what you say ?

tools/jobs.php
touhidurabir: @asmecher  No that is how it's works . basically a job will and should go (unless there is an issue with our implementation or design), it will go to failed table after 3 failed attempt. the implementation for cron or schedular do utilise laravel's worker but not through a `daemon` . so we need a way to pop and push jobs into the worker queue one by one through the `Illuminate\Queue\Worker::runNextJob()` . Also the DB query check of max attempt as a precaution to make sure if a job has been attempted max time and yet present in the `jobs` table , will not be picked up for processing . Again, a job that is reserved by marking the `reserved_at` column will not be picked up so we should be safe from a loop running for infinite time. But we can make it better by making a `max job processing limit in a single call` just to be safe . I will do that just to have an infinite time running trap handler but that can not be implemented for `daemon` feature as that is the purpose of `daemon` feature .

asmecher: Thanks, that sounds good!

classes/core/PKPRequest.php
asmecher: Thinking a little more about this -- I see your point. And the more verbose work-around I suggested above...
```php
$request = $application->getRequest();
$dispatcher = $application->getDispatcher();
$url = $dispatcher->url($request, PKPApplication::ROUTE_PAGE, ...);
```
...is hardly less dependent on the `$request` object. So let's go ahead with your proposal!

classes/task/ProcessQueueJobs.php
touhidurabir: But without this I am getting errors as 

```
[08-Aug-2022 17:39:15 Asia/Dhaka] PHP Fatal error:  Uncaught Exception: Cannot instantiate class. Class "ReviewReminder" is not declared in "lib.pkp.classes.task.ReviewReminder". in /Users/abir/Sites/code/ojs-main/lib/pkp/includes/functions.php:223
Stack trace:
#0 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/cliTool/ScheduledTaskTool.php(117): instantiate('lib.pkp.classes...', NULL, NULL, 'execute', Array)
#1 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/cliTool/ScheduledTaskTool.php(103): PKP\cliTool\ScheduledTaskTool->executeTask('lib.pkp.classes...', Array)
#2 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/cliTool/ScheduledTaskTool.php(73): PKP\cliTool\ScheduledTaskTool->parseTasks('registry/schedu...')
#3 /Users/abir/Sites/code/ojs-main/tools/runScheduledTasks.php(25): PKP\cliTool\ScheduledTaskTool->execute()
#4 {main}
  thrown in /Users/abir/Sites/code/ojs-main/lib/pkp/includes/functions.php on line 223
```
Similar errors for any schedule tasks . May be I am missing something .

classes/core/PKPContainer.php
```diff
@@ -112,13 +116,12 @@ public function registerConfiguredProviders()
...
         // Load main settings, this should be done before registering services, e.g., it's used by Database Service
         $this->loadConfiguration();
-
         $this->register(new PKPEventServiceProvider($this));
         $this->register(new LogServiceProvider($this));
         $this->register(new \Illuminate\Database\DatabaseServiceProvider($this));
         $this->register(new \Illuminate\Bus\BusServiceProvider($this));
-        $this->register(new \Illuminate\Queue\QueueServiceProvider($this));
-        $this->register(new PKPQueueProvider());
+        // $this->register(new \Illuminate\Queue\QueueServiceProvider($this));
```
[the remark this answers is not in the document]
touhidurabir: Removed.

Domains/Jobs/WorkerConfiguration.php
```diff
@@ -0,0 +1,338 @@
...
+     * @return self
+     */
+    public static function withOptions(array $options = []) : self
+    {
+        $self = new static;
+
+        collect($options)
+            ->each(fn($value, $option) => method_exists($self, 'set' . ucfirst($option)) 
+                ? $self->{'set' . ucfirst($option)}($value) 
+                : null
```
asmecher, line 113: What do you think about throwing an exception when a matching setter function can't be found?

Domains/Jobs/WorkerConfiguration.php
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /**
+     * Get the worker options
+     *
+     * @return array
+     */
+    public function getWorkerOptions() : array 
+    {
+        return [
```
asmecher, line 126: I wonder if this could be rewritten to use named arguments (and an associative array) rather than a long positional list.

Domains/Jobs/WorkerConfiguration.php
```diff
@@ -0,0 +1,338 @@
...
+    /**
+     * Get the number of seconds to wait before retrying the job
+     */
+    public function getBackoff(): int
+    {
+        return $this->backoff;
+    }
+
+    /**
+     * The maximum amount of RAM the worker may consume.
```
asmecher, line 178: Maybe specify in the self-documentation: is the unit in MB?

classes/core/PKPContainer.php
```diff
@@ -344,6 +342,16 @@ public function runningUnitTests()
     {
         return false;
     }
+
+    /**
+     * Determine if the application is currently down for maintenance.
+     *
+     * @return bool
+     */
+    public function isDownForMaintenance()
```
asmecher, line 351: Is there a reason for the `isDownForMaintenance` / `isUnderMaintenace` discrepancy?

classes/core/PKPQueueProvider.php
```diff
@@ -18,53 +18,162 @@
...
+            $connection,
+            $queue,
+            $this->getWorkerOptions($workerOptions)
+        );
+    }
+    
+    /**
+     * Run the queue worker to process queue the jobs
+     */
+    public function runJobsInQueue(): void
```
asmecher, line 88: This function only runs a single job, right? `runJobsInQueue` might be misleading.

classes/task/ProcessQueueJobs.php
asmecher: I think it should be possible to specify a fully qualified class name in registry/scheduledTasks.xml, in place of the `dot.separated.class.names` we have relied upon prior to adding classloading support.

classes/core/PKPQueueProvider.php
```diff
@@ -18,53 +18,162 @@
...
+        parent::register();
+
+        $this->registerDatabaseConnector(app()->get(\Illuminate\Queue\QueueManager::class));
+
+        $this->app->bind(
+            JobRepositoryInterface::class,
+            JobRepository::class
+        );
+
+        // register_shutdown_function([$this, 'runJobsAtShutdown']);
```
asmecher, line 152: May as well clear out the removed code!

asmecher (commented):
Looks good, @touhidurabir! A few initial comments.

Domains/Jobs/WorkerConfiguration.php
touhidurabir: Done.

Domains/Jobs/WorkerConfiguration.php
touhidurabir: good idea . done .

classes/task/ProcessQueueJobs.php
touhidurabir: providing the fully qualified namespace also does (for example : `\PKP\task\ProcessQueueJobs` or `PKP\task\ProcessQueueJobs`) not resolve this as throwing exception as such 

```
Cannot instantiate task class.
```

originated from https://github.com/pkp/pkp-lib/blob/b376a074f127e39b737c07787e1ccbf9cfc053ca/classes/cliTool/ScheduledTaskTool.php#L118 . The core of that lies in the defined regex here at https://github.com/pkp/pkp-lib/blob/b376a074f127e39b737c07787e1ccbf9cfc053ca/includes/functions.php#L178 which returning a `false` . 

I think it will be better to set up a new issue and update the code with backward compatibility to handle `.` and `/` class paths and update the schedule tasks as a PR of that issue . That way we can also do bit clean up for the schedule tasks also .

classes/core/PKPQueueProvider.php
touhidurabir: right. as it's handle one job at a time , it should be in singular like `runJobInQueue`. Done fixing .

Domains/Jobs/WorkerConfiguration.php
touhidurabir: @asmecher  returning back an `associative` make sense and updated accordingly. But I am bit confused about the part `rewritten to use named arguments`.  Can you provide bit more details about what kind of changes you are thinking ?

classes/core/PKPContainer.php
touhidurabir: right now there is none . This was done according to the implementation of the `Illuminate\Queue\QueueServiceProvider::registerWorker()` such as 

```php
$this->app->singleton('queue.worker', function ($app) {
    $isDownForMaintenance = function () {
        return $this->app->isDownForMaintenance();
    };

    $resetScope = function () use ($app) {
        if (method_exists($app['log']->driver(), 'withoutContext')) {
            $app['log']->withoutContext();
        }

        if (method_exists($app['db'], 'getConnections')) {
            foreach ($app['db']->getConnections() as $connection) {
                $connection->resetTotalQueryDuration();
                $connection->allowQueryDurationHandlersToRunAgain();
            }
        }

        $app->forgetScopedInstances();

        return Facade::clearResolvedInstances();
    };

    return new Worker(
        $app['queue'],
        $app['events'],
        $app[ExceptionHandler::class],
        $isDownForMaintenance,
        $resetScope
    );
});
```

basically a lot of laravel's core service providers depend on a container method `isDownForMaintenance()` which we will be need to override as it's not defined out container . This can be removed right now as I am overriding that method as it also require a logger and we dont have one but eventually we will need that method to be included in our container if we want to avoid overriding for more services that we may include in future .

classes/core/PKPQueueProvider.php
touhidurabir: we can and will unless we decide to implement something similar proposed here https://github.com/pkp/pkp-lib/issues/8043#issuecomment-1214777409 .

classes/task/ProcessQueueJobs.php
asmecher: Can you try this again without using the `class_alias`? It should be possible to just use the FQCN (per my comment on the OJS repo: https://github.com/pkp/ojs/pull/3476/files#r951951685) and if it's not working then it means I have to fix something over on pkp/pkp-lib#6091. This has probably been fixed since your last comment a week ago -- my apologies!

Domains/Jobs/WorkerConfiguration.php
asmecher: When this gets passed into the `WorkerOptions` constructor via the splat (`...`) operator, the array gets unpacked into a long list of parameters. We're trying to avoid long positional parameter lists in function calls, especially where the positions of them are arbitrary. Making this an associative array, and using named arguments, makes it less likely that some of the parameters could e.g. get put in the list in the wrong order since the function call will depend on string keys rather than array positions.

Unfortunately, now that I search it, applying `...` to keyed arrays is only available [starting with PHP8.1](https://php.watch/versions/8.1/spread-operator-string-array-keys) -- so on PHP8.0 I think [this line](https://github.com/pkp/pkp-lib/pull/8133/files#diff-d8bccac695c8cacb01f4be531434b57599b8747080dd26b56e8f1cb6d5568274R68) will raise an error. But I think you could work around that using https://www.php.net/manual/en/function.call-user-func-array.php.

asmecher (changes requested):
@touhidurabir, this is looking great -- I only have a couple of minor comments; if you can review those, then kick off a fresh set of Travis tests, I'll get this merged ASAP :+1:

classes/task/ProcessQueueJobs.php
touhidurabir: It's working fine now . Fixed .

Domains/Jobs/WorkerConfiguration.php
touhidurabir: > applying `...` to keyed arrays is only available [starting with PHP8.1](https://php.watch/versions/8.1/spread-operator-string-array-keys)

weirdly it is working on my set up which use `PHP 8.0.19` so I thought it's added to PHP8.0+. 

> But I think you could work around that using https://www.php.net/manual/en/function.call-user-func-array.php.

I don't think using the `call_user_func_array` is the right one here for us as it's take first param as a `callable` but we need to pass those params to the `\Illuminate\Queue\WorkerOptions::__construct` which has no other method to set those after the instance generated . I thought about extending the `\Illuminate\Queue\WorkerOptions` to have a static constructor so that we can passed named agrs but that seems like a bit too much workaround. Also the `PKP\Domains\Jobs\WorkerConfiguration::getWorkerOptions` set the params ordering to make it compatible with `\Illuminate\Queue\WorkerOptions::__construct` and that is why we have that class so we should not directly pass params to `\Illuminate\Queue\WorkerOptions` but build up those through the `PKP\Domains\Jobs\WorkerConfiguration`.  For now to make it compatible with PHP8.0+, I have extracted the values through `array_values`. 

Or may be I am missing something ?

classes/core/PKPQueueProvider.php
touhidurabir: Done . Completely removed the functionality to remove the run jobs as through `register_shutdown_function`.

Domains/Jobs/WorkerConfiguration.php
asmecher: Hmm, you're right -- unpacking associative arrays as named arguments does appear to work with PHP8.0; I've tested with 8.0.2, which is pretty old. I think we're OK to continue with what you've proposed :+1:

Domains/Jobs/WorkerConfiguration.php
```diff
@@ -0,0 +1,340 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * @file Domains/Jobs/Job.php
```
asmecher, line 5: Self-documentation needs update

asmecher (changes requested):
Just one small tweak -- but we also have to bash out an agreement on https://github.com/pkp/pkp-lib/issues/8043#issuecomment-1225759269 before this should be merged.

classes/task/ProcessQueueJobs.php
```diff
@@ -0,0 +1,55 @@
...
+
+        $jobQueue = app('pkpJobQueue');
+
+        $jobBuilder = $jobQueue->getJobModelBuilder();
+
+        if ($jobBuilder->count() <= 0) {
+            return true;
+        }
+
+        while ($jobBuilder->count()) {
```
asmecher, line 49: @touhidurabir, I think we agreed you were going to add a maximum job limit for the sake of avoiding restrictive execution time limits when relying on the Acron plugin. I'll hold off the merge until your say-so on that part, thanks!

touhidurabir: @asmecher improvement done . check it changes at https://github.com/touhidurabir/pkp-lib/commit/c6591eb61fee87909737b464e773ee4b5bc4e178

Issue #8093, opened 2022-07-11
UserGroup Entity: Change UserGroupDAO to Repository Pattern


defstat:
UserGroup Entity follows the deprecated DAO class (`UserGroupDAO`) to perform the code actions necessary for UserGroups. 

This issue will cover the appropriate changes for the UserGroups (and possibly related Entities like `UserGroupAssignments`) to follow the active Repository pattern instead.  

---
**Current PRs**

**Branch `main`**
[OJS]: pkp/ojs#3486
[OMP]: pkp/omp#1171
[OPS]: pkp/ops#325
[pkp-lib]: #8154 

**Plugins**
citationStyleLanguage: pkp/citationStyleLanguage#98
quickSubmitPlugin: pkp/quickSubmit#59

---
For documentation purposes I am adding the file bellow, pairing the removed `UserGroupDAO` functions to the new Repository pattern respective ones.

[UserGroupDAO-2-UserGroupRepo.txt](https://github.com/pkp/pkp-lib/files/9693359/UserGroupDAO-2-UserGroupRepo.txt)

defstat:
@asmecher I have resolved the PRs review comments. Waiting for the tests to run on [OJS](https://app.travis-ci.com/github/pkp/ojs/builds/254835640).

asmecher:
All merged to `main` -- thanks, @defstat, this is great! This just needs some documentation at the head of the issue so that @NateWr can schedule it for inclusion in the release notebook for 3.4; I'll leave this issue open until that's in place.

asmecher:
@defstat, it also looks like there's still a call to `UserGroupDAO` in https://github.com/pkp/citationStyleLanguage/blob/main/pages/CitationStyleLanguageHandler.inc.php#L138; could you refactor this?

ajnyga:
Quicksubmit is affected by this and broken at the moment

defstat:
> Quicksubmit is affected by this and broken at the moment

@ajnyga could you please provide the reproduction steps and possible php errors?

asmecher:
@defstat, I see in `QuickSubmitForm::initData`...
```php
$userGroupAssignmentDao = DAORegistry::getDAO('UserGroupAssignmentDAO');
$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
```
These DAOs no longer exist, which will cause fatal errors when the quickSubmit form is initialized.

defstat:
@asmecher @NateWr 

[...]

PR #8154, opened 2022-08-11
[pkp-lib][main] #8093 UserGroupDAO, UserGroupAssignmentDAO functions replaced

defstat:


classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+        if (isset($this->offset)) {
+            $q->offset($this->offset);
+        }
+
+        switch ($this->orderBy) {
+             case self::ORDERBY_ROLE_ID:
+                $q->orderBy('a.role_id', 'asc');
+                break;
+            case self::ORDERBY_ID:
+            default:
```
asmecher, line 278: I think I would prefer...
- Initialize `$orderBy` to `null` at the top of the class (for no sorting)
- Have `default:` apply no sorting at al

This would not apply sorting where it's not needed (better for performance) and allow for subclasses or plugins to extend it with additional options if needed.

classes/userGroup/DAO.php
```diff
@@ -0,0 +1,173 @@
...
+    {
+        parent::_delete($userGroup);
+    }
+
+    /**
+     * @copydoc EntityDAO::deleteById()
+     */
+    public function deleteById(int $userGroupId)
+    {
+        DB::beginTransaction();
```
asmecher, line 138: I'd recommend setting `user_user_groups` and `user_group_stage` to onDelete('cascade')`. That way you could get rid of this entire function.

classes/userGroup/Repository.php
```diff
@@ -0,0 +1,469 @@
...
+            $errors = $schemaService->formatValidationErrors($validator->errors());
+        }
+
+        HookRegistry::call('UserGroup::validate', [$errors, $userGroup, $props, $allowedLocales, $primaryLocale]);
+
+        return $errors;
+    }
+
+    /**
+     * @copydoc \PKP\services\entityProperties\EntityWriteInterface::add()
```
asmecher, line 135: These `copydoc`s don't add anything; they can be removed.

classes/userGroup/Repository.php
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    /**
+    * Delete all user groups assigned to a certain context by contextId
+    *
+    * @param int $contextId
+    */
+    public function deleteByContextId($contextId)
+    {
+        // I may have to use transactions here
```
asmecher, line 180: I'd like to figure out a consistent policy on transactions before we start to use them. We can't nest transactions, so applying them at one level means we can't apply them on another level.

classes/userGroup/Repository.php
```diff
@@ -0,0 +1,469 @@
...
+        $userGroupIds = Repo::userGroup()->getIds($collector);
+        foreach ($userGroupIds as $userGroupId) {
+            $this->dao->deleteById($userGroupId);
+        }
+    }
+
+    /**
+    * Get all user groups assigned to a certain context by contextId
+    */
+    public function getByContextId(int $contextId)
```
asmecher, line 191: I'm on the fence about whether to remove these functions now, in favour of consistently using `Collector`s in calling code. Since we're moving this function to a different class, this is a breaking change, so it makes sense to get rid of them now. What do you think?

classes/userGroup/Repository.php
```diff
@@ -0,0 +1,469 @@
...
+     * @param int $userGroupId The user group ID
+     *
+     * @return array
+     */
+    public function getAssignedStagesByUserGroupId(int $contextId, int $userGroupId) : array
+    {
+        return UserGroupStage::withContextId($contextId)
+            ->withUserGroupId($userGroupId)
+            ->pluck('stage_id')
+            ->toArray();
```
asmecher, line 362: I'm leaning towards having these functions return `Collection` objects, and only force conversion to arrays where it's needed.

classes/userGroup/UserGroup.php
```diff
@@ -134,12 +137,11 @@ public function getPermitSelfRegistration()
 
     /**
      * Set the "permit self-registration" flag
-     *
-     * @param bool $permitSelfRegistration
      */
-    public function setPermitSelfRegistration($permitSelfRegistration)
+    public function setPermitSelfRegistration(?bool $permitSelfRegistration)
```
asmecher, line 141: Is the `null` state necessary? It looks to me like this is always `true` or `false`.

classes/userGroup/UserGroup.php
```diff
@@ -242,15 +244,10 @@ public function getPermitMetadataEdit()
 
     /**
      * Setter for permitMetadataEdit attribute.
-     *
-     * @param bool $permitMetadataEdit
      */
-    public function setPermitMetadataEdit($permitMetadataEdit)
+    public function setPermitMetadataEdit(?bool $permitMetadataEdit)
```
asmecher, line 248: Same as above -- is the `null` state necessary?

classes/userGroup/UserGroup.php
```diff
@@ -242,15 +244,10 @@ public function getPermitMetadataEdit()
...
     {
-        $this->setData('permitMetadataEdit', $permitMetadataEdit);
+        if (isset($permitMetadataEdit))
+            $this->setData('permitMetadataEdit', $permitMetadataEdit);
     }
-}
-
-if (!PKP_STRICT_MODE) {
-    class_alias('\PKP\security\UserGroup', '\UserGroup');
```
asmecher, line 255: I'd recommend keeping the `class_alias` call for the sake of legacy plugins.

classes/userGroup/relationships/UserGroupStage.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @file classes/userGroup/relationships/UserUserGroup.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class \PKP\userGroup\UserGroup
+ * @ingroup userGroup
```
asmecher, line 11: Now that we're using namespaces, I've been removing `@ingroup` / `@defgroup` calls.

classes/userGroup/relationships/UserGroupStage.php
```diff
@@ -0,0 +1,99 @@
...
+    public function userGroup(): Attribute
+    {
+        return Attribute::make(
+            get: fn($value, $attributes) => Repo::userGroup()->get($attributes['user_group_id']),
+            set: fn($value) => $value->getId()
+        );
+    }
+
+    //
+    // The following two functions permit lowerCamelCase accessors in place of snake_case column names
```
asmecher, line 54: This is probably borrowed from my example Eloquent model; there are 3 of them in this case, but best to just remove these comments, I think.

classes/userGroup/relationships/UserGroupStage.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return Attribute::make(
+            get: fn($value, $attributes) => $attributes['context_id'],
+            set: fn($value) => ['context_id' => $value]
+        );
+    }
+
+    //
+    // The following two functions allow for queries to be executed that don't require the database column
+    // names to be exposed in calling code (as ->where would do). Called e.g.: UserUserGroup::withUserId(123)
```
asmecher, line 83: Likewise, these comments can be removed.

classes/userGroup/relationships/UserGroupStage.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+/**
+ * @file classes/userGroup/relationships/UserUserGroup.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class \PKP\userGroup\UserGroup
```
asmecher, line 10: The `@class` tags are supposed to be for Doxygen to identify the class that's being documented in the block, so they should always match the classname of the class being implemented in the file. (We'll probably move to PHPDocumentor at some point, which can intuit the class name -- but until then it's probably best to stay consistent.) There are a number of classes in this PR that have `@class` tags with the wrong classname in them.

controllers/grid/settings/roles/form/UserGroupForm.php
```diff
@@ -257,13 +257,12 @@ public function execute(...$functionParams)
...
     {
         $contextId = $this->getContextId();
-        $userGroupDao = DAORegistry::getDAO('UserGroupDAO'); /** @var UserGroupDAO $userGroupDao */
 
         // Current existing workflow stages.
         $stages = WorkflowStageDAO::getWorkflowStageTranslationKeys();
 
         foreach (array_keys($stages) as $stageId) {
-            $userGroupDao->removeGroupFromStage($contextId, $userGroupId, $stageId);
+           Repo::userGroup()->removeGroupFromStage($contextId, $userGroupId, $stageId);
```
asmecher, line 265: Wrong indentation (I suspect the code formatter isn't running as it should prior to commits)

tests/classes/core/PKPComponentRouterTest.php
```diff
@@ -271,6 +271,7 @@ public function testRoute()
         // with a reference to the request object as the first argument.
         Registry::set('request', $this->request);
         $user = new \PKP\user\User();
+        $user->setData('id', 0);
```
asmecher, line 274: Could you add a comment documenting why this is required? It's probably something that'll cause confusion later.

asmecher (changes requested):
Looks very good, @defstat! I have lots of small changes to propose but overall this is very promising. It did cause me to think about some areas where we don't have strong conventions yet, especially around e.g. Iterable / array / LazyCollector type hints. It's not necessary to get all of that 100% consistent before we get this merged; it'll take us a while to land on the right mix and apply it everywhere.

classes/userGroup/Repository.php
jonasraoni: Transactions can be nested using a [savepoint](https://www.tutorialspoint.com/mysql/mysql_savepoint.htm).
I was checking one issue in Laravel, and looks like they've got it implemented (https://github.com/laravel/framework/issues/1823#issuecomment-148954287), but better to test 😂

classes/components/forms/publication/PKPPublicationLicenseForm.php
```diff
@@ -34,9 +35,8 @@ class PKPPublicationLicenseForm extends FormComponent
      * @param array $locales Supported locales
      * @param Publication $publication The publication to change settings for
      * @param Context $context The publication's context
-     * @param array $userGroups User groups in this context
```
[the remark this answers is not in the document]
defstat: Changed here.

classes/components/listPanels/PKPSelectReviewerListPanel.php
```diff
@@ -188,7 +188,7 @@ public function getItemsMax()
      *
      * @return array
```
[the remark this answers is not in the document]
defstat: removed

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -449,11 +449,14 @@ public function getUsersNotAssignedAsReviewers($args, $request)
         $context = $request->getContext();
         $term = $request->getUserVar('term');
 
-        $userGroupDao = DAORegistry::getDAO('UserGroupDAO'); /** @var UserGroupDAO $userGroupDao */
-        $users = $userGroupDao->getUsersNotInRole(Role::ROLE_ID_REVIEWER, $context->getId(), $term);
+        $users =Repo::user()->getCollector()
```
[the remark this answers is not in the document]
defstat: Correct - space added

classes/core/DataObject.php
```diff
@@ -23,6 +23,7 @@
 
 class DataObject
 {
+    public $timestamps = false;
```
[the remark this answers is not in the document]
defstat: No it is not. It's part of a test I tried. Removed.

classes/core/PKPPageRouter.php
```diff
@@ -501,29 +502,24 @@ public function redirectHome($request)
...
         if ($context = $this->getContext($request, 1)) {
             // The user is in the context, see if they have zero or one roles only
-            $userGroups = $userGroupDao->getByUserId($userId, $context->getId());
-            $firstUserGroup = $userGroups->next();
-            $secondUserGroup = $userGroups->next();
-            if (!$secondUserGroup) {
-                if (!$firstUserGroup || $firstUserGroup->getRoleId() == Role::ROLE_ID_READER) {
-                    return $request->url(null, 'index');
-                }
+            $userGroups = Repo::userGroup()->userUserGroups($userId, $context->getId());
```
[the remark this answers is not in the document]
defstat: Correct. Changed to `if (!isset($userGroup) || ($userGroup->getRoleId() == Role::ROLE_ID_READER)) {`

classes/emailTemplate/DAO.php
```diff
@@ -278,8 +278,8 @@ public function installEmailTemplates(
 
             DB::table('email_templates_default')->insert([
                 'email_key' => $attrs['key'],
-                'can_disable' => isset($attrs['can_disable']) ? ((int) $attrs['can_disable']) : 0,
-                'can_edit' => isset($attrs['can_edit']) ? ((int) $attrs['can_edit']) : 0,
+                'can_disable' => (array_key_exists('can_disable', $attrs) && isset($attrs['can_disable'])) ? ((int) $attrs['can_disable']) : 0,
```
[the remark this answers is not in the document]
defstat: Correct, I got a conflict there and resolved it like so. 
Changed to the `main` branch's code.

classes/publication/PKPPublication.php
```diff
@@ -149,7 +150,7 @@ public function getTitles()
      *
      * @return string
      */
-    public function getAuthorString($userGroups, $includeInBrowseOnly = false)
+    public function getAuthorString(LazyCollection $userGroups, $includeInBrowseOnly = false)
```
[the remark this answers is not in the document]
defstat: - I changed the `@param` to `\Traversable<UserGroup>` - I am trying to use something like "`Generics`" in documentation where ever I can.
- I changed the function's signature to `\Traversable`

I understand that the `\Traversable` is more descriptive for the actual use of the `$userGroups` inside the function (only in `foreach` clause), and can lead to less issues regarding possible fatal errors that may appear because we pass an `array` for example, instead of a `LazyCollection`. I used `LazyCollection`, though, to force the use of a `Collection` object (for the reasons you described above), and also to be able to use the functions and possibilities of a `[Collection] (https://laravel.com/docs/9.x/collections)` if necessary.

classes/submission/maps/Schema.php
```diff
@@ -40,8 +39,8 @@ class Schema extends \PKP\core\maps\Schema
     /** @copydoc \PKP\core\maps\Schema::$schema */
     public string $schema = PKPSchemaService::SCHEMA_SUBMISSION;
 
-    /** @var UserGroup[] The user groups for this context. */
-    public array $userGroups;
+    /** @var LazyCollection The user groups for this context. */
+    public LazyCollection $userGroups;
```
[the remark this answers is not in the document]
defstat: Agreed. Changed to `Collection`. (elsewhere also)

classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * @file classes/userGroup/Collector.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class userGroup
+ *
+ * @brief A helper class to configure a Query Builder to get a collection of announcements
```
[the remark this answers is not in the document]
defstat: Changed

classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+    public const ORDERBY_ID = 'id';
+
+    /** @var string|null The default orderBy value for authors collector */
+    public $orderBy = null;
+
+    /** @var DAO */
+    public $dao;
+
+    /** @var array|null */
+    public $ids = null;
```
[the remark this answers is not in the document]
defstat: Changed

classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+    public ?int $count = null;
+
+    public ?int $offset = null;
+
+    public function __construct(DAO $dao)
+    {
+        $this->dao = $dao;
+    }
+
+    /** @copydoc DAO::getCount() */
```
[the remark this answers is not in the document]
defstat: Removed

classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+        $this->offset = $offset;
+        return $this;
+    }
+
+    /**
+     * @copydoc CollectorInterface::getQueryBuilder()
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $q = DB::table('user_groups as a');
```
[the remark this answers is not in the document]
defstat: added like so:

```
$q = DB::table('user_groups as a')
            ->select('a.*');
```

classes/userGroup/Repository.php
defstat: Removed from 
- public function add
- public function edit
- public function delete

classes/userGroup/Repository.php
defstat: I had the same question when I wrote it. It makes sense to remove it.

I retained it because they are being used in a lot of places in the code (33 hits in 21 files), thinking that it would be better to contain the change in a certain function. (especially before being sure that it works as it should).

classes/userGroup/Repository.php
defstat: I changed that.

classes/userGroup/UserGroup.php
defstat: You are right. the `$permitSelfRegistration` should not be null.

I changed that in order to "fix" cases where there where a code segment that tried to set `$permitSelfRegistration` to `null`. Instead of adding the check when the `setPermitSelfRegistration` is called, I added this business logic to the actual `setPermitSelfRegistration` function. 

I am wondering whether, if changing the `?bool` to `bool`, the code will fail if the code will try to pass null value there. Also I am wondering how the `->default` value of a column will be filled in the DB if the object has a value of `null` for the specified param, or if the object has no value for the specified param. Will the default db value be applied to both of those cases?

classes/userGroup/UserGroup.php
defstat: Added this

```
if (!PKP_STRICT_MODE) {
    class_alias('\PKP\security\UserGroup', '\UserGroup');
}
```

at the end of the class after the closing `}`

classes/userGroup/relationships/UserGroupStage.php
defstat: Removed from the newly created files in `/classes/userGroup/`

classes/userGroup/relationships/UserGroupStage.php
defstat: Removed from 
- `UserGroupStage`
- `UserUserGroup`

classes/userGroup/relationships/UserGroupStage.php
defstat: Removed from 
- `UserGroupStage`
- `UserUserGroup`

classes/userGroup/relationships/UserGroupStage.php
defstat: I think I fixed it for the newly created files in `/classes/userGroup/`

controllers/grid/settings/roles/form/UserGroupForm.php
defstat: Fixed

tests/classes/core/PKPComponentRouterTest.php
defstat: Added the following comment 

```
/*
* Set the id of the user here to something other than null in order for the UserRolesRequiredPolicy 
* to be able to work as it supposed to.
* Specifically, the UserRolesRequiredPolicy::effect calls the getByUserIdGroupedByContext function
* which needs a userId that is not nullable.
*/
```

classes/userGroup/DAO.php
defstat: Added the foreignKeys 

```
        Schema::table('user_group_settings', function (Blueprint $table) {
            $table->foreign('user_group_id')->references('user_group_id')->on('user_groups')->onDelete('cascade');
        });

        Schema::table('user_user_groups', function (Blueprint $table) {
            $table->foreign('user_group_id')->references('user_group_id')->on('user_groups')->onDelete('cascade');
        });

        Schema::table('user_group_stage', function (Blueprint $table) {
            $table->foreign('user_group_id')->references('user_group_id')->on('user_groups')->onDelete('cascade');
        });
```
to 
- Upgrade Migration (up, down)
- Installation

Removed the `deleteById` function

classes/userGroup/Collector.php
defstat: - `$orderBy` was already initialized to `null` (am I missing something?)
- Change the Switch clause removing the `default:` case.

classes/userGroup/Collector.php
```diff
@@ -0,0 +1,288 @@
...
+        $q->when($this->isRecommendOnly !== null, function (Builder $q) {
+            $q->whereIn('a.user_group_id', function (Builder $q) {
+                $q->select('user_group_id')
+                    ->from($this->dao->settingsTable)
+                    ->where('setting_name', '=', 'recommendOnly')
+                    ->where('setting_value', $this->isRecommendOnly);
+            });
+        });
+
+        if ($this->isDefault) {
```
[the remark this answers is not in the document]
defstat: Changed. Also changed the `\userGroup\Repository::getByRoleIds` to follow the change.

classes/userGroup/Collector.php
asmecher: > am I missing something?

Nope, it looks good!

classes/userGroup/UserGroup.php
asmecher: Yes, passing a `null` to a `bool`-typehinted parameter will fail; this is why we can't just batch-add type hints to classes without expecting a lot of breakage. I think we should be getting stricter comprehensively and consistently -- business logic, database, storage code, everything. The database definition of columns is usually pretty accurate (e.g. `permit_self_registration` does not permit `null` data, which is good).

If the PHP object has a `null` value, then the `INSERT` statement will pass the `NULL` along to the database. Inserting a `NULL` value in a `NOT NULL` column will cause an error, as it should. So when creating a new `UserGroup` object, the calling code will be responsible for setting an initial value of `true` or `false` for `setPermitSelfRegistration`. But that's not new -- it's always been the case.

classes/core/PKPPageRouter.php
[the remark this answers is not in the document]
defstat: Here there where a problem with the migration - The code change (I mean the initial, not the syntax change) did not reflect the business logic of the particular code segment. That was causing the following issue (among others): When the `first()` function was called, and the underlying connection was on a `MySQL` DB, then the returned role was different than the one that was returned when the underlying connection was on a `PostgreSQL` DB. 

Fixed the segment to reflect the initial business logic.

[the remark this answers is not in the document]
asmecher: Ah, I was barking up the wrong tree. Glad you found it!

classes/submission/maps/Schema.php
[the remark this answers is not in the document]
defstat: @asmecher this was changed back to `LazyCollection`. I thought that `LazyCollection` extends `Collection` but that is not the case.

classes/userGroup/UserGroup.php
defstat: Changed

classes/userGroup/UserGroup.php
defstat: Changed

classes/userGroup/Repository.php
defstat: Changed - `getByContextId` function removed

classes/user/Collector.php
```diff
@@ -241,6 +242,16 @@ public function filterExcludeSubmissionStage(int $submissionId, int $stageId, in
         return $this;
     }
 
+    /**
+     * Retrieve a set of users not assigned to the given roles
+     * (Replaces UserGroupDAO::getUsersNotInRole)
```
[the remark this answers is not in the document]
asmecher: Don't forget to add the implications that coders will need to know to the top of #8093!

Issue #8160, opened 2022-08-11
Flaky test failing most of OMP builds


jonasraoni:
**Describe the bug**
There's a flaky test at the `MdawsonSubmission.spec.js` which is breaking most of the OMP builds, it always ends up with the file `Segmentation of Vascular Ultrasound Image Sequences.` missing a genre assignment, which blocks the save button.

I couldn't simulate it locally, only in Travis.

**What application are you using?**
OMP /main

**PRs**
- https://github.com/pkp/pkp-lib/pull/8161
- https://github.com/pkp/omp/pull/1172 [TESTS ONLY]

**Additional information**
![image](https://user-images.githubusercontent.com/361921/184238689-edd185a3-7154-4a5e-a299-9d287fd605cc.png)

jonasraoni:
@asmecher The build has passed through the flaky test around 7x, so I think it's ok to be reviewed.

asmecher:
Merged, thanks!

asmecher:
@jonasraoni, I've reverted this -- it appears to have broken OPS tests. Could you have a look?

jonasraoni:
Yep, I'll check later!

jonasraoni:
The tests passed in all applications, so I've merged.

PR #8161, opened 2022-08-11
pkp/pkp-lib#8160 Fixed flaky test

jonasraoni:


cypress/support/commands.js
```diff
@@ -189,20 +197,30 @@ Cypress.Commands.add('createSubmission', (data, context) => {
 				cy.get('input[type=file]').attachFile(
 					{fileContent, 'filePath': file.fileName, 'mimeType': 'application/pdf', 'encoding': 'base64'}
 				);
-				var $row = cy.get('a:contains("' + file.fileName + '")').parents('.listPanel__item');
+				cy.wait('@fileUploaded').its('status').should('eq', 200);
+				const $row = cy.get('a:contains("' + file.fileName + '")').parents('.listPanel__item');
+				cy.wait(1000);
```
asmecher, line 202: Is this wait a critical part of the test being improved? It sounds from the original report like the defined route (`@fileUploaded`) might be enough... I consider every one of our `cy.wait` calls a FIXME, and there are lots!

jonasraoni: Unfortunately it was needed, I guess some text is appearing before the handler is attached to the button, so the click fails, I probably just can reduce the value to 200ms.

I think that a proper fix will have to wait for a signal from `Vue`, perhaps something generic using the [lifecycle events](https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram)... Like checking for a counter ruled by "created increases a counter, mounted decreases it"... But then, there's also jQuery, timers, etc. 😣

asmecher: OK! No need to drop it from 1000 to 250; let's deal with these systematically later, possibly when we have time to revisit our seeding/testing strategy.

PR #8217, opened 2022-08-29
[pkp-lib][stable-3_3_0] #7249 Force lowercase strings on user email and username (Forms, Migration, DAO)

defstat:


asmecher:
@defstat, this is obsolete, right?

PR #8220, opened 2022-08-30
[pkp-lib][main] #6610 Native import export Security issue fix

defstat:


templates/plugins/importexport/resultsExport.tpl
```diff
@@ -23,6 +23,7 @@
 		<form id="exportIssuesXmlForm" class="pkp_form" action="{plugin_url path="downloadExportFile"}" method="post">
 			{csrf}
 			<input type="hidden" name="downloadFilePath" id="downloadFilePath" value="{$exportPath}" />
+			<input type="hidden" name="downloadFileName" id="downloadFileName" value="{$exportFileName}" />
```
asmecher, line 26: Both these hidden variables need to be passed through `|escape`.

classes/plugins/ImportExportPlugin.php
```diff
@@ -412,26 +425,35 @@ public function getBounceTab($request, $title, $bounceUrl, $bounceParameterArray
      *
      * @param string $exportFileName
      */
-    public function downloadExportedFile($exportFileName)
+    public function downloadExportedFile(string $exportFileName, string $exportFilePath, PKPImportExportDeployment $deployment)
     {
+        $date = DateTime::createFromFormat(self::EXPORT_FILE_DATE_PART_FORMAT,  $exportFilePath);
```
asmecher, line 430: I'm finding `$exportFilePath` to be confusing; it's actually supposed to be a string-formatted date, not a path. Elsewhere it's called `$filename`, elsewhere `$downloadPath`. I'd recommend standardizing on something sensible, proabably relating to the date.

classes/plugins/ImportExportPlugin.php
```diff
@@ -412,26 +425,35 @@ public function getBounceTab($request, $title, $bounceUrl, $bounceParameterArray
...
+        return true;
     }
 
     /**
      * Create file given it's name and content
      *
      * @param string $filename
      * @param string $fileContent
-     * @param Context $context
+     * @param \Context $context
```
asmecher, line 448: Best not to refer to classes in the root namespace, since we'll be removing the aliases sometime in the future. Here, for example, it should be `\PKP\context\Context`. (Also below.)

asmecher (changes requested):
There's more leeway in the `main` branch to make this code sensible, so this is a good chance to clean up some of the variables/calls here that refer to the same thing in several different ways; see comments. Thanks!

templates/plugins/importexport/resultsExport.tpl
defstat: Fixed.

classes/plugins/ImportExportPlugin.php
defstat: Yes, completely agreed. Changed.

classes/plugins/ImportExportPlugin.php
defstat: Fixed

PR #8226, opened 2022-08-31
[pkp-lib][stable-3_3_0] #7249 Add lowercase search to user username and email

defstat:


classes/user/UserDAO.inc.php
```diff
@@ -513,9 +519,11 @@ function userExistsByUsername($username, $userId = null, $allowDisabled = true)
 	 * @return boolean
 	 */
 	function userExistsByEmail($email, $userId = null, $allowDisabled = true) {
+		$emailLower = strtolower($email);
```
asmecher, line 522: Mixing PHP-side string operations with database-side string operations may cause trouble with locales and collations. I'd recommend keeping everything in one place -- e.g. `LOWER(?)` in the database query.

classes/migration/CommonMigration.inc.php
```diff
@@ -88,10 +88,21 @@ public function up() {
...
-
+		
+		switch (Capsule::connection()->getDriverName()) {
+			case 'mysql':
+				Capsule::schema()->table('users', function (Blueprint $table) {
+					$table->unique(['username'], 'users_username');
+					$table->unique(['email'], 'users_email');
+				});
+				break;
+			case 'pgsql':
```
asmecher, line 100: This will cause problems if there were previously registrations that had mixed cases. The pre-flight check is helpful for this; I'd recommend flagging any cases where there will be collisions in the pre-flight check, and asking the user to resolve them (e.g. with "merge users") *before* running the upgrade.

asmecher: [sorry, that comment is more useful for `main`, not `stable-3_3_0`!)

asmecher:
Since this change requires a migration, it shouldn't be included in `stable-3_3_0`.

PR #8265, opened 2022-09-16
fix: PKPReviewerReviewStep3Form::initData() now calls parent::

prechelt:
My plugin development (based on OJS 3.3) came to a halt when I needed a callback at `reviewerreviewstep3form::initData`.  
The callback never got called, because `ReviewerReviewStep3Form::initData()` does not call `parent::initData();`  
This patch adds this call (and nothing else).

I have checked the other methods that override something from `Form`.
Their required `parent::` calls _are_ present.

I have also provided the equivalent patch in a version 3.4 form in [#8264](https://github.com/pkp/pkp-lib/pull/8264).


asmecher:
Merged both, thanks; I would very much like a better mechanism than the `Form` class implementations that only exist to call hooks, as they're janky and inconsistently called. But for now this should get you moving ahead!

asmecher:
(This will be included starting with 3.3.0-12, due out shortly.)

Issue #6849, opened 2021-03-10
Tool to convert old and apache log files into new format 


asmecher:
Previously the stats framework used Apache-style logs to handle both internal processing of log events generated by OJS, and retroactive processing of log files from another source (like the web server), e.g. for when it was necessary to reconstruct or rebuild the logs.

We are moving away from the internal use of Apache-style log files in order to reduce complexity around resolving actions based only on URLs. However, it would still be good to have a way to ingest Apache logs to rebuild the stats in some fashion.

Proposal: Put together a command-line tool to process Apache logs by matching against a set of URL pattern regular expressions using a configuration like...

```
; {$contextPath} matches context (e.g. journal) URL path
; {$submissionId} matches numeric submission ID
; {$submissionUrlPath} matches textual URL slug
; {$publicationId} matches publication ID
; {$currentPublicationId} matches submission's current publication ID
submission "/{$contextPath}/article/view/{$submissionId}$"
submission "/{$contextPath}/article/view/{$submissionUrlPath}$"
submission "/{$contextPath}/article/view/{$submissionUrlPath}/{$currentPublicationId}"
publication "/{$contextPath}/article/view/{$submissionId}/{$publicationId}$"
publication "/{$contextPath}/article/view/{$submissionUrlPath}/{$publicationId}$"
...
```
...where `submission` and `publication` define the type of stats event the match will map to.

The user will be able to feed logs into the tool using something like:
```
cat /path/to/access_log | php lib/pkp/tools/processUsageLog.php
```
...and the tool will do its best to ingest the events, providing warnings where e.g. a `{$submissionUrlPath}` doesn't match anything because an editor has changed it.

1. PRs:
pkp-lib: https://github.com/pkp/pkp-lib/pull/8271
ojs: https://github.com/pkp/ojs/pull/3594 (only submodule update)
omp: https://github.com/pkp/omp/pull/1235 (only submodule update)
ops: https://github.com/pkp/ops/pull/382 (only submodule update)

2. PRs:
https://github.com/pkp/pkp-lib/pull/8467
https://github.com/pkp/ojs/pull/3644
https://github.com/pkp/omp/pull/1262 (only submodule update)
https://github.com/pkp/ops/pull/404 (only submodule update)

asmecher:
@NateWr and @bozana, from our meeting today -- just a suggestion for the implementation based on what I was picturing during the meeting, details are all up for discussion!

bozana:
PR:
[...]

Issue #3030, opened 2017-11-08
[OJS] The application name for OJS 3 is stated as "ojs2"


MarHerUMR:
I noticed it when a fatalerror() wrote this into the error log: _ojs2: 404 Not Found_

In the [Application.inc.php](https://github.com/pkp/ojs/blob/master/classes/core/Application.inc.php) the name is still set to ojs2.

```
	/**
	 * Get the symbolic name of this application
	 * @return string
	 */
	function getName() {
		return 'ojs2';
	}
```

asmecher:
This is a known issue, but it's cosmetic and a little tricky to fix.  See https://github.com/pkp/pkp-lib/issues/1839.

PR #8271, opened 2022-09-19
pkp/pkp-lib#6849 convert usage stats log file CLI tool

bozana:
s. https://github.com/pkp/pkp-lib/issues/6849

### DIFF
--- a/tools/convertUsageStatsLogFile.php
+++ b/tools/convertUsageStatsLogFile.php
@@ -38,6 +38,14 @@ class ConvertUsageStatsLogFile extends \PKP\cliTool\CommandLineTool
     public const PARSEREGEX = '/^(?P<ip>\S+) \S+ \S+ "(?P<date>.*?)" (?P<url>\S+) (?P<returnCode>\S+) "(?P<userAgent>.*?)"/';
 
     /**
+     * Weather the URL parameters are used instead of CGI PATH_INFO.
+     * This is the former variable 'disable_path_info' in the config.inc.php
+     *
+     * This needs to be set to true if the URLs in the old log file contain the paramteres as URL query string.
+     */
+    public const PATH_INFO_DISABLED = false;
+
+    /**
      * Name of the log file that should be converted into the new format.
      *
      * The file needs to be in the folder usageStats/archive/
@@ -272,16 +280,14 @@ class ConvertUsageStatsLogFile extends \PKP\cliTool\CommandLineTool
 
         $expectedPageAndOp = $this->getExpectedPageAndOp();
 
-        $pathInfoDisabled = Config::getVar('general', 'disable_path_info');
-
         // Apache and usage stats plugin log files come with complete or partial base url,
         // remove it so we can retrieve path, page, operation and args.
         $url = Core::removeBaseUrl($url);
         if ($url) {
-            $contextPaths = Core::getContextPaths($url, !$pathInfoDisabled);
-            $page = Core::getPage($url, !$pathInfoDisabled);
-            $operation = Core::getOp($url, !$pathInfoDisabled);
-            $args = Core::getArgs($url, !$pathInfoDisabled);
+            $contextPaths = $this->getContextPaths($url, !self::PATH_INFO_DISABLED);
+            $page = Core::getPage($url, !self::PATH_INFO_DISABLED);
+            $operation = Core::getOp($url, !self::PATH_INFO_DISABLED);
+            $args = Core::getArgs($url, !self::PATH_INFO_DISABLED);
         } else {
             // Could not remove the base url, can't go on.
             // __('plugins.generic.usageStats.removeUrlError', array('file' => $filePath, 'lineNumber' => $lineNumber))
@@ -367,6 +373,8 @@ class ConvertUsageStatsLogFile extends \PKP\cliTool\CommandLineTool
                 ];
                 $pageAndOp[Application::getContextAssocType()][] = 'index';
                 break;
+            default:
+                throw new Exception('Unrecognized application name.');
         }
         return $pageAndOp;
     }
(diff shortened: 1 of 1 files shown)


tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+use PKP\config\Config;
+use PKP\context\Context;
+use PKP\core\Core;
+use PKP\core\Registry;
+use PKP\db\DAORegistry;
+use PKP\file\FileManager;
+use PKP\submission\Genre;
+use PKP\task\FileLoader;
+
+require(dirname(__FILE__, 4) . '/tools/bootstrap.php');
```
asmecher, line 28: :exploding_head: I didn't know about the 2nd parameter to `dirname`!

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    /**
+     * Get data from the passed log entry.
+     */
+    private function getDataFromLogEntry(string $entry): array
+    {
+        $entryData = [];
+        if (preg_match(self::PARSEREGEX, $entry, $m)) {
+            $associative = count(array_filter(array_keys($m), 'is_string')) > 0;
```
asmecher, line 239: I don't suppose it does any harm, but is the idea here that users might customized `self::PARSEREGEX` to use numeric indexes?

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+     * Get assoc type, page, operation and args from the passed url,
+     * if it matches any that's defined in getExpectedPageAndOp().
+     */
+    private function getUrlMatches(string $url, int $lineNumber): array
+    {
+        $noMatchesReturner = [null, null, null, null, null];
+
+        $expectedPageAndOp = $this->getExpectedPageAndOp();
+
+        $pathInfoDisabled = Config::getVar('general', 'disable_path_info');
```
asmecher, line 275: FYI, the `disable_path_info` option has been marked deprecated since 3.2.0 and I am ready to remove it. It's likely not working well (e.g. with API requests) anyway. If you find yourself doing a lot of coding to accommodate the feature, let me know and I'll prioritize its removal instead!

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+            case 'ops':
+                $pageAndOp = $pageAndOp + [
+                    Application::ASSOC_TYPE_SUBMISSION_FILE => [
+                        'preprint/download'],
+                    Application::ASSOC_TYPE_SUBMISSION => [
+                        'preprint/view']
+                ];
+                $pageAndOp[Application::getContextAssocType()][] = 'index';
+                break;
+        }
```
asmecher, line 370: Please add:
```php
`default: throw new Exception('Unrecognized application name.');
```
This will help us find affected areas if we ever get around to https://github.com/pkp/pkp-lib/issues/3030.

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+                }
+
+                $representationId = (int) $args[1];
+                $galley = Repo::galley()->get($representationId);
+                if (!$galley) {
+                    echo "Represantation (galley or publication format) with the ID {$representationId} does not exist.\n";
+                    break;
+                }
+
+                if (Application::get()->getName() == 'omp') {
```
asmecher, line 399: I'd prefer if this was a `switch` statement explicitly mentioning each valid case, rather than trusting existing differences between applications. For example,
```php
case 'ojs2':
case 'ops':
    break; // noop
case 'omp':
    ...run the conditonal code...
    break;
default: throw new Exception('Unrecognized application name!');
```
This will help pinpoint areas of the code needing attention if e.g. we add another application to the suite in the future.

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+                    if (!isset($args[2])) {
+                        echo "Missing URL parameter.\n";
+                        break;
+                    } else {
+                        $submissionFileId = (int) $args[2];
+                    }
+                } else {
+                    // consider this issue: https://github.com/pkp/pkp-lib/issues/6573
```
asmecher, line 407: Glad to see this attention to detail :)

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+                break;
+            case Application::ASSOC_TYPE_SUBMISSION:
+                $publicationId = null;
+                if (!isset($args[0])) {
+                    echo "Missing URL parameter.\n";
+                    break;
+                }
+                // If the operation is 'view' and the arguments count > 1
+                // the arguments must be: $submissionId/version/$publicationId.
+                // Else, it is not considered hier, as submission abstract count.
```
asmecher, line 446: *here

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+                if ($op == 'view' && count($args) > 1) {
+                    if ($args[1] !== 'version') {
+                        echo "Wrong URL parameter. Expected 'verison' found {$args[1]}.\n";
```
asmecher, line 449: *version

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+                        break;
+                    } elseif (count($args) != 3) {
+                        echo "Missing URL parameter.\n.\n";
```
asmecher, line 452: This could also be missing the publication parameter, right?

tools/convertUsageStatsLogFile.php
```diff
@@ -0,0 +1,570 @@
...
+            $application = Application::get();
+            $applicationName = $application->getName();
+            switch ($applicationName) {
+                case 'ojs2':
+                    $this->setOJSAssoc($assocType, $args, $newEntry);
+                    break;
+                case 'omp':
+                    $this->setOMPAssoc($assocType, $args, $newEntry);
+                    break;
+            }
```
asmecher, line 486:
```php
case 'ops':
    break; // noop
default: throw new Exception('Unrecognized application name!');
```

asmecher (changes requested):
Just some very minor issues, @bozana! Please go ahead and merge when you're ready, thanks.

tools/convertUsageStatsLogFile.php
bozana: I am not sure about numeric indexes :thinking: but yes, the idea here is that users might customized the regex -- above all if their apache log files are in a different format that this here.
I also do not think it can cause any harm -- it is only applied to that file that they want to convert...

tools/convertUsageStatsLogFile.php
bozana: Are the functions Core::getContextPaths, getPage, getOp and getArtgs going to change? Will they still have/consider `$isPathInfo` as parameter? If so I do not have to change much -- I would add a variable at the beginning of the script so that the user can adapt it if the log file contains such URLs. If not I would need to 'copy' those function here in the script...

tools/convertUsageStatsLogFile.php
bozana: Yes...

asmecher: Could you add that to the message?

tools/convertUsageStatsLogFile.php
asmecher: Since you're not using the 3rd parameter to `Core::_getUserVar`, I'd recommend just using `parse_url` directly, e.g. for the two use cases of `Core::_getUserVar` in this PR:
```php
parse_str((string) parse_url($urlInfo, PHP_URL_QUERY), $userVars);
$contextPaths[$key] = $userVars[$contextName] ?? null;
```
...and...
```php
parse_str((string) parse_url($urlInfo, PHP_URL_QUERY), $userVars);
$component = $userVars[$varName] ?? null;
```

bozana: Oh yes, that makes sense! :+1:

PR #8283, opened 2022-09-22
pkp/pkp-lib#7391 user partial updatebale feature added

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#7391

### DIFF
--- a/classes/security/Validation.php
+++ b/classes/security/Validation.php
@@ -488,53 +488,71 @@ class Validation
      *
      * @param int   $administeredUserId     User ID of user to potentially administer
      * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * @param int   $contextId              The journal/context Id
      * 
      * @return int The authorized adminstration level 
      */
-    public static function getAdministrationLevel($administeredUserId, $administratorUserId)
+    public static function getAdministrationLevel(int $administeredUserId, int $administratorUserId, int $contextId = null): int
     {
-        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var RoleDAO $roleDao */
-        $context = Application::get()?->getRequest()?->getContext();
-
         // You can administer yourself
         if ($administeredUserId == $administratorUserId) {
             return self::ADMINISTRATION_FULL;
         }
 
+        $filteredSiteAdminUserGroups = Repo::userGroup()
+            ->getCollector()
+            ->filterByContextIds([\PKP\core\PKPApplication::CONTEXT_SITE])
+            ->filterByRoleIds([Role::ROLE_ID_SITE_ADMIN]);
+
         // You cannot adminster administrators
-        if ($roleDao->userHasRole(\PKP\core\PKPApplication::CONTEXT_SITE, $administeredUserId, Role::ROLE_ID_SITE_ADMIN)) {
+        if ($filteredSiteAdminUserGroups->filterByUserIds([$administeredUserId])->getCount() > 0) {
             return self::ADMINISTRATION_PROHIBITED;
         }
 
         // Otherwise, administrators can administer everyone
-        if ($roleDao->userHasRole(\PKP\core\PKPApplication::CONTEXT_SITE, $administratorUserId, Role::ROLE_ID_SITE_ADMIN)) {
+        if ($filteredSiteAdminUserGroups->filterByUserIds([$administratorUserId])->getCount() > 0) {
             return self::ADMINISTRATION_FULL;
         }
 
         // Check for administered user group assignments in other contexts
         // that the administrator user doesn't have a manager role in.
-        $userGroups = Repo::userGroup()->userUserGroups($administeredUserId);
-        foreach ($userGroups as $userGroup) {
-            if ($userGroup->getContextId() != \PKP\core\PKPApplication::CONTEXT_SITE && !$roleDao->userHasRole($userGroup->getContextId(), $administratorUserId, Role::ROLE_ID_MANAGER)) {
-                // Found an assignment: disqualified.
-                // But also determine if a partial administrate is allowed
-                // if the Administrator User is a Journal Manager in the current context
-                if ($context && $roleDao->userHasRole($context->getId(), $administratorUserId, Role::ROLE_ID_MANAGER)) {
-                    return self::ADMINISTRATION_PARTIAL;
-                }
-                return self::ADMINISTRATION_PROHIBITED;
+        $userGroupsCount = Repo::userGroup()
+            ->userUserGroups($administeredUserId)
+            ->filter(fn($userGroup) => 
+                $userGroup->getContextId() != \PKP\core\PKPApplication::CONTEXT_SITE &&
+                !Repo::userGroup()
+                    ->getCollector()
+                    ->filterByContextIds([$userGroup->getContextId()])
+                    ->filterByUserIds([$administratorUserId])
+                    ->filterByRoleIds([Role::ROLE_ID_MANAGER])
+                    ->getCount()
+            )
+            ->count();
+        
+        if ( $userGroupsCount > 0 ) {
+            // Found an assignment: disqualified.
+            // But also determine if a partial administrate is allowed
+            // if the Administrator User is a Journal Manager in the current context
+            if ($contextId && 
+                Repo::userGroup()
+                    ->getCollector()
+                    ->filterByContextIds([$contextId])
+                    ->filterByUserIds([$administratorUserId])
+                    ->filterByRoleIds([Role::ROLE_ID_MANAGER])
+                    ->getCount()) {
+                return self::ADMINISTRATION_PARTIAL;
             }
+            return self::ADMINISTRATION_PROHIBITED;
         }
 
         // Make sure the administering user has a manager role somewhere
-        $foundManagerRole = false;
-        $roles = $roleDao->getByUserId($administratorUserId);
-        foreach ($roles as $role) {
-            if ($role->getRoleId() == Role::ROLE_ID_MANAGER) {
-                $foundManagerRole = true;
-            }
-        }
-        if (!$foundManagerRole) {
+        $roleManagerCount = Repo::userGroup()
+            ->getCollector()
+            ->filterByUserIds([$administratorUserId])
+            ->filterByRoleIds([Role::ROLE_ID_MANAGER])
+            ->getCount();
+
+        if ( $roleManagerCount <= 0 ) {
             return self::ADMINISTRATION_PROHIBITED;
         }
 
--- a/classes/security/Validation.php
+++ b/classes/security/Validation.php
@@ -26,8 +26,9 @@ use PKP\session\SessionManager;
 
 class Validation
 {
-    /** @var bool A flag to determine if partial adminstration(only user group) is allowed or not */
-    protected static bool $partialAdministration = false;
+    public const ADMINISTRATION_PROHIBITED = 0;
+    public const ADMINISTRATION_PARTIAL = 1;
+    public const ADMINISTRATION_FULL = 2;
     
     /**
      * Authenticate user credentials and mark the user as logged in in the current session.
@@ -428,29 +429,30 @@ class Validation
     /**
      * Check whether a user is allowed to administer another user.
      *
-     * @param int $administeredUserId User ID of user to potentially administer
-     * @param int $administratorUserId User ID of user who wants to do the administrating
-     *
-     * @return bool True IFF the administration operation is permitted
+     * @param int   $administeredUserId     User ID of user to potentially administer
+     * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * @param array $allowedLevels          The allowed adminstration levels
+     * 
+     * @return int The authorized adminstrative level 
      */
-    public static function canAdminister($administeredUserId, $administratorUserId)
+    public static function canAdminister($administeredUserId, $administratorUserId, $allowedLevels = [self::ADMINISTRATION_FULL])
     {
         $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var RoleDAO $roleDao */
         $context = Application::get()?->getRequest()?->getContext();
 
         // You can administer yourself
         if ($administeredUserId == $administratorUserId) {
-            return true;
+            return static::getApplicableAdministrationLevel(self::ADMINISTRATION_FULL, $allowedLevels);
         }
 
         // You cannot adminster administrators
         if ($roleDao->userHasRole(\PKP\core\PKPApplication::CONTEXT_SITE, $administeredUserId, Role::ROLE_ID_SITE_ADMIN)) {
-            return false;
+            return static::getApplicableAdministrationLevel(self::ADMINISTRATION_PROHIBITED, $allowedLevels);
         }
 
         // Otherwise, administrators can administer everyone
         if ($roleDao->userHasRole(\PKP\core\PKPApplication::CONTEXT_SITE, $administratorUserId, Role::ROLE_ID_SITE_ADMIN)) {
-            return true;
+            return static::getApplicableAdministrationLevel(self::ADMINISTRATION_FULL, $allowedLevels);
         }
 
         // Check for administered user group assignments in other contexts
@@ -462,9 +464,9 @@ class Validation
                 // But also determine if a partial administrate is allowed
                 // if the Administrator User is a Journal Manager in the current context
                 if ($context && $roleDao->userHasRole($context->getId(), $administratorUserId, Role::ROLE_ID_MANAGER)) {
-                    static::$partialAdministration = true;
+                    return static::getApplicableAdministrationLevel(self::ADMINISTRATION_PARTIAL, $allowedLevels);
                 }
-                return false;
+                return static::getApplicableAdministrationLevel(self::ADMINISTRATION_PROHIBITED, $allowedLevels);
             }
         }
 
@@ -477,21 +479,28 @@ class Validation
             }
         }
         if (!$foundManagerRole) {
-            return false;
+            return static::getApplicableAdministrationLevel(self::ADMINISTRATION_PROHIBITED, $allowedLevels);
         }
 
         // There were no conflicting roles. Permit administration.
-        return true;
+        return static::getApplicableAdministrationLevel(self::ADMINISTRATION_FULL, $allowedLevels);
     }
 
     /**
-     * Check if partial administration is allowed
+     * Determine the adminstation level
      *
-     * @return bool
+     * @param int $applicableLevel The possible applicable adminstration level
+     * @param array $allowedLevels The allowed applicable adminstration level
+     * 
+     * @return int Authorized adminstration level
      */
-    public static function canPartialAdministrate(): bool
+    protected static function getApplicableAdministrationLevel(int $applicableLevel, array $allowedLevels)
     {
-        return static::$partialAdministration;
+        if ( in_array($applicableLevel, $allowedLevels) ) {
+            return $applicableLevel;
+        }
+
+        return self::ADMINISTRATION_PROHIBITED;
     }
 }
 
--- a/classes/security/Validation.php
+++ b/classes/security/Validation.php
@@ -26,6 +26,9 @@ use PKP\session\SessionManager;
 
 class Validation
 {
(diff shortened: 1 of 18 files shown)


classes/security/Validation.php
```diff
@@ -455,6 +459,11 @@ public static function canAdminister($administeredUserId, $administratorUserId)
         foreach ($userGroups as $userGroup) {
             if ($userGroup->getContextId() != \PKP\core\PKPApplication::CONTEXT_SITE && !$roleDao->userHasRole($userGroup->getContextId(), $administratorUserId, Role::ROLE_ID_MANAGER)) {
                 // Found an assignment: disqualified.
+                // But also determine if a partial administrate is allowed
+                // if the Administrator User is a Journal Manager in the current context
+                if ($context && $roleDao->userHasRole($context->getId(), $administratorUserId, Role::ROLE_ID_MANAGER)) {
+                    static::$partialAdministration = true;
```
asmecher, line 465: I'd recommend against using a static property on `Validation`. Instead, I'd suggest introducing class constants. Two options:

## 1. Return a constant.
Make it so that `canAdminister` returns one of e.g.:
```
public const ADMIN_PROHIBITED = 0;
public const ADMIN_PARTIAL = 1;
public const ADMIN_FULL = 2;
```
(In the future we might use enums for these, but they require at least PHP8.1. Also they aren't extensible by plugins, which limits their use in other cases, though here I think it's OK.)

The downside to this is that calling code that expected a boolean will need to be adapted to check for one of the two true-ish cases -- but this function is rarely used in the system or by plugins, so I'm OK to document that in the notes and change the function's behaviour.

## 2. Specify a kind of administration being checked in the function call:
```php
public static function canAdminister($administeredUserId, $administratorUserId, $requiredLevel = ADMIN_LEVEL_FULL) {
```
In this case calling code can specify what "level" of administrative privileges are required. The default level can match the previous behaviour, so backwards compatibility isn't broken.

controllers/grid/settings/user/form/UserDetailsForm.php
```diff
@@ -265,6 +301,12 @@ public function execute(...$functionParams)
             $this->user->setInlineHelp(1); // default new users to having inline help visible
         }
 
+        // if doing only a partial update that includes only updating user's user group
+        if ($this->userGroupUpdateOnly) {
+            parent::execute(...$functionParams);
```
asmecher, line 306: I think because there are now security implications to how `execute` functions, I'd propose rearranging the implementation in this and related classes (`UserForm` and `UserRoleForm`, which also extends `UserForm`). Otherwise it might be possible that some change to `UserForm` accidentally permits users of `UserDetailsForm` to modify more or less of the user object than they should be able to.

I'd recommend renaming `UserForm::execute` to something like `saveUserGroupAssignments`, and then having both `UserDetailsForm` and `UserRoleFarm` call both `$this->saveUserGroupAssignments` **and** `parent::execute` (which will have become the `Form::execute` implementation, just for the sake of hooks).

(I don't like the way the `Form` class functions that take care of hook calls have had implications that ripple through subclass implementations -- but that can be cleaned up another time!)

templates/common/userDetailsReadOnly.tpl
```diff
@@ -0,0 +1,41 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * User details view only
+ *}
+<h4>{translate key="grid.user.userDetails"}</h4>
+
+<div id="userDetailsReadOnly" style="background-color: #EBEDEE; padding: 6px; margin-top: 5px; margin-bottom: 5px;">
+    <div class="section">				
+        <div class="inline pkp_helpers_third">
+            <span style="font-wight: bold; color: #777777;">{translate key="user.givenName"}</span><br />
```
asmecher, line 15: Style information should go into the `less` files rather than hard-coding it in the template.

asmecher (changes requested):
Hi Touhidur! Some changes required -- see comments.

classes/security/Validation.php
touhidurabir: @asmecher I tried to incorporate both of the option but not sure if I have made it more confusing in that process. please review .

controllers/grid/settings/user/form/UserDetailsForm.php
touhidurabir: done.

templates/common/userDetailsReadOnly.tpl
```diff
@@ -0,0 +1,41 @@
...
+            <span>{$givenName|array_pop}</span>
```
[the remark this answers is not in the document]
touhidurabir: done.

templates/common/userDetailsReadOnly.tpl
touhidurabir: done.

classes/security/Validation.php
```diff
@@ -468,11 +479,28 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+            return static::getApplicableAdministrationLevel(self::ADMINISTRATION_PROHIBITED, $allowedLevels);
         }
 
         // There were no conflicting roles. Permit administration.
-        return true;
+        return static::getApplicableAdministrationLevel(self::ADMINISTRATION_FULL, $allowedLevels);
+    }
+
+    /**
+     * Determine the adminstation level
```
asmecher, line 490: *administration

classes/security/Validation.php
```diff
@@ -425,28 +429,30 @@ public static function isSiteAdmin()
...
-     *
-     * @return bool True IFF the administration operation is permitted
+     * @param int   $administeredUserId     User ID of user to potentially administer
+     * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * @param array $allowedLevels          The allowed adminstration levels
+     * 
+     * @return int The authorized adminstrative level 
      */
-    public static function canAdminister($administeredUserId, $administratorUserId)
+    public static function canAdminister($administeredUserId, $administratorUserId, $allowedLevels = [self::ADMINISTRATION_FULL])
```
asmecher, line 438: Hmm, this feels a little bit arbitrary to me. As a coder I would expect `$allowedLevels` to essentially be an optional `array_intersect`, but here it's coded to default to just `ADMINISTRATION_FULL`. I think this is trying to do both parts of https://github.com/pkp/pkp-lib/pull/8283#discussion_r979016275 when it should be either one or the other; the combination is complicating things without preserving backwards compatibility.

asmecher (changes requested):

controllers/grid/settings/user/form/UserDetailsForm.php
```diff
@@ -58,19 +62,32 @@ public function __construct($request, $userId = null, $author = null)
...
+        $this->addSupportedFormLocale($request->getSite()->getPrimaryLocale());
+
+        if ($userId !== null) {
+            $this->user = Repo::user()->get($userId, true);
+        }
+    }
 
-        // Validation checks for this form
+    /**
+     * Attach the validatio checks for this form
```
asmecher, line 73: *validation

classes/security/Validation.php
touhidurabir: > this feels a little bit arbitrary to me
> think this is trying to do both suggestions in https://github.com/pkp/pkp-lib/pull/8283#discussion_r979016275 at once, when it should be either one or the other;

@asmecher  I agree 100% with that . So I propose that we introduce a new method `getAdministrationLevel` which will return appropriate level and deprecate the old `canAdminister` method but keep it for now(but marked it as `deprecated`) . that way we can ensure that no breaking change has introduced. Also the previous `canAdminister` was ok as that that time we have only `true/false` but now we have different levels . I have updated it accordingly . please review .

asmecher: @touhidurabir Did that proposal get pushed up? I don't see `getAdministrationLevel` implemented here.

touhidurabir: @asmecher I can confirm that I pushed those changes yesterday but weirdly they were not present when I checked today . More weird is that even that commit is missing in the commit history . Any idea what may cause this ? BTW, I have again pushed those changes they are there now .

classes/security/Validation.php
```diff
@@ -474,6 +482,65 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+     *
+     * @param int   $administeredUserId     User ID of user to potentially administer
+     * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * 
+     * @return int The authorized adminstration level 
+     */
+    public static function getAdministrationLevel($administeredUserId, $administratorUserId)
+    {
+        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var RoleDAO $roleDao */
+        $context = Application::get()?->getRequest()?->getContext();
```
asmecher, line 497: This binds the `Validation` class to the `Request`, which presumes web-based invocation. I'd rather not bind these together if we can avoid it, in case e.g. we need to call this from a command line tool or a queued job. I'd suggest passing the `$contextId` in as an optional 3rd variable.

classes/security/Validation.php
```diff
@@ -474,6 +482,65 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+
+    /**
+     * Get the user's administration level
+     *
+     * @param int   $administeredUserId     User ID of user to potentially administer
+     * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * 
+     * @return int The authorized adminstration level 
+     */
+    public static function getAdministrationLevel($administeredUserId, $administratorUserId)
```
asmecher, line 494: Use type hints for new code.

classes/security/Validation.php
```diff
@@ -474,6 +482,65 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+                    return self::ADMINISTRATION_PARTIAL;
+                }
+                return self::ADMINISTRATION_PROHIBITED;
+            }
+        }
+
+        // Make sure the administering user has a manager role somewhere
+        $foundManagerRole = false;
+        $roles = $roleDao->getByUserId($administratorUserId);
+        foreach ($roles as $role) {
```
asmecher, line 532: This could be rewritten using the `UserGroup` collector using e.g. (off the top of my head):
```php
Repo::userGroup()->collector()
    ->filterByUserIds([$administratorUserId])
    ->filterByRoleIds([ROLE_ID_MANAGER])->count()
```

classes/security/Validation.php
```diff
@@ -474,6 +482,65 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+     * Get the user's administration level
+     *
+     * @param int   $administeredUserId     User ID of user to potentially administer
+     * @param int   $administratorUserId    User ID of user who wants to do the administrating
+     * 
+     * @return int The authorized adminstration level 
+     */
+    public static function getAdministrationLevel($administeredUserId, $administratorUserId)
+    {
+        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var RoleDAO $roleDao */
```
asmecher, line 496: `RoleDAO` has been largely replaced by the UserGroup repository class and collector; I'd suggest rewriting this to use those instead. See notes below.

classes/security/Validation.php
```diff
@@ -474,6 +482,65 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+
+        // Otherwise, administrators can administer everyone
+        if ($roleDao->userHasRole(\PKP\core\PKPApplication::CONTEXT_SITE, $administratorUserId, Role::ROLE_ID_SITE_ADMIN)) {
+            return self::ADMINISTRATION_FULL;
+        }
+
+        // Check for administered user group assignments in other contexts
+        // that the administrator user doesn't have a manager role in.
+        $userGroups = Repo::userGroup()->userUserGroups($administeredUserId);
+        foreach ($userGroups as $userGroup) {
```
asmecher, line 517: This will be expensive when running through a list of users checking administration capabilities. I think you could probably rewrite this to make better use of the UserGroup collector to do the work in SQL rather than PHP.

controllers/grid/settings/user/form/UserForm.php
```diff
@@ -106,16 +106,28 @@ public function display($request = null, $template = null)
...
         }
 
-        parent::execute(...$functionArgs);
+        Repo::userGroup()
+            ->deleteAssignmentsByContextId(
+                Application::get()->getRequest()->getContext()->getId(), 
+                $this->userId
+            );
+
+        if ($this->getData('userGroupIds')) {
```
asmecher, line 127: This trusts the user-specified `userGroupIds` list; these could be in e.g. another context. (It looks like that was already the case but should be fixed here.)

asmecher (changes requested):
Thanks, @touhidurabir! I added some more comments. Some of it is issues that already existed but it's a good time to fix them.

classes/security/Validation.php
touhidurabir: done .

classes/security/Validation.php
touhidurabir: done .

classes/security/Validation.php
touhidurabir: done .

controllers/grid/settings/user/form/UserForm.php
touhidurabir: @asmecher bit confused about this requirement . Are you suggesting that we should check it against current context and this group if associated with this context ?

touhidurabir: I have implemented a context wise check to see if the user group id has any association with the given context, please check .

classes/security/Validation.php
touhidurabir: updated .

classes/security/Validation.php
touhidurabir: Removed and fully moved to `UserGroup Repository` .

controllers/grid/settings/user/form/UserForm.php
asmecher: That looks good, thanks!

classes/security/Validation.php
```diff
@@ -474,6 +482,83 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+                    ->filterByRoleIds([Role::ROLE_ID_MANAGER])
+                    ->getCount()
+            )
+            ->count();
+        
+        if ( $userGroupsCount > 0 ) {
+            // Found an assignment: disqualified.
+            // But also determine if a partial administrate is allowed
+            // if the Administrator User is a Journal Manager in the current context
+            if ($contextId &&
```
asmecher, line 536: I would suggest checking `$contextId !== null`, to distinguish from e.g. `CONTEXT_SITE` (which would be `0`).

classes/security/Validation.php
```diff
@@ -474,6 +482,83 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+        if ($administeredUserId == $administratorUserId) {
+            return self::ADMINISTRATION_FULL;
+        }
+
+        $filteredSiteAdminUserGroups = Repo::userGroup()
+            ->getCollector()
+            ->filterByContextIds([\PKP\core\PKPApplication::CONTEXT_SITE])
+            ->filterByRoleIds([Role::ROLE_ID_SITE_ADMIN]);
+
+        // You cannot adminster administrators
```
asmecher, line 507: *administer

classes/security/Validation.php
```diff
@@ -474,6 +482,83 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+        }
+
+        // Otherwise, administrators can administer everyone
+        if ($filteredSiteAdminUserGroups->filterByUserIds([$administratorUserId])->getCount() > 0) {
+            return self::ADMINISTRATION_FULL;
+        }
+
+        // Check for administered user group assignments in other contexts
+        // that the administrator user doesn't have a manager role in.
+        $userGroupsCount = Repo::userGroup()
```
asmecher, line 519: I suspect this won't perform well on large databases as it executes queries in a loop.

What we need is the number of user groups that the `$administeredUserId` is in, but that `$administratorUserId` doesn't manage. I would suggest getting a list of IDs for all user groups that `$administratorUserId` manages, and a list of all user groups that `$administeredUserId` participates in. If the second list intersected the first list contains any items, then full administration should be prevented. That would just require two queries. (We could rewrite it down to a single query, but that would require adding some single-use functions to the `Collector` class that I don't think would be justified.)

classes/security/Validation.php
```diff
@@ -474,6 +482,83 @@ public static function canAdminister($administeredUserId, $administratorUserId)
...
+                    ->filterByUserIds([$administratorUserId])
+                    ->filterByRoleIds([Role::ROLE_ID_MANAGER])
+                    ->getCount()) {
+                return self::ADMINISTRATION_PARTIAL;
+            }
+            return self::ADMINISTRATION_PROHIBITED;
+        }
+
+        // Make sure the administering user has a manager role somewhere
+        $roleManagerCount = Repo::userGroup()
```
asmecher, line 549: This check should be moved above any potential `ADMINISTRATION_FULL` or `ADMINISTRATION_PARTIAL` return. (You could re-use the query in https://github.com/pkp/pkp-lib/pull/8283/files#r989522297 for this for bonus points!)

templates/common/userDetailsReadOnly.tpl
```diff
@@ -0,0 +1,41 @@
+{**
+ * common/userDetailsReadOnly.tpl
```
asmecher, line 2: Should be `templates/common/userDetailsReadOnly.tpl` (include full path)

templates/controllers/grid/settings/user/form/userDetailsForm.tpl
```diff
@@ -29,27 +29,38 @@
 	<input type="hidden" id="sitePrimaryLocale" name="sitePrimaryLocale" value="{$sitePrimaryLocale|escape}" />
 	<div id="userDetailsFormContainer">
 		<div id="userDetails" class="full left">
-			{if $userId}
-				<h3>{translate key="grid.user.userDetails"}</h3>
-			{else}
-				<h3>{translate key="grid.user.step1"}</h3>
+			{if !$userGroupUpdateOnly}
```
asmecher, line 32: Are you sure this change is correct? At a glance, I think the `{if}` check here was to determine whether we're creating a new user (`$userId` not present) vs. whether we're editing an existing one (`$userId` present).

asmecher: Oops, NVM, I see the nested `if`.

templates/common/userDetailsReadOnly.tpl
touhidurabir: fixed.

classes/security/Validation.php
touhidurabir: fixed.

classes/security/Validation.php
touhidurabir: done.

classes/security/Validation.php
touhidurabir: @asmecher I think so but I think it should not be right at the top but right after following three checks 
1. user can administrate himself/herself.
2. can not administrate an administrator.
3. an administrator can administrate anyone .

updated accordingly .

classes/userGroup/Repository.php
```diff
@@ -247,13 +247,11 @@ public function userInGroup(int $userId, int $userGroupId): bool
...
     public function contextHasGroup(int $contextId, int $userGroupId): bool
     {
-        $collector = Repo::userGroup()
-            ->getCollector()
-            ->filterByContextIds([$contextId]);
-
-        $userGroups = $collector->getMany();
-
-        return $userGroups->where('id', $userGroupId)->count();
+		return Repo::userGroup()
```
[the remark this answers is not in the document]
touhidurabir: fixed.

classes/security/Validation.php
touhidurabir: @asmecher you are right and I should have considered to refactor it . But the implementation that I did is bit different than what you proposed . If I understand correctly , all we need to determine any context that has the `administered user group` assigned but the `administrator user` don't have nay `manager` role in that context. So basically we can do a diff between the both of the user case context list . And thats sounds logical and easier. But correct me if I understand it wrong .

asmecher: You're right -- it should be a diff, not an intersect!

PR #8289, opened 2022-09-26
pkp/pkp-lib#7352 added language locale to site languages list

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#7352


classes/i18n/Locale.php
```diff
@@ -353,6 +353,21 @@ public function getScripts(?string $locale = null): Scripts
         return $this->_getLocaleCache(__METHOD__, $locale, fn () => $this->_getIsoCodes($locale)->getScripts());
     }
 
+    /**
+     * Get the locales counts by only locale code
+     */
+    public function getLocaleCodesCount(array $allLocales = []): array
```
asmecher, line 359: This is a general coding style preference rather than something specific to this function, but I prefer to use an explicit `null` for special cases rather than an empty array for an array parameter. That way if the `$allLocales` parameter comes from an upstream array operation of some kind that legitimately returns a `[]`, the function will behave as expected.

classes/i18n/Locale.php
```diff
@@ -353,6 +353,21 @@ public function getScripts(?string $locale = null): Scripts
...
+    {
+        if (empty($allLocales)) {
+            $allLocales = $this->getLocales();
+        }
+        return array_count_values(
+            collect(array_keys($allLocales))
+                ->map(fn(string $value) => trim(explode('_', $value)[0]))
+                ->toarray()
```
asmecher, line 367: `toArray` (capitalized)

classes/i18n/Locale.php
```diff
@@ -353,6 +353,21 @@ public function getScripts(?string $locale = null): Scripts
...
+     * Get the locales counts by only locale code
+     */
+    public function getLocaleCodesCount(array $allLocales = []): array
+    {
+        if (empty($allLocales)) {
+            $allLocales = $this->getLocales();
+        }
+        return array_count_values(
+            collect(array_keys($allLocales))
+                ->map(fn(string $value) => trim(explode('_', $value)[0]))
```
asmecher, line 366: Have you seen cases where the `trim` is necessary? I'm pretty sure other parts of the code would break if there were spaces in the locale code. (Elsewhere also)

classes/site/Site.php
```diff
@@ -38,8 +38,20 @@ public function getSupportedLocaleNames()
...
         $supportedLocales = array_fill_keys($this->getSupportedLocales(), '');
-        array_walk($supportedLocales, fn (&$item, string $locale) => $item = Locale::getMetadata($locale)->getDisplayName());
+        $localeCodesCount = Locale::getLocaleCodesCount();
+
+        array_walk($supportedLocales, function(&$item, string $locale) use ($localeCodesCount) {
+            $localeCode = trim(explode('_', $locale)[0]);
+            $item = Locale::getMetadata($locale)
+                ->getDisplayName(
+                    null, 
+                    isset($localeCodesCount[$localeCode]) && $localeCodesCount[$localeCode] > 1 ? true : false,
```
asmecher, line 50: `($localeCodesCount[$localeCode] ?? 0) > 1` (also below)

controllers/grid/admin/languages/AdminLanguageGridHandler.php
```diff
@@ -416,7 +417,7 @@ protected function _updateContextLocaleSettings($request)
     {
         $site = $request->getSite();
         $siteSupportedLocales = $site->getSupportedLocales();
-        $contextService = \Services::get('context');
```
asmecher, line 419: Thanks for catching this, there will be lots similar!

controllers/grid/admin/languages/AdminLanguageGridHandler.php
```diff
@@ -453,7 +454,7 @@ protected function _canManage($request)
     {
         $contextDao = Application::getContextDAO();
         $contexts = $contextDao->getAll();
-        $userRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+        $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
```
asmecher, line 457: Likewise, thanks for catching this.

classes/i18n/Locale.php
asmecher: This is kind of a confusing function -- it's not clear by looking at its description what it does or how it might be used. Both of the use cases for it carry some duplicated code, so I suggest making this function do a little more -- e.g. get a list of language codes by display name. That'll both clear up this function and reduce the duplicated code in both calling cases.

asmecher (changes requested):
Getting close, thanks!

classes/i18n/Locale.php
touhidurabir: Not really . it's more like a habit of mine to do a bit of defensive programming. probably not necessary .

classes/i18n/Locale.php
touhidurabir: done .

classes/site/Site.php
touhidurabir: done .

classes/i18n/Locale.php
touhidurabir: Done .

classes/i18n/Locale.php
asmecher: I understand, but I'd rather not do this unless it's necessary. For a long time we allowed vague data to be stored in the database -- `0` and `null` and `''` and `[]` were interchangeable, for example. It led to overcomplicated/slow queries and lots of areas where data was acceptable in one part of the system but caused errors in another. But the most challenging thing was managing upgrades where it was necessary to anticipate boundary data conditions in the wild and remember them across several releases. In the last 3 or so years we've been changing our approach towards have clear, unambiguous expectations of data -- `null` is distinct from `[]` and each has a specific meaning; foreign key integrity is enforced in the database; types/nullability in PHP are hinted; etc. The idea is for data errors to surface as soon as possible and for the underlying cause to be fixed, rather than adding work-arounds ad-hoc and incompletely. That way we can catch/fix them quickly during the dev cycle rather than having end users find boundary conditions through variety of data.

touhidurabir: @asmecher I understand . I have removed the `trim` from `Locale@getFormattedDisplayNames` .

asmecher (changes requested):

controllers/grid/languages/LanguageGridHandler.php
```diff
@@ -241,6 +241,25 @@ public function addNameColumn()
...
+    {
+        $cellProvider = $this->getCellProvider();
+
+        // Locale code.
+        $this->addColumn(
+            new GridColumn(
+                'code',
+                'grid.columns.locale.code',
+                null,
+                'controllers/grid/languages/localeCodeCell.tpl',
```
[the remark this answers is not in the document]
touhidurabir: Right , I can reuse the `controllers/grid/languages/localeNameCell.tpl` . Updated .

classes/i18n/Locale.php
```diff
@@ -353,6 +357,69 @@ public function getScripts(?string $locale = null): Scripts
...
+    /**
+     * Get the formatted locale display names with country if same language code present multiple times
+     * 
+     * @param   array $filterByLocales  Optional list of locales code to filter by the returned formatted names list
+     * @param   array $locales          Optional list of availabel all locales
+     * @param   int   $langLocaleStatus The const value of one of LocaleMetadata:LANGUAGE_LOCALE_*
+     * 
+     * @return  array                   The list of locales with formatted display name
+     */
+    public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH): array
```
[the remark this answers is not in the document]
touhidurabir: Updated.

asmecher:
Manually cherry-picked (merge conflict)

Issue #8290, opened 2022-09-26
Inclusion of Data Availability Statement as submission metadata


iudizm:
**Describe the problem you would like to solve**
In issue #6278, @jalperin proposed the inclusion of a "Data Availability Statement" field to enable metadata for datasets to be published alongside articles.

As discussions in that issue expanded to Data Citations and indexing services, we created this child-issue to focus on the inclusion of the Data Availability Statement field.

**Describe the solution you'd like**
Make possible to enable the Data Availability Statement as submission metadata, so it can be requested/required to authors in the submission and revision processes.

It should be visible in the publication public page and also for reviewers.

It would be great to have it in OJS, OPS and OMP.

**Who is asking for this feature?**
As stated by Juan, "it has been a standard feature in virtually all journal submissions systems (in PLOS since 2014)". As the number of datasets published continues to grow, this also becomes a increasing demand from the community.

**Additional information**
There is more context to this in the parent issue: #6278

iudizm:
Are there any acceptance tests that should be implemented for this new submission metadata? if so, how could it be done?

asmecher:
@iudizm, for tests, there are two options:
1. Alter one of the existing submission tests (something in `cypress/tests/data/60-content/`) so that it provides a data availability statement and e.g. checks the reviewer process to ensure that it's present, or
2. Create a new test in e.g. `lib/pkp/cypress/tests/integration/` (if it can be kept application-agnostic) or `cypress/tests/integration/` (if not) that enters a data availability statement in an existing submission in the dataset, and checks an existing review assignment to ensure that it's visible.

I think option 2 will probably be better. You can get a fully built dataset to work with from https://github.com/pkp/datasets -- this is what the Travis CI tests generate for their own testing purposes, before running the integration tests.

asmecher:
@iudizm, if I read it correctly, the attached PRs...
- Add a new field to setup allowing the journal manager to enable / disable / require the Data Availability Statement
- Add a new field for the Data Availability Statement to the submission process, when enabled
- Expose the contents of that field to reviewers when they access the submission metadata in Peer Review

@jalperin, does that match your expectations?

jalperin:
* What about displaying on the abstract page? 
* does exposing to reviewers respect anonymity?

iudizm:
About the presentation of the Data Availability Statement in the landing page of the article/preprint/monograph:

[...]

PR #8291, opened 2022-09-26
pkp/pkp-lib#8290 Inclusion of 'Data Availability Statements' as submission metadata

iudizm:
Related issue: #8290


schemas/context.json
```diff
@@ -841,6 +841,14 @@
 			"type": "boolean",
 			"description": "Whether or not the access to the COUNTER SUSHI statistics API is public. If false, only users with admin or manager roles can access the API.",
 			"default": true
+		},
+		"dataAvailability": {
+			"type": "string",
+            "description": "Whether to not request, request or require a Data Availability Statement from the authors.",
```
asmecher, line 847: Sorry, I know this is trivial, but can you format the JSON changes with tab-based indentation? The changes in this PR use a mix of tabs and spaces.

templates/controllers/modals/submission/viewSubmissionMetadata.tpl
```diff
@@ -22,7 +22,7 @@
 					{if $metadataItem@iteration % 2 != 0}
 						<th scope="row">{$metadataItem}</th>
 					{else}
-						<td>{$metadataItem}</td>
+						<td>{$metadataItem|escape}</td>
```
asmecher, line 25: It looks like you've identified missing escaping here -- but if so, can you add it to the above `<th ...>` element as well?

schemas/context.json
iudizm: Sure, sorry for mixing indentation styles

templates/controllers/modals/submission/viewSubmissionMetadata.tpl
iudizm: done, escaping the `<th ...>` as well now

iudizm:
@asmecher, done implementing the Cypress test for the metadata, could you take a look?

note: 
You said the test should "check an existing review assignment to ensure that it's visible", but there wasn't any review assignments with the 'disclosed author' method in the [dataset](https://github.com/pkp/datasets/tree/main/ojs/main/mysql), so I had to [create a new assignment](https://github.com/pkp/pkp-lib/pull/8291/files#diff-bb23160115233ba107b181077d37776cd48b6b08bca81dc34a86f0a99c54cacdR29) to check out the presence of the metadata in this type of review.
I hope this is the right way to do it.

cypress/tests/integration/DataAvailabilityStatements.spec.js
```diff
@@ -0,0 +1,54 @@
...
+		cy.wait(500);
+        cy.contains('Enable Data Availability Statement metadata').find('[type="checkbox"]').check()
+        cy.wait(500);
+		cy.get('#metadata button').contains('Save').click();
+		cy.get('#metadata [role="status"]').contains('Saved');
+	});
+    
+    it('Adds a statement to a submission', function () {
+        cy.login('dbarnes');
+		cy.visit('index.php/publicknowledge/workflow/index/14/1');
```
asmecher, line 20: It's best not to embed database IDs in the test. The best approach is to have the test find the content e.g. via the editorial interface by title or author name, then store its ID in an [alias](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Aliases) for the testing code to use.

Admittedly, sometimes we do embed database IDs in our own tests; if you do continue that way, it's best to at least pull this out into a constant near the top of the file.

cypress/tests/integration/DataAvailabilityStatements.spec.js
```diff
@@ -0,0 +1,54 @@
...
+        cy.get('#metadata-dataAvailability-control-en_US').clear();
+        cy.get('#metadata-dataAvailability-control-en_US').type(statement);
+        cy.get('#metadata button').contains('Save').click();
+        cy.get('#metadata [role="status"]').contains('Saved');
+    });
+
+    it('Creates a review with disclosed author', function () {
+        cy.login('dbarnes');
+		cy.visit('index.php/publicknowledge/workflow/index/14/1');
+		cy.wait(1000);
```
asmecher, line 32: `cy.wait` statements [should be considered FIXMEs](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting). We do have a few cases where we use waits internally -- also needing to be fixed e.g. due to some sort of Javascript rewrite -- but they shouldn't be used indiscriminately or it'll lead to balancing random beakage of the tests against the whole suite taking too long to run.

(For example, there is a 5.5 second delay just below -- that's a lot of time for the server to sit around!)

schemas/publication.json
```diff
@@ -329,6 +329,14 @@
 			"validation": [
 				"min:1"
 			]
-		}
+		},
+        "dataAvailability": {
```
asmecher, line 333: The indentation contains a mix of spaces and tabs, here and elsewhere in the PR; please follow the existing conventions for each file. (We're gradually moving to autoformatting with git hooks, but that hasn't happened everywhere yet.)

asmecher (changes requested):
Sorry for the delay in getting around to a second review! I have added a couple of comments on the tests, plus one in the OJS PR that may affect the other application PRs as well.

cypress/tests/integration/DataAvailabilityStatements.spec.js
iudizm: Sorry for this smelly code, done removing the database IDs from the specs

cypress/tests/integration/DataAvailabilityStatements.spec.js
iudizm: Done, sorry for this bad practice

iudizm:
@asmecher,

We moved the test about the metadata in the Review to the OJS specific integration specs. 
So it is now [here](https://github.com/pkp/ojs/pull/3553/files).

NateWr:
Replaced by https://github.com/pkp/pkp-lib/pull/8570

PR #8324, opened 2022-10-11
pkp/pkp-lib#8183 Context contact name and email required at creation with preflight check

touhidurabir:
This PR aims to solve pkp/pkp-lib#8183

### DIFF
--- a/api/v1/contexts/PKPContextHandler.php
+++ b/api/v1/contexts/PKPContextHandler.php
@@ -16,7 +16,9 @@ namespace PKP\API\v1\contexts;
 
 use APP\core\Application;
 use APP\core\Services;
+use APP\services\ContextService;
 use APP\template\TemplateManager;
+use PKP\db\DAORegistry;
 use PKP\handler\APIHandler;
 use PKP\plugins\Hook;
 use PKP\plugins\PluginRegistry;
@@ -24,8 +26,10 @@ use PKP\security\authorization\PolicySet;
 use PKP\security\authorization\RoleBasedHandlerOperationPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
+use PKP\security\RoleDAO;
 use PKP\services\interfaces\EntityWriteInterface;
 use PKP\services\PKPSchemaService;
+
 
 class PKPContextHandler extends APIHandler
 {
@@ -152,7 +156,7 @@ class PKPContextHandler extends APIHandler
         // Anyone not a site admin should not be able to access contexts that are
         // not enabled
         if (empty($allowedParams['isEnabled'])) {
-            $userRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+            $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
             $canAccessDisabledContexts = !empty(array_intersect([Role::ROLE_ID_SITE_ADMIN], $userRoles));
             if (!$canAccessDisabledContexts) {
                 return $response->withStatus(403)->withJsonError('api.contexts.403.requestedDisabledContexts');
@@ -206,9 +210,9 @@ class PKPContextHandler extends APIHandler
         // A disabled journal can only be access by site admins and users with a
         // manager role in that journal
         if (!$context->getEnabled()) {
-            $userRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+            $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
             if (!in_array(Role::ROLE_ID_SITE_ADMIN, $userRoles)) {
-                $roleDao = DaoRegistry::getDao('RoleDAO');
+                $roleDao = DAORegistry::getDao('RoleDAO'); /** @var RoleDAO $roleDao */
                 if (!$roleDao->userHasRole($context->getId(), $user->getId(), Role::ROLE_ID_MANAGER)) {
                     return $response->withStatus(403)->withJsonError('api.contexts.403.notAllowed');
                 }
@@ -252,9 +256,9 @@ class PKPContextHandler extends APIHandler
         // A disabled journal can only be access by site admins and users with a
         // manager role in that journal
         if (!$context->getEnabled()) {
-            $userRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+            $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
             if (!in_array(Role::ROLE_ID_SITE_ADMIN, $userRoles)) {
-                $roleDao = DaoRegistry::getDao('RoleDAO');
+                $roleDao = DAORegistry::getDao('RoleDAO'); /** @var RoleDAO $roleDao */
                 if (!$roleDao->userHasRole($context->getId(), $user->getId(), Role::ROLE_ID_MANAGER)) {
                     return $response->withStatus(403)->withJsonError('api.contexts.403.notAllowed');
                 }
@@ -318,7 +322,7 @@ class PKPContextHandler extends APIHandler
             }
         }
 
-        $contextService = Services::get('context');
+        $contextService = Services::get('context'); /** @var ContextService $contextService */
         $errors = $contextService->validate(EntityWriteInterface::VALIDATE_ACTION_ADD, $params, $allowedLocales, $primaryLocale);
 
         if (!empty($errors)) {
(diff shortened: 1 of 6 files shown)


classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,74 @@ protected function setFallbackVersion(): ?string
...
+                $join
+                    ->on("context.{$this->getContextKeyField()}", "=", "context_setting.{$this->getContextKeyField()}")
+                    ->where("context_setting.locale", "=", $primaryLocale)
+                    ->where("context_setting.setting_name", "=", "name")
+            )
+            ->where("context.enabled", 1)
+            ->get();
+        
+        // Get all the context IDs for which contact name/email is set
+        $contextIdsWithValidContact = DB::table("{$this->getContextTable()} AS context")
```
Vitaliy-1, line 531: For brevity, I'd use a subquery instead of joins here and just check if those settings exist, something like:
```sql
SELECT `journal_id` FROM `journals` WHERE `journal_id` IN (SELECT `journal_id` FROM `journal_settings` WHERE `setting_name` IN ('contactEmail', 'contactName'))
...
```

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,74 @@ protected function setFallbackVersion(): ?string
...
+                $join
+                    ->on("contact_name_setting.{$this->getContextKeyField()}", "=", "context.{$this->getContextKeyField()}")
+                    ->where("contact_name_setting.setting_name", "=", "contactName")
+            )
+            ->leftJoin("{$this->getContextSettingsTable()} AS contact_email_setting", fn($join) => 
+                $join
+                    ->on("contact_email_setting.{$this->getContextKeyField()}", "=", "context.{$this->getContextKeyField()}")
+                    ->where("contact_email_setting.setting_name", "=", "contactEmail")
+            )
+            ->whereNotNull("contact_email_setting.{$this->getContextKeyField()}")
```
Vitaliy-1, line 543: Can journal ID be null here? Just clarifying for myself this check

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,74 @@ protected function setFallbackVersion(): ?string
...
+            ->whereIn("context.{$this->getContextKeyField()}", $contexts->pluck("{$this->getContextKeyField()}")->toArray())
```
Vitaliy-1, line 544: Is this to ensure that only enabled journals are retrieved?
What about using direct SQL transaction to find `journal_id`'s where contactName and contactEmail don't exist? Considering the previous example:
```sql
SELECT `journal_id` FROM `journals` WHERE `journal_id` NOT IN (SELECT `journal_id` FROM `journal_settings` WHERE `setting_name` IN ('contactEmail', 'contactName')) AND `enabled` = 1
....
```
For the message, you then can retrieve the journal name by its ID

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,74 @@ protected function setFallbackVersion(): ?string
...
+     */
+    protected function checkContactSetting(): void
+    {
+        $primaryLocale = DB::table('site')
+            ->select(['primary_locale'])
+            ->first()
+            ->primary_locale;
+
+        // Get all the enabled context with context name/title
+        $contexts = DB::table("{$this->getContextTable()} AS context")
```
Vitaliy-1, line 513: The curly syntax here for expressions inside strings is looking different in this method's body compared to the rest of the file. Shouldn't it be standardized? But this is for your and Alec's consideration.

classes/components/forms/context/PKPContextForm.php
```diff
@@ -71,6 +72,19 @@ public function __construct($action, $locales, $baseUrl, $context)
                 'groupId' => 'identity',
                 'value' => $context ? $context->getData('acronym') : null,
             ]))
+            ->addField(new FieldText('contactName', [
+                'label' => __('manager.setup.principalContact') .' '. __('common.name'),
+                'isRequired' => true,
+                // 'groupId' => 'principal',
```
Vitaliy-1, line 78: Probably, these commented lines should be removed before the merge, right?

Vitaliy-1 (commented):
Looks good! I left a couple of comments and tested these changes, this is working as expected

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
touhidurabir: No and it's no necessary. updated .

api/v1/contexts/PKPContextHandler.php
```diff
@@ -206,9 +210,9 @@ public function get($slimRequest, $response, $args)
         // A disabled journal can only be access by site admins and users with a
         // manager role in that journal
         if (!$context->getEnabled()) {
-            $userRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+            $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
             if (!in_array(Role::ROLE_ID_SITE_ADMIN, $userRoles)) {
-                $roleDao = DaoRegistry::getDao('RoleDAO');
+                $roleDao = DAORegistry::getDao('RoleDAO'); /** @var ROleDAO $roleDao */
```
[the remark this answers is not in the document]
touhidurabir: fixed.

api/v1/contexts/PKPContextHandler.php
```diff
@@ -252,9 +256,9 @@ public function getTheme($slimRequest, $response, $args)
...
```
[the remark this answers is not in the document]
touhidurabir: fixed.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
touhidurabir: Right, SubQuery is a better choice and through that I can do it in one single query rather than multiple query and diffing . Updated .

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
touhidurabir: Yeah, we only need to target the enable ones , right ? Or should we test all existing journals? Extra overhead can be associated in both case as if system have one or multiple disable contexts, then

1. If no plan to activate those in future, make no sense for them to be forced to have those information added during the upgrade 
2. If system admin has the plan to active those in future and not have these info added during upgrade but enable the contexts later after upgrade and then not providing those contact name/email, email and submission will not work properly and they may have a hard to track the issue . 

@Vitaliy-1   What is your suggestion in such case ?

classes/components/forms/context/PKPContextForm.php
touhidurabir: removed .

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
touhidurabir: The existing curly syntax `${VARIABEL}`  has been deprecated(with 8.2+) and should be updated to `{$VARIABLE}` more details here https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated . it will not show the deprecation message right now but better to update those.

Thanks for spotting those and I have fixed those .

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
Vitaliy-1: Thinking about this further, even if the journal is disabled from appearing publicly, it still can be used to send emails and some of them still rely on a context contact email. At least the system allows that. I think it makes sense to include them in the check.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,51 @@ protected function setFallbackVersion(): ?string
...
+            ->select(['primary_locale'])
+            ->first()
+            ->primary_locale;
+
+        $missingContactContexts = DB::table("{$this->getContextTable()}")
+            ->select([
+                "{$this->getContextKeyField()}", 
+                "enabled", 
+            ])
+            ->addSelect(["title" => DB::table("{$this->getContextSettingsTable()}")
```
Vitaliy-1, line 517: Nice, I didn't know that `addSelect` allows to include select subqueries

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -488,4 +495,51 @@ protected function setFallbackVersion(): ?string
...
+                ->select(["setting_value"])
+                ->whereColumn("{$this->getContextKeyField()}", "{$this->getContextTable()}.{$this->getContextKeyField()}")
+                ->where("locale", $primaryLocale)
+                ->where("setting_name", "name")
+                ->take(1)
```
Vitaliy-1, line 522: Probably, we don't need to limit the result here, I think it will always return 1 row with a context name as the combination of journal_id, locale and setting_name is unique

touhidurabir: right, removed .

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
touhidurabir: got it, removed the `enable` constrain.

PR #8349, opened 2022-10-19
pkp/pkp-lib#7191 Implement new submission wizard

NateWr:


classes/migration/upgrade/v3_4_0/I7191_InstallSubmissionHelpDefaults.php
```diff
@@ -0,0 +1,137 @@
...
+            'forTheEditorsHelp' => 'default.submission.step.forTheEditors',
+            'reviewHelp' => 'default.submission.step.review',
+            'uploadFilesHelp' => 'default.submission.step.uploadFiles',
+        ]);
+    }
+
+    /**
+     * Get the params needed to localize some of the default settings
+     */
+    protected function getLocalizationParams(stdClass $row, string $locale): array
```
asmecher, line 97: If this wasn't migration code, and if it didn't introduce just as many complications as it would solve, I'd suggest unpacking the `stdClass` into named parameters to allow the function to better self-document and enforce parameter requirements. But this is just migration code, and it would require the `->get` list to alias columns to match the function requirements, which is pretty wonky, so never mind.

classes/observers/listeners/SendSubmissionAcknowledgement.php
```diff
@@ -0,0 +1,126 @@
...
+            /** @var StageAssignment $stageAssignment */
+            $assignedUserIds[] = $stageAssignment->getUserId();
+        }
+
+        $submitterUsers = !empty($assignedUserIds)
+            ? Repo::user()
+                ->getCollector()
+                ->filterByUserIds($assignedUserIds)
+                ->getMany()
+            : collect();
```
asmecher, line 56: This is where writing the collectors to distinguish between `null` (filter not set) and `[]` (filter set, but with no acceptable values) will pay off. You should be able to just pass in `filterByUserIds($assignedUserIds)` even if it's an empty array and the filter will behave as expected. The ternary isn't necessary.

classes/observers/listeners/UpdateAuthorStageAssignments.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $userGroups = Repo::userGroup()
+            ->getCollector()
+            ->filterByContextIds([$event->context->getId()])
+            ->filterByRoleIds([Role::ROLE_ID_AUTHOR])
+            ->getMany();
+
+        /** @var StageAssignment $stageAssignment */
+        while ($stageAssignment = $stageAssigments->next()) {
+            $userGroup = $userGroups->first(fn (UserGroup $userGroup) => $userGroup->getId() === $stageAssignment->getUserGroupId());
```
asmecher, line 56: I think...
```php
$userGroup = $userGroups->get($stageAssignment->getUserGroupId()));
```
...would've worked here, since the UserGroup DAO is yielding by ID.

classes/publication/Repository.php
```diff
@@ -470,7 +485,12 @@ public function publish(Publication $publication)
                 $submission
             ]
         );
-        event(new PublishedEvent($newPublication, $publication, $submission));
+
+        $context = $submission->getData('contextId') === Application::get()->getRequest()->getContext()->getId()
```
asmecher, line 489: Since we're in the Repository class, far away from request handling, I think it's best not to assume we'll have a Request object (for a future where it's simply not available during non-interactive invocations). But this is an acceptable optimization, so I'd recommend using the null-safe operator:
```php
Application::get()->getRequest()?->getContext()?->getId()
```

classes/query/QueryDAO.php
```diff
@@ -105,6 +111,33 @@ public function getByAssoc($assocType, $assocId, $stageId = null, $userId = null
         );
     }
 
+    /**
+     * Retrieve a count of all queries totalled by stage
```
asmecher, line 115: *all *open* queries

classes/query/QueryDAO.php
```diff
@@ -394,23 +421,71 @@ public function addRecommendationQuery(int $recommenderUserId, int $submissionId
...
     }
+
+    /**
+     * Create a query with a submission's comments for the editors
+     *
+     * Creates the query and assigns all participants
+     *
+     * @return int new query id
+     */
+    public function addCommentsForEditorsQuery(Submission $submission): int
```
asmecher, line 451: Now that we're starting to have queries created by several workflow processes, maybe it's worth adding some kind of a "source" field to differentiate between them in case e.g. we need to later alter the way some of them are stored in the database. We've often had to resort to heuristics to distinguish between things after the fact. (I'm considering a similar addition to the submissions table, for example, to distinguish between imported, quickSubmit, and regular workflow submissions.)

classes/query/QueryDAO.php
```diff
@@ -394,23 +421,71 @@ public function addRecommendationQuery(int $recommenderUserId, int $submissionId
...
+            $submission->getId(),
+            [
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SUB_EDITOR,
+                Role::ROLE_ID_ASSISTANT,
+                Role::ROLE_ID_AUTHOR,
+            ],
+            $submission->getData('stageId')
+        );
+        $assigned = collect($assigned->toArray());
```
asmecher, line 466: Isn't this just round-tripping a collection to an array and back again?

asmecher (changes requested):
I'll add a summary on the issue, but here are some line-specific notes. I didn't get through all of it this week but will pick it up next.

pages/submission/PKPSubmissionHandler.php
```diff
@@ -10,257 +10,838 @@
...
+     */
+    protected function getPublicationApiUrl(Request $request, int $submissionId, int $publicationId): string
+    {
+        return $request
+            ->getDispatcher()
+            ->url(
+                $request,
+                Application::ROUTE_API,
+                $request->getContext()->getPath(),
+                'submissions/' . $submissionId . '/publications/' . $publicationId
```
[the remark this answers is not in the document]
NateWr: Can I pass an array in as the `$handler` to `$dispatcher->url()`? It gets passed down to `APIRouter` as the `$endpoint` and seems to be used as a string here:

https://github.com/pkp/pkp-lib/blob/main/classes/core/APIRouter.php#L227

Maybe there is another approach to constructing API urls?

pages/submission/PKPSubmissionHandler.php
```diff
@@ -10,257 +10,838 @@
...
+            $this->getSubmissionFilesApiUrl($request, $submission->getId()),
+            $genres
+        );
+
+        return [
+            'addFileLabel' => __('common.addFile'),
+            'apiUrl' => $this->getSubmissionFilesApiUrl($request, $submission->getId()),
+            'cancelUploadLabel' => __('form.dropzone.dictCancelUpload'),
+            'genrePromptLabel' => __('submission.submit.genre.label'),
+            'documentTypes' => [
```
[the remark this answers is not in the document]
NateWr: Probably. But at the moment this is the only place we use this array, so I thought I'd wait to see how long `FileManager` lives. When we introduced Flysystem, I think I thought `FileManager` would be deprecated soon, but that's not likely to happen for a while.

Happy to do whatever you think is best.

pages/submission/PKPSubmissionHandler.php
```diff
@@ -10,257 +10,838 @@
...
+            ),
+            'id' => 'submissionFiles',
+            'items' => Repo::submissionFile()
+                ->getSchemaMap()
+                ->summarizeMany($submissionFiles, $genres)
+                ->values(),
+            'options' => [
+                'maxFilesize' => Application::getIntMaxFileMBs(),
+                'timeout' => ini_get('max_execution_time') ? ini_get('max_execution_time') * 1000 : 0,
+                'dropzoneDictDefaultMessage' => __('form.dropzone.dictDefaultMessage'),
```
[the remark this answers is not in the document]
NateWr: Yes, this is on my list of "hard problems" for the UI Library at the moment. I have gone back and forth and so there are two approaches mixed in right now:

1. Common locale keys are added to state through the TemplateManager and can be used in Vue components with `this.__('common.close')`. See https://github.com/pkp/pkp-lib/blob/main/classes/template/PKPTemplateManager.php#L814.
2. Locale keys used by a single component are passed in as props, like `dropzoneDictDefaultMessage`. The benefit of this approach is that we can track every use of a locale key. Although the prop name is different, it must be tied to a server-side use of the locale key. That means that we can track all uses within the PHP code. With the global locale keys set in `TemplateManager`, we have to search the UI Library and template files to see if they are used. The other downside of using global locale keys is that UI components become more tightly coupled to a specific translation. Often we want a component to be reusable in different contexts, which may require different labels.

I would love to be able to use an approach like https://github.com/avast/vuei18n-po, but I've always thought that we have just too many locale strings. All .po files except `emails.po` in locale/en_US and lib/pkp/locale/en_US add up to about 530kb, and I don't think a minified JSON file would save more than 10-20%. That's a pretty big increase in our payload, which is already far too big (builds at 2.5mb now). We need to start reducing the built file size.

For loading locale files, none of the approaches I can think of sound good: a) segment our .po files so that what's needed in state can be bundled with the JS package, but that seems hard and prone to errors; b) request locale keys "on demand" from the Vue components, but that would require quick server response times which aren't common in our community; c) pass locale keys into components directly, like we do now, but in a way that preserves the locale key structure (eg - a global state object of some kind, which is kind of how `$templateMgr->setLocaleKeys(...)` works, but figure out a better way to do that).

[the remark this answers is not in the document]
NateWr: Another thing we can do, which would be difficult but I think would be possible, is to write an extension to our webpack build script which searched our component files for a syntax (like `this.__(...)`), extracted the locale keys, and generated built files of just the locale keys that are used. We'd also need to scan `.tpl` files. And it wouldn't cover plugins. And I have no idea how to do it. But I'm pretty sure it's possible. :laughing:

pages/submission/PKPSubmissionHandler.php
```diff
@@ -10,257 +10,838 @@
...
+    }
+
+    /**
+     * Get the user groups that a user can submit in
      *
-     * @param array $args
-     * @param Request $request
+     * @return UserGroup[]
+     */
+    protected function getSubmitUserGroups(Context $context, User $user): array
```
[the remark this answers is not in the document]
NateWr: I agree. Changed this to a `LazyCollection` to match `PKP\user\DAO::getMany()`.

templates/submission/review-files.tpl
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Copyright (c) 2014-2022 Simon Fraser University
+ * Copyright (c) 2003-2022 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * The template in the submission wizard when reviewing the upload files step.
+ *}
+<div class="submissionWizard__reviewPanel">
+    <div class="submissionWizard__reviewPanel__header">
+        <h3 id="review{$step.id}">
```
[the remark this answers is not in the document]
NateWr: Thanks for the reminder. I think I've got all the occurrences on the submission wizard templates.

api/v1/submissions/PKPSubmissionHandler.php
```diff
@@ -477,45 +570,119 @@ public function add($slimRequest, $response, $args)
...
+            return $response->withStatus(400)->withJson($readOnlyErrors);
         }
 
-        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_SUBMISSION, $slimRequest->getParsedBody());
         $params['id'] = $submission->getId();
         $params['contextId'] = $request->getContext()->getId();
 
-        // Don't allow a submission's context to be changed
-        if (isset($params['contextId']) && $params['contextId'] !== $submission->getData('contextId')) {
-            return $response->withStatus(403)->withJsonError('api.submissions.403.cantChangeContext');
```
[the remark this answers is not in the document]
NateWr: I removed the check because we now manually force the `contextId` to match the request, and `SubmissionRequiredPolicy` should only pass if the submission is in the current request.

(The `Repository::validate()` method only checks that the context exists, since it is meant to be able to run in the future without a request context, like from a CLI.)

classes/author/maps/Schema.php
```diff
@@ -91,12 +93,9 @@ protected function mapByProperties(array $props, Author $item): array
...
                 case 'userGroupName':
-                    $userGroupId = $item->getData('userGroupId');
-
-                    $output[$prop] = $this->authorUserGroups->contains('userGroupId', $userGroupId)
-                        ? $this->authorUserGroups->firstWhere('userGroupId', $userGroupId)->getName(null)
-                        : '';
-
+                    /** @var UserGroup $userGroup */
+                    $userGroup = $this->authorUserGroups->first(fn (UserGroup $userGroup) => $userGroup->getId() === $item->getData('userGroupId'));
+                    $output[$prop] = $userGroup ? $userGroup->getName(null) : new stdClass();
```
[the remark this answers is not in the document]
NateWr: This addresses a `json_encode` issue with associate array vs. object. `$userGroup->getName(null)` returns an assoc array, which is converted to an object in JSON, so the "empty" equivalent has to be an object in PHP.

Since `Schema` maps are only ever used for the REST API, it makes sense to map them to values appropriate for `json_encode`. But if we ever use them separately, we might have to rethink this.

classes/category/Repository.php
```diff
@@ -62,6 +63,34 @@ public function exists(int $id, int $contextId = null): bool
...
     }
 
+    /**
+     * Get the breadcrumb of a category
+     *
+     * @return string For example: Social Sciences > Anthropology
+     */
+    public function getBreadcrumb(Category $category, ?Category $parent = null): string
+    {
+        return !$parent
```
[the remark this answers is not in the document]
NateWr: Makes sense. I've implemented it the same as how we use the `common.commaListSeparator`, which still enforces a particular order but let's them change the character:

```po
msgid "common.breadcrumbSeparator"
msgstr " > "
```

```php
$parent->getLocalizedTitle()
    . __('common.breadcrumbSeparator')
    . $category->getLocalizedTitle();
```

Is that alright or do you want me to include the names in the full locale string?

classes/mail/variables/SubmissionEmailVariable.php
```diff
@@ -133,4 +129,30 @@ protected function getSubmissionUrl(Context $context): string
...
+            [
+                'id' => $this->submission->getId(),
+            ]
+        );
+    }
+
+    /**
+     * URL to the published submission
+     */
+    abstract protected function getSubmissionPublishedUrl(Context $context): string;
```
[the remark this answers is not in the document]
NateWr: Ah good catch. I've moved the relevant code into pkp-lib so consistency is enforced across the apps but there's no code duplication. :+1:

classes/migration/upgrade/v3_4_0/I7191_EditorAssignments.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Reverse the downgrades
+     *
+     * @throws DowngradeNotSupportedException
+     */
+    public function down(): void
+    {
+        Schema::table('subeditor_submission_group', function (Blueprint $table) {
+            $table->dropForeign('user_group_id');
```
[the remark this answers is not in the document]
NateWr: Thanks! I think you meant `subeditor_submission_group_user_group_id_foreign` but just want to double check.

classes/migration/upgrade/v3_4_0/I7191_EditorAssignments.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Set the user group for all existing assignments
+     *
+     * Uses the default user group for Role::ROLE_ID_SUB_EDITOR
+     * if one exists
+     */
+    protected function setUserGroup(): void
+    {
+        DB::table('user_groups')
```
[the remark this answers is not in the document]
NateWr: Maybe I've missed something, but I think that this will fall back as you suggest. The `orderBy('is_default')` is used to put the default user group first, but shouldn't exclude non-default user groups, right?

```php
DB::table('user_groups')
    ->where('role_id', '=', 17) // Role::ROLE_ID_SUB_EDITOR
    ->orderBy('is_default')
    ->get(['user_group_id', 'context_id'])
    ->each(function ($row) {
        DB::table('subeditor_submission_group')
            ->whereNull('user_group_id')
            ->where('context_id', '=', $row->context_id)
            ->update(['user_group_id' => $row->user_group_id]);
    });
```

Although now that I look at it, I think `->each()` should be `->first()`, right? That would get me the first user group in the set, with a default user group ordered before others?

> If there are several assumptions being made, maybe we ought to revisit e.g. being able to remove default user groups?

I would like to do this but maybe only once we have reduced the number of default user groups? The list feels built for a large publishing operation at the moment, and I can see lots of good UX reasons to delete the defaults for a small journal.

classes/migration/upgrade/v3_4_0/I7191_InstallSubmissionHelpDefaults.php
```diff
@@ -0,0 +1,137 @@
...
+
+abstract class I7191_InstallSubmissionHelpDefaults extends \PKP\migration\Migration
+{
+    protected string $CONTEXT_TABLE = '';
+    protected string $CONTEXT_SETTINGS_TABLE = '';
+    protected string $CONTEXT_COLUMN = '';
+
+    public function up(): void
+    {
+        if (empty($this->CONTEXT_TABLE) || empty($this->CONTEXT_SETTINGS_TABLE) || empty($this->CONTEXT_COLUMN)) {
```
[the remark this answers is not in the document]
NateWr: Makes sense. I'll do that going forward.

classes/observers/listeners/SendSubmissionAcknowledgement.php
NateWr: :+1: This collector is coded that way so it will work. The one thing the ternary does is avoid an unnecessary DB query. But since it will return 0 results it's probably not a big performance gain. I've removed the ternary condition.

classes/observers/listeners/UpdateAuthorStageAssignments.php
NateWr: Oh nice! Didn't know about `get().`

classes/publication/Repository.php
NateWr: Good call.

classes/query/QueryDAO.php
NateWr: Makes sense. I can already think of a use-case here: automatically adding an editor to the comments-for-the-editor discussion when they are assigned after submission. Want me to try to roll that into this PR or file a separate issue?

classes/query/QueryDAO.php
NateWr: In this case, `$assigned` is a `DAOResultFactory` so it's going `DAOResultFactory` -> `array` -> `Collection`.

pages/submission/PKPSubmissionHandler.php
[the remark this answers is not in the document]
asmecher: Let's leave that as a dev-infrastructure problem for later. I would be happy to start by prioritizing a locale file cleanup, to see what we're left with after that.

classes/category/Repository.php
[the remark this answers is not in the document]
asmecher: I would rather a locale key that includes both variables in the locale string, like:
```
msgid "common.breadcrumb"
msgstr "{$parentBreadcrumb} > {$childBreadcrumb}
```
This makes it clearer to translators, and makes it less likely that this will get inappropriately reused.

classes/migration/upgrade/v3_4_0/I7191_EditorAssignments.php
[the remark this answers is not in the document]
asmecher: Yes, that's right!

classes/query/QueryDAO.php
asmecher: Yes, please file it separately.

classes/migration/upgrade/v3_4_0/I7191_EditorAssignments.php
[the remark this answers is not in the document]
asmecher: Right on all counts, I think. It should be `first()`, and let's come back to defaults later.

pages/submission/PKPSubmissionHandler.php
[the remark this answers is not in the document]
asmecher: You're right, the `$handler` expects a string -- and I don't recommend we change it now. I've filed the removal of `disable_path_info` for 3.5, which should simplify URL generation; after that would be a good time to consider bigger-picture approaches to URL generation: https://github.com/pkp/pkp-lib/issues/8365

In other words, don't worry about this for now.

classes/query/QueryDAO.php
NateWr: Filed at https://github.com/pkp/pkp-lib/issues/8367.

templates/submission/review-files.tpl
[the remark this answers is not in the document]
asmecher: I think this might've gone too far -- I see one case where HTML is now being escaped:
image

classes/stageAssignment/StageAssignmentDAO.php
```diff
@@ -367,12 +371,12 @@ public function deleteByAll($submissionId, $userGroupId, $userId)
...
      * @param int $userGroupId optional
      * @param int $userId optional
-     * @param int $roleId optional ROLE_ID_...
+     * @param int[] $roleIds optional [ROLE_ID_...]
      * @param bool $single specify if only one stage assignment (default is a ResultFactory)
      *
      * @return StageAssignment|ResultFactory Mixed, depending on $single
      */
-    public function _getByIds($submissionId = null, $stageId = null, $userGroupId = null, $userId = null, $roleId = null, $single = false)
+    public function _getByIds($submissionId = null, $stageId = null, $userGroupId = null, $userId = null, $roleIds = null, $single = false)
```
asmecher, line 379: I would suggest adding a typehint for the `$roleIds` parameter in particular, in case any 3rd-party code is passing an `int` here and doesn't catch the change to `array`.

classes/userGroup/Repository.php
```diff
@@ -350,6 +350,24 @@ public function getUserCountByContextId(?int $contextId = null): Collection
...
     }
 
+    /**
+     * Get the user group a new author may be assigned to
+     * if they don't already have a role that can submit
+     *
+     * This returns the first user group with ROLE_ID_AUTHOR
+     * that permits self-registration.
+     */
+    public function getDefaultAuthorUserGroup(int $contextId): ?UserGroup
```
asmecher, line 360: The function name suggests that it'll *always* get the default user group, but this function won't behave that way. Do you think `getBestAuthorUserGroup`, `getCandidateAuthorUserGroup`, or `getPreferredAuthorUserGroup` would make more sense?

cypress/support/commands.js
```diff
@@ -360,6 +177,111 @@ Cypress.Commands.add('findSubmissionAsEditor', (username, password, familyName,
 	cy.contains('View ' + familyName).click({force: true});
 });
 
+Cypress.Commands.add('createSubmissionWithApi', (data, csrfToken) => {
```
asmecher, line 180: Really good to have some API-based data seeding patterns to build on. Not to mention API tests in the bargain.

schemas/submission.json
```diff
@@ -208,14 +216,20 @@
...
-			"type": "integer",
-			"description": "This number tracks a submission's progress before it is submitted to the journal or press. After submission, it will be set to `0`, indicating it is ready for an editor's consideration.",
+			"type": "string",
+			"description": "This is sent to the current step when a submission is in progress. After it is submitted, this will be an empty string, indicating it is ready for an editor's consideration.",
 			"apiSummary": true,
-			"default": 1,
+			"default": "files",
+			"writeDisabledInApi": true,
 			"validation": [
-				"in:0,1,2,3,4"
```
asmecher, line 216: Good riddance to magic numbers.

templates/controllers/grid/settings/category/form/categoryForm.tpl
```diff
@@ -74,13 +74,23 @@
...
+		{foreach from=$assignableUserGroups item="assignableUserGroup"}
+			{assign var="role" value=$assignableUserGroup.userGroup->getLocalizedName()}
+			{assign var="userGroupId" value=$assignableUserGroup.userGroup->getId()}
+			{foreach from=$assignableUserGroup.users item=$username key="id"}
+				{fbvElement
+					type="checkbox"
+					id="subEditors[{$userGroupId}][]"
+					value=$id
+					checked=in_array($id, $assignedSubeditors)
+					label={translate key="manager.sections.form.assignEditorAs" name=$username role=$role}
```
asmecher, line 88: Don't forget to escape user-supplied data (or anything that shouldn't be HTML) to `{translate}` calls as well -- 
```
{translate key="manager.sections.form.assignEditorAs" name=$username|escape role=$role|escape}

File: templates/submission/review-details.tpl [hunk 25]
```diff
@@ -0,0 +1,68 @@
...
+                                {{ error }}
+                            </notification>
+                        </template>
+                        <h4 class="submissionWizard__reviewPanel__item__header">
+                            {translate key="submission.citations"}
+                        </h4>
+                        <div
+                            class="submissionWizard__reviewPanel__item__value"
+                            v-html="
+                                publication.citationsRaw
```
asmecher, line 55: Should `citationsRaw` be allowed arbitrary HTML?

asmecher (changes requested):
Just a few more comments on the first round, mostly minor!

classes/components/forms/publication/TitleAbstractForm.php
```diff
@@ -0,0 +1,78 @@
...
+abstract class TitleAbstractForm extends FormComponent
+{
+    public $id = FORM_TITLE_ABSTRACT;
+    public $method = 'PUT';
+    public $publication;
+
+    /** @var bool Whether or not this form is for the submission wizard */
+    public bool $isSubmissionWizard = false;
+
+    /** @var int The abstract word limit for this submission */
```
asmecher, line 33: Add self-doc: "or 0 for no limit"

templates/submission/review-files.tpl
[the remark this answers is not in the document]
NateWr: Woops! Fixed that. :)

classes/security/authorization/internal/PublicationRequiredPolicy.php
```diff
@@ -16,25 +16,27 @@
...
+     * @param ?string $publicationParameterName the request parameter we expect
+     *  the submission id in. Pass null to look for the current publication in an authorized
+     *   submission object instead.
      * @param null|mixed $operations
      */
-    public function __construct($request, &$args, $publicationParameterName = 'publicationId', $operations = null)
+    public function __construct($request, &$args, $publicationParameterName = null, $operations = null)
     {
-        parent::__construct($request, $args, $publicationParameterName, 'user.authorization.invalidPublication', $operations);
+        parent::__construct($request, $args, $publicationParameterName ?? '', 'user.authorization.invalidPublication', $operations);
```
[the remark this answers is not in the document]
NateWr: I can make this change, but it means my code will now impact 10 policies that extend `DataObjectRequiredPolicy`. Maybe there won't be a problem, but I never feel confident making such a change with the base security policy classes because it's very hard to test for side effects. Happy to do whatever you think is best here.

classes/stageAssignment/StageAssignmentDAO.php
NateWr: The way that it's currently set up, the function will work if someone passes an `int` to the `$roleIds`. We coerce the value to an array, which should turn `1` into `[1]`:

```php
$sanitizedRoleIds = join(',', array_map(fn ($roleId) => (int) $roleId, (array) $roleIds));
```

It's not pretty, but again I was erring on the side of caution. Happy to add the type hint if you want.

classes/userGroup/Repository.php
NateWr: I see what you mean. Since this is only used when determining what a user should submit as, what about `getFirstSubmitAsAuthorUserGroup()`? It's a mouthful, but that will prevent people from using it for other purposes, which is I think the confusion we want to avoid.

templates/submission/review-details.tpl
NateWr: Good spot! I was trying to preserve line breaks but I've fixed that by splitting to an array and putting each citation in a div. :+1:

classes/stageAssignment/StageAssignmentDAO.php
asmecher: Ah -- that's fine, I missed that. For new code I prefer strict typing (it's so easy to just throw `[]` around a single value) but for deprecated code like this backwards compatibility is probably the lesser evil.

classes/userGroup/Repository.php
asmecher: ...yes, it is a mouthful, but it's clear. Go for it.

classes/security/authorization/internal/PublicationRequiredPolicy.php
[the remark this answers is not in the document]
asmecher: I'd be confident about this (i.e. if it causes downstream problems I'll deal with it :smile:).

It tightens up error handling, establishes `null` as a valid `$parameterName` indicating "we dont do things that way, thanks".
```diff
diff --git a/lib/pkp/classes/security/authorization/DataObjectRequiredPolicy.php b/lib/pkp/classes/security/authorization/DataObjectRequiredPolicy.php
index be8c1e290d..2fca04d274 100644
--- a/lib/pkp/classes/security/authorization/DataObjectRequiredPolicy.php
+++ b/lib/pkp/classes/security/authorization/DataObjectRequiredPolicy.php
@@ -22,8 +22,7 @@ class DataObjectRequiredPolicy extends AuthorizationPolicy
     /** @var array */
     public $_args;
 
-    /** @var string */
-    public $_parameterName;
+    public ?string $_parameterName;
 
     /** @var array */
     public $_operations;
@@ -56,11 +55,11 @@ class DataObjectRequiredPolicy extends AuthorizationPolicy
      *
      * @param PKPRequest $request
      * @param array $args request parameters
-     * @param string $parameterName the request parameter we expect
+     * @param ?string $parameterName the request parameter we expect
      * @param string $message
      * @param array $operations Optional list of operations for which this check takes effect. If specified, operations outside this set will not be checked against this policy.
      */
-    public function __construct($request, &$args, $parameterName, $message = null, $operations = null)
+    public function __construct($request, &$args, ?string $parameterName, $message = null, $operations = null)
     {
         parent::__construct($message);
         $this->_request = $request;
@@ -114,7 +113,7 @@ class DataObjectRequiredPolicy extends AuthorizationPolicy
         $router = $this->_request->getRouter();
         switch (true) {
             case $router instanceof \PKP\core\PKPPageRouter:
-                if (ctype_digit((string) $this->_request->getUserVar($this->_parameterName))) {
+                if ($this->_parameterName !== null && ctype_digit((string) $this->_request->getUserVar($this->_parameterName))) {
                     // We may expect a object id in the user vars
                     return (int) $this->_request->getUserVar($this->_parameterName);
                 } elseif (!$lookOnlyByParameterName && isset($this->_args[0]) && ctype_digit((string) $this->_args[0])) {
@@ -125,19 +124,20 @@ class DataObjectRequiredPolicy extends AuthorizationPolicy
 
             case $router instanceof \PKP\core\PKPComponentRouter:
                 // We expect a named object id argument.
-                if (isset($this->_args[$this->_parameterName])
+                if ($this->_parameterName !== null && isset($this->_args[$this->_parameterName])
                         && ctype_digit((string) $this->_args[$this->_parameterName])) {
                     return (int) $this->_args[$this->_parameterName];
                 }
                 break;
 
             case $router instanceof \PKP\core\APIRouter:
-                $handler = $router->getHandler();
-                return $handler->getParameter($this->_parameterName);
+                if ($this->_parameterName !== null) {
+                    $handler = $router->getHandler();
+                    return $handler->getParameter($this->_parameterName);
+                }
                 break;
 
-            default:
-                assert(false);
+            default: throw new Exception('DataObjectRequiredPolicy does not support routers of type ' . get_class($router) . '!');
         }
 
         return false;
diff --git a/classes/security/authorization/OjsIssueRequiredPolicy.php b/classes/security/authorization/OjsIssueRequiredPolicy.php
index 2fa69432d5..50451eff42 100644
--- a/classes/security/authorization/OjsIssueRequiredPolicy.php
+++ b/classes/security/authorization/OjsIssueRequiredPolicy.php
@@ -92,7 +92,7 @@ class OjsIssueRequiredPolicy extends DataObjectRequiredPolicy
         $router = $this->_request->getRouter();
         switch (true) {
             case $router instanceof \PKP\core\PKPPageRouter:
-                if (ctype_digit((string) $this->_request->getUserVar($this->_parameterName))) {
+                if ($this->_parameterName !== null && ctype_digit((string) $this->_request->getUserVar($this->_parameterName))) {
                     // We may expect a object id in the user vars
                     return (int) $this->_request->getUserVar($this->_parameterName);
                 } elseif (isset($this->_args[0])) {
```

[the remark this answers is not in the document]
NateWr: Thanks for the diff!

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -23,6 +23,7 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
     abstract protected function getContextTable(): string;
     abstract protected function getContextSettingsTable(): string;
     abstract protected function getContextKeyField(): string;
+    abstract protected function getContextSettingsTable(): string;
```
asmecher, line 26: Looks like there was a rebase hiccough -- this is declared twice.

asmecher:
It's not holding me back, but just a heads-up that the OJS and pkp-lib pulls are out of sync; OJS includes a reference to `PKP\migration\upgrade\v3_4_0\I7245_UpdateUserLocaleStringToParsableJsonString` in `dbscripts/xml/upgrade.xml` but this base version of pkp-lib doesn't include it yet. This should get resolved with the next rebase, I expect.

NateWr:
I've been playing a bit loose with the branches. I've now rebased, synced and pushed up for ui-library, pkp-lib and ojs.

classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
```diff
@@ -0,0 +1,70 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+class I7191_SubmissionProgressType extends \PKP\migration\Migration
+{
+    public function up(): void
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->text('submission_progress')->default('files')->change();
```
asmecher, line 26: I was getting an exception when running the updates here that suggested that the column type hadn't changed:
```php
ERROR: Upgrade failed: DB: SQLSTATE[22007]: Invalid datetime format: 1292 Truncated incorrect DOUBLE value: '' (SQL: update `submissions` set `submission_progress` = files where `submission_progress` = 1)
```
Oddly, checking in the MySQL client, I got a double type as expected, so there's some kind of "sure-it's-changed-but-not-quite-yet" situation going on.

I wasn't able to track down the cause, but I was able to work around it by re-using the new column creation you've done below on the `down()` function:

```php
diff --git a/classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php b/classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
index c37a6906d5..6da3d869fc 100644
--- a/classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
+++ b/classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
@@ -23,16 +23,21 @@ class I7191_SubmissionProgressType extends \PKP\migration\Migration
     public function up(): void
     {
         Schema::table('submissions', function (Blueprint $table) {
-            $table->text('submission_progress')->default('files')->change();
+            $table->text('submission_progress_temp')->default('files');
         });
 
         foreach ($this->getStepMap() as $oldValue => $newValue) {
             DB::table('submissions')
                 ->where('submission_progress', $oldValue)
                 ->update([
-                    'submission_progress' => $newValue,
+                    'submission_progress_temp' => $newValue,
                 ]);
         }
+
+        Schema::table('submissions', function (Blueprint $table) {
+            $table->dropColumn('submission_progress');
+            $table->renameColumn('submission_progress_temp', 'submission_progress');
+        });
     }
 
     public function down(): void
```

Since this is what you're using in `down()` anyway, I'd suggest we just apply this change and move on.

classes/migration/upgrade/v3_4_0/I7191_ResubscribeSubeditors.php
```diff
@@ -0,0 +1,80 @@
...
+                    ->on('ug.context_id', '=', 'nss.context')
+                    ->where('ug.role_id', '=', self::SUBEDITOR_ROLE);
+            })
+            ->distinct()
+            ->where('nss.setting_name', 'blocked_emailed_notification')
+            ->where('nss.setting_value', self::NOTIFICATION_TYPE)
+            ->whereNotNull('ug.user_group_id')
+            ->get(['nss.*', 'nss.context']);
+
+        DB::table('notification_subscription_settings as nss')
```
asmecher, line 62: At least on my version of MariaDB, it's not allowed to specify a table alias in a `DELETE` query. I get:
```
ERROR: Upgrade failed: DB: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'as `nss` where 0 = 1' at line 1 (SQL: delete from `notification_subscription_settings` as `nss` where 0 = 1)
```
Removing the `as nss` fixes it for me.

classes/migration/install/SubmissionsMigration.php
```diff
@@ -175,6 +177,10 @@ public function up(): void
             $table->foreign('user_id', 'subeditor_submission_group_user_id')->references('user_id')->on('users')->onDelete('cascade');
             $table->index(['user_id'], 'subeditor_submission_group_user_id');
 
+            $table->bigInteger('user_group_id');
+            $table->foreign('user_group_id')->references('user_group_id')->on('user_groups');
```
asmecher, line 181: I'd recommend adding `onDelete('cascade')` here. Currently, if someone attempts to delete the user group without first clearing out this table, it'll cause an error.

classes/context/PKPSectionDAO.php
```diff
@@ -36,6 +54,17 @@ abstract public function newDataObject();
...
     abstract public function getById($sectionId, $contextId = null);
 
+    /**
+     * Check if a section exists
+     */
+    public function exists(int $sectionId, ?int $contextId = null)
+    {
+        return DB::table($this->_getTableName())
+            ->where($this->_getIdColumnName(), $sectionId)
+            ->when($contextId, fn (Builder $q) => $q->where($this->_getContextIdColumnName(), $contextId))
```
asmecher, line 64: Use strict checks in `->when` conditions to avoid ambiguity between `0` and `null`. Preferred form:
```php
->when($contextId !== null, fn (Builder $q) => $q->where($this->_getContextIdColumnName(), $contextId))
```

classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
asmecher: This didn't 100% solve the problem as `doctrine/dbal` was renaming the column using e.g...
```sql
ALTER TABLE submissions CHANGE submission_progress_temp submission_progress TEXT NOT NULL
```
...which has the unfortunate side-effect of dropping e.g. the `DEFAULT 'files'` clause, causing errors on new article creation after an upgrade. Luckily this appears to have been fixed in a newer release of `doctrine/dbal`, and updating it solves the `DEFAULT` problem -- but not the need for the patch above.

So I recommend committing the patch above, plus also [updating `doctrine/dbal`](https://github.com/NateWr/pkp-lib/pull/11/files).

classes/components/forms/submission/ConfirmSubmission.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace PKP\components\forms\submission;
+
+use PKP\components\forms\FieldOptions;
+use PKP\components\forms\FormComponent;
+use PKP\context\Context;
+
+define('FORM_CONFIRM_SUBMISSION', 'confirmSubmission');
+
+class ConfirmSubmission extends FormComponent
```
asmecher, line 23: It appears to be possible to make a submission to an inactive section if the section is made inactive after the submission has been started. I'd suggest putting a check for the current section being inactive on the submission confirmation form.

classes/context/SubEditorsDAO.php
```diff
@@ -113,32 +119,22 @@ public function deleteBySubmissionGroupId($assocId, $assocType, $contextId = nul
...
     /**
      * Delete all submission group assignments for the specified user.
-     *
-     * @param int $userId
-     * @param int $contextId optional, include assignments only in this context
-     * @param int $assocId optional, include only this submission group
-     * @param int $assocType optional ASSOC_TYPE_SECTION or ASSOC_TYPE_CATEGORY
      */
-    public function deleteByUserId($userId, $contextId = null, $assocId = null, $assocType = null)
+    public function deleteByUserId(int $userId)
```
asmecher, line 123: If this were not such a new and niche function I would be tempted to add a check at the top:
```php
// TEMPORARY: This check was added with 3.4.0 when additional checks were removed from the function. Remove after next LTE.
if (func_num_args() !== 1) throw new Exception('Invalid number of parameters passed to SubEditorsDAO::deleteByUserId!');
```
...but the likelihood of this being encountered are close enough to 0 that I'm OK to leave it out if you'd prefer.

(Trivia: [this PHP-level proposal](https://wiki.php.net/rfc/strict_argcount) to warn about excess parameters was proposed but utterly defeated in the vote.)

classes/context/SubEditorsDAO.php
```diff
@@ -66,29 +83,18 @@ public function deleteEditor($contextId, $assocId, $userId, $assocType)
...
-        $users = [];
-        foreach ($result as $row) {
-            $users[$row->user_id] = $userDao->fromRow($row);
-        }
-        return $users;
+        return DB::table('subeditor_submission_group')
+            ->where('assoc_type', '=', $assocType)
+            ->where('context_id', '=', $contextId)
+            ->whereIn('assoc_id', $assocIds)
+            ->get(['user_id', 'user_group_id']);
```
asmecher, line 97: This is probably the first place in the system where we're exposing `database_column_naming` outside of a DAO's internal implementation rather than our usual convention of `camelCase` in PHP. I know it adds cruft but I'm pretty hesitant to break our current clean separation of styles. I'd suggest giving these lowerCamelCase aliases if this is going to be a public function.

classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php
NateWr: Thanks!

classes/components/forms/submission/ConfirmSubmission.php
NateWr: Ahh, yeah. These checks are there when creating a new submission but not submitting it. This presents a little bit of a challenge because there is not anywhere to put such an error on the review step. I've filed a separate issue to tackle this so it doesn't hold up this merge: https://github.com/pkp/pkp-lib/issues/8408

classes/context/SubEditorsDAO.php
NateWr: Funny, I guess I just assumed that PHP threw an exception in such cases! I've added in the exception. :+1:

classes/context/SubEditorsDAO.php
```diff
@@ -160,6 +156,136 @@ public function editorExists($contextId, $assocId, $userId, $assocType)
...
+                $unsubscribed = in_array(
+                    Notification::NOTIFICATION_TYPE_SUBMISSION_SUBMITTED,
+                    $notificationSubscriptionSettingsDao->getNotificationSubscriptionSettings(
+                        NotificationSubscriptionSettingsDAO::BLOCKED_EMAIL_NOTIFICATION_KEY,
+                        $editorAssignment->getUserId(),
+                        $context->getId()
+                    )
+                );
+
+                if ($unsubscribed && !in_array($editorAssignment->getUserId(), $notifiedEditors)) {
```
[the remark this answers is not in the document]
NateWr: Good catch! The `!` needed to be removed from `in_array()` too.

classes/context/SubEditorsDAO.php
NateWr: Good call. I've aliased the select columns as `userId` and `userGroupId`.

cypress/support/commands.js
```diff
@@ -360,6 +177,111 @@ Cypress.Commands.add('findSubmissionAsEditor', (username, password, familyName,
...
+					});
+				});
+		});
+});
+
+Cypress.Commands.add('submitSubmissionWithApi', (id, csrfToken) => {
+	const api = new Api(Cypress.env('baseUrl') + '/index.php/publicknowledge/api/v1');
+	return cy.request({
+			url: api.submit(id),
+			method: 'PUT',
```
asmecher, line 275: The OPS API does not appear to support a `PUT` to this endpoint. I get:
image

asmecher: Never mind -- this seems to have been a local problem. Not sure why but it's gone away now.

NateWr: Looks like the `baseUrl` bug. I see `/undefined/` in the request URL. :+1:

classes/observers/listeners/AssignEditors.php
```diff
@@ -0,0 +1,112 @@
...
+                    $manager->getId(),
+                    $event->context->getId()
+                )
+            );
+
+            if ($unsubscribed) {
+                continue;
+            }
+
+            Mail::send($mailable->recipients([$manager]));
```
Vitaliy-1, line 101: @NateWr, there were 2 associated notification emails before, one related to `NOTIFICATION_TYPE_EDITOR_ASSIGNMENT_REQUIRED` (A new article has been submitted to which an editor needs to be assigned) and another one `NOTIFICATION_TYPE_SUBMISSION_SUBMITTED` (A new article, "Title," has been submitted). Does this mean that you combined them? They both are still present in profile notifications settings

NateWr: I believe that `NOTIFICATION_TYPE_EDITOR_ASSIGNMENT_REQUIRED` is only triggered if no editors are automatically assigned.

PR #8353, opened 2022-10-19
pkp/pkp-lib#7318 stats issues page changes according to the code review

bozana:


locale/en_US/manager.po
```diff
@@ -2514,10 +2514,13 @@ msgstr "Monthly"
 msgid "stats.pdf"
 msgstr "PDF"
 
-msgid "stats.publications.downloadReport.betweenDates"
+msgid "stats.geographic.ccAttribution"
+msgstr "<a href='https://db-ip.com'>IP Geolocation by DB-IP</a>"
```
NateWr, line 2518: @asmecher Bozana said that she spoke with you about the CC attribution for IP geolocation. Do we need to put it in such a prominent place in the UI?

geolocation-attribution

Would it be acceptable to put it in with the stats settings instead, like this?

geolocation-attribution-in-settings

templates/stats/publications.tpl
```diff
@@ -233,13 +233,16 @@
...
+										<p id="report-type-geographic-description">
 											{translate key="stats.publications.downloadReport.downloadGeographic.description"}
 										</p>
+										<p id="report-type-geographic-ccAttribution">
+											{translate key="stats.geographic.ccAttribution"}
+										</p>
 										<template slot="actions">
 											<pkp-button
-												aria-describedby="report-type-article report-type-article-description"
+												aria-describedby="report-type-geographic report-type-geographic-description"
```
NateWr, line 245: I think my example code probably included the `aria-describedby` attributes but I don't think we need them (or the `id` attributes they reference). That's because the download buttons on the publications and issue stats pages all have descriptive labels: Download Articles, Download Files, etc.

We would use the `aria-describedby` labels in cases where we had several buttons with identical names -- like Download, Download, Download -- and needed to distinguish them for screen readers.

locale/en_US/manager.po
asmecher: The license for it says:
> You are free to use this IP to City Lite database in your application, provided you give attribution to DB-IP.com for the data.
>
> In the case of a web application, you must include a link back to DB-IP.com on pages that display or use results from the database. You may do it by pasting the HTML code snippet below into your code

Thus wherever we present results that draw from that library we have to also present a link to the source.

NateWr: Hmm, I guess that means the CSV file itself, then. Right? Because we don't show the geographic stats anywhere except in the CSV file.

bozana: Yes, till now it is only the API (CSV or JSON) and that Statistics > Articles/Monographs/Preprints page, where the user can download the CSV.
Regarding the API I would agree with Alec: "...APIs are meant to be machine-readable and the license appears to be talking about human-facing content".
I would think it is sufficient to give the attribution near the Download-Button in the UI (and e.g. in the API documentation), but I am not sure if we would need to put it somehow into the downloaded CSV-file (which somehow feels a little bit strange)?

NateWr: Ok, I guess we can't get around it. It just sticks out like a sore thumb there. Here are two alternatives.

A tooltip that says "Geolocation provided by [DB-IP](#)."

tooltip

Or as a small footer at the bottom of the article stats page.

footer

bozana: If no objection I would choose the tooltip for now...

bozana: @NateWr, what should be the tooltip label?

asmecher: ^ I'm OK with that solution as well. I think we can meet the terms without such a glaring diversion from our usual "neutral" UI practices.

NateWr: @bozana The tooltip should read:

> Geolocation provided by [DB-IP](#).

bozana: Also the label part, for the screenreader?

NateWr: It can be something like "About Geographic"

PR #8383, opened 2022-10-29
pkp/pkp-lib#7318 Stats Context Page

bozana:

### DIFF
--- a/templates/stats/context.tpl
+++ b/templates/stats/context.tpl
@@ -1,0 +1,189 @@
+{**
+ * templates/stats/context.tpl
+ *
+ * Copyright (c) 2022 Simon Fraser University
+ * Copyright (c) 2022 John Willinsky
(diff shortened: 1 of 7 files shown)


templates/stats/context.tpl
```diff
@@ -0,0 +1,189 @@
...
+									</p>
+									<template slot="actions">
+										<pkp-button
+											@click="downloadReport"
+										>
+											{translate key="stats.context.downloadReport.downloadContext"}
+										</pkp-button>
+									</template>
+								</action-panel>
+								<action-panel class="pkpStats__reportAction">
```
ewhanson, line 136: I'm not seeing where this second action panel shows up in the UI. The same for the issue and article stats pages. I may be missing something as I don't have actual stats data to test with, but I wanted to flag it.

Issue #8392, opened 2022-11-02
Email client safelink cause user password reset feature broken.


touhidurabir:
**Describe the bug**
This issue ported from forum and details at https://forum.pkp.sfu.ca/t/e-mail-client-safelink-feature-expires-password-reset-confirmation-emails-sorry-the-link-you-clicked-on-has-expired-or-is-not-valid-please-try-resetting-your-password-again/75112/3

**To Reproduce**
1. Request for a password reset
2. Open the mail in MS Outlook and click on the given reset link
3. See that it shows message as `Sorry, the link you clicked on has expired or is not valid. Please try resetting your password again.`
4. Check mailbox again and see that a new password already mailed which cause by `safelink` feature to execute the code of `LoginHandler::resetPassword` before user redirected to the give link properly . 

**What application are you using?**
 OJS, OMP or OPS version `3.3.0-13` and `main`

**Additional information**
one possible solution already mentioned at https://forum.pkp.sfu.ca/t/e-mail-client-safelink-feature-expires-password-reset-confirmation-emails-sorry-the-link-you-clicked-on-has-expired-or-is-not-valid-please-try-resetting-your-password-again/75112/2?u=touhidur

#### Update
decided to implement the feature request from https://github.com/pkp/pkp-lib/issues/2135 to address this issue . It will change the entire password reset process so that outlook's safelink feature will not have any impact on process while it verify an included link in the mail .

asmecher:
If someone is able to confirm whether or not the Safelink tool results in a `HEAD` request to OJS (as opposed to e.g. `GET`), it'll help us know how to fix this!

touhidurabir:
> someone is able to confirm whether or not the Safelink tool results in a HEAD request to OJS (as opposed to e.g. GET)

@asmecher I am thinking that what we can do it is to check for any request in the `LoginHandler::resetPassword` that is not a `GET` and redirect back to site/context home page . I think that way we can resolve this one .

touhidurabir:
@asmecher PRs to review . 

[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8400
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3602

if this PRs got accepted , I will forward port it for the `stable 3.3.0`

asmecher:
Thanks, @touhidurabir, I've added a comment on the PR:

[...]

PR #8400, opened 2022-11-03
pkp/pkp-lib#8392 redirection added to password reset for non GET request

touhidurabir:
This PR aims to solve pkp/pkp-lib#8392

### DIFF
--- a/pages/login/LoginHandler.php
+++ b/pages/login/LoginHandler.php
@@ -249,6 +249,10 @@ class LoginHandler extends Handler
      */
     public function resetPassword($args, $request)
     {
+        if (!$request->isGet()) {
+            return $request->redirect($request->getContext()?->getData('urlPath') ?? $request->getBasePath());
+        }
+
         $this->setupTemplate($request);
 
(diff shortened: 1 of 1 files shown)


pages/login/LoginHandler.php
```diff
@@ -249,6 +249,10 @@ public function requestResetPassword($args, $request)
      */
     public function resetPassword($args, $request)
     {
+        if (!$request->isGet()) {
```
asmecher, line 252: I would suggest using...
```php
if ($request->getRequestMethod() === 'HEAD') {
    return;
}
```
...and moving the code block just below `$templateMgr = TemplateManager::getManager($request);`.

Benefits:
- Explicitly checks for a `HEAD` response, which is the use case we're trying to capture
- A `HEAD` response should emulate the `GET` response, except without a response body. A redirect doesn't do that.

touhidurabir:
Closing the PR as we decided to implement https://github.com/pkp/pkp-lib/issues/2135 .

PR #8439, opened 2022-11-16
pkp/pkp-lib#7709 Fix editorial statistics

jonasraoni:


asmecher:
@bozana, a fix like this has already been merged into `stable-3_3_0`; here is the forward-port to `main`. Can you have a quick look over it for issues that might arise from the stats refactor? Do you have outstanding work to merge that this might conflict with?

bozana:
@asmecher, @jonasraoni, this is not in conflict with my work -- I have not touched the editorial stats :-) -- so yes, please go ahead :+1: 
But, after taking a look at the changes, I do have a question: why are we not considering the start and end date for these skipped submissions? -- sorry if this is already explained somewhere else...

jonasraoni:
@bozana Oh, I've seen this while testing, but looks like I forgot to handle, thanks for catching it =]

PR #8453, opened 2022-11-27
Bugfix/stable 3 3 0/8442 fix date filter

jonasraoni:


classes/components/PKPStatsComponent.inc.php
```diff
@@ -73,6 +73,7 @@ public function getConfig() {
 			'apiUrl' => $this->apiUrl,
 			'tableColumns' => $this->tableColumns,
 			'dateStart' => $this->dateStart,
+			'dateStartMin' =>  preg_replace('/^(\d{4})(\d{2})(\d{2})/', '$1-$2-$3', STATISTICS_EARLIEST_DATE),
```
jonasraoni, line 76: On `main` this constant is defined as '2001-01-01' while here '20010101'.

PR #8457, opened 2022-11-29
Add definition to avoid undefined variable warning

ewhanson:
Hey @asmecher, could you have a quick look at this? I came across it when hunting down why my tests were failing in OMP. It was only generating an "undefined variable" warning for `$administrationLevel`, but it was popping up a lot. Thanks!


ewhanson:
Closed as @touhidurabir has this covered in a separate PR.

PR #8460, opened 2022-12-01
pkp/pkp-lib#7836 Updated condition to display possible list to associate with discussion

touhidurabir:
This PR aims to solve pkp/pkp-lib#7836

### DIFF
--- a/controllers/grid/queries/form/QueryForm.inc.php
+++ b/controllers/grid/queries/form/QueryForm.inc.php
@@ -44,12 +44,12 @@ class QueryForm extends Form {
 		parent::__construct('controllers/grid/queries/form/queryForm.tpl');
 		$this->setStageId($stageId);
 
-		$queryDao = DAORegistry::getDAO('QueryDAO'); /* @var $queryDao QueryDAO */
+		$queryDao = DAORegistry::getDAO('QueryDAO'); /** @var QueryDAO $queryDao */
 		if (!$queryId) {
 			$this->_isNew = true;
 
 			// Create a query
-			$query = $queryDao->newDataObject();
+			$query = $queryDao->newDataObject(); /** @var Query $query */
 			$query->setAssocType($assocType);
 			$query->setAssocId($assocId);
 			$query->setStageId($stageId);
@@ -61,7 +61,7 @@ class QueryForm extends Form {
 			$queryDao->insertParticipant($query->getId(), $request->getUser()->getId());
 
 			// Create a head note
-			$noteDao = DAORegistry::getDAO('NoteDAO'); /* @var $noteDao NoteDAO */
+			$noteDao = DAORegistry::getDAO('NoteDAO'); /** @var NoteDAO $noteDao */
 			$headNote = $noteDao->newDataObject();
 			$headNote->setUserId($request->getUser()->getId());
 			$headNote->setAssocType(ASSOC_TYPE_QUERY);
@@ -170,7 +170,7 @@ class QueryForm extends Form {
 	 * Initialize form data from the associated author.
 	 */
 	function initData() {
-		$queryDao = DAORegistry::getDAO('QueryDAO'); /* @var $queryDao QueryDAO */
+		$queryDao = DAORegistry::getDAO('QueryDAO'); /** @var QueryDAO $queryDao */
 		if ($query = $this->getQuery()) {
 			$headNote = $query->getHeadNote();
 			$this->_data = array(
@@ -197,7 +197,7 @@ class QueryForm extends Form {
 
 		$query = $this->getQuery();
 		$headNote = $query->getHeadNote();
-		$user = $request->getUser();
+		$user = $request->getUser(); /** @var User $user */
 		$context = $request->getContext();
 
 		$templateMgr = TemplateManager::getManager($request);
@@ -211,10 +211,10 @@ class QueryForm extends Form {
 		// Queryies only support ASSOC_TYPE_SUBMISSION so far
 		if ($query->getAssocType() == ASSOC_TYPE_SUBMISSION) {
 
-			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
+			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /** @var StageAssignmentDAO $stageAssignmentDao */
 
 			// Get currently selected participants in the query
-			$queryDao = DAORegistry::getDAO('QueryDAO'); /* @var $queryDao QueryDAO */
+			$queryDao = DAORegistry::getDAO('QueryDAO'); /** @var QueryDAO $queryDao */
 			$assignedParticipants = $query->getId() ? $queryDao->getParticipantIds($query->getId()) : array();
 
 			// Always include current user, even if not with a stage assignment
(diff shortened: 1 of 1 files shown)


controllers/grid/queries/form/QueryForm.inc.php
```diff
@@ -226,21 +226,20 @@ function fetch($request, $template = null, $display = false, $actionArgs = array
...
 					$userGroup = $userGroupDao->getById($usersAssignment->getUserGroupId());
 					$assignedRoles[] = $userGroup->getRoleId();
 				}
-
-				// if current user is editor, add all reviewers
-				if ($user->hasRole([ROLE_ID_SITE_ADMIN], CONTEXT_SITE) || $user->hasRole([ROLE_ID_MANAGER], $context->getId()) ||
-						array_intersect([ROLE_ID_MANAGER, ROLE_ID_SUB_EDITOR], $assignedRoles)) {
+				
+				// if current user is editor/journal manager/site admin and not have author role , add all reviewers
+				if (($user->hasRole([ROLE_ID_SITE_ADMIN], CONTEXT_SITE) || $user->hasRole([ROLE_ID_MANAGER, ROLE_ID_SUB_EDITOR], $context->getId())) && empty(array_intersect($assignedRoles, [ROLE_ID_AUTHOR])) ) {
```
asmecher, line 242: @touhidurabir, I see the logic in adding the check to exclude users who are assigned with `ROLE_ID_AUTHOR`, but it looks like this change also turns the check for `ROLE_ID_MANAGER` and `ROLE_ID_SUB_EDITOR` from a check on `$assignedRoles` (which is about assignments to the submission) to a call to `$user->hasRole` (which is based on users being assigned to roles in the journal). Was that intentional?

touhidurabir: @asmecher yes that is intentional . It felt to me that any one who has `manager`, `site admin` or `editor` can  see associated users/reviewers, not just those who are associated with stage/submission and the exception is only when has an `author` role associated with that submission .  Now that you have mention this, wondering if I am wrong to such thoughts .

asmecher: @ajnyga might have a better sense than me; if I remember correctly he originally added the code to protect reviewer/author anonymization in cases where a user had multiple roles. @ajnyga, did you have a policy in mind for checking role assignments on the current submission vs. on the context?

NateWr: The logic around this is difficult, but we try to let stage assignments override context roles whenever possible. If a user is assigned to a submission, treat them only according to their assigned roles. If they are not assigned to a submission _in any role_, then we fall back to their context roles.

So I think the appropriate logic here, in plain terms, would be:

```php
if (
    array_intersect($assignedRoles, [ROLE_ID_MANAGER, ROLE_ID_SUB_EDITOR]) 
    || 
    (
        empty($assignedRoles)
        &&
        (
            $user->hasRole([ROLE_ID_MANAGER, $context->getId()])
            || 
            $user->hasRole([ROLE_ID_SITE_ADMIN, CONTEXT_SITE])
        )
    )
) {
```

Subeditors do not have any permissions on submissions when they are not assigned, so we don't include them in a `$user->hasRole()` check.

@touhidurabir would you like me to pick this up or do you have time to make the changes for me to review?

touhidurabir: > Subeditors do not have any permissions on submissions when they are not assigned, so we don't include them in a $user->hasRole() check.

@NateWr I was not aware of this particular check . In that sense your proposed check is the right approach as I did some manual testing . Updated it

controllers/grid/queries/form/QueryForm.inc.php
```diff
@@ -226,21 +226,20 @@ function fetch($request, $template = null, $display = false, $actionArgs = array
...
 					$userGroup = $userGroupDao->getById($usersAssignment->getUserGroupId());
 					$assignedRoles[] = $userGroup->getRoleId();
 				}
-
-				// if current user is editor, add all reviewers
-				if ($user->hasRole([ROLE_ID_SITE_ADMIN], CONTEXT_SITE) || $user->hasRole([ROLE_ID_MANAGER], $context->getId()) ||
-						array_intersect([ROLE_ID_MANAGER, ROLE_ID_SUB_EDITOR], $assignedRoles)) {
+				
+				// if current user is editor/journal manager/site admin and not have author role , add all reviewers
+				if ( array_intersect($assignedRoles, [ROLE_ID_MANAGER, ROLE_ID_SUB_EDITOR]) || ( empty($assignedRoles) && ( $user->hasRole([ROLE_ID_MANAGER, $context->getId()]) || $user->hasRole([ROLE_ID_SITE_ADMIN, CONTEXT_SITE]) ) ) ) {
```
NateWr, line 242: This is a typo from my sample code, but the `[]` brackets are wrong in `$user->hasRole([ROLE_ID_MANAGER, $context->getId()])`. It should be `$user->hasRole([ROLE_ID_MANAGER], $context->getId())`.

The same is true for the `$user->hasRole()` check for admin.

NateWr (changes requested):

controllers/grid/queries/form/QueryForm.inc.php
touhidurabir: Fixed it . I should have noticed it too, sorry.

NateWr:
Ok, this looks good. I've tested it myself and it's working. However, while testing I recalled one more condition :sweat_smile:. When a review assignment is open (`SUBMISSION_REVIEW_METHOD_OPEN`), then it should be possible for all assigned users to see and message each other.

There seems to be some logic for this at line 260 of `QueryForm.inc.php`. However, open reviewers weren't included in the liast of possible discussion recipients when I tested it just now.

To make a review assignment open, you can select "open" for the method when creating it or click "Edit Review" on an existing review and change the method.

touhidurabir:
@NateWr if I understand it correctly, if a review assignment marked as `Open`, then all assigned users will be able to view the associated users, even considering the case mentioned in the issue pkp/pkp-lib#7836 where an `author` also have `editor` role . 

I tested and it seems it's including all the associated users as discussion participants when set the review assignment type `Open`. see the screenshot below 

CleanShot 2023-01-19 at 13 06 08@2x

CleanShot 2023-01-19 at 13 04 53@2x

Testing Steps : 
1. Assigned `author` role to an existing user with only `editor` role . 
2. Submit a new submission.
3. Assigned a different editor to that submission
4. Add reviewer and accept as that reviewer in a Review Round 
5. when the review assignment type set as `Anonymous Reviewer/Anonymous Author`, as the author , starting a review discussion does not disclose the reviewer details . 
6. Setting the review assigned type as `Open` allow the author to view review name in the review discussion list 

If this is correct behaviour, then all seems ok . But I am not aware of the exact behaviour so please confirm if the above workflow ok or we need any modification when review assignment type set to `Open`.

NateWr:
@touhidurabir hmm, that's not what I'm seeing. Here are my reproduction steps.

1. Send a submission to Review.
2. Make sure there is an author assigned to that submission (In my case, the author has only the author role.)
3. Add an Anonymous review assignment.
4. Add an Open review assignment.
5. Login as the author.
6. Go to the author's submission view.
7. Go to the Review Stage.
8. Click Add Discussion.
9. See that the Open reviewer is not in the list of participants.

Here are screenshots showing my configuration. The submission from the editorial view:

Screenshot 2023-01-19 at 10-22-15 admin test Journal of Public Knowledge

The participants list when creating a discussion as the author:
Screenshot 2023-01-19 at 10-22-30 admin test Journal of Public Knowledge

touhidurabir:
> Here are my reproduction steps.

> 1. Send a submission to Review.
> 2. Make sure there is an author assigned to that submission (In my case, the author has only the author role.)
> 3. Add an Anonymous review assignment.
> 4. Add an Open review assignment.
> 5. Login as the author.
> 6. Go to the author's submission view.
> 7. Go to the Review Stage.
> 8. Click Add Discussion.
> 9. See that the Open reviewer is not in the list of participants.

@NateWr I was able to reproduce your case only when review still haven's accepted the review request . Once the reviewer accepted it , it shows in the list . In my previous screenshot, that was the same case but as I took that screenshot before reloading the page, so it does not show `Request Accepted` . Is this expected behaviour ?

PR #8464, opened 2022-12-02
pkp/pkp-lib#8424 Create map of mailables instead of a dir scanner

Vitaliy-1:


controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -38,7 +37,7 @@
 
 use PKP\core\Core;
 
-abstract class PKPStageParticipantNotifyForm extends Form
+class PKPStageParticipantNotifyForm extends Form
```
Vitaliy-1, line 40: This class is initialised in several places after the form was simplified and children classes removed.

Issue #8508, opened 2022-12-16
Convert current usage stats log file into new format at upgrade


bozana:
Convert the current usage stats log file into the new format at the upgrade, so when the logging for that day continues that the file can be fully processed (all usage considered).

PRs:

- https://github.com/pkp/pkp-lib/pull/8509
- https://github.com/pkp/ojs/pull/3674
- https://github.com/pkp/omp/pull/1300
- https://github.com/pkp/ops/pull/434

bozana:
@asmecher, one more for you to review :-) But, this assumes the code from here: https://github.com/pkp/pkp-lib/issues/6849. Thus, first that other issue, then this one :-)
Thanks a lot!

asmecher:
Thanks, @bozana, and sorry for the wait! Just one comment.

PR #8509, opened 2022-12-16
pkp/pkp-lib#8508 convert current usage stats log file at upgrade

bozana:
s. https://github.com/pkp/pkp-lib/issues/8508

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php
+++ b/classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php
+ *
+ * Copyright (c) 2022 Simon Fraser University
+ * Copyright (c) 2022 John Willinsky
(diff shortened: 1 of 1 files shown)


classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Reverse the downgrades
+     *
+     * @throws DowngradeNotSupportedException
+     */
+    public function down(): void
+    {
+        throw new DowngradeNotSupportedException();
```
asmecher, line 60: Rather than having this throw an exception, I would suggest just leaving this function empty.

bozana: @asmecher, I see many of the other upgrade scripts do it in this way. Shall I just remove it here, or also from the others -- at least my other upgrade scripts (for statistics upgrade)?

asmecher: Hmm, you're right, I hadn't realized this was so widespread. Please continue this way for now, and we'll talk about potentially better approaches for the next release.

PR #8512, opened 2022-12-19
pkp/pkp-lib#2135 Password reset process update

touhidurabir:
This PR aims to solve https://github.com/pkp/pkp-lib/issues/2135


classes/user/form/ResetPasswordForm.php
```diff
@@ -0,0 +1,154 @@
...
+		$this->readUserVars(array('username', 'hash', 'password', 'password2'));
+	}
+
+	/**
+	 * @copydoc Form::execute()
+	 */
+	public function execute(...$functionArgs) {
+        
+		$user = $this->getUser();
+        $user->setPassword(Validation::encryptCredentials($user->getUsername(), $this->getData('password')));
```
asmecher, line 111: Hmm, I see a mix of tab- and space-based indents here, but the automatic code formatting tools on `main` should take care of formatting these consistently. Do you see those run when you do a `git commit`?

asmecher: This still needs a look, I think!

touhidurabir: @asmecher this looks ok to me , see the diff at https://github.com/pkp/pkp-lib/pull/8512/files#diff-4550ee52519c4fe1edb10a6f377f4f7e92b0a728b53ef7119890c02af5d38b6d .

asmecher: It's consistent, but it uses tabs, which aren't appropriate for `main`. The `stable-3_3_0` branch (and previous) use tabs for indentation, without automatic formatting; `main` uses spaces, and should have automatic formatting applied. It looks like that's not working on your machine.

The automatic formatting works like this:
- A tool configures Git pre-commit hooks. In pkp-lib it's [Captain Hook](https://github.com/captainhookphp/captainhook) (pkp-lib), which is a Composer dependency; in the application repositories it's [Husky](https://typicode.github.io/husky/#/), which is a NPM dependency.
- For repositories using Husky, the pre-commit hook is configured in `package.json`:
    ```
    "husky": {
      "hooks": {
        "pre-commit": "lint-staged"
      }
    },
    "lint-staged": {
      "**/*.php": [
        "php lib/pkp/lib/vendor/bin/php-cs-fixer fix --config .php-cs-fixer.php --allow-risky=yes",
        "php lib/pkp/lib/vendor/bin/php-cs-fixer fix --config .php-cs-fixer.php --allow-risky=yes --dry-run"
      ]
    }
    ```
- For repositories using Captain Hook, the pre-commit hook is configured in `captainhook.json`:
    ```
    "pre-commit": {
        "enabled": true,
        "actions": [
            {
                "action": "\\CaptainHook\\App\\Hook\\PHP\\Action\\Linting",
                "options": [],
                "conditions": []
            },
            {
                "action": "./tools/pre-commit.sh",
                "options": [],
                "conditions": []
            }
        ]
    },

    ```
- Both of these use `php-cs-fixer` to apply code formatting to changed files, according to .php-cs-fixer.php and .php_cs_rules in the repo root.

You should see these run before any `git commit` is allowed to proceed.

touhidurabir: You are right, this is not running at least for the `pkp/lib` commit in my machine . I will check to see what causing the issue . I think it used to run but only when I make a commit from the separate checkout of `pkp-lib`, not from within any app (like `pkp/lib` folder reside within the checkout of OJS/OMP/OPS ) which is weird .

touhidurabir: My assumption is correct, those checks are not running when making commit from within an app checkout. Made a commit from separate checkout of `pkp-lib` and it ran. should be ok now but please check . In mean time I will fix it for the apps .

Issue #8524, opened 2022-12-27
Cypress tests login failing for Nginx or PHP built in web server


touhidurabir:
**Describe the bug**
With the merge of PRs for https://github.com/pkp/pkp-lib/issues/8239, the cypress version updated to `10.11.0`, user login is failing at tests run time when app is running via Nginx or PHP built in web server. 

**To Reproduce**
1. Make sure any app (`OJS`, `OPS` or `OMP`) running through Nginx or PHP built in web server(`php -S`). 
2. Set the proper env and clean the DB. 
3. Start tests from CLI or UI . 
4. See that after the first test(`10-Installation.cy.js `) complete, it halted at `20-CreateContext.cy.js`. 

![Data suite tests -- Creates a context (failed)](https://user-images.githubusercontent.com/2587979/209653500-b4fb75f4-63df-44d1-9246-dc5f71a10968.png)

Above screenshot points that a login attempt initiated and completed but then it was not redirected to intended url as cypress internal cookie issue . Further investigation points that a successful login is happening at https://github.com/touhidurabir/pkp-lib/blob/main/cypress/support/commands.js#L106-L113 but then cypress is setting a new cookie with different session id for which no `user_id` is associated with in `users` table in DB. As s result, it's failed to redirect to or so it seems. 

**What application are you using?**
 OJS, OMP or OPS version `3.4(pre-release)`

**Additional information**
@jonasraoni has confirmed this behaviour when running apps via PHP built in serve(`php -S`) and confirmed that using `Apache` it's ok . . @touhidurabir confirming this same behaviour running apps via `Nginx` . 

**Proposed Solution/Workaround**
it is tested that login via `login form` rather than `AJAX POST` request does not reset the generated cookie after a successful login . So having a `loginViaForm` command with the following implementation 
```JS
Cypress.Commands.add('loginViaForm', (username, password, context) => {
	context = context || 'index';
	password = password || (username + username);
	
	cy.visit('index.php/' + context + '/login');

	cy.get('input[name=username]').clear().type(username);
	cy.get('input[name=password]').clear().type(password);

	cy.get('button').contains('Login').click();
});
```
and controlling by a `env` variable named `loginViaForm` which will be set to `false` by default in the `cypress.config.js`. Then setting it `true` by command `export CYPRESS_loginViaForm=true` from terminal before running tests and having modified `cy.login()` as such 
```JS
Cypress.Commands.add('login', (username, password, context) => {
[...]

PR #8525, opened 2022-12-27
pkp/pkp-lib#8524 Cypress new command added to login via form

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8524


touhidurabir:
Resolved by https://github.com/pkp/pkp-lib/issues/8544 .

PR #8487, opened 2022-12-13
pkp#6340 Set secure flag to true

salmanm2003:
Hi @asmecher,
I have set the secure flag to `true` regarding your discussion with @fradeve.

Issue #7373, opened 2021-10-07
Set httpOnly flag on session cookie


asmecher:
The session cookie should be sent with the httpOnly flag in order to reduce the risk of session highjacking via a client-side script injection.

Original report: https://huntr.dev/bounties/bbe8d92d-0c70-4dca-b076-350bcb4df403/

asmecher:
Also back-ported to `stable-3_3_0`.

fradeve:
@asmecher one of our customers has found a little issue with the fix for this issue.

The problem is that the `HTTPOnly` flag is set on the cookie only _after_ login, but not before.
Vulnerability scans are highlighting this as an issue.

I have checked this personally and can confirm the problem on `OJS 3.3.0-13`, even when `force_ssl = On`.
Is there any chance that we could set the `HTTPOnly` flag on the cookie _before login_ as well?

I will also flag this to @salmanm2003 so that we address this on the (similar) issue he is working on, #6340.

PR #8533, opened 2023-01-08
Feature/main/6340 improve session management

jonasraoni:


classes/session/SessionManager.php
```diff
@@ -39,13 +39,16 @@ public function __construct(SessionDAO $sessionDao, PKPRequest $request)
...
-        ini_set('session.use_trans_sid', 0);
-        ini_set('session.serialize_handler', 'php');
-        ini_set('session.use_cookies', 1);
-        ini_set('session.name', Config::getVar('general', 'session_cookie_name')); // Cookie name
+        ini_set('session.name', Config::getVar('general', 'session_cookie_name'));
         ini_set('session.cookie_lifetime', 0);
         ini_set('session.cookie_path', Config::getVar('general', 'session_cookie_path', $request->getBasePath() . '/'));
         ini_set('session.cookie_domain', $request->getServerHost(null, false));
+        ini_set('session.cookie_httponly', 1);
+        ini_set('session.cookie_samesite', Config::getVar('general', 'same_site', 'lax'));
```
jonasraoni, line 47: Browsers assume `Lax` is the default value when no `SameSite` is defined.

classes/session/SessionManager.php
```diff
@@ -267,12 +258,11 @@ public function regenerateSessionId(): bool
         $success = false;
         $currentSessionId = session_id();
 
-        if (session_regenerate_id() && isset($this->userSession)) {
+        if (session_regenerate_id(true) && isset($this->userSession)) {
             // Delete old session and insert new session
             $this->sessionDao->deleteById($currentSessionId);
             $this->userSession->setId(session_id());
             $this->sessionDao->insertObject($this->userSession);
-            $this->updateSessionCookie(); // TODO: this might not be needed on >= 4.3.3
```
jonasraoni, line 275: The `session_regenerate_id()` above issues a `Set-Cookie`, so I guess this isn't needed.

classes/session/SessionManager.php
```diff
@@ -234,25 +237,13 @@ public function gc(int $lifetime): bool
...
     {
-        $domain = ini_get('session.cookie_domain');
-        // Specific domains must contain at least one '.' (e.g. Chrome)
-        if (strpos($domain, '.') === false) {
-            $domain = false;
-        }
-
-        // Clear cookies with no domain #8921
-        if ($domain) {
-            setcookie(session_name(), '', 0, ini_get('session.cookie_path'), false);
```
jonasraoni, line 245: This `Set-Cookie` will be overridden by the `Set-Cookie` that follows below (cookie names are unique at the browser, so isn't not needed to clear "old cookies").
This is also responsible for the unexpected issues at Cypress.

classes/session/SessionManager.php
```diff
@@ -234,25 +237,13 @@ public function gc(int $lifetime): bool
      */
     public function updateSessionCookie($sessionId = false, int $expireTime = 0): bool
     {
-        $domain = ini_get('session.cookie_domain');
-        // Specific domains must contain at least one '.' (e.g. Chrome)
-        if (strpos($domain, '.') === false) {
```
jonasraoni, line 239: I think this validation isn't needed (it's possible to setup the domain with `localhost`)

classes/session/SessionManager.php
```diff
@@ -39,13 +39,16 @@ public function __construct(SessionDAO $sessionDao, PKPRequest $request)
...
 
         // Configure PHP session parameters
-        ini_set('session.use_trans_sid', 0);
-        ini_set('session.serialize_handler', 'php');
-        ini_set('session.use_cookies', 1);
-        ini_set('session.name', Config::getVar('general', 'session_cookie_name')); // Cookie name
+        ini_set('session.name', Config::getVar('general', 'session_cookie_name'));
         ini_set('session.cookie_lifetime', 0);
         ini_set('session.cookie_path', Config::getVar('general', 'session_cookie_path', $request->getBasePath() . '/'));
         ini_set('session.cookie_domain', $request->getServerHost(null, false));
```
jonasraoni, line 45: Some lines below, there's some code updating the `session.cookie_domain`, I think it makes sense to bring it before the `session_start()` 🤔

classes/session/SessionManager.php
asmecher: I think `lax` should be `Lax`; also in the [OJS PR](https://github.com/pkp/ojs/pull/3696/files#diff-0e9ca26524adfdf100efce10bf37876578f7c505644f2886007be1eccb969e3bR51).

classes/session/SessionManager.php
asmecher: I'd suggest reviewing the issue associated with the code. This is back in Bugzilla days:

http://wayback.archive-it.org/7100/20180509142027/https://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8843

classes/session/SessionManager.php
asmecher: ...and 8921 is here: http://wayback.archive-it.org/7100/20180509190857/https://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8921

classes/session/SessionManager.php
```diff
@@ -39,13 +39,16 @@ public function __construct(SessionDAO $sessionDao, PKPRequest $request)
...
+        ini_set('session.cookie_httponly', 1);
+        ini_set('session.cookie_samesite', Config::getVar('general', 'same_site', 'lax'));
+        ini_set('session.cookie_secure', Config::getVar('security', 'force_ssl'));
```
asmecher, line 48: This will clobber the proposal at https://github.com/pkp/pkp-lib/pull/8487. Can you work with @salmanm2003 to ensure that the two are reconciled?

salmanm2003: @asmecher I have already added these two lines for the next review as it will make sure that the cookies will have secure flag and http_only enabled before and after login as being mentioned at #7373 by @fradeve :
```
ini_set('session.cookie_httponly', 1);
ini_set('session.cookie_secure', Config::getVar('security', 'force_ssl'));

Comment (jonasraoni) [reply classes/session/SessionManager.php:48]: Yep! For now the #8487 is updating only the line that I dropped, but as @salmanm2003 mentioned he will add the same code, then things will probably end looking the same...

Well, I'm not going to update anything more related to the "secure/httponly" flags, so I'll leave it this way for now, then I might incorporate his PR or vice-versa :)

File: classes/session/SessionManager.php (hunk 1 above)
Comment (jonasraoni) [reply classes/session/SessionManager.php:47]: AFAICR the value is case-insensitive (I've also checked on Firefox/Chrome), but better to avoid surprises 😂

File: classes/session/SessionManager.php (hunk 2 above)
Comment (jonasraoni) [reply classes/session/SessionManager.php:275]: On my test the `session_regenerate_id()` was issuing a `Set-Cookie` under the hood, but I'll double check.

File: classes/session/SessionManager.php (hunk 5 above)
Comment (jonasraoni) [reply classes/session/SessionManager.php:45]: @asmecher I tried to find this issue in GitHub and failed, thanks for the link!

The discussion makes sense, but the problem still exists (Chrome + OJS 3.3.0):
1. Accessed `journal.xyz`
   > Got a cookie with the domain `.journal.xyz`.
2. Accessed `www.journal.xyz`
   > The previous cookie is visible at the subdomain, and no extra cookie was created.
3. Tried to login and got redirected back to the login form
   > A new cookie with the domain `.www.journal.xyz` was created, but the previous cookie wasn't removed, because:
    - The code above only drops domainless cookies
    - The previous cookie failed to be re-used, because this line doesn't work (it's not possible to set these session constants when there's an active session): https://github.com/pkp/pkp-lib/blob/ef79d952408a9c91fa0b38cfbdb407df69429704/classes/session/SessionManager.php#L78

FYI, the login form generated these `Set-Cookie` headers at the response:
```http
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; path=/; domain=www.journal.xyz
Set-Cookie: OJSSID=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; path=/; domain=www.journal.xyz; HttpOnly
Set-Cookie: OJSSID=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; expires=Thu, 09-Feb-2023 19:43:34 GMT; Max-Age=2592000; path=/; domain=www.journal.xyz; HttpOnly
```

I'll check the other bugzilla link and update the PR.

classes/session/SessionManager.php
jonasraoni: @asmecher I tried to find this issue in GitHub and failed, thanks for the link!

The discussion makes sense, but the problem still exists (Chrome + OJS 3.3.0):
1. Accessed `journal.xyz`
   > Got a cookie with the domain `.journal.xyz`.
2. Accessed `www.journal.xyz`
   > The previous cookie is visible at the subdomain, and no extra cookie was created.
3. Tried to login and got redirected back to the login form
   > A new cookie with the domain `.www.journal.xyz` was created, but the previous cookie wasn't removed, because:
    - The code above only drops domainless cookies
    - The previous cookie failed to be re-used, because this line doesn't work (it's not possible to set these session constants when there's an active session): https://github.com/pkp/pkp-lib/blob/ef79d952408a9c91fa0b38cfbdb407df69429704/classes/session/SessionManager.php#L78

FYI, the login form generated these `Set-Cookie` headers at the response:
```http
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; path=/; domain=www.journal.xyz
Set-Cookie: OJSSID=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; path=/; domain=www.journal.xyz; HttpOnly
Set-Cookie: OJSSID=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/
Set-Cookie: OJSSID=mlamfo2l6deevkhnjmflsmk1tk; expires=Thu, 09-Feb-2023 19:43:34 GMT; Max-Age=2592000; path=/; domain=www.journal.xyz; HttpOnly
```

I'll check the other bugzilla link and update the PR.

Issue #8306, opened 2022-10-03
Enhance the queue failed jobs functionality. 


touhidurabir:
**Describe the enhancement**
We have a frontend UI implemented to view queue pending/waiting to execute queue jobs in pkp/pkp-lib#7171 also queue job execution/handling enhancement in pkp/pkp-lib#7105 . But have no UI or CLI functionality to view or manage failed jobs and associated error for those failed job. It can be a good enhancement to have following functionalities for better managing of failed jobs 

- UI for failed jobs with redispatch, remove and exception trace visualisation.
- CLI enhancement for listing failed jobs with ability to redispatch and clear out failed jobs. 
- Schedule task to clear out much older failed jobs such as older than 180 days . 

**Who is asking for this one?**
No one right now.  But this will give both the devs and site admins each access to failed jobs with better visualisation. 

**Possible UI Update(screenshots)**
<img width="991" alt="CleanShot 2023-01-11 at 14 54 38@2x" src="https://user-images.githubusercontent.com/2587979/211762460-e60403de-05cc-4bb2-a088-bb8eb068ba82.png">
<img width="1005" alt="CleanShot 2023-01-11 at 14 52 01@2x" src="https://user-images.githubusercontent.com/2587979/211762507-19953341-a3dc-4569-9a31-0d63c467357e.png">
![omp-main test_index php_index%2Fadmin%2FfailedJobDetails_73](https://user-images.githubusercontent.com/2587979/211762598-3d6731c2-4f17-4b45-a7a6-63be267300e7.png)

**Previous Idea**
Clear out much older failed jobs from the `failed_jobs` table. Suggestion is to run a schedule task once a week and clear out any failed jobs older that 2 weeks from the table to keep it cleaner . This can also include some enhancement to existing job cli tool for better failed jobs handling.

### PRs
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3763
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1298
[ops](https://github.com/touhidurabir/ops) --> https://github.com/pkp/ops/pull/472
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8534
[ui-library](https://github.com/pkp/ui-library) --> https://github.com/pkp/ui-library/pull/248

NateWr:
[...]

PR #8534, opened 2023-01-09
pkp/pkp-lib#8306 Failed jobs UI and CLI enhancement

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8306


api/v1/jobs/PKPJobHandler.php
```diff
@@ -0,0 +1,123 @@
...
+class PKPJobHandler extends APIHandler
+{
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        $this->_handlerPath = 'jobs';
+
+        $this->_pathPattern = '/admin/api/{version}/' . $this->_handlerPath;
```
asmecher, line 38: I think the URL pattern should be `/api/{version}/jobs`:
- Version number comes first (for consistency with other endpoints)
- `jobs` reflects the entity we're working with, rather than the person/role who is working with it

classes/job/repositories/FailedJob.php
```diff
@@ -0,0 +1,94 @@
...
+use PKP\job\resources\HttpFailedJobResource;
+
+class FailedJob extends BaseRepository
+{
+    public function __construct(PKPFailedJobModel $model)
+    {
+        $this->model = $model;
+    }
+
+    public function showJobs(): LengthAwarePaginator
```
asmecher, line 37: This should be moved out of the model (`classes/job`) and into the view.

classes/task/RemoveFailedJobs.php
```diff
@@ -0,0 +1,50 @@
...
+        return __('admin.scheduledTask.removeFailedJobs');
+    }
+
+
+    /**
+     * @copydoc ScheduledTask::executeActions()
+     */
+    public function executeActions()
+    {
+        $cleanUpPeriod = (int) Config::getVar('queues', 'failed_job_clean_period', null);
```
asmecher, line 39: The last `null` (the default value) should probably be the same as the default you're introducing into the `config.TEMPLATE.inc.php` file. That way folks to install a fresh application will get the same default as those who upgrade but forget to add the new `config.inc.php` option.

locale/en_US/admin.po
```diff
@@ -464,7 +467,7 @@ msgid "admin.cli.tool.jobs.maintenance.message"
 msgstr "Can not run queue jobs because the application maintenance mode is enabled."
 
 msgid "admin.cli.tool.jobs.available.options.list.description"
-msgstr "List all queued jobs. If you want to paginate results, use the parameters --page= and --perPage="
+msgstr "List all queued jobs. If you want to paginate results, use the parameters --page= and --perPage= . Pass flah --failed to list failied jobs."
```
asmecher, line 470: *`flah` should be removed (this reminds me of a nonsense word I use while I'm coding -- `blee`)
*`failied` should be `failed`

pages/admin/AdminHandler.php
```diff
@@ -552,4 +568,154 @@ protected function getJobsTableState(int $page = 1): array
...
+                    'label' => __('admin.jobs.list.failedAt'),
+                    'value' => 'failed_at',
+                ],
+                [
+                    'name' => 'actions',
+                    'label' => __('admin.jobs.list.actions'),
+                    'value' => 'action',
+                ],
+            ],
+            'rows' => collect($failedJobs->all())->map(fn($failedJob) => array_merge($failedJob, [
```
asmecher, line 651: It would be nice if the list of jobs was fetched via the API as well, but I'll leave that aspect to Nate to make sure this proposal is consistent with other examples.

classes/task/RemoveFailedJobs.php
touhidurabir: Done .

pages/admin/AdminHandler.php
touhidurabir: Fixed.

locale/en_US/admin.po
touhidurabir: Fixed.

pages/admin/AdminHandler.php
touhidurabir: I also agree with this idea . wondering what is the best approach to make make a request form a page route/handler to api/handler route ?  in most of cases I can see that we make such api request from the view to API end point . But here we are passing the required base data into the initialisation of `VUE` component . There is always way such as we can fetch those in the component initialisation life cycle methods such as `created` or `mounted` but that feels bit unconventional approach to me . Any other idea or may be I am over thinking it .

classes/job/repositories/FailedJob.php
touhidurabir: @asmecher bit confused about the requested change. `showJobs` method resides in the repository not model . Also not sure what you mean by view .

includes/functions.php
```diff
@@ -415,3 +416,21 @@ function convertHrToBytes(string $value): int
     }
 }
 
+/**
+ * Converts a shorthand byte value to an integer byte value.
```
NateWr, line 420: Is this comment correct?

pages/admin/AdminHandler.php
NateWr: Yeah this is basically the way we do it. The fetch is extracted to a repository or some other method, then that's called in the APIHandler and the PageHandler. In this case, I don't think it's very important to pre-populate the data in the page, so you could load the data from the API in the Vue.js component once the page is loaded (in the `mounted()` callback).

pages/admin/AdminHandler.php
```diff
@@ -552,4 +569,154 @@ protected function getJobsTableState(int $page = 1): array
...
+                'detailsPath' => $request->getDispatcher()->url($request, Application::ROUTE_PAGE, 'index', 'admin', 'failedJobDetails', $failedJob['id'])
```
NateWr, line 653: Our standard approach is to include a `_href` attribute which is a URL to the resource in the API. I'd be tempted to put this into your `HTTP****Resource.php` files. That is the equivalent of maps in our schema-based entities.

pages/admin/AdminHandler.php
```diff
@@ -552,4 +569,154 @@ protected function getJobsTableState(int $page = 1): array
...
+            'total' => $total,
+            'lastPage' => $failedJobs->lastPage(),
+            'currentPage' => $failedJobs->currentPage(),
+            'isLoadingItems' => false,
+            'apiUrl' => $request->getDispatcher()->url($request, Application::ROUTE_API, APIHandler::ADMIN_API_PREFIX, 'jobs'),
+        ];
+    }
+
+    /**
+     *
```
NateWr, line 664: Missing description

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,48 @@
...
+                        </pkp-button>
+                        <pkp-button is-warnable @click="this.pkp.eventBus.$emit('remove-job', slotProps.row)">
+                            {translate key="admin.jobs.failed.action.delete"}
+                        </pkp-button>
+                        <pkp-button element="a" is-link :href="slotProps.row.detailsPath">
+                            {translate key="admin.jobs.failed.action.details"}
+                        </pkp-button>
+                    </div>
+                </td>
+            </template>
```
NateWr, line 38: Some indentation problems here.

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,48 @@
...
+			:columns="columns"
+			:rows="rows"
+			:description="description"
+			:label="label"
+            :api-url="apiUrl"
+		>
+            <template slot="actions" slot-scope="slotProps">
+                <td>
+                    <div class="buttonRow">
+                        <pkp-button @click="this.pkp.eventBus.$emit('redispatch-job', slotProps.row)">
```
NateWr, line 27: Is there a reason you used the global `eventBus` instead of the click handlers for the page? I think this could be:

```
<pkp-button @click="redispatch(slotProps.row)">
```

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,48 @@
...
+	</h1>
+	<div class="app__contentPanel">
+		<pkp-table
+			:columns="columns"
+			:rows="rows"
+			:description="description"
+			:label="label"
+            :api-url="apiUrl"
+		>
+            <template slot="actions" slot-scope="slotProps">
```
NateWr, line 24: If `row` is the only property in `slotProps` that you need, we usually use deconstruction to pull it out:

```html
<template slot="actions" slot-scope="{row}">
    ...
    <pkp-button @click="example(row)">
```

NateWr: If there are more props, you can pull them out with:

```
slot-scope="{row, column, etc}"
```

classes/task/RemoveFailedJobs.php
```diff
@@ -0,0 +1,49 @@
...
+        return __('admin.scheduledTask.removeFailedJobs');
+    }
+
+
+    /**
+     * @copydoc ScheduledTask::executeActions()
+     */
+    public function executeActions()
+    {
+        $cleanUpPeriod = (int) Config::getVar('queues', 'failed_job_clean_period', 180);
```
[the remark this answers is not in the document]
asmecher: Yup, the 180 is the default. (Same as introduced to the template in the app PRs.)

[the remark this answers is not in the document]
NateWr: The config file says that if the line is commented out jobs will never be deleted:

```
; The number of days older failed jobs will be removed from failed list automatically
; Comment out this setting if do not want to clean much older failed jobs
failed_job_clean_period = 180
```

But from this line of code it seems that if you commented out the `failed_job_clean_period` they would still be deleted after 180 days.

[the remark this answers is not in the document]
touhidurabir: @NateWr you are right, even if it's commented out , the default(`180`) will be taken in that case. we need to pass the third params `0` or `null` to handle it .

pages/admin/AdminHandler.php
touhidurabir: Done . All data fetch moved to API .

classes/job/traits/JobResource.php
```diff
@@ -17,11 +17,22 @@
...
 use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Model;
 
 trait JobResource
 {
     protected string $dateFormat = 'Y-m-d G:i:s T Z';
 
+    public static function toResourceArray(Model $modelInstance): array
+    {
+        return (new static($modelInstance))->toArray(app()->get('request'));
```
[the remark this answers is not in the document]
touhidurabir: Done .

classes/handler/APIHandler.php
```diff
@@ -190,14 +196,21 @@ public function getApp()
      * Compiles the URI path pattern from the context, api version and the
      * unique string for the this handler.
      *
+     * @param  bool  $contextualApi
```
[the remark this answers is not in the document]
touhidurabir: > Can you explain this parameter in the docblock?

@NateWr  Before that I want to make sure if we really need it . So the idea behind it is able to use `Contextual API` even when we set the following in any of admin level API class
```php
/** @var bool Define if all the path building for admin api */
protected $_apiForAdmin = false;
```
This will allow us to define the api level but at the same time override if encounter a case when an API class may contains both the admin and contextual level api . But right now I can not think of such case, so this is just an enhancement placed based on future possibilities. But is it necessary or Should I remove it ?

classes/job/repositories/BaseRepository.php
```diff
@@ -19,23 +19,16 @@
...
 
-abstract class BaseRepository implements Repository
+abstract class BaseRepository
 {
-    /**
-     * @var Model
-     */
-    protected $model;
+    protected Model $model;
+    protected int $perPage = 10;
```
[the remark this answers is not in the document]
touhidurabir: done .

classes/job/repositories/BaseRepository.php
```diff
@@ -19,23 +19,16 @@
...
+use Illuminate\Pagination\LengthAwarePaginator;
 
-abstract class BaseRepository implements Repository
+abstract class BaseRepository
 {
-    /**
-     * @var Model
-     */
-    protected $model;
+    protected Model $model;
```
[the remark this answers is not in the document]
touhidurabir: Doable but do we need it ? also this will start pointing out error in the base repository class in the IDE if we moved ti `$model` prop out of it .

includes/functions.php
touhidurabir: fixed.

pages/admin/AdminHandler.php
touhidurabir: done .

pages/admin/AdminHandler.php
touhidurabir: added .

templates/admin/failedJobs.tpl
touhidurabir: should be fixed now .

templates/admin/failedJobs.tpl
touhidurabir: Right and I dont remember why I used global `eventbus` . Probably trying out something and forgot to update it .

templates/admin/failedJobs.tpl
touhidurabir: Done . and thanks for the tip .

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,54 @@
...
+		>
+            <template slot-scope="{ row, rowIndex }">
+                <table-cell
+                    v-for="(column, columnIndex) in columns"
+                    :key="column.name"
+                    :column="column"
+                    :row="row"
+                    :tabindex="!rowIndex && !columnIndex ? 0 : -1"
+                >
+                    <div v-if="column.name === 'actions'" class="buttonRow">
```
NateWr, line 32: If you want to use a `ButtonRow`, use the component, not just the class:

```html
<button-row v-if...>
    ...
</button-row>
```

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,54 @@
...
+                </table-cell>
+            </template>
+		</pkp-table>
+
+		<pagination
+			:current-page="currentPage"
+			:last-page="lastPage"
+			:is-loading="isLoadingItems"
+			@set-page="handlePagination"
+		/>
```
NateWr, line 52: Don't show the pagination if there are no items or only one page.

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,54 @@
...
+	<h1 class="app__pageHeading">
+		{translate key=$pageTitle}
+	</h1>
+	<div class="app__contentPanel">
+		<pkp-table
+			:columns="columns"
+			:rows="rows"
+			:description="description"
+			:label="label"
+            :api-url="apiUrl"
```
NateWr, line 22: This `api-url` doesn't need to be passed into the table.

locale/en_US/admin.po
```diff
@@ -536,36 +563,84 @@ msgstr "Did you mean one of the following?"
...
+msgstr "Actions"
+
+msgid "admin.jobs.failedAt"
+msgstr "Created at {$failedAt}"
+
 msgid "admin.jobs.createdAt"
 msgstr "Created at {$createdAt}"
 
+msgid "admin.jobs.failed.action.redispatch"
+msgstr "Redispatch"
```
NateWr, line 636: > Try Again

locale/en_US/admin.po
```diff
@@ -536,36 +563,84 @@ msgstr "Did you mean one of the following?"
...
+
+msgid "admin.jobs.failed.action.delete"
+msgstr "Delete"
```
NateWr, line 639: You can re-use an existing locale key for "delete".

NateWr (changes requested):

templates/admin/failedJobs.tpl
touhidurabir: done .

locale/en_US/admin.po
touhidurabir: done.

locale/en_US/admin.po
touhidurabir: done .

templates/admin/failedJobs.tpl
touhidurabir: removed.

templates/admin/failedJobs.tpl
touhidurabir: done .

templates/admin/failedJobs.tpl
```diff
@@ -0,0 +1,54 @@
...
+
+{block name="page"}
+	<h1 class="app__pageHeading">
+		{translate key=$pageTitle}
+	</h1>
+	<div class="app__contentPanel">
+		<pkp-table
+			:columns="columns"
+			:rows="rows"
+			:description="description"
```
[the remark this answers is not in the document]
touhidurabir: @NateWr for some reason it's now working for me as showing in console error message like key not found . I mentioned about this https://github.com/pkp/ui-library/pull/248#discussion_r1090022152

classes/job/repositories/BaseRepository.php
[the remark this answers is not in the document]
touhidurabir: Problem is redefining throws fatal error in `PHP8.1+` as such 
```
PHP Fatal error:  Type of PKP\job\repositories\FailedJob::$model must be Illuminate\Database\Eloquent\Model (as in class PKP\job\repositories\BaseRepository) in /Users/abir/Sites/code/omp-main/lib/pkp/classes/job/repositories/FailedJob.php on line 30
```

One way to handle it as remove the prop `$model` from `BaseRepository` and implement in both of the main repos . Then IDE will start pointing out error for all the methods in base repos as a lot of use such as `$this->model` . To overcome this , we can have a have an abstract method `getModel(): Model` but then we have to update all the use of `$this->model` to `$this->getModel()` in base repo .

templates/admin/failedJobs.tpl
touhidurabir: Should be fixed now . I configured captainsHook few weeks again and now it's not running again.

classes/handler/APIHandler.php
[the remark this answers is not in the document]
touhidurabir: Make sense , removed .

classes/job/repositories/BaseRepository.php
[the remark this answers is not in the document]
NateWr: Hmm, I think a type hint where the type has the same parent class is supposed to be allowed. Isn't that what's happening with this? https://github.com/pkp/ojs/blob/main/classes/submission/Repository.php#L29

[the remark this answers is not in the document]
touhidurabir: For https://github.com/pkp/ojs/blob/main/classes/submission/Repository.php#L29 , there is not type defined but here in base repos , we are defining it as 
```php
protected Model $model;
```

and if try to define it in Job/FailedJob repos which extends base repo , we can not redefine it as 
```php
protected PKPJobModel $model;
// or
protected PKPFailedJobModel $model;
```
even though `PKPJobModel/PKPFailedJobModel` extends `Model`, it has to be exact same(can not be the extended or inherited one also) or `PHP8.1+` throws fatal error .  I don't remember the exact behaviour for `PHP8.0+`(probably a warning that in future it will throw fatal error) but this is the case also for no type or primitive type when extending in `PHP8.1+`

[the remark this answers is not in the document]
NateWr: Ah right, I didn't notice that we weren't defining a type for the map. That's probably why we didn't define a type for the map. Ok, leave it as-is for now.

api/v1/jobs/PKPJobHandler.php
asmecher: I had dropped the ball on this earlier, but had to come back to it as part of #8706, which will forbid arbitrary values for the context path (like `admin`, which was introduced here). I've adjusted this to just use `index`, which is what we use elsewhere to indicate non-context-specific requests; see https://github.com/pkp/pkp-lib/pull/8717/commits/efaa7421768c3b3fb78c104de4bd10e509958955 (this link may change if I have to force-push).

PR #8538, opened 2023-01-11
pkp/pkp-lib#7523 Can save DOI settings with no suffix selected

ewhanson:


classes/submission/Repository.php
```diff
@@ -26,9 +26,10 @@
 use Illuminate\Support\Enumerable;
 use Illuminate\Support\LazyCollection;
 use PKP\context\Context;
+use PKP\context\PKPSection;
```
bozana, line 29: This class should not exist any more, after the conversion to the new Repository/DAO/Map pattern.
I see also some old paths in OJS... They will probably go away once you rebase again, but be careful if there are any conflicts... :-)

classes/doi/exceptions/DoiException.php
```diff
@@ -0,0 +1,19 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class DoiException
+ *
+ * @brief Base class to catch for all DOI-related exceptions
+ */
+
+namespace PKP\doi\exceptions;
+
+class DoiException extends \Exception
```
bozana, line 17: Hmmm... Why do we need this empty class?

bozana (commented):
just a question :-)

classes/submission/Repository.php
ewhanson: Not sure why that was added, but I've rebased and this use statement was the only reference to it. I've removed it.

classes/doi/exceptions/DoiException.php
ewhanson: The idea was to create a base class for DOI-related exceptions, so they could be caught, but looking again, I can refactor the existing `DoiActionException` to work in all cases so we won't need an empty base class.

PR #8545, opened 2023-01-14
Bugfix/main/8544 fix session issues

jonasraoni:


classes/session/SessionManager.php
```diff
@@ -15,106 +15,51 @@
...
-                ini_set('session.cookie_domain', $domain);
+        // If there's a session assigned to the session ID
+        if ($this->userSession) {
+            // Validates it and refresh
+            if ($this->isValid($this->userSession)) {
+                $this->refresh();
+                return;
             }
+            // When invalid, regenerates the session ID without destroying the failed session (perhaps it belongs to another user)
+            session_regenerate_id();
```
jonasraoni, line 59: Here is the fix for the session fixation issue that I had, after failing the validation, it was destroying the session data at the database.

classes/session/SessionManager.php
```diff
@@ -317,6 +223,171 @@ public static function hasSession(): bool
...
+        ini_set('session.cookie_samesite', Config::getVar('general', 'same_site', 'Lax'));
+        ini_set('session.cookie_secure', Config::getVar('security', 'force_ssl'));
+        ini_set('session.use_trans_sid', 0);
+        ini_set('session.serialize_handler', 'php');
+        ini_set('session.use_cookies', 1);
+        ini_set('session.gc_probability', 1);
+        ini_set('session.gc_maxlifetime', 60 * 60);
+        ini_set('session.cache_limiter', 'none');
+
+        session_set_save_handler($this, true);
```
jonasraoni, line 246: According to the manual, the second parameter has a default value of true, but I decided to ensure the `register_shutdown_function()` registration 👀

classes/session/SessionManager.php
```diff
@@ -317,6 +223,171 @@ public static function hasSession(): bool
...
+
+        $requestDomain = Stringy::create($this->request->getServerHost(includePort: false));
+        /** @var \Illuminate\Support\Collection<Session> */
+        $sessions = $sessionIds
+            ->map(fn (string $sessionId) => $this->sessionDao->getSession($sessionId))
+            // Only sessions with valid domains (empty domains are also accepted)
+            ->filter(fn (?Session $session) => $session && $requestDomain->endsWith($session->getDomain(), false));
+
+        /** @var ?Session */
+        $bestSession = $sessions->reduce(function (?Session $best, Session $current): Session {
```
jonasraoni, line 274: Here's the fix for the multi-session issue, it just tries to find the best possible and will probably stick to it.

PR #8566, opened 2023-01-24
pkp/pkp-lib#7744 Escape email variables appropriately

asmecher:


classes/mail/variables/ContextEmailVariable.php
```diff
@@ -109,7 +108,7 @@ protected function getContextUrl(): string
     protected function getContextSignature(array $values): string
     {
         $signature = Mail::compileParams(
-            (string) $this->context->getData('emailSignature'), 
+            (string) $this->context->getData('emailSignature'),
```
Vitaliy-1, line 111: Shouldn't we also escape email signature with `PKPString::stripUnsafeHtml()`?

Vitaliy-1 (changes requested):

classes/mail/variables/ContextEmailVariable.php
asmecher: You're right -- fixed, thanks!

PR #8572, opened 2023-01-25
pkp/pkp-lib#6432 Retrieve emailTemplates data for supported locales only

Vitaliy-1:


Vitaliy-1:
(work in progress)

classes/emailTemplate/DAO.php
```diff
@@ -153,9 +153,18 @@ public function fromRow(object $row): EmailTemplate
...
+        $supportedLocalesJson = $row->context_id === PKPApplication::CONTEXT_SITE ?
+            DB::table('site')->first()->supported_locales :
+            DB::table($contextDao->settingsTableName)
+                ->where($contextDao->primaryKeyColumn, $row->context_id)
+                ->where('setting_name', 'supportedLocales')
+                ->value('setting_value');
 
         $rows = DB::table($this->defaultTable)
             ->where('email_key', '=', $emailTemplate->getData('key'))
+            ->whereIn('locale', json_decode($supportedLocalesJson))
```
asmecher, line 167: This caused a fatal error for me -- `json_decode` returns a `stdClass`, and `whereIn` needs an array.

asmecher (changes requested):
This works for me, except for the fatal error I noted in the review, thanks! I'll add some general comments on the issue.

classes/emailTemplate/DAO.php
Vitaliy-1: Hmm, that's strange. Is `supportedLocales` an associative array in your case? I'd expect it to be in the format `["en_US","fr_CA"]`. I can enforce it to return an array but I'm wondering if it's a revenge of https://github.com/pkp/pkp-lib/issues/6757

asmecher: You're probably right -- but as a precaution it might be a good idea to set `$associative` to `true` in the call to `json_decode`.

asmecher: ...aaaand I see you've already done that :+1:

PR #8581, opened 2023-01-27
Add delegate for handling notifications about plugin configuration

wopsononock:


classes/notification/managerDelegate/ConfigurationNotificationManager.inc.php
```diff
@@ -0,0 +1,94 @@
...
+use PKP\db\DAORegistry;
+
+class ConfigurationNotificationManager extends NotificationManagerDelegate {
+
+	/**
+	 * Constructor.
+	 * @param $notificationType int NOTIFICATION_TYPE_...
+	 */
+	function __construct($notificationType) {
+		parent::__construct($notificationType);
```
asmecher, line 26: This constructor is dead code and can be removed.

classes/notification/managerDelegate/ConfigurationNotificationManager.inc.php
```diff
@@ -0,0 +1,94 @@
...
+		$notificationSettingsDao = DAORegistry::getDAO('NotificationSettingsDAO');
+		$notificationSettings = $notificationSettingsDao->getNotificationSettings($notification->getId());
+		assert(!is_null($notificationSettings['contents']));
+		switch ($notification->getType()) {
+			case PKPNotification::NOTIFICATION_TYPE_CONFIGURE_USER:
+				return $notificationSettings['contents'];
+			case PKPNotification::NOTIFICATION_TYPE_CONFIGURE_PLUGIN:
+				return $notificationSettings['contents'];
+			default:
+				assert(false);
```
asmecher, line 42: I know you're following existing patterns, but in new code, we're moving more to using `default: throw new \Exception(...)` in place of the old `default: assert(false);` patterns. (Here and below.)

classes/notification/managerDelegate/ConfigurationNotificationManager.inc.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+/**
+ * @file classes/notification/managerDelegate/ConfigurationNotificationManager.inc.php
```
asmecher, line 4: Was this taken from a 3.3.0-x class as a template? This should be renamed to `.php` and moved into a namespace; have a look at the other notification manager classes in the `main` branch as an example.

asmecher: There should be code auto-formatting tools in the codebase that take care of things like indentation; it looks like those aren't set up in your checkout. For pkp-lib, it's a tool called "Captain Hook" that runs PHP CS Fixer on a call to `git commit`. Can you check that toolset to make sure it's working? (It should e.g. convert tabs to spaces, which we've adopted in `main` per the [PSR](https://www.php-fig.org/psr/psr-2/).)

PR #8589, opened 2023-01-31
pkp/pkp-lib#7352 Locale presentation update

touhidurabir:
This PR introduce update/patch for pkp/pkp-lib#7352

### DIFF
--- a/classes/i18n/Locale.php
+++ b/classes/i18n/Locale.php
@@ -376,7 +376,7 @@ class Locale implements LocaleInterface
     {
         $locales ??= $this->getLocales();
 
-        if ($filterByLocales) {
+        if ($filterByLocales !== null) {
             $filterByLocales = array_intersect_key($locales, array_flip($filterByLocales));
         }
 
--- a/classes/i18n/Locale.php
+++ b/classes/i18n/Locale.php
@@ -365,52 +365,36 @@ class Locale implements LocaleInterface
     /**
      * Get the formatted locale display names with country if same language code present multiple times
      * 
-     * @param   array $filterByLocales  Optional list of locales code to filter by the returned formatted names list
-     * @param   array $locales          Optional list of availabel all locales
-     * @param   int   $langLocaleStatus The const value of one of LocaleMetadata:LANGUAGE_LOCALE_*
+     * @param   array $filterByLocales          Optional list of locales code to filter by the returned formatted names list
+     * @param   array $locales                  Optional list of availabel all locales
+     * @param   int   $langLocaleStatus         The const value of one of LocaleMetadata:LANGUAGE_LOCALE_*
+     * @param   bool  $omitLocaleCodeInDisplay  Should leave out the locale code from display. By deault leave out.
      * 
-     * @return  array                   The list of locales with formatted display name
+     * @return  array                           The list of locales with formatted display name
      */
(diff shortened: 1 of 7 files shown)


classes/i18n/Locale.php
```diff
@@ -365,54 +365,38 @@ public function getScripts(?string $locale = null): Scripts
...
      * 
-     * @return  array                   The list of locales with formatted display name
+     * @return  array                           The list of locales with formatted display name
      */
-    public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH): array
+    public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH, bool $omitLocaleCodeInDisplay = true): array
     {
         $locales ??= $this->getLocales();
 
+        if ($filterByLocales) {
```
asmecher, line 379: Since we're using `null` as a special case value signifying "don't filter anything", I'd suggest making this an exact `=== null` check. This will ensure that the behaviour doesn't get accidentally applied to `[]` as well.

classes/i18n/LocaleMetadata.php
```diff
@@ -34,6 +34,21 @@ class LocaleMetadata
 {
     use ExportableTrait;
 
+    /**
+     * The following constants define how the locale information will be presented
+     * 
+     * LANGUAGE_LOCALE_WITHOUT : The locale will be presented in the current selected language
+     * So, if English and French is availabel and user selected locale is French, it will be
```
asmecher, line 41: *available (here and elsewhere)

classes/i18n/Locale.php
```diff
@@ -365,54 +365,38 @@ public function getScripts(?string $locale = null): Scripts
...
+     * @param   array $filterByLocales          Optional list of locales code to filter by the returned formatted names list
+     * @param   array $locales                  Optional list of availabel all locales
+     * @param   int   $langLocaleStatus         The const value of one of LocaleMetadata:LANGUAGE_LOCALE_*
+     * @param   bool  $omitLocaleCodeInDisplay  Should leave out the locale code from display. By deault leave out.
      * 
-     * @return  array                   The list of locales with formatted display name
+     * @return  array                           The list of locales with formatted display name
      */
-    public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH): array
+    public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH, bool $omitLocaleCodeInDisplay = true): array
```
asmecher, line 375: What do you think about homogenizing the locale code's display or non-display into the `$langLocaleStatus` option? It seems like that option could handle the full set of display cases.

asmecher (changes requested):
Just a few comments, @touhidurabir!

classes/i18n/Locale.php
touhidurabir: @asmecher I thought about it but could not come up with a good solution . Right now we have 3 cases defined as `LANGUAGE_LOCALE_*` and each define different presentation. but adding `locale code` to each of it multiply it and make it harder to properly identify each presentation case or so it seems to me .

PR #8593, opened 2023-02-01
pkp/pkp-lib#7486 Review assignment consideration process update

touhidurabir:
This PR aims to solve https://github.com/pkp/pkp-lib/issues/7486


controllers/grid/users/reviewer/form/ThankReviewerForm.php
```diff
@@ -149,7 +149,6 @@ public function execute(...$functionArgs)
         $reviewAssignmentDao = DAORegistry::getDAO('ReviewAssignmentDAO'); /** @var ReviewAssignmentDAO $reviewAssignmentDao */
         $reviewAssignment->setDateAcknowledged(Core::getCurrentDate());
         $reviewAssignment->stampModified();
-        $reviewAssignment->setUnconsidered(ReviewAssignment::REVIEW_ASSIGNMENT_NOT_UNCONSIDERED);
```
touhidurabir, line 152: Removed as don't see why thanking a review needed to update the consideration status .

classes/decision/types/traits/NotifyReviewers.php
```diff
@@ -57,7 +57,6 @@ protected function sendReviewersEmail(DecisionNotifyReviewer|ReviewerUnassign $m
                 if ($reviewAssignment) {
                     $reviewAssignment->setDateAcknowledged(Core::getCurrentDate());
                     $reviewAssignment->stampModified();
-                    $reviewAssignment->setUnconsidered(ReviewAssignment::REVIEW_ASSIGNMENT_NOT_UNCONSIDERED);
```
touhidurabir, line 60: Don't get the point why a notifying a review needed to update the review assignment consideration status . Weird part is this is my code when I worked on https://github.com/pkp/pkp-lib/issues/2890

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -899,9 +869,10 @@ class_alias('\PKP\submission\reviewAssignment\ReviewAssignment', '\ReviewAssignm
         'SUBMISSION_REVIEW_METHOD_ANONYMOUS',
         'SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS',
         'SUBMISSION_REVIEW_METHOD_OPEN',
-        'REVIEW_ASSIGNMENT_NOT_UNCONSIDERED',
+        'REVIEW_ASSIGNMENT_NEW',
```
asmecher, line 872: These are only here for legacy code and we'll remove them at some point. Since old code won't refer to the new constants, I'd suggest removing them from this list.

controllers/grid/users/reviewer/form/ThankReviewerForm.php
asmecher: I think the logic is that if an editor is thanking a reviewer they must've taken a look over the review. So I'd suggest following [this same pattern](https://github.com/pkp/pkp-lib/pull/8593/files#diff-f2ed2863fe82393fce36f4a2aa609a7a89260e8cd77e770a36927e439303720aR716) here.

classes/decision/types/traits/NotifyReviewers.php
asmecher: I think this is the same as https://github.com/pkp/pkp-lib/pull/8593/files#r1095049305.

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -256,23 +255,23 @@ public function setRecommendation($recommendation)
     }
 
     /**
-     * Get unconsidered state.
+     * Get considered state.
      *
      * @return int
      */
-    public function getUnconsidered()
+    public function getConsidered()
```
asmecher, line 262: This (and the associated clean-up of constants) will make this so much clearer, thanks!

classes/decision/types/traits/NotifyReviewers.php
touhidurabir: @asmecher I think this is still wrong  . the trait used in several decision class and only in decision class such as 
1. CancelInternalReviewRound
2. Accept
3. CancelReviewRound
4. Decline
5. RequestRevisions
6. Resubmit
and the purpose of method `sendReviewersEmail` is to sent mail to any reviewer associated with submission at the current review round about the taken decision . So don't see any reason to update the review assignment `considered` status at time of notifying . 

Also you see previous status was 
```php
public const REVIEW_ASSIGNMENT_NOT_UNCONSIDERED = 0; // Has never been unconsidered
```
So basically at the notify time it's setting the review assignment consider status as `0` which make no sense. Right now this status set through different action such as 
```
Just when a new assignment created --> 0 (REVIEW_ASSIGNMENT_NEW)
Read and Confirm First Time               --> 3 (REVIEW_ASSIGNMENT_CONSIDERED)
Revert Decision                                     --> 1 (REVIEW_ASSIGNMENT_UNCONSIDERED)
Read and Confirm Again                       --> 2 (REVIEW_ASSIGNMENT_RECONSIDERED)
```
Also at the end of few decisions , review round and associated review assignment get deleted .

Sorry if I am bike shedding here .

asmecher: Gotcha, and since that code (https://github.com/pkp/pkp-lib/issues/2890) has never been released, we can change it however we like without affecting users. What I meant is that the logic was probably to cause the review to be flagged as considered when the editor had obviously moved past it in the workflow. I'm confident that thanking the reviewer is a good case for marking the review flagged, but I don't think editor decisions are. I'd support removing the change to the considered status in this use case.

controllers/grid/users/reviewer/form/ThankReviewerForm.php
touhidurabir: Added with slight change .

classes/submission/reviewAssignment/ReviewAssignment.php
touhidurabir: Removed .

PR #8620, opened 2023-02-10
pkp/pkp-lib#8611 Installer get installed locales instead of all available

Vitaliy-1:


classes/install/Installer.php
```diff
@@ -169,7 +170,8 @@ public function preInstall()
         }
 
         if (!isset($this->installedLocales)) {
-            $this->installedLocales = array_keys(Locale::getLocales());
+            $siteDao = DAORegistry::getDAO('SiteDAO'); /** @var SiteDAO $siteDao */
+            $this->installedLocales = array_keys($siteDao->getSite()->getInstalledLocales());
```
Vitaliy-1, line 174: I think it should include only installed locales rather than all. As far as I understand, the case when installed locales are not set, is during upgrade, right? If it's not the case, I'd need to modify `Plugin::installEmailTemplates()` to install templates only for installed locales.

Issue #8598, opened 2023-02-02
Merge locale folders to locale code folders


defstat:
For example there are `es_ES` and `es_MX` folders in the installation. Consider a solution/process that would merge them to an `es` locale code folder. 

Changes need to be made to 
1. the filesystem regarding the translation `.po` files, and their respective folders.
2. the DB, regarding the data that directly correspond and manage the installation's locale
3. the base code, where ever certain affected locales are hard coded 

-----------------------------------
**PRs**

**FOR MERGE**
~~OJS: https://github.com/pkp/ojs/pull/3774 (Alec: Additions made here https://github.com/pkp/ojs/pull/3775)~~
~~OMP: https://github.com/pkp/omp/pull/1344~~
~~OPS: https://github.com/pkp/ops/pull/478~~
~~PKP-LIB: https://github.com/pkp/pkp-lib/pull/8649 (Alec: Additions made here https://github.com/pkp/pkp-lib/pull/8650)~~
~~UI-LIBRARY: https://github.com/pkp/ui-library/pull/261~~
~~staticPages: https://github.com/pkp/staticPages/pull/74~~

**ADDITIONAL PRs**
OJS: https://github.com/pkp/ojs/pull/3779 **TESTS ONLY**
~~pkp-lib: https://github.com/pkp/pkp-lib/pull/8654 (merged)~~
pkp-lib: https://github.com/pkp/pkp-lib/pull/8658

**PLUGINS**
plagiarism :  https://github.com/pkp/plagiarism/pull/41
paperbuzz  :  https://github.com/pkp/paperbuzz/pull/27
pln:  https://github.com/pkp/pln/pull/62
pdfJsViewer:  https://github.com/pkp/pdfJsViewer/pull/65
crossref   :  https://github.com/pkp/crossref-ojs/pull/28
oaijats:  https://github.com/pkp/oaiJats/pull/34
makesubmission :  https://github.com/pkp/makeSubmission/pull/35
medra  :  https://github.com/pkp/medra/pull/8
lensGalley :  https://github.com/asmecher/lensGalley/pull/63
jatsTemplate   :  https://github.com/pkp/jatsTemplate/pull/23
JATSParser :  https://github.com/Vitaliy-1/JATSParserPlugin/pull/82
immersion  :  https://github.com/pkp/immersion/pull/84
hypothesis :  https://github.com/asmecher/hypothesis/pull/21
googleScholar  :  https://github.com/pkp/googleScholar/pull/13
[...]

PR #8621, opened 2023-02-10
[pkp-lib][main] #8598 MERGED LOCALE FILES FOR REVIEW

defstat:


xml/schema/common.xml
```diff
@@ -99,7 +99,7 @@
 		</field>
 		<field name="installed_locales" type="C2" size="1024">
 			<NOTNULL/>
-			<DEFAULT VALUE="en_US"/>
```
asmecher, line 102: This file shouldn't be changed; it's historical, and will be used for pre-3.3 upgrades.

classes/migration/upgrade/v3_4_0/MergeLocalesMigration.php
```diff
@@ -0,0 +1,366 @@
...
+
+        // Context
+        $contexts = DB::table($this->CONTEXT_TABLE)
+            ->get();
+
+        foreach ($contexts as $context) {
+            $this->updateSingleValueLocale($context->primary_locale, $this->CONTEXT_TABLE, 'primary_locale', $this->CONTEXT_COLUMN, $context->{$this->CONTEXT_COLUMN});
+        }
+
+        $journalSettingsFormLocales = DB::table($this->CONTEXT_SETTINGS_TABLE)
```
asmecher, line 150: OJS-specific language ("journal") shouldn't be used in pkp-lib; use application-independent "context".

classes/migration/install/CommonMigration.php
```diff
@@ -45,7 +45,7 @@ public function up(): void
             $table->bigInteger('redirect')->default(0)->comment('If not 0, redirect to the specified journal/conference/... site.');
             $table->string('primary_locale', 14)->comment('Primary locale for the site.');
             $table->smallInteger('min_password_length')->default(6);
-            $table->string('installed_locales', 1024)->default('en_US')->comment('Locales for which support has been installed.');
+            $table->string('installed_locales', 1024)->default('en')->comment('Locales for which support has been installed.');
```
asmecher, line 48: You'll also need to change the default for existing schema in the upgrade migration (here for `site` and below for `email_templates_default_data`). At a glance, https://stackoverflow.com/questions/37003469/laravel-migrations-change-default-value-of-column has a solution.

asmecher:
The XML files (e.g. `bic21qualifiers.xml`) that have landed in the `depr_xx_YY_depr` directories should be moved into the new locations. (They're used in OMP and will adapt to the new locale name without trouble.)

PR #8622, opened 2023-02-11
pkp/pkp-lib#7524 Make DOI filters more granular

ewhanson:


classes/components/listPanels/PKPDoiListPanel.php
```diff
@@ -199,6 +210,9 @@ public function getConfig()
             'manager.dois.status.error.description',
             'manager.dois.status.stale.description',
             'manager.dois.registration.manuallyMarkedRegistered',
+            'manager.dois.filters.doiAssigned',
+            'manager.dois.filters.doiAssigned.description',
+            'manager.dois.status.stale.filterTitle',
```
bozana, line 215: Do we need `manager.dois.status.error.filterTitle` here?

bozana: `manager.dois.status.needsDoi` is 2 times in the list

bozana (commented):
just two comments :-)

Issue #8369, opened 2022-10-25
Consider new DOI implementation in other parts of the code


bozana:
Consider new DOI implementation in other parts of the code:
- [x] DublinCoreMetaPlugin (s. https://github.com/pkp/pkp-lib/issues/8406) 
- [x] GoogleScholarPlugin (s. https://github.com/pkp/pkp-lib/issues/8478)
- [x] IssueNativeXmlFilter  
- [x] NativeXmlIssueFilter 
- [x] NativeXmlRepresentationFilter
- [x] NativeXmlSubmissionFileFilter
- [x] CounterReportAR1

PRs:
- https://github.com/pkp/pkp-lib/pull/8624
- ~https://github.com/pkp/orcidProfile/pull/234~ will be solved in another issue
- https://github.com/pkp/ojs/pull/3757
- https://github.com/pkp/omp/pull/1331
- https://github.com/pkp/ops/pull/466 (only submodule update)

bozana:
Hi @ewhanson, could you please take a look at the PRs/changes above? Thanks a lot!

ewhanson:
Hey @bozana, I had a look and only a few comments/questions. 

On the subject of DOI versioning with the native import export tools, I took a look and don't see any immediate issues. The DOIs are tied to versioned pubObjects, e.g. two publications on the same submission having two different DOIs would be updated the same way as if there were only one publication. This should give us the desired behaviour, but it's hard to say what edge cases there might be.

ewhanson:
@bozana, another thought I had about DOI versioning (or rather, lack of DOI versioning) is what happens when say, a submission has two publication versions that are each assigned the same DOI? When the native import works through the second publication, it would try and create the same DOI again. Would it be best to assume that this is correct (that this publication should be assigned the same DOI), and instead of creating a new DOI object, just update the publication's `doi_id` to use the existing one?

bozana:
Hi @ewhanson, yes, I would (as I thought here: https://github.com/pkp/pkp-lib/pull/8624#discussion_r1105862218) first check if the DOI exists and if it exists I would assign that DOIs ID to the object, else I would create a new one...
Thanks a lot!

bozana:
@ewhanson, I have considered it... Could you please take another look, if so OK?

ewhanson:
@bozana, looks good to me!

bozana:
Thanks a lot @ewhanson!

PR #8624, opened 2023-02-12
pkp/pkp-lib#8369 Consider new DOI implementation in other parts of th…

bozana:
…e code

s. https://github.com/pkp/pkp-lib/issues/8369

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlRepresentationFilter.php
+++ b/plugins/importexport/native/filter/NativeXmlRepresentationFilter.php
@@ -16,6 +16,7 @@
 namespace PKP\plugins\importexport\native\filter;
 
 use APP\core\Application;
+use APP\facades\Repo;
 use PKP\plugins\PluginRegistry;
 
 class NativeXmlRepresentationFilter extends NativeImportFilter
@@ -100,6 +101,7 @@ class NativeXmlRepresentationFilter extends NativeImportFilter
     public function parseIdentifier($element, $representation)
     {
         $deployment = $this->getDeployment();
+        $context = $deployment->getContext();
         $advice = $element->getAttribute('advice');
         switch ($element->getAttribute('type')) {
             case 'internal':
@@ -113,9 +115,24 @@ class NativeXmlRepresentationFilter extends NativeImportFilter
                 break;
             default:
                 if ($advice == 'update') {
-                    // Load pub id plugins
-                    $pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $deployment->getContext()->getId());
-                    $representation->setStoredPubId($element->getAttribute('type'), $element->textContent);
+                    if ($element->getAttribute('type') == 'doi') {
+                        if ($doiObject = $representation->getData('doiObject')) {
+                            Repo::doi()->edit($doiObject, ['doi' => $element->textContent]);
+                        } else {
+                            $newDoiObject = Repo::doi()->newDataObject(
+                                [
+                                    'doi' => $element->textContent,
+                                    'contextId' => $context->getId()
+                                ]
+                            );
+                            $doiId = Repo::doi()->add($newDoiObject);
+                            $representation->setData('doiId', $doiId);
+                        }
+                    } else {
+                        // Load pub id plugins
+                        $pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $context->getId());
+                        $representation->setStoredPubId($element->getAttribute('type'), $element->textContent);
+                    }
                 }
         }
     }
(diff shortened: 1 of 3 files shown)


plugins/importexport/native/filter/NativeXmlRepresentationFilter.php
```diff
@@ -113,9 +115,24 @@ public function parseIdentifier($element, $representation)
                 break;
             default:
                 if ($advice == 'update') {
-                    // Load pub id plugins
-                    $pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $deployment->getContext()->getId());
-                    $representation->setStoredPubId($element->getAttribute('type'), $element->textContent);
+                    if ($element->getAttribute('type') == 'doi') {
+                        if ($doiObject = $representation->getData('doiObject')) {
+                            Repo::doi()->edit($doiObject, ['doi' => $element->textContent]);
```
ewhanson, line 120: This might be a larger problem beyond the scope of this PR, but since the DOIs are either being directly edited or created here, if something fails later on, this data will already be written to the database. I don't know if there is a good solution currently for the within the Native XML toolset, but thought I would flag it.

ewhanson: When there already is a `$doiObject` present, I wonder if we should also update the `status` to `Doi::STATUS_STALE` since the metadata is changing and will need updating with whichever registration agency is being used?

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.php
```diff
@@ -389,9 +390,24 @@ public function parseIdentifier($element, $submissionFile)
                 break;
             default:
                 if ($advice == 'update') {
-                    // Load pub id plugins
-                    $pubIdPlugins = PluginRegistry::loadCategory('pubIds', true, $deployment->getContext()->getId());
-                    $submissionFile->setStoredPubId($element->getAttribute('type'), $element->textContent);
+                    if ($element->getAttribute('type') == 'doi') {
```
ewhanson, line 393: Same considerations as above.

plugins/importexport/native/filter/NativeXmlRepresentationFilter.php
bozana: Hmmmm.... Now that I think more about it: actually the object from any`NativeXml...Filter` will never already have a DOI, because this object has just been newly created and imported. The update of objects by using native import is currently not supported, s. e.g. https://github.com/pkp/pkp-lib/blob/main/plugins/importexport/native/filter/NativeXmlRepresentationFilter.php#L106.
Thus, I would suggest to remove that `if ($doiObject = $representation->getData('doiObject'))` completely. OK?

The other thing is to ensure that the DOI from the import XML file does not already exist (which is automatically done within `Repo::doi()->add()`, right?)  i.e. if it exist to use its ID. 
Thus in the code next/below I would first check if such a DOI exists (`filterByIdentifier()`?). 
If it exists the object will get its ID. This could, for example, be the case for all publications of an article. I am not sure if I could also check that the existing DOI is assigned to an object of the same type? :thinking: Else, we could for now just assume that the XML file is correct.
If it does not exist, a new DOI will be created and inserted, as in the code next below.

And, yes, if then a new DOI is created and something goes wrong, so that the object is not updated in the DB, then the DOI would not be linked to anything :-P I think it is similar with other metadata, e.g. keywords, subjects, or even authors... :thinking: ...so that I do not know the solution for that now...

ewhanson: > Thus, I would suggest to remove that if ($doiObject = $representation->getData('doiObject')) completely. OK?

That sounds good to me.

> The other thing is to ensure that the DOI from the import XML file does not already exist (which is automatically done within Repo::doi()->add(), right?)

I took a look and this is not done automatically as part of `Repo::doi()->add()`. It's instead handled as part of `Repo::doi()->validate()`. Here's the code that does the check, though it may be worth it running the `validate` method before saving anyway:

```php
// Check for duplicates across all contexts
$doiId = $object ? $object->getData('id') : null;
$doi = $props['doi'] ?? null;
if ($doi !== null && $this->isDuplicate($doi, $doiId)) {
    $validator->errors()->add('doi', __('doi.editor.doiSuffixCustomIdentifierNotUnique'));
}
```

`filterByIdentifier()` would do the check and is what's used within `isDuplicate()`.

> I think it is similar with other metadata, e.g. keywords, subjects, or even authors... 🤔 ...so that I do not know the solution for that now...

That's what I assumed, but thought I'd check.

Issue #6732, opened 2021-02-08
Upgrade to 3.3: default database engine doesn't support foreign key constraint


Vitaliy-1:
**Describe the bug**
Related forum thread: https://forum.pkp.sfu.ca/t/updating-problem-from-3-2-1-3-to-3-3-0-2/65621/9 
Upgrades from 3.2.1 to 3.3.0.
Users report Upgrade failure with the message:
```
ERROR: Upgrade failed: DB: SQLSTATE[HY000]: General error: 1005 Can’t create table ojs_database.submission_files (errno: 150 “Foreign key constraint is incorrectly formed”) (SQL: alter table submission_files add constraint submission_files_file_id_foreign foreign key (file_id) references files (file_id))
``` 
Although I cannot be sure about the causes, it seems that such an error can occur if the table's storage engine doesn't support foreign key constraints. E.g., `files` table is created using something different from `InnoDB`

**To Reproduce**
I'm unable to reproduce the problem but can emulate by explicitly passing `MyISAM` as a default engine to Laravel's database connector

**Additional information**
According to reports from users, it's relevant for MariaDB, at least versions 10.3.24 and 5.5.60. It seems that forcing to
use `InnoDB` engine by passing additional setting to a database connection (`'engine' => 'InnoDB'`) solves the issue.

asmecher:
@Vitaliy-1, one possible cause for this error is an attempt to form a foreign key constraint when one table uses MyISAM and the other uses InnoDB.

Vitaliy-1:
I'm still unable to reproduce this issue. For those 2 users, who reported it, explicitly stating to use InnoDB engine, solved the issue. Hence, I'm assuming that you are correct and problem is that there are old installs of MySQL/MariaDB that still use other than InnoDB engine as a default. It quite possible that during the upgrade they may not update default, at least I'm experiencing the same with some default parameters on my MySQL instance, e.g., default collation.   

One user reported that `submission_files` table was using `MyISAM` and directly altering the table before upgrade had helped.

semteacher:
Hello. Just faced this issue during my 3.1.1.4 -> 3.3 upgrade preparations.
I realized that my 3.1.1.4 test DB was entirely MyISAM (despite server's default engine - InnoDB). 
During upgrade one table become converted to InnoDB and referred - remains MyISAM (exactly as suggested https://github.com/pkp/pkp-lib/issues/6732#issuecomment-775474500) 
Obviously, update failed. 
[...]

PR #8630, opened 2023-02-14
pkp/pkp-lib#6732 Preflight check of database engine

Vitaliy-1:


classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -550,6 +551,31 @@ public function up(): void
...
+            if (DB::connection() instanceof MySqlConnection) {
+                $defaultEngine = DB::scalar('SELECT ENGINE FROM INFORMATION_SCHEMA.ENGINES WHERE SUPPORT = "DEFAULT"');
+                if ($defaultEngine !== 'InnoDB') {
+                    throw new Exception(
+                        'A default database engine ' . $defaultEngine . ' isn\'t supported, expecting InnoDB. ' .
+                        'Please change the default database engine to InnoDB to run the upgrade.'
+                    );
+                }
+
+                $result = DB::select(
```
asmecher, line 565: Since we've already checked that the default engine is `InnoDB` above, this query doesn't need `information_schema.engines`. It could just be:
```sql
SELECT t.table_name, t.engine AS table_engine
FROM information_schema.tables AS t
WHERE t.table_schema = :databaseName AND t.engine <> 'InnoDB';
```
(I'm a little nervous we're going to run into issues with more restrictive grants out there in the wild, so trying to keep these as simple as possible.)

I'd also suggest making the exception list any tables that need attention, rather than just throwing a generic exception if any tables match. Since the query is providing the information anyway :)

PR #8640, opened 2023-02-15
Fixes for usage information text

temin:
I've been studying the usage of this tool and noticed and fixed some inconsistencies in information text.


asmecher:
Thanks, @temin, good catches! I've also back-ported this to the `stable-3_3_0` branch for the next 3.3.0 release.

Issue #8651, opened 2023-02-17
Allow queue job runner to be enabled by default with increased job processing at request life cycle


touhidurabir:
**Description**
As the part of pkp/pkp-lib#7105, we have a optimised `Job Runner` implemented that allow jobs to run at web request like cycle with following constrains 
1. max 30 jobs
2. max 30 seconds run time
3. max 80% of allowed memory
4. with the determination if next job will be executed with left over or 30 seconds time frame

Though this implemented , it's only allow to run a full capacity when running via `Acron` (which also run jobs at the end of request but) once in a minute which mimics the behaviour of cron job for installation that can not have system cron setup. But for normal web request , we run this with limited capacity with default to disable . the reason for this as follow 

> we decided to have the only one jobs to run at each request life cycle and if thats the path any installation want to go with, they can have the acron which will also run the jobs via Job Runner with `30 seconds+30 MAX Jobs+80% max memory` once a minute . We decided to go that path considering worst case scenario such as if an installation dispatch many jobs (like several hundreds) at some time without running via cron/worker and we allow the Job runner to run  at same config that is `30 seconds+30 MAX Jobs+80% max memory`, that can cause user performance issue suddenly for a small or extended amount of time until large number of pending jobs got cleared.

**Proposal**
Allow the `Job runner` enable by default even if the config is missing in the `config` file . Also increase the capacity to same as what set for `Acron` . 

**Note**
This increased capacity of job runner is a target to assess the performance for `RC1` release to determine if we can keep it up or not . 

#### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8652
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3777 [Submodule Update Only]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1345 [Submodule Update Only]
[ops](https://github.com/touhidurabir/ops) --> https://github.com/pkp/ops/pull/480

touhidurabir:
@asmecher please check the PRs at https://github.com/pkp/pkp-lib/issues/8651#issue-1588815003 . Along with intended changes for this issue it also contains fix for tests after the locale merge as follow 
1. `OJS` --> Test failure fix for `ImportIssue.cy.js` where locale validation pattern regex ned to updated . 
2. `OPS` --> OAI Metadata DC unit test update to handle cross match with output file with expected file . 

[...]

PR #8652, opened 2023-02-17
pkp/pkp-lib#8651 job runner enabled by default with increased job processing at request life cycle

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8651

### DIFF
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -141,14 +141,6 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         $this->registerDatabaseConnector(app()->get(\Illuminate\Queue\QueueManager::class));
 
         if (!Application::get()->isUnderMaintenance() && Config::getVar('queues', 'job_runner', true)) {
-            
-            // This is required to make the upgrade test pass when updaring from 3.2.1 and 3.2.0
-            // As for those version we did not have any queue related functionality 
-            // and jobs related tables are missing
-            if (!Schema::hasTable('jobs')) {
-                return;
-            }
-            
             register_shutdown_function(function () {
                 (new JobRunner())
                     ->withMaxExecutionTimeConstrain()
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -112,6 +112,25 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
      */
     public function boot()
     {
+        if (Config::getVar('queues', 'job_runner', true)) {
+            register_shutdown_function(function() { 
+                
+                // As this runs at the current request's end but the 'register_shutdown_function' registered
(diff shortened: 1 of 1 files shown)


classes/core/PKPQueueProvider.php
```diff
@@ -139,11 +140,21 @@ public function register()
 
         $this->registerDatabaseConnector(app()->get(\Illuminate\Queue\QueueManager::class));
 
-        if (!Application::get()->isUnderMaintenance() && Config::getVar('queues', 'job_runner', false)) {
+        if (!Application::get()->isUnderMaintenance() && Config::getVar('queues', 'job_runner', true)) {
+            
+            // This is required to make the upgrade test pass when updaring from 3.2.1 and 3.2.0
```
asmecher, line 145: *upgrading

classes/core/PKPQueueProvider.php
```diff
@@ -139,11 +140,21 @@ public function register()
...
+            // As for those version we did not have any queue related functionality 
+            // and jobs related tables are missing
+            if (!Schema::hasTable('jobs')) {
```
asmecher, line 148: The job runner should not be running during maintenance due to the `isUnderMaintenance` check above; can you look into why that's not catching the upgrade?

xml/importexport.xsd
```diff
@@ -14,7 +14,7 @@
 	<!-- Identifies a particular I18N locale, e.g. en -->
 	<simpleType name="locale">
 		<restriction base="string">
-			<pattern value="[a-z][a-z]_[A-Z][A-Z](@[a-z]+)?" />
+			<pattern value="([a-z]{2})((_[A-Z]{2})?)(@[a-z]{0,})?" />
```
asmecher, line 17: :+1: I'll cherry-pick this in separately, thanks!

asmecher (changes requested):

xml/importexport.xsd
asmecher: I extended the regex slightly to allow e.g. `ckb` (Kurdish): https://github.com/pkp/pkp-lib/commit/66f096ef37dd5b25195dfafe51257c56a6f22827

classes/core/PKPQueueProvider.php
touhidurabir: The implementation was bit misleading . we need to check the maintenance mode with the `register_shutdown_function` as service provider `register` method get called at the initial app bootstrapping time . Also moved the job runner code to `boot` method as `register` is not the right place for such implementation .

asmecher: Thanks, @touhidurabir; is this ready for review/merging?

touhidurabir: @asmecher Yes, it is . `OMP` test has passed but `OJS/OPS` still running. But should be ready to merge .

PR #8655, opened 2023-02-17
Feature/main/8627 ignore missing files native xml

jonasraoni:


locale/en/manager.po
```diff
@@ -2254,6 +2254,15 @@ msgstr ""
 "The submission file {$id} was skipped because it is attached to a record "
 "that will not be imported, such as a review assignment or discussion."
 
+msgid "plugins.importexport.native.error.submissionFileWithoutRevision"
+msgstr "The submission file {$id} was skipped because it does not have a valid revision."
```
defstat, line 2258: Is this a sanity check, or this change could produce cases that the revision is not valid?

I am wondering whether skipping revised files could result a "break" in the "revision chain" which may result to inconsistences

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
```diff
@@ -148,48 +146,63 @@ public function createSubmissionFileNode($doc, $submissionFile)
...
+            if (!($this->opts['no-embed'] ?? false)) {
                 $embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($localPath)));
                 $embedNode->setAttribute('encoding', 'base64');
                 $revisionNode->appendChild($embedNode);
+                continue;
             }
 
-            $submissionFileNode->appendChild($revisionNode);
+            $hrefNode = $doc->createElementNS($deployment->getNamespace(), 'href');
+            $revisionNode->appendChild($hrefNode);
```
defstat, line 181: It does not change the result, but I would prefer the `$revisionNode->appendChild($hrefNode);` to came after all the attributes and children of `$hrefNode` have been added, if possible - `$revisionNode->appendChild($hrefNode);` being something like the "closure statement" of a process. What do you say?

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
```diff
@@ -148,48 +146,63 @@ public function createSubmissionFileNode($doc, $submissionFile)
...
+            ];
+            $params = $baseParams + ['fileId' => $revision->fileId];
+            $dispatcher ??= Application::get()->getDispatcher();
+            $request ??= Application::get()->getRequest();
+            $url = $dispatcher->url($request, PKPApplication::ROUTE_COMPONENT, $context->getPath(), 'api.file.FileApiHandler', 'downloadFile', null, $params);
+            $hrefNode->setAttribute('src', $url);
         }
 
+        // Report if no revision has been added
+        if (!$hasRevision) {
```
defstat, line 202: I would change this parameter's name to indicate that the `hasRevision` does not refer to whether the SubmissionFile has any revisions, but whether we are adding any revisions to the export (this is the case, right?)

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
```diff
@@ -148,48 +146,63 @@ public function createSubmissionFileNode($doc, $submissionFile)
...
         }
 
         // Create the revision nodes
         $revisions = Repo::submissionFile()->getRevisions(($submissionFile->getId()));
+        $basePath = rtrim(Config::getVar('files', 'files_dir'), '/') . '/';
+        $hasRevision = false;
         foreach ($revisions as $revision) {
-            $localPath = rtrim(Config::getVar('files', 'files_dir'), '/') . '/' . $revision->path;
+            $localPath = $basePath . $revision->path;
+            if (!file_exists($localPath)) {
```
defstat, line 161: So here we are not skipping the `SubmissionFile` if a revision specific file is missing, we are skipping the revision, right?

plugins/importexport/native/filter/SubmissionNativeXmlFilter.php
```diff
@@ -196,7 +197,7 @@ public function addPublications($doc, $submissionNode, $submission)
                 $submissionNode->appendChild($clone);
             } else {
                 $deployment = $this->getDeployment();
-                $deployment->addError(ASSOC_TYPE_SUBMISSION, $submission->getId(), __('plugins.importexport.publication.exportFailed'));
+                $deployment->addError(PKPApplication::ASSOC_TYPE_SUBMISSION, $submission->getId(), __('plugins.importexport.publication.exportFailed'));
```
defstat, line 200: Here, again, maybe we should pick a consistent way to deal with that `$deployment = $this->getDeployment();`. We could do that either `$deployment = $this->getDeployment();` and use the `$deployment` after or by using `$this->getDeployment()->`.

defstat (changes requested):

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
jonasraoni: I also prefer to build the tree sequentially.
I just moved this line up to "return early" without duplicating the `appendChild()`, I'll review how many early returns I added, if it's few, then it's ok to duplicate 👌

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
jonasraoni: Exactly, I'll think about a better name 😅

plugins/importexport/native/filter/SubmissionFileNativeXmlFilter.php
jonasraoni: Right, I'm trying to export as much as possible, so if there's at least one revision, it will be exported. Then we'll just show some warnings to let the user know we did the best we could.

On the forum I saw it's very common to have some missing files, especially from people that migrated from OJS 2.x.
Their journal is healthy and the production files are in place, but they have a bunch of missing files (perhaps orphan data which will be cleaned at OJS 3.4?!).

plugins/importexport/native/filter/SubmissionNativeXmlFilter.php
jonasraoni: Yep, in general I just create variables if I'm going to use them at least twice. The codebase has both patterns 🤔

Issue #8663, opened 2023-02-20
Native XML plugin fails to export submissions with language metadata under PHP 8


jonasraoni:
**Describe the bug**
When trying to export submissions that have at least one language defined in the metadata section, a fatal error is generated.

```php
PHP Fatal error:  Uncaught TypeError: DOMElement::setAttribute(): Argument #2 ($value) must be of type string, array given in lib/pkp/plugins/importexport/native/filter/PKPPublicationNativeXmlFilter.inc.php:89
Stack trace:
#0 lib/pkp/plugins/importexport/native/filter/PKPPublicationNativeXmlFilter.inc.php(89): DOMElement->setAttribute()
#1 plugins/importexport/native/filter/PublicationNativeXmlFilter.inc.php(52): PKPPublicationNativeXmlFilter->createEntityNode()
#2 lib/pkp/plugins/importexport/native/filter/PKPPublicationNativeXmlFilter.inc.php(52): PublicationNativeXmlFilter->createEntityNode()
#3 lib/pkp/classes/filter/Filter.inc.php(449): PKPPublicationNativeXmlFilter->process()
#4 lib/pkp/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php(182): Filter->execute()
#5 lib/pkp/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php(103): SubmissionNativeXmlFilter->addPublications()
#6 plugins/importexport/native/filter/ArticleNativeXmlFilter.inc.php(49): SubmissionNativeXmlFilter->createSubmissionNode()
#7 lib/pkp/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php(65): ArticleNativeXmlFilter->createSubmissionNode()
#8 lib/pkp/classes/filter/Filter.inc.php(449): SubmissionNativeXmlFilter->process()
#9 plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php(203): Filter->execute()
#10 plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php(108): IssueNativeXmlFilter->addArticles()
#11 plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php(57): IssueNativeXmlFilter->createIssueNode()
#12 lib/pkp/classes/filter/Filter.inc.php(449): IssueNativeXmlFilter->process()
#13 plugins/importexport/native/NativeImportExportPlugin.inc.php(289): Filter->execute()
```

[...]

PR #8664, opened 2023-02-20
pkp/pkp-lib#8663 Dropped language attribute from publications at the …

jonasraoni:
…Native XML plugin

### DIFF
--- a/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php
+++ b/plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php
@@ -98,8 +98,6 @@ class SubmissionNativeXmlFilter extends NativeExportFilter {
 		$workflowStageDao = DAORegistry::getDAO('WorkflowStageDAO'); /** @var $workflowStageDao WorkflowStageDAO */
 		$submissionNode->setAttribute('stage', WorkflowStageDAO::getPathFromId($submission->getData('stageId')));
 
-		// FIXME: language attribute (from old DTD). Necessary? Data migration needed?
-
 		$this->addIdentifiers($doc, $submissionNode, $submission);
 		$this->addFiles($doc, $submissionNode, $submission);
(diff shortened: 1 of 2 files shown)


plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php
```diff
@@ -98,8 +98,6 @@ function createSubmissionNode($doc, $submission) {
 		$workflowStageDao = DAORegistry::getDAO('WorkflowStageDAO'); /** @var $workflowStageDao WorkflowStageDAO */
 		$submissionNode->setAttribute('stage', WorkflowStageDAO::getPathFromId($submission->getData('stageId')));
 
-		// FIXME: language attribute (from old DTD). Necessary? Data migration needed?
```
defstat, line 101: Should it be removed from https://github.com/pkp/pkp-lib/blob/stable-3_3_0/plugins/importexport/native/pkp-native.xsd#L268 also?

defstat (commented):
Just a comment on that @jonasraoni, thanks

plugins/importexport/native/filter/SubmissionNativeXmlFilter.inc.php
jonasraoni: For the stable branch I decided to leave this attribute forgotten, just to avoid creating a breaking change on the validation of old exports.
FYI I've already removed this at the `main` branch previously.

defstat: I remember about the main branch.  

Maybe you are right. No harm in leaving it there (the code does not deal with that anyway)

jonasraoni:
Thanks @defstat, so I'll merge it later :)

The other issue is ready for review as well: https://github.com/pkp/pkp-lib/issues/8627, I just created extra PRs to run the tests.

PR #8694, opened 2023-02-24
pkp/pkp-lib#4073 Merge user now updates email_log_users table

jyhein:
Merge user now updates email_log_users table. Item_views table not used anymore?


asmecher:
@jyhein, at a glance, I think this SQL syntax won't work with PostgreSQL. Could you test it and see if there's a formulation that'll work for both MySQL/MariaDB and PostgreSQL?

Issue #8697, opened 2023-02-27
User search at stage participant assignment performing as case sensitive search for PostgreSQL


touhidurabir:
**Describe the bug**
Installation running on PostgreSQL DB, the user search for a submission to assign participant to a stage causing empty result to return because of case sensitive search. 

**To Reproduce**
See details at https://forum.pkp.sfu.ca/t/remove-case-sensitive-search-when-assigning-participants/77033

**What application are you using?**
 OJS 3.3.0-14

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8698
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3798 [TEST ONLY]

touhidurabir:
@asmecher Please review the PRs at https://github.com/pkp/pkp-lib/issues/8697#issue-1600903209 if test all passes .

touhidurabir:
@asmecher I think it's good to merge now .

asmecher:
Merged, thanks, @touhidurabir! I think we already confirmed that this doesn't need forward-porting to `main`, right?

touhidurabir:
> I think we already confirmed that this doesn't need forward-porting to `main`, right?

@asmecher  Yes . I have tested it in main and it works fine there so no forward porting needed . I am closing it as completed .

PR #8698, opened 2023-02-27
pkp/pkp-lib#8697 Fixed user assignment case insensitive search for postgres

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8697

### DIFF
--- a/classes/user/UserStageAssignmentDAO.inc.php
+++ b/classes/user/UserStageAssignmentDAO.inc.php
@@ -120,14 +120,14 @@ class UserStageAssignmentDAO extends UserDAO {
 			WHERE	uug.user_group_id = ? AND
 				s.user_group_id IS NULL'
 				. ($name !== null 
-					? ' AND (LOWER(usgs_pl.setting_value) LIKE CONCAT("%", LOWER(?), "%") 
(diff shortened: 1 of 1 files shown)


classes/user/UserStageAssignmentDAO.inc.php
```diff
@@ -104,7 +104,7 @@ function filterUsersNotAssignedToStageInUserGroup($submissionId, $stageId, $user
 			(int) $userGroupId,
 		];
 		if ($name !== null) {
-			$params = array_merge($params, array_fill(0, 6, '%'.(string) $name.'%'));
+			$params = array_merge($params, array_fill(0, 6, '%'.strtolower((string) $name).'%'));
```
asmecher, line 107: Doing the `strtolower` of the parameter in PHP and then matching with `LOWER(column) LIKE ?` in SQL invites tricky collation problems. I would suggest doing the case conversion 100% on the SQL side by using something like `LOWER(column) LIKE CONCAT('%', ?, '%')`.

asmecher: See this pattern: https://github.com/pkp/ojs/blob/858656d68a24bba35238667b0eda9d528a0e5d8e/classes/issue/Collector.php#L436

PR #8702, opened 2023-02-27
Feature/stable 3 3 0/8700 improve slow query

jonasraoni:


classes/db/DAO.inc.php
```diff
@@ -103,8 +103,15 @@ function retrieveRange($sql, $params = [], $dbResultRange = null, $callHooks = t
 	 * @return int
 	 */
 	public function countRecords($sql, $params = []) {
-		$result = $this->retrieve('SELECT COUNT(*) AS row_count FROM (' . $sql . ') AS count_subquery', $params);
-		return $result->current()->row_count;
+		foreach ([$this->_optimizeCountQuery($sql, $params), [$sql, $params]] as [$sql, $params]) {
```
jonasraoni, line 106: Considering that a `COUNT()` don't need to know anything about costly SELECT fields (`SELECT (SELECT ...) AS slow_sub_query`) nor `ORDER BY`s, I think it's safe to drop them, but MySQL doesn't seem to recognize this optimization opportunity.

So the "parser" is just dropping the SELECT and the ORDER BY clauses, taking care to not touch strings/sub-queries.

Besides skipping the `UNION`, other complex cases I decided to not cover would be:
- `SELECT field AS alias... GROUP BY alias`
- `SELECT function(or_sub_query) GROUP BY function(or_sub_query)`

Optimization failures should fall into the retry mechanism, which will use the original query.

classes/db/DAO.inc.php
```diff
@@ -583,4 +590,75 @@ function formatDateToDB($date, $defaultNumWeeks = null, $acceptPastDate = true)
 			return null;
 		}
 	}
+
+	/**
+	 * Retrieves a SELECT statement without the SELECT and ORDER BY clauses for optimization purposes
+	 * @return array The SQL query at the index 0 and the updated parameters at the index 1
+	 */
+	private static function _optimizeCountQuery(string $sql, array $params): array {
+		$findTopLevelExpression = static function (string $s, string $expression, int $index, ?int &$foundParams = null): int {
```
jonasraoni, line 599: If you think it's ok to keep this code, I can leave some comments to make it more reader friendly. Otherwise, I'll just save somewhere in my gists 😅

classes/security/UserGroupDAO.inc.php
```diff
@@ -910,7 +910,7 @@ function _getSearchSql($searchType, $search, $searchMatch, &$params) {
 			}
 		}
 
-		$searchSql .= $this->userDao->getOrderBy(); // FIXME Add "sort field" parameter?
+		$searchSql .= ' ORDER BY u.user_id'; // FIXME Add "sort field" parameter?
```
jonasraoni, line 913: I was going to update the `getOrderBy()` method, but I didn't check all of its usages, so I just updated here.
This piece improves the performance from 8s to 2s in my use-case.

cypress/support/commands.js
```diff
@@ -394,8 +394,14 @@ Cypress.Commands.add('createUser', user => {
...
 	});
-	cy.get('form[id=userRoleForm] button[id^=submitFormButton]').click();
-	cy.get('span[id$="-username"]:contains("' + Cypress.$.escapeSelector(user.username) + '")');
+	cy.server();
+	cy.route('POST', '**/grid/settings/user/user-grid/update-user-roles*').as('rolesSaved');
+	cy.get('div[id^=component-grid-settings-user-usergrid]').as('userGrid').then(staleUserGrid => {
+		cy.get('form[id=userRoleForm] button[id^=submitFormButton]').click();
+		cy.wait('@rolesSaved').its('status').should('eq', 200);
+		// Wait for the interface to refresh, the "User Grid" will be recreated and get a new ID.
+		cy.get('@userGrid').should('not.have.attr', 'id', staleUserGrid.attr('id'));
```
jonasraoni, line 403: Here it's basically just waiting for the list component to get re-rendered, the purpose is to avoid a `cy.wait()`.
I got curious to search for alternative fixes, probably something using Vue's nextTick()/life-cycle might be good enough for our use-case.

asmecher: Because this is a grid, it should be possible to use the `waitJQuery()` function (see `lib/pkp/cypress/support/commands.js`). But you could also argue as a UX pattern that a newly-created record *should* appear on the page after it's created, i.e. perhaps the grid should jump to the last page or similar.

classes/security/UserGroupDAO.inc.php
asmecher: This will mean that several features will lose their current sorting (by the user's name):
- Subscriber selection in the subscription/payments area (`controllers/grid/users/subscriberSelect/SubscriberSelectGridHandler.inc.php`)
- User grid in settings (`lib/pkp/controllers/grid/settings/user/UserGridHandler.inc.php`)
- User list in user export plugin (`lib/pkp/controllers/grid/users/exportableUsers/ExportableUsersGridHandler.inc.php`)
I suspect this will make those lists less usable. Maybe this could be applied as a stop-gap for better optimization to installations that are encountering performance problems? I'm hesitant to merge it.

classes/db/DAO.inc.php
asmecher: Hmm, this is an awful lot of magic to be adding, especially to a stable release. I did run into [an implementation of this optimization in ADODB](https://github.com/ADOdb/ADOdb/blob/abe86c12b27e428eeee1f933f42f0dcf2c0dd403/adodb-lib.inc.php#L391), but it shows signs of needing [fixes](https://github.com/ADOdb/ADOdb/issues/549) on [fixes](https://github.com/ADOdb/ADOdb/issues/870) even though it's a very mature codebase. (In my opinion this kind of thing is what has effectively killed ADODB; it's become impossible to maintain because of the potential for complicated interactions. Touch one thing and you're guaranteed to unbalance a dozen carefully-balanced other things.)

In the `main` branch, the relevant code has already been rewritten into the Laravel query builder toolset and Collector pattern (e.g. https://github.com/pkp/pkp-lib/blob/main/classes/user/Collector.php and https://github.com/pkp/pkp-lib/blob/main/classes/userGroup/Collector.php), which gives more of an ability to intervene in how SQL is built; I think this optimization to drop ordering could be applied to `main` without needing to build and then parse the entire SQL string. You could just drop the `ORDER BY` part of the query builder and get on with your day. I'd be more amenable to an approach like that.

asmecher: There's a very close parallel to what I'm proposing e.g. [here](https://github.com/pkp/pkp-lib/blob/d8244127bff8078f834f31f23ed29fee11a9a7fa/api/v1/users/PKPUserHandler.php#L142..L153), where the same QueryBuilder is first used to [get a page of results](https://github.com/pkp/pkp-lib/blob/d8244127bff8078f834f31f23ed29fee11a9a7fa/api/v1/users/PKPUserHandler.php#L142) (with a limit and offset), then the limit and offset are reset and the same query builder is [used to get the total count](https://github.com/pkp/pkp-lib/blob/d8244127bff8078f834f31f23ed29fee11a9a7fa/api/v1/users/PKPUserHandler.php#L151) with the limit and offset removed.

But this can only be done on the `main` branch, where the QueryBuilder pattern has been adopted. For `stable-3_3_0`, I'd recommend an intervention on the specific query that's causing the slowdown.

jonasraoni: @asmecher I just skimmed the ADODB implementation, and I think I'm able to break it yet 🙃

I was already expecting a negative answer, so I just wanted to check your opinion before (given the tests passed and there's a retry mechanism). So, I'll go with the less aggressive approach 👌

---

About the Laravel's builder, I did some tests and we can't blindly trust on its output, this simple query:
```php
Capsule::table('submissions')->groupBy('context_id')->count();
```

Outputs (the count return will be whatever comes up in the first record, doesn't fail, but completely wrong 🙈):
```sql
select count(*) as aggregate from `submissions` group by `context_id`
```

I did other tests, and not so complex constructions can easily break the output as well (I just know one near perfect query builder that also does optimizations, but it's not available for PHP).

This way, I think it's worth to review the `count()` usage! I didn't inspect alternatives yet, but looks like it will be safer for us to keep enclosing it with another `SELECT` to avoid "naive" failures (e.g. calling the `$builder->count()` on a random `Builder`).

classes/security/UserGroupDAO.inc.php
jonasraoni: Ok! I just wanted you to balance the pros/cons of doing it, as it's an easy improvement, on interfaces with a search by name, it's less of an issue, but I also prefer to keep this natural sorting 😅
I have a bunch of ideas, I'll check if I can come up with something simple to improve this.

cypress/support/commands.js
jonasraoni: Yeah, I was going to try the `waitJQuery()`, but I thought it was a Vue interface and end up not trying :)

In fact I was mostly trying to find a generic way to solve flaky tests, the `cy.get('selector').should('be.visible')` is supposed to retry the `cy.get()`, but it didn't, even a `cy.get('BODY selector').should('be.visible')` to ensure the element is attached to the DOM failed 😪

classes/db/DAO.inc.php
asmecher: The issue with `GROUP BY` / `count()` is a known Laravel behaviour (see e.g. https://github.com/laravel/framework/issues/44081), and IMO it it's going to be fixed it should be fixed there. If they consider it acceptable behaviour, we should code to Laravel's expectations rather than trying to alter its behaviour. In essence, I'm trying to keep us in our lane (application code) and keep framework concerns with the framework.

jonasraoni: Yep. I just wanted to point out that some review might be needed at our `count()` usage later :)

classes/core/PKPApplication.inc.php
```diff
@@ -224,6 +225,17 @@ public function initializeDatabaseConnection() {
 			error_log("Database query\n$query->sql\n" . json_encode($query->bindings));//\n Bindings: " . print_r($query->bindings, true));
 		});
 
+		Builder::macro('safeCount', function (): int {
```
jonasraoni, line 228: I've decided to add this macro to the Laravel Builder, besides the performance improvement, it's also safer (will work with GROUP BY).

classes/log/EmailLogDAO.inc.php
```diff
@@ -59,19 +59,22 @@ function _getByEventType($assocType, $assocId, $eventType, $userId = null, $rang
 		];
 		if ($userId) $params[] = $userId;
 
-		$result = $this->retrieveRange(
-			$sql = 'SELECT	e.*
-			FROM	email_log e' .
-			($userId ? ' LEFT JOIN email_log_users u ON e.log_id = u.email_log_id' : '') .
-			' WHERE	e.assoc_type = ? AND
+		$baseSql = '
+			FROM email_log e' .
```
jonasraoni, line 63: I did this update basically everywhere: extracted the `select`/`order by` and re-added them at the end.

classes/db/DAO.inc.php
```diff
@@ -98,11 +99,15 @@ function retrieveRange($sql, $params = [], $dbResultRange = null, $callHooks = t
 
 	/**
 	 * Count the number of records in the supplied SQL statement (with optional bind parameters parameters)
-	 * @param $sql string SQL query to be counted
+	 * @param $sql string|Builder SQL query to be counted
 	 * @param $params array Optional SQL query bind parameters
```
asmecher, line 103: Should note that if a Builder is passed for `$sql`, then `$params` will not be used.

classes/core/PKPApplication.inc.php
```diff
@@ -69,6 +69,7 @@
 define('WORKFLOW_TYPE_AUTHOR', 'author');
 
 use Illuminate\Database\Capsule\Manager as Capsule;
+use Illuminate\Database\Query\Builder;
```
asmecher, line 72: I don't think the changes to this class are necessary.

classes/security/UserGroupDAO.inc.php
```diff
@@ -537,12 +544,16 @@ function getUsersById($userGroupId = null, $contextId = null, $searchType = null
 					' . ($userGroupId ? 'AND ug.user_group_id = ?' : '') . '
 			)';
 		}
-		$sql .= ' ' . $this->_getSearchSql($searchType, $search, $searchMatch, $params);
+		$baseSql .= ' ' . $this->_getSearchSql($searchType, $search, $searchMatch, $params, '');
```
asmecher, line 547: Hrm, the difference in behaviour between `null` and `''` is confusing -- but as long as it's cleaner in `main`, I'm OK with it for `stable-3_3_0` and `stable-3_4_0`.

classes/core/PKPApplication.inc.php
jonasraoni: Yep, this is cruft that remained from removed code.

classes/db/DAO.inc.php
jonasraoni: Ok!

classes/security/UserGroupDAO.inc.php
jonasraoni: It's looking strange for me too! As the `_getSearchSQL()` is private and there's just one caller, I'll update it (remove the order by), which was basically transferred to here...

jonasraoni: On the `main` branch, this code doesn't exist anymore.

Issue #8704, opened 2023-02-28
Missing error message for Google reCaptcha


jonasraoni:
**Describe the bug**
When using invalid keys, Google returns `invalid-keys`, which gets converted to `common.captcha.error.invalid-keys`, a locale key which isn't mapped in the software.

As part of the task it makes sense to check if there's any other missing message and perhaps emit a warning for unknown messages (will be needed to try to translate the locale key to `en_US` before assuming it's not covered).

**What application are you using?**
OJS/stable-3_3_0

**Additional information**
![image](https://user-images.githubusercontent.com/361921/221920318-9c4b013d-970f-4658-9cfb-3ecb2d3c30f1.png)

**PRs**
- **[TESTS ONLY]** https://github.com/pkp/ojs/pull/3809
- https://github.com/pkp/pkp-lib/pull/8725
- https://github.com/pkp/pkp-lib/pull/8724

jonasraoni:
@asmecher PRs
- **[TESTS ONLY]** https://github.com/pkp/ojs/pull/3809
- https://github.com/pkp/pkp-lib/pull/8725
- https://github.com/pkp/pkp-lib/pull/8724

jonasraoni:
PRs merged partially.

PR #8724, opened 2023-03-03
pkp/pkp-lib#8704 Broke error messages into user/system and added miss…

jonasraoni:
…ing reCAPTCHA error key

### DIFF
--- a/classes/form/validation/FormValidatorReCaptcha.php
+++ b/classes/form/validation/FormValidatorReCaptcha.php
@@ -15,10 +15,16 @@
 
 namespace PKP\form\validation;
 
+use APP\core\Application;
 use Exception;
+use InvalidArgumentException;
+use PKP\config\Config;
+use PKP\form\Form;
 
 class FormValidatorReCaptcha extends FormValidator
 {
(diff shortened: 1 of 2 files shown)


asmecher:
We don't yet have a pattern for distinguishing between exception messages that are sent to the user and others that are logged by the system. This is one way of doing it, but before introducing a new pattern I'd like to consider some options for doing it consistently. As usual, Laravel's approach makes sense (https://laravel.com/docs/10.x/errors). I'd suggest...
- Just fixing the reported error, per the stable-3_3_0 branch fix
- Starting a discussion for the broader question, with a link to e.g. https://laravel.com/docs/10.x/errors for one option.

jonasraoni:
Okay! I've just applied the underlying fix to `main` through a [direct commit](https://github.com/pkp/pkp-lib/commit/efae92885e7f4afc023a59d88215286cf779f360), so I'll close everything and move to the discussion.

jonasraoni:
Ok, here it is: https://github.com/pkp/pkp-lib/discussions/8762, feel free to edit/add something that I've missed.

PR #8746, opened 2023-03-06
Bugfix/main/8723 fix plugin installation

jonasraoni:


classes/plugins/PluginHelper.php
```diff
@@ -43,51 +44,63 @@ class PluginHelper
...
-            throw new Exception(__('manager.plugins.invalidPluginArchive'));
-        }
 
-        // Create random dirname to avoid symlink attacks.
-        $pluginExtractDir = dirname($filePath) . "/{$pluginShortName}" . substr(md5(mt_rand()), 0, 10);
-        if (!mkdir($pluginExtractDir)) {
-            throw new Exception('Could not create directory ' . $pluginExtractDir);
+        // If the extension doesn't match, copy the file to another location to avoid issues with the PharData class
+        $filePathWithExtension = null;
+        if ($getExtension($filePath) !== $extension) {
```
asmecher, line 66: This is a fatal error, I think?

jonasraoni: Why exactly? FYI these updates are currently deployed at the OJS 3.4 test installation, the basic operations are working fine:
- Installation
- Installation from plugin gallery
- Removal
- Upgrade
- Upgrade from plugin gallery
- Error conditions (e.g. upgrade an already upgraded plugin)

classes/plugins/PluginHelper.php
```diff
@@ -192,74 +192,99 @@ protected function _getConnectionParams()
...
-            $params['additionalLocales'] = $site->getSupportedLocales();
-            $installer = new Upgrade($params, $upgradeFile, true);
-
-            if (!$installer->execute()) {
-                throw new Exception(__('manager.plugins.upgradeFailed', ['errorString' => $installer->getErrorString()]));
+    /**
+     * Attempts to create a locked and writable temporary file
+     * The prefix/suffix will receive a minor sanitization
+     */
+    public static function getTemporaryFile(string $prefix = '', string $suffix = ''): SplFileObject
```
jonasraoni, line 274: I was going to move this method to the `FileManager` class, but I wasn't sure if that class is going to be dropped later.

I found a similar method at the Symfony/Filesystem (which is available indirectly), but it returns a resource, which isn't very interesting 🤔

As a second note, I think the Symfony/Filesystem would be a good base for our `FileManager`, it's boxing the I/O operations (if an error happens at the `copy`, it throws an exception).

classes/plugins/PluginHelper.php
asmecher: Ah, the use of lambda functions looked at a glance like a typo.

We already have `FileManager::getExtension`; I'd suggest making sure that handles compound extensions, and working with it instead.

Otherwise I'd suggest just making a private or protected function in the class rather than using lambdas this way.

jonasraoni:
@asmecher I'd point to this commit: https://github.com/pkp/pkp-lib/pull/8746/commits/f66b8b8755152d36fc9d21416729bbf3b58f8ba0

Exactly these lines (which check if the given archive file has an extension compatible with the "original filename"):
https://github.com/pkp/pkp-lib/pull/8746/commits/f66b8b8755152d36fc9d21416729bbf3b58f8ba0#diff-32e2c802a4e691592052893cdb8dfe98495b5d84fff805327fedd6a682d3a27cR60-R66

classes/plugins/PluginHelper.php
jonasraoni: Yep, I just wanted to check if it's ok to keep the `FileManager` alive, I'll take a look if I can move some things to there and make use of the `Flysystem`.

---

About the `FileManager::getExtension()` it won't work with `.tar.gz` (also its implementation needs to be fixed: `getExtension('test')` returns "test"). The `FileManager::parseFileExtension()` works (it has an exception for this extension), but I decided to have something more generic 🤔

classes/plugins/PluginHelper.php
```diff
@@ -192,74 +173,86 @@ protected function _getConnectionParams()
...
+                $fileManager->rmtree($destinyPath);
+                throw $e;
             }
+        } finally {
+            // Discard the temporary plugin files
+            $fileManager->rmtree($sourcePath);
         }
+    }
 
-        $installedPlugin->setCurrent(0);
```
asmecher, line 259: Though I don't see an update call anywhere here, I would guess this was originally to set the `versions.current` column to `0` for the old version of the plugin. Is this being done elsewhere now that the code has been rewritten?

asmecher (commented):
Just a few minor changes to suggest. This is definitely a riskier changeset than I would've advised this late in the process, given all the potential for platform-specific quirks, but let's make sure it gets a good round of testing during RC2.

classes/file/FileManager.php
```diff
@@ -806,6 +808,28 @@ public function gzCompressFile(string $filePath, int $level = 9): string
...
+     * Attempts to create a locked and writable temporary file
+     * The prefix/suffix will receive a minor sanitization
+     */
+    public static function getTemporaryFile(?string $prefix = null, ?string $suffix = null, ?int $lockType = LOCK_EX, int $retries = 10): SplFileObject
+    {
+        $sanitize = fn (string $path) => preg_replace('/[^\w.-]/', '', $path);
+        $basePath = rtrim(sys_get_temp_dir(), '\\/') . "/";
+        for ($retries = abs($retries); $retries--; ) {
+            try {
+                $file = new SplFileObject($basePath . $sanitize($prefix) . substr(md5(mt_rand()), 0, 10) . $sanitize($suffix), 'x+');
```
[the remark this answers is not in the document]
jonasraoni: As I'm using the `new SplFileObject()` instead of `fopen()`, an exception will be raised =]

classes/file/FileManager.php
```diff
@@ -806,6 +808,28 @@ public function gzCompressFile(string $filePath, int $level = 9): string
...
+                if ($lockType) {
+                    $file->flock(LOCK_EX) || throw new Exception('Failed to acquire lock');
```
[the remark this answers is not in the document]
jonasraoni: Yep, I'll add it and also modify the `$file->flock(LOCK_EX)` to `$file->flock($lockType)`, which I forgot.

classes/plugins/PluginHelper.php
jonasraoni: I suspected this line wasn't needed because there was no `$versionDao->insertVersion($installedPlugin);` after that, then I confirmed here:
https://github.com/pkp/pkp-lib/blob/87e018ebe033574fa40749f0de110c5875f2f56b/classes/site/VersionDAO.php#L116

classes/file/FileManager.php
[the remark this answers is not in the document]
asmecher: Whoops, I clicked through to `fopen` to find the flags and got stuck thinking in C.

Issue #8750, opened 2023-03-07
Deleting journal section causes 500 system error when viewing orphan submission


israelcefrin:
**Describe the bug**
Deleting a section that contains an active submission causes a system error (500) if you attempt to load the submission details after the deletion. The system is not warning about existing submission on the section neither asking to move it to a different one.

**To Reproduce**
1. Go to 'Dashboard > Journal > Sections'
2. Click on the blue arrow beside any section that contains at least one submission and click 'Delete'
3. Go to 'Dashboard > Submissions'
4. Click "View" on any submission from the deleted section
5. See the error

**What application are you using?**
 OJS version 3.4.0rc1
PHP 8.0.18

**Additional information**

[Tab-1678150872979.webm](https://user-images.githubusercontent.com/4992902/223292040-850df175-1444-4f10-9843-0850a708ad9e.webm)

touhidurabir:
@asmecher PRs to review

[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8753
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3817 [TEST ONLY]

asmecher:
Merged -- thanks, @touhidurabir!

PR #8753, opened 2023-03-07
pkp/pkp-lib#8750 Section empty logic update

touhidurabir:
fix for pkp/pkp-lib#8750

### DIFF
--- a/classes/section/Repository.php
+++ b/classes/section/Repository.php
@@ -164,7 +164,7 @@ class Repository
     }
 
     /**
-     * Check i the section has any submissions assigned to it.
+     * Check if the section has any submissions assigned to it.
      */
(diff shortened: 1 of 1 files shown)


classes/section/Repository.php
```diff
@@ -172,7 +172,7 @@ public function isEmpty(int $sectionId, int $contextId): bool
             ->getCollector()
             ->filterByContextIds([$contextId])
             ->filterBySectionIds([$sectionId])
-            ->getCount() > 0;
+            ->getCount() <= 0;
```
asmecher, line 175: Could you update this to `=== 0` for clarity? It should always be numeric, and never <0.

touhidurabir: Done .

PR #8767, opened 2023-03-09
pkp/pkp-lib#8766 Dropped the locale merge for new setting tables, the…

jonasraoni:
… MergeLocalesMigration.php will run before them.


classes/migration/upgrade/v3_4_0/MergeLocalesMigration.php
```diff
@@ -444,12 +444,10 @@ protected static function getSettingsTables(): Collection
             'citation_settings' => 'citation_id',
             'controlled_vocab_entry_settings' => 'controlled_vocab_entry_id',
             'data_object_tombstone_settings' => 'tombstone_id',
-            'doi_settings' => 'doi_id',
             'email_templates_settings' => 'email_id',
             'event_log_settings' => 'log_id',
             'filter_settings' => 'filter_id',
             'genre_settings' => 'genre_id',
-            'institution_settings' => 'institution_id',
```
jonasraoni, line 452: I've just dropped the migration of these two entities, they were introduced at 3.4.0, so by the time they will be created and filled, the locales should be fine.

defstat (approved):

Issue #7670, opened 2022-02-02
Ensure database schemas are fully synchronized after upgrading


jonasraoni:
**Describe the bug**
At the [issue #7167](https://github.com/pkp/pkp-lib/issues/7167) I concluded the migration from the old database framework (< 3.3.0) has changed the way some database artifacts are created (at least on PostgreSQL).
Even though it's not critical, it's better to get it fixed to avoid unexpected issues.

**Solution**
- Grab different database versions (3.1.0, 3.2.0, 3.3.0, the more the better)
- Run the upgrade scripts until /main (3.4)
- Extract a schema-only script from them and compare the outputs for both MySQL and PostgreSQL
- Write a migration script in /main (3.4.0) to synchronize the found differences (e.g. modify constraints, types, default values, nullability, ...).

**What application are you using?**
/stable-3_2_1
/stable-3_3_0

**Additional information**
![image](https://user-images.githubusercontent.com/361921/152169055-67f49701-ca1e-42ae-9cc6-9c362b2e8747.png)

asmecher:
@jonasraoni, starting with OJS/OMP/OPS 3.3.0 the schema creation tools are now built on Laravel migrations, whereas on 3.2.1 and prior they were done with ADODB XMLSchema. This introduces a bunch of semantic changes, e.g. in the way default values are specified, along with a few minor column type changes. When I committed the changes I did what you proposed -- comparing the "before" and "after" schema dumps. So I think it's already OK.

Looking at your example above, if I inspect the PostgreSQL table describe command's output for the `user_settings` table in `stable-3_3_0`, I get:
```
Indexes:
    "user_settings_pkey" UNIQUE CONSTRAINT, btree (user_id, locale, setting_name, assoc_type, assoc_id)
    "user_settings_locale_setting_name_index" btree (setting_name, locale)
    "user_settings_user_id" btree (user_id)
```
With `stable-3_2_1`, I get:
```
Indexes:
    "user_settings_pkey" UNIQUE, btree (user_id, locale, setting_name, assoc_type, assoc_id)
    "user_settings_locale_setting_name_index" btree (setting_name, locale)
    "user_settings_user_id" btree (user_id)
```
[...]

Issue #3573, opened 2018-04-10
Change database tables schema for Clustering


marciofoz:
When running PKP/OJS in Percona XtraDB Cluster (MySql Cluster), all tables are required to have a primary key. Not all of the PKP/OJS tables contain a primary key.

Reference: https://dev.mysql.com/doc/refman/5.7/en/group-replication-requirements.html

"Primary Keys. Every table that is to be replicated by the group must have an explicit primary key defined. Primary keys play an extremely important role in determining which transactions conflict by identifying exactly which rows each transaction has modified."

We see this PKP/OJS tables that do not contain primary keys:

announcement_settings
announcement_type_settings
article_galley_settings
article_html_galley_images
article_search_object_keywords
article_settings
article_supp_file_settings
author_settings
books_for_review_settings
citation_settings
controlled_vocab_entry_settings
custom_issue_orders
custom_section_orders
data_object_tombstone_settings
email_log_users
email_templates_data
email_templates_default_data
event_log_settings
external_feed_settings
filter_settings
group_memberships
group_settings
issue_galley_settings
issue_settings
journal_settings
metadata_description_settings
metrics
notification_settings
oai_resumption_tokens
object_for_review_settings
plugin_settings
processes
referral_settings
review_form_element_settings
review_form_responses
review_form_settings
review_object_metadata_settings
review_object_type_settings
roles
scheduled_tasks
section_editors
section_settings
sessions
site
site_settings
static_page_settings
subscription_type_settings
usage_stats_temporary_records
user_interests
user_settings
versions

asmecher:
@marciofoz, this is a side-effect of the database toolset we use for schema management (ADODB), and our need to support both MySQL and PostgreSQL, which have their own peculiarities around schema management, indexing, etc. If you'll look at those tables you'll generally find a unique index with a name that has a `_pkey` suffix. This takes the place of explicitly defined primary keys in a way that ADODB and both databases can handle.

Our longer term plans are to [replace ADODB with something else](https://github.com/pkp/pkp-lib/issues/2493), but that's going to take a lot of work, and I'm not sure it'll permit us to explicitly define primary keys as you require.

Meanwhile, you can likely cook up some DDL SQL that'll add the explicit primary key definitions where they're needed by looking for the `_pkey`-suffixed indexes.

marciofoz:
@asmecher 
Ok, I understand.
Let me know: changed the tables definitions, should the application address these new integrity constraints?
BR

[...]

PR #8769, opened 2023-03-10
pkp/pkp-lib#8737 Section form multiple role associated user association fix

touhidurabir:
This PR aims to provide solution for or pkp/pkp-lib#8737.

### DIFF
--- a/classes/context/SubEditorsDAO.php
+++ b/classes/context/SubEditorsDAO.php
@@ -21,6 +21,7 @@ use APP\notification\Notification;
 use APP\notification\NotificationManager;
 use APP\submission\Submission;
 use Exception;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Mail;
@@ -296,6 +297,26 @@ class SubEditorsDAO extends \PKP\db\DAO
         }
 
         return $assignments->map(fn ($assignment) => $assignment->userId);
+    }
+
+    /**
+     * Get the section assigned sub eidtor's associated user groups for given section
+     *
+     * @param   int         $contextId
+     * @param   int         sectionId
+     * @param   int|array   $userId
+     *
+     * @return  Collection The user ids and user group id assosciation for given section
+     */
+    public function getSectionAssignEditorsUserGroups(int $contextId, int $sectionId, int|array $userIds): Collection
+    {
+        return DB::table('subeditor_submission_group')
+            ->select(['user_id', 'user_group_id'])
+            ->where('assoc_type', Application::ASSOC_TYPE_SECTION)
+            ->where('context_id', $contextId)
+            ->where('assoc_id', $sectionId)
+            ->whereIn('user_id', Arr::wrap($userIds))
+            ->get();
     }
 }
 
--- a/classes/context/SubEditorsDAO.php
+++ b/classes/context/SubEditorsDAO.php
@@ -300,7 +300,7 @@ class SubEditorsDAO extends \PKP\db\DAO
     }
 
     /**
-     * Get the section assigned sub eidtor's associated user groups ids for given section
+     * Get the section assigned sub editor's associated user groups ids for given section
      *
      * @return  Collection Collection A list of user group IDs for each user, keyed by user ID.
      */
--- a/classes/context/SubEditorsDAO.php
+++ b/classes/context/SubEditorsDAO.php
@@ -300,7 +300,7 @@ class SubEditorsDAO extends \PKP\db\DAO
     }
 
     /**
-     * Get the section assigned sub eidtor's associated user groups for given section
+     * Get the section assigned sub eidtor's associated user groups ids for given section
      *
      * @param   int         $contextId
      * @param   int         sectionId
@@ -308,7 +308,7 @@ class SubEditorsDAO extends \PKP\db\DAO
      *
      * @return  Collection The user ids and user group id assosciation for given section
      */
-    public function getSectionAssignEditorsUserGroups(int $contextId, int $sectionId, int|array $userIds): Collection
+    public function getAssignedUserGroupIds(int $contextId, int $sectionId, int|array $userIds): Collection
     {
         return DB::table('subeditor_submission_group')
             ->select(['user_id', 'user_group_id'])
@@ -316,7 +316,9 @@ class SubEditorsDAO extends \PKP\db\DAO
             ->where('context_id', $contextId)
             ->where('assoc_id', $sectionId)
             ->whereIn('user_id', Arr::wrap($userIds))
-            ->get();
+            ->get()
+            ->groupBy('user_id')
+            ->map(fn($userGroups) => $userGroups->pluck('user_group_id'));
     }
 }
 
--- a/classes/context/SubEditorsDAO.php
+++ b/classes/context/SubEditorsDAO.php
@@ -21,7 +21,6 @@ use APP\notification\Notification;
 use APP\notification\NotificationManager;
 use APP\submission\Submission;
 use Exception;
-use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Mail;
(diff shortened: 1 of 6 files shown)


controllers/grid/settings/sections/form/PKPSectionForm.php
```diff
@@ -138,22 +139,37 @@ public function fetch($request, $template = null, $display = false)
...
+                ->toArray();
+
+            if (!empty($assignedSubeditors)) {
+                $subEditorsDao = DAORegistry::getDAO('SubEditorsDAO'); /** @var SubEditorsDAO $subEditorsDao */
+                $subeditorUserGroups = $subEditorsDao->getSectionAssignEditorsUserGroups(
+                    $contextId, 
+                    $this->getSection()->getId(), 
+                    $assignedSubeditors
+                )
+                ->groupBy('user_id')
```
asmecher, line 162: This kind of extends DAO operations outside of the DAO (and if I'm reading right, divides some of the database-like operations between SQL in the DAO and collections in memory). Since `getSectionAssignEditorsUserGroups` only exists for this use case, I'd suggest moving the `groupBy` and `map` into the `getSectionAssignEditorsUserGroups` function, and giving it a slightly tighter name, something like `getAssignedUserGroupIds`.

controllers/grid/settings/sections/form/PKPSectionForm.php
```diff
@@ -138,22 +139,37 @@ public function fetch($request, $template = null, $display = false)
...
-                    ->assignedToSectionIds([$this->getSectionId()])
-                    ->getIds()
-                    ->toArray(),
-            ]);
-        } else {
-            $this->setData([
-                'assignedSubeditors' => [],
-            ]);
+            $contextId = Application::get()->getRequest()->getContext()->getId();
+            $assignedSubeditors = Repo::user()
```
asmecher, line 147: Is `$assignedSubeditors` made redundant by `$subeditorUserGroups`? My understanding is that `$assignedSubeditors` was only used to check whether the user had an assignment to the section, but that it was missing the potential for one user's assignment to multiple user groups. Now that `$subeditorUserGroups` handles that properly, can we drop `$assignedSubeditors`?

classes/context/SubEditorsDAO.php
```diff
@@ -297,6 +298,26 @@ public function assignEditors(Submission $submission, Context $context): Collect
...
+     */
+    public function getSectionAssignEditorsUserGroups(int $contextId, int $sectionId, int|array $userIds): Collection
+    {
+        return DB::table('subeditor_submission_group')
+            ->select(['user_id', 'user_group_id'])
+            ->where('assoc_type', Application::ASSOC_TYPE_SECTION)
+            ->where('context_id', $contextId)
+            ->where('assoc_id', $sectionId)
+            ->whereIn('user_id', Arr::wrap($userIds))
+            ->get();
```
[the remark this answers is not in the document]
touhidurabir: Not needed any more as I have moved the operation into the DAO class and only retrieving a collection of values only .

controllers/grid/settings/sections/form/PKPSectionForm.php
touhidurabir: Moved to DAO class .

controllers/grid/settings/sections/form/PKPSectionForm.php
touhidurabir: Thats better, done .

classes/context/SubEditorsDAO.php
```diff
@@ -297,6 +298,28 @@ public function assignEditors(Submission $submission, Context $context): Collect
...
     }
+
+    /**
+     * Get the section assigned sub eidtor's associated user groups ids for given section
+     *
+     * @param   int         $contextId
+     * @param   int         sectionId
+     * @param   int|array   $userId
+     *
+     * @return  Collection The user ids and user group id assosciation for given section
```
asmecher, line 309: *association

Suggest a different description:
```
@return Collection A list of user group IDs for each user, keyed by user ID.
```

classes/migration/upgrade/v3_4_0/I8737_SectionEditorsUniqueIndexUpdate.php
```diff
@@ -0,0 +1,75 @@
...
+        });
+    }
+
+    /**
+     * Reverse the migration.
+     */
+    public function down(): void
+    {
+        Schema::table('subeditor_submission_group', function (Blueprint $table) {
+            $this->dropUniqueIndexKey($table);
```
asmecher, line 41: Is there a need to check for the index to exist? I think it should always exist, so Laravel's `dropIndex` should be enough, no?

controllers/grid/settings/sections/form/PKPSectionForm.php
```diff
@@ -138,22 +139,32 @@ public function fetch($request, $template = null, $display = false)
...
+                ->getCollector()
+                ->filterByContextIds([$contextId])
+                ->filterByRoleIds($this->assignableRoles)
+                ->assignedToSectionIds([$this->getSectionId()])
+                ->getIds()
+                ->toArray();
+
+            if (!empty($assignedSubeditors)) {
+                $subEditorsDao = DAORegistry::getDAO('SubEditorsDAO'); /** @var SubEditorsDAO $subEditorsDao */
+                $subeditorUserGroups = $subEditorsDao->getAssignedUserGroupIds(
```
asmecher, line 156: Rather than passing a list of user IDs in here based on the previous query, why not just add the role filter to this function and do it all in one place?

asmecher (changes requested):

classes/migration/upgrade/v3_4_0/I8737_SectionEditorsUniqueIndexUpdate.php
touhidurabir: @asmecher I added those checks as upgrade tests were failing for `3.2.1` with `PostgreSQL` with error as 

`
ERROR: Upgrade failed: DB: SQLSTATE[42704]: Undefined object: 7 ERROR:  constraint "section_editors_pkey" of relation "subeditor_submission_group" does not exist (SQL: ALTER TABLE subeditor_submission_group DROP CONSTRAINT section_editors_pkey)
`

see at https://app.travis-ci.com/github/pkp/ojs/jobs/597974527 . 

But it seems that it does not solve it . looking into the possible solutions . 

Also `dropIndex` does not work for PostgreSQL as it need to run DB statement as 
```sql
ALTER TABLE subeditor_submission_group DROP CONSTRAINT section_editors_pkey
```
tested in my local `psql 13+` .

jonasraoni: I didn't watch the PR, but this sounds like https://github.com/pkp/pkp-lib/issues/7670

https://github.com/pkp/pkp-lib/blob/e13b11a2a799a4627430e5193fd83686bbc0295e/classes/migration/upgrade/v3_4_0/I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields.php#L98-L107

Due to differences on how the table was created (old framework vs Laravel stuff), the unique key might exist as a constraint or an index 😅

touhidurabir: @jonasraoni yes, this is the same problem happening for the PostgreSQL .

controllers/grid/settings/sections/form/PKPSectionForm.php
touhidurabir: I also thought about that but later decided not to do that as that makes the DAO method too big . Moved it .

touhidurabir: Reverted the changes as same issue spotted for `Category` . So the `DAO::getAssignedUserGroupIds()` can not have a generalised implementation as we need to filter by section and category, 2 different entity .

classes/context/SubEditorsDAO.php
```diff
@@ -297,6 +298,28 @@ public function assignEditors(Submission $submission, Context $context): Collect
 
         return $assignments->map(fn ($assignment) => $assignment->userId);
     }
+
+    /**
+     * Get the section assigned sub eidtor's associated user groups ids for given section
```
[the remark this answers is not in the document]
asmecher: Don't forget this one, thanks!

classes/migration/upgrade/v3_4_0/I8737_SectionEditorsUniqueIndexUpdate.php
asmecher: I'm probably already covering this with https://github.com/pkp/pkp-lib/issues/3573#issuecomment-1480464388 -- this should standardize the constraints/indexes by recreating them using `$table->unique` during the 3.3.0 to 3.4.0 upgrade. Those haven't been merged yet but hopefully this problem will go away then.

PR #8792, opened 2023-03-12
Feature/main/namespacing typehinting

jonasraoni:


classes/components/forms/context/PKPDoiRegistrationSettingsForm.php
```diff
@@ -62,7 +63,7 @@ public function __construct(string $action, array $locales, Context $context)
 
         $this->agencyFields = [];
 
-        $registrationAgencies->each(function (IDoiRegistrationAgency $agency) use (&$options, $context) {
+        $registrationAgencies->each(function (IDoiRegistrationAgency|Plugin $agency) use (&$options, $context) {
```
asmecher, line 66: @jonasraoni, are you sure this is necessary? It looks to me like this can only be `CrossRefPlugin` or `DatacitePlugin`, both of which are `Plugin`s but also implement `IDoiRegistrationAgency`.

classes/payment/Payment.php
```diff
@@ -148,9 +148,9 @@ public function setContextId($contextId)
     }
 
     /**
-     * Set the type for this payment (PAYMENT_TYPE_...)
+     * Set the type for this payment (OJSPaymentManager::PAYMENT_TYPE_...)
      *
-     * @param int $type PAYMENT_TYPE_...
+     * @param int $type OJSPaymentManager::PAYMENT_TYPE_...
```
asmecher, line 153: Shouldn't be referring to OJS-specific classes from pkp-lib.

classes/components/forms/context/PKPDoiRegistrationSettingsForm.php
jonasraoni: The reasoning for this is that some places are also using operations from the `Plugin` class (e.g. `$agency->getName()`).

The most meaningful way would be to use the intersection type `IDoiRegistrationAgency&Plugin`, but it requires PHP 8.1.

classes/payment/Payment.php
jonasraoni: Yep! I removed the prefix.

classes/components/forms/context/PKPDoiRegistrationSettingsForm.php
asmecher: OK, that makes sense to me.

Issue #8807, opened 2023-03-14
Plugin installation doesn't cleanup temporary data properly


jonasraoni:
**Describe the bug**
After attempting to install/upgrade a plugin, it always leave some cruft behind:
![image](https://user-images.githubusercontent.com/361921/224993527-13a8712d-e309-405c-91d3-ac88e4574f14.png)

**To Reproduce**
1. Install a plugin
2. Watch out the folder where it was extracted (at main `/temp`, at 3.3.0 "it depends")
3. See the temporary folders

**What application are you using?**
*

**PRs**
- **[TESTS ONLY]** https://github.com/pkp/ojs/pull/3833
- https://github.com/pkp/pkp-lib/pull/8808

jonasraoni:
@asmecher I just remembered of another issue at the plugin installation that I've seen in the wild, so I created this extra issue...
PRs
- **[TESTS ONLY]** https://github.com/pkp/ojs/pull/3833
- https://github.com/pkp/pkp-lib/pull/8808

asmecher:
Merged the pkp-lib PR; I'll take care of the app submodules separately.

PR #8808, opened 2023-03-14
pkp/pkp-lib#8807 Fixed the plugin installation cleanup

jonasraoni:


classes/plugins/PluginHelper.php
```diff
@@ -41,14 +41,16 @@ class PluginHelper
     public const PLUGIN_UPGRADE_FILE = 'upgrade.xml';
 
     /**
-     * Extract and validate a plugin (prior to installation)
+     * Extract the plugin, executes the callback, then cleanup the files
      *
+     * @template T of mixed
```
jonasraoni, line 46: Poor man's generics (it works though for the code editor, PHPStan, etc) 😅

classes/plugins/PluginHelper.php
```diff
@@ -72,17 +74,15 @@ private function extractPlugin(string $filePath, string $originalFileName): stri
             // Ensure there's a file named "version.xml" at the main directory or at the direct sub-directories
             foreach(new DirectoryIterator($extractPath) as $current) {
                 if ($current->isDir() && $current->getBasename() !== '..' && is_file(($path = "{$current->getPathname()}/") . static::PLUGIN_VERSION_FILE)) {
-                    return $path;
+                    return $onExtracted($path);
```
jonasraoni, line 77: The problem is exactly here, we retrieve the path where plugin was *found*, and the original extracted folder is kind of forgotten, it also happens at OJS 3.3.0, not a big deal as the main folder will be probably empty, but seeing this on the server made it a personal pet peeve 😁

classes/plugins/PluginHelper.php
```diff
@@ -100,10 +100,9 @@ private function extractPlugin(string $filePath, string $originalFileName): stri
      */
     public function installPlugin(string $path, string $originalFileName): Version
     {
-        $fileManager = new FileManager();
-        $sourcePath = $this->extractPlugin($path, $originalFileName);
-        try {
-            $versionFile = $sourcePath . static::PLUGIN_VERSION_FILE;
+        return $this->extractPlugin($path, $originalFileName, function (string $pluginFolder): Version {
```
jonasraoni, line 103: Just enclosed the extraction/cleanup process in a single place and processed through a callback, looks cleaner/safer for me.

classes/plugins/PluginHelper.php
asmecher: I'd really like to avoid further substantive changes for RC1 -- but as a caller of this code I wouldn't expect a callback. I think there are a few problems causing the complexity:
- The "create a container directory with a pseudorandom name" is wrapped into `extractPlugin`, rather than made available as a utility function and called outside of this function
- The "look for version.xml anywhere it might appear and return that directory" behaviour is unnecessary; we have a [standard toolset for building/releasing plugins](https://docs.pkp.sfu.ca/dev/plugin-guide/en/release#build-and-package-your-plugin) now and plugin installation code should expect plugins built accordingly.

jonasraoni: > (...) rather than made available as a utility function and called outside of this function

In general, I like such callbacks, especially in the form of language constructions (e.g. `using (var file = new File(...)) { code; }`), that disposes of the underlying resources, it decreases the amount of boilerplate code and brings extra safety (I'm unable to handle it wrong) 🤔

> The "look for version.xml anywhere it might appear and return that directory" behaviour is unnecessary

The previous code already had such checks in place, so I just made it more generic, I don't see it as a big deal, so I'm ok to keep these 3 lines :)

As the main ideas were preserved, it means this issue of leaving the main extracted folder forgotten, is also present at 3.3.0.

PR #8832, opened 2023-03-20
Bugfix/main/8800 fix filter

jonasraoni:


classes/filter/PersistableFilter.php
```diff
@@ -204,23 +204,15 @@ public function hasSettings()
...
-     * (This must be hard coded by sub-classes for PHP4 compatibility.
-     * PHP4 always returns class names lowercase which we cannot
-     * tolerate as we need this path to find the class on case sensitive
-     * file systems.)
      */
-    public function getClassName()
+    public function getClassName(): string
     {
-        assert(false);
+        return static::class;
```
jonasraoni, line 215: `static::class` addresses all the implementations. I was going to add a `final` keyword, but it would be a breaking change 🤔

PR #8833, opened 2023-03-20
Bugfix/main/8829 fix select multiple style

jonasraoni:


styles/form/form.less
```diff
@@ -106,6 +106,10 @@
 		border: @bg-border;
 	}
 
+	select[multiple] {
```
asmecher, line 109: This is a regression from 3.3.0-x, right? If so, I suspect a rule specified elsewhere is accidentally targeting these fields. Rather than overriding this, it might be better to find out whether another rule needs better specificity.

asmecher (changes requested):

styles/form/form.less
jonasraoni: I've just checked and it's happening at 3.3.0 as well.

Issue #8868, opened 2023-04-01
Installing a plugin using commandline tool  exits   if email migration is included


withanage:
**Describe the bug**

Cannot install the email templates of plugins using the 
`php  lib/pkp/tools/installPluginVersion.php `

**To Reproduce**

-  Download the orcidProfile Plugin to `$OJS/plugins/generic` e.g.  ` git clone https://github.com/pkp/orcidProfile`
- Execute inside  `$OJS`   folder the following command  `php  lib/pkp/tools/installPluginVersion.php plugins/generic/orcidProfile/version.xml`
- Result : `PHP Warning:  foreach() argument must be of type array|object, null given in /var/www/html/ojs-main/lib/pkp/classes/plugins/Plugin.php on line 592`

**What application are you using?**
 OJS 3.4

**Additional information**

-  In Command-Line , the installer may not have any information about the Context including [locales](https://github.com/pkp/pkp-lib/blob/main/classes/plugins/Plugin.php#L592)

asmecher:
@withanage, the `installPluginVersions.php` tool isn't intended to install email templates; that can only be done by uploading the plugin .tar.gz into the UI, or using the Plugin Gallery, or installing the system with the plugin present.

PR #8869, opened 2023-04-01
pkp-lib#8868 Defaulting to sitewide locale, when installer locale is missing

withanage:
# Issue

https://github.com/pkp/pkp-lib/issues/8868


withanage:
Hi @Vitaliy-1 ,

I was not sure, if there was a better way to solve this. But I could go to the plugin level , after  defaulting to site-locale fpr #8868

PR #8907, opened 2023-04-11
Bugfix/main/8905 fix event discovery

jonasraoni:


classes/core/PKPEventServiceProvider.php
```diff
@@ -39,22 +39,28 @@ class PKPEventServiceProvider extends EventServiceProvider
...
     public function getEvents()
     {
-        $cacheManager = CacheManager::getManager();
-        /** @var FileCache */
-        $cache = $cacheManager->getCache('event', PKPApplication::CONTEXT_SITE, function (FileCache $cache) {
-            $cache->setEntireCache($this->discoveredEvents());
-        });
-
+        $expiration = DateInterval::createFromDateString(static::MAX_CACHE_LIFETIME);
+        $events = Cache::remember(static::getCacheKey(), $expiration, fn () => $this->discoveredEvents());
```
Vitaliy-1, line 60: Do we still need this workaround after adopting Laravel cache or can we delegate it back to the Container?

tools/events.php
```diff
@@ -142,8 +141,7 @@ protected function list(): void
      */
     protected function clear(): void
     {
-        $cacheManager = CacheManager::getManager();
-        $cacheManager->flush('event');
+        PKPEventServiceProvider::clearCache();
```
Vitaliy-1, line 144: Likely a general problem, the cache can't be forcefully cleared from the administration area. I guess there was already discussion somewhere, right?

Vitaliy-1 (changes requested):

classes/core/Core.php
```diff
@@ -498,18 +498,16 @@ private static function _getUrlComponents($urlInfo, $isPathInfo, $offset, $varNa
...
-        if (Str::startsWith($path, 'classes')) {
-            $path = Str::replaceFirst('classes/', '', $path);
+        $libPath = realpath(base_path(PKP_LIB_PATH));
+        $isLib = str_starts_with($file->getRealPath(), $libPath);
+        $className = str_replace($isLib ? $libPath : realpath(base_path()), '', $file->getRealPath());
+        $className = preg_replace('#^[\\\\/]classes|\.php$#', '', $className);
+        $className = str_replace('/', '\\', '/' . ($isLib ? 'PKP' : 'APP') . $className);
+
+        if (!class_exists($className)) {
+            error_log(new Exception("Failed to map the file \"{$file->getRealPath()}\" to a class"));
```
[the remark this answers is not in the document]
jonasraoni: Considering that we might still have files that are not namespaced, and that we're close to release the 3.4, I thought it's better to avoid surprises, so I just left the message with a helpful trace :)

classes/core/PKPEventServiceProvider.php
jonasraoni: Here I wanted to inherit this class from the `Illuminate\Foundation\Support\Providers` instead of `Illuminate\Events\EventServiceProvider` and just override with customizations.

But I think it's worth to get this fix into 3.4.0, so I played safe and it was just a quick fix for a problem that I found while working in another task.

tools/events.php
jonasraoni: Now I remembered that I've talked about it on a task to retire the `CacheManager`, but it ended up forgotten. Here's a new fresh issue: https://github.com/pkp/pkp-lib/issues/8922 😅

classes/core/PKPEventServiceProvider.php
jonasraoni: FYI I've just checked the implementation, and it's a kind of "hard" cache, which is created through an "artisan" command, so I'll keep the customization.

jonasraoni:
@Vitaliy-1 Not sure if you want to review again, so I've sent a request 😅
Basically I changed the parent class of the `PKPEventServiceProvider` to reuse more code and also imported some extra pieces from the Laravel's Application, to support better the expectations of `ServiceProviders`.

Vitaliy-1 (approved):
I've reviewed this, should be safe to merge. I don't mind adding a possibility to register bindings from service providers and booting callbacks. And it doesn't look like extending Foundation/.../EventServiceProvider should have any drawbacks.

PR #8908, opened 2023-04-11
Bugfix/main/8906 fix submission indexing

jonasraoni:


jobs/submissions/UpdateSubmissionSearchJob.php
```diff
@@ -53,14 +53,15 @@ public function handle(): void
             throw new JobException(JobException::INVALID_PAYLOAD);
         }
 
+        $submissionSearchIndex = Application::getSubmissionSearchIndex();
         if ($submission->getData('status') !== PKPSubmission::STATUS_PUBLISHED) {
-            Application::getSubmissionSearchIndex()->deleteTextIndex($submission->getId());
-            Application::getSubmissionSearchIndex()->clearSubmissionFiles($submission);
+            $submissionSearchIndex->deleteTextIndex($submission->getId());
+        } else {
```
jonasraoni, line 59: On OJS 3.3 we have this switch to decide whether the submission should be indexed.

In my opinion we should index unpublished content as well, then the search index can be used at the administrative panel as well. If you agree, I can create another issue to change this behavior...

asmecher: Right now the built-in search index is not intended for use outside the public search feature. Including it would mean a pretty big rewrite of the search filters etc. in the submission lists, and we haven't had any requests for that. I'd suggest keeping it as is for now.

asmecher:
Merged, thanks!

jobs/submissions/UpdateSubmissionSearchJob.php
jonasraoni: Ok! Just asked because I did a related work on another PR (used the indexed content) 😁

PR #8912, opened 2023-04-12
Feature/main/8710 address recommend by similarity performance

jonasraoni:


classes/submission/Collector.php
```diff
@@ -288,22 +303,44 @@ public function getQueryBuilder(): Builder
...
                 $q->addSelect([DB::raw($coalesceTitles)]);
                 $q->orderBy(DB::raw($coalesceTitles), $this->orderDirection);
                 break;
+            case self::ORDERBY_SEARCH_RANKING:
+                if (!$keywords->count()) {
+                    $q->orderBy('s.date_submitted', $this->orderDirection);
+                    break;
+                }
+                // Retrieves the number of matches for all keywords
+                $orderByMatchCount = DB::table('submission_search_objects', 'sso')
```
jonasraoni, line 328: Comment for here and below, a single query using `keyword IN (x, y, z)` becomes very slow if one of the values doesn't exist (I tested locally and on a production database with a large dataset) o_O

classes/submission/Collector.php
```diff
@@ -383,40 +419,79 @@ public function getQueryBuilder(): Builder
...
+                    ->whereIn('ra.reviewer_id', $this->assignedTo)
+                    ->select(DB::raw('1 AS value'))
+                    ->limit(1),
+                    'any_assignment', 'any_assignment.value', '=', DB::raw('1')
+                );
             }
+            // Builds the filters
+            $q->where(fn (Builder $q) => $keywords
+                ->map(fn (string $keyword) => $q
+                    // Look for matches on the indexed data
```
jonasraoni, line 438: This filter can be slightly improved by nesting an extra check `Is submission published? Use indexed data : search elsewhere`.

That's why I asked if it was ok to let the [search indexer handle non-published content](https://github.com/pkp/pkp-lib/pull/8908#pullrequestreview-1381470606) :)

classes/submission/Collector.php
```diff
@@ -383,40 +419,79 @@ public function getQueryBuilder(): Builder
...
+                        ->whereIn('aus.setting_name', [
+                            Identity::IDENTITY_SETTING_GIVENNAME,
+                            Identity::IDENTITY_SETTING_FAMILYNAME,
+                            'orcid'
+                        ])
+                        // Don't permit reviewers to search on author names
+                        ->when(!empty($this->assignedTo), fn (Builder $q) => $q
+                            ->where(fn (Builder $q) => $q
+                                ->whereNull('any_assignment.value')
+                                ->orWhereNotIn('aus.setting_name', [
```
jonasraoni, line 476: The previous code would skip searching the ORCID field

classes/submission/Collector.php
```diff
@@ -270,6 +281,10 @@ public function getQueryBuilder(): Builder
             $q->whereIn('s.context_id', $this->contextIds);
         }
 
+        // Prepare keywords (allows short and numeric words)
+        $keywords = collect(Application::getSubmissionSearchIndex()->filterKeywords($this->searchPhrase, false, true, true))->unique();
```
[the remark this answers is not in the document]
asmecher: I'd like to keep rate limits etc. in the controller part of the app rather than the model.

classes/submission/Collector.php
```diff
@@ -425,14 +500,13 @@ public function getQueryBuilder(): Builder
...
         // By any child pub object's DOI status
-        $q->when($this->doiStatuses !== null, function (Builder $q) {
-            $this->addDoiStatusFilterToQuery($q);
-        });
+        $q->when($this->doiStatuses !== null, fn (Builder $q) => $this->addDoiStatusFilterToQuery($q));
+
+        // By whether any child pub objects have DOIs assigned
+        $q->when($this->hasDois !== null, fn (Builder $q) => $this->addHasDoisFilterToQuery($q));
 
         // By whether any child pub objects have DOIs assigned
```
Vitaliy-1, line 508: This comment needs to be updated

Vitaliy-1 (commented):
Looks great! I'd added a comment regarding possible enhancement

PR #8916, opened 2023-04-13
Feature/main/8915 improve search indexer

jonasraoni:


classes/search/SearchFileParser.php
```diff
@@ -68,8 +70,10 @@ public function setFilePath($filePath)
      */
     public function open()
     {
-        $this->fp = @fopen($this->filePath, 'rb');
-        return $this->fp ? true : false;
+        if (!($this->fp = @fopen($this->filePath, 'rb'))) {
+            throw new Exception("Failed to parse the file \"{$this->filePath}\"\nLast error: " . error_get_last());
```
asmecher, line 74: I think the search parser should be made resilient to missing files; no objection to an exception being thrown here, but if so, then the indexing code should be prepared to catch it and move on with indexing other content.

classes/search/SearchFileParser.php
```diff
@@ -127,22 +131,23 @@ public static function fromFile($submissionFile)
...
-                break;
-            case 'text/html':
-            case 'text/xml':
-            case 'application/xhtml':
-            case 'application/xml':
-                return new \PKP\search\SearchHTMLParser($path);
-                break;
-        }
-        return new \PKP\search\SearchHelperParser($type, $path);
+        $type = Config::getVar('search', "index[{$type}]")
```
asmecher, line 138: This kind of double-ternary construct is pretty hard to read. I guess you could simplify it by giving a default value to `Config::getVar`?
```php
$type = Config::getVar('search', "index[{$type}]", 'default');
```
OTOH this will give you `default` if the configuration line is not specified at all, but will still respect an `Off` setting if you've configured that indexer to be entirely disabled.

classes/config/Config.php
```diff
@@ -124,4 +124,14 @@ public static function &getContextBaseUrls()
 
         return $contextBaseUrls;
     }
+
+    /**
+     * Retrieve wether the specified configuration variable is defined, even if it's null.
```
asmecher, line 129: *whether

asmecher (changes requested):
Thanks, @jonasraoni! I've made some specific comments, but I'd also like to investigate the relationship between the UTF8 characters that are causing problems and the default configuration file suggestions for indexers, which attempt to strip out UTF8. Is there something we can tweak in the default suggestions that will resolve the problem without code changes? If so, we can add a `config.inc.php` default change and release that in 3.3.0-x. It appears that these changes moot those cleanups; if that's the case, I'd favour removing them from the configuration file template in `stable-3_4_0` and `main`.

classes/search/SearchFileParser.php
jonasraoni: This is supposed to be executed inside a job, so I thought about letting the `Exception` leak, in order to retry the job.

But it's better to index the maximum we can, before failing the job, so I'll recover the previous behavior, but let the `Exception` leak at the end.

jonasraoni: Ah, regarding your comment, I just the code is already fine, there's a try/catch in the outer scope.
But I'll let an `Exception` leak at the end, in order to retry.

classes/search/SearchFileParser.php
jonasraoni: I'm ok with the nested stuff, as long as they are not very long, but no worries, I'll update 😁
Yeah, I saw the `getVar()` accepts a default, but it's not suitable for my use case.

PR #8941, opened 2023-04-24
pkp/pkp#8933 Restore original file after cancelling file upload wizard

Vitaliy-1:


controllers/api/file/PKPManageFileApiHandler.php
```diff
@@ -88,6 +90,51 @@ public function deleteFile($args, $request)
...
+		$revisedFileId = $request->getUserVar('fileId') ? (int) $request->getUserVar('fileId') : null;
+
+		// Get revisions and check file IDs
+		$revisions = Repo::submissionFile()->getRevisions($submissionFile->getId());
+		$revisionIds = [];
+		foreach ($revisions as $revision) {
+			$revisionIds[] = $revision->fileId;
+		}
+
+		if (!$revisedFileId && !in_array($revisedFileId, $revisionIds)) {
```
NateWr, line 113: Is this meant to be `||`?

controllers/api/file/PKPManageFileApiHandler.php
```diff
@@ -88,6 +90,51 @@ public function deleteFile($args, $request)
...
+			return new JSONMessage(false);
+		}
+
+		if (!isset($originalFile['fileId']) && !in_array($originalFile['fileId'], $revisionIds)) {
```
NateWr, line 117: Supposed to be `||`?

controllers/api/file/PKPManageFileApiHandler.php
```diff
@@ -88,6 +90,51 @@ public function deleteFile($args, $request)
...
+			return new JSONMessage(false);
+		}
+
+		// Restore original submission file
+		Repo::submissionFile()->edit(
+			$submissionFile,
+			[
+				'fileId' => $originalFile['fileId'],
```
NateWr, line 125: This will work. But in theory I could use this to switch to any revision, not just the last revision. It wouldn't cause any big problems. But instead of getting the revision from a request user var, you could reset the `fileId` to the most recent of the `$revisionIds`.

The method `getRevisions()` returns with the most recent first, so you could use the first revision id in that array.

controllers/api/file/PKPManageFileApiHandler.php
```diff
@@ -88,6 +90,51 @@ public function deleteFile($args, $request)
...
+				'name' => $originalFile['name'],
+				'uploaderUserId' => $originalFile['uploaderUserId'],
```
NateWr, line 127: Again, I'm wary of using request data here, because there's nothing enforcing it to be the original user id. It might be that we need to store this information along with revisions so we can retrieve it safely.

You may also be able to grab this data from the submission event log, which will have an entry for the last revision of this submission file.

controllers/wizard/fileUpload/FileUploadWizardHandler.php
```diff
@@ -418,13 +416,19 @@ public function uploadFile($args, $request)
             return new JSONMessage(false, $uploadForm->fetch($request));
         }
 
+		$submissionFileId = $uploadForm->getRevisedFileId();
+		// Store the data of the submission file before it's replaced by the revised file
+		if ($submissionFileId) {
+			$originalFile = Repo::submissionFile()->get($submissionFileId);
+		}
```
NateWr, line 423: An extra indent in this code.

js/controllers/wizard/fileUpload/FileUploadWizardHandler.js
```diff
@@ -243,7 +258,8 @@
 				this.uploadedFile_.csrfToken = this.csrfToken_;
 				// Authorization policy expects to find the submissionFileId para
 				this.uploadedFile_.submissionFileId = this.uploadedFile_.id;
-				$.post(this.deleteUrl_, this.uploadedFile_,
```
NateWr, line 246: It looks like maybe `deleteUrl_` is no longer used?

NateWr (changes requested):

controllers/api/file/PKPManageFileApiHandler.php
Vitaliy-1: Yes, I thought about that.
User is able to upload indefinite number of revisions through the submission file wizard, thus, I don't know which revision I need to retrieve.

controllers/api/file/PKPManageFileApiHandler.php
Vitaliy-1: Thanks, done!

js/controllers/wizard/fileUpload/FileUploadWizardHandler.js
Vitaliy-1: I think it's still used for file deletion

schemas/eventLog.json
```diff
@@ -0,0 +1,245 @@
...
+				"date_format:Y-m-d H:i:s"
+			]
+		},
+		"decision": {
+			"type": "integer",
+			"description": "The decision type identifier. One of the Decision::* constants.",
+			"validation": [
+				"nullable",
+				"min:1",
+				"max:32"
```
NateWr, line 43: This might cause problems, for example, if a third-party plugin adds a decision and gives it a constant `555`.

schemas/eventLog.json
```diff
@@ -0,0 +1,245 @@
...
+			"type": "string",
+			"description": "The full name of the email sender.",
+			"multilingual": true,
+			"validation": [
+				"nullable"
+			]
+		},
+		"stageId": {
+			"type": "integer",
+			"description": "The workflow stage ID associated submission is in. One of the `WORKFLOW_STAGE_ID_` constants",
```
NateWr, line 185: "is in" --> "was in when this event was logged"

classes/log/event/DAO.php
```diff
@@ -0,0 +1,305 @@
...
+            {
+                $schema = parent::get($schemaName, $forceReload);
+                if (!$this->getCustomProps()) {
+                    return $schema;
+                }
+
+                $customSchema = new \stdClass();
+                $customSchema->properties = json_decode(json_encode($this->getCustomProps()));
+                $schema = $this->merge($schema, $customSchema);
+                $this->_schemas[$schemaName] = $schema;;
```
NateWr, line 125: Two `;;` at the end of the line

NateWr (changes requested):
Just one longer comment about modifying the schema at run-time like this. I don't know if you'll have time to rethink that approach, but something to think about anyway. Otherwise, it's all small comments. :+1:

classes/log/event/DAO.php
```diff
@@ -0,0 +1,305 @@
...
+        $this->schemaService = new class extends PKPSchemaService {
+            /**
+             * Schema for the custom data to be recorded in the event log
+             * [
+             *   'propName' => [
+             *        'type' => ...
+             *        'multilingual' => ... (optional)
+             * ]
+             */
+            protected ?array $customProps = null;
```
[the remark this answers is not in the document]
Vitaliy-1: @asmecher, for 3.4 release I believe we can define all event log properties in schema and use a generic `data` property for outstanding cases, what do you think?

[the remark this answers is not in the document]
asmecher: Yes, this sounds reasonable to me.

api/v1/submissions/PKPSubmissionHandler.php
```diff
@@ -708,17 +710,21 @@ public function submit(SlimRequest $slimRequest, APIResponse $response, array $a
...
-                ]
-            );
+            $user = $request->getUser();
+            $eventLog = Repo::eventLog()->newDataObject([
+                'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION,
+                'assocId' => $submission->getId(),
+                'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_COPYRIGHT_AGREED,
+                'userId' => Validation::loggedInAs() ?? $user->getId(),
+                'message' => 'submission.event.copyrightAgreed',
+                'isTranslated' => 0,
```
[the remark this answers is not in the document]
Vitaliy-1: Ahh, yes, it appears that PostgreSQL supports true boolean values.

classes/services/PKPSchemaService.php
```diff
@@ -42,6 +42,7 @@ class PKPSchemaService
     public const SCHEMA_SUBMISSION_FILE = 'submissionFile';
     public const SCHEMA_USER = 'user';
     public const SCHEMA_USER_GROUP = 'userGroup';
+	public const SCHEMA_EVENT_LOG = 'eventLog';
```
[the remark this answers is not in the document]
Vitaliy-1: All those tabs are due to switching to/from them when editing schema json. I think we need to standardise that also.

classes/log/event/Repository.php
```diff
@@ -0,0 +1,159 @@
...
+     * @return array A key/value array with validation errors. Empty if no errors
+     */
+    public function validate(?EventLogEntry $object, array $props, array $allowedLocales, string $primaryLocale): array
+    {
+        $validator = ValidatorFactory::make(
+            $props,
+            $this->schemaService->getValidationRules($this->dao->schema, $allowedLocales)
+        );
+
+        // Check required fields if we're adding a context
```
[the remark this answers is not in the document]
Vitaliy-1: That was copypasted quite a lot :)

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
```diff
@@ -0,0 +1,247 @@
...
+                // Determine locale
+                $locale = $this->getContextPrimaryLocale($row, $sitePrimaryLocale);
+                if (!$locale) {
+                    return;
+                }
+                DB::table('event_log_settings')
+                    ->where('event_log_setting_id', $row->event_log_setting_id)
+                    ->update(['locale' => $locale]);
+            });
+    }
```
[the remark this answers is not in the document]
Vitaliy-1: I think I know hot to make it more performant - by 1) moving event types to the `whereIn` statement, 2) updating multiple rows at once (for each determined locale). Nevertheless, I'll preserve chunking results.

classes/migration/install/LogMigration.php
```diff
@@ -49,10 +49,11 @@ public function up(): void
             $table->foreign('log_id', 'event_log_settings_log_id')->references('log_id')->on('event_log')->onDelete('cascade');
             $table->index(['log_id'], 'event_log_settings_log_id');
 
+            $table->string('locale', 14)->default('');
             $table->string('setting_name', 255);
             $table->mediumText('setting_value')->nullable();
-            $table->string('setting_type', 6)->comment('(bool|int|float|string|object)');
-            $table->unique(['log_id', 'setting_name'], 'event_log_settings_unique');
+            $table->string('setting_type', 6)->comment('(bool|int|float|string|object)')->nullable();
```
asmecher, line 55: I think I'd rather not introduce a `nullable` here, for consistency with the `setting_type` columns on other `_settings` tables. Feel free to force a `string` to restored entries, since that should work for all existing entries.

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
```diff
@@ -0,0 +1,243 @@
...
+            $table->unique(['log_id', 'locale', 'setting_name'], 'event_log_settings_unique');
+            $table->string('setting_type', 6)->nullable()->change();
+        });
+
+        // Events can be triggered without a user, e.g., in schedule tasks
+        Schema::table('event_log', function (Blueprint $table) {
+            $table->dropForeign('event_log_user_id_foreign');
+            $table->dropIndex('event_log_user_id');
+            $table->bigInteger('user_id')->nullable()->change();
+            $table->foreign('user_id')->references('user_id')->on('users')->onDelete('cascade');
```
asmecher, line 51: Are you sure you want an `ON DELETE CASCADE` here? I would think `ON DELETE SET NULL` would be more appropriate -- the point is to be able to continue to log events when a user account doesn't exist, right?

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
```diff
@@ -0,0 +1,243 @@
...
+                SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_UPLOAD,
+                SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_EDIT,
+                SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_REVISION_UPLOAD,
+            ])
+            ->orderBy('event_log_setting_id')
+            ->chunk(100, function (Collection $logChunks) use ($sitePrimaryLocale) {
+                $mapLocaleWithSettingIds = [];
+                foreach ($logChunks as $row) {
+                    // Get locale based on a submission file ID log entry
+                    $locale = $this->getContextPrimaryLocale($row, $sitePrimaryLocale);
```
asmecher, line 76: I think you can do this all in SQL for a significant performance boost. Do a `LEFT JOIN` to the appropriate table for each `assoc_type`, then join on journals (contexts) using `COALESCE() around the candidate IDs; that'll give you a single `journals` join regardless of what kind of `assoc_type` was present.

asmecher: Thinking further about performance: would we be able to turn this around to optimize it? Currently it's...
- for each file upload, edit, or revise event,
  - get its context ID's primary locale, and
  - update the setting's locale column to match.

If I understand it right, it would probably be a lot faster to go the other way around:
- For each context,
  - Get its primary locale, and
  - Update any file upload, edit, or revise event log entry settings to match.

classes/migration/install/LogMigration.php
Vitaliy-1: Oоps, setting_type is not longer needed and should be removed - I made it nullable in my attempt to record custom log entries.

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
Vitaliy-1: Thanks, that was also on my mind as a belated idea but I wasn't sure if the performance boost would be worth to re-implement it

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
```diff
@@ -0,0 +1,243 @@
...
+        // Rename ambiguous settings
+        $this->renameSettings();
+
+        // Localize existing submission file name entries
+        $sitePrimaryLocale = DB::table('site')->value('primary_locale');
+        DB::table('event_log_settings AS es')
+            ->join('event_log AS e', 'es.log_id', '=', 'e.log_id')
+            ->where('setting_name', 'filename')
+            ->whereIn('event_type', [
+                SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_UPLOAD,
```
asmecher, line 67: I would suggest hard-coding the constant values here, with e.g. `SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_UPLOAD` in an inline comment. This means grepping will still work but the constants will be able to evolve in future code without affecting the upgrade.

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
Vitaliy-1: Correct me if I'm wrong: the only appropriate way to remove a user account is to merge it with another one. For this specific case we also combine their log entries. Thus, on delete cascade here would prevent data corruption in unexpected cases. If not, I can set to null on delete.

Forgot to mention, the idea behind making `user_id` nullable, is to allow to log events from the system, automated events or other cases when a specific user isn't involved.

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
asmecher: Just for reference, there are 3M entries in SciELO's event log table; 567k are in (SUBMISSION_LOG_FILE_UPLOAD, SUBMISSION_LOG_FILE_EDIT, SUBMISSION_LOG_FILE_REVISION_UPLOAD). So the cost of going through these individually would be very high.

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
asmecher: That makes sense to me, thanks, please go ahead. But please do add a `->comment` to the `user_id` column in `LogMigration` to explain what a null value means (we'll add more of these in future releases)!

classes/migration/upgrade/v3_4_0/I8933_EventLogLocalized.php
Vitaliy-1: I'm getting context primary locale having only submission or submission file. Event log doesn't have `context_id` column, thus I need to iterate over all those entries. I meant to move statement that selects context locale out of the loop.

I've optimised migration where possible. The major change is setting chunking to 10000, which means that the update statement will run once for 10k entries. On my laptop 500k records are updated within 15 minutes. If it's not enough, I can try to create a wider join statement, a series of something like:
```sql
SELECT * FROM event_log AS e
JOIN event_log_settings AS es on e.log_id = es.log_id
JOIN submissions AS s ON e.assoc_id = s.submission_id
WHERE e.assoc_type = 1048585 
AND es.setting_name = 'filename'
AND s.context_id = '?'
```
Then updating log entries altogether, knowing the associated context and its primary locale. It might be faster but less prone to corrupted data.

asmecher: That sounds good, thanks -- please go ahead, and I'll do a sanity check with a large dataset once all upgrade-related merges are complete in order to make sure we don't regress on performance.

Issue #8871, opened 2023-04-03
Add ability to login via Email


touhidurabir:
**Describe the enhancement**
Current OJS/OMP/OPS system only allow login via username . As the username is not changeable and a lot of time set via JM/System admin, it's sometime become lot difficult for user to remember it . Allowing the login via user email can be considered a good solution as in the system of OJS/OMP/OPS, as both the username and email of a user is unique .

**What application are you using?**
 OJS, OMP or OPS version 3.4.0 

**Who is asking for it**
It's a feature that has been requested quite a few times in the PKP forum . 

**Additional information**
This will provide alternative solution for a regularly requested feature to allow the update of username . One challenge with username is that it's used as the salt for the password hashing and need some consideration to implement it . As updating email is allowed, it can provide an alternative approach for it . 

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9012 [TRANSLATION UPDATE ONLY] (merged)

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/8950 (merged)
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3886 [TEST ONLY]

#### stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9416
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4066 [TEST ONLY]

jonasraoni:
There's a related discussion here: https://github.com/pkp/pkp-lib/discussions/7965

asmecher:
Thanks, @touhidurabir -- I can't merge this for 3.4.0-0 as most of the translators have already completed their work and this would be a breaking change for them. Here's what I suggest:
- Let's leave the PRs for the feature addition for merge later, *but*...
- Please create and merge a PR that adds the new locale key to the English translation, so that translators will be prompted to add it for other translations in the meantime.

ajnyga:
This is a very important new feature, thank you!

But do we actually need the username at all after this?

asmecher:
Thanks! I've merged the translation PR and the rest can wait until a later 3.4.0-x.

mfelczak:
[...]

PR #8950, opened 2023-04-30
pkp/pkp-lib#8871 Added login via email along with username

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8871

### DIFF
--- a/classes/user/form/BaseProfileForm.php
+++ b/classes/user/form/BaseProfileForm.php
@@ -18,6 +18,7 @@ namespace PKP\user\form;
 
 use APP\core\Application;
 use APP\facades\Repo;
+use PKP\session\SessionManager;
 use PKP\form\Form;
 use PKP\user\User;
 
@@ -61,6 +62,17 @@ abstract class BaseProfileForm extends Form
         $request = Application::get()->getRequest();
         $user = $request->getUser();
         Repo::user()->edit($user);
+
+        if ($functionArgs['emailUpdated'] ?? false) {
+            $sessionManager = SessionManager::getManager();
+            $session = $sessionManager->getUserSession();
+
+            if ($session->getSessionVar('email')) {
+                $session->setSessionVar('email', $user->getEmail());
+            }
+
+            $sessionManager->invalidateSessions($user->getId(), $sessionManager->getUserSession()->getId());
+        }
     }
 }
 
--- a/classes/user/form/ContactForm.php
+++ b/classes/user/form/ContactForm.php
@@ -114,6 +114,7 @@ class ContactForm extends BaseProfileForm
     public function execute(...$functionArgs)
     {
         $user = $this->getUser();
+        $functionArgs['emailUpdated'] = $user->getEmail() !== $this->getData('email');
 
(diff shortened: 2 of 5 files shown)


pages/login/LoginHandler.php
```diff
@@ -75,7 +75,7 @@ public function index($args, $request)
         $templateMgr = TemplateManager::getManager($request);
         $templateMgr->assign([
             'loginMessage' => $request->getUserVar('loginMessage'),
-            'username' => $request->getUserVar('username') ?? $session->getSessionVar('username'),
+            'username' => $session->getSessionVar('email') ?? $session->getSessionVar('username'),
```
asmecher, line 78: @touhidurabir, can you explain this change? I'm not sure I'm following.

touhidurabir: @asmecher this is to prefill the auth key (username/email) with the email if one of last login done via email instead of username . So if user login using the email, then logout and if we only have 
```php
$request->getUserVar('username') ?? $session->getSessionVar('username')
```
the the login form is prefilled with username only even if their last login is via email . as once a login done via email, we are putting the email information in session , we can prefill it in the login form . 

I am happy that you raised this point . This lead me to think 2 more cases 
1. If we keep the changes that I have proposed, I think we need to update the behaviour a bit when a user update the email . Otherwise it will sill prefill the older/incorrect email address . So we need to update the session data on email update . 
2. As we are allowing email as auth key and which is possible to update unlike the `username`, I think we need to implement the session invalidation like we did at https://github.com/pkp/pkp-lib/issues/6983 . 

what you think above 2 points ?

classes/user/form/ContactForm.php
```diff
@@ -114,6 +114,7 @@ public function readInputData()
     public function execute(...$functionArgs)
     {
         $user = $this->getUser();
+        $functionArgs['emailUpdated'] = $user->getEmail() !== $this->getData('email');
```
touhidurabir, line 117: @asmecher  The reason we are not updating the session and clearing out all but existing session here is because the data commit is not happening here but in the `PKP\user\form\BaseProfileForm::execute` and we don't make those changes until the changes are pushed to DB itself . Thats why we are setting just a flag here .

classes/user/form/BaseProfileForm.php
```diff
@@ -61,6 +62,17 @@ public function execute(...$functionArgs)
...
+        if ($functionArgs['emailUpdated'] ?? false) {
+            $sessionManager = SessionManager::getManager();
+            $session = $sessionManager->getUserSession();
+
+            if ($session->getSessionVar('email')) {
+                $session->setSessionVar('email', $user->getEmail());
+            }
+
+            $sessionManager->invalidateSessions($user->getId(), $sessionManager->getUserSession()->getId());
+        }
```
touhidurabir, line 75: see the comment at https://github.com/pkp/pkp-lib/pull/8950/files#r1358343133

pages/login/LoginHandler.php
touhidurabir: @asmecher implementation done based on above those 2 point .

PR #8952, opened 2023-05-01
Bugfix/main/8951 fix native xml plugin

jonasraoni:


plugins/importexport/native/filter/NativeXmlPKPPublicationFilter.php
```diff
@@ -80,11 +80,12 @@ public function handleElement($node)
         $publication->setData('seq', $node->getAttribute('seq'));
         $publication->setData('accessStatus', $node->getAttribute('access_status'));
         $publication->setData('status', $node->getAttribute('status'));
-        $publication->setData('primaryContactId', $node->getAttribute('primary_contact_id'));
         $publication->setData('urlPath', $node->getAttribute('url_path'));
 
         $publicationId = Repo::publication()->dao->insert($publication);
         $publication = Repo::publication()->get($publicationId);
+        // Non-persisted temporary ID, will be updated and stored once the authors get parsed
+        $publication->setData('primaryContactId', $node->getAttribute('primary_contact_id'));
```
jonasraoni, line 88: This isn't very safe (e.g. in case another part of the code decides to use the `$deployment->setPublication()`), but instead of creating/spreading extra workarounds, maybe it's better to proceed with the refactoring or retirement (JATS?!) of this plugin.
If you want, I can come up with a better approach (store the data in a class property).

The same applies to the other fix.

classes/cliTool/CommandLineTool.php
```diff
@@ -67,10 +67,13 @@ public function __construct($argv = [])
         $request->setRouter($router);
 
         // Initialize the locale and load generic plugins.
-        $contextDao = Application::getContextDAO(); /** @var ContextDAO $contextDao */
-        $contextFactory = $contextDao->getAll(); /** @var DAOResultFactory $contextFactory */
-        foreach ($contextFactory->toIterator() as $context) { /** @var Context $context */
-            PluginRegistry::loadCategory('generic', false, $context->getId());
+        PluginRegistry::loadCategory('generic');
+        if (!$application->isUnderMaintenance()) {
```
jonasraoni, line 71: @bozana This was breaking the OPS build, so I did this update.

bozana: Hmmm... When the first `PluginRegistry::loadCategory('generic'); ` is executed, the plugins will be loaded and registered but without context.
Then when the next `PluginRegistry::loadCategory('generic', false, $context->getId());` is called the plugin will not be registered again, because of this line https://github.com/pkp/pkp-lib/blob/7d62c31d1438a43df1402df503d9f42ed826cb3b/classes/plugins/PluginRegistry.php#L76. There `isset($plugins[$category][$pluginName])` will be `true` and the `register` will not be called at all. Thus the plugins will again need to extra call their register function, in order for the hooks to work.
Maybe we should then leave it so, but then the whole block `if (!$application->isUnderMaintenance()) {` is not needed.
Or, do you maybe know where exactly in OPS there was a problem? It seems like a CLI tool is executed before a context is created? Else, I can maybe run the tests locally... So it can be differently solved in OPS, or to leave it as it was earlier...

jonasraoni: > The plugins will be loaded and registered but without context.

This part `but without context` makes me curious about the right behavior here (I didn't inspect the original issue), errrrr if we have N journals with the plugin enabled, which one should be the chosen one?

I didn't inspect the original issue, but if calling the register with a specific context is important, then I think the `isset($plugins[$category][$pluginName])` should be `isset($plugins[$category][$pluginName][$mainContextId])`.

> do you maybe know where exactly in OPS there was a problem?

OPS had a `journals` table, which became `servers` in 3.4. So if you try to load the contexts before the migration, it will try to load data from the `servers` table, which doesn't exist yet (it will be created here: https://github.com/pkp/ops/blob/main/classes/migration/upgrade/v3_4_0/I6759_RenameVariables.php#L58).

bozana: Yes... I noticed that `php tools/upgrade.php` was called... causing that...
Hmmm... I am not sure if generic plugins are ever need to be loaded during install or upgrade... but... maybe to be on the safe side I would return it how it was and will register the plugin for a context within the plugin's tool itself, as it was earlier...

bozana: Theoretically, we could maybe only leave that part `if (!$application->isUnderMaintenance()) {` i.e. without that first `PluginRegistry::loadCategory('generic');` and leave the install and upgrade to load the generic plugins differently if needed, but I hesitate to make that change now. Maybe this can be considered some time later...

bozana: > This part but without context makes me curious about the right behavior here (I didn't inspect the original issue), errrrr if we have N journals with the plugin enabled, which one should be the chosen one? I mean, is it ok that only one will be "connected"?

I think then the plugin defines the behavior -- the site-wide plugins work normally, but journal specific will not register for the hooks and thus can only be partly used (e.g. only those function that do not require hook registration).

> I didn't inspect the original issue, but if calling the register with a specific context is important, then I think the isset($plugins[$category][$pluginName]) should be isset($plugins[$category][$pluginName][$mainContextId]).

`$plugins[$category][$pluginName]` does not contain context IDs keys, only plugin object, I think. And as said above, there is then only one plugin object that registers for hooks for different contexts.
So it seems like plugin loading and hook (registration) maintenance is different...

Issue #8960, opened 2023-05-04
Load generic plugins for all contexts in the CLI tool construct


bozana:
Here https://github.com/pkp/pkp-lib/blob/main/classes/cliTool/CommandLineTool.php#L70 the generic plugins are loaded but without a context. However, the generic plugins register for Hooks only if they are enabled for a context. 
Load generic plugins here for all contexts, so they can register for Hooks and work fully/properly in any CLI too.

PRs:
- https://github.com/pkp/pkp-lib/pull/8961
- https://github.com/pkp/ojs/pull/3891 (only submodule update)

bozana:
@asmecher, could you please take a look?

asmecher:
Hmm, I was sure I gave the go-ahead for this already, but I don't see it now! Anyway, yes, please proceed.

PR #8961, opened 2023-05-04
pkp/pkp-lib#8960 load generic plugins for all contexts in CommandLine…

bozana:
…Tool

s. https://github.com/pkp/pkp-lib/issues/8960

### DIFF
--- a/classes/cliTool/CommandLineTool.php
+++ b/classes/cliTool/CommandLineTool.php
@@ -67,7 +67,11 @@ class CommandLineTool
         $request->setRouter($router);
 
         // Initialize the locale and load generic plugins.
-        PluginRegistry::loadCategory('generic');
+        $contextDao = Application::getContextDAO(); /** @var ContextDAO $contextDao */
+        $contextFactory = $contextDao->getAll(); /** @var DAOResultFactory $contextFactory */
+        foreach ($contextFactory->toIterator() as $context) { /** @var Context $context */
+            PluginRegistry::loadCategory('generic', false, $context->getId());
+        }
 
         $this->argv = isset($argv) && is_array($argv) ? $argv : [];
 


classes/cliTool/CommandLineTool.php
```diff
@@ -67,7 +67,11 @@ public function __construct($argv = [])
         $request->setRouter($router);
 
         // Initialize the locale and load generic plugins.
-        PluginRegistry::loadCategory('generic');
+        $contextDao = Application::getContextDAO(); /** @var ContextDAO $contextDao */
+        $contextFactory = $contextDao->getAll(); /** @var DAOResultFactory $contextFactory */
```
jonasraoni, line 71: Ah, if you want to make the code more generic, the `DAOResultFactory` has a `toIterator()` which you could use with a `foreach`.

classes/cliTool/CommandLineTool.php
```diff
@@ -67,7 +67,11 @@ public function __construct($argv = [])
...
+        while ($context = $contextFactory->next()) { /** @var Context $context */
+            PluginRegistry::loadCategory('generic', false, $context->getId());
```
jonasraoni, line 73: Does it this cover site-wide plugins (I didn't inspect)?
Another possible issue is running a CLI tool when there's no contexts (is there a use-case for that?)

Perhaps it's ok to keep the previous `PluginRegistry::loadCategory('generic');` together with this new code (I remember there was something to avoid registering the same plugin twice).

bozana: I think it does not matter for side-wide plugins -- they do nothing with mainContextId -- they register their hooks in any case and do not check if they are enable for that context. 
And regarding loading plugins: 
mainContextId matters here:
https://github.com/pkp/pkp-lib/blob/main/classes/plugins/PluginRegistry.php#L112
then here:
https://github.com/pkp/pkp-lib/blob/main/classes/plugins/PluginRegistry.php#L216
then here:
https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPApplication.php#L427
i.e. https://github.com/pkp/pkp-lib/blob/main/classes/site/VersionDAO.php#L190
all plugins, also side-wide will be considered.
But, I am not 100% sure, maybe @asmecher will know better.

bozana: I'll double check the plugins regarding `$this->context = $journalDao->getById($mainContextId);` but I don't think so -- it could cause a problem earlier too...

Issue #8975, opened 2023-05-10
(OJS-3.4rc3) On click Add and edit templates (email) HTTP ERROR 500 appeared.


shantanu198713:
**OJS-3.4rc3
PHP-8.1.13**

I want to edit or view the email template, but after clicking on 'Add & Edit Template' HTTP ERROR 500 appeared.

![email error](https://github.com/pkp/pkp-lib/assets/108086810/7d1ba25d-edbc-49b9-b73d-e9534a80cc51)
![500 error](https://github.com/pkp/pkp-lib/assets/108086810/ce5d8530-fc2b-4106-85e3-5955d09a1d99)

**Error log**

`[10-May-2023 07:01:08 UTC] Exception: Plugin referral expected to inherit from ReferralPlugin, actual type NULL in F:\wamp64\www\ojs34rc3\lib\pkp\classes\plugins\PluginRegistry.php:202
Stack trace:
#0 F:\wamp64\www\ojs34rc3\lib\pkp\classes\plugins\PluginRegistry.php(218): PKP\plugins\PluginRegistry::_instantiatePlugin('generic', 'referral', 'ReferralPlugin')
#1 F:\wamp64\www\ojs34rc3\lib\pkp\classes\plugins\PluginRegistry.php(111): PKP\plugins\PluginRegistry::_loadFromDatabase('generic', NULL)
#2 F:\wamp64\www\ojs34rc3\lib\pkp\classes\core\Dispatcher.php(155): PKP\plugins\PluginRegistry::loadCategory('generic', true)
#3 F:\wamp64\www\ojs34rc3\lib\pkp\classes\core\PKPApplication.php(372): PKP\core\Dispatcher->dispatch(Object(APP\core\Request))
#4 F:\wamp64\www\ojs34rc3\index.php(21): PKP\core\PKPApplication->execute()
#5 {main}

[10-May-2023 07:01:08 UTC] Exception: Plugin coins expected to inherit from CoinsPlugin, actual type NULL in F:\wamp64\www\ojs34rc3\lib\pkp\classes\plugins\PluginRegistry.php:202
Stack trace:
#0 F:\wamp64\www\ojs34rc3\lib\pkp\classes\plugins\PluginRegistry.php(218): PKP\plugins\PluginRegistry::_instantiatePlugin('generic', 'coins', 'CoinsPlugin')
[...]

PR #8983, opened 2023-05-10
pkp/pkp-lib#8975 Fixed argument reflection handling

jonasraoni:


classes/mail/Mailable.php
```diff
@@ -451,23 +455,43 @@ public static function getDataDescriptions(): array
...
         }
 
         return $descriptions;
     }
 
+    /**
+     * Retrieve the list of type names that might compose a given type
+     * @return string[]
+     */
+    protected static function getTypeNames(ReflectionType $type): array
```
jonasraoni, line 481: This is supposed to retrieve all types that compose the argument, e.g.:
```php
Mailable::getTypeNames((new ReflectionFunction(fn (int|bool|(stdClass&Application) $a) => 0))->getParameters()[0]->getType());
```
Retrieves ['stdClass', 'APP\core\Application', 'int', 'bool']

classes/mail/Mailable.php
```diff
@@ -451,23 +455,43 @@ public static function getDataDescriptions(): array
...
+    {
+        if ($type instanceof ReflectionNamedType) {
+            return [$type->getName()];
+        }
+        if ($type instanceof ReflectionUnionType || $type instanceof ReflectionIntersectionType) {
```
Vitaliy-1, line 486: On the top of my head, Mailables can support intersection types but not union types, except cases, like in ORCID Plugin, where classes belong to different applications (but nevertheless, I'd check if it works as expected)

Constructor arguments are identified before Mailable instantiation to forecast what email template variables are going to be available when sending an email. E.g., say class1 is a associated with email template variable1 and class2 with variable2, then using the union
```php
public function __construct(class1|class2 $object)
```   
 would mean that both template variables, 1 and 2, are available, which isn't true.  

As a workaround we can ensure that both class names in the union don't exist in the application at the same time and exclude the one that isn't defined. Or forbid union types and deal with it on the plugin side. Or maybe there is a cleaner way...

Vitaliy-1 (changes requested):

classes/mail/Mailable.php
jonasraoni: Ops, I forgot to answer 😅

I thought about this case, and it will work out of the box for our main case `Server|Journal|Preprint` (only one of the classes will exist, and the `is_a()` will skip the other variants properly).

jonasraoni: For the other cases (where the class might actually exist), like let's say an email that might receive a `Failure|Success`... I think we have the following options:
- A. Build a separate email class for each variant
  This is the simplest for us and users, so that would be my choice. It will become unmaintainable just if we had many of such "intersection" variables in a single email, but then we can revisit this topic.
- B. Add "Smarty" capabilities to the template, so the users can setup loops/conditions/etc (`We {if Failure)failed{else}succeeded{/if}`)
  This would require adding a "view code" / building a powerful and intuitive editor 😅

jonasraoni: So, if you don't have anything against, I think it's ok to merge it this way, and use the option A 🤔

jonasraoni: And for future updates, I think that maybe it's better to have a method/interface, like errrrrr, `public static function getVariables()`, then we can call it and the class will still have a chance to customize its variables.

Vitaliy-1: @jonasraoni, thanks for double checking that it works. Can you also add an exception to unsure that both classes in the union don't exist at the same time? It's easy to forget it, especially for someone who haven't worked with mailables before.

Vitaliy-1 (approved):
Thanks, @jonasraoni. I left a comment but feel free to merge as is if you don't have a time for it right now, in this case please open an issue to not forget to return to it after the release.

classes/mail/Mailable.php
jonasraoni: When there's more than one active class in the union type, I've updated the code to retrieve only the first class and emit a warning.
Anyway, I think it's worth to open an issue to revisit how these variables are retrieved later.

PR #8990, opened 2023-05-11
Bugfix/main/8989 fix mariadb build

jonasraoni:


.php-cs-fixer.php
```diff
@@ -2,24 +2,16 @@
...
+    // The next two rules are enabled by default, kept for clarity
     ->ignoreDotFiles(true)
-    ->ignoreVCS(true);
+    ->ignoreVCS(true)
+    // The pattern is matched against each found filename, thus:
+    // - The "/" is needed to avoid having "vendor" match "Newsvendor.php"
+    // - The presence of "node_modules" here doesn't prevent the Finder from recursing on it, so we merge these paths below at the "exclude()"
+    ->notPath($ignoredDirectories = ['cypress/', 'js/', 'locale/', 'node_modules/', 'styles/', 'templates/', 'vendor/'])
+    // Ignore root based directories
+    ->exclude(array_merge($ignoredDirectories, ['dtd', 'lib', 'registry', 'schemas', 'xml']));
```
jonasraoni, line 14: The `notPath()` is called while the recursion is in progress, for each matched file. Which means, there's no way to break the loop (e.g. skip the whole `node_modules`).

composer.json
```diff
@@ -76,7 +77,11 @@
 			"PKP\\plugins\\": "plugins/",
 			"APP\\plugins\\": "../../plugins/",
 			"PKP\\jobs\\": "jobs/",
-			"APP\\jobs\\": "../../jobs/",
+			"APP\\jobs\\": "../../jobs/"
+		}
+	},
+	"autoload-dev": {
+		"psr-4": {
```
jonasraoni, line 84: The "tests" namespaces was introduced by me and aren't used within the application, so I moved them to the `autoload-dev`.

Issue #8999, opened 2023-05-15
Native XML plugin doesn't handle galleys properly


jonasraoni:
**Describe the bug**
After importing a submission with a galley, the galley is created, but no file is assigned to it.

**To Reproduce**
1. Import a submission that has a galley
2. Access it on the frontend
3. No galleys will be displayed, even though they can be seen at the dashboard

**What application are you using?**
OJS 3.4rc2

PR #9000, opened 2023-05-16
pkp/pkp-lib#8999 Fixed galley ID handling

jonasraoni:


plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.php
```diff
@@ -76,12 +76,11 @@ public function handleElement($node)
         $submission = $deployment->getSubmission();
         $context = $deployment->getContext();
         $stageName = $node->getAttribute('stage');
-        $submissionFileId = $node->getAttribute('id');
+        $submissionFileIdFromXml = $node->getAttribute('id');
```
jonasraoni, line 79: This variable got overwritten down below and messed up things even more.

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.php
```diff
@@ -234,13 +233,10 @@ public function handleElement($node)
             });
             $allRevisionIds = array_values($allRevisionIds);
 
-            $submissionFileId = $submissionFile->getId();
-
             foreach ($allRevisionIds as $i => $fileId) {
                 if ($i === 0) {
                     $submissionFile->setData('fileId', $fileId);
-                    $id = Repo::submissionFile()->add($submissionFile);
```
jonasraoni, line 242: This `$id` should be used below, but there was a typo.

Issue #7916, opened 2022-05-09
Replace HTMLPurifier with a maintained dependency


asmecher:
[HTMLPurifier](http://htmlpurifier.org) is not maintained anymore. Find another replacement toolset for "safe" HTML filtering.

It does not support HTML5: https://forum.pkp.sfu.ca/t/tinymce-stripping-anchor-tags-within-or-around-divs/72749/12

We currently use HTMLPurifier in the `PKPString::stripUnsafeHtml` function (lib/pkp/classes/core/PKPString.inc.php). This takes a configuration of allowed elements/attributes from `config.inc.php`.

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9257
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4025 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1581 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/696 [TEST ONLY]
[jatsTemplate](https://github.com/pkp/jatsTemplate) --> https://github.com/pkp/jatsTemplate/pull/30
[oaiJats](https://github.com/pkp/oaiJats) --> https://github.com/pkp/oaiJats/pull/36

**Update**
Based on following reasons we have decided to revert back the [Symfony HTML Sanitizer](https://github.com/symfony/html-sanitizer) and keep using the [HTMLPurifier](https://github.com/ezyang/htmlpurifier)
- Bug in the [Symfony html sanitizer ](https://github.com/symfony/html-sanitizer)
  - [Html Sanitizer] Sanitization remove single < (less than) char which is not part of any tag(https://github.com/symfony/symfony/issues/57597)
  - Bug actually in the <https://github.com/Masterminds/html5-php> parser
  - [Parser remove the single < (less than) character from given html string](https://github.com/Masterminds/html5-php/issues/250)
- Also the use of [strip_tags](https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPHtmlSanitizer.php#L78) will cause problem 
  - use of strip_tags omits anything after <=
  - we can overcome it by not using strip_tags and that better
[...]

PR #9002, opened 2023-05-16
pkp/pkp-lib#7916 Replaced html purifier with symfony html sanitizer

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#7916

### DIFF
--- a/classes/core/PKPString.php
+++ b/classes/core/PKPString.php
@@ -403,30 +403,21 @@ class PKPString
      * onclick(...) attributes, javascript: urls, and special characters.
      *
      * @param string $input input string
-     * @param string $configKey The config section key['allowed_html', 'allowed_title_html']
+     * @param string $key   The config section key['allowed_html', 'allowed_title_html']
      *
      * @return string
      */
-    public static function stripUnsafeHtml($input, $configKey = 'allowed_html')
+    public static function stripUnsafeHtml(string $input, string $key = 'allowed_html'): string
     {
-        // static $purifier;
-        // if (!isset($purifier)) {
-        //     $config = HTMLPurifier_Config::createDefault();
-        //     $config->set('Core.Encoding', 'utf-8');
-        //     $config->set('HTML.Doctype', 'HTML 4.01 Transitional');
-        //     $config->set('HTML.Allowed', Config::getVar('security', $configKey));
-        //     $config->set('Cache.SerializerPath', 'cache');
-        //     $purifier = new HTMLPurifier($config);
-        // }
-        // return $purifier->purify((string) $input);
-
         static $sanitizer;
+        static $configKey;
         static $allowedTagToAttributeMap;
 
-        if (!isset($htmlTagToAttributeMap)) {
-            Str::of(Config::getVar('security', $configKey))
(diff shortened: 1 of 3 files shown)


classes/core/PKPString.php
```diff
@@ -399,23 +402,68 @@ public static function getAmbiguousExtensionsMap()
...
     {
-        static $purifier;
-        if (!isset($purifier)) {
-            $config = HTMLPurifier_Config::createDefault();
-            $config->set('Core.Encoding', 'utf-8');
-            $config->set('HTML.Doctype', 'HTML 4.01 Transitional');
-            $config->set('HTML.Allowed', Config::getVar('security', $configKey));
-            $config->set('Cache.SerializerPath', 'cache');
-            $purifier = new HTMLPurifier($config);
+        if (!$input) {
```
asmecher, line 412: This is strictly supposed to deal with a `null` case, right? If so, then please use an exact test:
```php
if ($input === null) {
    return '';
}
```
Otherwise, non-string values will get coerced into strings by the function's type hint and the `if` will not behave like it looks at first glance.

classes/core/PKPString.php
```diff
@@ -399,23 +402,68 @@ public static function getAmbiguousExtensionsMap()
...
+            return '';
         }
-        return $purifier->purify((string) $input);
+
+        static $sanitizer;
+        static $configKey;
+        static $allowedTagToAttributeMap;
+
+        if ($configKey !== $key) {
```
asmecher, line 420: This is meant to cache repeat calls with the same `$configKey`, right? We only have two different configurations for `$configKey` we use in practice at the moment; at the moment flipping back and forth between two values of `$configKey` will cause the parsing to be repeated. I'd suggest storing the parsed data in an array by `$configKey`.

classes/core/PKPString.php
```diff
@@ -399,23 +402,68 @@ public static function getAmbiguousExtensionsMap()
...
+                                ->filter()
+                                ->toArray()
+                        ];
+                    }
+            
+                    return [];
+                });
+        }
+
+        if(!isset($sanitizer)) {
```
asmecher, line 447: This will instantiate a sanitizer for the first value of `$configKey`, and after that it will be re-used regardless of whether the `$configKey` matches.

asmecher (changes requested):
Thanks, @touhidurabir! I've given this a first review and added some comments. Could you move the work on the `php-jwt` update over to a separate PR? I've added some review comments separately on that to https://github.com/pkp/pkp-lib/issues/9110.

classes/core/PKPString.php
touhidurabir: Done .

classes/core/PKPString.php
touhidurabir: Moved both the attributes tag map and sanitizer as static cache as both needed at the end .

classes/core/PKPString.php
touhidurabir: Moved both the attributes tag map and sanitizer as static cache as both needed at the end .

PR #9004, opened 2023-05-17
Try fixing locale migration issue with es_AR

defstat:


jonasraoni:
@marcbria FYI this is related to that issue I've mentioned that could happen on 3.4 (e.g. failure to merge data when a journal is using variants of a "merged" locale, es_AR + es_MX + es_ES).

By the way, I've also raised the issue related to the submission language on Mattermost (that may lose information if we drop the country): 
image

marcbria:
Hello Jonas,

I have to admit that I'm a bit lost with all this and reading the code doesn't help me. :-/

Could you explain what the change implemented by defstat does?

Actually, I still don't understand why we use different criteria in the different languages that have dialectal variants (fr, pt, es) because common sense tells me that it would be a better idea to unify criteria.

I don't remember either why it is important to have one dialect above the rest (es_ES -> es) and why we can't continue to work with all dialects at the same level (es_ES, es_AR, es_MX...) as we did until 3.4.

The only thing I can say is that the strings of a dialect should never be transferred to others, because even if it is the same language there are very important stylistic issues that do not allow mixing. 

For example, a translation may choose to use the treatment of "tu" instead of "vos" or "usted", or not to split the gender (or do it from the article) and all that without going into vocabulary issues where regional terms can have very different meanings.

Although it makes a lot of sense to offer a translation es_ES (instead of en_US) in the defaultTranslations, as it will always be better to use another dialect of the same language than a totally different language, but what I would like to clarify is that, if we want translations with a minimum of quality, we cannot automatically transfer strings from one dialect to another.

@alec, if you didn't do it already in some issue, do you mind refreshing here your arguments for the merge of all the es_* into a single "es" and give answers to the questions that arise for us left in a thread.

Thank you all for your work.

jonasraoni:
@marcbria I think this is still a work in progress, so I've just tagged you to follow-up, there are many ways to solve it, so I'll be watching :)

marcbria:
Thanks Dimitris.

> merge only es_MX and es_ES into **es**

Please, don't do that. 

If for some reason we need a "es" tag, looks rasonable to assign this "alias" to the most complete translation... (at least as we take a final decision about this)... but for the reasons I explained before (and some others), strings should not be mixed between dialects.

asmecher:
Hi all -- I think this has gotten a little confused over the last months. We are not merging Spanish translations into each other; we are preserving `es_AR` (though nobody should be using it as it was never very complete and only available for 2.x) and `es_MX`. The `es_ES` translation is being renamed to `es`.

defstat:
Closing in favor of #9011

Issue #8964, opened 2023-05-04
Import/Export users is not working between same version of OJS 3.3.0-rc3


marcbria:
**Describe the bug**
Importation fails due key constrain and no feedback is sent to users.

Log throws following error:

```
[...]

PR #9006, opened 2023-05-19
pkp/pkp-lib#8964 Check user assign already to group on user import

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8964

### DIFF
--- a/classes/security/Role.php
+++ b/classes/security/Role.php
@@ -21,14 +21,14 @@ namespace PKP\security;
 class Role extends \PKP\core\DataObject
 {
     // ID codes and paths for all default roles
-    public const ROLE_ID_MANAGER = 0x00000010;
-    public const ROLE_ID_SITE_ADMIN = 0x00000001;
-    public const ROLE_ID_SUB_EDITOR = 0x00000011;
-    public const ROLE_ID_AUTHOR = 0x00010000;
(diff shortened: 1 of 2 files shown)


classes/security/Role.php
```diff
@@ -21,14 +21,14 @@
...
     // ID codes and paths for all default roles
-    public const ROLE_ID_MANAGER = 0x00000010;
-    public const ROLE_ID_SITE_ADMIN = 0x00000001;
-    public const ROLE_ID_SUB_EDITOR = 0x00000011;
-    public const ROLE_ID_AUTHOR = 0x00010000;
-    public const ROLE_ID_REVIEWER = 0x00001000;
-    public const ROLE_ID_ASSISTANT = 0x00001001;
-    public const ROLE_ID_READER = 0x00100000;
-    public const ROLE_ID_SUBSCRIPTION_MANAGER = 0x00200000;
+    public const ROLE_ID_MANAGER = 0x00000010;              // Decimal : 16
```
asmecher, line 24: @touhidurabir, feel free to just replace the hex constants with their decimal equivalents. This is a frequent point of frustration and there is no reason to have them in hex.

asmecher (changes requested):

classes/security/Role.php
touhidurabir: @asmecher updated .

PR #9007, opened 2023-05-19
pkp/pkp-lib#8993 null check applied on original publication for preview

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8993


jonasraoni (changes requested):
The fix should work, the `getTimeline()` has a method to cover the lack of a `dateStart`.

But considering there's no *published* publications, there won't be statistics to retrieve, so it's ok to just early return a "default" value.

touhidurabir:
> But considering there's no published publications, there won't be statistics to retrieve, so it's ok to just early return a "default" value.

@jonasraoni added default empty array return when no publication available .

jonasraoni (approved):

Issue #8985, opened 2023-05-10
[OJS 3.4 rc3] submission status when you cancel copyediting 


forgive38:
One minor thing, when you “Cancel copyediting” after accept submission, the status go back in Review step but the status of the submission is still “submission accepted”.
Maybe status should be the status of the last round of review ?
 (in the activity log : “##{$editorName} has sent this submission back from the copyediting stage.##” )

Thank you

touhidurabir:
@Vitaliy-1 PRs to review

[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9013
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3913

@Vitaliy-1 we still need to finalise few key behaviour .First of all , should we allow the status to show as `Return back to review.` or what is the current status of that round before moved to Copy Editing stage . for example , if no reviewer was assigned to round before moved to copy editing and then return back to review, it can show `Waiting for reviewers to be assigned.` which actually points the current status of the round . 

Also if we want to add a new status for which we need a new local key, may be we should wait for `3.4.0-1` to get this merged as for `3.4.0` translations are done .

Vitaliy-1:
> First of all , should we allow the status to show as Return back to review. or what is the current status of that round before moved to Copy Editing stage

Ideally, I believe we should show the status which was before the submission was moved to the next stage. Let's show `Returned back to review` if it can't be determined otherwise for now.  

> for example , if no reviewer was assigned to round before moved to copy editing and then return back to review, it can show Waiting for reviewers to be assigned

Yeah, as it's based on review assignments, it's possible.

> Also if we want to add a new status for which we need a new local key, may be we should wait for 3.4.0-1 to get this merged as for 3.4.0 translations are done

I think it's ok to introduce a new locale key here, it's better than leaving an incorrect status. 

Thanks, @touhidurabir for looking into this!

Vitaliy-1:
Thanks, merged!

PR #9013, opened 2023-05-22
pkp/pkp-lib#8985 added review status on back from copy editing

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8985

### DIFF
--- a/classes/submission/reviewRound/ReviewRound.php
+++ b/classes/submission/reviewRound/ReviewRound.php
@@ -57,6 +57,9 @@ class ReviewRound extends \PKP\core\DataObject
     // at least one revision file has been uploaded.
     public const REVIEW_ROUND_STATUS_RESUBMIT_FOR_REVIEW_SUBMITTED = 15;
 
+    // The following status is set when a submission return back from copyediting stage to last review round again
+    public const REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW = 16;
+
     //
     // Get/set methods
     //
@@ -151,6 +154,11 @@ class ReviewRound extends \PKP\core\DataObject
      */
     public function determineStatus()
     {
+        // The submission back form copy editing stage to last review round
+        if ($this->getStatus() == self::REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW) {
+            return self::REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW;
+        }
+
         // If revisions have been requested, check to see if any have been
         // submitted
         if ($this->getStatus() == self::REVIEW_ROUND_STATUS_REVISIONS_REQUESTED || $this->getStatus() == self::REVIEW_ROUND_STATUS_REVISIONS_SUBMITTED) {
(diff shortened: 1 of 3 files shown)


classes/submission/reviewRound/ReviewRound.php
```diff
@@ -151,6 +154,11 @@ public function setStatus($status)
      */
     public function determineStatus()
     {
+        // The submission back form copy editing stage to last review round
+        if ($this->getStatus() == self::REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW) {
+            return self::REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW;
+        }
+
```
touhidurabir, line 161: what can we do is to move this block in some later part of `determineStatus` function so that before showing status such as `Retune back to review`, it can calculate/determine other more important status such as `Resubmit for Review` or other more important action that require attention right before it passed to copy editing stage .

Vitaliy-1: Good idea. We can put it just before `REVIEW_ROUND_STATUS_REVIEWS_COMPLETED` so, at least, the round status based on review assignments will be shown.

classes/submission/reviewRound/ReviewRound.php
```diff
@@ -312,6 +320,8 @@ public function getStatusKey($isAuthor = false)
                 return 'editor.submission.roundStatus.recommendationsReady';
             case self::REVIEW_ROUND_STATUS_RECOMMENDATIONS_COMPLETED:
                 return 'editor.submission.roundStatus.recommendationsCompleted';
+            case self::REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW:
+                return 'editor.submission.roundStatus.returnToReview';
```
Vitaliy-1, line 324: As it's already completed action, let it be `editor.submission.roundStatus.returnedToReview`

locale/en/submission.po
```diff
@@ -1278,6 +1278,9 @@ msgstr "New editorial recommendations have been submitted."
 msgid "editor.submission.roundStatus.recommendationsCompleted"
 msgstr "All recommendations are in and a decision is needed."
 
+msgid "editor.submission.roundStatus.returnToReview"
+msgstr "Return back to review."
```
Vitaliy-1, line 1282: `Returned back to review.`

Vitaliy-1 (commented):
Thanks, @touhidurabir. Looks good! Just minor comments and it's ready to be merged

classes/submission/reviewRound/ReviewRound.php
touhidurabir: Done .

classes/submission/reviewRound/ReviewRound.php
touhidurabir: Done .

locale/en/submission.po
touhidurabir: Done .

Issue #8845, opened 2023-03-25
General code issues


jonasraoni:
**Describe the bug**
### pkp-lib
- `classes\metadata\MetadataRecord.php`
The methods `getMetadataSchema()` and `getApplicationEntityIdFromMetadataDescription()` are not declared at the class:
https://github.com/pkp/pkp-lib/blob/87e018ebe033574fa40749f0de110c5875f2f56b/classes/metadata/MetadataRecord.php#L46
https://github.com/pkp/pkp-lib/blob/87e018ebe033574fa40749f0de110c5875f2f56b/classes/metadata/MetadataRecord.php#L53

- `classes\notification\PKPNotificationOperationManager.php` and `classes\security\authorization\internal\ApiTokenDecodingMiddleware.php`
Given the `JWT::decode()` returns an object, it's curious to have it inside a `json_decode()` (there are other similar cases below).
https://github.com/pkp/pkp-lib/blob/87e018ebe033574fa40749f0de110c5875f2f56b/classes/notification/PKPNotificationOperationManager.php#L404
https://github.com/pkp/pkp-lib/blob/87e018ebe033574fa40749f0de110c5875f2f56b/classes/security/authorization/internal/ApiTokenDecodingMiddleware.php#L83-L84

### OJS
- `classes\payment\ojs\OJSPaymentManager.php` and `pages\issue\IssueHandler.php`
Calls to `getSetting()`/'updateSetting()` not available at this class:
https://github.com/pkp/ojs/blob/9fa385a2b795985f3b745577a6400414c89016c4/classes/payment/ojs/OJSPaymentManager.php#L233
https://github.com/pkp/ojs/blob/9fa385a2b795985f3b745577a6400414c89016c4/classes/payment/ojs/OJSPaymentManager.php#L245
https://github.com/pkp/ojs/blob/ecadfa2a654efe544d2c0d155936ee89ff5b72da/pages/issue/IssueHandler.php#L279
⚠ Before updating the code to `setData()`/`getData()`, it's probably needed to check if the field is part of the schema.

- `pages\article\ArticleHandler.php` and `pages\oai\OAIHandler.php`
JWT::decode() vs string
https://github.com/pkp/ojs/blob/9fa385a2b795985f3b745577a6400414c89016c4/pages/article/ArticleHandler.php#L81-L86
https://github.com/pkp/ojs/blob/9fa385a2b795985f3b745577a6400414c89016c4/pages/oai/OAIHandler.php#L68-L74

[...]

PR #9033, opened 2023-05-30
Bugfix/main/minor fixes

jonasraoni:


pages/reviewer/PKPReviewerHandler.php
```diff
@@ -205,13 +207,13 @@ public function getReviewForm(
...
+                return new PKPReviewerReviewStep1Form($request, $reviewSubmission, $reviewAssignment);
             case 2:
-                return new \PKP\submission\reviewer\form\PKPReviewerReviewStep2Form($request, $reviewSubmission, $reviewAssignment);
+                return new PKPReviewerReviewStep2Form($request, $reviewSubmission, $reviewAssignment);
             case 3:
-                return new \PKP\submission\reviewer\form\PKPReviewerReviewStep3Form($request, $reviewSubmission, $reviewAssignment);
+                return new PKPReviewerReviewStep3Form($request, $reviewSubmission, $reviewAssignment);
             default:
-                return null;
+                return throw new Exception("Invalid review form step {$step}");
```
asmecher, line 216: It does appear to work, but is `return throw` an intentional choice rather than just `throw`? The `return throw` syntax is not a common formulation in PHP.

classes/statistics/PKPStatisticsHelper.php
```diff
@@ -305,7 +305,7 @@ public function getInstitutionIds(int $contextId, string $ip, string $hashedIp,
 
         if ($flush) {
             // Salt and thus hashed IPs changed, empty the cache.
-            $this->institutionDataCacheMiss($this->institutionDataCache, 'ID');
+            $this->institutionDataCacheMiss($this->institutionDataCache);
```
asmecher, line 308: Merged this commit -- but heads-up @bozana just in case.

classes/statistics/TemporaryInstitutionsDAO.php
```diff
@@ -18,8 +18,9 @@
 
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
+use PKP\db\DAO;
 
-class TemporaryInstitutionsDAO
+class TemporaryInstitutionsDAO extends DAO
```
asmecher, line 23: @bozana, could you review this commit?

classes/announcement/DAO.php
```diff
@@ -116,14 +116,6 @@ public function getMany(Collector $query): LazyCollection
         });
     }
 
-    /**
-     * @copydoc EntityDAO::fromRow()
-     */
-    public function fromRow(object $row): Announcement
-    {
-        return parent::fromRow($row);
```
asmecher, line 124: This shouldn't be removed; it provides accurate type hinting for the return type.

classes/db/DAO.php
```diff
@@ -161,7 +161,7 @@ public function update($sql, $params = [], $callHooks = true, $dieOnError = true
             // this method is only called by a subclass. Results
             // in hook calls named e.g. "sessiondao::_updateobject"
             // (all lowercase)
-            $value = null;
```
asmecher, line 164: This is a long-standing behaviour dating back to our use of ADODB before we shifted the DB layer to Laravel, and it may or may not affect third-party plugins. This function is deprecated, so I'd rather leave it unaffected until it can be removed entirely.

classes/file/ContextFileManager.php
```diff
@@ -41,7 +41,7 @@ public function __construct($contextId)
     public function getBasePath()
     {
         $dirNames = Application::getFileDirectories();
-        return parent::getBasePath() . $dirNames['context'] . $this->contextId . '/';
+        return rtrim(parent::getBasePath(), '/\\') . $dirNames['context'] . $this->contextId . '/';
```
asmecher, line 44: This is more than a syntax change; can you file it separately?

classes/payment/PaymentManager.php
```diff
@@ -96,10 +92,10 @@ public function isConfigured()
     public function getPaymentForm($queuedPayment)
     {
         $paymentPlugin = $this->getPaymentPlugin();
-        if ($paymentPlugin !== null && $paymentPlugin->isConfigured($this->_context)) {
-            return $paymentPlugin->getPaymentForm($this->_context, $queuedPayment);
+        if (!$paymentPlugin?->isConfigured($this->_context)) {
+            throw new Exception('Payment not configured');
```
asmecher, line 96: Was there a friendly user-facing error than the default exception handler prior to this change?

classes/query/QueryDAO.php
```diff
@@ -138,7 +138,8 @@ public function countOpenPerStage(int $submissionId, ?array $participantIds = nu
             ->toArray();
 
         return collect(Application::get()->getApplicationStages())
-            ->mapWithKeys(fn ($stageId, $key) => [$stageId => $counts[$stageId] ?? 0]);
+            ->mapWithKeys(fn ($stageId, $key) => [$stageId => $counts[$stageId] ?? 0])
+            ->toArray();
```
asmecher, line 142: Is this strictly to satisfy the documented return type? If so, I'd rather not merge changes like this so late in the 3.4.0 release. Once we have a `stable-3_4_0` branch distinct from `main`, we can start adding explicit type declarations to `main` and scaring up any bugs that get introduced early in the dev cycle.

classes/section/Collector.php
```diff
@@ -166,7 +166,7 @@ public function getQueryBuilder(): Builder
                     $qb->select('p.*')
                         ->from('publications AS p')
                         ->whereNotNull('p.' . $this->dao->primaryKeyColumn)
-                        ->where(DB::raw('p.' . $this->dao->primaryKeyColumn, '=', 's.' . $this->dao->primaryKeyColumn))
```
asmecher, line 169: Was this actually not working in the previous form?

controllers/grid/eventLog/EventLogGridCellProvider.php
```diff
@@ -87,7 +87,7 @@ public function getTemplateVarsFromRowColumn($row, $column)
                             if (isset($params['reviewAssignmentId'])) {
                                 $reviewAssignment = $reviewAssignmentDao->getById($params['reviewAssignmentId']);
                                 if ($reviewAssignment && $reviewAssignment->getReviewMethod() === ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN) {
-                                    $userName = $reviewAssignment->getUserFullName();
+                                    $userName = $element->getUserFullName();
```
asmecher, line 90: @Vitaliy-1, can you review this change?

controllers/grid/files/form/ManageSubmissionFilesForm.php
```diff
@@ -101,11 +101,12 @@ public function execute($stageSubmissionFiles = null, $fileStage = null, ...$fun
             if ($this->fileExistsInStage($submissionFile, $stageSubmissionFiles, $fileStage)) {
                 // ...update the "viewable" flag accordingly.
                 if ($isViewable != $submissionFile->getData('viewable')) {
-                    $submissionFile = Repo::submissionFile()
+                    Repo::submissionFile()
                         ->edit(
                             $submissionFile,
                             ['viewable' => $isViewable]
                         );
+                    $submissionFile = Repo::submissionFile()->get($submissionFile->getId());
```
asmecher, line 109: @jonasraoni, can you file this separately with details?

classes/section/Collector.php
jonasraoni: The sample below outputs `p.x`:
```php
$primaryKeyColumn = 'x';
echo DB::raw('p.' . $primaryKeyColumn, '=', 's.' . $primaryKeyColumn); 
```

classes/controllers/grid/feature/selectableItems/SelectableItemsFeature.php
```diff
@@ -54,7 +54,8 @@ public function getInitializedRowInstance($args)
 
         if ($grid instanceof \PKP\controllers\grid\CategoryGridHandler) {
             $categoryId = $grid->getCurrentCategoryId();
-            $row->addFlag('selected', (bool) $grid->isDataElementInCategorySelected($categoryId, $row->getData()));
+            $data = $row->getData();
```
asmecher, line 57: `$row->getData()` returns a reference; is this change necessary?

jonasraoni: This is the signature of the `isDataElementInCategorySelected`:
```php
public function isDataElementInCategorySelected($categoryId, &$gridDataElement){}
```

So it requires a reference.

asmecher:
@jonasraoni, I've started going through this but I'm still finding it hard to review as it mixes small-but-important changes in with uncontroversial changes (e.g. spelling corrections). I'd suggest two next steps:
- For the very uncontroversial changes, please just go ahead and cherry-pick them into `main`. I've gone ahead and done this already with a few of them. But please just keep it the trivial ones.
- For issues that may cause problems if left unresolved in 3.4.0 (e.g. https://github.com/pkp/pkp-lib/commit/96b8b8e37bf197db6051d62700da2c63791f0de9), please merge them directly if they are trivial and did not exist in 3.3.0, and break them out separately into their own issues if they introduce any risk or were present in 3.3.0 already.
- Leave the rest in the open PRs and we can consider them post-3.4.0.

classes/payment/PaymentManager.php
jonasraoni: AFAICR due to the lack of `null` (`false`) checks, it was just going to fail with a fatal error:
```php
$paymentForm = $paymentManager->getPaymentForm($queuedPayment);
$paymentForm->display($this->request);
```

classes/controllers/grid/feature/selectableItems/SelectableItemsFeature.php
asmecher: Yes, and `GridRow::getData` returns a reference:

https://github.com/pkp/pkp-lib/blob/96b8b8e37bf197db6051d62700da2c63791f0de9/classes/controllers/grid/GridRow.php#L126

classes/file/ContextFileManager.php
jonasraoni: Given that the `getBasePath()` is basically a `Config::getVar('files', 'files_dir')`, I imagine that a `files_dir` ending with `/` would lead to double bars `//`, but I didn't test tough.

classes/query/QueryDAO.php
asmecher: I'd rather we just left it as is, since it's already deprecated.

classes/db/DAO.php
jonasraoni: No complaints here, I just wanted to match the expected return :)

pages/reviewer/PKPReviewerHandler.php
jonasraoni: Typo :)

classes/file/ContextFileManager.php
asmecher: And I think I can recall seeing cases where `//` exists in practice -- OSs tend to be forgiving in this case, which is why it hasn't been fixed yet. But I think this should be kept back for the 3.5.0 cycle.

classes/payment/PaymentManager.php
asmecher: If I recall the payment code generally checks the configured status before calling functions like this in the first place, so any error handling would have already happened. But neither of us is sure, and the payment aspects of the software already went through the community testing process and are a little finicky to set up for testing, so I'd rather leave this for now.

classes/announcement/DAO.php
jonasraoni: While I was adding type hints, I've added some harmless "generics" to the code:
image

image

So the proper type hint should be covered by this:
image

classes/controllers/grid/feature/selectableItems/SelectableItemsFeature.php
jonasraoni: Hmm, so it must be ok, I'll double check later 👌

jonasraoni:
@asmecher I still didn't review these PRs, I've just shared to let you take a first-look 😁
I'll take a look after I clear the migration issue.

classes/section/Collector.php
asmecher: Just wondering whether an important scenario wasn't tested. This is [only used in OMP](https://github.com/pkp/omp/blob/main/pages/catalog/CatalogHandler.php#L102), and it would be complicated or impossible to get the wrong data to show up due to this. Cherry-picked into `main`, thanks!

jonasraoni:
After the latest type hints, I think the code must be almost ready to run with the `[general].strict = On`, at least on the build. But I remember some parts still need to be fixed (e.g. non-namespaced class names inside strings).

classes/announcement/DAO.php
asmecher: If it makes the IDEs used by the dev team happy, then I'm all for removing the subclass functions. But this is great stuff to change at the start of a dev cycle, and to bring to the team as a suggested change in practice so we're all aware of it. Any changes going in now should be limited to bug fixes until we get `stable-3_4_0` branched from `main`.

asmecher:
I'd like to set aside a bit of time after 3.4.0 is released to work through some of the strict mode warnings/errors, but I'm pretty sure strict mode is going to be broken until after the next LTS release (looking like 3.5.0 at this point), when we can formally remove some of the backwards compatibility stuff that was introduced with this cycle.

classes/payment/PaymentManager.php
jonasraoni: No problem! I can move this, together with the other things that I didn't push, to the post-3.4 era.

classes/statistics/TemporaryInstitutionsDAO.php
bozana: Hmmm... All those DAOs in classes/statistics/ do actually not need to extend DAO -- are a little bit different than other objects DAOs -- but I suppose there is no harm this way... :thinking:

jonasraoni: My reasoning here was the following: If all DAOs are extending this base class, so I think it's better to keep the pattern, it's not going to harm, and might be useful somewhere (e.g. `$dao instanceof DAO` ? a : b`).

In another PR I was prototyping something that would require such check 🤔

classes/statistics/PKPStatisticsHelper.php
bozana: All fine! :+1: Thanks!

classes/db/DAO.php
```diff
@@ -330,7 +330,7 @@ public function convertFromDB($value, $type, $nullable = false)
                 return (float) $value;
             case 'object':
             case 'array':
-                $decodedValue = json_decode($value, true);
+                $decodedValue = json_decode($value ?? ($type === 'array' ? '[]' : '{}'), true);
```
jonasraoni, line 333: AFAICR I got this from a warning in a user log.

A not null field is attempting to convert a null value. I thought it's better to return an empty array/object than `false`.

jonasraoni:
@asmecher Done! I've created some new issues. From this PR the only item that might be worth to review is the one below, the rest + things that I didn't push, I'll leave for post-3.4:
https://github.com/pkp/pkp-lib/pull/9033/files#diff-baca326285c1780182c7a44aae2ef03a56e8dc332efbf90fe66acfa8bc11b636R333

As a reminder, the other issues that I've met before must be unsolved yet, see: https://github.com/pkp/pkp-lib/issues/8845 (as I wrote in the description, the PRs on the issue **are not** covering anything that I wrote in the issue description).

controllers/grid/eventLog/EventLogGridCellProvider.php
Vitaliy-1: Oops, forgot to remove deprecated methods, working on it: https://github.com/pkp/pkp-lib/issues/9067

classes/db/DAO.php
```diff
@@ -330,10 +330,10 @@ public function convertFromDB($value, $type, $nullable = false)
...
                 // FIXME: pkp/pkp-lib#6250 Remove after 3.3.x upgrade code is removed (see also pkp/pkp-lib#5772)
                 if (!is_null($decodedValue)) {
-                    return $decodedValue;
+                    return $type === 'array' ? array_values((array) $decodedValue) : (object) $decodedValue;
```
jonasraoni, line 336: For here and below, the database might be holding a valid JSON, with an unexpected value.
For example, if we encode the array `[1 => 0]` using the `$type` `array`, it will be stored as a JSON object.

There are already some known cases out in the wild.

classes/statistics/TemporaryInstitutionsDAO.php
asmecher: I have no problem with extending the DAOs as a general principle, but we don't have a pressing need to do it for 3.4.0 and I'd rather not make speculative changes just before the release. Can we leave this for `main` or at least `stable-3_4_0` post 3.4.0-0?

jonasraoni: Yep, this is definitely not a requirement.

classes/user/DAO.php
```diff
@@ -80,6 +80,8 @@ class DAO extends EntityDAO
 
     /**
      * Construct a new User object.
+     *
+     * @return User
```
asmecher, line 84: Since I think this PR is most suitable for `main`, I'd rather we just started adding first-class type hints.

asmecher (changes requested):
Just a small tweak, and if you're satisfied that this is ready to go, you can merge to `main`. I don't think this needs to be brought into `stable-3_4_0`.

PR #8970, opened 2023-05-09
Bugfix/main/8929 migration issues

jonasraoni:

Issue #8333, opened 2022-10-14
Fix remaining missing foreign key declarations


asmecher:
This issue is a continuation of https://github.com/pkp/pkp-lib/issues/6093. 

The remaining missing foreign key constraints will require some code changes to move away from the use of `0` as a reserved case.

- `lib/pkp/classes/migration/install/CommonMigration.php`:
  - [x]  `site.redirect` (renamed to redirect_context_id)
  - [x]  `plugin_settings.context_id` (uses `0` for site-wide)
  - [x] `notifications.context_id` (uses `0` for site-wide)
  - [x] `notifications.user_id` (uses `0` for "anyone" IIRC)
  - [x] `notification_subscription_settings` (just needs the field to be renamed from `context` to `context_id`)
- `lib/pkp/classes/migration/install/LogMigration.php`:
  - [x] `email_log.sender_id` (uses `0` for "nobody in particular")
- `lib/pkp/classes/migration/install/MetadataMigration.php`:
  - [x] `filters.context_id` (uses `0` for site-wide)
  - [x] `filters.parent_filter_id` (uses `0` for none)
- `lib/pkp/classes/migration/install/NavigationMenusMigration.php`:
  - [x] `navigation_menus.context_id` (`0` used for site-wide)
  - [x] `navigation_menu_items.context_id` (`0` used for site-wide)
  - [x] `navigation_menu_items_assignments.parent_id` (`0` apparently used for none, even though the field appears nullable)
- `lib/pkp/classes/migration/install/classes/RolesAndUserGroupsMigration.php`:
  - [x] `user_groups.context_id` (`0` used for side-wide)
notification_subscription_settings

Some of these areas are slated for rewriting, so this work may be a lower priority than the rewrite.

It makes sense to keep the `email_log` even when a sender (user) is deleted, but it's probably needed to review GDPR (e.g. right to be forgotten), as the email might keep personal data.

jonasraoni:
I've updated the issue description (added the missing `user_groups.context_id` and marked others as done), and created initial PRs from detached commits of another issue.

jonasraoni:
Added the `site.redirect` field, I think it's also better to get it renamed to `redirect_context_id`.

asmecher:
@jonasraoni, can you dust off any remaining PRs here and summarize?

jonasraoni:
Yep, I've got a couple of PRs to dust off, I'll add this one to the list 😁

asmecher:
[...]

PR #9071, opened 2023-06-06
Bugfix/main/9069 migration issues

jonasraoni:


classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -31,745 +34,1135 @@ abstract protected function getContextTable(): string;
...
+            $affectedRows += $this->deleteRequiredReference('publications', 'submission_id', 'submissions', 'submission_id');
+            $affectedRows += $this->cleanOptionalReference('publications', 'primary_contact_id', 'authors', 'author_id');
+            return $affectedRows;
+        });
+
+        $this->addTableProcessor('user_groups', function (): int {
+            $affectedRows = 0;
+            // Depends directly on ~1 entities: context_id->context_table.context_id
+            // Custom field (not found in at least one of the softwares)
+            $affectedRows += $this->cleanOptionalReference('user_groups', 'context_id', $this->getContextTable(), $this->getContextKeyField(), $this->ignoreZero('context_id'));
```
asmecher, line 498: I think we [still have this problem](https://github.com/pkp/pkp-lib/pull/8970#discussion_r1207505674) with the preflight check trying to update `context_id` to `null`. I get:
```
ERROR: Upgrade failed: DB: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'context_id' cannot be null (SQL: update `user_groups` set `context_id` = ? where `context_id` in (180))
```
This is a case where an entry in `journals` was deleted (`journal_id = 180`) and data in `user_groups` was not cleaned up.

Inspecting the code, it looks like there are a couple of problems:
- The `cleanOptionalReference` function expects to set missing references to `null`, but in this case that will generate bad data. `context_id` should only be `null` in `user_groups` for the administrator role, but this will result in numerous other entries getting set to `null` where their respective `journal_id` entries are missing.
- The `context_id` column will not be nullable anyway at this point in the upgrade, I don't think.

jonasraoni: That's right! I thought that I kept all changes related to the extra FKs on two separated commits, and overlooked the rebased modifications. I'll regenerate the automated code using a fresh database and do a diff, it should be enough to capture the divergences.

jonasraoni: I've updated the PR. After regenerating the cleanup code, I flagged one more divergence related to recent changes on the nullability of the `event_log.user_id`

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -31,745 +34,1137 @@ abstract protected function getContextTable(): string;
...
-            // Clean orphaned submission_comments entries by author_id
-            $orphanedIds = DB::table('submission_comments AS sc')->leftJoin('users AS u', 'u.user_id', '=', 'sc.author_id')->whereNull('u.user_id')->distinct()->pluck('sc.author_id');
-            foreach ($orphanedIds as $userId) {
-                $this->_installer->log("Removing orphaned submission_comments entry for missing author_id {$userId}");
-                DB::table('submission_comments')->where('author_id', '=', $userId)->delete();
-            }
+        $this->addTableProcessor('email_log', function (): int {
+            $affectedRows = 0;
+            // Depends directly on ~1 entities: sender_id->users.user_id
+            $affectedRows += $this->deleteRequiredReference('email_log', 'sender_id', 'users', 'user_id', $this->ignoreZero('sender_id'));
```
asmecher, line 625: The `email_log.sender_id` relationship does not have a foreign key constraint (that's filed for future work in https://github.com/pkp/pkp-lib/issues/8333), and we should not be removing email log entries when a user account is removed. If anything, the missing entries (when #8333 is implemented) should be set to `null` and the email log should be otherwise preserved. Removing a user shouldn't destroy evidence of emails sent on a submission.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -31,745 +34,1137 @@ abstract protected function getContextTable(): string;
...
-                $this->_installer->log("Removing orphaned settings for missing category ID {$categoryId}");
-                DB::table('category_settings')->where('category_id', '=', $categoryId)->delete();
-            }
+        $this->addTableProcessor('navigation_menu_item_assignments', function (): int {
+            $affectedRows = 0;
+            // Depends directly on ~3 entities: navigation_menu_id->navigation_menus.navigation_menu_id navigation_menu_item_id->navigation_menu_items.navigation_menu_item_id parent_id->navigation_menu_item_assignments.navigation_menu_item_assignment_id
+            $affectedRows += $this->deleteRequiredReference('navigation_menu_item_assignments', 'navigation_menu_item_id', 'navigation_menu_items', 'navigation_menu_item_id');
+            $affectedRows += $this->deleteRequiredReference('navigation_menu_item_assignments', 'navigation_menu_id', 'navigation_menus', 'navigation_menu_id');
+            $affectedRows += $this->deleteOptionalReference('navigation_menu_item_assignments', 'parent_id', 'navigation_menu_item_assignments', 'navigation_menu_item_assignment_id', $this->ignoreZero('parent_id'));
```
asmecher, line 579: I think this is the wrong foreign key reference -- `parent_id` refers to an entry in `navigation_menu_items`, not `navigation_menu_item_assignments`. Also, this FK constraint is slated for later (https://github.com/pkp/pkp-lib/issues/8333) so cleanup shouldn't be done now -- it will just need to be done again, before the FK constraint is added.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
jonasraoni: I generally agree, I just don't know about the GDPR stuff (right to be forgotten) 🤔

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
jonasraoni: Yep, the cleanup will have to be redone.
I decided to preserve the cleanups for the future foreign keys, but I can drop all of them.

asmecher: OK, but the more critical issue is that this is checking the wrong table, and deleting entries that are not invalid.

jonasraoni: Definitely yes, the automation is "right", the problem came from here (blindly thought that `parent_id` referred to itself):
https://github.com/pkp/pkp-lib/pull/9076/commits/a8154863bec82d7e06e9980bbdacb85546a186d0#diff-8723a186af1085addb70eb012702a54559d8b36a3a2def7900edbf3f80512e96R80

jonasraoni: These are the "extra" cleanups:
user_groups.context_id
filters.parent_filter_id
filters.context_id
navigation_menu_item_assignments.parent_id
navigation_menu_items.context_id
email_log.sender_id
navigation_menus.context_id
plugin_settings.context_id

PR #9076, opened 2023-06-07
Feature/main/8333 Add remaining FKs

jonasraoni:


classes/migration/install/NavigationMenusMigration.php
```diff
@@ -72,6 +77,8 @@ public function up(): void
             $table->index(['navigation_menu_item_id'], 'navigation_menu_item_assignments_navigation_menu_item_id');
 
             $table->bigInteger('parent_id')->nullable();
+            $table->foreign('parent_id', 'navigation_menu_item_assignments_parent_id')->references('navigation_menu_item_assignment_id')->on('navigation_menu_item_assignments')->onDelete('cascade');
```
jonasraoni, line 80: Invalid reference, `parent_id` refers in fact to the `navigation_menu_items`, it's needed to fix here and on the migration.

classes/user/Collector.php
```diff
@@ -455,7 +455,7 @@ protected function buildUserGroupFilter(Builder $query): self
                 )
                 ->when($this->excludeRoles !== null, fn ($query) => $query->whereNotIn('ug.role_id', $this->excludeRoles))
                 ->when($this->roleIds !== null, fn ($query) => $query->whereIn('ug.role_id', $this->roleIds))
-                ->when($this->contextIds !== null, fn ($query) => $query->whereIn('ug.context_id', $this->contextIds))
+                ->when($this->contextIds !== null, fn ($query) => $query->whereIn(DB::raw('COALESCE(ug.context_id, 0)'), $this->contextIds))
```
asmecher, line 458: We're going to run into a collision here between the conventions we use for "this applies to the site level rather than the journal level" (for which we previously used `CONTEXT_ID_NONE` but are changing to `null`) and "there is no filter on this condition", for which in the new DAO/Collector patterns we use `null`.

What you're implying in this change is that `$this->contextIds` should be `null` when no filter is set, and should be an array containing `0` when we want to query site-wide user groups (e.g. the administrator user group). (This is workable, though I'd prefer an array containing `null`, as set out below.)

However, in other cases that don't use the new query builder toolset, you're moving away from `CONTEXT_ID_NONE` and instead preferring `null` directly (e.g. [this change](https://github.com/pkp/pkp-lib/pull/9076/files#diff-5d0146e413cc69e673f0e04a928ef6ac5eadac70bd11dd9af3c0e45f89ff4826R74)).

This makes it ambiguous whether coders should use `null` or `CONTEXT_ID_NONE`. As we convert more DAOs over to the `Collector` pattern it'll get muddier.

Do you think it's feasible to code consistently around `null`s rather than a mix of `null` and `0`, and change the `CONTEXT_ID_NONE` constant so that it's synonymous with `null`?

This will also reduce the use of `COALESCE` in the SQL, which risks proliferating and might have repercussions on performance if they start to spread into complex queries.

classes/userGroup/UserGroup.php
```diff
@@ -70,11 +70,11 @@ public function getContextId()
     /**
      * Set the context ID
      *
-     * @param int $contextId
+     * @param ?int $contextId
```
asmecher, line 73: Since we're early in the 3.4.0 dev cycle, I'd suggest just making this a first-class type hint. (Also elsewhere)

classes/userGroup/UserGroup.php
```diff
@@ -70,11 +70,11 @@ public function getContextId()
...
      */
     public function setContextId($contextId)
     {
-        $this->setData('contextId', $contextId);
+        $this->setData('contextId', (int) $contextId ?: null);
```
asmecher, line 77: I'd prefer if we didn't add extra casts etc. throughout the codebase (unless they are useful for backwards compatibility and explicitly flagged as such, for later removal). Casts like this aren't consistently used and they cause values to flip types unpredictably in places where you think you're just calling a getter/setter. It would be better if we adopted good first-class type hints throughout, and completely removed cases where there were special values, e.g. a special `0` (as noted above); then, if a `0` somehow snuck into the system (e.g. due to an out of date plugin or something), it would immediately trigger a referential integrity error. That would be quick to fix and not leave a trail of ambiguity behind it.

asmecher (changes requested):
Thanks, @jonasraoni! Some work still to be done on the relationship between `null` and `0` on context IDs; see comments.

classes/context/ContextDAO.php
```diff
@@ -192,7 +192,7 @@ public function resequence()
         $result = $this->retrieve('SELECT ' . $this->primaryKeyColumn . ' AS context_id FROM ' . $this->tableName . ' ORDER BY seq');
         $i = 1;
         for ($i = 1; $row = (array) $result->current(); $i += 2 && $result->next()) {
-            $this->update('UPDATE ' . $this->tableName . ' SET seq = ? WHERE ' . $this->primaryKeyColumn . ' = ?', [$i, $row['context_id']]);
+            $this->update('UPDATE ' . $this->tableName . ' SET seq = ? WHERE COALESCE(' . $this->primaryKeyColumn . ', 0) = ?', [$i, (int) $row['context_id']]);
```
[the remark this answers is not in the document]
jonasraoni: Thanks for the initial review! I'll double check the changes.
I've just added coalesces/value coercions everywhere that made minimal sense, to include it on the 3.4 release 😁

Now that we have time, it's better to fix it properly and get rid of the `0`s.

Issue #9084, opened 2023-06-10
HTML should be stripped from submission titles when used in email titles


asmecher:
**Describe the bug**
When a submission with HTML in the title contains HTML formatting, the submission acknowledgement email passes the HTML through in the title:
![image](https://github.com/pkp/pkp-lib/assets/200411/4a0d6dc1-8e37-4cdf-909d-a4bcc3de9108)

**What application are you using?**
 OJS, OMP or OPS version 3.4.0-0

touhidurabir:
@asmecher do we plan to change it and only send the title as non-html text now ? If i remember correctly, when we implemented the pkp/pkp-lib#2564, we decided to have the localized html title as submission title . Also in any modern mail client, it should be presented nicely . which mail client it causing this issue ? see below example screenshot 

<img width="1397" alt="CleanShot 2023-06-16 at 15 39 43@2x" src="https://github.com/pkp/pkp-lib/assets/2587979/3a98a562-5d7b-48d6-99c6-0900ccdae68f">

it can be reverted easily by changing line at https://github.com/pkp/pkp-lib/blob/main/classes/mail/variables/SubmissionEmailVariable.php#L83

asmecher:
@touhidurabir, it's the email title, not the submission title, that's causing problems. When the submission title contains HTML it's presented OK in the body of the email, but not in the title of the email (the subject). The screenshot at the top is taken from the list of messages in the inbox in my email client, not the body of a single email.

touhidurabir:
> it's the email title, not the submission title, that's causing problems

@asmecher  got it . 

@Vitaliy-1 can you review the PRs 
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9101
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3949 [TEST ONLY]

This is only for `stable-3_4_0` branch for now . if all ok, I will forward port it to `main` .

asmecher:
Hi @touhidurabir and @Vitaliy-1! I've taken the liberty of reviewing, testing, and merging this to get it into the 3.4.0-1 build. Looks great, @touhidurabir!

PR #9101, opened 2023-06-16
pkp/pkp-lib#9084 Mailable subject construction removing any html chars

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#9084


asmecher:
Manually cherry-picked.

Issue #7432, opened 2021-11-05
Break tags are visible in sent emails when editorial signature variable is added to templates


kshuttle:
**Describe the bug**
Break tags [< br/ >] are visible in sent emails when the {$editorialContactSignature} variable is added to templates.

Email templates that already have the {$editorialContactSignature} variable there by default don't seem to have this problem. It is happening when a user edits an email template and adds the variable in order to include the editor's signature at the end. 

**To Reproduce**
1. Edit the "Resubmit for Review" email template so it includes the {$editorialContactSignature}
2. Find an article in the review stage, select "Request Revisions," and select the option for another review round so that the "Resubmit for Review" email template is populated in the preview
3. Note that the signature variable populates as follows: Editor's Name<br/>editor@email.com
4. Send the email to the author and notice that the email they receive has the break tag in the signature, as above

<img width="585" alt="Revise-resubmit-break-tag-preview-and-email" src="https://user-images.githubusercontent.com/59741486/140440599-e39c4092-3596-41dc-a73a-760fe5f453cc.png">

Note that this issue does not occur when sending emails that already have the {$editorialContactSignature} in the default text.
1. Find an article in the review stage with an overdue review, and click "Send reminder" to a reviewer
2. Notice that the reminder email preview shows the signature variable with the break tag visible
3. Send the email and notice that the email received by the reviewer has the signature variable populated correctly (without the break tag)

<img width="590" alt="Review-remind-break-tag-preview-only" src="https://user-images.githubusercontent.com/59741486/140440676-e796e629-75f8-444b-9b83-3fc4c26a3bd0.png">

<img width="747" alt="signature-correct-review-remind-email" src="https://user-images.githubusercontent.com/59741486/140440732-0a5638f8-4a0c-426a-a528-3faff4767c06.png">

1. Find an article in the review stage and add a new reviewer
2. Notice that the review invitation email preview shows the signature variable populated correctly (without the break tag), and it stays this way when the email is received
![image](https://user-images.githubusercontent.com/59741486/140440838-6df41d0b-9c48-4c2b-b529-a4fa8f2651fb.png)

**What application are you using?**
A user reported this problem in OJS 3.3, and I was able to reproduce the problem in OJS 3.2.

harshakokel:
Hello, 

Is there a fix for this? We are facing the same issue on the [JAIR website](https://www.jair.org/) that recently upgraded to OJS 3.3. 

Regards,
hkokel@utdallas.edu

[...]

PR #9103, opened 2023-06-17
pkp/pkp-lib#7432 Allow breaks in editorial contact signature variable

asmecher:


jardakotesovec:
@asmecher Looks good to me.

My understanding is that the reason why its happening only for this particular screen is that the other ones do some of the variable replacements on server side using `$template->replaceParams();`

So basically its relying on these tinyMCE helpers to the replacement in this particular case. So handling it there make sense.

And I see that this screen is completely overhauled with Vue.js for 3.4, where the variable replacement is happening within FieldPreparedContent.vue component.

Also tried your changes and checked both in UI and in resulting email and looks ok.

PR #9108, opened 2023-06-21
pkp/pkp-lib#5953 email variable locale update to support new variable addition

touhidurabir:
for pkp/pkp-lib#5953

### DIFF
--- a/locale/en/manager.po
+++ b/locale/en/manager.po
@@ -2990,6 +2990,9 @@ msgstr "Issue identification, including volume, number, year, and title"
 msgid "emailTemplate.variable.issue.issuePublishedUrl"
 msgstr "The URL to the published issue"
 
+msgid "emailTemplate.variable.issue.issueTableOfContent"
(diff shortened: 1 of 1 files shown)


locale/en/manager.po
```diff
@@ -2990,6 +2990,9 @@ msgstr "Issue identification, including volume, number, year, and title"
 msgid "emailTemplate.variable.issue.issuePublishedUrl"
 msgstr "The URL to the published issue"
 
+msgid "emailTemplate.variable.issue.issueTableOfContent"
+msgstr "The issue table of content such as list of all associated submissions."
```
asmecher, line 2994: Does this need to go into the pkp-lib locale files? The rest of the feature is OJS-only, so the locale keys should stay there too.

asmecher (changes requested):

locale/en/manager.po
touhidurabir: Right, moved to OJS app locale files .

Issue #9073, opened 2023-06-07
Update laravel version from 9.x to 10.x


touhidurabir:
**Describe the Proposal**
We need to update the current laravel version from 9.x to 10.x which will make the PHP version requirement 8.1 or higher . 

**Additional Information**
As several issues(such as pkp/pkp-lib#7916, pkp/pkp-lib#7698) for future version require or may require PHP 8.1+, it will be better to update the laravel version to 10.x which required PHP 8.1+ . 

**Upgrade Doc/Resource**
Laravel Upgrade Doc : https://laravel.com/docs/10.x/upgrade#updating-dependencies

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9115
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3954 [TEST ONLY]

asmecher:
Resolved via https://github.com/pkp/pkp-lib/pull/9128!

PR #9115, opened 2023-06-27
pkp/pkp-lib#9073 laravel update from 9.x to 10.x

touhidurabir:
for pkp/pkp-lib#9073


asmecher:
@touhidurabir, I think you can get rid of all the extensions and keep it simple if you cover a second `DB::raw` use case in `DAO.php`; see: https://github.com/pkp/pkp-lib/pull/9128/files for details.

Issue #9127, opened 2023-06-29
Users assigned to submissions after changing their roles lose access to files to which they should have it


Vitaliy-1:
**Describe the bug**
Described in this [comment](https://github.com/pkp/pkp-lib/issues/9104#issuecomment-1612043621)
If user was assigned to a submission in one role and then the role has changed, `SubmissionFileAccessPolicy` won't grant access to the file even if new role should allow the access. E.g., user was assigned as a journal editor and then demoted to a section editor, or assigned as designer and promoted to the section editor.

Such behaviour was introduced by this [commit](https://github.com/NateWr/pkp-lib/commit/58ec77bd3559f53b294ae515227e5718003ef59f) as part of [this issue](https://github.com/pkp/pkp-lib/issues/3666) aimed to prevent double-assigned editors from accessing unauthorized files.

**To Reproduce**
See: https://github.com/pkp/pkp-lib/issues/9104#issuecomment-1612043621
In this case this policy is failing: https://github.com/pkp/pkp-lib/blob/e93c832fc028eb01186626d1ca64d84aa2064259/classes/security/authorization/SubmissionFileAccessPolicy.php#L243

**What application are you using?**
 OJS, OMP or OPS version stable 3.3.0, 3.4.0, main branch

Vitaliy-1:
User with manager role assigned as an assistant with stage restrictions, e.g., designer, will have an access to all workflow stages and submission files. I think `AssignedStageRoleHandlerOperationPolicy` was created to prevent this

Vitaliy-1:
PRs pkp-lib
main: https://github.com/pkp/pkp-lib/pull/9133
tests: https://github.com/pkp/ojs/pull/3961

Requires porting to stable branches

Vitaliy-1:
@asmecher, can you review this PR?

Vitaliy-1:
This PR modifies `AssignedStageRoleHandlerOperationPolicy` in the way that it grants the access to submission files according to stage assignments, even for roles with global access, like managers and admin. E.g., journal editor assigned as a layout editor won't have access to files in the submission stage. When this user loses the layout editor global role, the access to submission files is restored - meaning that assignment as a layout editor isn't taken into account. This prevents the bug described above. 

[...]

PR #9133, opened 2023-07-04
pkp/pkp-lib#9127 Restrict submission file access for global roles when dual-assigned

Vitaliy-1:


classes/security/authorization/AssignedStageRoleHandlerOperationPolicy.php
```diff
@@ -59,13 +63,49 @@ public function effect()
     {
         // Check whether the user has one of the allowed roles
         // assigned. If that's the case we'll permit access.
-        // Get user roles grouped by context.
-        $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_ACCESSIBLE_WORKFLOW_STAGES);
```
Vitaliy-1, line 63: As far as I understand, we don't need workflow stages, which are accessible for the user here, as they also include stages accessible because of local roles, like admin or manager. Instead stage assignments should be used.

Issue #9131, opened 2023-07-02
Typo in ReviewAssignmentFileWritePolicy which permits managers file access when assigned to a submission


Vitaliy-1:
**Describe the bug**
See https://github.com/pkp/pkp-lib/blob/abe282d110b5cd32f78d13e30b0f4af54932fabc/classes/security/authorization/ReviewAssignmentFileWritePolicy.php#L85
```php
if (empty($stageAssignments)...
```
Should be replaced with
```php
if (empty($assignedStages)...
```

As `stageAssignment` variable is always empty, I'm assuming that admin/managers have access to files even if assigned in a lower-level roles.

**To Reproduce**
1. Create a user and assign roles of journal editor and layout editor
2. Make a submission and it to the review stage, assign a reviewer and make a decision on it
3. Pass submission to the production stage and assign a user created in a step 1 as a layout editor
4. log in as this user, go to the review stage, click on Review Details and upload a file

Expected behaviour - the user shouldn't have access either for review file upload, nor to all other stages except production

**What application are you using?**
 OJS, OMP  from version stable 3.3.0

Vitaliy-1:
PRs
pkp-lib main: https://github.com/pkp/pkp-lib/pull/9135
tests: https://github.com/pkp/ojs/pull/3963

Requires porting to stable branches

Vitaliy-1:
@asmecher, can you take a look at this PR. I've applied here a general solution to the issue with dual assigned users with manager roles and related bugs. This solves also following issues:
https://github.com/pkp/pkp-lib/issues/9127
https://github.com/pkp/pkp-lib/issues/9104
And likely other related problems, e.g., described here: https://github.com/pkp/pkp-lib/issues/9104#issuecomment-1604571431

asmecher:
@Vitaliy-1, rather than coding to `main` and then back-porting to `stable-3_4_0` and `stable-3_3_0`, could you first code to `stable-3_4_0` and (when ready) forward-port from there? That'll help reduce the chances of a regression on a stable branch, plus I'd like to run the proposal past both @ajnyga (who worked on the original multiple-assignment logic) and the PKP|PS team (who have been encountering problems) for validation; they'll be working with `stable-3_3_0` installs.

asmecher:
[...]

PR #9135, opened 2023-07-06
pkp/pkp-lib#9131 Restrict access of managers if assigned to the submission in lower-level roles

Vitaliy-1:


classes/security/authorization/ReviewAssignmentFileWritePolicy.php
```diff
@@ -82,7 +82,7 @@ public function effect()
         }
 
         // Managers can write review attachments when they are not assigned to a submission
-        if (empty($stageAssignments) && count(array_intersect([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $userRoles))) {
+        if (empty($assignedStages) && count(array_intersect([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $userRoles))) {
```
asmecher, line 85: Good catch!

classes/user/Repository.php
```diff
@@ -197,41 +196,51 @@ public function canUserAccessStage($stageId, $workflowType, $userAccessibleStage
...
+                continue;
             }
-            if ($roleDao->userHasRole(Application::CONTEXT_SITE, $userId, Role::ROLE_ID_SITE_ADMIN)) {
-                $accessibleStageRoles[] = Role::ROLE_ID_SITE_ADMIN;
+
+            $accessibleWorkflowStages[$stageAssignment->getStageId()][] = $roleId;
+        }
+
+        // Managers and admin have access if not assigned to the submission or are assigned in a revoked role
+        $managerRoles = array_intersect($userRoleIds, [Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER]);
```
asmecher, line 235: I think `$userRoleIds` will never contain `Role::ROLE_ID_SITE_ADMIN` when the roles are fetched within the function, since that is a site-level role (`context_id = 0`) and `$userRoleIds` is populated from a specific `$contextId`.

asmecher (changes requested):

classes/user/Repository.php
Vitaliy-1: Thanks, fixed! It now checks for the admin role in a proper way.

Issue #9111, opened 2023-06-26
Wrong variable name in PASSWORD_RESET_CONFIRM email template


Vitaliy-1:
**Describe the bug**
See https://forum.pkp.sfu.ca/t/password-reset-isnt-working-as-expected/79422
`lostPasswordUrl` email template variable isn't replaced by the actual value.
This regression seems arises from https://github.com/pkp/pkp-lib/issues/8635#issuecomment-1431698517 where `passwordResetUrl` [was replaced](https://github.com/Vitaliy-1/ojs/commit/b144964191e8c6c0de819d8d55ff96626b7d24d2) by `lostPasswordUrl` in `PASSWORD_RESET_CONFIRM`. The initial intent for this replacement was to avoid exposition of password reset hash in the log for review reminder emails.

We don't have `lostPasswordUrl` but have `passwordLostUrl` email template variable composed from the Context, see `PKP\mail\variables\ContextEmailVariable::PASSWORD_LOST_URL`. But I think we can return `passwordResetUrl` with the hash there as it's not exposed anywhere and user requests it.  

It looks relevant only for fresh installs as this regression wasn't included in the migration script

**To Reproduce**
1. Install OJS 3.4
2. Set the default mailer to the log in OJS config
3. Register a user
4. Go to the reset password from 
5. Send an email with the link to reset the password to that user
6. See the log for the variable not being replaced

**What application are you using?**
 OJS, OMP or OPS 3.4 stable, main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/9118
OJS: https://github.com/pkp/ojs/pull/3956

(still requires porting to OMP/OPS)

Vitaliy-1:
I've made modification, which, I suspect, were intended in the original issue. Once merged, I'll open an issue to made necessary migration changes for new installs in 3.4.1

There are still old review-related templates that may have unsubstituted variables, e.g., `passwordResetUrl` in `REVIEW_RESPONSE_OVERDUE_AUTO` (previously `REVIEW_REQUEST_REMIND_AUTO`), see https://github.com/pkp/ojs/blob/stable-3_3_0/locale/en_US/emails.po#L253. This is relevant for locales, in which `emails.reviewRequestRemindAuto.body` isn't translated. The problem is that this wasn't included in the migration script, this means that upgraded instances are aIso affected and I can fix it only in 3.4.1.

PR #9151, opened 2023-07-13
[stable-3_4_0] pkp/pkp-lib#9111

Vitaliy-1:


classes/mail/mailables/ReviewRemindAuto.php
```diff
@@ -58,5 +59,8 @@ public function setData(?string $locale = null): void
         parent::setData($locale);
 
         $this->setOneClickAccessUrl($this->context, $this->reviewAssignment);
+
+        // See pkp/pkp-lib#9111
+        $this->addData(['lostPasswordUrl' => $this->viewData[ContextEmailVariable::PASSWORD_LOST_URL]]);
```
jonasraoni, line 64: Hi @Vitaliy-1, one PS client had the `passwordResetUrl` variable for this email, I was going to do a fast commit and include it, but perhaps there's something more to review.

Vitaliy-1: Can you explain I'm more details what is happening?

jonasraoni: This is a recent OJS 3.4 upgrade, the template `Review Reminder (Automated)` is using the variable `{$passwordResetUrl}`, but looks like it doesn't exist.

I've updated the application to use the latest commits, and the problem wasn't solved, so the email is sent with the plain-text `{$passwordResetUrl}`.

Issue #9152, opened 2023-07-13
Locale migration does not take into account plugin_settings table


defstat:
**Describe the bug**
Because the Locale migration does not take into account `plugin_settings` table, localized content from `customBlockManager` plugin is not getting migrated.

**What application are you using?**
 OJS `main` branch

**Additional information**
It seems that the localized content is in `setting_name`: [`blockContent`, `blockTitle`]

---------------------------

**PRs**

**Branch `main`**
PKP-LIB: #9156
OJS: pkp/ojs#3968 (***TESTS ONLY***)

defstat:
@asmecher I have added the proposed PR fix in the [top](https://github.com/pkp/pkp-lib/issues/9152#issue-1803138776) comment.

asmecher:
Just one comment needing attention, thanks! https://github.com/pkp/pkp-lib/pull/9156#issuecomment-1634711333

defstat:
@asmecher I have pushed the change to PR

PR #9156, opened 2023-07-13
[PKP-LIB][main] #9152 Add locale migration for customBlocksManager plugin

defstat:


asmecher:
This works, @defstat, but don't forget to flush the cache when working with settings directly. Otherwise the right values might be in the database but they won't be visible due to an out of date settings cache file.

defstat:
> This works, @defstat, but don't forget to flush the cache when working with settings directly. Otherwise the right values might be in the database but they won't be visible due to an out of date settings cache file.

Thanks @asmecher! I added the code bellow in every context iteration (code pushed):
```
$pluginSettingsDao = DAORegistry::getDAO('PluginSettingsDAO'); /** @var PluginSettingsDAO $pluginSettingsDao */
            
$cache = $pluginSettingsDao->_getCache($contextId, $blockPluginName);
$cache->flush();
```

Issue #9132, opened 2023-07-04
Logic Error in SQL query invalidates editorial statistics


ronste:
**Describe the bug**
When compiling the number of published submissions for all our journals I recognized differences in the reported OJS stats (editorial stats) as compared to the actual number of publications listed in the frontend. For some journals even zero published submissions are shown.

Some of these differences can be accounted for by the known issues with missing publication dates and imported articles.

However, sticking to one particular case with no imported articles at all, 37 articles listed in the frontend and 0 published submissions reported in editorial stats over the last 2 years, I debugged the code and came across the followling lines intended to exclude imported submissions:

https://github.com/pkp/pkp-lib/blob/5cad3a1534781d682b242b528158b957702948f1/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.inc.php#L365-L371

In line 369 all publications with no publication date are selected, i.e. included in the response instead of excluded as was intended. Changeing line 369 to:

```
$q->whereNotNull('pi.date_published')
```
generates the correct number of 37 published articles in the editorial stats backend.

**To Reproduce**
1. Count the number of publsihed submission in the frontend (maybe for a certian period to make it easier)
4. Compare to the number of published submissions in the editorial stats in the backend

**What application are you using?**
 OJS version 3.3.0-14

**Additional Information**
Original discussion of this code: https://github.com/pkp/pkp-lib/issues/6011

<hr>

PRs:
- https://github.com/pkp/pkp-lib/pull/9163 (stable-3_3_0)
- https://github.com/pkp/ojs/pull/3973 (stable-3_3_0 only submodule update)
- https://github.com/pkp/pkp-lib/pull/9162 (stable-3_4_0)
- https://github.com/pkp/ojs/pull/3972 (stable-3_4_0 only submodule update)
- https://github.com/pkp/pkp-lib/pull/9172
- https://github.com/pkp/ojs/pull/3978  (only submodule update)

Fix1:
- https://github.com/pkp/pkp-lib/pull/9218 (stable-3_3_0)
- https://github.com/pkp/ojs/pull/4000 (stable-3_3_0 only submodule update)
- https://github.com/pkp/omp/pull/1437 (stable-3_3_0 only submodule update)
- https://github.com/pkp/ops/pull/551 (stable-3_3_0 only submodule update)
- https://github.com/pkp/pkp-lib/pull/9221 (stable-3_4_0)
[...]

PR #9162, opened 2023-07-14
pkp/pkp-lib#9132 fix SQL datetime and date comparison for editorial s…

bozana:
s. https://github.com/pkp/pkp-lib/issues/9132

### DIFF
--- a/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
+++ b/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
@@ -394,7 +394,7 @@ abstract class PKPStatsEditorialQueryBuilder
         // submissions from being counted in editorial stats.
         $q->where(
             fn (Builder $q) => $q->whereNull('pi.date_published')
-                ->orWhere('s.date_submitted', '<=', DB::raw('pi.date_published'))
+                ->orWhere(DB::raw('CAST(s.date_submitted AS DATE)'), '<=', DB::raw('pi.date_published'))
         );
 
         Hook::call('Stats::editorial::queryObject', [&$q, $this]);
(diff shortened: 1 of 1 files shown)


bozana:
@asmecher or @jonasraoni, could one of you review this change?
I think DATE() can be used for both, MySQL and PostgreSQL, right?

jonasraoni:
@bozana the DATE isn't a function in PostgreSQL, but a construction, you can use it this way: `SELECT DATE '2023-01-01 10:22:22'` :D
But it works, so it's fine, another option that works for both would be `CAST('2023-01-01 10:22:22' AS DATE)`

jonasraoni:
And about the change... I've already had issues with it, see https://github.com/pkp/pkp-lib/issues/7709 and https://github.com/pkp/pkp-lib/issues/6011, AFAICR, it's not possible to reach a perfect solution (e.g. a solution might fit the QuickSubmit, but not the Native XML plugin and other import tools out there).

Personally, I'd drop this heuristic. It doesn't work very well and will have to be maintained forever.

Basically, to drop the concept of imported records for now, and recover it once the https://github.com/pkp/pkp-lib/issues/8389 gets implemented. Users that imported content, should understand that it might affect statistics, but that's just my opinion =]

bozana:
Thanks @jonasraoni. To be on the safe side, I will then use `CAST(X AS DATE)`...

bozana:
@jonasraoni, now I used CAST(X AS DATE). OK so?
I will not remove that heuristic now -- it is already there for some time, and the user might got used to it in a way, and it did solved a problem... So, lets hope the right solution will come soon... :-)

jonasraoni:
I think it's better/more standard.

About the heuristic, we would need the approval from @asmecher, I just wanted to raise the flag regarding the existing issues and future maintenance burden.

Issue #9166, opened 2023-07-14
Submission checklist migration sometimes presents an error on upgrade


asmecher:
**Describe the bug**
See: https://forum.pkp.sfu.ca/t/upgrade-from-3-3-0-10-to-latest-release/79880

```
2023-07-14 10:59:07 [migration: APP\migration\upgrade\v3_4_0\I7191_SubmissionChecklistMigration]
PHP Fatal error:  Uncaught TypeError: property_exists(): Argument #1 ($object_or_class) must be of type object|string, array given in /var/www/siteuser/data/www/library.uars.info/lib/pkp/classes/migration/upgrade/v3_4_0/I7191_SubmissionChecklistMigration.php:58
```

This is happening because PHP's `json_decode` will sometimes return an array and sometimes an object, depending on the array indexes for a provided input.

asmecher:
PRs:
- https://github.com/pkp/ojs/pull/3975 (TESTS ONLY)
- https://github.com/pkp/pkp-lib/pull/9167

Needs cherry-picking forward to `main` on merge!

asmecher:
@jonasraoni, could you have a look at the attached PR?

asmecher:
Merged and cherry-picked forward to `main`.

PR #9167, opened 2023-07-14
pkp/pkp-lib#9166 Ensure decoded type is treated consistently

asmecher:


classes/migration/upgrade/v3_4_0/I7191_SubmissionChecklistMigration.php
```diff
@@ -49,16 +49,16 @@ public function up(): void
                         ->delete();
                     return;
                 }
-                $checklist = json_decode($row->setting_value);
+                $checklist = json_decode($row->setting_value, true);
```
jonasraoni, line 52: It looks correct to me 👌

FYI, I had added some fixes for it here as well:
https://github.com/pkp/pkp-lib/blob/a1dabe01c9c52291ea63b3c500792ad731f3b82c/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php#L293-L305

jonasraoni (approved):

classes/migration/upgrade/v3_4_0/I7191_SubmissionChecklistMigration.php
asmecher: Thanks; I'm tempted not to disturb that earlier code, since it's already released and won't do harm in this case.

PR #9176, opened 2023-07-19
pkp/pkp-lib#7698 Laravel router to handle API routing

touhidurabir:
for pkp/pkp-lib#7698


api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,191 @@
...
+            "has.user",
+            "has.roles:{$roles}",
+        ];
+    }
+
+    public function getGroupRoutesCallback(): Closure
+    {
+        return function() {
+            
+            Route::get('all', [static::class, 'getJobs'])
```
asmecher, line 55: Let's use first-class callables when possible: `$this->getJobs(...)` [should work](https://laravel-code.tips/you-can-use-first-class-callables-for-routes-in-php-81/), I think.

classes/core/APIRouter.php
```diff
@@ -107,7 +107,11 @@ public function route($request)
 
         $handler = require('./' . $sourceFile);
         $this->setHandler($handler);
-        $handler->getApp()->run();
+        
+        $handlerApp = $handler->getApp();
+        if ($handlerApp !== null) {
```
asmecher, line 112: What would cause `$handlerApp` to be null? (If it's a possible case, then shouldn't there be error handling here?)

classes/core/PKPApiRoutingHandler.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace PKP\core;
+
+use PKP\handler\APIHandler;
+
+class PKPApiRoutingHandler extends APIHandler
```
asmecher, line 7: Odd -- I can't see where this class is attached to anything! What am I missing?

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,181 @@
...
+    public function registerRouter(): void
+    {
+        $this->app->singleton('router', function ($app) {
+            $request = Request::capture();
+            $app->instance(\Illuminate\Http\Request::class, $request);
+            return new Router($app['events'], $app);
+        });
+    }
+
+    public function registerRouteMiddlewares(): void
```
asmecher, line 82: Middleware is plural already -- please change anywhere that refers to `Middlewares` to `Middleware` instead (unless that's a Laravel convention).

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,181 @@
...
+        ValidatePostSize::class,
+        TrimStrings::class,
+        ConvertEmptyStringsToNull::class,
+    ];
+
+    /**
+     * The application's route middleware.
+     * These middleware can/should be assigned to specific routes or routes groups individually.
+     */
+    protected array $routeMiddlewares = [
```
asmecher, line 42: This list is going to grow out of control pretty quickly when we start making broad use of the new routing toolset, I think - especially if we eventually unify our authorization classes with Laravel. There are lots of those and they are complex!

Would it be possible to delegate building this list to the individual controller, rather than centralizing it here?

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        return self::$globalMiddlewares;
+    }
+
+    /**
+     * Register the service provider.
+     *
+     * @return void
+     */
+    public function register()
```
asmecher, line 58: This could present a maintenance headache as it duplicates and extends the parent implementation rather than using `parent::register()`. Would it be possible to let `parent::register()` take care of the built-ins, and have this function just add the missing pieces (`registerRouteMiddlewards()` at a glance)? Or is the order important?

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,181 @@
...
+        $container->bind(\Illuminate\Contracts\View\Factory::class, \Illuminate\View\Factory::class);
+        $container->bind(\Illuminate\Contracts\Routing\ResponseFactory::class, \Illuminate\Routing\ResponseFactory::class);
+    }
+
+    /**
+     * Register the URL generator service.
+     *
+     * @return void
+     */
+    protected function registerUrlGenerator()
```
asmecher, line 117: Is this a copy from the parent class?

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Register a binding for the PSR-7 request implementation.
+     *
+     * @return void
+     *
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
+     */
+    protected function registerPsrRequest()
```
asmecher, line 168: This looks like it's just copied in from the parent class; if so, please remove it. Anyway, the packages that are being checked for in our package list.

asmecher (changes requested):
@touhidurabir, thanks, this is excellent! I've added some comments that I hope are helpful -- I'm still learning the Laravel way of thinking and trying to strike a compromise between that and our own internal toolset.

Once we're both satisfied with the design of the Laravel replacements, the next step will be to convert *all existing handlers* over to the new toolset. Obviously that'll be a lot of work and we'll need to coordinate with e.g. Vitaliy and others who are working on related work like the submission lists.

Then the last step will be to remove the old toolset (Slim, the old middleware classes, the routing stuff that kicks it off, etc). Obviously you can't remove that stuff quite yet but please feel free to flag the code with a FIXME comment so we don't forget to clean it out before the merge.

classes/middlewares/RemoveTrailingSlash.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $uriPath = $request->path();
+
+        if ($uriPath === '/' || substr($uriPath, -1) !== '/') {
+            return $next($request);
+        }
+
+        $uriPath = substr($uriPath, 0, -1);
+
+        $newRequest = Request::create(
```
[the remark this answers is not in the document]
touhidurabir: Right, this is not needed at all . I did not noticed that comment from the symfony class .

classes/middlewares/HasRoles.php
```diff
@@ -0,0 +1,49 @@
...
+     * 
+     * 
+     * @param \Illuminate\Http\Request  $request
+     * @param Closure                   $next
+     * @param string                    $matchableRoles The passed comma separated roles e.g. , 1,2,3
+     * @param bool                      $string         Should the passed roles match all(strict) or any(loose)
+     * 
+     * @return mixed
+     */
+    public function handle(Request $request, Closure $next, string $matchableRoles, bool $strict = false)
```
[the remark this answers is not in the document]
touhidurabir: Updated .

classes/handler/APIHandler.php
```diff
@@ -54,6 +62,54 @@ class APIHandler extends PKPHandler
...
+                        : $exception->getMessage(),
+                ], $exception instanceof NotFoundHttpException 
+                    ? Response::HTTP_NOT_FOUND 
+                    : (in_array($exception->getCode(), array_keys(Response::$statusTexts)) 
+                        ? $exception->getCode() 
+                        : Response::HTTP_INTERNAL_SERVER_ERROR)
+                )->send();
+            }
+
+            return;
```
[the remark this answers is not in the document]
touhidurabir: Yes, it should be removed .

classes/core/PKPBaseController.php
```diff
@@ -0,0 +1,32 @@
...
+    use AuthorizesRequests;
+    use DispatchesJobs;
+    use ValidatesRequests;
+
+    public function getPathPattern(): ?string
+    {
+        return null;
+    }
+
+    public function nonContextualApi(): bool
```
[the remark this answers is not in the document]
touhidurabir: method name updated to `isSiteWide` .

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,191 @@
...
+    {
+        $roles = implode('|', [Role::ROLE_ID_SITE_ADMIN]);
+
+        return [
+            "has.user",
+            "has.roles:{$roles}",
+        ];
+    }
+
+    public function getGroupRoutesCallback(): Closure
```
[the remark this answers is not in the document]
touhidurabir: This is excellent , I knew about the `First class callable syntax` but never used it or thought about it . seems pretty useful, update .

api/v1/jobs/PKPJobController.php
touhidurabir: Update . But note that it's only useable for now as we have decided keep the routes in their own controller classes . If we decide to move it to separate `routes` file, then we have to user something like `ControllerClass@method` or [Controller::class, 'method'] syntax .

classes/core/APIRouter.php
touhidurabir: This become a case with applying and matching laravel routing system which applied [here](https://github.com/pkp/pkp-lib/pull/9176/files#diff-d6cd68c8ec4cd5788ad82e8f44a3fbe49186121c580151a3ac59315742e0dee6R66-R110) . Here we are returning early if match with any laravel routes or if any routes defined(as for not we have those define in each controller class) . so `app` never get to initialized . that why we need that check. It will become unnecessary once we refactor all handler to controller and remove the slim from the system .

classes/core/PKPApiRoutingHandler.php
touhidurabir: This is to facilitate our current implementation of API routing system . Our current `APIRouter` required a handler that extends `APIHandler` to work . But if we plan to use laravel routing system , we need controller . So this one single class will work as a bridge between those 2 . See how it's use for [User](https://github.com/pkp/ojs/pull/3980/files#diff-a67dc7f90e012f5dfa7eef2a5424a828c1d83a6ef1017d8509615d414de18fa9) and [Job](https://github.com/pkp/ojs/pull/3980/files#diff-cdfbe01ee2eb75a66b1d59b1af4625d2e48331be728e64ad7973c08da4ea9f7c) how it's used where we inject the controller class itself .

classes/core/PKPRoutingProvider.php
touhidurabir: > Middleware is plural already

That is correct . Update

classes/core/PKPRoutingProvider.php
touhidurabir: This is just to maintain a list of middleware that may use a lot and given those an alias name. we can always apply any new middleware on each of routes as we like 
```php
Route::put('/post/{id}', function (string $id) {
    // ...
})->middleware('role:editor');

// or

Route::put('/post/{id}', function (string $id) {
    // ...
})->middleware([SomeMiddleware::class, ...]);
```

classes/core/PKPRoutingProvider.php
touhidurabir: > Would it be possible to let parent::register() take care of the built-ins,

Yes 

> Or is the order important?

No 

Thanks for spotting it . It was an unintended left over from routing test locally . Fixed .

classes/core/PKPRoutingProvider.php
touhidurabir: The only reason for this is because laravel `app('request')` expects to return an instance `Illuminate\Http\Request` but because of our [override](https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPContainer.php#L163), now the `app('request')` return an instance of our `PKPRequest` so I had to override it with `Request::capture()` . 

I am wondering if the [override](https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPContainer.php#L163) is necessary at all ? I don't see any usage of of `app('request')` or `app()->get('request')` in our codebase but not too sure about the plugins . If not, we can clean it up a lot .

classes/core/PKPRoutingProvider.php
touhidurabir: see comment at https://github.com/pkp/pkp-lib/pull/9176#discussion_r1288266039

classes/core/PKPRoutingProvider.php
asmecher: Ah, that makes sense!

classes/core/PKPRoutingProvider.php
asmecher: OK, great!

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,187 @@
...
+            ->name('redispatchFailedJob');
+        
+        Route::delete('failed/delete/{jobId}', $this->deleteFailedJob(...))
+            ->name('deleteFailedJob');
+    }
+
+    /**
+     * Get all pending jobs in the queue waiting to get executed
+     * 
+     * @param   \Illuminate\Http\Request $request
```
asmecher, line 71: The `@param and `@return` self-documentation doesn't add anything now that we have good typehints; please remove them (here and elsewhere).

classes/core/APIRouter.php
asmecher: OK, please add a `// FIXME pkp/pkp-lib#7698` here so we don't forget before merging (and/or a checkbox in the issue description)!

touhidurabir: Added .

api/v1/jobs/PKPJobController.php
touhidurabir: Removed where it seems unnecessary .

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,187 @@
...
+     */
+    public function redispatchAllFailedJob(Request $request): JsonResponse
+    {
+        if (Repo::failedJob()->total() <= 0) {
+            return response()->json([
+                'errorMessage' => __('api.jobs.406.failedJobEmpty')
+            ], Response::HTTP_NOT_ACCEPTABLE);
+        }
+
+        $redispatableFailedJobs = Repo::failedJob()->getRedispatchableJobsInQueue(null, ['id']);
```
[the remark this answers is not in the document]
touhidurabir: fixed .

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,172 @@
...
+                'currentPage' => $failedJobs->currentPage(),
+            ],
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Redispatch all failed jobs back to queue
+     * It will only redispatch failed jobs that has valid payload attribute
+     */
+    public function redispatchAllFailedJob(Request $request): JsonResponse
```
jonasraoni, line 116: I think it should be `redispatchAllFailedJobs`

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,172 @@
+<?php
+
+/**
+ * @file api/v1/jobs/PKPJobController.php
+ *
+ * Copyright (c) 2014-2023 Simon Fraser University
+ * Copyright (c) 2003-2023 John Willinsky
```
jonasraoni, line 7: On new files we're adding just the current year instead of the range.

api/v1/jobs/PKPJobController.php
```diff
@@ -0,0 +1,172 @@
...
+        return 'jobs';
+    }
+
+    public function getRouteGroupMiddleware(): array
+    {
+        $roles = implode('|', [Role::ROLE_ID_SITE_ADMIN]);
+
+        return [
+            "has.user",
+            "has.roles:{$roles}",
```
jonasraoni, line 46: I think it would be better to have these strings coming from a class, to avoid typos (e.g. `return [HasUserMiddleware::class, HasRolesMiddleware::for(Role::ROLE_ID_SITE_ADMIN)]` or something like that).

classes/core/PKPRoutingProvider.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Boot the service provider.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        Response::macro('withCSV', function (array $rows, array $columns, int $maxRows) {
```
jonasraoni, line 74: I'm not sure @asmecher enjoys macros, I've tried to add one and it was rejected.
The `$rows`/`$columns` arguments could be iterators to give room for future optimizations.

jonasraoni (approved):
I've approved, my comments are just nitpicks 😁

asmecher:
@touhidurabir, with apologies for the delays, there are two changes still required:
- There's a merge conflict on locale/en/api.po
- I've just merged https://github.com/pkp/pkp-lib/issues/9262, which introduces a new handler that will also be adapted (api/v1/highlights in each app and pkp-lib)

Could you clear those up? No need to kick off tests on all apps -- the changes for the OMP, OPS, and OJS repos should all be identical.

includes/functions.php
```diff
@@ -431,3 +431,27 @@ function isValidJson(mixed $data): bool
...
+ * Convert a query parameter to an array
+ *
+ * This method will convert a query parameter to an array, and
+ * supports a comma-separated list of values
+ * 
+ * @param mixed $value
+ * @return array
+ */
+if (!function_exists('paramToArray')) {
+    function paramToArray(mixed $value): array
```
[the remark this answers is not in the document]
asmecher: Agreed, but I'll adjust this myself after merge. (@touhidurabir has suffered through enough rebases already!)

Issue #9147, opened 2023-07-12
Invitation to review Email - Display abstract problem


defstat:
**Describe the bug**
When trying to send a review request for the submission `Sodium butyrate improves growth performance of weaned piglets during the first period after weaning` of the `datasets` db, the email send to the reviewer seems to have included an abstract that is not the same as the submission's abstract. 

The email send is displayed below:
![image](https://github.com/pkp/pkp-lib/assets/560456/a4d75b79-e9db-4592-b6d5-08eb436faae0)

The abstract of the submission is:
`The aim of this study was to assess the influence of long-term fat supplementation on the fatty acid profile of heavy pig adipose tissue. Fifty-four Large White barrows, averaging 25 kg LW, were randomized (matched weights) to one of three isoenergetic diets supplemented with either tallow (TA), maize oil (MO), or rapeseed oil (RO). The fats were supplement- ed at 3% as fed from 25 to 110 kg LW, and at 2.5 % from 110 kg to slaughtering. Following slaughter at about 160 kg LW, backfat samples were collected from ten animals per treatment and analyzed. Fatty acid composition of backfat close- ly reflected the fatty acid composition of the supplemented fats. The backfat of pigs fed TA had the highest saturated fatty acid content (SFA) (P<0.01); those fed MO had the highest content in polyunsaturated fatty acid (PUFA) and the lowest in monounsaturated fatty acid (MUFA) content; those fed RO had the highest content of linolenic acid (C18:3) and cis 11- ecosenoic acid (C20:1). Only MO treatment had an effect on linoleic acid levels and the iodine value (IV) of backfat, result- ing in levels higher than those (IV = 70; C18:2 = 15%) accepted by the Parma Consortium for dry-cured ham. The IV and unsaturation index in both layers of subcutaneous backfat tissue differed significantly between treatments. These results show that long-term dietary supplementation with different fats changes the fatty acid profile of heavy pig adipose tissue. Supplementation with rapeseed oil increases the proportion of “healthy” fatty acids in pig fat, thereby improving the nutritional quality, however the effects on the technological quality of the fat must be carefully assessed.`

**To Reproduce**
1. Go to the workflow of a submission that is in the review stage
2. Click on `Add Reviewer`
3. Select one new Reviewer
4. Add the reviewer and see the above behavior in the sent email.

**What application are you using?**
 OJS `main branch`

**Additional information**
It seems that the `<` character has something to do with the result.

**PRs**
[...]

PR #9178, opened 2023-07-20
pkp/pkp-lib#9147 update string sanitization process for < char

touhidurabir:
fix for pkp/pkp-lib#9147


asmecher:
See https://github.com/pkp/pkp-lib/issues/9147#issuecomment-1679711510

Issue #9180, opened 2023-07-24
Update to cypress 12, reduce flakiness on M1


jardakotesovec:
While trying to run cypress test on dev machine (apple M1), I ran into couple of failures.

Main reason to update to cypress 12, because it runs correct chrome (native, not intel binary which is super slow) and it comes with newer electron, which is not causing (intermittent) exception within tinyMCE initialization..

Challenge while debugging the issues was interdependency between tests, thats something that we should address in longer term.

For shorter term, I moved NativeXmlImportExportIssue.cy to be run almost at the end. Because it re-imports two submissions with exactly the same name - which was causing failure (for me it was 100%, not sure why on CI its intermittent) for Subscriptions.js, because the url for these duplicated article was not resolving to correct submission (as url is based on the title).

Only migration migration that was necessary due cypress 12 was migrate  cy.server&cy.route to cy.intercept, that change will go to pkp-lib..

asmecher:
@jardakotesovec, the PRs for this are the following, right?

- https://github.com/pkp/ojs/pull/3984/files
- https://github.com/pkp/ops/pull/547
- https://github.com/pkp/omp/pull/1430
- https://github.com/pkp/pkp-lib/pull/9181/files

Not sure if there are others.

I've reviewed the PRs and made a couple of comments, but nothing major; once you're satisfied with those, please go ahead and merge (or bug me to). Fingers crossed that this will take a chunk out of the random failures we've been seeing recently and we can build on things from there.

@jardakotesovec, you were asking about tools to restore the database state in order to make tests less interdependent. We do have a [rough-and-ready way to do that](https://github.com/pkp/pkp-lib/blob/main/tests/PKPTestHelper.php) with the PHPUnit tests; in PHP-land you can call `PKPTestHelper::restoreDB`. But that doesn't help you much in Javascript. The best thing to do might be to code up a bash script that's a companion to [dump-database.sh](https://github.com/pkp/pkp-lib/blob/main/tools/travis/dump-database.sh) and then `exec` it from within Cypress when needed or as part of setup/teardown. (The database and files directory are dumped [here](https://github.com/pkp/pkp-lib/blob/e5f5f14ef7f180c485d90b55a718e61f5e0364f5/tools/travis/run-tests.sh#L19..L20).)

asmecher:
[...]

PR #9181, opened 2023-07-24
pkp/pkp-lib#9180 update cypress12

jardakotesovec:


jonasraoni (approved):
LGTM. Would be nice to update the remaining tests, but we can do it on demand.

PR #9198, opened 2023-07-31
pkp/pkp-lib#7495 submission lists UI

Vitaliy-1:


api/v1/_submissions/PKPBackendSubmissionsHandler.php
```diff
@@ -108,18 +134,39 @@ public function getMany($slimRequest, $response, $args)
...
+                    break;
+
+                case 'isIncomplete':
+                    $collector->filterByIncomplete(true);
+                    break;
+            }
+        }
+
+        // In new submission lists this endpoint is dedicated to retrieve all submissions only by admins and managers
+        if (!Config::getVar('features', 'enable_new_submission_listing')) {
```
asmecher, line 158: To decrease the chances of parts of this getting left in the `main` codebase, please file an issue against the `3.5 Internal` milestone for cleanup, and add FIXME notes to the parts of the code that'll require cleanup before release:
```
// FIXME: Clean up before release pkp/pkp-lib#1234
```
That way we can grep for cases like this and be sure we get them all.

classes/components/forms/dashboard/SubmissionFilters.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        $this
+            ->addSectionFields()
+            ->addAssignedTo()
+            ->addIssues()
+            ->addCategories()
+        ;
+    }
+
+    protected function isManager(): bool
```
asmecher, line 56: `isManager` is misleading; `isManagerOrAdmin` would be more accurate. Maybe name the function for its purpose (`canAssign`) rather than how it works?

classes/submission/Collector.php
```diff
@@ -434,6 +441,15 @@ public function getQueryBuilder(): Builder
                 ->whereIn('pc.category_id', $this->categoryIds);
         }
 
+        if ($this->isReviewedBy !== null) {
```
asmecher, line 444: I suppose the `TODO` might cover this, but this `if` is redundant. This is personal taste, but I find the `->when` formulations clearer to read when everything is written with them consistently, rather than mixing `if` with `->when`.

pages/dashboard/Column.php
```diff
@@ -0,0 +1,28 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Column
+ *
+ * @ingroup pages_dashboard
+ *
+ * @brief A class that represents a column in the submissions table
+ */
+
+namespace PKP\pages\dashboard;
```
asmecher, line 16: This should probably be somewhere in `classes`; the `pages` directory should only be used for handler classes.

pages/dashboard/DashboardHandlerNext.php
```diff
@@ -0,0 +1,358 @@
...
+
+        $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
+        $apiUrl = $dispatcher->url($request, PKPApplication::ROUTE_API, $context->getPath(), '_submissions');
+
+        $sections = Repo::section()
+            ->getCollector()
+            ->filterByContextIds([$context->getId()])
+            ->getMany();
+
+        $categories = Repo::category()
```
asmecher, line 94: There's definitely going to be optimization work required on this page; we're loading all categories, sections, user groups, and potentially all queued submissions into memory for each page load.

Categories, for example, could be quite a long list -- but I don't think they'll frequently be relevant to the editorial workflow.

templates/dashboard/column-title.tpl
```diff
@@ -0,0 +1,21 @@
...
+	is="table-cell"
+	class="submissions__list__item__title"
+	:id="'submission-title-' + submission.id"
+	:is-row-header="true"
+>
+	<span class="submissions__list__item__author">
+		{{ submission.publications[0].authorsStringShort }}
+	</span>
+	<template v-if="submission.publications[0].authorsStringShort">—</template>
+	{{ submission.publications[0].fullTitle.en }}
```
asmecher, line 20: Be careful of hard-coding `en` here without a FIXME -- these will be hard to spot later!

asmecher (commented):
Hi all! I've added a quick review with two big caveats:
- I'm somewhat of an outsider to this work, so it's not a fine-detail review, and
- It's still a WIP.

That said, I've just added a few comments that I'm OK to consider immediately or later.

classes/submission/Collector.php
Vitaliy-1: Yes, this should be replaced in the whole query builder. Leaving for later.

pages/dashboard/Column.php
Vitaliy-1: I've replaced it with anonymous class for now

PR #9207, opened 2023-08-04
pkp/pkp-lib#8025 static cache of locale precedence implementation update to resolve job failure on locale update

touhidurabir:
for pkp/pkp-lib#8025

### DIFF
--- a/classes/core/DataObject.php
+++ b/classes/core/DataObject.php
@@ -49,10 +49,27 @@ class DataObject
     public $_injectionAdaptersLoaded = false;
 
     /**
+     * Cached locale precedence on a single reqyest life cycle
+     * 
+     * @var null|array 
+     */
+    private static $localePrecedence = null;
+
+    /**
      * Constructor
      */
     public function __construct()
     {
+    }
+
+    /**
+     * Reset/Clear the locale precedence during a single request life cycle
+     * 
+     * @return void
+     */
+    public static function resetLocalePrecedence(): void
+    {
+        static::$localePrecedence = null;
     }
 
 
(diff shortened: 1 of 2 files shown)


classes/core/DataObject.php
```diff
@@ -48,13 +48,30 @@ class DataObject
     /** @var bool whether injection adapters have already been loaded from the database */
     public $_injectionAdaptersLoaded = false;
 
+    /**
+     * Cached locale precedence on a single reqyest life cycle
```
asmecher, line 52: *request

asmecher (changes requested):
Thanks, Touhidur!

Could you open this PR against `stable-3_4_0` rather than `main`? The issue is filed against 3.4.0-x, and it's safer to develop against a stable branch then forward-port to `main` rather than the opposite.

Just a thought, but I think I see a way of fixing this that requires a little less entanglement between classes. Rather than maintaining a list of locales (e.g. `['en', 'fr_CA']`) and having to reset it when something changes, why not maintain a list of authorities on what locale should be used (e.g. `[Locale::getLocale(...), $context->getPrimaryLocale(...), $site->getPrimaryLocale(...)]`)? Then that list can stay unchanged; all that will change is the return value from `Locale::getLocale()` when it's called. No need to reset a static variable or entangle `Locale` and `DataObject` further.

(This is the first-class callable syntax I've been recently using elsewhere in the system. It's a lot nicer to work with callbacks this way than it was with `[$object, 'methodName']`.)

Eventually we'll probably have to go one step further and account for the `$request` object to change -- in fact I'm surprised we haven't yet. But let's not get ahead of ourselves!

touhidurabir:
> maintain a list of authorities on what locale should be used 
> Then that list can stay unchanged; all that will change is the return value from `Locale::getLocale()` when it's called. No need to reset a static variable or entangle `Locale` and `DataObject` further.

@asmecher I see your point . And I think that way we can remove the `private` method `DataObject::_getLocalePrecedence`  as that should be something managed by `Locale` class, not `DataObject` . But for the performance sake, we still should maintain a static cached data of that list(that's my thought and maybe you are pointing to that also) . That way we can avoid the entanglement also .

touhidurabir:
Closing this PR as it will be implemented first for `stable-3_4_0` branch and will be ported to `main` branch from there .

PR #9209, opened 2023-08-07
pkp/pkp-lib#7916 updated implementation for sanitization to support plugin/general usage

touhidurabir:
for pkp/pkp-lib#7916

### DIFF
--- a/classes/core/PKPString.php
+++ b/classes/core/PKPString.php
@@ -442,22 +442,32 @@ class PKPString
     /**
      * Sanitize HTML string based on given allowed tags with attributes
      *
-     * @param string                                                    $input                       input string
-     * @param string                                                    $allowedTagsWithAttributes   Should be in format as 'a[href|target|title],em,strong,cite,...'
-     * @param \Symfony\Component\HtmlSanitizer\HtmlSanitizerConfig|null $config
+     * @param string                                                $input                       input string
+     * @param string                                                $allowedTagsWithAttributes   Should be in format as 'a[href|target|title],em,strong,cite,...'
+     * @param \Symfony\Component\HtmlSanitizer\HtmlSanitizer|null   $sanitizer
      *
      * @return string
      */
-    public static function sanitizeHtmlString(string $input, string $allowedTagsWithAttributes, HtmlSanitizerConfig $config = null): string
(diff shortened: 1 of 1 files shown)


classes/core/PKPString.php
```diff
@@ -460,6 +439,78 @@ public static function stripUnsafeHtml(?string $input, string $configKey = 'allo
...
+            );
+    }
+
+    /**
+     * Generate the allowed tags to attributes map for sanitization process
+     *
+     * @param   string $allowedTagsWithAttributes   Should be in format as 'a[href|target|title],em,strong,cite,...'
+     * @return \Illuminate\Support\Collection
+     */
+    public static function generateSanitizableAllowedTagToAttributeMap(string $allowedTagsWithAttributes): Collection
```
asmecher, line 487: This is a mouthful of a function! Functions like these tend to collect in helper classes and gradually become quite a bit of technical debt.

The majority of the code here is used to parse allowed tags from strings, which is only really used in the `stripTags` case -- for the plugin-based cases it's always hard-coded as only `<p>` being allowed.

I think it would be better to keep `stripTags` as it was, and for the two plugins, just wrap a call to `strip_tags` around the contents before they are passed into the sanitizer. It doesn't add complexity to the plugins, and avoids us having to support `sanitizeHtmlString` until the end of time (as we do once something is used in a plugin).

touhidurabir: @asmecher In that case we can just simply drop this PR and update the implementation of the plugins with the use of `strip_tags` method . But the problem is the behaviour of `strip_tags` which remove anything after char `<` as noted at https://github.com/pkp/pkp-lib/issues/9147 . perhaps we should add a new function such as PKPString::stripTags with implementation as 

```php
public static function stripTags(string $input, string|array $allowedTags = [])
{
    $allowedTags = Arr::wrap($allowedTags);
    return html_entity_decode(
        strip_tags(preg_replace("/<([^>]*(<|$))/", "&lt;$1", $input), $allowedTags)
    );
}
```

asmecher: I think I've mooted that problem -- see https://github.com/pkp/pkp-lib/issues/9147#issuecomment-1679711510.

Issue #9195, opened 2023-07-28
Add controlling configuration to queue jobs


touhidurabir:
**Describe the bug**
We have queue jobs that can managed and executed via JobRunner or Cron or Worker Thread . But it works with default jobs configurations such as tries, backoff with no maxException etc . It will be a good enhancement to have this added to the `PKP\jobs\BaseJob` class. 

**Who will be benefited from it**
This will help users to handle case such as https://forum.pkp.sfu.ca/t/ojs-3-4-0-1-pkp-jobs-doi-depositsubmission-has-been-attempted-too-many-times-or-run-too-long/80154/3 . In fact `job has been attempted too many times or run too long` is a very common occurrence for which can be handled via worker command options but have no default configs when running via JobRunner. 

**What configs should be added**
1. tries = 3
2. backoff = 5 seconds
3. maxExceptions = 3
4. timeout = 60 seconds with increased to 180 for long running jobs .
5. failOnTimeout = true
6. retry_after = 240 seconds 

**What application are you using?**
 OJS, OMP or OPS version `3.4.0` and `main`

**PRs**

#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9213
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/3998 [TEST ONLY]

Need to forward port to `main`

touhidurabir:
@asmecher can you please review the PR at https://github.com/pkp/pkp-lib/issues/9195#issue-1826346096

asmecher:
@touhidurabir, no objections here, but are the two PRs syced?

touhidurabir:
@asmecher I suggest to discard the `main` branch PR and forward port the `stable-3.4.0` branch PR . I am closing the `main` branch PR as that has some outdated changes .

asmecher:
Thanks, @touhidurabir -- merged to `stable-3_4_0` for `pkp-lib`, and cherry-picked that forward to `main`. I'll handle the submodule updates separately.

touhidurabir:
@asmecher Patch PR

### stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9263
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4029 [TEST ONLY]

asmecher:
[...]

PR #9213, opened 2023-08-08
pkp/pkp-lib#9195 Added queue jobs control configs

touhidurabir:
for pkp/pkp-lib#9195

### DIFF
--- a/jobs/BaseJob.php
+++ b/jobs/BaseJob.php
@@ -34,56 +34,32 @@ abstract class BaseJob implements ShouldQueue
     use SerializesModels;
 
     /**
-     * The name of the connection the job should be sent to.
-     *
-     * @var string|null
-     */
-    public $connection;
-
-    /**
-     * The queue's name where the job will be consumed
-     *
-     * @var string
-     */
-    public $queue;
-
-    /**
      * The number of times the job may be attempted.
-     *
-     * @var int
      */
-    public $tries = 3;
+    public int $tries = 3;
 
     /**
      * The number of SECONDS to wait before retrying the job.
-     *
-     * @var int
      */
-    public $backoff = 5;
+    public int $backoff = 5;
 
     /**
      * The maximum number of SECONDS a job should get processed before consider failed
-     *
-     * @var int
      */
-    public $timeout = 120;
+    public int $timeout = 60;
 
     /**
      * The maximum number of unhandled exceptions to allow before failing.
-     *
-     * @var int
      */
-    public $maxExceptions = 3;
+    public int $maxExceptions = 3;
 
     /**
      * Indicate if the job should be marked as failed on timeout.
-     *
-     * @var bool
      */
-    public $failOnTimeout = true;
+    public bool $failOnTimeout = true;
 
     /**
-     * Initiate the job
+     * Initialize the job
      */
     public function __construct()
     {
--- a/jobs/BaseJob.php
+++ b/jobs/BaseJob.php
@@ -47,12 +47,53 @@ abstract class BaseJob implements ShouldQueue
      */
     public $queue;
 
+    /**
+     * The number of times the job may be attempted.
+     *
+     * @var int
+     */
+    public $tries = 3;
+
+    /**
+     * The number of SECONDS to wait before retrying the job.
+     *
+     * @var int
+     */
+    public $backoff = 5;
+
+    /**
+     * The maximum number of SECONDS a job should get processed before consider failed
+     *
+     * @var int
+     */
+    public $timeout = 120;
+
+    /**
+     * The maximum number of unhandled exceptions to allow before failing.
+     *
+     * @var int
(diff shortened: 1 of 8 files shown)


jobs/BaseJob.php
```diff
@@ -47,12 +47,53 @@ abstract class BaseJob implements ShouldQueue
...
+     * The number of times the job may be attempted.
+     *
+     * @var int
+     */
+    public $tries = 3;
+
+    /**
+     * The number of SECONDS to wait before retrying the job.
+     *
+     * @var int
```
asmecher, line 60: Rather than `@var` self-documentation, better to just use the `public int $backoff = 5;` directly. (Also below)

jobs/BaseJob.php
```diff
@@ -47,12 +47,53 @@ abstract class BaseJob implements ShouldQueue
...
+
+    /**
+     * Indicate if the job should be marked as failed on timeout.
+     *
+     * @var bool
+     */
+    public $failOnTimeout = true;
+
+    /**
+     * Initiate the job
```
asmecher, line 86: Should be `Initialize` rather than `Initiate`, I think -- `Initiate` implies that calling the constructor starts the job. (Also below)

jobs/email/EditorialReminder.php
```diff
@@ -81,7 +81,7 @@ public function handle(): void
 
         /** @var int $submissionId */
         foreach ($submissionIds as $submissionId) {
-            $submission = Repo::submission()->get($submissionId);
+            $submission = Repo::submission()->get((int)$submissionId);
```
asmecher, line 84: Is this change necessary?

jobs/email/EditorialReminder.php
```diff
@@ -81,7 +81,7 @@ public function handle(): void
...
             $submissions[$submissionId] = $submission;
 
             if ($submission->getData('stageId') === WORKFLOW_STAGE_ID_SUBMISSION) {
```
asmecher, line 87: This could be one of those weird data type issues, I think. `$submission->getData('stageId')` might come back from the DB as a string. To be safe, use a `==` comparison.

asmecher (commented):
Thanks, @touhidurabir! Am I reading correctly that this adds the capability for additional configurations but doesn't use it anywhere? I'm not a big fan of adding capabilities unless we at least demonstrate their use somewhere in the codebase; otherwise it's dead code. Is there a case or two where this would prove useful to end users that we could add to the PR?

touhidurabir:
> this adds the capability for additional configurations but doesn't use it anywhere

@asmecher not the case . once these added , this will have reflect on both case when job run via `JobRunner` or from `CLI` or `worker` . And these add few enhancement like specially the `backOff` property that can have impact when failed job picked too fast to execute again . The other few just add some extra enhancement just to make sure it a bit higher threshold than default like the `timeout` , `maxException` and `failOnTimeout` . 

for example, `timeout` will be helpful for long running jobs such as balk mailing which may need more than 60 seconds which is default . The same goes for `failOnTimeout` as it's right to make a job marked as failed when passed the timeout limit before complete and exit .

jobs/email/EditorialReminder.php
touhidurabir: @asmecher this is added because of this error posted in forum https://forum.pkp.sfu.ca/t/ojs-3-4-0-1-failed-jobs/80170/4?u=touhidur . I also could not reproduce it in my testing but may be this is one of those platform specific type issues that we are experiencing too randomly .

jobs/email/EditorialReminder.php
touhidurabir: Done .

jobs/BaseJob.php
touhidurabir: done .

jobs/BaseJob.php
touhidurabir: Done .

jobs/email/EditorialReminder.php
asmecher: I think this is only happening because you've got [strict types](https://github.com/pkp/pkp-lib/pull/9213/files#diff-83d61d197c61d91e16b32cc318edb9f2c3ba21032384d2337d75ce12c7189fdfR3) enabled on the file; otherwise the `string` would be automatically coerced to an `int` when passed into the submission repository's `get` function.

We don't yet have a set of development patterns that will gracefully allow us to adopt strict typing -- so rather than peppering the code with manual typecasts when errors arise, I'd rather just disable strict typing.

asmecher:
Thanks, @touhidurabir! I just added [one last comment](https://github.com/pkp/pkp-lib/pull/9213#discussion_r1295046909); please let me know when you're ready for me to take another look.

touhidurabir:
@asmecher please take another look. if all ok, should be good to merge .

Issue #9214, opened 2023-08-08
Adding support for github actions for PKP applications


withanage:
#  Github Actions support for PKP applications

|Application | Unittests | Cypress integration tests |  validation tests  |  Upgrade tests |
| ---- | ----| ----| ---- | ----|
| OJS | ||||
| OPS | ||||
| OMP | ||||

PR #9216, opened 2023-08-08
pkp/pkp-lib#9214 Cypress move Cypress.env('baseUrl') to Cypress.config('baseUrl')

withanage:
#### Issue
https://github.com/pkp/pkp-lib/issues/9214

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -284,7 +284,7 @@ Cypress.Commands.add('addSubmissionAuthorsWithApi', (api, data, csrfToken) => {
 });
 
 Cypress.Commands.add('submitSubmissionWithApi', (id, csrfToken) => {
-	const api = new Api(Cypress.env('baseUrl') + '/index.php/publicknowledge/api/v1');
+	const api = new Api(Cypress.config('baseUrl') + '/index.php/publicknowledge/api/v1');
 	return cy.get('@submissionId').then((submissionId) => {
 		cy.request({
 			url: api.submit(submissionId),


cypress/support/commands.js
```diff
@@ -284,7 +284,7 @@ Cypress.Commands.add('addSubmissionAuthorsWithApi', (api, data, csrfToken) => {
 });
 
 Cypress.Commands.add('submitSubmissionWithApi', (id, csrfToken) => {
-	const api = new Api(Cypress.env('baseUrl') + '/index.php/publicknowledge/api/v1');
+	const api = new Api(Cypress.config('baseUrl') + '/index.php/publicknowledge/api/v1');
```
asmecher, line 287: I think this change also needs to be made in `cypress/tests/integration/SubmissionWizard.cy.js`.

asmecher: Would this suggest that local developers will need to customize `cypress.config.js`? Also, should some kind of default be added to `cypress.travis.env.json`? These files serve as templates for developers to use.

asmecher (commented):
Thanks, Dulip! A couple of comments on this.

cypress/support/commands.js
withanage: > I think this change also needs to be made in cypress/tests/integration/SubmissionWizard.cy.js.
sure, cypress/support/commands.js had also a little, will pack them in one-commit.


> Would this suggest that local developers will need to customize cypress.config.js?  Also, should some kind of default be added to cypress.travis.env.json? 

we  do not need the cypress.config.js file copying , as  in the current travis integration. We can use the cypress variables for this purpose. Their defalut github action will run with no manual intervention. It is not in a PR, but https://github.com/withanage/ojs/commit/035a7c0c51ed6ad943df0b926594bc44c5472000#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R45 here is the current status.

asmecher: OK, and will local developers need to change their setups to work with Cypress?

withanage: @asmecher
> will local developers need to change their setups to work with Cypress?

No they can use their older cypress.env.json  as  they have been using.

Issue #9136, opened 2023-07-07
[Beacon] Missing uniqueSiteId on new installations


jonasraoni:
**Describe the bug**
The `uniqueSiteId`, which is used for disambiguation purposes on the Beacon, comes from the `UsageEventPlugin`.
The statistics code, which used to be centralized on this plugin, suffered a major rewrite in the latest release, and the code which sets the `uniqueSiteId` is probably unreachable by now.

On a sidenote, is the `UsageEventPlugin` plugin even needed at this point?

**Solution:**
- The Beacon must be updated to accept an empty installation ID and any requirement over this ID must be relaxed in order to avoid losing input.
- Migrate existing `uniqueSiteId` settings to the `site_settings` table.
- Implement its retrieval through a getter (`Application::getGUID()`?!), which should set a new value if empty.
- Old values must be kept unchanged, new ones can make use of the `PKPString::generateUUID()` (and close this discussion: https://github.com/pkp/pkp-lib/discussions/7988).
- Given the Beacon is a global setting, inspect the feasibility of forwarding also the OAI URL for the whole installation.
- We have indirect dependencies over the package `ramsey/uuid`, we can promote it to a direct dependency, and call it at the `PKPString::generateUUID()` to get rid of some code.

**To Reproduce**
1. Access a fresh installation installed from zero
2. Create a context
3. Run the query `SELECT * FROM plugin_settings WHERE setting_name = 'uniqueSiteId'`
4. There will be an empty `uniqueSiteId` setting (which comes from the `version.xml`).

**What application are you using?**
OJS 3.4.0-1

<hr>

PRs:

stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9229 
- https://github.com/pkp/ojs/pull/4005
- https://github.com/pkp/omp/pull/1442
- https://github.com/pkp/ops/pull/556

main:
- https://github.com/pkp/pkp-lib/pull/9232
- https://github.com/pkp/ojs/pull/4007
- https://github.com/pkp/omp/pull/1443
- https://github.com/pkp/ops/pull/557

asmecher:
@bozana, could you have a look at this? Essentially we used to have a site setting called `uniqueSiteId` that was used to disambiguate installations for stats purposes, but was also provided to the beacon to help with disambiguation there. It's no longer showing up with the beacon since the stats rewrite happened.

bozana:
[...]

PR #9229, opened 2023-08-14
pkp/pkp-lib#9136 introduce new and migrate old uniqueSiteId

bozana:
s. https://github.com/pkp/pkp-lib/issues/9136

### DIFF
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -294,7 +294,16 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
      */
     public function getUUID(): string
     {
-        return $this->getRequest()->getSite()->getUniqueSiteID();
+        $site = $this->getRequest()->getSite();
+        $uniqueSiteId = $site->getUniqueSiteID();
+        if (!strlen((string) $uniqueSiteId)) {
+            $uniqueSiteId = PKPString::generateUUID();
+            $site->setUniqueSiteID($uniqueSiteId);
+            /** @var SiteDAO */
+            $siteDao = DAORegistry::getDAO('SiteDAO');
+            $siteDao->updateObject($site);
+        }
+        return $uniqueSiteId;
     }
 
     /**
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -290,6 +290,14 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     }
 
     /**
+     * Get the unique site ID
+     */
+    public function getUUID(): string
+    {
+        return $this->getRequest()->getSite()->getUniqueSiteID();
+    }
+
+    /**
      * Return a HTTP client implementation.
      *
      * @return Client
--- a/classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
+++ b/classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
  *
- * Copyright (c) 2014-2023 Simon Fraser University
- * Copyright (c) 2000-2023 John Willinsky
+ * Copyright (c) 2023 Simon Fraser University
+ * Copyright (c) 2023 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class I9136_MigrateUniqueSiteId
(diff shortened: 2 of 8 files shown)


classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
```diff
@@ -0,0 +1,59 @@
...
+            ->where('setting_name', '=', 'uniqueSiteId')
+            ->value('setting_value');
+        if (is_null($newUniqueSiteId)) {
+            $uniqueSiteId = DB::table('plugin_settings')
+                ->where('plugin_name', '=', 'usageeventplugin')
+                ->where('setting_name', '=', 'uniqueSiteId')
+                ->value('setting_value')
+                ??
+                PKPString::generateUUID();
+            DB::table('site_settings')->insert([
```
jonasraoni, line 38: I don't think it's going to fail, but an `updateOrInsert()` would be safer here.

classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $newUniqueSiteId = DB::table('site_settings')
+            ->where('setting_name', '=', 'uniqueSiteId')
+            ->value('setting_value');
+        if (is_null($newUniqueSiteId)) {
+            $uniqueSiteId = DB::table('plugin_settings')
+                ->where('plugin_name', '=', 'usageeventplugin')
+                ->where('setting_name', '=', 'uniqueSiteId')
+                ->value('setting_value')
+                ??
```
jonasraoni, line 36: I also don't think we'll see such case, but sounds safer to use `?:`

classes/site/Site.php
```diff
@@ -235,6 +235,22 @@ public function setSupportedLocales($supportedLocales)
     {
         $this->setData('supportedLocales', $supportedLocales);
     }
+
+    /**
+    * Get the unique site ID.
+    */
+    public function getUniqueSiteID(): string
+    {
+        return $this->getData('uniqueSiteId');
```
jonasraoni, line 244: Even though the migration should handle things properly, I'd keep the pattern from the previous code for safety reasons 😁
I mean the part that checks whether the `uniqueSiteId` value is falsy, then regenerate it on the fly.

classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
+ *
+ * Copyright (c) 2014-2023 Simon Fraser University
+ * Copyright (c) 2000-2023 John Willinsky
```
jonasraoni, line 7: Looks like we're using another pattern for new files: `Copyright (c) 2023 Simon Fraser University`

jonasraoni (approved):
Everything makes sense for me, I just left minor comments =]

classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
bozana: Hmmm... if there is already uniqueSiteId setting then it should not be overwritten, correct? So maybe it is better for the upgrade script to fail here?

classes/site/Site.php
bozana: OK...
I then changed this function so that it returns `?string` and consider that check in the PKPApplication::getUUID()

classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
jonasraoni: If it's falsy, I think it's ok to overwrite... But as this situation isn't going to happen anyway, anything is ok 🙂

classes/core/PKPApplication.php
```diff
@@ -294,7 +294,16 @@ public static function get()
      */
     public function getUUID(): string
     {
-        return $this->getRequest()->getSite()->getUniqueSiteID();
+        $site = $this->getRequest()->getSite();
+        $uniqueSiteId = $site->getUniqueSiteID();
+        if (is_null($uniqueSiteId) || strlen($uniqueSiteId) == 0) {
```
jonasraoni, line 299: Another way could be `if (!strlen((string) $uniqueSiteId)))`, but that's just my personal preference :)

jonasraoni (approved):

classes/migration/upgrade/v3_4_0/I9136_MigrateUniqueSiteId.php
bozana: How we could check if it is falsy, what would be a falsy ID?

jonasraoni: If the value is empty/null. But this condition isn't going to happen here, so it's not needed to change anything :)

I just think the `updateOrInsert()` is the best choice when there are unique keys inside the entity, to avoid any kind of unexpected condition/failure.

bozana: Thanks @jonasraoni. I will then leave it as it is for now...

Issue #9227, opened 2023-08-14
[OJS-3.4.0-1] Incorrect (encrypted) password sent to the user after the user was added by Admin/Journal Manager


shantanu198713:
OJS-3.4.0-1
PHP-8.1

Incorrect (encrypted) password sent to the user after the user was added by Admin/Journal Manager.
**I think the registration portal URL can be included in the mail.**

![Screenshot 2023-08-14 101858](https://github.com/pkp/pkp-lib/assets/108086810/89e998d3-8b5b-4412-a8f6-934dd8767447)

touhidurabir:
@Vitaliy-1 PRs to review 

[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9230
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4006 [TEST ONLY]

we probably need to port the changes to `main` branch and user export import plugin also .

touhidurabir:
@Vitaliy-1 can you check again ? Updated the PR based in review .

Vitaliy-1:
PR pkp-lib main: https://github.com/pkp/pkp-lib/pull/9233

Vitaliy-1:
Thanks, merged!

PR #9230, opened 2023-08-14
pkp/pkp-lib#9227 Corrent password value in user registration email

touhidurabir:
for pkp/pkp-lib#9227

### DIFF
--- a/classes/mail/mailables/UserCreated.php
+++ b/classes/mail/mailables/UserCreated.php
@@ -49,9 +49,12 @@ class UserCreated extends Mailable
     ];
     protected static ?string $variablePassword = 'password';
 
-    public function __construct(Context $context)
+    protected ?string $password;
+
+    public function __construct(Context $context, string $password = null)
     {
-        parent::__construct(func_get_args());
+        parent::__construct([$context]);
+        $this->password = $password;
     }
 
     /**
@@ -79,7 +82,7 @@ class UserCreated extends Mailable
     {
         $this->traitRecipients([$recipient], $locale);
         $this->addData([
-            static::$variablePassword => $recipient->getPassword()
+            static::$variablePassword => $this->password ?? $recipient->getPassword()
         ]);
 
         return $this;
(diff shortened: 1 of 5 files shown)


classes/mail/mailables/UserCreated.php
```diff
@@ -79,7 +82,7 @@ public function recipients(User $recipient, ?string $locale = null): Mailable
     {
         $this->traitRecipients([$recipient], $locale);
         $this->addData([
-            static::$variablePassword => $recipient->getPassword()
+            static::$variablePassword => $this->password ?? $recipient->getPassword()
```
Vitaliy-1, line 85: It looks that User::getPassword() returns encrypted password. As you passing the original password to the constructor, you can remove `$recipient->getPassword()`.

classes/mail/mailables/UserCreated.php
```diff
@@ -49,9 +49,12 @@ class UserCreated extends Mailable
     ];
     protected static ?string $variablePassword = 'password';
 
-    public function __construct(Context $context)
+    protected ?string $password;
+
+    public function __construct(Context $context, string $password = null)
```
Vitaliy-1, line 54: Let's make the `$password` argument mandatory and add it when the mailable is instantiated, e.g., in `UserXmlPKPUserFilter::parseUser()`

Vitaliy-1 (changes requested):
Thanks, @touhidurabir! Can you also fix this in `ReviewerRegister` mailable too?

classes/mail/mailables/UserCreated.php
Vitaliy-1: And we need to escape it with `htmlspecialchars()`

touhidurabir: Done .

classes/mail/mailables/UserCreated.php
touhidurabir: Done .

touhidurabir:
> also fix this in ReviewerRegister mailable too?

@Vitaliy-1 done .

Vitaliy-1 (approved):

PR #9245, opened 2023-08-21
Change submission language

jyhein:
- Change the submission language after submission
- Show the language of the submission in the workflow page and the author dashboard
-  In the modal window editors can change the language and add/edit required metadata
- The submission's metadata can be edited even if the submission's locale is not one of the currently supported submission locales
- The language of the submission can not be changed if the status of the publication is published or there are more than one version

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -17,6 +17,7 @@
 
 namespace PKP\API\v1\submissions;
 
+use APP\author\Author;
 use APP\core\Application;
 use APP\facades\Repo;
 use APP\mail\variables\ContextEmailVariable;
@@ -65,6 +66,7 @@ use PKP\services\PKPSchemaService;
 use PKP\submission\GenreDAO;
 use PKP\submission\PKPSubmission;
 use PKP\submission\reviewAssignment\ReviewAssignment;
+use PKP\submissionFile\SubmissionFile;
 use PKP\userGroup\UserGroup;
 
 class PKPSubmissionController extends PKPBaseController
@@ -84,6 +86,7 @@ class PKPSubmissionController extends PKPBaseController
         'saveForLater',
         'submit',
         'delete',
+        'changeLocale',
         'getGalleys',
         'getDecisions',
         'getParticipants',
@@ -116,6 +119,7 @@ class PKPSubmissionController extends PKPBaseController
         'deleteContributor',
         'editContributor',
         'saveContributorsOrder',
+        'changeLocale',
     ];
 
     /** @var array Handlers that must be authorized to access a submission's production stage */
@@ -218,6 +222,10 @@ class PKPSubmissionController extends PKPBaseController
             Route::delete('{submissionId}', $this->delete(...))
                 ->name('submission.delete')
                 ->whereNumber('submissionId');
+
+            Route::put('{submissionId}/publications/{publicationId}/changeLocale', $this->changeLocale(...))
+                ->name('submission.publication.changeLocale')
+                ->whereNumber(['submissionId', 'publicationId']);
         });
 
         Route::middleware([
@@ -865,6 +873,59 @@ class PKPSubmissionController extends PKPBaseController
         Repo::submission()->delete($submission);
 
         return response()->json($submissionProps, Response::HTTP_OK);
+    }
+
+    /**
+     * Change submission language
+     */
+    public function changeLocale(Request $illuminateRequest): JsonResponse
+    {
+        $publication = Repo::publication()->get((int) $illuminateRequest->route('publicationId'));
+
+        if (!$publication) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $publication->getData('submissionId')) {
+            return response()->json([
+                'error' => __('api.publications.403.submissionsDidNotMatch'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $paramsSubmission = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_SUBMISSION, $illuminateRequest->input());
+        $newLocale = $paramsSubmission['locale'] ?? null;
+
+        // Submission language can not be changed when there are more than one publication or a publication's status is published
+        if (!$newLocale || count($submission->getData('publications')) > 1 || $publication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
+            return response()->json(['error' => __('api.submission.403.cantChangeSubmissionLanguage')], Response::HTTP_FORBIDDEN);
+        }
+
+        // Convert a form field value to multilingual (if it is not) and merge rest values
+        collect(app()->get('schema')->getMultilingualProps(PKPSchemaService::SCHEMA_PUBLICATION))
+            ->each(fn (string $prop) =>
+                $illuminateRequest->whenHas($prop, fn ($value) =>
+                    $illuminateRequest->merge([
+                        $prop => array_merge(
+                            $publication->getData($prop) ?? [],
+                            (is_array($value) && array_key_exists($newLocale, $value)) ? $value : [$newLocale => $value]
+                        )
+                    ])
+                )
+            );
+
+        $responsePublication = $this->editPublication($illuminateRequest);
+
+        if ($responsePublication->status() !== 200) {
+            return $responsePublication;
+        }
+
+        $this->copyMultilingualData($submission, $newLocale);
+
+        return $this->edit($illuminateRequest);
     }
 
     /**
@@ -1986,4 +2047,40 @@ class PKPSubmissionController extends PKPBaseController
         ];
     }
 
+    /**
+     * Copy author, files, etc. multilingual fields from old to new changed language
+     */
+    protected function copyMultilingualData(Submission $submission, string $newLocale): void {
+        $oldLocale = $submission->getData('locale');
+        $editProps = fn (Author|SubmissionFile $item, array $props): array => collect($props)
+            ->mapWithKeys(fn (string $prop): array => [$prop => ($data = $item->getData($prop)[$oldLocale] ?? null) ? [$newLocale => $data] : null])
+            ->filter()
+            ->toArray();
+
+        // Submission files
+        $fileProps = [
+            'name',
+        ];
+        Repo::submissionFile()
+            ->getCollector()
+            ->filterBySubmissionIds([$submission->getId()])
+            ->getMany()
+            ->each(fn (SubmissionFile $file) => Repo::submissionFile()->edit($file, $editProps($file, $fileProps)));
+        
+        // Contributor
+        $contributorProps = [
+            'givenName',
+            'familyName',
+            'preferredPublicName',
+        ];
+        Repo::author()
+            ->getCollector()
+            ->filterByPublicationIds([$submission->getLatestPublication()->getId()])
+            ->getMany()
+            ->each(function (Author $contributor) use ($contributorProps, $editProps, $newLocale) {
+                if (!($contributor->getData('givenName')[$newLocale] ?? null)) {
+                    Repo::author()->edit($contributor, $editProps($contributor, $contributorProps));
+                }
+            });
+    }
 }
--- a/classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
+++ b/classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
@@ -1,0 +1,118 @@
+<?php
+/**
+ * @file classes/components/form/submission/ChangeSubmissionLanguageMetadataForm.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class ChangeSubmissionLanguageMetadataForm
+ *
+ * @brief A form for changing submission's metadata before language change.
+ */
+
+namespace PKP\components\forms\submission;
+
+use APP\facades\Repo;
+use APP\publication\Publication;
+use APP\submission\Submission;
+use PKP\components\forms\Field;
+use PKP\components\forms\FieldHTML;
+use PKP\components\forms\FieldRadioInput;
+use PKP\components\forms\FormComponent;
+use PKP\components\forms\publication\TitleAbstractForm;
+use PKP\context\Context;
+use PKP\facades\Locale;
+
+class ChangeSubmissionLanguageMetadataForm extends FormComponent
+{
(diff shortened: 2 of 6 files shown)


classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        $this->action = $action;
+        $this->locales = $locales;
+
+        $this->addGroup([
+            'id' => 'lang',
+        ]);
+
+        $submissionLocale = $submission->getData('locale');
+        $supportedLocaleNames = collect($context->getSupportedSubmissionLocaleNames());
```
bozana, line 45: Shouldn't here publication->getLanguageNames() or submission->getPublicationLanguageNames() be used?

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
```diff
@@ -0,0 +1,120 @@
+<?php
+/**
+ * @file classes/components/form/submission/ChangeSubmissionLanguageMetadataForm.php
+ *
+ * Copyright (c) 2023 Simon Fraser University
+ * Copyright (c) 2023 John Willinsky
```
bozana, line 6: 2024 :-)

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
jyhein: I'm not sure, shouldn't this be the same list as in submission wizard where you choose the submission locale ?

bozana: Hmmm... I am not sure as well :-D
I think it is a rare case, but: Lets say an author entered a submission with a submission full text file in Finish, and filled out metadata in German and English, and has chosen wrongly English (instead of Finish) as primary locale. Then the JM removes Finish from submission languages. Then an editor wants to fix the submission language for this submission and to chose Finish. It would not be possible, right?
It is maybe slightly different than submission wizard, because the submission could exist in the system for a while (having the possibility that submission languages have changed in the mean time), before the editor realizes that the submission primary language is wrong, no?

bozana: Lets leave it as it is for now...

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -612,6 +612,11 @@ public function edit(Request $illuminateRequest): JsonResponse
 
         $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_SUBMISSION, $illuminateRequest->input());
 
+        // Submission language can not be changed when there are more than one publication or a publication's status is published
+        if (count($submission->getData('publications')) > 1 || $submission->getLatestPublication()->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
```
bozana, line 616: Hmmm... This edit can be called also for other changes, not only the submission language, so this condition should not be applied in general, I think

locale/en/submission.po
```diff
@@ -2043,6 +2043,28 @@ msgstr "You must assign at least one participant to this submission before initi
...
+
+msgid "submission.list.changeSubmissionLanguage.title"
+msgstr "Change Submission Language For"
+
+msgid "submission.list.changeSubmissionLanguage.languageDescription"
+msgstr "This is the primary submission language. Changing this will have effects on the submission and the metadata entered"
+
+msgid "submission.list.changeSubmissionLanguage.metadataDescription"
+msgstr "<strong>Before changing the submission language, ensure you have filled out the following metadata fields to maintain system integrity. "
+"Also, note that contributor information and file names will be copied from previously entered information</strong>"
```
bozana, line 2060: There is a full dot '.' here at the end.

api/v1/submissions/PKPSubmissionController.php
jyhein: Yes, forgot to add `if (isset($params['locale']) && ...`

locale/en/api.po
```diff
@@ -272,6 +272,9 @@ msgstr "The requested volume, number or year is not valid."
 msgid "api.submissions.400.invalidSubmitAs"
 msgstr "You are not allowed to submit in this user role."
 
+msgid "api.submission.403.cantChangeSubmissionLanguage"
```
ewhanson, line 275: This isn't really a comment about the locale key, but why can't the submission locale be changed if there's more than one publication version? I can think of a few, but I'm wondering if there was a specific technical/editorial process reason for limiting it like this?

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        $this->action = $action;
+
+        $this->addGroup([
+            'id' => 'language',
+        ]);
+
+        $submissionLocale = $submission->getData('locale');
+        $supportedLocaleNames = collect($context->getSupportedSubmissionLocaleNames());
+        $submissionLocaleNameArr = collect(Locale::getSubmissionLocaleDisplayNames([$submissionLocale]));
```
ewhanson, line 43: I know it's small, but having this appended with `Arr` at the end makes it seem like it's different from the `$supportedLocaleNames` Eloquent collection. Especially with all the transformations going on, I would suggest keeping the naming conventions the same and calling this $submissionLocaleNames` or something like that.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -819,6 +827,50 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+        $newLocale = $paramsSubmission['locale'] ?? null;
+
+        // Submission language can not be changed when there are more than one publication or a publication's status is published
+        if (!$newLocale || count($submission->getData('publications')) > 1 || $publication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
+            return response()->json(['error' => __('api.submission.403.cantChangeSubmissionLanguage')], Response::HTTP_FORBIDDEN);
+        }
+
+        // Convert a form field value to multilingual (if it is not) and merge rest values
+        collect(app()->get('schema')->getMultilingualProps(PKPSchemaService::SCHEMA_PUBLICATION))
+            ->each(fn (string $p) => $illuminateRequest->whenHas($p, fn ($v) => $illuminateRequest->merge([$p => array_merge($publication->getData($p) ?? [], is_array($v) ? $v : [$newLocale => $v])])));
```
ewhanson, line 861: With this using deeply nested arrow functions, the single-letter variables and everything being on one line make it kind of difficult to parse what's going on. I would recommend more explicit variable names and possibly breaking the `each` into more than one line for readability.

ewhanson, line 861: I don't have any immediate solution or recommendation, but just wanted to flag it here while reviewing the rest: I'm a little unsure about directly modifying the incoming request in order to complete this work, especially with the approach of passing on the request to two further API endpoints. I agree that this is a good solution to not having to rewrite code that already exist, but I'm not sure if this is a pattern we'd want to adopt/encourage. I'll do a bit of thinking and come back to it.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1706,4 +1758,41 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+            'name',
+        ];
+        Repo::submissionFile()
+            ->getCollector()
+            ->filterBySubmissionIds([$submission->getId()])
+            ->getMany()
+            ->each(fn (SubmissionFile $f) => Repo::submissionFile()->edit($f, $editProps($f, $fileProps)));
+        
+        // Contributor
+        $contributorProps = [
```
ewhanson, line 1783: There are other localized fields such as bio and affiliation. Just wanted to flag that to make sure those shouldn't be copied over as well.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1706,4 +1758,41 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+			'givenName',
```
ewhanson, line 1784: This is the only place I've seen it so far, but this is using tabs when the rest of the codebase here is using spaces. I'd double check all the additions to make sure they're using the same code styling as the specific repo.

locale/en/submission.po
```diff
@@ -2043,6 +2043,28 @@ msgstr "You must assign at least one participant to this submission before initi
...
+msgstr "Current Submission Language: <strong>{$language}</strong>"
+
+msgid "submission.list.changeSubmissionLanguage.buttonLabel"
+msgstr "Change"
+
+msgid "submission.list.changeSubmissionLanguage.title"
+msgstr "Change Submission Language For"
+
+msgid "submission.list.changeSubmissionLanguage.languageDescription"
+msgstr "This is the primary submission language. Changing this will have effects on the submission and the metadata entered"
```
ewhanson, line 2056: This and a few others don't have periods at the end of sentences. I think it's fine when it's just a word or phrase, but this and the others below should probably have them.

locale/en/submission.po
```diff
@@ -2430,3 +2452,5 @@ msgstr "Competing Interests"
 msgid "author.competingInterests.description"
 msgstr "Please disclose any competing interests this author may have with the research subject."
 
+msgid "submission.localeNotSupported"
+msgstr "The language of the submission ({$language}) is not one of the supported submission languages. You can still edit the submission details but are not currently accepting new submissions with this language."
```
ewhanson, line 2456: There's a missing word here at `but [missing word] are not currently accepting`.

locale/en/submission.po
```diff
@@ -2043,6 +2043,28 @@ msgstr "You must assign at least one participant to this submission before initi
 msgid "submission.query.allowedEditTime"
 msgstr "You can update this discussion for {$allowedEditTimeNoticeLimit} minutes."
 
+msgid "submission.list.changeSubmissionLanguage.currentLanguage"
```
ewhanson, line 2046: Just flagging this as another place where the workflow/author dashboard styling is happening.

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
```diff
@@ -0,0 +1,118 @@
...
+    protected function getTitleAbstractForm(string $publicationApiUrl, array $locales, Publication $publication, Context $context): TitleAbstractForm
+    {
+        $pubSecId = $publication->getData('sectionId');
+        $section = $pubSecId ? Repo::section()->get($pubSecId, $context->getId()) : null;
+        return new TitleAbstractForm(
+            $publicationApiUrl,
+            $locales,
+            $publication,
+            $section ? (int) $section->getData('wordCount') : 0,
+            $section && !$section->getData('abstractsNotRequired')
```
jardakotesovec, line 103: If I follow correctly, you create TitleAbstractForm to cherry-pick the title and abstract fields from it? And during cherry-picking you also convert it from multilingual to not multilingual, right? Would not be easier just to add title and abstract fields directly to your form and not using the TitleAbstractForm?

api/v1/submissions/PKPSubmissionController.php
jardakotesovec: @ewhanson Just a thought on this and also not sure how its applicable to our existing code base. But one strategy to take is to keep the API handlers very slim&lightweight, with just API specific things like validation. And than having business logic in other application layers, where that logic can be invoked from anywhere and its easier to reuse.

ewhanson: Thanks @jardakotesovec. I like that idea in general and I think this use case would be a good candidate for testing that out. I think that kind of change would be outside the scope of this issue so in the interest in keeping things moving, I would recommend keeping this as-is for this issue, and creating a new issue to look into implementing that with this as a first test case.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -819,6 +827,59 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+        }
+
+        // Convert a form field value to multilingual (if it is not) and merge rest values
+        collect(app()->get('schema')->getMultilingualProps(PKPSchemaService::SCHEMA_PUBLICATION))
+            ->each(fn (string $prop) =>
+                $illuminateRequest->whenHas($prop, fn ($value) =>
+                    $illuminateRequest->merge([
+                        $prop => array_merge(
+                            $publication->getData($prop) ?? [],
+                            (is_array($value) && array_key_exists($newLocale)) ? $value : [$newLocale => $value]
```
bozana, line 866: array_key_exists misses the second parameter

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1706,4 +1767,41 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+            'familyName',
+            'preferredPublicName',
+        ];
+        Repo::author()
+            ->getCollector()
+            ->filterByPublicationIds([$submission->getLatestPublication()->getId()])
+            ->getMany()
+            ->each(function (Author $c) use ($contributorProps, $editProps, $newLocale) {
+                if (!($c->getData('givenName')[$newLocale] ?? null)) {
+                    Repo::author()->edit($c, $editProps($c, $contributorProps));
```
bozana, line 1803: variable name $c

locale/en/submission.po
bozana: This should be so according to some design rules -- I have already asked Devika :-)

api/v1/submissions/PKPSubmissionController.php
bozana: I forgot why exactly, but we decided back then to only copy the minimal needed information

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
bozana: @jyhein, I am also wondering why have you chosen this way? Something specific?
It seems it is generic right now, how you have solved it, in case we need consider additional field, but... just curious...

locale/en/api.po
bozana: We realized that having versions would mean lots more to consider, and that the main use case actually happens before the publication. The main use case when this functionality is really needed is when author wrongly chooses the submission locale at the submission. Editors would mostly realize this immediately, and the fix should happen before the actual publication.

locale/en/submission.po
bozana: @ewhanson, what do you exactly mean?

ewhanson: Sorry, @bozana. It's actually in the line below. The styling for how it appears in the UI is applied via a `<strong>` tag that's baked into the localized string.

bozana: @ewhanson, how should this be actually solved? Having different span element with different classes and having only 'Current Submissino Language:' in the locale key? :-)
Thanks!

classes/components/forms/submission/ChangeSubmissionLanguageMetadataForm.php
jyhein: Currently there are only 2 fields but if more are added, then it is easier this way than repeating the code from forms. This is same if added new fields from forms, e.g. PKPMetadataForm

PR #9257, opened 2023-08-25
pkp/pkp-lib#7916 Added core PKPHtmlSanitizer class

touhidurabir:
patch for pkp/pkp-lib#7916

### DIFF
--- a/classes/core/PKPHtmlSanitizer.php
+++ b/classes/core/PKPHtmlSanitizer.php
@@ -1,0 +1,144 @@
+<?php
+
+/**
+ * @file classes/core/PKPHtmlSanitizer.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class PKPHtmlSanitizer
+ *
+ * @ingroup core
+ *
+ * @brief Wrapper on the top of Symfony's HtmlSanitizer implementation
+ *
+ */
+
+namespace PKP\core;
+
+use DOMDocument;
+use Illuminate\Support\Str;
+use Illuminate\Support\Collection;
+use PKP\config\Config;
+use Symfony\Component\HtmlSanitizer\HtmlSanitizer;
+use Symfony\Component\HtmlSanitizer\HtmlSanitizerConfig;
+
+class PKPHtmlSanitizer
+{
+    /**
+     * Collection of allowed tags to allowed attributes map as key/value[array] structure
+     */
+    protected Collection $allowedTagToAttributeMap;
+
+    /**
+     * Instance of HtmlSanitizerConfig config
+     */
+    protected ?HtmlSanitizerConfig $htmlSanitizerConfig = null;
+
+    /**
+     * Create a new instance
+     */
+    public function __construct(string $allowable)
+    {
+        $this->allowedTagToAttributeMap = $this->generateAllowedTagToAttributeMap(Config::getVar('security', $allowable, null) ?? $allowable);
+    }
+
+    /**
+     * Set a pre defined instance of \Symfony\Component\HtmlSanitizer\HtmlSanitizerConfig
+     */
+    public function setSanitizerConfig(HtmlSanitizerConfig $htmlSanitizerConfig): self
+    {
+        $this->htmlSanitizerConfig = $htmlSanitizerConfig;
+
+        return $this;
+    }
+
+    /**
+     * Get the instance of \Symfony\Component\HtmlSanitizer\HtmlSanitizerConfig
+     */
+    public function getSanitizerConfig(): HtmlSanitizerConfig
+    {
+        if (!$this->htmlSanitizerConfig) {
+            $this->buildSanitizerConfig();
+        }
+
+        return $this->htmlSanitizerConfig;
+    }
+
+    /**
+     * Sanitize the given html string
+     */
+    public function sanitize(string $html): string
+    {   
+        $config = clone $this->getSanitizerConfig();
+
+        $this
+            ->getAllNotAllowedHtmlTags($html)
+            ->each(function(string $tag) use (&$config) {
+                $config = $config->blockElement($tag);
+            });
+        
+        return (new HtmlSanitizer($config))->sanitize($html);
+    }
+
+    /**
+     * Build up the HtmlSanitizerConfig instance if no predefined instance provided
+     */
+    protected function buildSanitizerConfig(): void
+    {
+        $this->htmlSanitizerConfig = (new HtmlSanitizerConfig())
+            ->allowLinkSchemes(['https', 'http', 'mailto'])
+            ->allowMediaSchemes(['https', 'http']);
+        
+        $this->allowedTagToAttributeMap->each(
+            fn (array $attributes, string $tag) => $this->htmlSanitizerConfig = $this->htmlSanitizerConfig->allowElement($tag, $attributes)
+        );
+    }
+
+    /**
+     * Get the collection of non allowed tags in the given html srting
+     */
+    protected function getAllNotAllowedHtmlTags(string $html): Collection
+    {
+        $dom = new DOMDocument();
+
+        $errorState = libxml_use_internal_errors(true); // Don't generate warnings on malformed HTML
+        $dom->loadHTML($html);
+        libxml_use_internal_errors($errorState); // Restore the error state to its previous value
+
+        return collect($dom->getElementsByTagName('*'))
+            ->map(fn ($child) => $child->nodeName)
+            ->diff($this->allowedTagToAttributeMap->keys());
+    }
+
+    /**
+     * Generate the collection of allowed tags to allowed attributes map as key/value[array] structure
+     */
+    protected function generateAllowedTagToAttributeMap(string $allowable): Collection
+    {
+        return Str::of($allowable)
+            ->explode(',')
+            ->mapWithKeys(function(string $allowedTagWithAttr) {
+                
+                // Extract the tag itself (e.g. div, p, a ...)
+                preg_match('/\[[^][]+]\K|\w+/', $allowedTagWithAttr, $matches);
+                $allowedTag = collect($matches)->first();
+
+                // Extract the attributes associated with tag (e.g. class, href ...)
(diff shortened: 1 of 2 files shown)


classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Sanitize the given html string
+     */
+    public function sanitize(string $html): string
+    {   
+        $config = clone $this->getSanitizerConfig();
+
+        $this
+            ->getAllNotAllowedHtmlTags($html)
```
asmecher, line 77: We only have two use cases for `PKPHtmlSanitizer` currently, right?

1. When sanitizing a snippet of HTML that has generally been entered by TinyMCE
2. When getting a small piece out of an entire HTML / XML document (as in the two plugins)

Probably 99% of our usage is going to be the first case; using `getAllNotAllowedHtmlTags` is going to incur an entire DOM load for each of these cases, when it's not generally relevant to that use case. And the load could be substantial, as `|strip_unsafe_html` is used in many front-end templates.

I think we might need to give up trying to solve those use cases separately. And in fact they're different enough that I think it's justified; in 1) we're sanitizing user-entered HTML, and in 2) we're extracting part of a JATS XML document.

With use case 2 removed from consideration, is there still justification for the `PKPHtmlSanitizer` class?

touhidurabir: > going to incur an entire DOM load for each of these cases, when it's not generally relevant to that use case. And the load could be substantial

agreed and better to avoid that . 

> is there still justification for the PKPHtmlSanitizer class

I still think we should have one default sanitization process rather than multiple just because one is more frequently used than other . 

@asmecher I think my previous way to thinking for this implementation was flawed . Basically we need to work with W3C sanitization specification and and let other things handle differently . This newer implementation of `PKPHtmlSanitizer` use the following approach 
1. Build up allowed list
2. cross match with W3C provided list and build up the non allowed list (also include dangerous tags e.g. `script`,  etc)
3. any other tags not in the combined list of allowed and non-allowed list, should be stripped . 

This make it much better and faster to work with and still we can maintain a proper standard specify via W3C sanitization API . 

more about the default configuration at https://wicg.github.io/sanitizer-api/#default-configuration . 

can you review it again ?

classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,159 @@
...
+    {
+        return $this->htmlSanitizerConfig;
+    }
+
+    /**
+     * Sanitize the given html string
+     */
+    public function sanitize(string $html): string
+    {   
+        return (new HtmlSanitizer($this->htmlSanitizerConfig))->sanitize(
```
asmecher, line 81: @touhidurabir, is the cost of the sanitizer in building the configuration (cached) or instantiating the sanitizer (not cached)? If the `HtmlSanitizer` is intended to be a reusable object, would it make sense to store it as a property in this class?

classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,159 @@
...
+     */
+    protected function buildSanitizerConfig(Collection $allowedTagToAttributeMap): void
+    {
+        $this->htmlSanitizerConfig = (new HtmlSanitizerConfig())
+            ->allowLinkSchemes(['https', 'http', 'mailto'])
+            ->allowMediaSchemes(['https', 'http']);
+        
+        if ($allowedTagToAttributeMap->count()) {
+            $allowedTagToAttributeMap->each(
+                fn (array $attributes, string $tag) => $this->htmlSanitizerConfig = $this->htmlSanitizerConfig->allowElement($tag, $attributes)
```
asmecher, line 98: Shouldn't `$this->htmlSanitizerConfig` be appended to rather than reinitialized on multiple collection elements?

classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,159 @@
...
+        return $this->w3cValidElements
+            ->merge(collect($this->htmlSanitizerConfig->getAllowedElements())->keys())
+            ->merge(collect([ // list of dangerous tags that should be only handled by sanitization library
+                'script',
+            ]))
+            ->unique();
+    }
+
+    /**
+     * Generate the collection of allowed tags to allowed attributes map as key/value[array] structure
```
asmecher, line 131: Needs self-documentation on parameter and return; they aren't clear to me at a glance over the function definition.

classes/core/PKPHtmlSanitizer.php
touhidurabir: The main cost is the configure the sanitizer config as on each call of `allowElement/blockElement/...` , it clone the current object . The instantiating of the `HtmlSanitizer` only take the config object and a parser or set the default one of `Symfony\Component\HtmlSanitizer\Parser\MastermindsParser` with empty constructor . Core sanitization happens in the sanitize/For methods. But considering your suggestion, I see no harm to have it set as properly and done it .

classes/core/PKPHtmlSanitizer.php
touhidurabir: the config initialized right before the looping of `allowedTagToAttributeMap` as
```php
$this->htmlSanitizerConfig = (new HtmlSanitizerConfig())
            ->allowLinkSchemes(['https', 'http', 'mailto'])
            ->allowMediaSchemes(['https', 'http']);
```

classes/core/PKPHtmlSanitizer.php
touhidurabir: Right, it's a bit complex logic and hard to understand what it's doing , DocBlock added .

classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Instance of HtmlSanitizer
+     */
+    protected HtmlSanitizer $htmlSanitizer;
+
+    /**
+     * Create a new instance
+     * 
+     * @param string $allowable Config key such security.[allowed_html/allowed_title_html]
```
asmecher, line 50: It's confusing to have this operate in two modes. I think you could just move the `Config` interaction into `PKPString`, and make this function always accept a string of allowable elements. (Also I think `attribited` should be `attributes`.)

classes/core/PKPHtmlSanitizer.php
asmecher: Ah, I was getting confused by the habit of cloning of the config object within the Symfony toolset! I see why they did it that way, but it certainly makes this part harder to read.

classes/core/PKPHtmlSanitizer.php
```diff
@@ -0,0 +1,192 @@
...
+                    Config::getVar('security', $allowable, null) ?? $allowable
+                )
+            )
+        );
+    }
+
+    /**
+     * Set a pre defined instance of \Symfony\Component\HtmlSanitizer\HtmlSanitizerConfig
+     */
+    public function setSanitizerConfig(HtmlSanitizerConfig $htmlSanitizerConfig): self
```
asmecher, line 72: Are the getter/setter ever used? It doesn't look like it (and the recreation of `htmlSanitizer` seems like an opportunity for side-effects).

asmecher (changes requested):
@touhidurabir, sorry to be a stick in the mud about this, but this adds a lot of complexity (and more than doubles the execution time) of a core function in order to accommodate a use case that's really uncommon. Is there a downside to leaving `stripUnsafeHtml` as it is and having the plugins use libxml to fish out paragraph text per [this snippet](https://github.com/pkp/jatsTemplate/pull/30#discussion_r1297687873)?

touhidurabir:
> this adds a lot of complexity

Yes, but it takes into the consideration of `W3C` specification . see more at https://wicg.github.io/sanitizer-api/#default-configuration. 

> more than doubles the execution time

@asmecher as per my calculation, this implementation takes less time and memory comparing to our [`HTMLPurifier`](https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/core/PKPString.php#L407-L419) or [current](https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPString.php#L166-L219) implementation as 

**With current implementation**
```
Total memory usage in KB : 1633.7421875
Total execution time seconds : 0.016600847244263
```

**With `HTMLPurifier` implementation**
```
Total memory usage in KB : 1644.7421875
Total execution time seconds : 0.0094790458679199
```

**With new implementation in PR**
```
Total memory usage in KB : 1574.265625
Total execution time seconds : 0.0080928802490234
```

I used the following approach(very simple and probably bit too naive)
```php
$beforeMemoryUsage = memory_get_usage();
$beforeTime = microtime(true);

echo PKPString::stripUnsafeHtml(
    "&lt;img src=&quot;invalid-url&quot; onerror=&quot;alert(&#039;XSS Attack prefix&#039;)&quot; /&gt; <b>Bold</b> <i>Italic</i> <u>Underline</u> <sup>Superscript</sup> <sub>Subscript</sub> _______ 000001 <b>MORE BOLD</b>",
    "allowed_title_html"
) .
    PHP_EOL .
    PHP_EOL;

$afterMemoryUsage = memory_get_usage();
$afterTime = microtime(true);

echo "Total memory usage in KB : " .
    (float) (($afterMemoryUsage - $beforeMemoryUsage) / 1024) .
    PHP_EOL .
    PHP_EOL;

echo "Total execution time seconds : " .
    (float) ($afterTime - $beforeTime) .
    PHP_EOL .
    PHP_EOL;
```

> Is there a downside to leaving `stripUnsafeHtml` as it is 

I strongly believe that current implementation is not ideal and heavily depends on `strip_tags` . We do need to use the `strip_tags` as both the `HTMLPurifier` and `symfony html-sanitizer` use those internally but only to handle invalid tags just to make sure we do get the inner content of those tags but not the tags as current implementation does .

classes/core/PKPHtmlSanitizer.php
touhidurabir: done .

classes/core/PKPHtmlSanitizer.php
touhidurabir: I am thinking if from more of a generalized class that can be reuse other places so make the `HtmlSanitizerConfig` modifiable but we can remove those as those not needed anywhere right now and later introduce it if ever needed to .

PR #9263, opened 2023-08-28
pkp/pkp-lib#9195 removed int type hint for tries public property from jobs

touhidurabir:
patch for pkp/pkp-lib#9195 as of issue reported at https://forum.pkp.sfu.ca/t/ojs-3-4-0-2-php-fatal-error-usage-statistics-stopped-procesing-records/80762

### DIFF
--- a/jobs/BaseJob.php
+++ b/jobs/BaseJob.php
@@ -35,8 +35,10 @@ abstract class BaseJob implements ShouldQueue
 
     /**
      * The number of times the job may be attempted.
+     * 
+     * @var int
      */
-    public int $tries = 3;
+    public $tries = 3;
 
     /**
      * The number of SECONDS to wait before retrying the job.
--- a/jobs/testJobs/TestJobFailure.php
+++ b/jobs/testJobs/TestJobFailure.php
@@ -27,8 +27,10 @@ class TestJobFailure extends BaseJob
 
     /**
      * The number of times the job may be attempted.
+     * 
+     * @var int
      */
-    public int $tries = 1;
+    public $tries = 1;
 
     /**
      * The maximum number of unhandled exceptions to allow before failing.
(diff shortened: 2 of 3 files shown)


asmecher:
@touhidurabir, isn't the unserialize going to encounter the same problem with `$backoff` and the other typehinted properties that were added as well?

touhidurabir:
@asmecher the problem lies with the already used cases of at https://github.com/pkp/pkp-lib/blob/main/jobs/statistics/CompileMonthlyMetrics.php#L28 and https://github.com/pkp/ojs/blob/main/jobs/statistics/CompileUsageStatsFromTemporaryRecords.php#L34 (this one for `OJS/OPS/OMP`) . 

Now the problem only happens for these 2 jobs . we can remove the type hint from other properties but in that case all pending jobs in the installation of `3.4.0-2` that has dispatched after the upgrade to `3.4.0-2` will fail to execute . However if we can safely assume that only a handful have upgraded to `3.4.0-2` , we can remove the other property type hints . May be better to remove as it seems odd as one config property has type hint and other does not . But I am concerned if a lot of installations have already upgraded to `3.4.0-2`

and if we keep those , just need to make sure when override in extending class, we have same type hint . weirdly VS code does not show any issue which is in my case I did not notice it .

asmecher:
Ah, thanks, I misunderstood the original issue as a serialize/unserialize incompatibility across releases, when it's actually about internal consistency of subclasses *within* the 3.4.0-2 release. This works, thanks!

bozana:
I am wondering if we haven't changed those two cases to have `int tries = 1` i.e. to be compatible with the base class? Or is it however better to remove the type hint?

touhidurabir:
@bozana I can see that the type hint `int` for property `tries` been removed from `\PKP\jobs\BaseJob` class in the both branch `stable-3.4.0` and `main` . Is there any issue still persist ?

bozana:
No no, I am just wondering why we decided to removed that type hing (int) and did not added (int) type hint the two classes that did not have it...

PR #9267, opened 2023-08-29
fix error in PKPPublication.inc.php

murilodbva:
Fix the error described (and resolved ) in this link:

https://forum.pkp.sfu.ca/t/when-importing-an-issue-system-returns-success-but-only-partial-data-appears-on-the-journal-webpage/79200

p.s: maybe someone, after, can find a solution fixing inside database.

### DIFF
--- a/classes/publication/PKPPublication.inc.php
+++ b/classes/publication/PKPPublication.inc.php
@@ -146,7 +146,7 @@ class PKPPublication extends DataObject {
 			if (!empty($str)) {
 				if ($lastUserGroupId != $author->getData('userGroupId')) {
 					foreach ($userGroups as $userGroup) {
-						if ($lastUserGroupId === $userGroup->getId()) {
+						if ($userGroup && $lastUserGroupId === $userGroup->getId()) {
 							if ($userGroup->getData('showTitle')) {
 								$str .= ' (' . $userGroup->getLocalizedData('name') . ')';
 							}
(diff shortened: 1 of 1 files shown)


asmecher:
@murilodbva, what theme are you using?

murilodbva:
Hi @asmecher . Thanks for responsing.
I'm using default theme. How can see in the print below:

murilodbva:
@asmecher, I remmebered a important `detail:`

This bug occur **just** in the publications **before** from **2017** age.

murilodbva:
@asmecher, I remmebered a important detail:

This bug occur **just** in the publications **before** from **2017** age.

asmecher:
@murilodbva, I think this might be a data error. Try the following query:
```sql
SELECT COUNT(*) FROM authors a LEFT JOIN user_groups ug ON (a.user_group_id = ug.user_group_id) WHERE ug.user_group_id IS NULL;
```
If you get any results, you'll need to correct them. Probably what happened is that someone deleted an entry in `user_groups` that was already in use, so now records in `authors` are pointing to non-existent entries. We've added formal foreign key declarations in OJS/OMP/OPS 3.4.0, so this kind of inconsistency will be caught at the database level.

murilodbva:
@asmecher, that select returned 3.495 results from my database.

asmecher:
@murilodbva, OK, I'm pretty sure that's the data error. That's a lot of rows but the good news is that there is probably only one or two `user_group` entries missing.

This was probably an Author user group, and you've probably re-created a new one; find its `user_group_id` (e.g. from a working author entry, and set the `user_group_id` of the problem rows to the same value. (Make sure to take a good backup before you do this.)

I'll close this PR, as the best solution will be to fix the data error.

Issue #9278, opened 2023-09-06
Use ISO639-2b instead of ISO639-3 for locale conversion


bozana:
We now use `sokil/php-isocodes` library for locale conversion. The function LocaleConversion::get3LetterFrom2LetterIsoLanguage (https://github.com/pkp/pkp-lib/blob/main/classes/i18n/LocaleConversion.php#L34) should convert to ISO 639-2b, but sokil/php-isocodes only returns ISO 639-3, which is not the same.
Our metadata servies (mEDRA, DOAJ, PubMed, MARC) use/expect  ISO 639-2b.
Thus, use the same old functions, e.g. LocaleConversion::get3LetterFrom2LetterIsoLanguage, for the conversions they were also earlier used for, and as the function comments say for conversion to/from ISO 639-2b.
`sokil/php-isocodes-db-i18n` (used by `sokil/php-isocodes`) comes also with `iso_639-2.json` file/database. So maybe for those functions in LocaleConversion we could get the values from that file.

<hr>

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9279
- https://github.com/pkp/ojs/pull/4031 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/9429
- https://github.com/pkp/ojs/pull/4071 (only submodule update)

bozana:
@jonasraoni or @asmecher, could one of you review the PR above?

PR #9279, opened 2023-09-06
pkp/pkp-lib#9278 Use ISO639-2b instead of ISO639-3 for locale conversion

bozana:
s. https://github.com/pkp/pkp-lib/issues/9278

### DIFF
--- a/classes/dev/ComposerScript.php
+++ b/classes/dev/ComposerScript.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * @file classes/dev/ComposerScript.php
+ *
+ * Copyright (c) 2023 Simon Fraser University
+ * Copyright (c) 2023 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class ComposerScript
+ *
+ * @brief Custom composer script that checks if the file iso_639-2.json exists in sokil library
+ */
+
+namespace PKP\dev;
+
+use Exception;
+
+class ComposerScript
+{
+    /**
+     * A post-install-cmd custom composer script that checks if
+     * the file iso_639-2.json exists in the installed sokil library
+     *
+     * @throw Exception
+     */
+    public static function isoFileCheck(): void
+    {
+        // We use dirname(__FILE__, 3) and not Core::getBaseDir() because
+        // this funciton is called by Composer, where INDEX_FILE_LOCATION is not defined.
+        $iso6392bFile = dirname(__FILE__, 3) . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
+        if (!file_exists($iso6392bFile)) {
+            throw new Exception("The ISO639-2b file {$iso6392bFile} does not exist.");
+        }
+    }
+}
--- a/classes/i18n/LocaleConversion.php
+++ b/classes/i18n/LocaleConversion.php
@@ -23,19 +23,53 @@ declare(strict_types=1);
 
 namespace PKP\i18n;
 
+use DateInterval;
+use Exception;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Cache;
+use PKP\core\Core;
 use PKP\facades\Locale;
 
 class LocaleConversion
 {
+    /** @var string Max lifetime for the ISO639-2b cache. */
+    protected const MAX_ISO6392B_CACHE_LIFETIME = '1 year';
+
+    /**
+     * Get ISO639-2b array
+     *
+     * @throw Exception
+     */
+    protected static function getISO6392b(): array
+    {
+        $iso6392bFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
+        if (!file_exists($iso6392bFile)) {
+            throw new Exception("The ISO639-2b file {$iso6392bFile} does not exist.");
+        }
+        $key = __METHOD__ . 'iso639-2b' . self::MAX_ISO6392B_CACHE_LIFETIME . filemtime($iso6392bFile);
+        $expiration = DateInterval::createFromDateString(static::MAX_ISO6392B_CACHE_LIFETIME);
+        return Cache::remember($key, $expiration, function () use ($iso6392bFile) {
+            return json_decode(file_get_contents($iso6392bFile), true);
+        });
+    }
+
     /**
      * Translate the ISO 2-letter language string (ISO639-1) into a ISO compatible 3-letter string (ISO639-2b).
      */
     public static function get3LetterFrom2LetterIsoLanguage(?string $iso2Letter): ?string
     {
-        assert(strlen($iso2Letter) === 2);
-        $locale = Arr::first(Locale::getLocales(), fn (LocaleMetadata $locale) => $locale->getIsoAlpha2() === $iso2Letter);
-        return $locale ? $locale->getIsoAlpha3() : null;
+        try {
+            $languages = self::getISO6392b();
+        } catch (Exception $e) {
+            error_log($e->getMessage());
+            return null;
+        }
+        foreach (reset($languages) as $languageRaw) {
+            if (($languageRaw['alpha_2'] ?? null) === $iso2Letter) {
+                return $languageRaw['bibliographic'] ?? $languageRaw['alpha_3'];
+            }
+        }
+        return null;
     }
 
     /**
@@ -43,9 +77,18 @@ class LocaleConversion
      */
     public static function get2LetterFrom3LetterIsoLanguage(?string $iso3Letter): ?string
     {
-        assert(strlen($iso3Letter) === 3);
-        $locale = Arr::first(Locale::getLocales(), fn (LocaleMetadata $locale) => $locale->getIsoAlpha3() === $iso3Letter);
-        return $locale ? $locale->getIsoAlpha2() : null;
+        try {
+            $languages = self::getISO6392b();
+        } catch (Exception $e) {
+            error_log($e->getMessage());
+            return null;
+        }
+        foreach (reset($languages) as $languageRaw) {
+            if (($languageRaw['bibliographic'] ?? null) === $iso3Letter || $languageRaw['alpha_3'] === $iso3Letter) {
+                return $languageRaw['alpha_2'] ?? null;
+            }
+        }
+        return null;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
+    /**
+     * Get ISO639-2b array
+     */
+    protected static function getISO6392b(): array
+    {
+        $iso6392BFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
+        if (!file_exists($iso6392BFile)) {
+            return [];
+        }
+        return json_decode(file_get_contents($iso6392BFile), true);
```
jonasraoni, line 41: Given the file is static, it might be worth to cache this.

If you use the Laravel Cache, the underlying implementation will attempt to make use of the OPcache (if available on the server). If the data is small, it should be ok to cache it locally (`static $cache`).

ps: I'm considering this method can be called several times.

classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
+    }
+
     /**
      * Translate the ISO 2-letter language string (ISO639-1) into a ISO compatible 3-letter string (ISO639-2b).
      */
     public static function get3LetterFrom2LetterIsoLanguage(?string $iso2Letter): ?string
     {
         assert(strlen($iso2Letter) === 2);
```
jonasraoni, line 49: Unrelated to this task, but I think these asserts can be removed now.
If they are really important, they should be turned into an `InvalidArgumentException`

classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
-        $locale = Arr::first(Locale::getLocales(), fn (LocaleMetadata $locale) => $locale->getIsoAlpha2() === $iso2Letter);
-        return $locale ? $locale->getIsoAlpha3() : null;
+        $iso6392b = null;
+        $languages = self::getISO6392b();
+        foreach (reset($languages) as $languageRaw) {
+            if (array_key_exists('alpha_2', $languageRaw) && $languageRaw['alpha_2'] === $iso2Letter) {
```
jonasraoni, line 53: Not sure if it's ok for your use-case, but I'd use: `($languageRaw['alpha_2'] ?? null) === $iso2Letter` and below `$iso6392b = $languageRaw['bibliographic'] ?? $languageRaw['alpha_3']`

classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
+                if (array_key_exists('bibliographic', $languageRaw)) {
+                    $iso6392b = $languageRaw['bibliographic'];
+                } else {
+                    $iso6392b = $languageRaw['alpha_3'];
+                }
+            }
```
jonasraoni, line 59: I think there should be a `break` here, there's also a functional way through Laravel (`Arr::first()`) to quit the loop once a truthy value is found.

classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
      * Translate the ISO 2-letter language string (ISO639-1) into a ISO compatible 3-letter string (ISO639-2b).
      */
     public static function get3LetterFrom2LetterIsoLanguage(?string $iso2Letter): ?string
     {
         assert(strlen($iso2Letter) === 2);
-        $locale = Arr::first(Locale::getLocales(), fn (LocaleMetadata $locale) => $locale->getIsoAlpha2() === $iso2Letter);
-        return $locale ? $locale->getIsoAlpha3() : null;
+        $iso6392b = null;
+        $languages = self::getISO6392b();
+        foreach (reset($languages) as $languageRaw) {
```
jonasraoni, line 52: I'm not sure if it's a good idea, but there's room for adding cache here and on the other methods.

jonasraoni (approved):
@bozana I've approved, I just think it's important to add the breaks and to add some extra safety to the method `getISO6392b()`.

bozana:
Thanks a lot @jonasraoni! All good comments!

classes/i18n/LocaleConversion.php
bozana: I will leave it without caching...

classes/i18n/LocaleConversion.php
```diff
@@ -24,18 +24,41 @@
...
 use PKP\facades\Locale;
 
 class LocaleConversion
 {
+    /**
+     * Get ISO639-2b array
+     */
+    protected static function getISO6392b(): array
+    {
+        $iso6392BFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
```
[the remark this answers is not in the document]
bozana: I copied the file localy, to lib/pkp/lib/iso_639-2.json, and am falling back to it, as you suggested.

classes/i18n/LocaleConversion.php
bozana: OK, I am using Laravel Cache now...

classes/i18n/LocaleConversion.php
bozana: Removed...

bozana:
@jonasraoni, followed your comments. Would you like to see the changes (the last commit) now (above all how I now use the copy of the iso_639-2b file and Laravel Cache?
Thanks a lot!

bozana:
Maybe also to double check with @asmecher, that we can provide and 'wear' the iso_639-2b.json file this way? Can/shell we note somewhere where from do we have the file? (https://github.com/sokil/php-isocodes-db-only/blob/4.0/databases/iso_639-2.json)

asmecher:
> Maybe also to double check with @asmecher, that we can provide and 'wear' the iso_639-2b.json file this way?

I actually just commented on this :)

https://github.com/pkp/pkp-lib/pull/9279#discussion_r1318905022

classes/i18n/LocaleConversion.php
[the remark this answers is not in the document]
jonasraoni: Yeah, I was thinking about a kind of "build step", I just think the file should be committed to avoid surprises, like you depend on the "package^1.0", then composer might pick the 1.1 depending on how the person installed (which could break things).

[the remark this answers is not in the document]
bozana: Hmmm... I am not sure if this could help/if it is the right solution... :thinking: 
Actually, we only need the copy of that file if sokil at some point stops providing it -- in that case when we/users get their code with composer we will not have it. 
As far as I understand those solutions with composer: they all get the remote files first and then do some things. Is this so? If so, it is not so useful for us, because if it gets the sokil library that has the file, no need to copy it. And if it gets the sokil library that does not have the file, we cannot copy it anyway.
Thus, actually, I think, we would need a script, that would download only that file if there is no such file in the downloaded sokil library. Could we then implement it so in that PHP function?
Also, this file is actually the file from debian package. So maybe there is another way for us to keep and maintain it locally?
:thinking:

classes/i18n/LocaleConversion.php
```diff
@@ -34,51 +35,43 @@ class LocaleConversion
...
-        $iso6392BFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
-        if (!file_exists($iso6392BFile)) {
-            return [];
+        $key = __METHOD__ . 'iso639-2b';
+        $iso6392bFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
+        if (!file_exists($iso6392bFile)) {
+            $iso6392bFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/iso_639-2.json';
         }
-        return json_decode(file_get_contents($iso6392BFile), true);
+        return Cache::rememberForever($key, function () use ($iso6392bFile) {
```
jonasraoni, line 43: I think the key should consider the modified date of the file.

jonasraoni (approved):

classes/i18n/LocaleConversion.php
bozana: If I use the date modified from the iso file as the key, and I store it forever (rememberForever), will the old cache files keep to exist?
Maybe then to use some expiration date, e.g. 1 year?

bozana:
@jonasraoni and @asmecher, could you please see the changes here again?
Now I am using a post-install-cmd composer script to check if the iso639-2b.json file from the sokil library exists. The function throws an exception if the file does not exist.
I do not copy that ISO file locally any more.
@jonasraoni, could you please also take a look at the remaining unresolved comments: how the key and cache is used?
Thanks a lot!

asmecher:
@bozana, sure, that's OK with me.

classes/i18n/LocaleConversion.php
```diff
@@ -23,29 +23,76 @@
...
+use Exception;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Cache;
+use PKP\core\Core;
 use PKP\facades\Locale;
 
 class LocaleConversion
 {
+    /** @var string Max lifetime for the ISO639-2b cache. */
+    protected const MAX_ISO6392b_CACHE_LIFETIME = '1 year';
```
jonasraoni, line 35: I think the `b` should be uppercased.

classes/i18n/LocaleConversion.php
```diff
@@ -23,29 +23,76 @@
...
+
+    /**
+     * Get ISO639-2b array
+     */
+    protected static function getISO6392b(): array
+    {
+        $iso6392bFile = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/vendor/sokil/php-isocodes-db-i18n/databases/iso_639-2.json';
+        if (!file_exists($iso6392bFile)) {
+            error_log("The ISO639-2b file {$iso6392bFile} does not exist.");
+            return [];
```
jonasraoni, line 61: Is this feature essential or cause strange issues if non-functional? If yes, then an exception sounds better.

classes/i18n/LocaleConversion.php
```diff
@@ -23,29 +23,76 @@
...
+        }
+        $key = __METHOD__ . 'iso639-2b' . self::MAX_ISO6392b_CACHE_LIFETIME . filemtime($iso6392bFile);
+        $languages = json_decode(file_get_contents($iso6392bFile), true);
+        if (is_array($languages) && !empty($languages)) {
```
jonasraoni, line 65: Here I also think it's better to fire an exception, the file is supposed to never be empty and always be parseable 🤔

classes/i18n/LocaleConversion.php
```diff
@@ -64,30 +111,36 @@ public static function get3LetterIsoFromLocale(?string $locale): ?string
...
     {
-        assert(strlen($iso3Letter) === 3);
         $primaryLocale = Locale::getPrimaryLocale();
 
-        $candidates = [];
+        $alpha2Candidates = $localeCandidates = [];
+        $languages = self::getISO6392b();
+        foreach (reset($languages) as $languageRaw) {
+            if (($languageRaw['bibliographic'] ?? null) === $iso3Letter || $languageRaw['alpha_3'] === $iso3Letter) {
+                if (array_key_exists('alpha_2', $languageRaw)) {
```
jonasraoni, line 120: I'm reviewing from my mobile, so it's complex to analyze the json 😁
Doesn't it always provide the `alpha_2`? If not, ignore my comment.

classes/i18n/LocaleConversion.php
```diff
@@ -23,29 +23,76 @@
...
+    /** @var string Max lifetime for the ISO639-2b cache. */
+    protected const MAX_ISO6392b_CACHE_LIFETIME = '1 year';
+
+    /**
+     * A post-install-cmd custom composer script that checks if
+     * the iso_639-2b.json file exists in the installed sokil library
+     *
+     * @throw Exception
+     */
+    public static function composerIsoFileCheckScript(): void
```
jonasraoni, line 43: As this isn't useful for the codebase, I think we should move it to a separated "build script", like "composer-commands.php".

jonasraoni (approved):
@bozana I've approved, just left some observations, check if it makes sense to update or not 🙂

classes/i18n/LocaleConversion.php
jonasraoni: Yep, it's better to add an expiration date.

p.s. At this moment, we don't have a job to clear the expired cache.

classes/i18n/LocaleConversion.php
bozana: Yes, I was thinking about having a class ComposerScript or so, but I did not know where to put it -- I thought in lib/pkp/classes/composer/ -- and then I thought it currently only for this one small function, not even using any composer script functionality... Thus, I left it here for now...
Hmmm... I myself am not sure...
@jonasraoni or @asmecher, where would you put it?

bozana: I added the class ComposerScript in lib/pkp/dev/ that contains this function...

bozana:
@jonasraoni, sorry, could you please please take one more look -- I hope now I have everything... :-)

classes/dev/ComposerScript.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
```
jonasraoni, line 3: We've decided to not add `strict_types` anymore (maybe the existing ones will be removed later).
I don't think you need to change anything here (i.e. it's just a script), so I'm just sharing the information.

jonasraoni (approved):

classes/dev/ComposerScript.php
bozana: Thanks a lot @jonasraoni! I removed it and will merge once the tests pass...

PR #9280, opened 2023-09-06
I8919 vue3&vite migration

jardakotesovec:


templates/admin/failedJobs.tpl
```diff
@@ -32,25 +32,27 @@
...
-							{translate key="admin.jobs.failed.action.redispatch"}
-						</pkp-button>
-						<pkp-button is-warnable @click="remove(row)">
-							{translate key="common.delete"}
-						</pkp-button>
-						<pkp-button element="a" is-link :href="row._hrefs._details">
-							{translate key="common.details"}
-						</pkp-button>
-					</button-row>
+					<template #default v-if="column.name === 'actions'">
```
jardakotesovec, line 43: This is nuanced, but important. If its not wrapped in template - table-cell is having hard time whether something is passed to the slot. Because even if the v-if is false in previous implementation it still receives some abstract v-if element to the slot.

templates/admin/failedJobDetails.tpl
```diff
@@ -19,17 +19,19 @@
...
 					:key="column.name"
 					:column="column"
 					:row="row"
 					:tabindex="!rowIndex && !columnIndex ? 0 : -1"
 				>
-					<span v-if="column.name === 'value' && isValidJson(row.value)">
-						<pre>{{ row.value }}</pre>
-					</span>
+					<template #default v-if="column.name === 'value' && isValidJson(row.value)">
+						<span v-if="column.name === 'value' && isValidJson(row.value)">
```
jonasraoni, line 31: The `v-if` seems to be duplicated.

js/classes/VueRegistry.js
```diff
@@ -53,7 +56,7 @@ export default {
 		$parents.each(function (i) {
 			if ($.pkp.classes.Handler.hasHandler($($parents[i]))) {
 				$.pkp.classes.Handler.getHandler($($parents[i])).handlerChildren_.push(
-					pkp.registry._instances[id]
+					pkp.registry._instances[id],
```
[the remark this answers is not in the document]
jardakotesovec: @jonasraoni Its all going through build step and replacing some new patterns with older based on our target browsers, so we don't have to worry about it.

PR #9293, opened 2023-09-10
Fix null locale when inserting into "email_templates_settings"

alvgalrus:
Hello!

I had a problem when upgrading from 3.3.0-14 into 3.4.0-3:

```
PHP Warning:  Attempt to read property "primary_locale" on null in /lib/pkp/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php on line 461
SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'locale' cannot be null (SQL: insert into `email_templates_settings` (`email_id`, `locale`, `setting_name`, `setting_value`) values (...)
```

As we only have one journal, `context_id` was equal to `1` in our installation. Removing the strict comparison fixed the migration flow.

Feel free to suggest any changes to this PR.

Regards!

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php
+++ b/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php
@@ -457,7 +457,7 @@ abstract class I5716_EmailTemplateAssignments extends Migration
 
         $contextIds->each(function (int $contextId) use ($primaryLocales) {
             $primaryLocale = $primaryLocales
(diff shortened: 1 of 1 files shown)


alvgalrus:
This may fix the following unresolved thread in the forum: https://forum.pkp.sfu.ca/t/error-during-upgrade-from-3-3-0-14-to-3-4-0-1/79280

asmecher:
@Vitaliy-1, could you take a look at this one? Thanks!

Issue #9275, opened 2023-09-05
Unable to create a new publication through submission API


Vitaliy-1:
**Describe the bug**
Reported on the forum: https://forum.pkp.sfu.ca/t/is-there-any-documentation-on-how-to-fully-create-a-submission-via-rest-api/74730/8

**To Reproduce**
1. Enable API key for the user with enough permissions level, e.g., site admin
2. Make a post request to create a new submission, e.g.: `http://localhost:8000/index.php/publicknowledge/api/v1/submissions`
3. See the `submissionId` in the response
5. Make a post request to create a new publication, using the newly created submission ID, e.g., if it's `20`: `http://localhost:8000/index.php/publicknowledge/api/v1/submissions/20/publications`
6. See the response: `<h1>404 Not Found</h1>`

**What application are you using?**
 OJS, OMP and OPS stable 3.3.0 

**Additional information**
`PKPSubmissionHandler::addPublication()` requires publication write access policy but the publication doesn't exist yet, see: https://github.com/pkp/pkp-lib/blob/stable-3_3_0/api/v1/submissions/PKPSubmissionHandler.inc.php#L41

asmecher:
@Vitaliy-1, if it's OK with you, I've assigned this to @defstat for a look.

defstat:
@Vitaliy-1 @asmecher 

I have added a PR in the first comment of this issue. As @Vitaliy-1 suggested, removing the `addPublication` from the `$requiresPublicationWriteAccess` list resolves the issue. 

`stable-3_4_0` and `main` branches have this fix in place already

Vitaliy-1:
Thanks, @defstat. I've merged this.

PR #9314, opened 2023-09-20
[PKP-LIB][stable-3_3_0] #9275 Remove addPublication from requiresPublicationWriteAccess list of functions

defstat:


Vitaliy-1 (approved):
Thanks, looks good!

Issue #9301, opened 2023-09-13
Remove user email address confirmation from password reset message


mfelczak:
**Describe the bug**
The current password reset message confirms the existence of a user account with the provided email address:
"A confirmation has been sent to your email address. Please follow the instructions in the email to reset your password."

If a site visitor is fishing for valid emails and no user exists with the provided email address, the following is displayed:
"No user exists with the specified email address."

The above approach provides information about registered users and their email addresses to unauthenticated site visitors. This doesn't meet [OWASP best practices](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#forgot-password-service) and likely violates GDPR.

User feedback should be updated to a generic message for both valid and invalid email addresses, e.g. "If a user exists with this email address a password reset email has been sent to that email address."

**To Reproduce**
1. Go to 'Forgot my password'
2. Enter known user account email address.
3. Enter invalid email address.

**What application are you using?**
 All versions of OJS, OMP and OPS.

**PRs**
#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9317
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4046 [TEST ONLY]

#### Main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9519
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4103 [TEST ONLY]

#### Stable-3.3.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9520
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4104

asmecher:
(@touhidurabir, this might make a good small task to work in with the larger stuff you're working on)

touhidurabir:
@asmecher ok, I will take a look at this .

touhidurabir:
@asmecher can you review the PR at https://github.com/pkp/pkp-lib/issues/9301#issue-1894620042 . just a very minor change and once finalise, I can forward port to `main` and back port to `stable-3.3.0` .

asmecher:
@touhidurabir, there's a merge conflict there; can you rebase the PR to resolve?

touhidurabir:
[...]

Issue #9169, opened 2023-07-15
Password reset process does not provide username to user


asmecher:
Starting with OJS 3.3.0-14, the password reset process was simplified:

https://github.com/pkp/pkp-lib/issues/2135

However, the new password reset email does not include the username in the text (or available variables). This causes confusion when users know their password but are using the wrong username.

Re-add the username to the available list of variables, and consider adding it to the default text.

asmecher:
Changes:
- For 3.3.0-x users, a new variable `{$recipientUsername}` can be added to the `PASSWORD_RESET_CONFIRM` email template. (This is already available in 3.4.0.)
- At the conclusion of the password reset process, clicking "Login" will pre-fill the login form with the appropriate username.

Users of OJS 3.3.0-x older than 3.3.0-16 can apply this patch:
```diff
diff --git a/pages/login/LoginHandler.inc.php b/pages/login/LoginHandler.inc.php
index ef61d9d82c..761caca888 100644
--- a/pages/login/LoginHandler.inc.php
+++ b/pages/login/LoginHandler.inc.php
@@ -51,7 +51,7 @@ class LoginHandler extends Handler {
 		$templateMgr = TemplateManager::getManager($request);
 		$templateMgr->assign(array(
 			'loginMessage' => $request->getUserVar('loginMessage'),
-			'username' => $session->getSessionVar('username'),
+			'username' => $request->getUserVar('username') ?? $session->getSessionVar('username'),
 			'remember' => $request->getUserVar('remember'),
 			'source' => $request->getUserVar('source'),
 			'showRemember' => Config::getVar('general', 'session_lifetime') > 0,
@@ -202,6 +202,7 @@ class LoginHandler extends Handler {
 		$mail->assignParams([
 			'url' => $request->url(null, 'login', 'resetPassword', $user->getUsername(), array('confirm' => $hash)),
 			'siteTitle' => htmlspecialchars($site->getLocalizedTitle()),
+			'recipientUsername' => $user->getUsername(),
 		]);
 		$mail->addRecipient($user->getEmail(), $user->getFullName());
 		$mail->send();
@@ -209,7 +210,7 @@ class LoginHandler extends Handler {
 		$templateMgr->assign([
 			'pageTitle' => 'user.login.resetPassword',
 			'message' => 'user.login.lostPassword.confirmationSent',
-			'backLink' => $request->url(null, $request->getRequestedPage()),
[...]

PR #9317, opened 2023-09-22
pkp/pkp-lib#9301 removed user email address confirmation from password reset message

touhidurabir:
for pkp/pkp-lib#9301

### DIFF
--- a/pages/login/LoginHandler.php
+++ b/pages/login/LoginHandler.php
@@ -223,43 +223,36 @@ class LoginHandler extends Handler
         $email = $request->getUserVar('email');
         $user = Repo::user()->getByEmail($email, true); /** @var User $user */
 
-        if ($user === null) {
-            $templateMgr->assign([
-                'pageTitle' => 'user.login.resetPassword',
-                'message' => 'user.login.lostPassword.confirmationSent',
-                'backLink' => $request->url(null, $request->getRequestedPage(), null, null),
-                'backLinkLabel' => 'user.login',
-            ])->display('frontend/pages/message.tpl');
+        if ($user !== null) {
+            
+            if ($user->getDisabled()) {
+                $templateMgr
+                    ->assign([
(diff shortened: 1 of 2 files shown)


pages/login/LoginHandler.php
```diff
@@ -223,45 +223,41 @@ public function requestResetPassword($args, $request)
...
-            ->from($site->getLocalizedContactEmail(), $site->getLocalizedContactName())
-            ->body($template->getLocalizedData('body'))
-            ->subject($template->getLocalizedData('subject'));
-        Mail::send($mailable);
-
         $templateMgr->assign([
             'pageTitle' => 'user.login.resetPassword',
             'message' => 'user.login.lostPassword.confirmationSent',
-            'backLink' => $request->url(null, $request->getRequestedPage(), null, null, ['username' => $user->getUsername()]),
+            'backLink' => $request->url(null, $request->getRequestedPage(), null, null),
```
asmecher, line 260: This reverts https://github.com/pkp/pkp-lib/issues/9169. Since we're looking up the user by email address, and planning to add login by email, what do you think about moving ahead with merging https://github.com/pkp/pkp-lib/issues/8871 and using the email address in place of username here?

asmecher (changes requested):

pages/login/LoginHandler.php
asmecher: Sorry for the delay in getting to this! Thinking this over from a higher level, I think allowing users to log in via email will solve the underlying problem, which is that usernames are easily forgotten and getting less common. I would be OK with reverting #9169 in the process of merging email-based logins.

touhidurabir: @asmecher PRs are done But note that reversion of https://github.com/pkp/pkp-lib/issues/9169 has not  implemented in `stable-3.3.0` as `stable-3.3.0` does not have login via email functionality . We only have login via email for `stable-3.4.0` and `main` .

Issue #9310, opened 2023-09-18
OJS 3.4 Editorial Activity: "Accept and Skip Review" does not mark article as "Accepted"


ronste:
**Describe the bug**
When accepting a submission without review the submission is not counted as "Accepted" in the editorial activity statistics. It is only counted as "Accepted" in case the review process was initiated.

**To Reproduce**
1. Go to "Editorial activity" and see the number of "Submissions Accepted"
2. Create a new submission
3. Go to the editorial workflow and click "Accept and Skip Review"
4. Go to "Editorial activity" and see that the number of "Submissions Accepted" didn't increase (as is the behaviour in OJS 3.3)

**What application are you using?**
 OJS 3.4.0.3

<hr>

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9318
- https://github.com/pkp/ojs/pull/4047 (only submodule update)
- https://github.com/pkp/omp/pull/1462 (only submodule update)
- https://github.com/pkp/ops/pull/573 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/9378
- https://github.com/pkp/ojs/pull/4055 (only submodule update)
- https://github.com/pkp/omp/pull/1468 (only submodule update)
- https://github.com/pkp/ops/pull/576 (only submodule update)

bozana:
@jonasraoni, do you maybe remember if this needs to be so, and why have we decided to be so, or if "Accept and Skip Review" should be counted as well?

jonasraoni:
@bozana I've initially written a SQL query to retrieve the statistics (performance purposes), then Nate decided to break the code into an object-oriented approach. I'm not sure when this was introduced, but it's clearly a bug and it will probably need to be addressed at 3.3 as well.

ronste:
> It seems 3.3 is OK, correct @ronste? I see the code is different -- those decision constants have changed in 3.4....

Yes, seems to be a 3.4 bug only.

bozana:
@jonasraoni, could you take a look at the PR above?

PR #9318, opened 2023-09-22
pkp/pkp-lib#9310 consider all accept decisions for editorial stats

bozana:
s. https://github.com/pkp/pkp-lib/issues/9310

### DIFF
--- a/classes/services/PKPStatsEditorialService.php
+++ b/classes/services/PKPStatsEditorialService.php
@@ -32,7 +32,7 @@ class PKPStatsEditorialService
     public function getOverview($args = [])
     {
         $received = $this->countSubmissionsReceived($args);
-        $accepted = $this->countByDecisions(Decision::ACCEPT, $args);
+        $accepted = $this->countByDecisions($this->getAcceptedDecisions(), $args);
         $submissionsPublished = $this->countSubmissionsPublished($args);
         $submissionsInProgress = $this->countSubmissionsInProgress($args);
         $submissionsImported = $this->countSubmissionsImported($args);


jonasraoni (changes requested):
I think the update makes sense. But if you drill down inside the `countByDecisions()`, you'll reach a place where there's a left join with the `edit_decisions` table, that might fail (when there's a `dateStart`/`dateEnd`):
https://github.com/pkp/pkp-lib/blob/ddb2d5f0f64726dff1ecba65c0b30210bce4359c/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php#L146-L154

p.s.: I didn't investigate further/tested.

bozana:
@jonasraoni, why do you think that the left join could fail? :thinking:

jonasraoni:
@bozana because the `where()` filter that I've highlighted will require at least one row to exist on the `edit_decisions` table, and that's not going to happen if the submission was approved without an editorial decision.
But as I said, I didn't test anything, I've just skimmed the code 🙂

bozana:
Hmm... @jonasraoni, I think everything is OK. The query to get the accepted counts is:
```
select COUNT(DISTINCT s.submission_id) as count from `submissions` as `s` 
left join `publications` as `pi` on `pi`.`publication_id` = (select `pi2`.`publication_id` from
 `publications` as `pi2` where `pi2`.`submission_id` = `s`.`submission_id` and `pi2`.`status` = 3 order by `pi2`.`date_published` asc limit 1) 
left join `edit_decisions` as `ed` on `s`.`submission_id` = `ed`.`submission_id` 
 where `s`.`context_id` in (1) and `s`.`submission_progress` = 0 and (`pi`.`date_published` is null or CAST(s.date_
submitted AS DATE) <= pi.date_published) and `ed`.`decision` in (2, 17, 7) and `ed`.`date_decided` >= 2023-07-03 and `ed`.`date_decided` < 2023-10-02 and `s`.`status` != 4
```
Thus, if there is no editorial decision for a submission, it will be excluded by `left join edit_decisions as ed on s.submission_id = ed.submission_id`. The submissions that have an editorial decision will be further filtered by `where` clause, e.g. by the accept decision (Decision::ACCEPT, Decision::SKIP_EXTERNAL_REVIEW,       Decision::SEND_TO_PRODUCTION) and the decision date.
The accepted submissions should always have an entry in the table `edit_decisions`, having either decision = Decision::ACCEPT or Decision::SKIP_EXTERNAL_REVIEW or / and sometimes also Decision::SEND_TO_PRODUCTION. And they should always have `date_decided`.

jonasraoni (approved):
@bozana, I've approved! I've mixed the lack of a review with the lack of a decision.

PR #9324, opened 2023-09-25
pkp#9323 Add use Memcached to MemcacheCache.php

henriquejsfj:
For me this solved the issue. 

I'm new to PHP, so I don't know if this is the right way to do this.


asmecher:
Thanks, @henriquejsfj, that's right! I'll also take care of cherry-picking this to `stable-3_4_0`. But I'll add a heads-up about Memcache support to the related issue.

PR #9332, opened 2023-09-27
pkp/pkp-lib#8886 Add Laravel API Resources for ReviewAssignments and the submission API endpoint for retrieving review assignment by id

defstat:


api/v1/submissions/PKPSubmissionHandler.php
```diff
@@ -1673,4 +1681,34 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+     *
+     * @param SlimRequest $slimRequest Slim request object
+     * @param APIResponse $response object
+     * @param array $args arguments
+     *
+     * @return APIResponse
+     */
+    public function getReviewAssignment($slimRequest, $response, $args)
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
```
Vitaliy-1, line 1696: I don't think submission will be set as an authorised context object. It's the scope of the submissions list (dashboard view). Can you double check if we'll need another way to determine if the subeditor has access to it?

api/v1/submissions/PKPSubmissionHandler.php
```diff
@@ -170,6 +173,11 @@ public function __construct()
                     'handler' => [$this, 'getContributor'],
                     'roles' => [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_ASSISTANT, Role::ROLE_ID_REVIEWER, Role::ROLE_ID_AUTHOR],
                 ],
+                [
+                    'pattern' => $this->getEndpointPattern() . '/{submissionId:\d+}/reviewAssignments/{reviewAssignmentId:\d+}',
+                    'handler' => [$this, 'getReviewAssignment'],
+                    'roles' => [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR],
```
Vitaliy-1, line 179: What about assistants? Do they have access to the review assignments if they are assigned to the review stage?

Vitaliy-1 (commented):
Thanks! I left a couple of comments. And I'm not aware if this pattern is inline with the current work on APIs refactoring. @touhidurabir or Alec should know more

Issue #9364, opened 2023-10-02
Translation performance improvement


jardakotesovec:
## Problem
Using xdebug I noticed that lots of time was spent on calculating following key, which literally was calculated for every translation call `__.(...))`.
```
$key = __METHOD__ . static::MAX_CACHE_LIFETIME . array_reduce(array_keys($this->paths), fn (string $hash, string $path): string => sha1($hash . $path . filemtime($path)), '');
```

## Solution
Did small refactor to make key calculation faster and happen only if needed. (PR will be attached shortly)

## Potential improvement in future
There is still one performance opportunity. That new LocaleBundle gets created multiple times during the request (in my local setup 3 times). And reason is that `__.()` is called before all locale paths from plugins are registered. Example is `init` function of default theme:
```
    public function init()
    {
        // Register theme options
        $this->addOption('typography', 'FieldOptions', [
            'type' => 'radio',
            'label' => __('plugins.themes.default.option.typography.label'),

```
Every time new path is registered the LocaleBundle gets removed and needs to be loaded from cache again.

To solve this problem I think it would require to basically decouple locale path registering from plugin registration process. So at first all locale paths would register before running actual plugin registration.

Given that `addLocaleData` is called on different places depending on plugin category - it would be likely breaking change for plugins.

jardakotesovec:
pkp-lib PR: https://github.com/pkp/pkp-lib/pull/9365

And performance different (unit is % of overall time):
article landing page:  
  before: 8.42%
  after: 2.22%
submissing listing:
  before: 11.74%
  after: 3.96%

Submission After
<img width="1868" alt="perf_submission_listing_after" src="https://github.com/pkp/pkp-lib/assets/1590681/2c523def-4b62-4023-92a7-d68928eb927a">
Submission Before
<img width="1858" alt="perf_submission_listing_before" src="https://github.com/pkp/pkp-lib/assets/1590681/53a980c2-e213-466d-ab22-aea4788cfcd5">
Article After
<img width="1871" alt="perf_article_after" src="https://github.com/pkp/pkp-lib/assets/1590681/69ca27e3-9e12-469f-a21a-c04e942d09e6">
Article Before
<img width="1716" alt="perf_article_before" src="https://github.com/pkp/pkp-lib/assets/1590681/e8067139-1516-4aa2-9ce4-6608a52c90be">

jonasraoni:
[...]

PR #9365, opened 2023-10-02
pkp/pkp-lib#9364 performance improvement, by reducing key calculation…

jardakotesovec:
… and making key calculation more efficient


classes/i18n/translation/LocaleBundle.php
```diff
@@ -132,7 +135,12 @@ public function getTranslator(): Translator
             }
             return $translator;
         };
-        $key = __METHOD__ . static::MAX_CACHE_LIFETIME . array_reduce(array_keys($this->paths), fn (string $hash, string $path): string => sha1($hash . $path . filemtime($path)), '');
+
+        $pathKeys = [];
+        foreach (array_keys($this->paths) as $path) {
+            $pathKeys[] = $path . filemtime($path);
+        }
+        $key = __METHOD__ . static::MAX_CACHE_LIFETIME . md5(implode(',', $pathKeys));
```
jonasraoni, line 143: About calculating the hash only once: I prefer to avoid building long strings on memory, so I'd rather leave the way it was (the [update above](https://github.com/pkp/pkp-lib/pull/9365/files#diff-406bc041be541dae2d1a2c9e33f29ebb189cb0d519b4aceed41392dec05aa193R122-R124) should be enough to address the `$key` recreation).

I think crc32 could replace the md5/sha1 on places where collisions are unlikely to happen, but there are worse things to address :)

jonasraoni (approved):
@jardakotesovec It makes sense to avoid recreating the `$key`, so I just left one comment.

classes/i18n/translation/LocaleBundle.php
jardakotesovec: Good point, its bit of memory/performance trade off. Calling sha1 on every path seemed as significant performance hit compared to somewhat long string (couple of kBs), but in fact its not really as bad as I thought.
Difference seemed significant on original measurements, because it was called so many times.

I tried quickly compare implementation using md5&foreach vs reduce&sha1 and my version is bit faster and it uses more memory, but now we are getting to overall very low numbers territory so it kind of does not make much difference, both implementations are "good enough". Here is comparison just for interest..  and I am pretty sure xdebug itself is somewhat skewing the results as well so its good to take it with grain of salt. 
Screenshot 2023-10-02 at 17 31 27
Screenshot 2023-10-02 at 17 30 49

jonasraoni: We don't need to go into this level of detail, while there are nested queries being fired against the database. So feel free to revert or keep, I just personally prefer the other as the number of paths is unknown (but unlikely to grow too much) :)

PR #9379, opened 2023-10-04
pkp/pkp-lib#9371 Exposing locale keys used in vue.js via endpoint

jardakotesovec:
pkp/pkp-lib#9371 CR Improvements


classes/i18n/interfaces/LocaleInterface.php
```diff
@@ -169,4 +170,11 @@ public function getScripts(?string $locale = null): Scripts;
      * @return array The list of locales with formatted display name
      */
     public function getFormattedDisplayNames(array $filterByLocales = null, array $locales = null, int $langLocaleStatus = LocaleMetadata::LANGUAGE_LOCALE_WITH, bool $omitLocaleCodeInDisplay = true): array;
+
+    /**
+     *  Get UI Translator, which provides all translations used in backend UI (vue.js)
+     *
+     * @return UITranslator provides UITranslator
+    */
+    public function getUiTranslator(): UITranslator;
```
jonasraoni, line 179: When adding methods to this interface, there's another place that needs to be updated (just a workaround to provide hints to the code editor), see:
https://github.com/pkp/pkp-lib/blob/777b10995699b8419a61f62adbceb5cfcba8522d/classes/facades/Locale.php#L44

classes/i18n/translation/LocaleBundle.php
```diff
@@ -114,6 +116,17 @@ public function setEntries(array $paths): void
         $this->translator = null;
     }
 
+    /**
+     * Returns last cache key, which is useful for LocaleUITranslations cache key to ensure invalidating cache when po files changes
+     *
+     */
+    public function getLastCacheKey(): string
+    {
+        return $this->lastCacheKey ?? $this->_getCacheKey();
```
jonasraoni, line 125: I would move the `_getCacheKey()` implementation to here and remove the private method.

jonasraoni (approved):

api/v1/_i18n/I18nHandler.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Provides javascript file which includes all translations used in Vue.js UI.
+     */
+    public function getTranslations(SlimRequest $slimRequest, APIResponse $response, array $args): Response
+    {
+
+        $translations = Locale::getUiTranslator()->getTranslationStrings();
+
+        $jsContent = 'pkp.localeKeys = ' . json_encode($translations) . ';';
```
jonasraoni, line 57: Ah, I've left a comment/explanation about the update that I requested for the `json_encode()` on the other PR.

classes/i18n/translation/LocaleBundle.php
jardakotesovec: I think its better to keep them separate as for getTranslator its important to not get lastCacheKey, but always fresh one. 
While getLastCacheKey has different purpose to provide last cache key to outside world.

Issue #9382, opened 2023-10-04
Display users’ roles in OJS user list 


pilasou:
As an editor, I want to see the roles given to users in my OJS journal user list. 

Improvements:  

- Add a column in user list to displayed roles given to each user. 

![image](https://github.com/pkp/pkp-lib/assets/19291880/4ab2467b-3f6f-4280-b4c0-a45a70cefb37)

jonasraoni:
@asmecher as a user, I think this is useful and I've already reviewed the linked PR, anything against?

asmecher:
Merged and cherry-picked!

pilasou:
Hi @jonasraoni, sorry I am late in sending you our big thanks for reviewing and suggesting improvment to our code! This is our first contribution making its way to OJS code! We have a few more coming soon!

jonasraoni:
@pilasou great, help is always welcome 🤗

PR #9385, opened 2023-10-05
pkp/pkp-lib#9382 Display a role column in the user list

nibou230:
### How to test? 

- Go to Settings, User and Roles
- Roles are displayed in the user list for each user

### DIFF
--- a/controllers/grid/settings/user/UserGridHandler.php
+++ b/controllers/grid/settings/user/UserGridHandler.php
@@ -18,8 +18,10 @@ namespace PKP\controllers\grid\settings\user;
 
 use APP\facades\Repo;
 use APP\notification\NotificationManager;
+use PKP\controllers\grid\ColumnBasedGridCellProvider;
 use PKP\controllers\grid\DataObjectGridCellProvider;
 use PKP\controllers\grid\feature\PagingFeature;
+use PKP\controllers\grid\GridCellProvider;
 use PKP\controllers\grid\GridColumn;
 use PKP\controllers\grid\GridHandler;
 use PKP\controllers\grid\settings\user\form\UserDetailsForm;
@@ -38,6 +40,8 @@ use PKP\security\authorization\ContextAccessPolicy;
 use PKP\security\Role;
 use PKP\security\RoleDAO;
 use PKP\security\Validation;
+use PKP\user\User;
+use PKP\userGroup\UserGroup;
 
 class UserGridHandler extends GridHandler
 {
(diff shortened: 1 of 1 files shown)


controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -203,7 +214,17 @@ protected function loadData($request, $filter)
         $collector->limit($rangeInfo->getCount());
         $collector->offset($rangeInfo->getOffset() + max(0, $rangeInfo->getPage() - 1) * $rangeInfo->getCount());
         $iterator = $collector->getMany();
-        return new VirtualArrayIterator(iterator_to_array($iterator, true), $totalCount, $rangeInfo->getPage(), $rangeInfo->getCount());
+        $users = iterator_to_array($iterator, true);
```
jonasraoni, line 217: Is this really needed? If it's an iterator, it should work at the `foreach`.

controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -203,7 +214,17 @@ protected function loadData($request, $filter)
...
+        foreach ($users as $user) {
+            $roles = array();
```
jonasraoni, line 219: We're using the newer syntax to declare arrays `[]`.

controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -203,7 +214,17 @@ protected function loadData($request, $filter)
...
+            $userGroupsIterator = Repo::userGroup()->userUserGroups($user->getId(), $context->getId());
+            $userGroups = iterator_to_array($userGroupsIterator);
```
jonasraoni, line 221: The same comment from above, I think it's not needed.

controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -203,7 +214,17 @@ protected function loadData($request, $filter)
...
+            foreach ($userGroups as $userGroup) {
+                $roles[] = $userGroup->getLocalizedName();
+            }
```
jonasraoni, line 224: Not sure if you like, but you might use a functional approach (either way are ok for us):

```php
$userGroupsIterator->map(fn (UserGroup $userGroup) => $userGroup->getLocalizedName())->join(', ');
```

controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -203,7 +214,17 @@ protected function loadData($request, $filter)
...
+            $user->setData('roles', join(', ', $roles));
```
jonasraoni, line 225: The separator is localized, so you can use: `__('common.commaListSeparator')` instead of `', '`.

jonasraoni, line 225: Still about this line, I think `setData('roles')` might be dangerous, I see the potencial to raise conflicts in the future.

So, I think the right solution is to use another type of "cell provider" on the `$this->addColumn()`, probably the `ColumnBasedGridCellProvider`. See an example here:
https://github.com/pkp/pkp-lib/blob/777b10995699b8419a61f62adbceb5cfcba8522d/classes/controllers/grid/feature/selectableItems/ItemSelectionGridColumn.php#L38

And the implementation here:
https://github.com/pkp/pkp-lib/blob/777b10995699b8419a61f62adbceb5cfcba8522d/classes/controllers/grid/feature/selectableItems/ItemSelectionGridColumn.php#L73-L80

To avoid creating a new file/class just for this, you can make use of a dynamic class, which would look something like this:
```php
$columnBasedGridCellProvider = new ColumnBasedGridCellProvider();
$this->addColumn(
    new class (
        'roles',
        'user.roles',
        null,
        null,
        $columnBasedGridCellProvider
    ) extends GridColumn {
        public function getTemplateVarsFromRow($row) {
            //implementation
        }
    }
);
```

jonasraoni (changes requested):
Hi @nibou230!
I think it's ok, but I've suggested changes that will make the code more resistant to future updates, basically to use the `ColumnBasedGridCellProvider` instead of adding custom data to the User objects.

controllers/grid/settings/user/UserGridHandler.php
nibou230: Hi @jonasraoni, thanks for all the detailled review comments, it's much appreciated!

I did every requested change and I hope it aligns more with the standard of OJS this way. =)

controllers/grid/settings/user/UserGridHandler.php
```diff
@@ -140,6 +144,56 @@ public function initialize($request, $args = null)
...
+                    $titleTranslated = null,
+                    $template = null,
+                    $cellProvider = null,
+                    $flags = [],
+                    int $contextId = null
+                ) {
+                    $this->_contextId = $contextId;
+
+                    parent::__construct($id, $title, $titleTranslated, $template, $cellProvider, $flags);
+                }
```
jonasraoni, line 184: Given that we're using a lot of `Application::get()->getRequest()` in the codebase, I think it's ok to use it on [the line 190](https://github.com/pkp/pkp-lib/pull/9385/files#diff-9bdf4df1fd7942644c46d93df32286ddd88a6f6f5fd6ad7dacf3b9c378b4b8a2R190) to get the `Request > Context` and remove all these lines that I highlighted. Then I don't have nothing more to comment 😁

controllers/grid/settings/user/UserGridHandler.php
jonasraoni: Thanks for your patience 🤗

controllers/grid/settings/user/UserGridHandler.php
nibou230: Hello, I'm glad we can remove all those lines!

asmecher:
Thanks, @jonasraoni and @nibou230! I've merged this to `main` and cherry-picked back to `stable-3_4_0`, since it doesn't introduce any e.g. new locale keys.

Issue #9386, opened 2023-10-06
[A11Y] Missing label on Google reCaptcha form triggering accessibility non-compliance issue


israelcefrin:
## Issue Description
// What is the nature of the issue? How does it impact users?
Google reCaptcha form is missing a label to the invisible textarea. It triggers an accessibility issue compliance when running automated accessibility audit tools. Despite of it may be a false positive since hidden elements are removed from the Accessibility tree, it is annoying and might become overwhelming for those running different tools.

## How to reproduce / Current behavior
// Step-by-step instructions explaining how to reproduce the problem. The environment required, explicit identification of the component in question
1. Click on  the "Register" in a journal with Google reCaptcha enabled
2. Run an automated accessibility audit tool (e.g. WAVE)
3. See the issue

## What application are you using?
 OJS, OMP or OPS version X.X.X
OJS 3.3 

## Stack used 
- Device: Laptop
- OS: MacOS
- Browser: Chrome
- Version: 117

## Testing method
- Automated test: WAVE extension

## Action
// Describe action that can be taken to remediate the issue. 
Adding a hidden label to the form will fix the error. 
`<label for="g-recaptcha-response" style="display:none;" >Recaptcha response</label>`

## Additional information
// Please add any screenshots, code snippets, or other information we can use to investigate this issue.
![image](https://github.com/pkp/pkp-lib/assets/4992902/7a14602d-d0d4-4572-a5a4-8c098d2aa141)

israelcefrin:
Hi @jonasraoni 

There are 2 PRs open, one for eacho OJS version:
OJS 3.3 - https://github.com/pkp/pkp-lib/pull/9387
OJS 3.4 - https://github.com/pkp/pkp-lib/pull/9389

On the second one I've added the `hidden` property as you have suggested, but I had to keep the` display:none; `just as a fallback for older browsers.

jonasraoni:
Great, I've merged both and cherry-picked the commit from `stable-3_4_0` into `main` (now we're supporting 3 branches/releases: LTS + current + next) 😁

PR #9387, opened 2023-10-06
Adding missing label to google reCaptcha form

israelcefrin:
This PR adds a label to the google reCaptcha form and prevents accessibility non-compliance errors from being triggered when users are running automated accessibility audit tools like WAVE.

Issue: https://github.com/pkp/pkp-lib/issues/9386

### DIFF
--- a/classes/user/form/RegistrationForm.inc.php
+++ b/classes/user/form/RegistrationForm.inc.php
@@ -91,7 +91,7 @@ class RegistrationForm extends Form {
 
 		if ($this->captchaEnabled) {
 			$publicKey = Config::getVar('captcha', 'recaptcha_public_key');
-			$reCaptchaHtml = '<div class="g-recaptcha" data-sitekey="' . $publicKey . '"></div>';
+			$reCaptchaHtml = '<div class="g-recaptcha" data-sitekey="' . $publicKey . '"></div><label for="g-recaptcha-response" style="display:none;">Recaptcha response</label>';
 			$templateMgr->assign(array(
 				'reCaptchaHtml' => $reCaptchaHtml,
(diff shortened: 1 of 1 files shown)


classes/user/form/RegistrationForm.inc.php
```diff
@@ -91,7 +91,7 @@ function fetch($request, $template = null, $display = false) {
 
 		if ($this->captchaEnabled) {
 			$publicKey = Config::getVar('captcha', 'recaptcha_public_key');
-			$reCaptchaHtml = '<div class="g-recaptcha" data-sitekey="' . $publicKey . '"></div>';
+			$reCaptchaHtml = '<div class="g-recaptcha" data-sitekey="' . $publicKey . '"></div><label for="g-recaptcha-response" style="display:none;">Recaptcha response</label>';
```
jonasraoni, line 94: @israelcefrin I think you can use the [hidden attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden).

jonasraoni (approved):

classes/user/form/RegistrationForm.inc.php
israelcefrin: Hi @jonasraoni, I can use/add hidden to the element on the 3.4 porting. Would you like me to add on OJS 3.3 as well?

jonasraoni: As you wish. Both ways are going to work, so I'm just sharing knowledge about this attribute 😁

israelcefrin: I will port it to OJS3.4 and see how it works.

Issue #9381, opened 2023-10-04
Limit reviewers' statistics to a journal in multi-journals OJS instance


pilasou:
As an editor of a journal hosted in a multi-journals OJS instance, I want to be able to limit the review statistics to my journal only. 

By default, when inviting reviewers, OJS provide statistics (number of invitations, already invited or not, etc.) that are global for all journals hosted. 

Improvement:  

- Add a site setting to setup wheter the statistics should be shared, defaulting to yes, to avoid breaking changes.
  
  ![image](https://github.com/pkp/pkp-lib/assets/31410053/ce4c017e-6edc-4468-afdd-e75bda44eaef)

- ~~Add a high-level configuration in the config.inc.php (reviewers_stats_local_only ; On/Off) which offer the ability to limit statistics to each of the journals or for the complete journal platform.
https://github.com/pkp/pkp-lib/assets/19291880/945d6f54-210d-42e0-99b4-784c3de46e63~~

This feature was already documented on PKP Forum (by me :°) https://forum.pkp.sfu.ca/t/ojs-3-reviewer-invite-window-statistics-configuration/75490

kristinmhoffmann:
As a host for a multi-journal OJS instance, I would appreciate having this option. Some of our journals have reviewers in common, but editors assume that they are seeing statistics that pertain to their journal only, and are confused to learn that that's not the case.

PR #9388, opened 2023-10-06
pkp/pkp-lib#9381 Limit reviewer statistics to the current journal

nibou230:
### How to test?

There are many ways to test this configuration. The test below illustrates the configuration using the “Active” statistics which indicates if a reviewer is already reviewing another article.  

- Use an OJS instance with at least 2 journals on it.
- Ensure that at least one reviewer is associated with both journals.
- Make a new submission in journal A and send it to review stage.
- Invite the reviewer.
- Connect as the reviewer and accept the invitation.
- Make a new submission in journal B and send it to review stage.
- Invite the reviewer and in the invite window:
  - If show_local_reviewer_statistics configuration is set to “On”, in the reviewer selection window, the reviewer will not appear as having 1 active review.
  - If show_local_reviewer_statistics configuration is set to "Off", in the reviewer selection window, the reviewer will appear as having 1 active review.

Use the configuration below in the **[interface]** section.

```
; Display local reviewer statistics (restricted to the current journal)
show_local_reviewer_statistics = On
```


jonasraoni (changes requested):
@nibou230 the introduction of new configuration settings, requires:
- Adding some description for the users (see https://github.com/pkp/ojs/blob/main/docs/release-notes/README-3.4.0#L7, but on this case a `3.5.0` file have to be created).
- A default value should also be added to the configuration template (https://github.com/pkp/ojs/blob/main/config.TEMPLATE.inc.php)

But in general, I'm against adding this setting to the `config.inc.php`. I think it should be a site/installation setting, then the administrator might change it using the UI at the URL `admin/settings`. Tagging @asmecher for a short opinion about the setting location.

p.s.: This setting isn't useful for OPS, as it doesn't have reviews.

classes/user/Collector.php
```diff
@@ -530,6 +531,7 @@ protected function buildReviewerStatistics(Builder $query): self
             return $this;
         }
 
+        $showLocalStats = Config::getVar('interface', 'show_local_reviewer_statistics');
```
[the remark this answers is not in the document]
nibou230: @jonasraoni, all good I will try to include this parameter as a site setting and I will give you some feedback when I'm done.

On a side note, if I need to make a change in both projects (pkp-lib and ojs), how am I suppose to contribute it? By making two pull requests? In this case, should I make a submodule update to link the PR done on the pkp-lib side? Those submodules are getting me crazy sometimes.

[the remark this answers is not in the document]
jonasraoni: That's right! You'll have to create a PR for the OJS repository. The last commit, should be updating the submodule to the right commit, and its commit message should contain the string `##username/branch` (e.g. `##nibou230/9381##`).

There's a script here: https://github.com/pkp/ojs/blob/main/tools/startSubmodulesTRAVIS.sh#L78, which will take care of downloading the code from your branch to run the tests based on the commit message.

See an example of such commit here: https://github.com/pkp/ojs/pull/4095/commits/cf48ca4174f744c802f91dda07b9d6e52aacbb98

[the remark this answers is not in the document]
nibou230: Hello @jonasraoni , I moved the setting to the site. For now, it's in the main **"Settings"** menu, but feel free to indicate what would be the best location. You can see it here:

image

pilasou:
Hi @jonasraoni, do changes we made comply with what you were recommanding ? Let us know if you have any comments or suggestions! Thanks very much!

classes/components/forms/site/PKPSiteConfigForm.php
```diff
@@ -73,5 +76,17 @@ public function __construct($action, $locales, $site)
             'size' => 'small',
             'value' => $site->getData('minPasswordLength'),
         ]));
+
+        // This parameter is only valuable in a multi-instances installation.
+        if ($numberOfJournals > 1) {
```
jonasraoni, line 81: I think this field should be always visible. It's not clear for me, as a user, that there's a hidden option. Also, it will allow the administrator to pre-configure everything properly, before creating the journals.

classes/components/forms/site/PKPSiteConfigForm.php
```diff
@@ -73,5 +76,17 @@ public function __construct($action, $locales, $site)
...
+            $this->addField(new FieldOptions('enableSharedReviewerStatistics', [
+                'label' => __('admin.settings.sharedReviewerStatistics'),
+                'description' => __('admin.settings.sharedReviewerStatistics.description'),
+                'options' => [
+                    ['value' => true, 'label' => __('admin.settings.sharedReviewerStatistics.enable')]
+                ],
+                'value' => $site->getData('enableSharedReviewerStatistics') ? $site->getData('enableSharedReviewerStatistics') : true,
```
jonasraoni, line 88: Not important, but on another place you've used `getData('enableSharedReviewerStatistics') ?? true`, which looks cleaner.

jonasraoni:
Hi @pilasou! I was having a short vacation. I've just left minor comments :)

After handling those, I think it's ready to be merged.

schemas/site.json
```diff
@@ -112,6 +112,11 @@
 				"min:4"
 			]
 		},
+		"enableSharedReviewerStatistics": {
+			"type": "boolean",
+			"description": "Whether to share the reviewers statistics between all journals or keep them contextual to each individual journal.",
```
jonasraoni, line 117: Perhaps `Specify whether reviewer statistics can be shared with other contexts of the installation.`

locale/en/admin.po
```diff
@@ -318,6 +318,19 @@ msgstr "Introduction"
...
+
+msgid "admin.settings.sharedReviewerStatistics.description"
+msgstr "On a multi-journals OJS instance, the reviewer statistics displayed "
+"in the review request window (number of reviews submitted, one active or "
+"not, etc.) can be either for all journals hosted on the OJS instance or "
+"specific to the journal only. By default, reviewer statistics counts are for "
+"all journals hosted."
+
+msgid "admin.settings.sharedReviewerStatistics.enable"
+msgstr "Display reviewer statistics counts of all journals hosted"
```
jonasraoni, line 332: Perhaps `Display aggregated reviewer statistics`

locale/en/admin.po
```diff
@@ -318,6 +318,19 @@ msgstr "Introduction"
...
 
+msgid "admin.settings.sharedReviewerStatistics"
+msgstr "Reviewer statistics"
+
+msgid "admin.settings.sharedReviewerStatistics.description"
+msgstr "On a multi-journals OJS instance, the reviewer statistics displayed "
+"in the review request window (number of reviews submitted, one active or "
+"not, etc.) can be either for all journals hosted on the OJS instance or "
+"specific to the journal only. By default, reviewer statistics counts are for "
+"all journals hosted."
```
jonasraoni, line 329: The `pkp-lib` is shared with other applications (OJS/OMP/OPS), so we should avoid using specific terms (e.g. OJS/journal), otherwise we'll need to customize these texts for each application.

Perhaps `In a multi-context installation, reviewer statistics, such as the count of submitted reviews, can be displayed either individually for each context or aggregated collectively`

nibou230:
Hi @jonasraoni, hope you had a great vacation! =)

I will try to be less specific in `pkp-lib` next time. I did the required changes, thanks for the detailed review, as always!

jonasraoni (approved):

Issue #9434, opened 2023-10-20
Allow plugins to add/modify API endpoints of existing entity


touhidurabir:
**Describe the problem**
With the merge of https://github.com/pkp/pkp-lib/issues/7698 in the `main branch(Future 3.5.0 LTS)` , slim routing toolset set has been removed as one of the core dependency and the API implementation moved to laravel routing toolset completely . This has made the previously applied hook `APIHandler::endpoints` to tap into and modify/add/update obsolete and removed . 

**Describe the proposal**
Possible enhancement is to bring back the hook similar kind of hook `APIHandler::endpoints` to allow dynamic insertion of API routes on run time using laravel router which oversees the each routing process . This will be helpful for plugins to inject routes to any existing routes collection . 

**What application are you using?**
OJS, OMP or OPS version `main`

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9451
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4081 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1625 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/724 [TEST ONLY]

An example plugin to demonstrate the usage of plugin level api implementation at https://github.com/touhidurabir/apiExample

**Additional Information**
Feature request : https://forum.pkp.sfu.ca/t/create-custom-api-handler-ver-3-5/89101/4

touhidurabir:
@asmecher can you please preview the PR at https://github.com/pkp/pkp-lib/issues/9434#issue-1953801910

asmecher:
Thanks, @touhidurabir! I added a review: https://github.com/pkp/pkp-lib/pull/9451#pullrequestreview-1723622313

touhidurabir:
@asmecher can you take a look at the PRs at https://github.com/pkp/pkp-lib/issues/9434#issue-1953801910

asmecher:
@touhidurabir, it's promising, but it feels like the `APIRouter` is trying to be two things at once. I wonder if maybe it's worth coming up with a `PluginAPIRouter` class that lives parallel to `APIRouter`? That would keep `APIRouter` from having to accommodate two distinctly different URL patterns.

touhidurabir:
@asmecher All tests have passed so should be good to merge .

[...]

PR #9451, opened 2023-10-23
pkp/pkp-lib#9434 Allow plugin level API call

touhidurabir:
for pkp/pkp-lib#9434

### DIFF
--- a/classes/core/APIRouter.php
+++ b/classes/core/APIRouter.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/core/APIRouter.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2023 Simon Fraser University
+ * Copyright (c) 2023 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class APIRouter
@@ -23,12 +23,20 @@ namespace PKP\core;
 use APP\core\Application;
 use Exception;
 use Illuminate\Http\Response;
+use PKP\core\PKPBaseController;
+use PKP\core\PKPRequest;
+use PKP\handler\APIHandler;
+use PKP\session\SessionManager;
 
 class APIRouter extends PKPRouter
 {
+    /** 
+     * Define if the api call for a plugin implemented endpoint 
+     */
+    protected bool $_pluginApi = false;
+
     /**
      * Determines path info parts
-     *
      */
     protected function getPathInfoParts(): array
     {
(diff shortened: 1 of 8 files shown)


classes/core/APIRouter.php
```diff
@@ -3,8 +3,8 @@
 /**
  * @file classes/core/APIRouter.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2023 Simon Fraser University
+ * Copyright (c) 2023 John Willinsky
```
asmecher, line 7: (I'm not sure it's at all necessary, but our general practice is to extend the date when editing classes -- so these should be changed to `2014-2023` and `2000-2023`, not just `2023`.)

asmecher (changes requested):
This is not a bad idea, @touhidurabir, but can we come back to it later? We don't yet have a required use case for a plugin to add its own namespace to the API, and this does build in greater complexity to the `APIHandler`. At some point (I hope sooner than later) we can continue to replace our own routing/dispatching code with Laravel's, building on your work with the Laravel API router; when we do that, the fewer cases we have to replicate in Laravel, the better.

touhidurabir:
> This is not a bad idea, @touhidurabir, but can we come back to it later? We don't yet have a required use case for a plugin to add its own namespace to the API, and this does build in greater complexity to the APIHandler.

@asmecher I agree . This is something I did as just to show a proof of concept that allow us now an easier way to have this functionality . 

BTW, in the previous implementation of 3.3.0/3.4.0, we have hook to allow plugin to intervene the API endpoints , see at https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/handler/APIHandler.php#L192 . Now that we have moved to `Laravel Routing System` it's been removed and previous approach will not work . Do we need such functionality ? If so, we can bring in that part only which is included in this PR(in that case I can create a new issue and only port the changes for the part) . If no needed, we can leave it for now .

classes/core/APIRouter.php
```diff
@@ -90,14 +187,22 @@ public function route(PKPRequest $request): void
...
         $this->setHandler($handler);
+        $handler->runRoutes();
     }
 
     /**
      * Get the requested operation
+     *
+     * @return string
      */
-    public function getRequestedOp(PKPRequest $request): string
```
asmecher, line 100: Is this an accidental revert of adding typehints?

classes/core/PKPBaseController.php
```diff
@@ -114,8 +113,19 @@ public static function getRouteController(Request $request = null): ?static
         if (!$requestedRoute = static::getRequestedRoute($request)) {
             return null;
         }
+        
+        $calledRouteController = (new ReflectionFunction($requestedRoute->action['uses']))->getClosureThis();
+
+        // When the routes are added to router as a closure/callbale from other section like from a
```
asmecher, line 119: *callable

classes/handler/APIHandler.php
```diff
@@ -105,23 +144,66 @@ public function __construct(PKPBaseController $controller)
...
+            'roles' => $roles
+        ]);
+    }
+
+    /**
+     * Register the routes in the routes collection which was added via hook
+     */
+    protected function registerRoute(): void
+    {
+        if (empty($this->routesFromHook)) {
```
asmecher, line 190: I think these 3 lines can be removed.

asmecher (changes requested):
Just a few minor comments -- thanks, @touhidurabir!

classes/core/APIRouter.php
```diff
@@ -156,10 +179,8 @@ public function url(
         //
         // Assemble URL
         //
-        $pathInfoArray = array_merge(
-            $context,
-            ['api', Application::API_VERSION, $endpoint]
-        );
+        $pathInfoArray = ['api', Application::API_VERSION, $endpoint];
+        $pathInfoArray = array_merge($context, $pathInfoArray);
```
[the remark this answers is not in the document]
touhidurabir: rebased and reverted back .

classes/core/APIRouter.php
```diff
@@ -23,31 +21,47 @@
...
+        if (is_null($pathInfoParts)) {
+            return false;
+        }
+
+        if (count($pathInfoParts) < 2) {
+            return false;
+        }
+
+        // Context-specific API requests: [index.php]/{contextPath}/api
+        if (strtolower($pathInfoParts[1]) === 'api') {
```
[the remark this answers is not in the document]
touhidurabir: good catch , it's a leftover change from removing plugin own level api

classes/core/Dispatcher.php
```diff
@@ -198,15 +196,16 @@ public function setUserResolver(): void
...
-     * @param $shortcut the short name of the router that should be used to construct the URL
-     * @param $newContext Optional contextual path
-     * @param $handler Optional name of the handler to invoke
-     * @param $op Optional name of operation to invoke
-     * @param $path Optional string or array of args to pass to handler
-     * @param $params Optional set of name => value pairs to pass as user parameters
-     * @param $anchor Optional name of anchor to add to URL
-     * @param $escape Whether or not to escape ampersands for this URL; default false.
-     * @param $urlLocaleForPage Whether or not to override locale for this URL; Use '' to exclude.
+     * @param PKPRequest    $request            The request object
```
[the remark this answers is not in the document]
touhidurabir: reverted back

classes/handler/APIHandler.php
touhidurabir: removed .

touhidurabir:
@asmecher all requested changes have been addressed. I see that you have already added the removed hook `APIHandler::endpoints` in the `Hook::addUnsupportedHooks` .

asmecher:
Thanks, @touhidurabir; this needs a rebase before it can be merged, and could you also run it past the developer who was looking for the feature, along with a link to your example plugin?

Issue #9400, opened 2023-10-11
Display competing Interests in OJS once a reviewer accepts an invitation


pilasou:
As an editor, I want to be able to see if a reviewer has declared any competing interests when responding to a review request. 

Right now, in OJS, if a reviewer declare a potential competing interest, editors cannot see it when the reviewer complete his/her report. 

**Improvement:**
Display competing interest to the editor after the reviewer respond to the review request (no matter if the reviewer accept or reject the invitation), as illustrated below: 

![image](https://github.com/pkp/pkp-lib/assets/19291880/8bdb0345-3d54-4b0a-bb78-ea468f8bcd3f)

The competing interest declaration can be seen (as shown below) when opening the review details :  
![image](https://github.com/pkp/pkp-lib/assets/19291880/bde36c7b-9169-45f4-965d-9e21e42564bb)

PR #9452, opened 2023-10-23
pkp/pkp-lib#9400 Display competing interests once a reviewer accepts an invitation

nibou230:
### How to test?

1. Activate competing interest by putting text in "Settings, Workflow, Review, Reviewer Guidance, Competing Interests".
2. Make a new submission and send it to review stage.
3. Invite a new reviewer, declare competing interest and accept the invitation.
4. As editor, open the submission, next to reviewer's name, the label Competing interest must display.
5. Open the Review details window. Text typed by reviewer is displayed at the top of the window.

Note: this should also work when a reviewer is declining the review, but for an unknown reason, the competing interests are not saved to the database using this workflow. I can't find a way to bring that information to the appropriate handler method.

### DIFF
--- a/controllers/grid/users/reviewer/AuthorReviewerGridCellProvider.php
+++ b/controllers/grid/users/reviewer/AuthorReviewerGridCellProvider.php
@@ -75,7 +75,13 @@ class AuthorReviewerGridCellProvider extends DataObjectGridCellProvider
                 return ['label' => __($element->getReviewMethodKey())];
 
             case 'considered':
-                return ['label' => $this->_getStatusText($this->getCellState($row, $column), $row)];
+                $statusText = $this->_getStatusText($this->getCellState($row, $column), $row);
+                $reviewAssignment = $row->getData();
+                $competingInterests = $reviewAssignment->getCompetingInterests();
+                if ($competingInterests) {
+                    $statusText .= '<span class="details">' . __('reviewer.competingInterests') . '</span>';
+                }
+                return ['label' => $statusText];
 
             case 'actions':
                 // Only attach actions to this column. See self::getCellActions()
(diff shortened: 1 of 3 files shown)


controllers/grid/users/reviewer/AuthorReviewerGridCellProvider.php
```diff
@@ -75,7 +75,13 @@ public function getTemplateVarsFromRowColumn($row, $column)
                 return ['label' => __($element->getReviewMethodKey())];
 
             case 'considered':
-                return ['label' => $this->_getStatusText($this->getCellState($row, $column), $row)];
+                $statusText = $this->_getStatusText($this->getCellState($row, $column), $row);
+                $reviewAssignment = $row->getData();
+                $competingInterests = $reviewAssignment->getCompetingInterests();
+                if ($competingInterests) {
+                    $statusText .= '<span class="details">' . __('reviewer.competingInterests') . '</span>';
```
jonasraoni, line 82: @nibou230 Does it look very bad if we remove the `<span>`?
I just think it's better to not bring HTML into the PHP side (even though there are already some cases on the codebase).

If it doesn't look very nice, then I think it's ok to leave it this way (all these `Grid` classes are supposed to be removed sooner or later) 👌

jonasraoni (approved):

jonasraoni (commented):
Hi @nibou230, I've just left a small comment.
I didn't download the branch to test, but LGTM.

controllers/grid/users/reviewer/AuthorReviewerGridCellProvider.php
nibou230: Hi @jonasraoni, it looks alright. See the image below:

image

As you said, those grids are already heavily using HTML inside the PHP code. To make it look inline with other labels, I also propose to let the `<span>` in place for now if it's good for you!

jonasraoni: Yeah, I think it's better to leave the `<span>`. I'll merge 😁

PR #9457, opened 2023-10-24
Update PKPSubmissionFileService.inc.php

eirikhanssen:
See the discussion in https://github.com/pkp/pkp-lib/issues/7453

If $reviewRound is null here (which it might be), and we try to return null->getStageId(), then the activity log will hang.

Returning null if $reviewRound is null here instead of trying to return $reviewRound->getStageId() will make sure it is possible to see the activity log even if $reviewRound is null in this place.

This fix has solved our issue with activity log which hangs in ojs 3.3.0.8 and 3.3.0.13.

I reported this fix before, but I apologize for not making a pull request earlier.


jonasraoni (approved):
I didn't investigate it deeper, but I see a similar approach below this line, so I guess it must be ok (considering the problem of a non-existing review round already exists on the wild).

Issue #9450, opened 2023-10-22
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array


janevsky:
**Describe the bug**
Getting a php fatal error on submission step 1 when clicking save and continue without selecting category section. There is no notification "Please select an appropriate journal section for this submission." because the site keeps loading, and it should refresh and show the notification on the right and additional info under section selection. After a manual refresh the notification pops up. I've tested it on my old version 3.3.0-11 and its working as it should.

**To Reproduce**
1. Click on New submission 
2. Check everything in Submission Requirements and Corresponding Contact, leave sections empty.
3. Scroll down to Save and continue (without section selected)
4. Site keeps loading and the error occurs

**What application are you using?**
OJS version 3.3.0-14
PHP version 8.0.26

**Additional information**
PHP error log:
`[23-Oct-2023 00:19:23 Europe/Ljubljana] PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /home/xxx/domains/XXXdomainXXX/public_html/cache/t_compile/f3f61f20c8cf263a6a0dca9d99469f9f95568d26^4ef1996850f75e89a53007d1a8243f48d8f4148f_0.app.submissionformcategories..php:24
Stack trace:
#0 /home/xxx/domains/XXXdomainXXX/public_html/lib/pkp/lib/vendor/smarty/smarty/libs/sysplugins/smarty_template_resource_base.php(123): content_6532454bdff141_16311372()
#1 /home/xxx/domains/XXXdomainXXX/public_html/lib/pkp/lib/vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php(114): Smarty_Template_Resource_Base->getRenderedTemplateCode()
#2 /home/xxx/domains/XXXdomainXXX/public_html/lib/pkp/lib/vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php(216): Smarty_Template_Compiled->render()
#3 /home/xxx/domains/XXXdomainXXX/public_html/lib/pkp/lib/vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php(385): Smarty_Internal_Template->render()
[...]

PR #9460, opened 2023-10-25
pkp/pkp-lib#9450 fixed issue when no category/section selected at submission time

touhidurabir:
for pkp/pkp-lib#9450

### DIFF
--- a/templates/submission/form/categories.tpl
+++ b/templates/submission/form/categories.tpl
@@ -7,7 +7,7 @@
  *
  * Include categories for submissions.
  *}
-{if count($categories)}
+{if $categories && is_array($categories) && count($categories)}
 	{if $readOnly}
 		{fbvFormSection title="grid.category.categories" list=true}
 			{foreach from=$categories item="category" key="id"}


templates/submission/form/categories.tpl
```diff
@@ -7,7 +7,7 @@
  *
  * Include categories for submissions.
  *}
-{if count($categories)}
+{if $categories && is_array($categories) && count($categories)}
```
Vitaliy-1, line 10: Aren't the last two checks enough for this case?  Or  is `is_array()` throwing a warning here without the first check?

Vitaliy-1:
Thanks, @touhidurabir!

Just a small comment. Don't think we need anything else here, looks straightforward.

templates/submission/form/categories.tpl
touhidurabir: Problem is when `$categories` is `null` which is the cause of this issue . Again PHP now more strict as if a non countable is passed to `count` method, then it throws fatal error . just some extra precaution to make sure anything like previous error does not arise .

Issue #9345, opened 2023-09-29
All applications 3.4.0-3 MetadataChangedJob and UpdateSubmissionSearchJob do not load search plugins


nils-stefan-weiher:
**Describe the bug**
Since moving the submission search updates to jobs the hooks called from the class `MonographSearchIndex` (https://github.com/pkp/omp/blob/main/classes/search/MonographSearchIndex.php) or the class `ArcticleSearchIndex` (https://github.com/pkp/ojs/blob/main/classes/search/ArticleSearchIndex.php). Do not trigger, because no plugin is loaded for the queue worker.

The hooks for OMP are:
- `MonographSearchIndex::submissionMetadataChanged`
- `MonographSearchIndex::submissionFilesChanged`
- `MonographSearchIndex::monographChangesFinished`

For OJS:
- 'ArticleSearchIndex::articleMetadataChanged'
- 'ArticleSearchIndex::submissionFileChanged'
- 'ArticleSearchIndex::submissionFilesChanged'

**To Reproduce**
A search plugin has to be enabled which implements some of the hooks above.
Maybe important: Jobs are processed by an external worker process, i.e. 'job_runner = Off` in `config.inc.php`.

1. Update submission metadata or upload a submission file
2. Publish a submission
3. Wait for jobs to be processed
4. Observe that no hook function from the plugin is called

**What application are you using?**
OMP 3.4.0-3, but OJS should be the same.

**Additional information**

I believe this happens when you setup an external worker process and do not use the internal job runner, i.e., `job_runner = Off` because for the internal job queue the generic plugins would be loaded during the normal processing of the request from the web server.

nils-stefan-weiher:
Updated title for

asmecher:
@touhidurabir, could you have a look at this one?

Tribunal33:
@touhidurabir will this one be merged soon and will there be something to test?

touhidurabir:
>  will this one be merged soon and will there be something to test?

@Tribunal33 probably not. I have updated the milestone to 3.6 . This is a long outstanding issue and related to another core issue . I do have a fix here add as PR but I dont like the approach . we are looking for alternative solution .

touhidurabir:
Hi @bozana , PR updated and test is passing for OJS in branch 3.4.0. Please take a look at the PR at https://github.com/pkp/pkp-lib/issues/9345#issue-1918768222 . 

[...]

Issue #7604, opened 2022-01-10
Add ability to include user/context with CLI request


ewhanson:
**Describe the problem you would like to solve**
As the CLI tools are operating in a server context and do not originate from a request initiated by the browser/user, they do not include some pieces of data (e.g. a user object or a context) that parts of the application expect to be there. This typically happens when a user or context is retrieved from the request, e.g. `$request->getUser()->getUsername()` or `$request->getContext()->getId()` (see #7603). Reaching these parts of the application result in a fatal error, e.g. `PHP Fatal error:  Uncaught Error: Call to a member function getUsername() on null`.

**Describe the solution you'd like**
It would be useful to be able to specify these at the time of running the tool, along the lines of including these options as flags, e.g.:
```bash
php importExport.php --user=username --context=journalPath NativeImportExportPlugin import [...]
```

This option assumes the tool is operating in an environment in which it's safe to assume whoever is running the tool should be able to act as any user just by specifying a username. Another option would be to specify the user by API key.

**Who is asking for this feature?**
This has arisen from needs of the PKP hosting team, but would be applicable to journal administrators working with CLI tools.

ewhanson:
@asmecher, what's your take on this, and what do you think of a (user-facing) approach like the one I've suggested? In terms of testing implementation strategies, I've successfully added users to the request as a temporary workaround (`Registry::set('user', $user)`) as well as by adding an API key to the request, but haven't had any success with attaching a context yet.

asmecher:
We have a lot of clean-up to do with our long-standing practices of deep-passing objects like `$request` (and sometimes `$user`), or jamming things into pseudo-globals like `Registry` and `Application::get`, rather than better practices like dependency injection (which we're starting to make better use of).

Meanwhile the fatal errors are actually helpful in determining where we're making assumptions about data being available. In the example you point out (https://github.com/pkp/pkp-lib/issues/7603).

So for the short term I support your proposal, but for the `main` branch it might be worth disabling that behaviour when using strict mode (see `config.inc.php`).

jonasraoni:
Hi!
Last time I used the `NativeImportExportPlugin` I had some issues, which I've shared here: https://github.com/pkp/pkp-lib/issues/7143#issuecomment-952174311

jonasraoni:
[...]

PR #9463, opened 2023-10-26
pkp/pkp-lib#9345 added ability to retrive context for jobs on CLI session

touhidurabir:
for pkp/pkp-lib#9345

### DIFF
--- a/classes/cliTool/CommandLineTool.php
+++ b/classes/cliTool/CommandLineTool.php
@@ -57,7 +57,7 @@ class CommandLineTool
     /** @var \PKP\user\User the user provided */
     public ?User $user = null;
 
-    public function __construct($argv = [])
+    public function __construct($argv = [], $loadPlugins = true)
     {
         // Initialize the request object with a page router
         $application = Application::get();
@@ -68,8 +68,10 @@ class CommandLineTool
         $router->setApplication($application);
         $request->setRouter($router);
 
-        // Initialize the locale and load generic plugins.
-        PluginRegistry::loadCategory('generic');
+        if ($loadPlugins) {
+            // Initialize the locale and load generic plugins.
+            PluginRegistry::loadCategory('generic');
+        }
 
         $this->argv = isset($argv) && is_array($argv) ? $argv : [];
 
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -27,6 +27,7 @@ use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Database\MySqlConnection;
 use Illuminate\Support\Facades\DB;
 use PKP\config\Config;
+use PKP\context\Context;
 use PKP\db\DAORegistry;
 use PKP\facades\Locale;
 use PKP\security\Role;
@@ -128,6 +129,13 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     public $allProducts;
 
     /**
+     * Instance of context(journal/press/server) when running on CLI
+     *
+     * @var Context|null
+     */
+    protected ?Context $cliContext = null;
+
+    /**
      * Constructor
      */
     public function __construct()
@@ -208,6 +216,44 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
             $appVersion = $versionDao->getCurrentVersion()->getVersionString();
             Registry::set('appVersion', $appVersion);
         }
+    }
+
+    /**
+     * Set the context set when app running on CLI
+     */
+    public function setCliContext(int|Context $context = null): void
+    {
+        if (is_null($context)) {
+            return;
+        }
+        
+        if ($context instanceof Context) {
+            $this->cliContext = $context;
+            return;
+        }
+
+        /** @var \PKP\context\ContextDAO $contextDao */
+        $contextDao = DAORegistry::getDAO(ucfirst($this->getContextName()) . 'DAO');
+
+        if($context = $contextDao->getById($context)) {
+            $this->cliContext = $context;
+        }
+    }
+
+    /**
+     * Clear/unset the context set on CLI
+     */
+    public function clearCliContext(): void
+    {
+        $this->cliContext = null;
+    }
+
+    /**
+     * Get the context set when app running on CLI
+     */
+    public function getCliContext(): ?Context
+    {
+        return $this->cliContext;
     }
 
     /**
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -824,6 +824,30 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     }
 
     /**
+     * Check if application currently running on CLI
+     */
+    public static function isRunningOnCLI(string $scriptPath = null): bool
+    {
+        if (php_sapi_name() && strtolower(php_sapi_name()) === 'cli') {
+            return true;
+        }
+
+        if ($scriptPath) {
+            $serverVars = $_SERVER;
+
+            if (isset($serverVars['SCRIPT_NAME']) && strpos(strtolower($serverVars['SCRIPT_NAME']), strtolower($scriptPath)) !== false) {
+                return true;
+            }
+
+            if (isset($serverVars['SCRIPT_FILENAME']) && strpos(strtolower($serverVars['SCRIPT_FILENAME']), strtolower($scriptPath)) !== false) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
      * Get the property name for a section id
      *
      * In OMP, the section is referred to as a series and the
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -221,9 +221,10 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     /**
      * Set the context set when app running on CLI
      */
-    public function setCliContext(int|Context $context = null): void
+    public function setCliContext(int|Context|null $context = null): void
     {
         if (is_null($context)) {
+            $this->cliContext = null;
             return;
         }
         
@@ -235,9 +236,12 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
         /** @var \PKP\context\ContextDAO $contextDao */
         $contextDao = static::getContextDAO();
 
-        if($context = $contextDao->getById($context)) {
+        if ($context = $contextDao->getById($context)) {
             $this->cliContext = $context;
+            return;
         }
+
+        throw new \InvalidArgumentException('Invalid context ID');
     }
 
     /**
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -233,7 +233,7 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
         }
 
         /** @var \PKP\context\ContextDAO $contextDao */
-        $contextDao = DAORegistry::getDAO(ucfirst($this->getContextName()) . 'DAO');
+        $contextDao = static::getContextDAO();
 
         if($context = $contextDao->getById($context)) {
             $this->cliContext = $context;
@@ -821,30 +821,6 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
         if (!defined('RUNNING_UPGRADE')) {
             define('RUNNING_UPGRADE', true);
         }
-    }
-
-    /**
-     * Check if application currently running on CLI
-     */
-    public static function isRunningOnCLI(string $scriptPath = null): bool
-    {
-        if (php_sapi_name() && strtolower(php_sapi_name()) === 'cli') {
-            return true;
-        }
-
-        if ($scriptPath) {
-            $serverVars = $_SERVER;
-
-            if (isset($serverVars['SCRIPT_NAME']) && strpos(strtolower($serverVars['SCRIPT_NAME']), strtolower($scriptPath)) !== false) {
-                return true;
-            }
-
-            if (isset($serverVars['SCRIPT_FILENAME']) && strpos(strtolower($serverVars['SCRIPT_FILENAME']), strtolower($scriptPath)) !== false) {
-                return true;
-            }
-        }
-
-        return false;
     }
 
     /**
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -29,6 +29,7 @@ use Illuminate\Queue\Failed\DatabaseFailedJobProvider;
 use Illuminate\Support\Facades\Facade;
 use PKP\config\Config;
 use PKP\i18n\LocaleServiceProvider;
+use PKP\core\PKPQueueProvider;
 use PKP\proxy\ProxyParser;
 use Throwable;
 
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -406,6 +406,30 @@ class PKPContainer extends Container
     {
         return PKPApplication::isUnderMaintenance();
     }
+
+    /**
+     * Determine if the application is running in the console.
+     */
+    public function runningInConsole(?string $scriptPath = null): bool
+    {
+        if (strtolower(php_sapi_name() ?: '') === 'cli') {
+            return true;
+        }
+
+        if (!$scriptPath) {
+            return false;
+        }
+
+        if (mb_stripos($_SERVER['SCRIPT_NAME'] ?? '', $scriptPath) !== false) {
+            return true;
+        }
+
+        if (mb_stripos($_SERVER['SCRIPT_FILENAME'] ?? '', $scriptPath) !== false) {
+            return true;
+        }
+
+        return false;
+    }
 }
 
 if (!PKP_STRICT_MODE) {
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -465,9 +465,9 @@ class PKPContainer extends Container
         }
 
         // Flush output buffer and send response and allow script to continue if client disconnects.
-        // Flush and end output buffer (if started) and also the system buffer.
+        // Flush and end ALL output buffer levels (if started) and also the system buffer.
         ignore_user_abort(true);
-        if (ob_get_level() > 0) {
+        while (ob_get_level() > 0) {
             ob_end_flush();
         }
         flush();
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -435,6 +435,48 @@ class PKPContainer extends Container
 
         return false;
     }
+
+    /**
+     * Flush the output buffer to ensure all output is sent to the client before any background
+     * task processing (e.g. job processing, schedule task running) to avoid any potential output
+     * buffering issues which may cause client page load time and performance degradation.
+     * 
+     * This should be called before the background tasks starts to process and only when there are
+     * background tasks available to process. 
+     */
+    public function flushOutputBuffer(): void
+    {
+        // Disable flushing output buffer for unit tests as PHPUnit is quite sensitive to output buffer 
+        // manipulation during tests. The root cause is The PHPUnit configuration has
+        // `beStrictAboutOutputDuringTests="true"` which makes PHPUnit very sensitive to any output
+        // buffer manipulation during tests and mark it as risky.
+        if ($this->runningUnitTests()) {
+            return;
+        }
+
+        // Force flush and close connection for non-blocking behavior
+        // and set headers to close connection and specify content length (if buffer exists)
+        if (headers_sent() === false) {
+            header('Connection: close');
+            header('Content-Encoding: none');
+            if (ob_get_length() > 0) {
+                header('Content-Length: ' . ob_get_length());
+            }
+        }
+
+        // Flush output buffer and send response and allow script to continue if client disconnects.
+        // Flush and end output buffer (if started) and also the system buffer.
+        ignore_user_abort(true);
+        if (ob_get_level() > 0) {
+            ob_end_flush();
+        }
+        flush();
+
+        // For PHP-FPM (Nginx/Apache with FPM): Explicitly finish FastCGI request
+        if (function_exists('fastcgi_finish_request')) {
+            fastcgi_finish_request();
+        }
+    }
 }
 
 if (!PKP_STRICT_MODE) {
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -111,6 +111,11 @@ class PKPContainer extends Container
             }
         );
 
+        $this->singleton(
+            'jobRunner',
+            fn ($app) => \PKP\queue\JobRunner::getInstance($app['pkpJobQueue'])
+        );
+
         Facade::setFacadeApplication($this);
     }
 
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -456,7 +456,7 @@ class PKPContainer extends Container
 
         // Force flush and close connection for non-blocking behavior
         // and set headers to close connection and specify content length (if buffer exists)
-        if (headers_sent() === false) {
+        if (!headers_sent()) {
             header('Connection: close');
             header('Content-Encoding: none');
             if (ob_get_length() > 0) {
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -22,6 +22,8 @@ use APP\core\Application;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Queue\Events\JobFailed;
+use Illuminate\Queue\Events\JobProcessed;
+use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\QueueServiceProvider as IlluminateQueueServiceProvider;
 use Illuminate\Queue\Worker;
 use Illuminate\Queue\WorkerOptions;
@@ -151,6 +153,36 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                     'trace' => $event->exception->getTrace(),
                 ])
             );
+        });
+
+        Queue::createPayloadUsing(function(string $connection, string $queue, array $payload) {
+            
+            // If a job dispatch on CLI (e.g. TestJobSuccess/TestJobFailure),
+            // there is simple no way to get the context(journal/press/server) information
+            // and no need to set the context id information in job payload
+            if (runOnCLI()) {
+                return [];
+            }
+
+            return [
+                'context_id' => Application::get()->getRequest()->getContext(),
+            ];
+        });
+
+        Queue::before(function(JobProcessing $event){
+            // Set the context for current CLI session if available right before job start processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (runOnCLI()) {
+                Application::get()->setCliContext($event->job->payload()['context_id'] ?? null);
+            }
+        });
+
+        Queue::after(function(JobProcessed $event){
+            // Clear the context for current CLI session if available when job finish the processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (runOnCLI()) {
+                Application::get()->clearCliContext();
+            }
         });
     }
 
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -165,7 +165,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
             }
 
             return [
-                'context_id' => Application::get()->getRequest()->getContext(),
+                'context_id' => Application::get()->getRequest()->getContext()->getId(),
             ];
         });
 
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -161,7 +161,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
 
         Queue::failing(function (JobFailed $event) {
             $contextId = $event->job->payload()['context_id'] ?? 'unknown';
-            trigger_error("Job failed for context_id {$contextId}: {$event->exception->__toString()}");
+            error_log("Job failed for context_id {$contextId}: {$event->exception->__toString()}");
 
             app('queue.failer')->log(
                 $event->connectionName,
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -92,21 +92,23 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
     /**
      * Run the queue worker to process queue the jobs
      */
-    public function runJobInQueue(): void
+    public function runJobInQueue(): bool
     {
         $job = $this->getJobModelBuilder()->limit(1)->first();
 
         if ($job === null) {
-            return;
+            return false; // this will signal that there are no jobs to run
         }
 
         $laravelContainer = PKPContainer::getInstance();
 
         $laravelContainer['queue.worker']->runNextJob(
-            'database',
+            Config::getVar('queues', 'default_connection', 'database'),
             $job->queue ?? Config::getVar('queues', 'default_queue', 'queue'),
             $this->getWorkerOptions()
         );
+
+        return true;
     }
 
     /**
@@ -116,7 +118,13 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
     public function boot()
     {
         if (Config::getVar('queues', 'job_runner', true)) {
-            register_shutdown_function(function () {
+            $currentWorkingDir = getcwd();
+            register_shutdown_function(function () use ($currentWorkingDir) {
+                
+                // restore the current working directory
+                // see: https://www.php.net/manual/en/function.register-shutdown-function.php#refsect1-function.register-shutdown-function-notes
+                chdir($currentWorkingDir);
+
                 // As this runs at the current request's end but the 'register_shutdown_function' registered
                 // at the service provider's registration time at application initial bootstrapping,
                 // need to check the maintenance status within the 'register_shutdown_function'
@@ -139,7 +147,8 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         }
 
         Queue::failing(function (JobFailed $event) {
-            error_log($event->exception->__toString());
+            $contextId = $event->job->payload()['context_id'] ?? 'unknown';
+            error_log("Job failed for context_id {$contextId}: {$event->exception->__toString()}");
 
             app('queue.failer')->log(
                 $event->connectionName,
@@ -153,6 +162,12 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                     'trace' => $event->exception->getTrace(),
                 ])
             );
+
+            // Clear the context for current CLI session when job failed to process
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (app()->runningInConsole()) {
+                Application::get()->clearCliContext();
+            }
         });
 
         Queue::createPayloadUsing(function(string $connection, string $queue, array $payload) {
@@ -164,7 +179,8 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         Queue::before(function(JobProcessing $event){
             // Set the context for current CLI session if available right before job start processing
             // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
-            if (Application::isRunningOnCLI()) {
+            if (app()->runningInConsole()) {
+                // FIXME: should validate the context and fail the job is invalid ?
                 Application::get()->setCliContext($event->job->payload()['context_id'] ?? null);
             }
         });
@@ -172,7 +188,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         Queue::after(function(JobProcessed $event){
             // Clear the context for current CLI session if available when job finish the processing
             // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
-            if (Application::isRunningOnCLI()) {
+            if (app()->runningInConsole()) {
                 Application::get()->clearCliContext();
             }
         });
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -33,6 +33,7 @@ use PKP\config\Config;
 use PKP\job\models\Job as PKPJobModel;
 use PKP\queue\JobRunner;
 use PKP\queue\WorkerConfiguration;
+use Throwable;
 
 class PKPQueueProvider extends IlluminateQueueServiceProvider
 {
@@ -100,9 +101,9 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
             return false; // this will signal that there are no jobs to run
         }
 
-        $laravelContainer = PKPContainer::getInstance();
+        $queueWorker = app()->get('queue.worker'); /** @var \Illuminate\Queue\Worker $queueWorker */
 
-        $laravelContainer['queue.worker']->runNextJob(
+        $queueWorker->runNextJob(
             Config::getVar('queues', 'default_connection', 'database'),
             $job->queue ?? Config::getVar('queues', 'default_queue', 'queue'),
             $this->getWorkerOptions()
@@ -143,7 +144,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                     return;
                 }
 
-                // Not to run in unit test mode
+                // Not to run in unit test mode as part of the application lifecycle
                 if (app()->runningUnitTests()) {
                     return;
                 }
@@ -186,13 +187,56 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         if (!Application::get()->isUnderMaintenance()) {
             Queue::createPayloadUsing(function(string $connection, string $queue, array $payload) {
                 // if running in unit tests, no change to payload and immediate return
-                if (app()->runningUnitTests()) {
-                    return $payload;
+                // if (app()->runningUnitTests()) {
+                //     return $payload;
+                // }
+
+                // If a `context_id` already exists, will not try to set a new one.
+                if (array_key_exists('context_id', $payload)) {
+                    return [];
                 }
 
                 $contextId = null;
+
+                $jobClass = $payload['data']['commandName']; /** @var \PKP\jobs\BaseJob $jobClass */
+                $jobInstance = $payload['data']['command'];
+
+                // will not try to set context id if the job is not context aware
+                if (!$jobClass::contextAware()) {
+                    return [];
+                }
+
+                // will only try to deduce context id if the job explicitly defines it
+                if ($jobClass::shouldTryToDeduceContextFromArgs()
+                    && $jobInstance instanceof \Illuminate\Contracts\Queue\ShouldQueue
+                ) {
+                    try {
+                        foreach ($jobClass::contextDeductionArgMap() as $arg => $method) {
+
+                            if (!method_exists($jobClass, $method)) {
+                                error_log("Warning: No {$method} method defined for property {$arg} in job class {$jobClass}");
+                                continue;
+                            }
+
+                            $contextId = $jobClass::$method($jobInstance);
+
+                            if ($contextId) {
+                                break;
+                            }
+                        }
+                    } catch (Throwable $e) {
+                        error_log(
+                            sprintf(
+                                'failed to deduce context ID for job class %s exception %s',
+                                $jobClass,
+                                $e->__toString()
+                            )
+                        );
+                    }
+                }
                 
-                if (!array_key_exists('context_id', $payload)) {
+                // Fallback to determine the context id from app reqeust or CLI context if available
+                if (!$contextId) {
                     // This try/catch block is to facilitate the case when the application
                     // running in CLI mode and a job get dispatched in CLI where it trying to
                     // determine the context form the request() . But as the request delegate
@@ -201,17 +245,18 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                     // directly from there if exists a context and set it.
                     try {
                         $contextId = Application::get()->getRequest()->getContext()?->getId();
-                    } catch (\Throwable $e) {
+                    } catch (Throwable $e) {
                         // error_log('Failed to retrieve context ID from request on queue payload creator with exception: ' . $e->__toString());
-                        $contextId = Application::get()->getCliContext();
-                    }
-
-                    if ($contextId) {
-                        $payload['context_id'] = $contextId;
+                        $contextId = Application::get()->getCliContext()?->getId();
                     }
                 }
+                
 
-                return $payload;
+                if ($contextId) {
+                    return ['context_id' => $contextId];
+                }
+
+                return [];
             });
         }
         
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -19,8 +19,11 @@ declare(strict_types=1);
 namespace PKP\core;
 
 use APP\core\Application;
+use Illuminate\Database\PostgresConnection;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Contracts\Debug\ExceptionHandler;
-use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Events\JobProcessing;
@@ -31,7 +34,6 @@ use Illuminate\Support\Facades\Facade;
 use Illuminate\Support\Facades\Queue;
 use PKP\config\Config;
 use PKP\job\models\Job as PKPJobModel;
-use PKP\queue\JobRunner;
 use PKP\queue\WorkerConfiguration;
 use PKP\queue\PKPQueueDatabaseConnector;
 use Throwable;
@@ -54,9 +56,37 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
     }
 
     /**
-     * Get a job model builder instance to query the jobs table
+     * Apply context-aware filtering to the job query.
      */
-    public function getJobModelBuilder(): Builder
+    public function applyJobContextAwareFilter(
+        EloquentBuilder|QueryBuilder $jobQuery,
+        ?int $contextId = null
+    ): EloquentBuilder|QueryBuilder
+    {
+        if (DB::connection() instanceof PostgresConnection) {
+            return $jobQuery->where(
+                fn ($query) => $query
+                    ->whereRaw(
+                        "REGEXP_REPLACE(payload, '.*\"context_id\":\\s*([0-9]+).*', '\\1') = ?",
+                        [(string) $contextId]
+                    )
+                    ->orWhereRaw(
+                        "payload !~ '\"context_id\":\\s*[0-9]+'"
+                    )
+            );
+        }
+
+        return $jobQuery->where(
+            fn ($query) => $query
+                ->where('payload->context_id', $contextId)
+                ->orWhereNull('payload->context_id')
+        );
+    }
+
+    /**
+     * Get a job model eloquent builder instance to query the jobs table
+     */
+    public function getJobModelBuilder(): EloquentBuilder
     {
         return PKPJobModel::isAvailable()
             ->nonEmptyQueue()
@@ -148,6 +178,8 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                 if (app()->runningUnitTests()) {
                     return;
                 }
+                
+                // error_log('Shutdown function started at: ' . microtime(true));
 
                 $jobRunner = app('jobRunner'); /** @var \PKP\queue\JobRunner $jobRunner */
 
@@ -158,6 +190,8 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
                     ->withMaxMemoryConstrain()
                     ->withEstimatedTimeToProcessNextJobConstrain()
                     ->processJobs();
+
+                // error_log('Shutdown function ended at: ' . microtime(true));
             });
         }
 
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -160,7 +160,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
             // If a job dispatch on CLI (e.g. TestJobSuccess/TestJobFailure),
             // there is simple no way to get the context(journal/press/server) information
             // and no need to set the context id information in job payload
-            if (runOnCLI()) {
+            if (Application::isRunningOnCLI()) {
                 return [];
             }
 
@@ -172,7 +172,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         Queue::before(function(JobProcessing $event){
             // Set the context for current CLI session if available right before job start processing
             // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
-            if (runOnCLI()) {
+            if (Application::isRunningOnCLI()) {
                 Application::get()->setCliContext($event->job->payload()['context_id'] ?? null);
             }
         });
@@ -180,7 +180,7 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
         Queue::after(function(JobProcessed $event){
             // Clear the context for current CLI session if available when job finish the processing
             // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
-            if (runOnCLI()) {
+            if (Application::isRunningOnCLI()) {
                 Application::get()->clearCliContext();
             }
         });
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -158,7 +158,9 @@ class PKPQueueProvider extends IlluminateQueueServiceProvider
     public function boot()
     {
         if (Application::isInstalled() && !Application::isUpgrading()) {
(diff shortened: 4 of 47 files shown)


classes/core/PKPQueueProvider.php
```diff
@@ -145,6 +147,36 @@ public function boot()
...
+            // there is simple no way to get the context(journal/press/server) information
+            // and no need to set the context id information in job payload
+            if (runOnCLI()) {
+                return [];
+            }
+
+            return [
+                'context_id' => Application::get()->getRequest()->getContext(),
+            ];
+        });
```
touhidurabir, line 163: we need the context information and as mostly will be dispatched at web request life cycle, we can have the context information . But attaching it to job data is risky as some jobs already have it and it may cause unforseen impact, just adding it to payload is safe as it that has no impact on serialisation of data . 

only catch is that is a job dispatch another job as in those cases we simple have no way to determine the context information when running on CLI .

classes/core/PKPQueueProvider.php
```diff
@@ -145,6 +147,36 @@ public function boot()
...
+
+        Queue::before(function(JobProcessing $event){
+            // Set the context for current CLI session if available right before job start processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (runOnCLI()) {
+                Application::get()->setCliContext($event->job->payload()['context_id'] ?? null);
+            }
+        });
```
touhidurabir, line 171: we set the context information to application for each job if available right before job get to process so that following code can get retrieve context information 

```php
Application::get()->getRequest()->getContext();
```

classes/core/PKPQueueProvider.php
```diff
@@ -145,6 +147,36 @@ public function boot()
...
+
+        Queue::after(function(JobProcessed $event){
+            // Clear the context for current CLI session if available when job finish the processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (runOnCLI()) {
+                Application::get()->clearCliContext();
+            }
+        });
```
touhidurabir, line 179: As each job may have different context associated, it's just a bit safer clear out the cli session based context information once each job get processed .

classes/core/PKPRouter.php
```diff
@@ -178,6 +178,11 @@ public function getRequestedContextPath(PKPRequest $request): string
      */
     public function getContext($request, $forceReload = false)
     {
+        // If the app currently run on CLI, return the on CLI context if set
+        if (runOnCLI()) {
+            return Application::get()->getCliContext();
+        }
```
touhidurabir, line 184: If the app running on CLI, we for sure only have option to retrieve the context information if a cli session based context information is set. So trying to retrieve it from there (e.g. either the context object or null returned) .

asmecher (changes requested):
Thanks, @touhidurabir! I like the use of `createPayloadUsing` to provide the context ID, but I'm not sure about adding e.g. `setCliContext` to `PKPApplication`. Can we not simply provide `context_id` to the job and have it use as needed? Where in the call stack is the call to `Request::getContext` that you're needing to change?

What will this mean for jobs that get queued before `context_id` is added?

touhidurabir:
> Can we not simply provide `context_id` to the job and have it use as needed?

The use of this is not for the jobs to have the context information but to have it set so that our use of `Application::get()->getRequest()->getContext()` can retrieve the context information in any part of the execution flow when executing on CLI through jobs . That is why we are setting the context information on application level when running on CLI(via worker) though event `JobProcessing(right before job get to process)` and clearing it out at event `JobProcessed(right after a job get processed)` . 

> Where in the call stack is the call to `Request::getContext` that you're needing to change?

The real use of it to make the context available through app level so that call of `Application::get()->getRequest()->getContext()` is able to retrieve the context information any where . For example , the plugins `register` method where we have extensive use of `getEnabled($mainContextId)` which use the above call to determine the context. A job may dispatch some class but it can not pass the context information to a plugin as that loaded right when the app get booted and it same for both web and CLI life cycle . 

> What will this mean for jobs that get queued before `context_id` is added?

This will have no impact on jobs serialization/unserialization as it's added to job `payload`, not to the `data` key of a payload which used to reconstruct the job instance when it's picked from the queue to execute . As an example see the following jobs payload 
```json
{
    "uuid": "7c7bc4d0-0390-46cb-b7a3-76a95dab956e",
    "displayName": "PKP\\jobs\\metadata\\MetadataChangedJob",
    "job": "Illuminate\\Queue\\CallQueuedHandler@call",
    "maxTries": 3,
    "maxExceptions": 3,
    "failOnTimeout": true,
    "backoff": "5",
    "timeout": 60,
    "retryUntil": null,
    "data": {
        "commandName": "PKP\\jobs\\metadata\\MetadataChangedJob",
        "command": "O:36:\"PKP\\jobs\\metadata\\MetadataChangedJob\":3:{s:15:\"\\u0000*\\u0000submissionId\";i:1;s:10:\"connection\";s:8:\"database\";s:5:\"queue\";s:5:\"queue\";}"
    },
    "context_id": null
}
```
It's allow us a way to know which job is should run for which context and retrieve that information and set it on APP level without reconstructing the job instance . 

@asmecher I agree this is not an elegant solution and I am not happy with it either . But the purpose of this change is to provide a way to set the context on APP level when jobs are running on CLI thread (such as worker or through `run`  command via cron etc) . even if jobs are passed to have the context information, some part of the application does not get to directly communicate with a job so job can not pass the context information . And those parts depend on `Request::getContext` to get the context information . One such example if the `plugins` which loads at the app boot time and the `Hooks` which called depends on implementation . Another case can be when a jobs dispatch one or more jobs which get picked by worker immediately(which is absolutely normal if worker is free) where not possible to set up context information on APP level unless we set it explicitly .

asmecher:
Thanks, @touhidurabir, that's very clearly explained.

I consider much of the `Application` / `PKPApplication` classes to be a sort-of-global anti-pattern where we're just building up functions supporting entanglements between classes. Since the jobs toolkit is relatively new, I'm hoping we can avoid repeating the same problems there.

It's true that we don't have a good pattern for loading plugins independent of the context, and I think adding one in this task is asking for trouble. Especially for `stable-3_4_0`.

However, the plugin will know whether it needs to consider CLI (job) use cases. We have a pattern most plugins use by convention which demonstrates the problem:
```php
function register($category, $path, $mainContextId) {
    if (parent::register($category, $path, $mainContextId)) {
        if ($this->getEnabled($mainContextId)) {
            Hook::register('Some::Hook::Name', $this->handlerFunction(...));
        }
    }
    function handlerFunction($hookName, $args) {
        // Run some code knowing that the plugin is enabled for $mainContextId
    }
}
```
This uses the `getEnabled` function to determine the enabled status based on`$mainContextId`, which might be incorrect. So this won't work for CLI-based usage (unless the plugin is e.g. enabled on a site-wide basis, like the Lucene plugin).

However, that's just a convention. We could use instead:
```php
function register($category, $path, $mainContextId) {
    if (parent::register($category, $path, $mainContextId)) {
        Hook::register('Some::Hook::Name', $this->handlerFunction(...));
    }
    function handlerFunction($hookName, $args) {
        // This will get called when Some::Hook::Name is encountered, regardless of the plugin enabled status.
        // Now we can use $this->getEnabled to check enabled status based on a context ID provided by something in $args.
    }
}
```
This depends entirely on the plugin to know what it needs, and I think it'll avoid the problem of entanglements with `Application::get()->getContext()->getId()`.

It doesn't solve the problem of other parts of the code that expect to use `Application::get()->getContext()->getId()`, but those are also anti-patterns that I'd like us to find alternatives for. That's not new; we've run into those (and tried to resolve them) for the previous jobs toolkit.

touhidurabir:
> However, that's just a convention. We could use instead:
```php
function register($category, $path, $mainContextId) {
    if (parent::register($category, $path, $mainContextId)) {
        Hook::register('Some::Hook::Name', $this->handlerFunction(...));
    }
    function handlerFunction($hookName, $args) {
        // This will get called when Some::Hook::Name is encountered, regardless of the plugin enabled status.
        // Now we can use $this->getEnabled to check enabled status based on a context ID provided by something in $args.
    }
}
```

@asmecher  This will solve the issue when plugin checked for enable or not but this still does not solve the issue of getting context information in `Hook` callbacks and they will continue to throw fatal error when try to execute . for example , in the [lucene](https://github.com/pkp/lucene) plugin, still have implementation that try to retrieve the context in hook callbacks such as 

```php
// Get the journal object from the context (optimized).
$router = $request->getRouter();
$journal = $router->getContext($request);
/* @var $journal Journal */
```
at https://github.com/pkp/lucene/blob/main/LucenePlugin.php#L394-L398

```php
// Get the request.
// Get the context
$journal = $request->getContext();
```
at https://github.com/pkp/lucene/blob/main/LucenePlugin.php#L791-L794

However as that hook pass `$submission` we can retrieve the context from there but need to update the plugin code to make it compatible with CLI/Worker thread based jobs. 

This is one example . Also this is not the only case . A lot of hook registration does not pass context information at hook registration time and in those cases a lot of implementations use the `Application::get()->getRequest()->getContext()` to retrieve the context information . And our only option is to update the code base everywhere to make the retrieval of context from passed objects which is probably the best solution but not an easy fix and may introduce some breaking changes .

asmecher:
The careless use of `Request` in potentially "offline" scenarios is a common error in our codebase with a lot of history. Have a look at this issue/discussion:

https://github.com/pkp/pkp-lib/issues/7604

There will be similar problems with `Request::getUser` too, for example -- even worse in that case, it may return the wrong user entirely (if the scheduled task/job is being processed on the back of another user's request)!

I would prefer that we resolved the entanglement of model (jobs, entities, etc) and controller (handler) areas of the MVC design, rather than masking them by permitting e.g. `Request::getContext` to be executed in headless cases. This does mean resolving them one-by-one, but in practice they aren't too complicated. I'm happy to help if you have a stack trace that shows a problem.

In the examples you included [in the comment above](https://github.com/pkp/pkp-lib/pull/9463#issuecomment-1786687599), I think the code is actually OK as it is -- those are for hook handlers for the `LoadHandler` and `TemplateManager::display`, both of which are intended to be invoked when handling a web-based request, not a headless operation like an indexing process. Those are for actual user-facing operations, not offline jobs.

touhidurabir:
> In the examples you included https://github.com/pkp/pkp-lib/pull/9463#issuecomment-1786687599, I think the code is actually OK as it is -- those are for hook handlers for the LoadHandler and TemplateManager::display, both of which are intended to be invoked when handling a web-based request

@asmecher you are right . However such usage e.g. `Application::get()->getRequest()->getContext()` can be and probably will be used in many implementation of hook callbacks and unknowingly used by others . But I agree, we should start moving away from similar implementation . 

But the approach that you have mentioned 
```php
function register($category, $path, $mainContextId) {
    if (parent::register($category, $path, $mainContextId)) {
        Hook::register('Some::Hook::Name', $this->handlerFunction(...));
    }
    function handlerFunction($hookName, $args) {
        // This will get called when Some::Hook::Name is encountered, regardless of the plugin enabled status.
        // Now we can use $this->getEnabled to check enabled status based on a context ID provided by something in $args.
    }
}
```
To have it, we need to update all the plugins to remove the use of `getEnabled()` use from the `register` method . But it seems to me that will rise another issue . If a plugin is disable , the following code always return `true` on CLI mode 
```php
parent::register($category, $path, $mainContextId)
```
So that means, even for a context for which a plugin(for example Lucene plugin in my testing), it will continue to run even when the plugin is disable in CLI mode . The `getEnabled` helps it to determine if it should run only if the plugin is enabled for the appropriate context . without that check, it will run and may cause unexpected or undesired output , please correct me if I am mistaken . 

Another option I can think about the if the above assumption is wrong is to add a check in the implementation of `getEnabled` method to see if it running on CLI mode and if no context id no passed and so pass true . That allow us handle this case without changing each of plugin implementation .

asmecher:
Sorry for the long wait, @touhidurabir!

> To have it, we need to update all the plugins to remove the use of getEnabled() use from the register method.

The vast majority of plugins are enabled and used only within a single journal, and will not use scheduled tasks; the usual `getEnabled(...)` convention will continue to work fine for them. This really is an unusual combination of needs, at least for now.

However as we continue to adopt jobs etc. more broadly it'll become a greater need. Can you file a new discussion outlining the issue and writing up a few options? Then tag me and I'll add some thoughts/ideas over there. It's probably not going to be an easy one to solve properly.

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +213,98 @@ public function boot()
...
+                    // determine the context form the request() . But as the request delegate
+                    // to router and in CLI mode, perhaps the router is not available which throws
+                    // as \AssertionError. In that case it will fallback to CLI context to get
+                    // directly from there if exists a context and set it.
+                    try {
+                        $contextId = Application::get()->getRequest()->getContext()?->getId();
+                    } catch (Throwable $e) {
+                        // error_log('Failed to retrieve context ID from request on queue payload creator with exception: ' . $e->__toString());
+                        $contextId = Application::get()->getCliContext()?->getId();
+                    }
```
touhidurabir: wondering should we be more strict that lenient here ? that is if a job is context aware, it must provide a context id and we will not try to second guess it from the `request` which is not very reliable in CLI . IMO, we should be strict .

classes/queue/ContextAwareJob.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+namespace PKP\queue;
+
+interface ContextAwareJob
+{
+    /**
+     * Get the context id of this context aware job
+     */
+    public function getContextId(): ?int;
```
touhidurabir: Again , should we be strict or lenient ? that is , if this is a context aware job, must pass a valid context id of int type and nothing else (e.g. not even nullable) . IMO, we should be strict .

classes/cliTool/CommandLineTool.php
```diff
@@ -57,7 +57,7 @@ class CommandLineTool
     /** @var \PKP\user\User the user provided */
     public ?User $user = null;
 
-    public function __construct($argv = [])
+    public function __construct($argv = [], $loadPlugins = true)
```
jonasraoni, line 60:
```suggestion
    public function __construct(?array $argv = [], bool $loadGenericPlugins = true)
```

Or perhaps:

```php
    public function __construct(?array $argv = [], ?array $loadPlugins = ['generic'])
```

classes/core/PKPApplication.php
```diff
@@ -127,6 +128,13 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     public $enabledProducts = [];
     public $allProducts;
 
+    /**
+     * Instance of context(journal/press/server) when running on CLI
+     *
+     * @var Context|null
```
jonasraoni, line 134:
```suggestion
```

classes/core/PKPApplication.php
```diff
@@ -210,6 +218,44 @@ class_alias('\PKP\core\PKPApplication', '\PKPApplication');
...
+        if (is_null($context)) {
+            return;
+        }
+        
+        if ($context instanceof Context) {
+            $this->cliContext = $context;
+            return;
+        }
+
+        /** @var \PKP\context\ContextDAO $contextDao */
```
jonasraoni, line 236: I think it's better to add the type directly to the method.

```suggestion
```

classes/core/PKPApplication.php
```diff
@@ -210,6 +218,44 @@ class_alias('\PKP\core\PKPApplication', '\PKPApplication');
...
     }
 
+    /**
+     * Set the context set when app running on CLI
+     */
+    public function setCliContext(int|Context $context = null): void
+    {
+        if (is_null($context)) {
+            return;
+        }
```
jonasraoni: I think it should be possible to setup with `null` (contextless execution). If you don't agree, then the default value (`= null`) should be removed.

As we're not using the `int|` part, I just saw one call passing a `Context` object. Then, I'd opt to not include code that's not being used.

In fact, I think we should turn the property `$context` into `public` and remove the related methods (get/set/clear)...

To be clearer, I'm thinking ahead, when we'll have [getters/setters](https://wiki.php.net/rfc/property-hooks) (PHP 8.4)... The "getX()/setX()" is an old practice ("what if...") which should be long gone.

Therefore, I think we shouldn't introduce more getters/setters on the codebase if not required.

classes/core/PKPApplication.php
```diff
@@ -210,6 +218,44 @@ class_alias('\PKP\core\PKPApplication', '\PKPApplication');
...
+            $this->cliContext = $context;
+            return;
+        }
+
+        /** @var \PKP\context\ContextDAO $contextDao */
+        $contextDao = static::getContextDAO();
+
+        if($context = $contextDao->getById($context)) {
+            $this->cliContext = $context;
+        }
```
jonasraoni: I think the method should always do what's meant to do (setup the current context) or fail with some noise.
Failures can be handled at the calling site (e.g. skip executing a job if the underlying context was removed).

```suggestion
        $this->cliContext = $contextDao->getById($context) ?? throw new InvalidArgumentException('Invalid context ID');
```

classes/core/PKPApplication.php
```diff
@@ -210,6 +218,44 @@ class_alias('\PKP\core\PKPApplication', '\PKPApplication');
...
+    }
+
+    /**
+     * Clear/unset the context set on CLI
+     */
+    public function clearCliContext(): void
+    {
+        $this->cliContext = null;
+    }
+
```
jonasraoni, line 254: If we allow setting up the a `null` context on the `setCliContext()`, then this method can be removed.

classes/core/PKPContainer.php
```diff
@@ -405,6 +411,72 @@ public function isDownForMaintenance()
...
+
+        if (mb_stripos($_SERVER['SCRIPT_NAME'] ?? '', $scriptPath) !== false) {
+            return true;
+        }
+
+        if (mb_stripos($_SERVER['SCRIPT_FILENAME'] ?? '', $scriptPath) !== false) {
+            return true;
+        }
+
+        return false;
```
jonasraoni, line 436: Out of curiosity, why can't we go with the code below?

```suggestion
        return strtolower(php_sapi_name() ?: '') === 'cli';
```

classes/core/PKPContainer.php
```diff
@@ -405,6 +411,72 @@ public function isDownForMaintenance()
...
+        // manipulation during tests. The root cause is The PHPUnit configuration has
+        // `beStrictAboutOutputDuringTests="true"` which makes PHPUnit very sensitive to any output
+        // buffer manipulation during tests and mark it as risky.
+        if ($this->runningUnitTests()) {
+            return;
+        }
+
+        // Force flush and close connection for non-blocking behavior
+        // and set headers to close connection and specify content length (if buffer exists)
+        if (headers_sent() === false) {
```
jonasraoni:
```suggestion
        if (!headers_sent()) {
```

classes/core/PKPQueueProvider.php
```diff
@@ -50,9 +62,37 @@ public function forQueue(string $queue): self
...
+                    ->whereRaw(
+                        "REGEXP_REPLACE(payload, '.*\"context_id\":\\s*([0-9]+).*', '\\1') = ?",
+                        [(string) $contextId]
+                    )
+                    ->orWhereRaw(
+                        "payload !~ '\"context_id\":\\s*[0-9]+'"
+                    )
+            );
+        }
+
```
jonasraoni: PostgreSQL 9.2 supports JSON queries and we're requiring 9.5 for the 3.4 applications. So I'd guess the query below, should cover both (I believe the `payload->context_id` is a "JSON syntax" provided by Laravel).

```suggestion
```

classes/core/PKPQueueProvider.php
```diff
@@ -90,21 +130,25 @@ public function runJobsViaDaemon(string $connection, string $queue, array $worke
     /**
      * Run the queue worker to process queue the jobs
      */
-    public function runJobInQueue(): void
+    public function runJobInQueue(?EloquentBuilder $jobBuilder = null): bool
     {
-        $job = $this->getJobModelBuilder()->limit(1)->first();
+        $job = $jobBuilder
+            ? $jobBuilder->limit(1)->first()
+            : $this->getJobModelBuilder()->limit(1)->first();
```
jonasraoni, line 141:
```suggestion
        $job = ($jobBuilder ?? $this->getJobModelBuilder())->limit(1)->first();
```

classes/core/PKPQueueProvider.php
```diff
@@ -113,8 +157,19 @@ public function runJobInQueue(): void
      */
     public function boot()
     {
+        if (Application::isInstalled() && !Application::isUpgrading()) {
```
jonasraoni: I guess this can be replaced by the `isUnderMaintenance()`, but as I see a couple of checks below, perhaps it makes sense to have an early return here.
```suggestion
        if (Application::isUnderMaintenance()) {
            return;
        }
```

classes/core/PKPQueueProvider.php
```diff
@@ -113,8 +157,19 @@ public function runJobInQueue(): void
...
+            $this->isMultiContextSite = DB::table(Application::getContextDAO()->tableName)
+                ->where('enabled', true)
```
jonasraoni: Not a big deal, but I think it makes more sense to use the same data type.

```suggestion
                ->where('enabled', 1)
```

Also, conceptually, if there are many registered contexts, the system considers it's a multi-context site, even though not all are enabled.

classes/core/PKPQueueProvider.php
```diff
@@ -127,17 +182,36 @@ public function boot()
...
+
+                $jobRunner
+                    ->setCurrentContextId(Application::get()->getRequest()->getContext()?->getId())
                     ->withMaxExecutionTimeConstrain()
                     ->withMaxJobsConstrain()
                     ->withMaxMemoryConstrain()
                     ->withEstimatedTimeToProcessNextJobConstrain()
                     ->processJobs();
+
+                // error_log('Shutdown function ended at: ' . microtime(true));
```
jonasraoni, line 216:
```suggestion
```

classes/core/PKPQueueProvider.php
```diff
@@ -127,17 +182,36 @@ public function boot()
...
             });
         }
 
         Queue::failing(function (JobFailed $event) {
-            error_log($event->exception->__toString());
+            $contextId = $event->job->payload()['context_id'] ?? 'unknown';
+            error_log("Job failed for context_id {$contextId}: {$event->exception->__toString()}");
```
jonasraoni:
```suggestion
            error_log("Job failed for context_id {$contextId}: {$event->exception}");
```

classes/core/PKPQueueProvider.php
```diff
@@ -127,17 +182,36 @@ public function boot()
                     return;
                 }
 
-                (new JobRunner($this))
+                // We only want to Job Runner for the web request life cycle
+                // not in any CLI based request life cycle
+                if (app()->runningInConsole()) {
```
jonasraoni, line 195: I think all of these safe checks followed by a `return;` can be moved out of the `register_shutdown_function()`, I mean, if we know before hand that it's not going to run, we don't need to register the shutdown function.

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+        });
+
+        // We will only register the payload creator if the application is not under maintenance
+        // to prevent any unintended DB access.
+        if (!Application::get()->isUnderMaintenance()) {
+            Queue::createPayloadUsing(function(string $connection, string $queue, array $payload) {
+                // if running in unit tests, no change to payload and immediate return
+                // if (app()->runningUnitTests()) {
+                //     return $payload;
+                // }
```
jonasraoni:
```suggestion
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                if ($jobInstance instanceof \PKP\queue\ContextAwareJob) {
+                    $contextId = $jobInstance->getContextId();
+
+                    // Specifically log the case when a context aware job return null context id
+                    if ($contextId === null) {
+                        error_log("Context aware job calss {$payload['displayName']} returned null contextId");
+                    }
+                }
+
+                // Fallback to determine the context id from app reqeust or CLI context if available
```
jonasraoni:
```suggestion
                // Fallback to determine the context id from app request or CLI context if available
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                    // This try/catch block is to facilitate the case when the application
+                    // running in CLI mode and a job get dispatched in CLI where it trying to
+                    // determine the context form the request() . But as the request delegate
+                    // to router and in CLI mode, perhaps the router is not available which throws
+                    // as \AssertionError. In that case it will fallback to CLI context to get
+                    // directly from there if exists a context and set it.
+                    try {
+                        $contextId = Application::get()->getRequest()->getContext()?->getId();
+                    } catch (Throwable $e) {
+                        // error_log('Failed to retrieve context ID from request on queue payload creator with exception: ' . $e->__toString());
```
jonasraoni:
```suggestion
```

classes/queue/ContextAwareJob.php
jonasraoni: I think it shouldn't be nullable, otherwise it's a `MaybeContextAwareJob`.

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                $contextId = null;
+                $jobInstance = $payload['data']['command']; /** @var \Illuminate\Contracts\Queue\ShouldQueue $jobInstance */
+
+                if ($jobInstance instanceof \PKP\queue\ContextAwareJob) {
+                    $contextId = $jobInstance->getContextId();
+
+                    // Specifically log the case when a context aware job return null context id
+                    if ($contextId === null) {
+                        error_log("Context aware job calss {$payload['displayName']} returned null contextId");
+                    }
```
jonasraoni: I think a `ContextAwareJob` should always have a contextId (whether it's still valid or not, that's another history).

```suggestion
```

classes/core/PKPQueueProvider.php
jonasraoni: A good reason to keep it nullable for now would be to address old jobs that are pending in the database and without a `contextId` defined, they are probably going to fail.

But it's better to avoid this type of technical debt (which tends to get worse/out of control) and force users to run all pending jobs before upgrading.

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+            });
+        }
+        
+        Queue::before(function(JobProcessing $event) {
+            // Set the context for current CLI session if available right before job start processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (app()->runningInConsole() && !Application::get()->isUnderMaintenance()) {
+                $contextId = $event->job->payload()['context_id'] ?? null;
+
+                // Validate the context exists if a context_id is provided
```
jonasraoni:
```suggestion
                // Validate if the context exists
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
                     'trace' => $event->exception->getTrace(),
                 ])
             );
+
+            // Clear the context for current CLI session when job failed to process
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (app()->runningInConsole() && !Application::get()->isUnderMaintenance()) {
+                Application::get()->clearCliContext();
+            }
```
jonasraoni, line 241: I guess the context will be set to a valid context (or null) later on, when the job is being initialized, then this block could be removed?!

```suggestion
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                }
+
+                return [];
+            });
+        }
+        
+        Queue::before(function(JobProcessing $event) {
+            // Set the context for current CLI session if available right before job start processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (app()->runningInConsole() && !Application::get()->isUnderMaintenance()) {
```
jonasraoni: Early return

```suggestion
            if (!app()->runningInConsole() || Application::get()->isUnderMaintenance()) {
                return;
            }
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                $contextId = $event->job->payload()['context_id'] ?? null;
+
+                // Validate the context exists if a context_id is provided
+                if ($contextId !== null) {
```
jonasraoni: Early return

```suggestion
                if (!$contextId) {
                    return;
                }
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                        // Fail the job immediately with a meaningful exception
+                        throw new \RuntimeException(
+                            "Job execution failed: Invalid context_id {$contextId}. The context does not exist in the database."
+                        );
+                    }
+
+                    Application::get()->setCliContext($context);
+
+                    // Initialize the locale and load generic plugins.
+                    \PKP\plugins\PluginRegistry::loadCategory('generic', false, $contextId);
```
jonasraoni: I've skimmed the `PluginRegistry` code and I'm not sure this will work very well.

Looks like all plugins will be loaded from disk, but the registration/initialization will have a kind of random behavior... For example an installation with 2 contexts and a single plugin:
- **Context A** has the plugin **inactive**: the plugin will be registered, but it won't be executed/enabled (e.g. register hooks and etc), due to the call to `getEnabled()` returning false (probably all plugins have this check), but its registration will succeed.
- **Context B** has the plugin **active**: As the plugin was already registered previously (it's stored on a global variable), then looks like its `register()` method will not be called.

So I think extra workarounds are needed, even though it might lead to unexpected consequences (I think no plugin was written with the assumption that it will be registered many times, so there's room for unexpected things regarding many hooks being registered and/or static variables stored on classes).

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+        });
+
+        Queue::after(function(JobProcessed $event) {
+            // Clear the context for current CLI session if available when job finish the processing
+            // Not necessary when jobs are running via JobRunner as that runs at the end of request life cycle
+            if (app()->runningInConsole() && !Application::get()->isUnderMaintenance()) {
+                Application::get()->clearCliContext();
+            }
+        });
+
```
jonasraoni, line 305: In my mind this isn't needed.

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+            if (!$this->isMultiContextSite) {
+                return true;
+            }
+
+            static $lockedContextId = null;
+
+            // Peek at next job WITHOUT popping (no lock, no reserve, no attempt increment)
+            $nextJob = DB::table('jobs')
+                ->where('queue', $event->queue)
+                ->where(fn($q) => $q->whereNull('reserved_at')
```
jonasraoni:
```suggestion
                ->where(fn (Builder $q) => $q->whereNull('reserved_at')
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+            // Non-context-aware jobs don't affect locking
+            if ($nextContextId === null) {
+                return true;
+            }
+
+            // First context-aware job sets the locked context
+            if ($lockedContextId === null) {
+                $lockedContextId = $nextContextId;
+                return true;
+            }
```
jonasraoni, line 345:
```suggestion
            $lockedContextId ??= $nextContextId;
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+
+            // Different context detected - quit BEFORE popping
+            if ($nextContextId !== $lockedContextId) {
```
jonasraoni:
```suggestion
            if ($nextContextId && $nextContextId !== $lockedContextId) {
```

classes/core/PKPRouter.php
```diff
@@ -178,6 +178,11 @@ public function getRequestedContextPath(PKPRequest $request): string
      */
     public function getContext($request, $forceReload = false)
     {
+        // If the app currently run on CLI, return the on CLI context if set
+        if (app()->runningInConsole() && ($cliContext = Application::get()->getCliContext())) {
+            return $cliContext;
+        }
```
jonasraoni, line 184: I guess this will probably conflict with other plugins, which are currently attempting to setup the context through workarounds (e.g. see the Native XML plugin, which setups a current user and context).

classes/task/ProcessQueueJobs.php
```diff
@@ -41,25 +41,40 @@ public function executeActions()
...
+        if (PKPContainer::getInstance()->runningInConsole('runScheduledTasks.php')) {
+            $maxJobCountToProcess = abs(Config::getVar('queues', 'job_runner_max_jobs', 30));
+            
+            while ($jobBuilder->count() && $maxJobCountToProcess) {
+                // if there is no more jobs to run, exit the loop
+                if ($jobQueue->runJobInQueue() === false) {
+                    break;
+                }
+                
+                --$maxJobCountToProcess;
```
jonasraoni, line 63:
```suggestion
            while ($jobBuilder->count() && $maxJobCountToProcess-- && $jobQueue->runJobInQueue()) {
```

tools/jobs.php
```diff
@@ -43,6 +43,8 @@
 use Symfony\Component\Console\Output\StreamOutput;
 use Throwable;
 
+error_reporting(E_ALL & ~E_DEPRECATED);
```
jonasraoni: There's a setting on the `config.inc.php` that controls this. Changing it here has room for possible side-effects (e.g. a script managing the output through `ob_*()` would capture errors), perhaps it's better to defer it for the `main` branch.

classes/core/PKPContainer.php
```diff
@@ -110,6 +111,11 @@ function ($app) {
             }
         );
 
+        $this->singleton(
+            'jobRunner',
+            fn ($app) => \PKP\queue\JobRunner::getInstance($app['pkpJobQueue'])
```
jonasraoni, line 116:
```suggestion
            fn (Application $app) => \PKP\queue\JobRunner::getInstance($app['pkpJobQueue'])
```

classes/core/PKPContainer.php
```diff
@@ -405,6 +411,72 @@ public function isDownForMaintenance()
...
+            header('Content-Encoding: none');
+            if (ob_get_length() > 0) {
+                header('Content-Length: ' . ob_get_length());
+            }
+        }
+
+        // Flush output buffer and send response and allow script to continue if client disconnects.
+        // Flush and end ALL output buffer levels (if started) and also the system buffer.
+        ignore_user_abort(true);
+        while (ob_get_level() > 0) {
```
jonasraoni, line 470:
```suggestion
        while (ob_get_level()) {
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+
+                    // Specifically log the case when a context aware job return null context id
+                    if ($contextId === null) {
+                        error_log("Context aware job calss {$payload['displayName']} returned null contextId");
+                    }
+                }
+
+                // Fallback to determine the context id from app reqeust or CLI context if available
+                // This is a placeholder solution of until all the needed job implements the ContextAwareJob
+                // interface and strictly return the context id as required
```
jonasraoni:
```suggestion
                // @todo: This workaround can be removed once all the jobs that require a context are implementing the ContextAwareJob
```

classes/core/PKPQueueProvider.php
```diff
@@ -151,6 +225,149 @@ public function boot()
...
+                if (!$contextId) {
+                    // This try/catch block is to facilitate the case when the application
+                    // running in CLI mode and a job get dispatched in CLI where it trying to
+                    // determine the context form the request() . But as the request delegate
+                    // to router and in CLI mode, perhaps the router is not available which throws
+                    // as \AssertionError. In that case it will fallback to CLI context to get
+                    // directly from there if exists a context and set it.
```
jonasraoni:
```suggestion
                    // determine the context from the request. The request delegates the call
                    // to the router in CLI mode, and the router might not be available, which throws
                    // an \AssertionError. In that case it will fallback to the default CLI context if available.
```

classes/queue/DatabaseQueue.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Set the context ID for context-aware job picking.
+     */
+    public function setContextId(?int $contextId): self
+    {
+        $this->contextId = $contextId;
+
+        return $this;
+    }
```
jonasraoni, line 51:
```suggestion
```

As I wrote before, I think we can avoid adding trivial getters/setters, unless it's required (e.g. N+1 arguments/processing).

classes/queue/DatabaseQueue.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Enable context-aware job picking.
+     */
+    public function enableJobInContextAwareMode(): self
+    {
+        $this->contextAware = true;
+
+        return $this;
+    }
+
```
jonasraoni, line 42: Having a `$contextId` defined is enough to know if it's a context aware queue.

```suggestion
    /**
     * The context id to use for context-aware job picking
     */
    protected ?int $contextId = null;

```

classes/queue/DatabaseQueue.php
```diff
@@ -0,0 +1,105 @@
...
+        // Determine the job popping method to use considering the context awareness
+        $jobPoppingMethod = $this->contextAware
+            ? 'getNextAvailableContextAwareJob'
+            : 'getNextAvailableJob';
+
+        return $this->database->transaction(function () use ($queue, $jobPoppingMethod) {
+            if ($job = $this->{$jobPoppingMethod}($queue)) {
+                return $this->marshalJob($queue, $job);
+            }
+        });
```
jonasraoni, line 76:
```suggestion
        return $this->database->transaction(
            fn () => ($job = $this->contextId ? $this->getNextAvailableContextAwareJob($queue) : $this->getNextAvailableJob($queue)) ? $this->marshalJob($queue, $job) : null;
        );
```

classes/queue/DatabaseQueue.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    protected function getNextAvailableContextAwareJob($queue)
+    {
+        $jobPickQuery = $this->database->table($this->table)
+                    ->lock($this->getLockForPopping())
+                    ->where('queue', $this->getQueue($queue))
+                    ->where(function ($query) {
+                        $this->isAvailable($query);
+                        $this->isReservedButExpired($query);
+                    });
```
jonasraoni, line 96:
```suggestion
            ->lock($this->getLockForPopping())
            ->where('queue', $this->getQueue($queue))
            ->where(function (Builder $query) {
                $this->isAvailable($query);
                $this->isReservedButExpired($query);
            });
```

classes/queue/DatabaseQueue.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $jobQueue = app('pkpJobQueue'); /** @var \PKP\core\PKPQueueProvider $jobQueue */
+        $job =  $jobQueue
+            ->applyJobContextAwareFilter($jobPickQuery, $this->contextId)
+            ->orderBy('id', 'asc')->first();
```
jonasraoni:
```suggestion
            ->orderBy('id', 'asc')
            ->first();
```

classes/queue/JobRunner.php
```diff
@@ -19,11 +19,23 @@
...
+    /*
+     * Singleton instance
+     */
+    private static $instance = null;
+
+    /**
+     * Static flag to prevent multiple runs within the same request.
+     */
+    protected bool $jobProcessing = false;
+
```
jonasraoni, line 38: Do we have a condition where the job runner would run twice? If not, then I think we can drop this.

classes/queue/JobRunner.php
```diff
@@ -65,12 +77,50 @@ class JobRunner
...
+    public static function getInstance(?PKPQueueProvider $jobQueue = null): self
+    {
+        if (self::$instance === null) {
+            self::$instance = new self();
+            self::$instance->jobQueue = $jobQueue ?? app('pkpJobQueue');
+        }
+
+        return self::$instance;
+    }
+
```
jonasraoni, line 115: Given the singleton management is done at the container level, we don't need to have another singleton management here.

By looking at the code, the `$jobQueue` initialization is required, so I think the code should be moved to the constructor (which matches the previous implementation).

```suggestion
    /**
     * Constructor
     */
    public function __construct(?PKPQueueProvider $jobQueue = null) {
        $this->jobQueue = $jobQueue ?? app('pkpJobQueue');
    }

```

classes/queue/JobRunner.php
```diff
@@ -65,12 +77,50 @@ class JobRunner
...
+    /*
+     * Set/Update the job queue
      */
-    public function __construct(PKPQueueProvider $jobQueue = null)
+    public function setJobQueue(PKPQueueProvider $jobQueue): static
```
jonasraoni, line 127: I think we can drop trivial setters/getters

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
         return $this;
     }
 
+    /**
+     * Check if the job runner is running in context-aware mode
+     */
+    public function isRunningInContextAwareMode(): bool
+    {
+        return $this->runInContextAwareMode;
+    }
```
jonasraoni, line 256: If the `$currentContextId` property is enough to understand if it's in a context aware mode, then this can be dropped.

```suggestion
```

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
...
+
+    /**
+     * Get the current context id
+     */
+    public function getCurrentContextId(): ?int
+    {
+        return $this->currentContextId;
+    }
+
```
jonasraoni, line 285: Drop trivial getters/setters.

```suggestion
```

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
...
+                // If Cache::get() returned NULL, the lock may have been cleared between
+                // add() and get(). Add a small delay and re-check to be safe.
+                if (!$cachedLock) {
+                    usleep(10000); // 10ms delay
+                    $cachedLock = Cache::get($this->getCacheKey());
+                    if ($cachedLock && $cachedLock['token'] !== $newToken) {
+                        return false;
+                    }
+                }
```
jonasraoni, line 348: The race condition check should be using an atomic operation, then these extra checks/re-checks won't be needed. I see Laravel has a `Cache::lock()`, but AFAICS, it's also not enough. If the installation is served by more than one server, a disk based file lock will not work (unless the cache folder is shared across all nodes and the file system is good enough to honor a network lock).

So, if the lock is really needed here, we should do the atomic operation using the database.

Is there a reason to not process jobs in parallel? In my mind it should work if coded properly (e.g. there shouldn't be N job runners processing the same job)..

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
...
+                if ($this->mayExceedMemoryLimitAtNextJob($this->jobProcessedOnRunner, $jobProcessingStartTime)) {
+                    return true;
+                }
+
+                // if there is no more jobs to run, exit the loop
+                if ($this->jobQueue->runJobInQueue($jobBuilder) === false) {
+                    return true;
+                }
+
+                $this->jobProcessedOnRunner = $this->jobProcessedOnRunner + 1;
```
jonasraoni, line 380:
```suggestion
                ++$this->jobProcessedOnRunner;
```

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
...
+     */
+    public function getJobProcessedCount(): int
+    {
+        return $this->jobProcessedOnRunner;
+    }
+
+    /**
+     * Get the current status of job runner to see if this is processing jobs.
+     * It will check for both in the current request life cycle and also
+     * in the other request life cycle.
```
jonasraoni, line 402:
```suggestion
     * Get the current status of job runner to see if it's processing jobs.
```

classes/queue/JobRunner.php
```diff
@@ -189,40 +239,200 @@ public function withEstimatedTimeToProcessNextJobConstrain(): self
...
+                }
+
+                if ($this->mayExceedMemoryLimitAtNextJob($this->jobProcessedOnRunner, $jobProcessingStartTime)) {
+                    return true;
+                }
+
+                // if there is no more jobs to run, exit the loop
+                if ($this->jobQueue->runJobInQueue($jobBuilder) === false) {
+                    return true;
+                }
```
jonasraoni, line 378:
```suggestion
            while (
                $jobBuilder->count()
                && !$this->exceededJobLimit($this->jobProcessedOnRunner)
                && !$this->exceededTimeLimit($jobProcessingStartTime)
                && !$this->exceededMemoryLimit()
                && !$this->mayExceedMemoryLimitAtNextJob($this->jobProcessedOnRunner, $jobProcessingStartTime)
                && $this->jobQueue->runJobInQueue($jobBuilder)
            ) {
```

jonasraoni (changes requested):
Hi! I left some comments, but I think it's important to check the comments regarding:
- Loading plugins (`loadCategory('generic', false, $contextId);`)
- Lock/parallelization (`$lockData = Cache::get($this->getCacheKey());`)

classes/core/PKPQueueProvider.php
touhidurabir: > A good reason to keep it nullable for now would be to address old jobs that are pending in the database and without a contextId defined, they are probably going to fail.

I think we are good there as this callback run as `Queue::createPayloadUsing` which only invoke when the job is being dispatched to the queue(basically right before being stored), have nothing to do with when job is being picked from queue and executed . so any jobs already in exists, have no impacts . 

> But it's better to avoid this type of technical debt (which tends to get worse/out of control) and force users to run all pending jobs before upgrading.

I fully agree that so better to remove it .

classes/core/PKPQueueProvider.php
touhidurabir: Yeah, but should be `!Application::isUnderMaintenance()` . Also we can not do a early return here as `boot` method may get more logics related to other part so we need the boot to continue .

classes/core/PKPQueueProvider.php
touhidurabir: As we can not do a early return, for now lets keep those checks there .

classes/core/PKPQueueProvider.php
touhidurabir: Not always as there can not non contextware jobs also so this provide some extra safe mechanism .

classes/core/PKPQueueProvider.php
touhidurabir: I think it's safe to have for following case : 
1. Job A is context aware and we have a cli context to 1 
2. Job B is not a context aware without clearing it, it will still have cli context to 1 . 

given that above case, it a extra safe mechanism .

classes/queue/JobRunner.php
touhidurabir: > The race condition check should be using an atomic operation, then these extra checks/re-checks won't be needed. I see Laravel has a Cache::lock()

Thats true but unfortunately we are using opcache for 3.4 which does not support atomic lock . `file` driver has been introduced in 3.5/main but as that's a configurable options, if someone moved to opcache, it's same . so we need this extensive(and i say bit ugly) checks. 

> So, if the lock is really needed here, we should do the atomic operation using the database.

Yes, but we not support database driver for cache yet if I am not mistaken . 

> Is there a reason to not process jobs in parallel

By al means jobs are supposed to run via CLI and determined by app deployment mechanism to finalise the number of worker processes . But jobs runner which runs as normal http request life cycle, there can be no limit . Also as part of normal request process, it can turn out quite expensive . So my idea is to allow just one instance of job runner at a time so that server does not get sudden spikes . if needed more, the number of jobs, time limits, and memory limits can be configured in config for job runner . 

> there shouldn't be N job runners processing the same job

I think we are good there as it's been handled by the worker with job reservation and release process .

classes/queue/JobRunner.php
touhidurabir: I agree this is better but in this case I would prefer the dumb code which is bit more readable and easier to understand .

classes/queue/JobRunner.php
touhidurabir: > Given the singleton management is done at the container level, we don't need to have another singleton management here.

thats true , but this done so that we also accidentally do a direct initialisation either

classes/queue/JobRunner.php
touhidurabir: this works with cross request locking mechanism so that another http request initialing the job runner while it's running form a previous request .

classes/core/PKPQueueProvider.php
touhidurabir: isn't that covered by the following logic
```php
$this->app['events']->listen(Looping::class, function (Looping $event) {

    if (!app()->runningInConsole() || app()->runningUnitTests()) {
        return true;
    }

    if (!$this->isMultiContextSite) {
        return true;
    }

    static $lockedContextId = null;

    // Peek at next job WITHOUT popping (no lock, no reserve, no attempt increment)
    $nextJob = DB::table('jobs')
        ->where('queue', $event->queue)
        ->where(fn(QueryBuilder $query) => $query->whereNull('reserved_at')
            ->orWhere('reserved_at', '<=', now()->subSeconds(90)))
        ->orderBy('id', 'asc')
        ->first();

    if (!$nextJob) {
        return true; // No jobs available, continue looping
    }

    $payload = json_decode($nextJob->payload, true);
    $nextContextId = $payload['context_id'] ?? null;

    // Non-context-aware jobs don't affect locking
    if ($nextContextId === null) {
        return true;
    }

    // First context-aware job sets the locked context
    if ($lockedContextId === null) {
        $lockedContextId = $nextContextId;
        return true;
    }

    // Different context detected - quit BEFORE popping
    if ($nextContextId && $nextContextId !== $lockedContextId) {
        app('queue.worker')->shouldQuit = true;
        return false; // Skip this iteration, pauseWorker() will check shouldQuit and exit
    }

    return true;
});
```

so this is how it's works in multi context env :-
we have a plugin P 
we have 2 context , A and B
Plugin P is active in context A
Plugin P is inactive in context B
The above mechanism peek ahead before picking a job . so 
1. job of context A is getting picked .
 - Plugin P is active in context A
 - looks for the plugins to load and registered
 - it's hooks will registered
 - next job is from context B
 - it's a different context aware jobs, so worker will quit and supervisor will restart the worker
2. job of context B is getting picked
 - Plugin P is inactive in context B
 - looks for the plugins to load and registered
 - will not get registered plugin P hooks as it's inactive
 
that is the design plan I went with . Basically on picking up a job of different context of that is just get executed, worker will quit right before the picking it up. Supervisor will restart the worker and before picking up the job of different context, it will load/register the plugins again .

classes/core/PKPApplication.php
touhidurabir: are we not supporting context less execution already ? thats why it can allow a null . and the clearing of cli context after each job processing or in case of failure ensure that the last context information not perserved for next job and it can be context less .

tools/jobs.php
touhidurabir: yeah, it's just that in cli, it flooding with deprecation warning and notices . But I will comment out it for now .

classes/core/PKPRouter.php
touhidurabir: @jonasraoni that is concerning . can you please provide a example flow where it will conflict ?

classes/core/PKPContainer.php
touhidurabir: it's same just moving it to the core app which provide better usage . I added that to the global functions at very beginning which was not a good idea .

classes/core/PKPQueueProvider.php
touhidurabir: I will double check it . I had some issue the postgres 11 .

classes/core/PKPQueueProvider.php
```diff
@@ -127,17 +190,36 @@ public function boot()
...
+                if (app()->runningInConsole()) {
+                    return;
+                }
+
+                // Not to run in unit test mode as part of the application lifecycle
+                if (app()->runningUnitTests()) {
+                    return;
+                }
+                
+                // error_log('Shutdown function started at: ' . microtime(true));
```
asmecher, line 204:
```suggestion
```

asmecher (commented):
@touhidurabir, I see there are quite a few unresolved comments from Jonas; are these reviewed and addressed (maybe in another branch)?

classes/core/PKPRouter.php
asmecher: @jonasraoni, a reminder about this question!

classes/queue/JobRunner.php
asmecher: @jonasraoni, a reminder about this question two!

classes/core/PKPQueueProvider.php
asmecher: @jonasraoni, a reminder about this question three!

classes/queue/JobRunner.php
touhidurabir: I think we can keep it to match the current pattern of the class

classes/core/PKPQueueProvider.php
touhidurabir: unfortunately IDE can not understand it properly and flag it as error.

classes/task/ProcessQueueJobs.php
touhidurabir: I would like the keep the dumb code here which is easy to read and understand .

classes/queue/DatabaseQueue.php
touhidurabir: True but this provide a more fluent API control of JobRunner .

classes/core/PKPRouter.php
jonasraoni: This is the type of thing that I meant, but AFAICS it doesn't exist anymore in +3.5:
https://github.com/pkp/ojs/blob/5019df5a6ecccf9058e86ff996439f268bf9555a/plugins/importexport/native/NativeImportExportPlugin.inc.php#L380-L392

classes/queue/JobRunner.php
jonasraoni: Long story short... Laravel isn't able to honor the `Cache::lock()` properly, so we'll need a package (or an in-house solution) that **uses the database**, probably the only resource we can currently rely on for a distributed lock.

> If needed more, the number of jobs, time limits, and memory limits can be configured in config for job runner .

I think it's ok to enable parallelism, it's up to the user to use it or not... But I'm not going to block the PR because of that, just wanted to share my view 😊

classes/core/PKPApplication.php
jonasraoni: Yeah, what I mean is that the current type should be `?Context` instead of `int|Context`, then the default `null` will be ok.

And I really think we should stop including getters/setters, unless they are required (e.g. extra operations), to have a better alignment with PHP 8.4.

classes/core/PKPContainer.php
jonasraoni: I mean... Is there any blocker for using the one-line code that I mentioned above or are all those `ifs` really required?

jonasraoni:
1. I think it's way too late to say it... But this PR is handling too many non-related, but important, issues.

   If there's still some energy, what about creating separated issues for them? I remember I flagged some on a Mattermost discussion (e.g. job execution blocking the response output), so I can try to help identifying them... They were simple to fix and should have been merged already in 3.4.

2. I barely can remember if I had previously [read the discussion/comment from](https://github.com/pkp/pkp-lib/pull/9463#issuecomment-1826177358) @asmecher (Nov/2023), so I don't know what was the final decision.

   Anyway, I fully agree with the point of avoiding adding new glue-code/classes to keep things working at the user-space... Because once we add something, it's tough to get it removed later (i.e. other developers will use the workaround).

   Therefore, I support dropping the fixes (e.g. `ContextAwareJob`) and let things break, at least for the `main` branch (need to evaluate how risky it currently is or will be for ORE) 😁

touhidurabir:
> 1. I think it's way too late to say it... But this PR is handling too many non-related, but important, issues.
>    If there's still some energy, what about creating separated issues for them? I remember I flagged some on a Mattermost discussion (e.g. job execution blocking the response output), so I can try to help identifying them... They were simple to fix and should have been merged already in 3.4.
> 2. I barely can remember if I had previously [read the discussion/comment from](https://github.com/pkp/pkp-lib/pull/9463#issuecomment-1826177358) @asmecher (Nov/2023), so I don't know what was the final decision.
>    Anyway, I fully agree with the point of avoiding adding new glue-code/classes to keep things working at the user-space... Because once we add something, it's tough to get it removed later (i.e. other developers will use the workaround).
>    Therefore, I support dropping the fixes (e.g. `ContextAwareJob`) and let things break, at least for the `main` branch (need to evaluate how risky it currently is or will be for ORE) 😁

@jonasraoni I agree (and also @asmecher does as per our last discussion) that this PR got too big and complex. 

We definitely need the enhancement of [Enhancement of Queue JobRunner](https://github.com/pkp/pkp-lib/issues/12532) that I just created for to be merged in 3.5 and main . 

So I have split the PR in 2 groups 
 1. Job runner enhancement with buffer flush and cross request locking
 2. Worker mode handling enhancement with context aware jobs.
 
PRs for the No.1 is done as part of [Enhancement of Queue JobRunner](https://github.com/pkp/pkp-lib/issues/12532), it's same but specific changes ported from this PR . can you take a quick look as I need to merged this for 3.5/main asap .

Issue #9276, opened 2023-09-05
OJS/OMP/OPS 3.5.0 Change Orientation


asmecher:
## For the Release Notebook ##

Items needing addition to the 3.5.0 Release Notebook should be noted here.
- [x] Invitations toolset https://github.com/pkp/pkp-lib/issues/9197
- [x] Slim API endpoints moved to Laravel https://github.com/pkp/pkp-lib/issues/7698#issue-1132905079
- [ ] Hook self-documentation https://github.com/pkp/pkp-lib/discussions/8089#discussioncomment-7139458
- [x] The Stringy dependency has been removed in favour of Laravel's `Str`; many PKPString wrapper functions have been removed
- [x] Add documentation note for those who are upgrading on how to add an app key (https://github.com/pkp/pkp-lib/issues/9895 -- see release notes)
- [x] Add documentation note about new API endpoint hook and how to add/modify API endpoint from plugin (https://github.com/pkp/pkp-lib/issues/9434)
- ...`_codesafe` [string wrapper functions removed](https://github.com/pkp/pkp-lib/issues/10164) (no longer necessary due to PHP8.2 baseline)
- Previously deprecated behaviours removed:
  - [x] `.inc.php` file suffixes are no longer supported
  - [x] Use of `HANDLER_CLASS` no longer supported
  - [x] All calls to `Router::url` or `Dispatcher::url` should use `array`s for the `$path` parameter -- scalars no longer supported
  - [x] Non-namespaced plugins are no longer supported.
- [x] The `fatalError` error handling function has been removed; use exceptions instead.
- [x] Email logging now uses an Eloquent model in place of a DAO (https://github.com/pkp/pkp-lib/issues/10129).
- [x] Notes (https://github.com/pkp/pkp-lib/issues/10188) and Queries (https://github.com/pkp/pkp-lib/issues/10189) now use Eloquent models in place of a DAO.
- [x] UserGroup now uses an Eloquent model in place of a DAO (https://github.com/pkp/pkp-lib/issues/10506).
- [x] Add documentation about new scheduler tool with new configs and deprecation of XML based tasks along with deprecation of Acron Plugin (https://github.com/pkp/pkp-lib/issues/9678) [comment](https://github.com/pkp/pkp-lib/issues/9678#issue-2110219392)
- [x] Add documentation about deprecation of `Services` and `PKPServices` classes and moving the service classes resolving via laravel container (https://github.com/pkp/pkp-lib/issues/9913)
[...]

PR #9469, opened 2023-10-30
pkp/pkp-lib#8887 REST API for dashboard views

Vitaliy-1:


classes/submission/reviewAssignment/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+    {
+        $object = $this->dao->newDataObject();
+        if (!empty($params)) {
+            $object->setAllData($params);
+        }
+        return $object;
+    }
+
+    /** @copydoc DAO::get() */
+    public function get(int $id): ?ReviewAssignment
```
asmecher, line 61: Please add a second (optional) `$submisisonId` parameter; the `SchemaDAO::get()` function supports the concept of a parent ID, and this can be used to ensure that calling code verifies that the ID belongs to the expected submission. (I'll tag a couple of examples with a link to this comment in a sec.)

classes/submission/reviewAssignment/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+        // Check for input from disallowed locales
+        ValidatorFactory::allowedLocales($validator, $this->schemaService->getMultilingualProps($this->dao->schema), $allowedLocales);
+
+        $errors = [];
+
+        if ($validator->fails()) {
+            $errors = $this->schemaService->formatValidationErrors($validator->errors());
+        }
+
+        Hook::call('Category::validate', [&$errors, $object, $props, $allowedLocales, $primaryLocale]);
```
asmecher, line 157: Wrong hook name.

asmecher: Sorry, (hopefully) one more request related to hooks :) Can you document the hooks that got removed in the issue, along with anything else that should go into the Release Notebook, and then drop a link to that comment into https://github.com/pkp/pkp-lib/issues/9276 where we track these things?

asmecher: One more :) When this code gets linted, you should see the pre-commit hook come in to add hook self-documentation in the function self-docs for the new hooks. If that doesn't happen, can you check whether your pre-commit hooks are running as expected?

classes/submission/reviewAssignment/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+            ReviewAssignment::SUBMISSION_REVIEW_METHOD_ANONYMOUS => 'editor.submissionReview.anonymous',
+            ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN => 'editor.submissionReview.open',
+        ];
+    }
+
+    /**
+     * Update the status of the review round an assignment is attached to. This
+     * should be fired whenever a reviewer assignment is modified.
+     */
+    protected function updateReviewRoundStatus(ReviewAssignment $reviewAssignment): bool
```
asmecher, line 239: I suspect functions like this would be better placed in the DAO, but since it's protected, no biggie if we decide to rearrange things later.

classes/submission/reviewAssignment/Collector.php
```diff
@@ -0,0 +1,298 @@
...
+    public function filterByIsOverdue(?bool $isOverdue): self
+    {
+        $this->isOverdue = $isOverdue;
+        return $this;
+    }
+
+    /**
+     * Filter by review round ids
+     */
+    public function filterByReviewRoundIds(?array $reviewRoundIds): self
```
asmecher, line 124: Better to type-hint a `static` return type, rather than `self`, I think: https://wiki.php.net/rfc/static_return_type

classes/submission/reviewAssignment/Collector.php
```diff
@@ -0,0 +1,298 @@
+<?php
+/**
+ * @file classes/submission/reviewAssignment/Collector.php
```
asmecher, line 3: This is very nice and clean and demonstrates very well the latest features we want to add throughout the codebase -- thanks!

asmecher (changes requested):
I just made a few comments, @Vitaliy-1; it's *excellent* to start to see this stuff coming into the newer coding conventions, and will lead to many downstream improvements!

PR #9482, opened 2023-11-02
Fix issue 5780 for stable-3_4_0

wopsononock:
Allows user to replace files uploaded to Publisher Library.  #5780

### DIFF
--- a/classes/components/listPanels/HighlightsListPanel.php
+++ b/classes/components/listPanels/HighlightsListPanel.php
@@ -38,7 +38,7 @@ class HighlightsListPanel extends ListPanel
             $config,
             [
                 'apiUrl' => $this->apiUrl,
-                'form' => $this->form->getConfig(),
+                'form' => $this->form?->getConfig(),
                 'getParams' => $this->getParams,
                 'i18nAdd' => __('manager.highlights.add'),
                 'i18nConfirmDelete' => __('manager.highlights.confirmDelete'),
--- a/controllers/grid/settings/library/form/EditLibraryFileForm.php
+++ b/controllers/grid/settings/library/form/EditLibraryFileForm.php
@@ -16,6 +16,8 @@
 
 namespace PKP\controllers\grid\settings\library\form;
 
+use APP\core\Application;
+use APP\file\LibraryFileManager;
 use PKP\context\LibraryFile;
 use PKP\context\LibraryFileDAO;
 use PKP\controllers\grid\files\form\LibraryFileForm;
--- a/pages/admin/AdminHandler.php
+++ b/pages/admin/AdminHandler.php
@@ -706,6 +706,14 @@ class AdminHandler extends Handler
      */
     protected function getHighlightsListPanel(): HighlightsListPanel
     {
+        if (!Config::getVar('features', 'highlights')) {
+            return new HighlightsListPanel(
+                'highlights',
+                '',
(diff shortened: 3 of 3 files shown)


classes/file/PKPLibraryFileManager.php
```diff
@@ -112,16 +112,49 @@ public function &copyFromTemporaryFile(&$temporaryFile, $libraryFileType)
...
+    }
+
+    /**
+     * Routine to replace a library file from a temporary file.
+     * @param $temporaryFile object
+     * @param $libraryFileType int LIBRARY_FILE_TYPE_...
+     * @param $libraryFile LibraryFile 
+     * @return LibraryFile|false the updated LibraryFile, or false on error
+     */
+    function &replaceFromTemporaryFile(&$temporaryFile, $libraryFileType, $libraryFile) {
```
asmecher, line 130: You're probably just following existing (very old) patterns, but the return reference here isn't necessary and should be removed.

asmecher: Also, not crucial, but for `stable-3_4_0` and newer, our preference is to use PHP8.0-compatible type hints for new code. For `main` it'll be PHP8.1-compatible. With that in place, `@param` tags are encouraged only when they add something more than the type hints do.

classes/file/PKPLibraryFileManager.php
```diff
@@ -112,16 +112,49 @@ public function &copyFromTemporaryFile(&$temporaryFile, $libraryFileType)
...
+     * @return LibraryFile the updated LibraryFile
+     */
+    function &assignFromTemporaryFile(&$temporaryFile, $libraryFileType, $libraryFile) {
         $libraryFile->setDateUploaded($temporaryFile->getDateUploaded());
         $libraryFile->setDateModified($temporaryFile->getDateUploaded());
         $libraryFile->setFileType($temporaryFile->getFileType());
         $libraryFile->setFileSize($temporaryFile->getFileSize());
         $libraryFile->setServerFileName($this->generateFileName($libraryFileType, $temporaryFile->getOriginalFileName()));
         $libraryFile->setOriginalFileName($temporaryFile->getOriginalFileName());
-        if (!$this->copyFile($temporaryFile->getFilePath(), $this->getBasePath() . $libraryFile->getServerFileName())) {
```
asmecher, line 121: Could you write a little bit about this change? At a glance, I think it might have unintended consequences.

asmecher (changes requested):
Thanks, just a few comments/questions!

classes/file/PKPLibraryFileManager.php
ctgraham: Previously the `copyFromTemporaryFile()` method handled both the metadata assignment and the file copy operation.  Breaking this into `copyFromTemporaryFile()` and `replaceFromTemporaryFile()`, we move the metadata operations into their own method.  It didn't seem appropriate for the `assignFromTemporaryFile()` to perform the copy, so each of `copyFromTemporaryFile()` (line 116) and `replaceFromTemporaryFile()` (line 134) to this themselves.

asmecher: That makes sense, thanks!

asmecher:
Replaced by https://github.com/pkp/pkp-lib/pull/9522.

Issue #9494, opened 2023-11-09
Add API to register vue.js components from plugins in 3.5 (vue3)


jardakotesovec:
### Issue

In Vue2 there was one global Vue instance (available via pkp.Vue) that allowed to globally register components and that was possible to use also from plugins.

In Vue3 there [is no global Vue instance](https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp), where would be possible to globally register components.New vue instance is created everytime it needs to be mounted (can be multiple times on one page until we have fully vue.js app) and therefore also the global components are registered for each instance. 

### Solution
Create new API available to the plugins
```
pkp.registry.registerComponent('componentName', component);
pkp.registry.getComponent('componentName);
```
Additionally expose vue module (to be able to use advance functions from vue, including composition api) via 
```
pkp.modules.vue
```

jardakotesovec:
pkp-lib: https://github.com/pkp/pkp-lib/pull/9496
ops (just to run tests): https://github.com/pkp/ops/pull/593

@touhidurabir Would you mind do the code review? After this is merged you would use pkp.registry.registerComponent instead of pkp.Vue.component.

@jonasraoni Could you please also have a quick look? Thanks!

jardakotesovec:
Merged

PR #9496, opened 2023-11-09
pkp/pkp-lib#9494

jardakotesovec:
- Add API for component registry from plugins
- Expose vue module globally

### DIFF
--- a/js/load.js
+++ b/js/load.js
@@ -10,8 +10,9 @@
 
 // Vue lib and custom mixins
 import {createApp} from 'vue';
-import {createPinia} from 'pinia'
+import * as vue from 'vue';
 
+import {createPinia} from 'pinia';
 import GlobalMixins from '@/mixins/global.js';
 import VueAnnouncer from '@vue-a11y/announcer';
 import FloatingVue from 'floating-vue';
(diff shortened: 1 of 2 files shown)


js/load.js
```diff
@@ -10,8 +10,9 @@
 
 // Vue lib and custom mixins
 import {createApp} from 'vue';
-import {createPinia} from 'pinia'
+import * as vue from 'vue';
```
jonasraoni, line 13: I think you can join both lines into a single import, but not important :)

js/load.js
```diff
@@ -99,18 +100,116 @@ import ListPanel from '@/components/ListPanel/ListPanel.vue';
 import VueRegistry from './classes/VueRegistry.js';
 import i18nPlugin from '@/piniaPlugins/i18n.plugin.js';
 
+// Register global components
+VueRegistry.registerComponent('Badge', Badge);
```
jonasraoni, line 104: If all components are properly named, I think we could extract the name dynamically from them (e.g. `Vue.component(component.name, component);`), and move these registrations into a loop.

I don't know what's the reasoning for registering `Spinner` and `PkpSpinner` (I imagine that it started with `Button`, then conflicted with real HTML elements, and was moved to `PkpButton`). Perhaps it's a good time to standardize the naming, given that probably almost nobody extended the Vue components 🤔

js/load.js
```diff
@@ -130,98 +229,22 @@ function pkpCreateVueApp(createAppArgs) {
...
-
-	// Required by the URN plugin, to be migrated at some point to pkp prefix
-	vueApp.component('field-text', FieldText);
-	vueApp.component('field-pub-id', FieldPubId);
-
-	// Register ListPanel
-	vueApp.component('PkpListPanel', ListPanel);
+	// register all global components
+	const allGlobalComponents = VueRegistry.getAllComponents();
+	Object.keys(allGlobalComponents).forEach((componentName) => {
```
jonasraoni, line 234: I think you can replace by `Object.entries()`, then use `([name, component]) => ...`.

jonasraoni (approved):

js/load.js
jardakotesovec: Naming was bit mixed, likely due some name conflicts with native elements.  So when I was exposing bigger part of ui-library to be available for plugins for 3.4, I kept the existing ones to avoid any surprises as 3.4 was already released and basically just exposed everything with pkp prefix.  But its good point for 3.5 that might be good opportunity to get rid of the ones without prefix.  Will consider it in separate PR.

Getting name from the component itself is also good suggestion, will keep it in mind when I review this logic. I need to consider whether to possibly expose all components globally, so they can be easily replaced from plugin via registry, that could be quite powerful way if someone needs some more significant customisation.

PR #9498, opened 2023-11-09
getTemporaryFilesByUserId function added to TemporaryFileDao

MrRob100:


classes/file/TemporaryFileDAO.inc.php
```diff
@@ -34,6 +34,18 @@ function getTemporaryFile($fileId, $userId) {
 		$row = (array) $result->current();
 		return $row?$this->_returnTemporaryFileFromRow($row):null;
 	}
+	
+	/**
+	 * Retrieve temporary files by user Id
+	 * @param $userId int
+	 * @return Generator
+	 */
+	public function getTemporaryFilesByUserId($userId) {
```
asmecher, line 43: Could you modify this function to return a `DAOResultFactory` invoking `_returnTemporaryFileFromRow`? It should be...
```php
$result = $this->retrieve(
	'SELECT * FROM temporary_files WHERE user_id = ? ORDER BY date_uploaded DESC LIMIT 1',
	[(int) $userId]
);
return new DAOResultFactory($result, $this, '_returnTemporaryFileFromRow');
```
That way this function takes care of instantiating the `TemporaryFile` objects, and calling code doesn't need to know about the raw database columns.

classes/file/TemporaryFileDAO.inc.php
```diff
@@ -34,6 +34,18 @@ function getTemporaryFile($fileId, $userId) {
...
+		return $this->retrieve(
+			'SELECT * FROM temporary_files WHERE user_id = ? ORDER BY date_uploaded DESC LIMIT 1',
```
asmecher, line 45: The `LIMIT 1` probably shouldn't be here!

classes/file/TemporaryFileDAO.inc.php
MrRob100: @asmecher Okay was there a way of doing it not using ORDER BY in the DAO method?

asmecher: I'm afraid I'm not following -- can you write a little more?

MrRob100: Your suggestion 
`$result = $this->retrieve(
	'SELECT * FROM temporary_files WHERE user_id = ?',
	[(int) $userId]
);
return new DAOResultFactory($result, $this, '_returnTemporaryFileFromRow');`

didn't have the `ORDER BY` in the statement which I have added, I wondered if you had removed it for a reason

`		$result = $this->retrieve(
			'SELECT * FROM temporary_files WHERE user_id = ? ORDER BY date_uploaded DESC',
			[(int) $userId]
		);
		return new DAOResultFactory($result, $this, '_returnTemporaryFileFromRow');`

asmecher: No objection to having the `ORDER BY` coded into the query!

Issue #9500, opened 2023-11-10
Hook `Template::SubmissionWizard::Section::Review ` not inside `submissionWizard__reviewPanel`


henriquejsfj:
**Describe the bug**
The Hook `Template::SubmissionWizard::Section::Review` called on the `'submission/wizard.tpl'` is supposed (I imagined) to permit devs to add a review to custom fields added in the steps. Although it is possible to add it, the visualization is not what is expected.

![image](https://github.com/pkp/pkp-lib/assets/16216517/aa34387c-7069-4b72-8cc9-2bb52e312185)

In the image above I added the check on `'editors'` step, however the fronted rendered it outside. This issue comes from the local where the hook is called. In the `'submission/wizard.tpl'` the reviewTemplate is rendered before closing the divs. 
```tpl
{foreach from=$reviewSteps item=$step}
    {if $step.reviewTemplate}
        {include file=$step.reviewTemplate}
    {/if}
    {call_hook name="Template::SubmissionWizard::Section::Review" submission=$submission step=$step.id}
{/foreach}
```

**To Reproduce**
I added this function to the Hook:
```php
function handleSubmissionReview($hookName, $args) {
    $step = $args[0]['step'];
    $templateMgr =& $args[1];
    $output = &$args[2];
    
    switch ($step) {
	    case 'editors':
		    $output .= $templateMgr->fetch($this->getTemplateResource('reviewEditors.tpl'));
		    break;
    }
    
    return false;
}
```
**What application are you using?**
 OJS version 3.4

**Additional information**
The `$output = &$args[2]` is coming with `null`value. I guess adding the result from reviewTemplate would be enough. Besides, it would be awesome if we have the new fields added to the form component generating it automatically.

henriquejsfj:
Sorry, no PR this time, I don´t know exactly how to solve it.

asmecher:
@henriquejsfj, if I understand right, basically you'd like to make a change to the markup but the hook you identified as the best candidate (`Template::SubmissionWizard::Section::Review`) is not at the right place in the markup for your needs. Do I have that right?

henriquejsfj:
@asmecher Yes! In the print I sent I would like to have my field inside the Editor compartment.

I understand that the hook is useful when someone make a new step.

Besides, I can't use the TemplateManager's hooks since they are included in the tpl...

jardakotesovec:
[...]

PR #9513, opened 2023-11-15
pkp/pkp-lib#9500 Adding template hooks to be able to extend review-de…

jardakotesovec:
…tails and review-editors on review page


henriquejsfj (commented):
Thanks, this PR solves my Issue #9500 

Wouldn't it be better to add those specific hooks on `review-contributors.tpl` and `review-files.tpl` to follow the pattern?

Please add to stable-3_4 too!

PR #9522, opened 2023-11-17
Publisher Library can swap files

### LINKED ISSUE #5780: Replace files in Publisher Library

wopsononock:
Fixes #5780

### DIFF
--- a/classes/file/PKPLibraryFileManager.php
+++ b/classes/file/PKPLibraryFileManager.php
@@ -18,6 +18,7 @@ namespace PKP\file;
 
 use PKP\context\LibraryFile;
 use PKP\context\LibraryFileDAO;
+use PKP\file\TemporaryFile;
 use PKP\core\PKPString;
 use PKP\db\DAORegistry;
 
@@ -125,7 +126,7 @@ class PKPLibraryFileManager extends PrivateFileManager
      * @param $libraryFileType int LIBRARY_FILE_TYPE_...
      * @return LibraryFile|false the updated LibraryFile, or false on error
      */
-    function replaceFromTemporaryFile(object &$temporaryFile, int $libraryFileType, LibraryFile $libraryFile) {
+    function replaceFromTemporaryFile(TemporaryFile $temporaryFile, int $libraryFileType, LibraryFile $libraryFile) {
         $originalServerFilename = $libraryFile->getServerFileName();
 
         $libraryFile = $this->assignFromTemporaryFile($temporaryFile, $libraryFileType, $libraryFile);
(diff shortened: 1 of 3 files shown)


wopsononock:
This PR supersedes #9482, which had problems with my commit name and email that I couldn't resolve without starting over. This addresses the code issues raised there and tests successfully with stable-3_4_0.

classes/file/PKPLibraryFileManager.php
```diff
@@ -112,16 +112,47 @@ public function &copyFromTemporaryFile(&$temporaryFile, $libraryFileType)
...
+
+        return $libraryFile;
+    }
+
+    /**
+     * Routine to replace a library file from a temporary file.
+     * @param $libraryFileType int LIBRARY_FILE_TYPE_...
+     * @return LibraryFile|false the updated LibraryFile, or false on error
+     */
+    function replaceFromTemporaryFile(object &$temporaryFile, int $libraryFileType, LibraryFile $libraryFile) {
```
asmecher, line 128: `object` should be `TemporaryFile` (and a `use` statement for that should be added at the top).

asmecher: Also, the reference isn't necessary here.

classes/file/PKPLibraryFileManager.php
```diff
@@ -112,16 +112,47 @@ public function &copyFromTemporaryFile(&$temporaryFile, $libraryFileType)
...
+
+        if ($originalServerFilename !== $libraryFile->getServerFileName()) {
+                unlink($this->getBasePath() . $originalServerFilename);
+        }
+        return $libraryFile;
+    }
+
+    /**
+     * Routine to assign metadata to a library file from a temporary file
+     * @param $temporaryFile object
```
asmecher, line 144: Should be `TemporaryFile`, not `object`

classes/file/PKPLibraryFileManager.php
```diff
@@ -112,16 +112,47 @@ public function &copyFromTemporaryFile(&$temporaryFile, $libraryFileType)
...
+     * @param $libraryFileType int LIBRARY_FILE_TYPE_...
+     * @param $libraryFile LibraryFile 
+     * @return LibraryFile the updated LibraryFile
+     */
+    function &assignFromTemporaryFile(&$temporaryFile, $libraryFileType, $libraryFile) {
```
asmecher, line 149: References are unnecessary here.

asmecher (changes requested):
Just a few small changes to request, thanks!

wopsononock:
@asmecher I made the changes you requested.

Issue #9518, opened 2023-11-16
[ OJS ] Error in I5716_EmailTemplateAssignments from 3.3.0.8 to 3.4.0-3


rubrodapa:
**Describe the bug**
Upgrading from 3.3.0.8 to 3.4.0-3 (also tested with 3.4.0-0) fails on APP\migration\upgrade\v3_4_0\InstallEmailTemplates.

ERROR: Upgrade failed: DB: Tried to install email template as an alternate to `DISCUSSION_NOTIFICATION_SUBMISSION`, but no default template exists with this key.

**To Reproduce**
Launch db check then upgrade

`php tools/upgrade.php check`

[Ok, no warn, no err]

`php tools/upgrade.php upgrade`

Results below

**What application are you using?**
ojs 3.3.0-8 attempt to upgrade 3.4.0-3

**Additional information**

Log before error

```
2023-11-16 12:58:15 [migration: APP\migration\upgrade\v3_4_0\I6782_MetricsIssue]
2023-11-16 12:58:18 [migration: APP\migration\upgrade\v3_4_0\I6782_MetricsSubmission]
2023-11-16 12:58:48 [migration: APP\migration\upgrade\v3_4_0\I6782_MetricsGeo]
2023-11-16 12:58:54 [migration: APP\migration\upgrade\v3_4_0\I6782_CleanOldMetrics]
2023-11-16 12:59:40 [migration: APP\migration\upgrade\v3_4_0\I6782_RemovePlugins]
2023-11-16 12:59:40 [migration: PKP\migration\upgrade\v3_4_0\I7286_BatchesMigration]
2023-11-16 12:59:40 [migration: APP\migration\upgrade\v3_4_0\I8151_ExtendSettingValues]
2023-11-16 12:59:42 [migration: PKP\migration\upgrade\v3_4_0\I8151_ExtendSettingValues]
2023-11-16 12:59:52 [migration: PKP\migration\upgrade\v3_4_0\I2890_EmailTemplatesVarcharLengthUpdate]
2023-11-16 12:59:53 [migration: PKP\migration\upgrade\v3_4_0\I2890_AddSetNullForOnDeleteToReviewRoundIdForeign]
2023-11-16 12:59:54 [migration: APP\migration\upgrade\v3_4_0\I7901_Duplicate_OAI_IDs]
2023-11-16 12:59:54 [migration: PKP\migration\upgrade\v3_4_0\I7366_UpdateUserAPIKeySettings]
[...]

PR #9531, opened 2023-11-23
pkp/pkp-lib#9518 Try to install missing template before throwing an error

Vitaliy-1:


classes/emailTemplate/DAO.php
```diff
@@ -379,7 +379,18 @@ public function installAlternateEmailTemplates(int $contextId, ?string $emailKey
...
             if (!$exists) {
-                throw new Exception('Tried to install email template as an alternate to `' . $alternateTo . '`, but no default template exists with this key.');
+                trigger_error(
+                    'Tried to install email template as an alternate to `' . $alternateTo . '`, but no default template exists with this key. Installing ' . $alternateTo . ' email template first',
+                    E_USER_WARNING
+                );
+                try {
+                    $siteDao = DAORegistry::getDAO('SiteDAO'); /** @var SiteDAO $siteDao */
+                    $site = $siteDao->getSite(); /** @var Site $site */
+                    $locales = $site->getInstalledLocales();
```
jonasraoni, line 389: Looks like these lines can be dropped, if you pass an empty array at the `$locales` argument, it will have the same effect.

classes/emailTemplate/DAO.php
```diff
@@ -379,7 +379,18 @@ public function installAlternateEmailTemplates(int $contextId, ?string $emailKey
...
+                    $this->installEmailTemplates(Repo::emailTemplate()->dao->getMainEmailTemplatesFilename(), $locales, $alternateTo);
+                } catch (Exception $e) {
+                    error_log($e->getMessage());
```
jonasraoni, line 392: If the previous behavior was right, regarding not executing the next instruction if the `$alternateTo` entry doesn't exist, then I think the try/catch should be removed, to allow the exception to leak.

jonasraoni (changes requested):

PR #9536, opened 2023-11-24
[PKP-LIB][main] #7505 Initial Commit Upload - Delete - Download JATS File

defstat:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1691,4 +1708,151 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+
+    /**
+     * Get JATS XML Files
+     */
+    public function getJats(Request $illuminateRequest): JsonResponse
+    {
+        $request = $this->getRequest();
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+        
+        $publication = Repo::publication()->get((int) $illuminateRequest->route('publicationId'));
```
asmecher, line 1720: I see you've got this checked manually below -- but if you use the 2nd parameter to `Repository::get` to specify the submission ID, then the auth policies for the submission will take care of the rest. That'll eliminate the need for extra checking, which is easily forgotten.

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+            $fileProps['jatsContent'] = $xmlString;
+        }
+
+        return $fileProps;
+    }
+
+    /**
+     * Returns the SubmissionFile, if any, that corresponds to the JATS contents of the given submission/publication
+     */
+    public function getJatsFiles(int $submissionId, int $publicationId): ?SubmissionFile
```
asmecher, line 882: Misnomer -- should be `getJatsFile`.

classes/components/listPanels/JatsListPanel.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * http://example.org/api/v1/submissions/1/publications/__publicationId__
+     */
+    protected function getPublicationUrlFormat(): string
+    {
+        return Application::get()->getRequest()->getDispatcher()->url(
+            Application::get()->getRequest(),
+            Application::ROUTE_API,
+            $this->context->getPath(),
+            'submissions/' . $this->submission->getId() . '/publications/__publicationId__/jats'
```
asmecher, line 95: Just for consistency's sake, use `{$publicationId}` as the replacement slug. (also at https://github.com/pkp/ui-library/pull/300/files#diff-564efc80c3d827c62c2b608403f1b0363b0a56c98aae4053ad745001ff1a7c70R182)

classes/db/DAOResultFactory.php
```diff
@@ -259,6 +259,22 @@ public function toAssociativeArray($idField = 'id')
...
+     * Convert this iterator to a collection.
+     *
+     * @return Collection<T>
+     */
+    public function toCollection(): Collection
+    {
+        $retCollection = collect();
+
+        while ($object = $this->next()) {
+            $retCollection->push($object);
```
asmecher, line 273: This will cause a big result set to be instantiated in memory all at once; I'd suggest using a generator instead. https://josephsilber.com/posts/2020/07/29/lazy-collections-in-laravel#lazy-collections-wrap-a-generator-function

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
             'username' => $user?->getUsername(),
         ];
     }
+
+    /**
+     * Add jatsContent for Submission files that correspont to
```
asmecher, line 864: *correspond

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+    /**
+     * Returns the SubmissionFile, if any, that corresponds to the JATS contents of the given submission/publication
+     */
+    public function getJatsFiles(int $submissionId, int $publicationId): ?SubmissionFile
+    {
+        $this->validateSubmissionAndPublication($submissionId, $publicationId);
+
+        $submissionFile = Repo::submissionFile()
+            ->getCollector()
+            ->filterBySubmissionIds([$this->submission->getId()])
```
asmecher, line 888: The `$publicationId` filter implicitly limits by submission ID as well. I'd suggest modifying this function to make `$publicationId` the first parameter, and making the `$submissionId` optional; when it's specified, then this filter is applied as well. That way it can be used optionally as an extra guardrail (and follows a familiar pattern we use elsewhere). And we can do away with extra functions like `validateSubmissionAndPublication`, which can quickly collect and be used inconsistently.

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+        
+        return $submissionFile;
+    }
+
+    /**
+     * Returns the name of the file that will contain the default JATS content 
+     */
+    public function getDefaultJatsFileName(int $submissionId, int $publicationId): string
+    {
+        return 'jats-' . $submissionId . '-' . $publicationId . '-' . (string) mt_rand(100000, 999999) . '.xml';
```
asmecher, line 902: This is just a download filename, right? Better to use a timestamp than a random number.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1691,4 +1708,151 @@ protected function getWriteDisabledErrors(string $schemaName, array $params): ar
...
+     */
+    protected function getJatsUIProps($request, $submissionId, $publicationId): array
+    {
+        $jatsFile = Repo::submissionFile()
+            ->getJatsFiles($submissionId, $publicationId);
+
+        $jatsFilesProp = [];
+        if (!isset($jatsFile)) {
+            $jatsFilesProp['defaultJatsContent'] = Repo::submissionFile()
+                ->createDefaultContentJatsFile($submissionId, $publicationId);
```
asmecher, line 1844: This assumes we'll be storing the "default" JATS as a file, right? I think that still needs removal.

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+    }
+
+    /**
+     * Creates the default JATS XML Content from the given submission/publication metadata
+     */
+    public function createDefaultContentJatsFile(int $submissionId, int $publicationId): string
+    {
+        $submission = Repo::submission()->get($submissionId);
```
asmecher, line 910: I think this will already be removed, but wherever possible it's a good idea to ensure that your IDs match appropriately where it's possible to specify several. That would mean either simplifying this down to only accept a `$publicationId`, or making sure that `$publicationId` and `$submissionId` are *both* passed in to the `Repo::publication()::get` function.

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+
+        $exportXml = $this->convertSubmissionToJatsXml($submission, $context, $section, $issue, $publication, Application::get()->getRequest());
+
+        return $exportXml;
+    }
+
+    /**
+     * Base function that will add a new JATS file
+     */
+    public function addJatsFile(
```
asmecher, line 928: A lot of this stuff (especially anything that includes the names of request parameters) should be moved to the handler class. To avoid having that grow too long, I'd suggest spinning the JATS stuff out into its own handler class.

classes/submissionFile/Repository.php
```diff
@@ -19,19 +19,31 @@
...
 use PKP\db\DAORegistry;
+use PKP\file\FileManager;
+use PKP\file\TemporaryFileManager;
 use PKP\log\event\SubmissionFileEventLogEntry;
 use PKP\log\SubmissionEmailLogDAO;
 use PKP\log\SubmissionEmailLogEntry;
 use PKP\mail\mailables\RevisedVersionNotify;
 use PKP\note\NoteDAO;
 use PKP\notification\PKPNotification;
+use PKP\oai\OAIRecord;
```
asmecher, line 46: Is this necessary?

asmecher (changes requested):
Hi @defstat! A handful of comments, mostly minor. Looking very good!

schemas/submissionFile.json
```diff
@@ -27,7 +27,7 @@
 			"apiSummary": true,
 			"description": "Used with `assocId` to associate this file with an object such as a galley. One of the following constants: `ASSOC_TYPE_SUBMISSION_FILE` (dependent files), `ASSOC_TYPE_REVIEW_ASSIGNMENT` (files uploaded by a reviewer), `ASSOC_TYPE_NOTE` (files uploaded with a discussion), `ASSOC_TYPE_REPRESENTATION` (files uploaded to a galley or publication format), `ASSOC_TYPE_REVIEW_ROUND` (review files and revisions for a particular review round).",
 			"validation": [
-				"in:515,517,520,521,523"
+				"in:515,517,520,521,523,1048588"
```
asmecher, line 30: What does this constant correspond to?

classes/submissionFile/Repository.php
```diff
@@ -843,4 +859,262 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+        $submissionFile = Repo::submissionFile()
+            ->get($submissionFileId);
+
+        return $submissionFile;
+    }
+
+    /**
+     * Function to get the Context's genres
+     */
+    public function getGenres(int $contextId) : Collection
```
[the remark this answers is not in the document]
defstat: Removed.

classes/submissionFile/Repository.php
defstat: Changed.

classes/db/DAOResultFactory.php
defstat: Maybe something like that 

```
public function toLazyCollection(): LazyCollection
{
    return LazyCollection::make(function () {
        while ($object = $this->next()) {
            yield $object;
        }
    });
}
```
In any case I removed the function all together because it is not referenced anywhere in the code anymore.

classes/submissionFile/Repository.php
defstat: Sure! Fixed.

classes/submissionFile/Repository.php
defstat: Removed

schemas/submissionFile.json
defstat: It's the Publication type DB representation (`Application::ASSOC_TYPE_PUBLICATION`)

api/v1/submissions/PKPSubmissionController.php
defstat: I used the same function name, which, now, creates the JATS content using the `JatsTemplate` code. I changed the name of the function to `createDefaultJatsContent`

classes/submissionFile/Repository.php
defstat: Changed it to `return 'jats-' . $submissionId . '-' . $publicationId . '-' . date('Ymd-His') . '.xml';`

classes/submissionFile/Repository.php
defstat: I removed the `validateSubmissionAndPublication` function.

classes/submissionFile/Repository.php
defstat: passed both `publicationId` and `$submissionId` in the getter.

api/v1/submissions/PKPSubmissionController.php
defstat: I passed the `$submissionId` as a second parameter to the function.

api/v1/jats/PKPJatsController.php
```diff
@@ -0,0 +1,213 @@
+<?php
+
+/**
+ * @file api/v1/jats/PKPJatsController.php
```
asmecher, line 4: If this is going to take care of the `submissions/{submissionId}/publications/{publicationId}/jats` endpoint, then it should be in `api/v1/submissions` or `api/v1/submissions/jats`. I think you can direct execution there by adding the new handler to `api/v1/submissions/index.php`; then you wouldn't need to e.g. list the new operations in the `PKPSubmissionController`.

classes/submissionFile/Repository.php
```diff
@@ -843,4 +849,59 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+            ->add($newSubmissionFile);
+
+        $submissionFile = Repo::submissionFile()
+            ->get($submissionFileId);
+
+        return $submissionFile;
+    }
+
+    /**
+     * returns jatsContent for Submission files that correspond to the content of the file
```
asmecher, line 897: I think this function can be removed. It shouldn't be necessary to use the `->summarize` to get the pathname (that data should be available directly from `$submissionFile`); the XML-specific parts of this (`jatsContent`, `jatsFileName`, `xmlString`) aren't actually specialized for JATS, and none of the other parts of the codebase (e.g. XML export) that need to read files from the disk use this. Plus I think you could get rid of the use of `$genres` entirely (also from calling code) if you didn't use `->summarize`.

classes/jats/Repository.php
```diff
@@ -0,0 +1,221 @@
+<?php
+/**
+ * @file classes/jats/Repository.php
```
asmecher, line 3: I'm ambivalent about the use of a repository class for this. Actually, almost everything here is closely tied to the handling code, and will only be called once from the `PKPJatsController` -- so I'd be happy if it these functions were moved into `PKPJatsController` and this controller was removed. Then `classes/jats/JatsFile.php` could be moved into `lib/pkp/classes/submissionFile` alongside the parent class, which makes more sense to me.

But if you're not sure that's the best approach, I'd be OK to merge this as is too.

asmecher (changes requested):
Looks very close, @defstat, thanks! I just made a few rearrangement suggestions.

classes/jats/Repository.php
defstat: I had the relationship with the `SubmissionFile` and the `SubmissionFile/Repository` dropped for `JatsFile` object, and that's why I took both files out of the SubmissionFile folder and in it's own `jats` folder. This can be used as the "code home" for existing business logic and future possible merging of the code from, let's say, `jatsTemplate` plugin for generating the "default" Jats content.

classes/submissionFile/Repository.php
defstat: You are right that let's say `jatsContent` is not specific to the `JatsFile` except, in my opinion, when this content is auto generated by the submission metadata. When a file is uploaded, then the content is the submission file content. The same goes for the other `JatsFile` properties as well. 

After you pointed that out, that the `JatsFile` is not always a `SubmissionFile`, which makes perfect sense, I thought that keeping the `SubmissionFile` parent relationship on the `JatsFile` and the `SubmissionFile/Repository` parent relationship on the `Jats/Repository` could be generally misleading. To remedy that, I choose to drop the `JatsFile` relationships and keep the `JatsFile` the reference class for Jats actions/UI interactions. 

In that sense, in order to use this object to interact with `vue.js` I had to `summarize` it to the "appropriate" set of array properties, for the UI to work. I thought using a separate Schema for that, but I thought it does not fit here, because the `JatsFile` is not a "legit" entity but more of a conceptual placeholder for keeping the necessary exchange data together in one place.

api/v1/jats/PKPJatsController.php
defstat: I think I did add the reference to the `PKPJatsController` API Controller in the `submissions/index.php`, and removed all references/endpoints regarding JATS from the `PKPSubmissionController`. 

Are you referring to something else here?

asmecher: The operations in `PKPJatsController` [are still listed](https://github.com/pkp/pkp-lib/pull/9536/files#diff-3673647c7b54136c9bd00654b29a449ee67288a533f1f08e14452b4777ca01edR106) in `PKPSubmissionController`; can that be moved to `PKPJatsController`?

classes/submissionFile/Repository.php
defstat: @asmecher you are right, I don't have to summarize the `$submissionFile` there in order to get the `path` that I need. 

Changed.

api/v1/jats/PKPJatsController.php
defstat: Removed from `PKPSubmissionController`. Added the `$this->addPolicy(new PublicationWritePolicy($request, $args, $roleAssignments));` in the `PKPJatsController` instead of the `PublicationAccessPolicy`

api/v1/jats/PKPJatsController.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
+
+        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+
+        $this->addPolicy(new PublicationWritePolicy($request, $args, $roleAssignments));
+
+        if ($actionName === 'add') {
+            $params = $illuminateRequest->input();
+            $fileStage = isset($params['fileStage']) ? (int) $params['fileStage'] : 21;
```
asmecher, line 100: File stage shouldn't be hard-coded (`21`).

api/v1/jats/PKPJatsController.php
```diff
@@ -0,0 +1,214 @@
...
+            self::roleAuthorizer([
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SITE_ADMIN,
+                Role::ROLE_ID_SUB_EDITOR,
+                Role::ROLE_ID_ASSISTANT,
+                Role::ROLE_ID_AUTHOR,
+            ]),
+        ])->group(function () {
+
+            Route::get('', $this->get(...))
```
asmecher, line 72: Probably my ignorance about Laravel's routes, but should these routes actually be empty strings?

classes/submissionFile/Repository.php
```diff
@@ -843,4 +847,58 @@ protected function getSubmissionFileLogData(SubmissionFile $submissionFile): arr
...
+     */
+    public function getSubmissionFileContent(SubmissionFile $submissionFile): string
+    {
+        $jatsFileName = Config::getVar('files', 'files_dir') . '/' . $submissionFile->getData('path') .'';
```
asmecher, line 899: Change `$jatsFileName` to just `$fileName`, as this can be used for non-JATS files; also remove the empty string concatenation at the end.

api/v1/jats/PKPJatsController.php
defstat: Changed to `SubmissionFile::SUBMISSION_FILE_JATS`

classes/submissionFile/Repository.php
defstat: Sure! Fixed.

jardakotesovec:
@asmecher Are we confident that we want to have dependency from pkp-lib to jats plugin? Is not that opportunity to actually include jats plugin as part of the core if core functionality is dependent on it?

asmecher:
> @asmecher Are we confident that we want to have dependency from pkp-lib to jats plugin?

Yes, by the time this is released I think we'll have moved the JATS Template plugin into core.

> I am getting following error, when loading workflow page for any submission...

@defstat, could you look at that?

jardakotesovec:
@asmecher @defstat Just to clarify. Cause of this problem is that I did not have JATS template plugin installed.  Once I did clone that it works fine..  but it means that it would break workflow page for anyone who does not install JATS template plugin after merging.

defstat:
> I am getting following error, when loading workflow page for any submission:

@jardakotesovec I suppose that you don't have the `jatsTemplate` plugin in your installation.

I could perhaps catch any JATS creation error and display something different/special in the case of that case - something like "Make sure the JatsTemplate plugin is installed" or something.

js/load.js
```diff
@@ -198,6 +201,9 @@ VueRegistry.registerComponent('field-pub-id', FieldPubId);
 // Register ListPanel
 VueRegistry.registerComponent('PkpListPanel', ListPanel);
 
+// Register PublicationSectionJats
+VueRegistry.registerComponent('PkpPublicationJats', PublicationSectionJats);
+
```
jardakotesovec, line 206: Lets not register this component here. Here are only general components, so its possible to use them from plugins. Its perfectly sufficients that its 'components' field in workflow.vue

asmecher:
Thanks, @defstat, I'm OK with the use of custom exceptions in this code. (I'd like to brainstorm a little with the team before we go too wild creating custom exception classes, because they are really going to multiply if the pattern is used in general! But it's OK here.)

js/load.js
defstat: I think I found it - I can put `publication-section-jats` because I have named the component `PublicationSectionJats` in the `WorkflowPage.vue`

Fixed.

Issue #9455, opened 2023-10-23
Multiple Reviewer automatic reminder in OJS


pilasou:
As an editor, I want to be able to configure multiples occurrences of reminder to be sent to reviewers. 

Those configurations are added to the Review setup page (Settings > Workflow > Review) as shown below: 

![image](https://github.com/pkp/pkp-lib/assets/19291880/90cfed2e-f8f8-4e20-bd64-598805ae8a23)

The editor can configure the number of occurrences for the reminders both for the invitation and the evaluation report. 

No specific emails templates will be added. Already existing one are used (Review Reminder (Automated))

asmecher:
@nibou230, thanks, I added a couple of comments to the PRs above! They are very minor -- this looks good, thanks.

nibou230:
@asmecher thanks, I did the required modifications, hope it's good enough. Let me know if something is missing!

pilasou:
Hi, This issue would be mostly resolved by [5885](https://github.com/pkp/pkp-lib/issues/5885), if a configurable additional reminder is implemented (https://github.com/pkp/pkp-lib/issues/5885#issuecomment-1949317971).

lpanebr:
> Hi, This issue would be mostly resolved by [5885](https://github.com/pkp/pkp-lib/issues/5885), if a configurable additional reminder is implemented ([#5885 (comment)](https://github.com/pkp/pkp-lib/issues/5885#issuecomment-1949317971)).

Actually sending multiple reminders is a different and a very useful one.

**Below I propose an interface that requests only 2 inputs, instead of 4 and allows for better flexibility.**

![image](https://github.com/user-attachments/assets/dbb391b6-503f-401e-baa2-395937c12ac5)


**Response Reminder**
Comma separated list of reminders to be sent after the initial invite (in days). Sent only if the reviewer has not answered yet.
For example, setting 5, 7, 10 will send 3 reminders.

**Review Reminder**
Comma separated list of reminders to be sent before/after the review due date (in days). Sent only if the reviewer has not submitted a recommendation yet.
For example, setting -7, -2, 2, 7 will send 4 reminders (2 before, and 2 after the due date).

PR #9548, opened 2023-11-30
pkp/pkp-lib#9455 Add support for multiple occurrences on review reminders

nibou230:
pkp/ojs#4101


xml/schema/reviews.xml
```diff
@@ -53,7 +53,16 @@
 		</field>
 		<field name="reviewer_file_id" type="I8"/>
 		<field name="date_rated" type="T"/>
-		<field name="date_reminded" type="T"/>
+        <field name="date_invite_reminded" type="T"/>
```
asmecher, line 56: The old XML schema is only used for legacy parts of the upgrade process and shouldn't need to be modified. For anything post-3.3.0, the migrations classes take care of it.

Issue #9456, opened 2023-10-23
Allowing user private editorial notes in OJS


pilasou:
As an editor of a journal in a multi-journals OJS instance, I want to be able to add editorial notes on a user only available to my journal staff.  

OJS only offer one type of note (Editorial notes) that are shared among all journals on the instance.

We improve this feature by adding a new type of notes “Private editorial notes” for notes specific to a journal and note share among others. Exiting Editorial note are renamed “Public editorial notes”, as shown below: 

![image](https://github.com/pkp/pkp-lib/assets/19291880/d47fb4d6-1b81-43e7-b984-4b38a621d1c6)

This editorial note appears in the reviewer list when inviting a reviewer (read only) and once a reviewer invited, in the Editorial notes menu (read and edit). The new private editorial note appears the same way. 

Both private and public editorial notes can be added in 2 ways:  

1. If the user is only in one journal: To add user notes, edit a user, in the window a new text box will be displayed as illustrated below: 
2. If the user is in more than one journal: the notes can be edited once a reviewer as been added to a submission.

kristinmhoffmann:
Adding a comment in support of this. We host a multi-journal instance of OJS and several of our journals have reviewers in common. Editors assume that they see information for their journal only.

nibou230:
Hello @asmecher and @jonasraoni, you can find the PR links below:

https://github.com/pkp/pkp-lib/pull/9549
https://github.com/pkp/ui-library/pull/303
https://github.com/pkp/ojs/pull/4110

asmecher:
@Devika008, a question for you about this. Currently we have an "Editorial Notes" field associated with each user, which any editor/manager can use to keep notes. The issue is that the field is shared between journals, which can be unexpected.

The proposed feature adds a second field, "Private Editorial Notes", which is specific to the journal. The old "Editorial Notes" field is renamed to "Public Editorial Notes".

I'm not sure "Public" is the right name for the old field -- these are still private to editors -- and uncertain whether adding a second field is the right approach. This might be one of those cases where some installations expect the field to be shared (e.g. small publishers with shared editorial teams between journals) and others don't (commercial service providers sharing an install for several journals). I think there are similar issues that hosting has raise (@pmangahis / @mfelczak) and it might be worth considering them as a group rather than one at a time.

nibou230:
Hi @asmecher and @Devika008,

> I'm not sure "Public" is the right name for the old field -- these are still private to editors

[...]

PR #9549, opened 2023-11-30
pkp/pkp-lib#9456 Add user private notes

nibou230:
https://github.com/pkp/ojs/pull/4110


classes/userPrivateNote/UserPrivateNote.php
```diff
@@ -0,0 +1,89 @@
...
+use PKP\core\DataObject;
+
+class UserPrivateNote extends DataObject
+{
+	/**
+	 * Get private note context ID.
+     *
+	 * @return int
+	 */
+    public function getContextId(): int
```
asmecher, line 30: Indentation is a bit garbled here; there should be a git pre-commit hook that applies automatic formatting. Can you check if that's working?

classes/userPrivateNote/UserPrivateNote.php
```diff
@@ -0,0 +1,89 @@
...
+     *
+	 * @param $note string
+	 */
+    public function setNote(string $note): void
+    {
+		$this->setData('note', $note);
+	}
+}
+
+if (!PKP_STRICT_MODE) {
```
asmecher, line 87: For new code, it's best not to add `class_alias`es like this unless for some reason it's necessary (I think that's only for plugin classes at the moment). We'll remove these at some point in the future; they're only for backward compatibility.

classes/migration/install/CommonMigration.php
```diff
@@ -15,6 +15,7 @@
 namespace PKP\migration\install;
 
 use APP\core\Application;
+use APP\facades\Repo;
```
asmecher, line 18: I think this is probably leftover and not necessary.

classes/userPrivateNote/UserPrivateNote.php
```diff
@@ -0,0 +1,89 @@
...
+ * @see DAO
+ *
+ * @brief Basic class describing user private note existing in the system.
+ */
+
+namespace PKP\userPrivateNote;
+
+use PKP\core\DataObject;
+
+class UserPrivateNote extends DataObject
```
asmecher, line 23: No need to change this unless you're motivated to -- but in general we're moving towards Eloquent models rather than `DataObject`s. At some point in the distant future we'll deprecate `DataObject` entirely.

classes/userPrivateNote/Repository.php
```diff
@@ -0,0 +1,128 @@
...
+
+        Hook::call('UserPrivateNote::delete', [$userPrivateNote]);
+    }
+
+    /**
+     * Get the user private note for the specified context.
+     *
+     * This returns the first user private note, as the "user ID/context ID" key should be unique.
+     */
+    public function getFirstUserPrivateNote(int $userId, int $contextId): ?UserPrivateNote
```
asmecher, line 118: Since there's a unique index on this (i.e. it's impossible to have more than one private note per journal per user), I'd suggest dropping `First` in the function name.

classes/userPrivateNote/DAO.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+/**
+ * @file classes/userPrivateNote/DAO.php
```
asmecher, line 4: Rather than a bespoke table for this feature (with all the attendant management that'll require), another option would've been to use `NoteDAO` with an `assoc_type` of  `ASSOC_TYPE_CONTEXT`. Notes are currently used for a few things, and the assoc_type/assoc_id pairing is similar to a Laravel "morphs" relationship.

asmecher (commented):
A few general comments! Before you start reading these over, I'll start a convo with Devika about the feature; start over there first. Thanks, this is a very clean contribution (as the others have been too)!

classes/userPrivateNote/DAO.php
asmecher: Yes, I think on balance it'll be worth reusing the notes feature. That would be my preference, above e.g. moving to Laravel or sticking with DAOs. We already have >100 tables and all the attendant classes etc. to go with them, and it's a lot of overhead to maintain!

classes/userPrivateNote/UserPrivateNote.php
asmecher: ...but [this](https://github.com/pkp/pkp-lib/pull/9549#discussion_r1419499935) would be my preference rather than moving to Eloquent.

PR #9555, opened 2023-12-02
Update Repository.php fill missing user_id for cli mode (user_id cons…

Igor-In:
…taint index)

When I did 
php tools/importExport.php NativeImportExportPlugin import bio_6v3+.xml bio oleg 

received
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`ojsbio`.`event_log`, CONSTRAINT `event_log_user_id_foreign` FOREIGN KEY (`user_id`) REFERENCES `users` (`user_id`) ON DELETE CASCADE) (SQL: insert into `event_log` (`assoc_type`, `assoc_id`, `user_id`, `date_logged`, `event_type`, `message`, `is_translated`) values (515, 1538, 0, 2023-12-02 11:59:21, 1342177296, submission.event.fileEdited, 0))

so let them fill

### DIFF
--- a/classes/submissionFile/Repository.php
+++ b/classes/submissionFile/Repository.php
@@ -270,6 +270,7 @@ abstract class Repository
             [
                 'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION_FILE,
                 'assocId' => $submissionFile->getId(),
+                'userId' => $submissionFile->getUploaderUserId(),
                 'eventType' => SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_UPLOAD,
                 'dateLogged' => Core::getCurrentDate(),
                 'message' => 'submission.event.fileUploaded',
@@ -285,6 +286,7 @@ abstract class Repository
             [
                 'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION,
                 'assocId' => $submission->getId(),
+                'userId' => $submissionFile->getUploaderUserId(),
                 'eventType' => SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_REVISION_UPLOAD,
                 'dateLogged' => Core::getCurrentDate(),
                 'message' => 'submission.event.fileRevised',
@@ -364,6 +366,7 @@ abstract class Repository
             [
                 'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION_FILE,
                 'assocId' => $submissionFile->getId(),
+                'userId' => $submissionFile->getUploaderUserId(),
                 'eventType' => $newFileUploaded ? SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_REVISION_UPLOAD : SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_EDIT,
                 'message' => $newFileUploaded ? 'submission.event.revisionUploaded' : 'submission.event.fileEdited',
                 'isTranslated' => false,


asmecher:
@defstat, could you have a look at this issue? Thanks!

classes/submissionFile/Repository.php
```diff
@@ -270,6 +270,7 @@ public function add(SubmissionFile $submissionFile): int
             [
                 'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION_FILE,
                 'assocId' => $submissionFile->getId(),
+                'userId' => $submissionFile->getUploaderUserId(),
```
defstat, line 273: This should not be overridden here. The `userId` is already defined in "https://github.com/pkp/pkp-lib/blob/main/classes/submissionFile/Repository.php#L836" `getSubmissionFileLogData` function.

This is true for the other occurrences of this addition through this PR

defstat (changes requested):

classes/submissionFile/Repository.php
Igor-In: For cli mode <which is recommended for import sometimes> iserId stay empty. It calls above mentioned fatal error and import fails.
If overridden (thou should not be) gui mode import tool works samely, but cli mode import will be successful.

defstat:
@Igor-In thanks for this PR! 

I have added a comment in your code change. Generally any fix in a core Repository class like the `submissionFile/Repository` is not an easy task, as it may impact a large portion of the code/functionality of the software. 

In your case, and in general in the case of the CLI execution of the import/export code the problem is that a user name is defined, but the `getSubmissionFileLogData` function, is not taking that into account. 

This is a more central problem, and it may require a broader discussion regarding user management in CLI operations.

classes/submissionFile/Repository.php
defstat: @Igor-In I understand that this change may add a `userId` in the `add` function, thus resolving your the DB error in this specific case, but changing the functionality of this core function may cause your installation to behave unexpectantly.

The problem in this case is in the calling code (the import/export plugin) - that's why a broader discussion on user management in those cases may be in order.

asmecher: Thanks, @Igor-In! As @defstat suggests, we've got a bit of infrastructural work needed in order to clean up code that assumes it'll be invoked from a user request when it could sometimes be CLI. Meanwhile, I suggest editing the `getSubmissionFileLogData` function so that it doesn't just look to the request for the current user's identification.

Command line tools can allow for a user to be specified there (the `--user_name` option). When that's specified, the user is stored in the `Registry` [under the `user` key](https://github.com/pkp/pkp-lib/blob/fa3c7bf2ead5a202435927a66324735968a549ac/classes/cliTool/CommandLineTool.php#L148). For cases when `Request::getUser` returns `null` in `getSubmissionFileLogData`, this could be used as a fallback.

defstat: > Thanks, @Igor-In! As @defstat suggests, we've got a bit of infrastructural work needed in order to clean up code that assumes it'll be invoked from a user request when it could sometimes be CLI. Meanwhile, I suggest editing the `getSubmissionFileLogData` function so that it doesn't just look to the request for the current user's identification.
> 
> Command line tools can allow for a user to be specified there (the `--user_name` option). When that's specified, the user is stored in the `Registry` [under the `user` key](https://github.com/pkp/pkp-lib/blob/fa3c7bf2ead5a202435927a66324735968a549ac/classes/cliTool/CommandLineTool.php#L148). For cases when `Request::getUser` returns `null` in `getSubmissionFileLogData`, this could be used as a fallback.

@Igor-In taking into account the note that @asmecher made, please try to execute the CLI command using the `-u` switch - try

```
php tools/importExport.php NativeImportExportPlugin import bio_6v3+.xml bio -u oleg
```

instead of 

```
php tools/importExport.php NativeImportExportPlugin import bio_6v3+.xml bio oleg
```

Igor-In: Yes, you are right, that switch is obligatory for user in cli mode, sorry had not founded.
Thank You!

Igor-In:
This needs to be documented

classes/submissionFile/Repository.php
defstat: @Igor-In not a problem, the documentation on Import/Export may not have been up-to-date with this change, so it's not your fault. Please feel free to contact us for any related problem that you may encounter, and keep your involvement with the code base with possible suggestions for fixes with your contributions.

Issue #9582, opened 2023-12-20
CompileUsageStatsFromTemporaryRecords fails on PostgreSQL due to non-existing function (OJS 3.4.0.4)


bdcaf:
**Describe the bug**
in OJS 3.4.0.4 with PostgreSQL usage statistics are no longer generated. Trying to reprocess yields:
```
    "message": "SQLSTATE[42883]: Undefined function: 7 ERROR:  function date(integer) does not exist\nLINE 1: ...metrics_context\" where \"load_id\" = $1 or \"date\" = DATE(20231...\n                                                             ^\nHINT:  No function matches the given name and argument types. You might need to add explicit type casts. (SQL: delete from \"metrics_context\" where \"load_id\" = usage_events_20231214.log or \"date\" = DATE(20231214))",
```

**To Reproduce**
1. move unprocessed log file from `/var/www/files/usageStats/usageEventLogs/` into `/var/www/files/usageStats/stage/` 
2. trigger reprocess ` php lib/pkp/tools/reprocessUsageStatsMonth.php 202312`
3. start jobs `php lib/pkp/tools/jobs.php run`

**What application are you using?**
 OJS 3.4

**Additional information**
This is a relatively fresh migrated journal, we didn't check the statistics in the past - so not sure at which point this issue was introduced.

*Payload*
```
{
    "uuid": "99822df5-c608-43c9-9579-a05c6e9926cc",
    "displayName": "APP\\jobs\\statistics\\CompileUsageStatsFromTemporaryRecords",
    "job": "Illuminate\\Queue\\CallQueuedHandler@call",
    "maxTries": 1,
    "maxExceptions": 3,
    "failOnTimeout": true,
    "backoff": "5",
    "timeout": 60,
    "retryUntil": null,
    "data": {
        "commandName": "APP\\jobs\\statistics\\CompileUsageStatsFromTemporaryRecords",
        "command": "O:57:\"APP\\jobs\\statistics\\CompileUsageStatsFromTemporaryRecords\":3:{s:9:\"\u0000*\u0000loadId\";s:25:\"usage_events_20231211.log\";s:10:\"connection\";s:8:\"database\";s:5:\"queue\";s:5:\"queue\";}"
    }
}
```

*Exception*
```
{
[...]

PR #9585, opened 2023-12-22
pkp/pkp-lib#9582 fix get date from string

bozana:
s. https://github.com/pkp/pkp-lib/issues/9582

### DIFF
--- a/classes/statistics/PKPTemporaryTotalsDAO.php
+++ b/classes/statistics/PKPTemporaryTotalsDAO.php
@@ -23,6 +23,7 @@
 namespace PKP\statistics;
 
 use APP\core\Application;
+use DateTimeImmutable;
 use Illuminate\Support\Facades\DB;
 use PKP\config\Config;
 use PKP\db\DAORegistry;
@@ -107,8 +108,8 @@ abstract class PKPTemporaryTotalsDAO
      */
     public function compileContextMetrics(string $loadId): void
     {
-        $date = substr($loadId, -12, 8);
-        DB::table('metrics_context')->where('load_id', '=', $loadId)->orWhere('date', '=', DB::raw("DATE({$date})"))->delete();
+        $date = DateTimeImmutable::createFromFormat('Ymd', substr($loadId, -12, 8));
+        DB::table('metrics_context')->where('load_id', '=', $loadId)->orWhereDate('date', '=', $date)->delete();
         $selectContextMetrics = DB::table($this->table)
             ->select(DB::raw('load_id, context_id, DATE(date) as date, count(*) as metric'))
             ->where('load_id', '=', $loadId)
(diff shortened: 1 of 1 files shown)


classes/statistics/PKPTemporaryTotalsDAO.php
```diff
@@ -108,7 +108,11 @@ public function removeDoubleClicks(int $counterDoubleClickTimeFilter): void
     public function compileContextMetrics(string $loadId): void
     {
         $date = substr($loadId, -12, 8);
-        DB::table('metrics_context')->where('load_id', '=', $loadId)->orWhere('date', '=', DB::raw("DATE({$date})"))->delete();
+        $dateFormatSql = "STR_TO_DATE('{$date}', '%Y%m%d')";
+        if (substr(Config::getVar('database', 'driver'), 0, strlen('postgres')) === 'postgres') {
```
asmecher, line 112: Rather than a loose test on the driver name, I'd suggest checking the class name of the driver, e.g.:
https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/migration/upgrade/PKPv3_3_0UpgradeMigration.php#L155

But better still -- since we have the date string in PHP-land, why not use something like https://www.php.net/manual/en/datetimeimmutable.createfromformat.php to parse the string into a saner format, then use e.g. Laravel's `whereDate` to pass into the database regardless of which DBMS is used?

asmecher (changes requested):

bozana:
Hi @asmecher, yes, good idea :+1:  :sweat_smile: Could you please take another look?

classes/statistics/PKPTemporaryTotalsDAO.php
```diff
@@ -107,8 +108,8 @@ public function removeDoubleClicks(int $counterDoubleClickTimeFilter): void
      */
     public function compileContextMetrics(string $loadId): void
     {
-        $date = substr($loadId, -12, 8);
-        DB::table('metrics_context')->where('load_id', '=', $loadId)->orWhere('date', '=', DB::raw("DATE({$date})"))->delete();
+        $date = DateTimeImmutable::createFromFormat('Ymd', substr($loadId, -12, 8));
+        DB::table('metrics_context')->where('load_id', '=', $loadId)->orWhereDate('date', '=', $date->format('Y-m-d'))->delete();
```
asmecher, line 112: I think `orWhereDate` will take a `DateTimeInterface` for `$value` directly (and `DateTimeImmutable::createFromFormat` returns one -- so there shouldn't be a need to convert it back to a string with `$date->format(...)`. (Also below.)

asmecher (changes requested):
Just a small suggestion, then please go ahead, thanks!

classes/statistics/PKPTemporaryTotalsDAO.php
bozana: Oh, yes, good catch!
Changed...

asmecher:
No need, thanks! It's good to go.

Issue #9552, opened 2023-12-01
Add masthead option to user roles


bozana:
Implement the possibility for JM to define if a role should be considered and displayed on/in the masthead. This should be a further option in the add/edit role form.

PRs:
- https://github.com/pkp/pkp-lib/pull/9603
- https://github.com/pkp/ojs/pull/4138
- https://github.com/pkp/omp/pull/1499
- https://github.com/pkp/ops/pull/614

bozana:
@asmecher, could you please review the PRs above? 
Thanks a lot!

asmecher:
@bozana, looks good -- the plan is to allow individual registrations for a user group to override the user group's "default", right?

bozana:
Hi @asmecher, yes, I understand that the JII users will not have the possibility to change it on the invitation page, but the others will have that possibility, i.e. for the other there will be a possibility to define it on the user level, if a user will be considered for the masthead). Is this what you mean?

asmecher:
I am at risk of forgetting our thought process on this :) but I think it will not be uncommon for journals to want e.g. some editors listed on the masthead, others not. No matter how much everyone swears before we code it that this won't happen. :)

So the default role configuration for Editors should be `masthead=true`, as you've currently coded it :+1:

...but...

When an invitation is sent, the user who creates it (typically the journal manager) should also have a checkbox for the masthead status (only showing for user groups with `masthead=true`) allowing them override it and hide certain enrolments. The user receiving the invitation maybe will be informed of it, but won't be able to change it.

bozana:
Yes, this is so. There will be a select box with "Appears on the Masthead" and "Does not appear in the Masthead", that JM can choose from for each users and role the user is invited in.
Hmmm... Maybe there should then also be the masthead select box on the user edit form, so that JM can change it later? (e.g. if users change their mind)... :thinking: 
But, for JII plugin users there will not be this possibility, as far as I understand.

bozana:
Ah, yes, correct. This is also on my list, the change of the user edit form according to the second image here: https://github.com/pkp/pkp-lib/issues/9271#issuecomment-1700822600. So yes, there will then also be the select box on the user edit form... :-)
I will then merge this and proceed with the UIs next... :-)

bozana:
So actually, we would also need a masthead column in the user_user_groups table? :thinking:

asmecher:
> So actually, we would also need a masthead column in the user_user_groups table? 🤔

Yes, I think so!

PR #9603, opened 2024-01-09
pkp/pkp-lib#9552 add masthead option to user roles

bozana:
s. https://github.com/pkp/pkp-lib/issues/9552


asmecher (approved):

bozana:
@asmecher, may I ask you to take a look at this second commit here: https://github.com/pkp/pkp-lib/pull/9603/commits/60b5f0ada9d59973ae9b44d7b0ff7fdaf8c25443 ? -- For some reason now the PostgreSQL does not return the user groups ordered by ID and for some reason we assume that on a several cases in the code -- for example: [here](https://github.com/pkp/pkp-lib/blob/main/classes/components/forms/submission/StartSubmission.php#L178) and [here](https://github.com/pkp/pkp-lib/blob/main/classes/userGroup/Repository.php#L364) -- so that now the PostgreSQL tests could not pass without it. 
Just for you to confirm that it is OK so.

Issue #9607, opened 2024-01-10
Selecting an email template in Discussions uses the template name as subject, rather than its actual subject.


asmecher:
Using `stable-3_4_0`.

Currently when selecting an email template during discussion creation, the Subject field will be pre-filled based on the template's name:
![image](https://github.com/pkp/pkp-lib/assets/200411/f353a408-ce25-46c0-8e24-b4a3803a5167)
However, we should actually be using the Subject field in email settings instead ("Author, please check your submission" in the case):
![image](https://github.com/pkp/pkp-lib/assets/200411/7f9a9037-29ee-499e-97ef-d975bd09f8a3)

PR:
- https://github.com/pkp/pkp-lib/pull/9608

Needs forward-porting to `main`!

asmecher:
@Vitaliy-1, could you check the PR on this one? Thanks!

asmecher:
Thanks, @Vitaliy-1! Merged and forward-ported to `main`.

PR #9608, opened 2024-01-10
pkp/pkp-lib#9607 Use subject field rather than template name when creating a discussion

asmecher:


Vitaliy-1 (approved):
Looks good, thanks!

Issue #9296, opened 2023-09-12
Implement sandbox mode


asmecher:
**Describe the bug**
When taking an installation from a production environment into a testing environment, certain activities should be prevented -- sending emails, depositing issues/articles, etc.

So far we have maintained lists of individual features to disable, but it would be better to have a single configuration setting (e.g. in `config.inc.php`) to refer to.

(To be considered: what about making LessCSS cache less obtrusive for developers? The LessCSS compilation takes too long to disable entirely, but perhaps sandbox mode could consider setting no-cache directives and letting e.g. inotifywait be used on the developer side to automatically delete cache files when a .less file changes.)

Things to disable:
- Emails
- CrossRef deposits
- DataCite deposits
- ORCiD
- PayPal

**PRs**
#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9616
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4143
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/631
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1516
[crossref-ojs](https://github.com/pkp/crossref-ojs) --> https://github.com/pkp/crossref-ojs/pull/43
[crossref-ops](https://github.com/pkp/crossref-ops) --> https://github.com/pkp/crossref-ops/pull/35
[orcidProfile](https://github.com/pkp/orcidProfile) --> https://github.com/pkp/orcidProfile/pull/310

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9699
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4175
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/632
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1517
[orcidProfile](https://github.com/pkp/orcidProfile) --> https://github.com/pkp/orcidProfile/pull/311
[crossref-ojs](https://github.com/pkp/crossref-ojs) --> https://github.com/pkp/crossref-ojs/pull/44
[crossref-ops](https://github.com/pkp/crossref-ops) --> https://github.com/pkp/crossref-ops/pull/36

#### Stable-3.4.0
[...]

PR #9616, opened 2024-01-15
pkp/pkp-lib#9296 Sandbox settings

touhidurabir:
for pkp/pkp-lib#9296

### DIFF
--- a/classes/mail/Mailer.php
+++ b/classes/mail/Mailer.php
@@ -120,7 +120,8 @@ class Mailer extends IlluminateMailer
 
         // Application is set to sandbox mode and will not sent any emails
         if (Config::getVar('general', 'sandbox', false)) {
-            return null;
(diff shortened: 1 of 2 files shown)


classes/mail/Mailer.php
```diff
@@ -117,6 +117,12 @@ public function send($view, array $data = [], $callback = null)
             /** @var Mailable $view */
             $view->setData();
         }
+
+        // Application is set to sandbox mode and will not sent any emails
+        if (Config::getVar('general', 'sandbox', false)) {
+            return null;
```
asmecher, line 123: I would suggest an `error_log` call wherever an action is skipped due to sandbox mode. (And mostly stylistic, but maybe use `return;` rather than `return null;` while you're at it -- this function doesn't return anything.)

touhidurabir: done .

PR #9620, opened 2024-01-16
[PKP-LIB][main] pkp/openid#1 Initial commit - port to main

defstat:


classes/user/DAO.php
```diff
@@ -356,6 +357,20 @@ public function deleteUnvalidatedExpiredUsers(Carbon $dateTillValid, array $excl
         return $users->count();
     }
 
+    /**
+     * Get publication ids that have a matching setting
+     */
+    public function getIdsBySetting(string $settingName, $settingValue): Enumerable
```
asmecher, line 363: I think you can use the existing `Collector` class in calling code rather than adding this new function. Off the top of my head, something like:
```php
$ids = Repo::user()->getCollector()
  ->filterBySettings(['setting name' => 'setting value'])
  ->getIds();
```

classes/user/Repository.php
```diff
@@ -428,6 +429,12 @@ public function deleteUnvalidatedExpiredUsers(Carbon $dateTillValid, array $excl
         return $this->dao->deleteUnvalidatedExpiredUsers($dateTillValid, $excludableUsersId);
     }
 
+    /** @copydoc DAO:: getIdsBySetting()*/
+    public function getIdsBySetting(string $settingName, $settingValue): Enumerable
```
asmecher, line 433: (As above, this shouldn't be needed)

classes/services/PKPSchemaService.php
```diff
@@ -70,6 +70,8 @@ class PKPSchemaService
      */
     public function get($schemaName, $forceReload = false)
     {
+        Hook::call('Schema::get::before::' . $schemaName, [&$forceReload]);
```
asmecher, line 73: (Same comment as on `stable-3_4_0` PR)

asmecher (changes requested):

defstat:
Closed in favor of #10083

PR #9621, opened 2024-01-16
[PKP-LIB][stable-3_4_0] pkp/openid#1 Initial Commit for stable-3.4.0

defstat:


classes/user/DAO.php
```diff
@@ -350,4 +351,18 @@ public function deleteUnvalidatedExpiredUsers(Carbon $dateTillValid, array $excl
 
         return $users->count();
     }
+
+    /**
+     * Get publication ids that have a matching setting
+     */
+    public function getIdsBySetting(string $settingName, $settingValue): Enumerable
```
asmecher, line 358: Why not use the existing functions instead...
```php
$repo = Repo::user();
$repo->getIds($repo->getCollector()->filterBySettings(['settingName' => 'settingValue']));
```
(^ that's just off the top of my head, I might've gotten details a bit wrong)

asmecher (changes requested):
I think the DAO changes here aren't needed -- see comment; also for https://github.com/pkp/pkp-lib/pull/9620/files.

classes/services/PKPSchemaService.php
```diff
@@ -67,6 +67,8 @@ class PKPSchemaService
      */
     public function get($schemaName, $forceReload = false)
     {
+        Hook::call('Schema::get::before::' . $schemaName, [&$forceReload]);
```
asmecher, line 70: Use `Hook::run` for new hooks, and improved calling conventions for hook callback functions. See https://github.com/pkp/pkp-lib/discussions/8089 for details.

defstat: Right - fixed

PR #9628, opened 2024-01-19
pkp/pkp-lib#699 Show locale in url in multilingual contexts

jyhein:
- Show the current locale in the url in multilingual contexts.
- Old links without the locale are redirected to include the current locale.
- Monolingual contexts do not have the locale


bozana (commented):
Hi @jyhein, great work! Just a few comments/questions...

classes/core/PKPPageRouter.php
```diff
@@ -495,6 +514,75 @@ private function _getRequestedUrlParts($callback, $request): array|string|null
...
+                $request->setCookieVar('currentLocale', Locale::getLocale());
+            }
+            return $session->get('currentLocale');
+        })($setLocale ?? $urlLocale);
+
+        if (preg_match('#^/\w#', $source = $request->getUserVar('source') ?? "")) {
+            $request->redirectUrl($source);
+        }
+
+        $uri = Core::removeBaseUrl($setLocale ? ($_SERVER['HTTP_REFERER'] ?? "") : ($_SERVER['REQUEST_URI'] ?? ""));
```
bozana, line 578: Hi @jyhein, in order to support the case that Alec mentioned, I think this line should be something like:
```
        $uri = $_SERVER['PATH_INFO'];
        if ($setLocale) {
            $uri = Core::removeBaseUrl($_SERVER['HTTP_REFERER'] ?? "");
            if (!empty($request->getBasePath())) {
                $uri = str_replace($request->getBasePath(), '', $uri);
            }
        }

```
Maybe we should also somehow test if it also works if the journal is using/has configured the following part of the config.inc.php:
```

; Base URL override settings: Entries like the following examples can
; be used to override the base URLs used by OJS. If you want to use a
; proxy to rewrite URLs to OJS, configure your proxy's URL with this format.
; Syntax: base_url[journal_path] = http://www.example.com
;
; Example1: URLs that aren't part of a particular journal.
;    Example1: base_url[index] = http://www.example.com
; Example2: URLs that map to a subdirectory.
;    Example2: base_url[myJournal] = http://www.example.com/myJournal
; Example3: URLs that map to a subdomain.
;    Example3: base_url[myOtherJournal] = http://myOtherJournal.example.com
 
```
I think you have such installations, no?

Issue #9627, opened 2024-01-18
SQL in removeDoubleClicks needs improvement


bozana:
**Describe the bug**
The query from the function PKPTemporaryTotalsDAO::removeDoubleClicks() never finishes for big installation with lots of log file entries. 
S. https://forum.pkp.sfu.ca/t/ojs-3-4-0-4-statistic-problem/85132, but also University Heidelberg reported the same problem.

Thus use an combined index on load_id, context_id and ip column.
Consider also the similar function compileUniqueClicks().

**What application are you using?**
 OJS, OMP, OPS version 3.4.0.x

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9629
- https://github.com/pkp/ojs/pull/4147
- https://github.com/pkp/omp/pull/1502
- https://github.com/pkp/ops/pull/617

main:
- https://github.com/pkp/pkp-lib/pull/9640
- https://github.com/pkp/ojs/pull/4152
- https://github.com/pkp/omp/pull/1503
- https://github.com/pkp/ops/pull/618

bozana:
@asmecher and/or @jonasraoni, could (one of) your take a look into the PRs above?
I added the indexes (load_id, context_id, ip(64)) to the usage stats temporary tables. This does have a great impact. (The query now finishes withing 50s for 250000 entries... comparing with not finishing at all because of the timeout i.e. finishing in 24h as the user from the forum said).
Once these PRs are merged, I would ask affected users to apply the patch, so that their jobs do not fail any more. Because of that -- because some users will have the indexes already before the next release -- I am checking if the indexes exist in the upgrade script (which we else do not do). OK?
Unfortunately, when testing I found another bug :see_no_evil: in OMP -- I have not considered the inheritance for the insert function in two classes correctly back then. Now I corrected it too.
Thanks a lot!

jonasraoni:
I've left some notes, but it's always positive to have another bunch of eyes taking a look :)

PR #9629, opened 2024-01-19
pkp/pkp-lib#9627 Add indexes to usage stats temporary tables

bozana:
s. https://github.com/pkp/pkp-lib/issues/9627

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
+++ b/classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
@@ -1,0 +1,72 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I9627_AddUsageStatsTemporaryTablesIndexes
+ *
+ * @brief Add an index to temporary usage stats tables to fix/improve the removeDoubleClicks and compileUniqueClicks query.
+ */
+
+namespace PKP\migration\upgrade\v3_4_0;
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use PKP\install\DowngradeNotSupportedException;
+use PKP\migration\Migration;
+
+class I9627_AddUsageStatsTemporaryTablesIndexes extends Migration
+{
+    public $ipTables = [
+        'usage_stats_total_temporary_records',
+        'usage_stats_unique_item_investigations_temporary_records',
+        'usage_stats_unique_item_requests_temporary_records',
+    ];
+
+    /**
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        // Decrease the size of the column ip to 64 characters
+        foreach ($this->ipTables as $ipTable) {
+            Schema::table($ipTable, function (Blueprint $table) {
+                $table->string('ip', 64)->change();
+            });
+        }
+
+        $sm = Schema::getConnection()->getDoctrineSchemaManager();
+        Schema::table('usage_stats_total_temporary_records', function (Blueprint $table) use ($sm) {
+            $indexesFound = $sm->listTableIndexes('usage_stats_total_temporary_records');
+            if (!array_key_exists('ust_load_id_context_id_ip', $indexesFound)) {
+                $table->index(['load_id', 'context_id', 'ip'], 'ust_load_id_context_id_ip');
+            }
+        });
+        Schema::table('usage_stats_unique_item_investigations_temporary_records', function (Blueprint $table) use ($sm) {
+            $indexesFound = $sm->listTableIndexes('usage_stats_unique_item_investigations_temporary_records');
+            if (!array_key_exists('usii_load_id_context_id_ip', $indexesFound)) {
+                $table->index(['load_id', 'context_id', 'ip'], 'usii_load_id_context_id_ip');
+            }
+        });
+        Schema::table('usage_stats_unique_item_requests_temporary_records', function (Blueprint $table) use ($sm) {
+            $indexesFound = $sm->listTableIndexes('usage_stats_unique_item_requests_temporary_records');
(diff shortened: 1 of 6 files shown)


classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
```diff
@@ -0,0 +1,81 @@
...
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        $sm = Schema::getConnection()->getDoctrineSchemaManager();
+        Schema::table('usage_stats_total_temporary_records', function (Blueprint $table) use ($sm) {
+            $indexesFound = $sm->listTableIndexes('usage_stats_total_temporary_records');
+            if (!array_key_exists('ust_load_id_context_id_ip', $indexesFound)) {
+                switch (DB::getDriverName()) {
+                    case 'mysql':
```
jonasraoni, line 35: I think this way is safer `if (DB::connection() instanceof PostgresConnection) {` (i.e. using an identifier instead of a string)

classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
```diff
@@ -0,0 +1,81 @@
...
+                        $table->index([DB::raw('load_id, context_id, ip(64)')], 'ust_load_id_context_id_ip');
```
jonasraoni, line 36: I'd decrease the column size from 255 to 64 and remove the `switch`/standardize the index creation.

classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    /**
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        $sm = Schema::getConnection()->getDoctrineSchemaManager();
+        Schema::table('usage_stats_total_temporary_records', function (Blueprint $table) use ($sm) {
+            $indexesFound = $sm->listTableIndexes('usage_stats_total_temporary_records');
+            if (!array_key_exists('ust_load_id_context_id_ip', $indexesFound)) {
```
jonasraoni, line 33: I think you've added this to cover the users that already added the indexes, right? I guess the users won't be using the same index name, then it's not going to work.

In case the index name is the same, and an error happens, it's also not our fault (given the user isn't supposed to change the schema).

bozana: I added it because I would ask the two affected users to immediately patch the installation with the PRs that I add to this issue...

classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
bozana: OK, I will do that... Will then make a new commit...

bozana:
@jonasraoni, could you please take a look at the last two commits? I decreased the size of the column IP to 64 characters, so I do not need swicht {...} any more. I also removed the deprecated mhash and left only hash function.

classes/statistics/PKPTemporaryItemRequestsDAO.php
```diff
@@ -70,16 +80,42 @@ public function deleteByLoadId(string $loadId): void
      *
      * See https://www.projectcounter.org/code-of-practice-five-sections/7-processing-rules-underlying-counter-reporting-data/#counting
      */
-    public function compileUniqueClicks(): void
+    public function compileUniqueClicks(string $loadId): void
     {
         if (substr(Config::getVar('database', 'driver'), 0, strlen('postgres')) === 'postgres') {
-            DB::statement("DELETE FROM {$this->table} usur WHERE EXISTS (SELECT * FROM (SELECT 1 FROM {$this->table} usurt WHERE usurt.load_id = usur.load_id AND usurt.ip = usur.ip AND usurt.user_agent = usur.user_agent AND usurt.context_id = usur.context_id AND usurt.submission_id = usur.submission_id AND EXTRACT(HOUR FROM usurt.date) = EXTRACT(HOUR FROM usur.date) AND usur.line_number < usurt.line_number) AS tmp)");
+            DB::statement(
```
jonasraoni, line 86: I didn't inspect the query, but isn't it possible to write it using the Laravel stuff to have a single query (and just add exceptions for the `EXTRACT`/`TIMESTAMPDIFF`)?

classes/migration/upgrade/v3_4_0/I9627_AddUsageStatsTemporaryTablesIndexes.php
jonasraoni: Just tagging @asmecher to check if he agree and possibly look out for this on next reviews (ensure migrations can be re-executed).

classes/statistics/PKPTemporaryItemRequestsDAO.php
bozana: @jonasraoni, PostgreSQL does not support inner joint with delete, and the other way around too (MySQL does not understand the PostgreSQL's query here), so I would leave this as is for now, OK?

jonasraoni: For the query `DB::table('submissions', 'a')->join('submissions AS b', 'b.submission_id', '=', 'a.submission_id')->delete();`, Laravel outputs this for PostgreSQL:

```sql
delete from "submissions" as "a" where "ctid" in (
	select "a"."ctid" from "submissions" as "a" inner join "submissions" as "b" on "b"."submission_id" = "a"."submission_id"
)
```

I think it will work fine for your case, but I don't think this is a blocker 😁

bozana: OK, let me try it with PostgreSQL... I am a little bit afraid to change anything... but I will test it thoroughly... :sweat_smile:

bozana: I created a new issue for that -- to test and see if we could use the Laravel query builder -- in order to finish this issue as soon as possible and provide affected users with the patch. S. https://github.com/pkp/pkp-lib/issues/9636.

jonasraoni: Not a big deal, I marked this as resolved 😁

Issue #9541, opened 2023-11-28
Fatal error when adding reviewer


jonasraoni:
**Describe the bug**
When trying to create a reviewer out of an existing user, the fatal error below will be shown if the user already has the reviewer role.

The main problem lies on the search, OJS 3.3 doesn't display users which are already reviewers, but OJS 3.4 has a regression.
But code that inserts unique data into the database should be more defensive, so there are two places to be fixed.
 
```php
PHP Fatal error:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '186-1082' for key 'user_user_groups.user_user_groups_unique' in /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php:545
Stack trace:
    #0 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(545): PDOStatement->execute()
#1 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(753): Illuminate\\Database\\Connection->Illuminate\\Database\\{closure}('insert into `us...', Array)
#2 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(720): Illuminate\\Database\\Connection->runQueryCallback('insert into `us...', Array, Object(Closure))
#3 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(546): Illuminate\\Database\\Connection->run('insert into `us...', Array, Object(Closure))
#4 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(498): Illuminate\\Database\\Connection->statement('insert into `us...', Array)
#5 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(3274): Illuminate\\Database\\Connection->insert('insert into `us...', Array)
#6 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(1869): Illuminate\\Database\\Query\\Builder->insert(Array)
#7 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1306): Illuminate\\Database\\Eloquent\\Builder->__call('insert', Array)
[...]

PR #9641, opened 2024-01-24
[WIP] pkp/pkp-lib#9541 Validate user and userGroup before executing EnrollExistingReviewerForm

Vitaliy-1:


classes/user/Collector.php
```diff
@@ -457,42 +457,61 @@ protected function buildUserGroupFilter(Builder $query): self
...
+                    ->where('uug.date_end', '<=', $currentDateTime)
+            );
+
+        $joinSub = clone $subQuery;
+        $query
+            ->whereExists(
+                $subQuery->when(
+                    $this->excludeRoles != null,
+                    // This aggregates a column role_count, which holds an information whether user holds specified role
+                    fn (Builder $subQuery) => $subQuery->leftJoinSub(
```
bozana, line 501: What do you think about using this here instead:
```
fn (Builder $subQuery) => $subQuery->whereNotIn(
  'u.user_id',
  $joinSub
    ->select('uug.user_id')
    ->whereIn('ug.role_id', $this->excludeRoles)
)
```

I wanted to double check and test that tomorrow, but wanted to write it already to you too, in case... :-)

bozana: Hi @Vitaliy-1, I believe the following would be the best query to use:
```
                    fn (Builder $subQuery) => $subQuery->leftJoinSub(
                        $joinSub
                            ->select('uug.user_id')
                            ->whereIn('ug.role_id', $this->excludeRoles),
                        'agr',
                        'uug.user_id',
                        'agr.user_id'
                    )->whereNull('agr.user_id')

```

What do you think?

Vitaliy-1: Agree!

Issue #9643, opened 2024-01-25
Fatal error on latest PHP 8.x releases


jonasraoni:
**Describe the bug**
More details at the "how to reproduce" section.
The code `(new IntlDateFormatter('C', 0, 0))->format(new DateTime())` generates a fatal error on the latest PHP 8.x releases, see a [reproduction case here](https://onlinephp.io?s=s7EvyCjg5UpNzshX0MhLLVfwzCvJcUksSXXLL8pNLClJLdJQd1bX1NS1SwMLgNWA5EMyc1M1NDWtAQ%2C%2C&v=8.3.1%2C8.2.14%2C8.1.27%2C7.4.33) and the exception below.

```php
PHP Fatal error:  Uncaught Error: Found unconstructed IntlDateFormatter in /var/www/html/lib/pkp/lib/vendor/php81_bc/strftime/src/php-8.1-strftime.php:101
Stack trace:
#0 /var/www/html/lib/pkp/lib/vendor/php81_bc/strftime/src/php-8.1-strftime.php(101): IntlDateFormatter->format(Object(DateTime))
#1 /var/www/html/lib/pkp/lib/vendor/php81_bc/strftime/src/php-8.1-strftime.php(202): PHP81_BC\\{closure}(Object(DateTime), '%B')
#2 [internal function]: PHP81_BC\\{closure}(Array)
#3 /var/www/html/lib/pkp/lib/vendor/php81_bc/strftime/src/php-8.1-strftime.php(181): preg_replace_callback('/(?<!%)%([_#-]?...', Object(Closure), '%B')
#4 /var/www/html/lib/pkp/classes/template/PKPTemplateManager.inc.php(2266): PHP81_BC\\strftime('%B', Object(DateTime))
#5 /var/www/html/cache/t_compile/91c4852bbf0ab9d68a26a0a5d3411137ee5758f8^756a7119cb48731a8fd45c250b45f4bcca2d6b01_0.app.frontendpagessearch.tpl.php(88): PKPTemplateManager->smartyHtmlSelectDateA11y(Array, Object(Smarty_Internal_Template))
#6 /var/www/html/lib/pkp/lib/vendor/smarty/smarty/libs/sysplugins/smarty_template_resource_base.php(123): content_64c2d4d46b0704_54914599(Object(Smarty_Internal_Template))
[...]

PR #9647, opened 2024-01-25
pkp/pkp-lib#9643 Added patch to strftime polyfill to retrieve the loc…

jonasraoni:
…ale from the application


lib/strftime-locale-patch.diff
```diff
@@ -0,0 +1,5 @@
+diff -u -r src/php-8.1-strftime.php src/php-8.1-strftime.php
+--- php-8.1-strftime.php	2022-10-27 21:37:02.000000000 +0300
++++ php-8.1-strftime.php	2024-01-25 21:34:27.000000000 +0300
+@@ -30,0 +31 @@
++    $locale = $locale ?? \AppLocale::getLocale();
```
jonasraoni, line 5: By the way, the `$locale` is the 3rd parameter of the polyfill, which doesn't exist in the polyfilled function (see: https://www.php.net/manual/en/function.strftime.php).

Issue #9664, opened 2024-01-29
Use jobs chain for usage stats log files processing


bozana:
Use a job for usage stats log file processing, split the job CompileUsageStatsFromTemporaryRecords into several jobs, and chain all usage stats jobs:

Above all the function removeDoubleClicks() and compileUniqueClicks() executed in the job CompileUsageStatsFromTemporaryRecords take some time for a huge installations (e.g. ca. 50sec. for 20000 rows). In order to avoid a timeout, the job CompileUsageStatsFromTemporaryRecords can be separated into several smaller jobs.

Till now the scheduled task UsageStatsLoader has been used to read an usage stats log file and insert the entries into the usage stats temporary tables. Then the job CompileUsageStatsFromTemporaryRecords would compile the stats form the temporary tables. Because the job is executed asynchronously/at some later point, it could happen that the UsageStatsLoader will fill the temporary tables with new data, from the next log file, while the job CompileUsageStatsFromTemporaryRecords still have not compiled the stats from the previous log file. Which leads to a huge amount of data in the temporary tables, which makes the processing difficult and can cause a timeout. (This is above all the case when the user wants to reprocess the old log files -- the old log files need to be reprocessed for the whole month). Thus, the log file processing (parsing and inserting into the temporary tables) can be done from a job as well. Also, all the usage stats jobs needs to be chained i.e. they need to be executed one after another -- so that no job can write into the temporary tables before the other has finished, but also to ensure the correct stats when the job CompileUsageStatsFromTemporaryRecords is separated into several jobs.

The scheduled task UsageStatsLoader would log a message if something would be wrong with a line in the log file, or if the inserting the line into the temporary tables would not work for some reason, however this would not stop the scheduled task from working and processing other log file lines. A similar solution would be needed for jobs -- a way of logging or informing a user that something was wrong with a line.

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9672
- https://github.com/pkp/ojs/pull/4164
- https://github.com/pkp/omp/pull/1511
- https://github.com/pkp/ops/pull/627

main:
- https://github.com/pkp/pkp-lib/pull/9710
- https://github.com/pkp/ojs/pull/4179
- https://github.com/pkp/omp/pull/1520
- https://github.com/pkp/ops/pull/635

jonasraoni:
@bozana Just something to consider when using jobs https://github.com/pkp/pkp-lib/issues/9591

bozana:
[...]

PR #9672, opened 2024-01-30
pkp/pkp-lib#9664 Use jobs chain for usage stats log files processing 

bozana:
s. https://github.com/pkp/pkp-lib/issues/9664

### DIFF
--- a/classes/task/FileLoader.php
+++ b/classes/task/FileLoader.php
@@ -28,6 +28,7 @@ use PKP\site\SiteDAO;
 abstract class FileLoader extends ScheduledTask
 {
     public const FILE_LOADER_RETURN_TO_STAGING = 0x01;
+    public const FILE_LOADER_RETURN_TO_DISPATCH = 0x02;
     public const FILE_LOADER_ERROR_MESSAGE_TYPE = 'common.error';
     public const FILE_LOADER_WARNING_MESSAGE_TYPE = 'common.warning';
 
@@ -35,6 +36,7 @@ abstract class FileLoader extends ScheduledTask
     public const FILE_LOADER_PATH_PROCESSING = 'processing';
     public const FILE_LOADER_PATH_REJECT = 'reject';
     public const FILE_LOADER_PATH_ARCHIVE = 'archive';
+    public const FILE_LOADER_PATH_DISPATCH = 'dispatch';
 
     /** The current claimed filename that the script is working on. */
     private string $_claimedFilename;
@@ -50,6 +52,9 @@ abstract class FileLoader extends ScheduledTask
 
     /** Archive directory path. */
     private string $_archivePath;
+
+    /** Dispatch directory path. */
+    private string $_dispatchPath;
 
     /** Reject directory path. */
     private string $_rejectPath;
@@ -97,6 +102,7 @@ abstract class FileLoader extends ScheduledTask
             $this->_archivePath = "{$basePath}/" . self::FILE_LOADER_PATH_ARCHIVE;
             $this->_rejectPath = "{$basePath}/" . self::FILE_LOADER_PATH_REJECT;
             $this->_processingPath = "{$basePath}/" . self::FILE_LOADER_PATH_PROCESSING;
+            $this->_dispatchPath = "{$basePath}/" . self::FILE_LOADER_PATH_DISPATCH;
         }
 
         // Set admin email and name.
@@ -140,6 +146,14 @@ abstract class FileLoader extends ScheduledTask
     public function getArchivePath(): string
     {
         return $this->_archivePath;
+    }
+
+    /**
+     * Return the dispatch path.
+     */
+    public function getDispatchPath(): string
+    {
+        return $this->_dispatchPath;
     }
 
     /**
@@ -206,7 +220,8 @@ abstract class FileLoader extends ScheduledTask
             $this->_stagePath,
             $this->_archivePath,
             $this->_rejectPath,
-            $this->_processingPath
+            $this->_processingPath,
+            $this->_dispatchPath
         ];
         $fileManager = null;
         foreach ($pathsToCheck as $path) {
@@ -269,6 +284,13 @@ abstract class FileLoader extends ScheduledTask
                 // Let the script know what files were sent back to staging,
                 // so it doesn't claim them again thereby entering an infinite loop.
                 $this->_stagedBackFiles[] = $this->_claimedFilename;
+            } elseif ($result === self::FILE_LOADER_RETURN_TO_DISPATCH) {
+                // Move the file to dispatch folder, where a dispatched job will find it
+                $this->_dispatchFile();
+                $this->addExecutionLogEntry(__(
+                    'admin.fileLoader.fileDispatched',
+                    ['filename' => $filePath]
+                ), ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_NOTICE);
             } else {
                 $this->_archiveFile();
             }
@@ -369,6 +391,14 @@ abstract class FileLoader extends ScheduledTask
     private function _rejectFile(): void
     {
         $this->moveFile($this->_processingPath, $this->_rejectPath, $this->_claimedFilename);
+    }
+
+    /**
+     * Move the current claimed file into the dispatch folder.
+     */
+    protected function _dispatchFile(): void
+    {
+        $this->moveFile($this->_processingPath, $this->_dispatchPath, $this->_claimedFilename);
     }
 
     /**
--- a/classes/task/FileLoader.php
+++ b/classes/task/FileLoader.php
@@ -27,8 +27,8 @@ use PKP\site\SiteDAO;
 
 abstract class FileLoader extends ScheduledTask
 {
-    public const FILE_LOADER_RETURN_TO_STAGING = 0x01;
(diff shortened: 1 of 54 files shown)


classes/task/FileLoader.php
```diff
@@ -269,6 +284,13 @@ protected function executeActions(): bool
                 // Let the script know what files were sent back to staging,
                 // so it doesn't claim them again thereby entering an infinite loop.
                 $this->_stagedBackFiles[] = $this->_claimedFilename;
```
jonasraoni, line 286: Unrelated to the changes, but if you replace the `opendir()` by the `FilesystemIterator`, then this won't be needed.

classes/task/PKPUsageStatsLoader.php
```diff
@@ -114,88 +105,31 @@ protected function executeActions(): bool
...
-        // check canonicalUrl ?
-        if (!is_int($entry->contextId)) {
-            throw new Exception(__('admin.scheduledTask.usageStatsLoader.invalidLogEntry.contextId'));
-        }
-        if (!empty($entry->submissionId) && !is_int($entry->submissionId)) {
-            throw new Exception(__('admin.scheduledTask.usageStatsLoader.invalidLogEntry.submissionId'));
-        }
+            foreach ($this->months as $month) {
+                $compileMonthlyMetricsJob = new CompileMonthlyMetrics($month, $site);
+                $jobs = array_merge($jobs, [$compileMonthlyMetricsJob]);
```
jonasraoni, line 118: Is it possible for the list of files to grow into an unmanageable size?
For example, if the user decides to reprocess his whole archive.
If yes, then it could be interesting to add a limit.

classes/task/PKPUsageStatsLoader.php
```diff
@@ -114,88 +105,31 @@ protected function executeActions(): bool
...
+            }
+            // Bus::chain() cannot accept an empty array
+            if (!empty($jobs)) {
+                Bus::chain($jobs)
+                    ->catch(function (Throwable $e) {
```
jonasraoni, line 123: What types of errors can come here? If there's something useful, it might be better to log, the `Throwable` can catch even code errors.

classes/task/PKPUsageStatsLoader.php
```diff
@@ -114,88 +105,31 @@ protected function executeActions(): bool
...
-    {
-        try {
-            $this->temporaryTotalsDao->insert($entry, $lineNumber, $loadId);
-            $this->temporaryInstitutionsDao->insert($entry->institutionIds, $lineNumber, $loadId);
-            if (!empty($entry->submissionId)) {
-                $this->temporaryItemInvestigationsDao->insert($entry, $lineNumber, $loadId);
-                if ($entry->assocType == Application::ASSOC_TYPE_SUBMISSION_FILE) {
-                    $this->temporaryItemRequestsDao->insert($entry, $lineNumber, $loadId);
-                }
+        if ($processFilesResult) {
```
jonasraoni, line 109: I think it's better to use early returns when possible, to decrease the code indentation and make it easier to read.
So here, I'd change to:
```php
if (!$processFilesResult) {
    return false;
}
```

jobs/statistics/ArchiveUsageStatsLogFile.php
```diff
@@ -0,0 +1,80 @@
...
+                ]
+            );
+            throw new JobException($message);
+        }
+        if ($this->site->getData('compressStatsLogs')) {
+            try {
+                $fileMgr = new FileManager();
+                $fileMgr->gzCompressFile($archiveFilePath);
+            } catch (Exception $e) {
+                error_log($e->getMessage());
```
jonasraoni, line 76: I think it's better to use `error_log($e)` to display the stack trace together with the error, or we won't be able to understand where it happened.

jobs/statistics/ArchiveUsageStatsLogFile.php
```diff
@@ -0,0 +1,80 @@
...
+use PKP\jobs\BaseJob;
+use PKP\site\Site;
+use PKP\task\FileLoader;
+
+class ArchiveUsageStatsLogFile extends BaseJob
+{
+    /**
+     * The number of times the job may be attempted.
+     */
+    public $tries = 1;
```
jonasraoni, line 32: Shouldn't we try again in a couple of days and/or warn the sysadmin if there's something not working?

jonasraoni (approved):
@bozana I didn't test in my environment, I think everything makes sense, so I just left some notes.

classes/task/FileLoader.php
```diff
@@ -28,13 +28,15 @@
 abstract class FileLoader extends ScheduledTask
```
jonasraoni, line 28: Imagine that this task ran, and created many jobs to process the files.
Then, while the jobs are still being processed (e.g. maybe there are so many that it's taking days to finish), this task gets executed again.

Is it going to generate issues (reprocessing)? Is it worth to give IDs to the jobs to avoid them from being duplicated on the queue?

jobs/statistics/ArchiveUsageStatsLogFile.php
jonasraoni: About the retry, as we have files waiting to be processed, they are enough to trigger the retry 😁

classes/task/FileLoader.php
bozana: There are two cases that would need attention:
- If the re-processing of the same log file has started next day: This case can only occur if the admin explicitly does it. But this we could eventually solve by having an ID, to avoid them from being duplicated on the queue. I will need to think about how to solve it. Maybe in a new issue. It is probably also a rare case, because I hope that admin will not do such a think :-)
- If log files processing (e.g. when the admin re-process the whole month) takes longer than a day, and the normal scheduled task that runs every day (and that would process the log file from yesterday) is run on the next day: Then the new scheduled task would create a chain of the new jobs (for the new log file) and those can be run in parallel with the jobs from the re-processing the whole month. This would again lead to enormous DB tables that would break the processing. This I do not know how to solve, I would need to investigate what is possible. But also this I would like to leave for a new issue.

jonasraoni: I've skimmed through the documentation of Laravel queue and I remember there were some features to address such problems. I'm just trying to help with questions, feel free to create new issues for the things that make sense :)

classes/task/PKPUsageStatsLoader.php
bozana: The CLI tool for re-processing takes 'only' one month.
If the user keeps the daily stats, it would be possible to copy all archived log files into the usageEventLogs folder and run the scheduled task -- then as you said all would be considered...
I hope the admins know what are they doing :sweat_smile: 
Hmmm... If we would add limit for 10-30 files, the admin could however ignore it and start a new scheduled task... So maybe first we will need to solve the two cases from the beginning.
I will document it all in a new issue...

classes/task/PKPUsageStatsLoader.php
bozana: > When chaining jobs, you may use the catch method to specify a closure that should be invoked if a job within the chain fails. The given callback will receive the Throwable instance that caused the job failure

OK, yes, I will log the error...

bozana: Hmmm... I think the error message is then visible in the failed_jobs table -- that's why I haven't done anything here...

Maybe to ask @touhidurabir if I should also log the error here?

jobs/statistics/ArchiveUsageStatsLogFile.php
bozana: I believe it is OK to have tries = 1. If the job fails the message will be visible in the failed_jobs table, and then the admin can decide if it can be re-run or if the problem first needs to be solved.

jonasraoni: FYI, there's a setting to delete failed jobs here (which I think should be "off" by default!), so they will be lost after some time:
https://github.com/pkp/ojs/blob/f60b51191ded50462eb76a75d1c9c48882365640/config.TEMPLATE.inc.php#L560

classes/task/PKPUsageStatsLoader.php
bozana: I think this needs to be added to the https://docs.pkp.sfu.ca/admin-guide/en/statistics#reprocess-log-files.

classes/task/FileLoader.php
touhidurabir: @bozana @jonasraoni I think what we are looking here in this case is making the jobs unique . It should be very easy just by making any job class implementing `Illuminate\Contracts\Queue\ShouldBeUnique` . Need to bit checking to make sure it can properly interact with the cache but seems ok to me and no changes require . see more at https://laravel.com/docs/10.x/queues#unique-jobs .

classes/task/PKPUsageStatsLoader.php
touhidurabir: we don't need to log the anything if we only need to log the exception as those do get logged in failed_jobs and also in the error log file as we have the `JobFailed` event register trough `Queue::failing`. We can use that `catch` if we need to run anything extra if a chain job failed. more details at https://laravel.com/docs/10.x/queues#chain-failures .

Issue #9686, opened 2024-02-02
The arguments for the jobs.php are not working


jonasraoni:
**Describe the bug**
When calling `php lib/pkp/tools/jobs.php`, all arguments that have values are ignored and their default values are used instead.

**To Reproduce**
1.  Run `php lib/pkp/tools/jobs.php work --max-jobs=1`
2. See that the value was ignored

**What application are you using?**
OJS 3.4

jonasraoni:
@touhidurabir I think there's nothing special here, but if you'd like to take a look, see the linked PRs.

touhidurabir:
@jonasraoni I added one comment for both PR, please check that . Just you need to make the same change at 2 more places and it should be good to merge .

jonasraoni:
Thanks!

PR #9687, opened 2024-02-02
pkp/pkp-lib#9686 Fixed the setup for the worker configuration

jonasraoni:


touhidurabir (commented):
@jonasraoni seems fine . There is 2 more places the same silly mistake slips in , see the `failed` method with line as `$this->getParameterValue('--queue')` as those need the fix too .

Issue #9302, opened 2023-09-13
Internal PHP locale fails to be defined under Windows and others


jonasraoni:
**Describe the bug**
The [locale format used by Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/language-strings?view=msvc-170) uses a `-` instead of `_`, therefore the call to `setlocale()` should consider the OS or retry with another combination (sounds like the best choice).

Besides this, when the OS doesn't have the expected locale, it fails silently and keeps the last locale that worked. Which is bad for long-lived processes (e.g. PHP-FPM).

**What application are you using?**
OJS 3.3

asmecher:
@defstat, have you experienced this, or is it maybe something that gets configured/mapped with WSL?

jonasraoni:
I've tested on my machine, I can create a small PR later :)

defstat:
@asmecher I have not experienced that. @jonasraoni can you point out when you are experiencing it?

jonasraoni:
If you change the PHP's internal locale to an invalid value, some functions that depend on it, will not work properly (under WSL should work, as it uses a Linux distribution).

```php
setlocale(LC_ALL, 'fr_FR');
echo strftime('%B'); //September
setlocale(LC_ALL, 'fr-FR');
echo strftime('%B'); //septembre
```

defstat:
That should have affected ojs also, if it was running under windows (not WSL), right?

jonasraoni:
@asmecher Regarding `*nix`... For the `setlocale()` to work properly, it's expected from the server to have the needed locales at `locale -a` (we currently have this issue on our servers). So in the long run, I think we should avoid depending on the `setlocale()` 🤔

On another hand, such functions work fine: https://www.php.net/manual/en/class.intldateformatter.php.

defstat:
@jonasraoni I have never experienced issues with locales on my windows installations (I am exclusively working on windows for ojs for some years now).. 

But something might have changed on the php version that you are working on, or I may have missed something on the certain process that you are experiencing this...

jonasraoni:
Here I've used an updated Windows 11 + PHP 8.2.
I've just ran `php -a`, then pasted [the code above](https://github.com/pkp/pkp-lib/issues/9302#issuecomment-1721526514), and got `September`, then `septembre`.

defstat:
Do you have an ojs example to demonstrate the problem so I can test it on the software?

jonasraoni:
[...]

PR #9700, opened 2024-02-06
pkp/pkp-lib#9302 Use POSIX locale and setup the default locale for th…

jonasraoni:
…e PHP's intl functions

### DIFF
--- a/classes/i18n/Locale.php
+++ b/classes/i18n/Locale.php
@@ -40,6 +40,7 @@ use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use RecursiveRegexIterator;
 use RegexIterator;
+use ResourceBundle;
 use Sokil\IsoCodes\Database\Countries;
 use Sokil\IsoCodes\Database\Currencies;
 use Sokil\IsoCodes\Database\LanguagesInterface;
@@ -143,11 +144,8 @@ class Locale implements LocaleInterface
         }
 
         $this->locale = $locale;
-        $dashedLocale = str_replace('_', '-', $locale);
-        if (!setlocale(LC_ALL, "{$locale}.utf8", $locale, "{$dashedLocale}.utf8", $dashedLocale)) {
-            error_log("Failed to set locale \"{$locale}.utf8\", ensure that all locales enabled in the application are available in your operational system");
-        }
-        putenv("LC_ALL={$locale}");
+        setlocale(LC_ALL, 'C.utf8', 'C');
+        \Locale::setDefault(\Locale::lookup(ResourceBundle::getLocales(''), $locale, true));
     }
 
     /**


classes/i18n/Locale.php
```diff
@@ -143,11 +144,8 @@ public function setLocale($locale): void
         }
 
         $this->locale = $locale;
-        $dashedLocale = str_replace('_', '-', $locale);
-        if (!setlocale(LC_ALL, "{$locale}.utf8", $locale, "{$dashedLocale}.utf8", $dashedLocale)) {
-            error_log("Failed to set locale \"{$locale}.utf8\", ensure that all locales enabled in the application are available in your operational system");
-        }
-        putenv("LC_ALL={$locale}");
+        setlocale(LC_ALL, 'C.utf8', 'C');
```
asmecher, line 147: @jonasraoni, do you know what the effect of `setlocale` is when `Locale::setDefault` is also used? If we're just going to force `C` or `C.utf8`, why not just leave out the `setlocale` call entirely?

jonasraoni: do you know what the effect of setlocale is when Locale::setDefault is also used? >

@asmecher The `Locale::setDefault()` is applied just to the [intl functions](https://www.php.net/manual/en/book.intl.php), it doesn't affect the `strftime()` for example.

> why not just leave out the setlocale call entirely?

I've left just because of this warning: https://www.php.net/setlocale#:~:text=Warning,of%20PHP%207.0.5.
The use case would be a server, like Apache, holding PHP under a thread, and one of them calls the `setlocale()` (perhaps another application like WordPress)... Basically a race-condition 😁

I'd leave it this way, but if you prefer to drop, no objections. I'll wait for a direction 😁

asmecher: OK, but in the previous behaviour, functions like `strftime` would be localized properly via the user of `setlocale` with the appropriate `$locale` option (assuming that's available on the server, which ideally is the case). Now these functions will behave with a hard-coded `C` locale. We're not using `strftime` anymore (deprecated), but off the top of my head, this would still include e.g. number formatting etc.

jonasraoni: Yeah! So, what I'm proposing is to use the `C`, and fix any problem that might arise (if there's something bad, it went unnoticed for all these years, so I'm not very scared about it) through any other way (the `intl` classes must cover probably all the cases).

jonasraoni: Another option is to map our locale to a better one (using the `ResourceBundle::getLocales('')`) in `3.4`, and drop the `setlocale()` just in `3.5`.

asmecher:
I don't think I see much value in making any changes to `stable-3_3_0`, if we don't have any reports of problems. In other words, what's the harm in reverting https://github.com/pkp/pkp-lib/commit/c80e90ca19f209f3deae2cb1d813d4f34daa8966 and leaving `stable-3_3_0` as is?

(I don't consider `stable-3_4_0` to be quite the same case -- `setlocale` doesn't expect RFC 4646 locale names, which we've adopted more broadly starting with 3.4.0-0.)

classes/i18n/Locale.php
asmecher: I don't think I see much value in making any changes to stable-3_3_0, if we don't have any reports of problems. In other words, what's the harm in reverting https://github.com/pkp/pkp-lib/commit/c80e90ca19f209f3deae2cb1d813d4f34daa8966 and leaving stable-3_3_0 as is?

(I don't consider stable-3_4_0 to be quite the same case -- setlocale doesn't expect RFC 4646 locale names, which we've adopted more broadly starting with 3.4.0-0.)

jonasraoni: About `stable-3_3_0`, if I have a more predictable option available, that's the one I take, but I don't have anything against reverting.
I'll wait for a final direction, my preferences are expressed on the PRs :)

PR #9726, opened 2024-02-13
Move Hook call before insert

forgive38:
Hello

With this little change, we can give the possibility to change Author object before insert...

Thank you

### DIFF
--- a/classes/author/Repository.php
+++ b/classes/author/Repository.php
@@ -159,10 +159,10 @@ class Repository
             $author->setData('seq', $nextSeq);
         }
 
+        Hook::call('Author::add', [$author]);
+
         $authorId = $this->dao->insert($author);
         $author = Repo::author()->get($authorId);
-
-        Hook::call('Author::add', [$author]);
 
         return $author->getId();
     }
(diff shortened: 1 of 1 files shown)


classes/author/Repository.php
```diff
@@ -159,11 +159,11 @@ public function add(Author $author): int
             $author->setData('seq', $nextSeq);
         }
 
+        Hook::call('Author::add', [$author]);
```
jonasraoni, line 162: There's a `Hook::call('Author::delete::before', [$author]);`, so instead of moving this block (I consider it a breaking change), I think it's better to follow the pattern and add an `Author::add::before`.

Given the other "repositories" are following the same pattern, it would be better to update them as well (the `Author::edit` is also missing a `before` hook).

forgive38: Ok, I understand. 
Even if introducing breaking change is not really a problem :)
 I have developed some plugins for ojs 3.3 and I have to update them to follow changes between 3.3 and 3.4

jonasraoni: I agree, the codebase had way too many breaking changes. Hopefully it will decrease, it's really tough to understand what you've got to change without a well-written changelog.

Issue #9736, opened 2024-02-15
Allow ordering of masthead roles


bozana:
There should be a possibility for JM to order the roles selected to appear on masthead. This should be available under Website > Appearance > Masthead, and it should look similar to Website > Appearance > Setup > Sidebar, but if possible without the checkbox.

PRs:
- https://github.com/pkp/ui-library/pull/331
- https://github.com/pkp/pkp-lib/pull/9738
- https://github.com/pkp/ojs/pull/4191
- https://github.com/pkp/omp/pull/1537
- https://github.com/pkp/ops/pull/650

bozana:
Hi @asmecher, could you take a look at the PRs above? @jardakotesovec will review the ui-library part, but if you could take a look how the order of the masthead roles is saved (as a context setting) and if it is OK so.

asmecher:
Sorry for the wait! I just added a small comment; it's ready to go otherwise.

PR #9738, opened 2024-02-16
pkp/pkp-lib#9736 Allow ordering of masthead roles

bozana:
s. https://github.com/pkp/pkp-lib/issues/9736

### DIFF
--- a/locale/sq/manager.po
+++ b/locale/sq/manager.po
@@ -1,0 +1,17 @@
+# Rudin Dakli <rdakli@gmail.com>, 2023.
+msgid ""
+msgstr ""
+"PO-Revision-Date: 2023-02-21 17:56+0000\n"
+"Last-Translator: Rudin Dakli <rdakli@gmail.com>\n"
+"Language-Team: Albanian <http://translate.pkp.sfu.ca/projects/ojs/manager/sq/"
+">\n"
+"Language: sq\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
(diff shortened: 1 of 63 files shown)


locale/se/manager.po
```diff
@@ -0,0 +1,5 @@
+# Huw Robert Grange <huw.r.grange@uit.no>, 2023.
+msgid ""
+msgstr "X-Generator: Weblate\nMIME-Version: 1.0\nContent-Type: text/plain; charset=UTF-8\nContent-Transfer-Encoding: 8bit"
```
asmecher, line 3: It looks like you intended to add a translation here but there's just a bit of metadata.

asmecher (changes requested):
Just a couple of minor comments, thanks!

locale/sq/manager.po
```diff
@@ -0,0 +1,17 @@
...
+"Language: sq\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=n != 1;\n"
+"X-Generator: Weblate 4.13.1\n"
+
+
+msgid "manager.setup.masthead"
+msgstr ""
```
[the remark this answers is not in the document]
bozana: Yes, I moved this key from OJS to pkp-lib and it was also empty in the OJS. So I will leave it as it is...

locale/se/manager.po
bozana: I will then remove this file. It happened when using the CLI tool to move a locale key from OJS into PKP-LIB...
I haven't notices this. Thanks!

Issue #9759, opened 2024-02-22
Additional file validation causes error


jardakotesovec:
**Describe the bug**
Please tell us what happens, what you expected to happen, and why you think it is a bug in the software.

**To Reproduce**
1. Enable a plugin that triggers the `submissionfilesuploadform::validate` hook and uses `submissionfilesuploadform->addError()`, like https://github.com/ajnyga/allowedUploads
2. Enter the valid file extensions to the plugin settings
3. Go to a submission workflow and try to upload a file that is not permitted in the plugin settings

**What application are you using?**
 OJS 3.4 stable

**Additional information**
![Näyttökuva 2024-2-22 kello 14 46 25](https://github.com/pkp/pkp-lib/assets/16347527/12cf9d3a-3eea-4efb-b105-629e411f7941)

jardakotesovec:
From looking to usual uses of JSONMessage. Second parameter should be error message to be displayed via alert dialog. But given that form validation errors seems to be communicated from fetchNotification, its probably correct just return JSONMessage(false).

pkp-lib PR https://github.com/pkp/pkp-lib/pull/9760

jardakotesovec:
3.4.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/9760
ojs (tests only): https://github.com/pkp/ojs/pull/4201

jardakotesovec:
@jonasraoni Can I ask you for review?

jonasraoni:
Merged and cherry picked in main.

PR #9760, opened 2024-02-22
pkp/pkp-lib#9759 form validation error is communicated via notificati…

jardakotesovec:
…ons, passing fetch() I think is typo


controllers/wizard/fileUpload/FileUploadWizardHandler.php
```diff
@@ -415,7 +415,7 @@ public function uploadFile($args, $request)
 
         // Validate the form and upload the file.
         if (!$uploadForm->validate()) {
-            return new JSONMessage(false, $uploadForm->fetch($request));
```
asmecher, line 418: The pattern of including the form markup in the JSON failure message was to replace the "failed" form markup with an updated version that might include more details. If it helps: this is processed e.g. in the AjaxFormHandler in our old JS stack; see e.g.:

https://github.com/pkp/pkp-lib/blob/bf99547b8cbe7dd3014b306c165fabb6b05372f8/js/controllers/form/AjaxFormHandler.js#L123..L140

controllers/wizard/fileUpload/FileUploadWizardHandler.php
```diff
@@ -415,7 +415,7 @@ public function uploadFile($args, $request)
...
+            return new JSONMessage(true, $uploadForm->fetch($request));
```
jonasraoni, line 418: I think this doesn't make sense, everywhere else is using `JSONMessage(false)` for failures 🤔

js/controllers/wizard/fileUpload/form/FileUploadFormHandler.js
```diff
@@ -205,7 +205,10 @@
 
 		if (jsonData !== false) {
 			// Trigger the file uploaded event.
-			this.trigger('fileUploaded', jsonData.uploadedFile);
+			// If form validation fails, response does not have uploadedFile
```
jonasraoni, line 208: Isn't it enough to move the code below to the outer scope and conditionally call the `this.handleJson()` (only if `jsonData.status` is `true`)?

```js
if (jsonData.content !== '') {
   this.replaceWith(jsonData.content);
}
```

jonasraoni (changes requested):
@jardakotesovec sorry for the delay, I've just reviewed.

controllers/wizard/fileUpload/FileUploadWizardHandler.php
jardakotesovec: Thats what I thought as well initially. But JSONMessage(false, 'error') these are errors that are supposed to be communicated via alert. But Form validations works differently, where you refetch the form so it can show updated (validation error state). If you search for `->validate()` you will see this pattern all over.

js/controllers/wizard/fileUpload/form/FileUploadFormHandler.js
jardakotesovec: handleJson is important also if status is false to display the error.

controllers/wizard/fileUpload/FileUploadWizardHandler.php
jonasraoni: Yep! I've just seen it, I'll change my decision to approved. I don't think it's worth to fix it properly, this code will eventually disappear.

js/controllers/wizard/fileUpload/form/FileUploadFormHandler.js
jonasraoni: I thought there was just this type of "HTML" errors on this form, but I see it has both 😁

Anyway, as you've pointed out, this pattern already exists, so no reason to come up with something new here.

jonasraoni (approved):

Issue #8365, opened 2022-10-24
Remove disable_path_info URL generation mode


asmecher:
The `disable_path_info` setting in `config.inc.php` has been marked as deprecated since 3.2.0. It was originally introduced to work around servers that did not properly support `PATH_INFO` URLs, but those seem to have dwindled over the past ~10 years. The `disable_path_info` setting is known to misbehave with API-based requests, but we haven't had much complaint about it, suggesting the mode is not widely used (or that users who require it are using very old versions).

With this mode deprecated for several years, it's time to remove it and simplify all the related code.

asmecher:
Once this is done, a logical next step would be to look into URL generation more broadly -- do better request-independent tools need solidifying? Can we delegate it to a Laravel-based toolset?

asmecher:
PRs:
- https://github.com/pkp/pkp-lib/pull/9761
- https://github.com/pkp/ojs/pull/4199

asmecher:
@bozana, could you review the PR [linked above](https://github.com/pkp/pkp-lib/issues/8365#issuecomment-1960098852)? The `disable_path_info` feature has been deprecated since 3.2.0, partially broken since at least 3.3.0, and removed entirely since 3.4.0; I think we're safe to remove the last bits and pieces for 3.5.0, but I'm not sure if you intended to keep especially long backwards compatibility in the log file parsing stuff?

asmecher:
Also, some of the log files couldn't be parsed in the upgrade tests -- but they don't appear to have path info disabled. Could you help me spot the problem there?

jonasraoni:
@asmecher / @bozana, the upgrade is failing due to the lines below, so I've created a dumb PR (https://github.com/pkp/pkp-lib/pull/9836), but I just read the comments and I see the issue might goes deeper than that. Are we going to recover the old code (regarding `disable_path_info`) just to make it works? I think it's good to leave a couple of "@todo" to remember about removing everything later.

https://github.com/pkp/pkp-lib/blob/69e7baad781699d6bbdbf65124c3a0c35db2c0fe/classes/cliTool/traits/ConvertLogFile.php#L323-L325

bozana:
Hi @jonasraoni, we will not recover the old code i.e. I only recovered it for log file conversion, for the users to be able to convert their old log files, also if they were suing disable_path_info. Thus, I think it should stay there as it is. Else, there are no other leftovers.
[...]

PR #9761, opened 2024-02-22
pkp/pkp-lib#8365 Remove vestiges of path info disabled mode support

asmecher:


classes/cliTool/traits/ConvertLogFile.php
```diff
@@ -69,12 +69,6 @@ abstract public function getParseRegex(): string;
      */
     abstract public function getPhpDateTimeFormat(): string;
 
-    /**
-     * Weather the URL parameters are used instead of CGI PATH_INFO.
-     * This will determine how URLs are parsed.
-     */
-    abstract public function isPathInfoDisabled(): bool;
-
```
bozana, line 77: Nothing should be removed from this file -- I extra added the functions here to be able to parse the old log files that used disabled path info once all the code is removed :-)

classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php
```diff
@@ -113,11 +112,6 @@ public function getPhpDateTimeFormat(): string
         return 'Y-m-d H:i:s';
     }
 
-    public function isPathInfoDisabled(): bool
```
bozana, line 116: Because this function is needed to be implemented -- it is an abstract function in ConvertLogFile, maybe to here only return false ?

tools/convertApacheAccessLogFile.php
```diff
@@ -70,14 +70,6 @@ class ConvertApacheAccessLogFile extends ConvertLogFileTool
...
     public const EGREP_PATH = '/bin/egrep';
 
-    /**
-     * Weather the URL parameters are used instead of CGI PATH_INFO.
-     * This is the former variable 'disable_path_info' in the config.inc.php
-     *
-     * This needs to be set to true if the URLs in the old log file contain the paramteres as URL query string.
-     */
-    public const PATH_INFO_DISABLED = false;
-
```
bozana, line 80: This file should also not be changed...

tools/convertUsageStatsLogFile.php
```diff
@@ -23,14 +23,6 @@
...
 class ConvertUsageStatsLogFile extends ConvertLogFileTool
 {
-    /**
-     * Weather the URL parameters are used instead of CGI PATH_INFO.
-     * This is the former variable 'disable_path_info' in the config.inc.php
-     *
-     * This needs to be set to true if the URLs in the old log file contain the paramteres as URL query string.
-     */
-    public const PATH_INFO_DISABLED = false;
-
```
bozana, line 33: And this one as well...

bozana (changes requested):
Hi @asmecher, the functions in ConvertLogFile, as well as other classes/files that extends it, should not be removed -- the old log files could have been created with disabled path info and we want to provide re-processing of those log files too (at least for some time).
Thus, I think only changes in Core and upgrade script are needed here.

asmecher:
Thanks for the review, @bozana! I adjusted as you suggested and merged.

PR #9766, opened 2024-02-28
Updated OAI stylesheet

jonasraoni:


jonasraoni:
@Vitaliy-1 as Alec is away, let me know if you're ok to review this. It's just an update to the OAI stylesheet, it's maintained by the [same entity](https://github.com/eprints/eprints3.4/blob/master/lib/static/oai2.xsl), so it should be safe.

jonasraoni:
@asmecher, here I'm just applying updates from upstream regarding the OAI stylesheet. If there's nothing against, I'll update the other branches.

asmecher:
@jonasraoni, sure, no objections.

asmecher:
@jonasraoni, I think this is ready for closing?

Issue #9776, opened 2024-03-07
Support format date string localization


defstat:
**Describe the bug**
For examples check #9303, #8768

**What application are you using?**
 OJS, OMP, OPS main branch

defstat:
@asmecher I have added the proposed PRs for that at the [initial comment](https://github.com/pkp/pkp-lib/issues/9776#issue-2173613326) of this issue and they are ready for review.

asmecher:
Reviewed, thanks, @defstat! https://github.com/pkp/pkp-lib/pull/9778#pullrequestreview-2126633243

asmecher:
@defstat, not a high priority, but please don't forget this one.

PR #9778, opened 2024-03-07
[pkp-lib][main] #9776 Support localization

defstat:


classes/services/PKPStatsServiceTrait.php
```diff
@@ -89,7 +89,7 @@ public function getEmptyTimelineIntervals(string $startDate, string $endDate, st
             $interval = 'P1M';
         } elseif ($timelineInterval === PKPStatisticsHelper::STATISTICS_DIMENSION_DAY) {
             $dateFormat = 'Y-m-d';
-            $labelFormat = PKPString::convertStrftimeFormat(Application::get()->getRequest()->getContext()->getLocalizedDateFormatLong());
+            $labelFormat =Application::get()->getRequest()->getContext()->getLocalizedDateFormatLong();
```
asmecher, line 92: @defstat, there's a missing space here, but the code auto-formatting tools should've taken care of it. In `pkp-lib`, we have [Captain Hook](https://github.com/captainhookphp/captainhook) installed to set up a git pre-commit hook; can you double-check that it's set up and working on your machine (e.g. on a personal fork)?

classes/core/PKPString.php
```diff
@@ -392,6 +397,127 @@ public static function getWordCount(string $str): int
...
+        return $locale;
+    }
+
+    /**
+     * Convert any datetime object to a localized string taking into account the 
+     * datetime format.
+     * 
+     * @throws \Exception General exception for potential errors during the date format
+     */
+    public static function getLocalizedDate(string|int|DateTimeInterface|null $date, string $format, ?string $locale = null): ?string
```
asmecher, line 488: @defstat, I'd consider this PR as-is, but part of our trouble with dates is that we throw around a lot of different kinds of dates without distinguishing clearly between them -- numeric timestamps and string representations, mainly. A side-effect of this is that I'm never sure what I'm dealing with, and any date processing ends up with a mess of conditionally called `strtotime` or `Carbon::parse` calls. Now that we're adopting typehints and able to use the `IntlDateFormatter`, I think we should standardize on `DateTimeImmutable`, and write functions that expect them consistently as inputs. Conversion to/from other formats should move close to the database layer (for stored values) and to the handler layer (for user inputs) and everything in between can use properly typed data. What do you think?

asmecher (changes requested):
Two comments to consider; thanks, @defstat!

Issue #9767, opened 2024-02-29
Offline App: Creating the PHP Fatal error


henriquejsfj:
See: https://github.com/pkp/pkp-lib/issues/7022

Seems that in version 3.4 this error persists. I tried the configuration said by @asmecher without success.

With setting `enable_beacon` and `show_upgrade_warning` to On in `config.inc.php` I got the following error:
```
PHP Fatal error:  Uncaught GuzzleHttp\\Exception\\ConnectException: cURL error 7: Failed to connect to pkp.sfu.ca port 443 after 1 ms: Couldn't connect to server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://pkp.sfu.ca/ojs/xml/ojs-version.xml?id=&oai=http%3A%2F%2Flocalhost%3A8081%2Findex%2Foai in /var/www/html/lib/pkp/lib/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:210\nStack trace:   [...] 
```
After setting `enable_beacon` and `show_upgrade_warning` to Off in `config.inc.php` I got the following error:
```
PHP Fatal error:  Uncaught GuzzleHttp\\Exception\\ConnectException: cURL error 7: Failed to connect to pkp.sfu.ca port 443 after 2 ms: Couldn't connect to server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://pkp.sfu.ca/ojs/xml/ojs-version.xml in /var/www/html/lib/pkp/lib/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:210\nStack trace: [...]
```

Note that after changing the configuration the requested URL does not have the OAI stuff anymore. 

I investigated it deeper and saw that in the [AdminHandler](https://github.com/pkp/pkp-lib/blob/8ce4daced2a4ab5e08eb95779df5e92d91951f00/pages/admin/AdminHandler.php#L122C9-L123C66) the function `VersionCheck::checkIfNewVersionExists()` is being called and VersionCheck only uses `enable_beacon` to add the OAI stuff in the URL.

A hotfix would move this function call to be inside the if with `show_upgrade_warning` right below.

Going even deeper, I investigate the stable 3.3 branch and that error would persist the same way after configuring...

jonasraoni:
Cherry-piked into `stable-3_3_0` and `main`.

PR #9779, opened 2024-03-07
#9767 Check if new version exists only if show_upgrade_warning is activated

henriquejsfj:
This is necessary to avoid Fatal Error if server has no internet connection.

#9767

### DIFF
--- a/classes/site/VersionCheck.php
+++ b/classes/site/VersionCheck.php
@@ -184,7 +184,12 @@ class VersionCheck
      */
     public static function checkIfNewVersionExists()
     {
-        $versionInfo = self::getLatestVersion();
+        try {
+            $versionInfo = self::getLatestVersion();
+        } catch (\GuzzleHttp\Exception\ConnectException $e) {
+            error_log('Failed to retrieve the latest version info: ' . $e->getMessage());
+            return false;
+        }
         $latestVersion = $versionInfo['release'];
 
         $currentVersion = self::getCurrentDBVersion();


pages/admin/AdminHandler.php
```diff
@@ -119,11 +119,11 @@ public function initialize($request)
             ]);
         }
 
-        // Interact with the beacon (if enabled) and determine if a new version exists
-        $latestVersion = VersionCheck::checkIfNewVersionExists();
-
         // Display a warning message if there is a new application version available
         if (Config::getVar('general', 'show_upgrade_warning') && $latestVersion) {
+            // Interact with the beacon (if enabled) and determine if a new version exists
+            $latestVersion = VersionCheck::checkIfNewVersionExists();
```
asmecher, line 125: This change will result in the version check never being run, and a warning, because `$latestVersion` (one line above) will never be defined.

henriquejsfj: Sorry about that, fixed it.

pages/admin/AdminHandler.php
```diff
@@ -119,11 +119,15 @@ public function initialize($request)
             ]);
         }
 
+        if (!Config::getVar('general', 'show_upgrade_warning')) {
+            return parent::initialize($request);
+        }
```
jonasraoni, line 124: @henriquejsfj the `VersionCheck::checkIfNewVersionExists()` must be always executed. So, the proper fix is to catch the `GuzzleHttp\Exception\ConnectException` on the call site.

pages/management/ManagementHandler.php
```diff
@@ -144,11 +144,13 @@ public function context($args, $request)
             ],
         ]);
 
+        $showUpgradeWarning = Config::getVar('general', 'show_upgrade_warning');
+
         // Interact with the beacon (if enabled) and determine if a new version exists
-        $latestVersion = VersionCheck::checkIfNewVersionExists();
+        if ($showUpgradeWarning) $latestVersion = VersionCheck::checkIfNewVersionExists();
```
jonasraoni, line 150: This line should also be always executed, for now the [Beacon](https://pkp.sfu.ca/software/ojs/usage-data/) depends on this, therefore disabling the upgrade warning shouldn't disable the Beacon.

pages/admin/AdminHandler.php
henriquejsfj: Obrigado @jonasraoni 
I was focused to make it work with setting the options, as the previous Issue solution, that I didn't realize that the proper fix is to really avoid throwing the fatal error hahah.

henriquejsfj:
I don't know if I used the proper way to log it, Laravel's `Log` didn't worked so I used `error_log` as I saw in other codes...
The output is like:
```
[Tue Mar 12 10:29:00.896581 2024] [php:notice] [pid 13] [client 172.19.0.1:50792] Failed to retrieve the latest version info: cURL error 7: Failed to connect to pkp.sfu.ca port 443 after 2 ms: Couldn't connect to server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://pkp.sfu.ca/ojs/xml/ojs-version.xml, referer: http://localhost/index/login/signIn
```
I would prefer to log it as a warning instead of notice...

jonasraoni (approved):
I think that's fine now 👌

jonasraoni:
Regarding the Laravel log, [there's a conversation to integrate it](https://github.com/pkp/pkp-lib/discussions/7966)... I'll restart the dialogue, I can't see any reasons to block it...

Issue #9658, opened 2024-01-26
Allow Journal Managers to invite users to adopt a role - Editors, Authors, Readers, etc


Devika008:
Hello,

Since [Allow journal managers to invite users to adopt a role](https://github.com/pkp/pkp-lib/issues/3022#top) was too big, upon discussion we decided to divide it into four smaller issues, the first one being this. 

This issue highlights only the user flow of journal managers inviting users like editors, authors, readers, etc to OJS. The user flow reflects all the research, discussions and feedback received on this issue over the last few years along with some minor testing with a small user group at the Copenhagen Sprint. 

**Summary of the requirement**

1. In the “Users & Roles” section of the settings, we add a new “Invite user to take a role” action that can be used to invite any user to the journal OJS software based on an email address. The email sent will add the following:

- Invitation request along with details of the role, starting date etc.
- ORCiD verification details
- GDPR compliance

2. Multiple hosted Journal OJS Installation: For Multiple Hosted Journals, Editors/ JMs cannot access information of users of one Journal to another. In spite of the journals being hosted on the same OJS instance, the user will have to go through the verification again to protect the information to be GDPR compliant. This is the same method followed by likes of Slack wherein even if the channel is hosted by the same company, the invitation process needs to be followed and accepted by the user. However, I as a user will not have to create a new account, I can use my existing OJS logins and verify and accept the role.

3. Other important considerations

- Even if the user information is publicly available, the JM / editor cannot input the information without consent. This will be translated in the journey in a way where the editors can input the information but they wont be stored in the system. They are shown to the user as suggestions when they accept the invite and the user approves or changes it. Post which the information is stored in the system
- Co-author reauthorization is required even if the user is an existing OJS user

### Below are the user flows for inviting users to a role in OJS

**A. Landing Page ( Settings > Users & Roles > Users)**
I have tweaked the interface to make it more accessible, instead of the arrows at the start of each row, I have added a "more options" icon at the end. I have shown three states of it below.

1. ![Image](https://github.com/pkp/pkp-lib/assets/112014397/3a3380d8-c9a4-4931-b673-c27bed59fd18)

2. ![Image](https://github.com/pkp/pkp-lib/assets/112014397/eb7c9e87-5ec0-49b9-89e6-894ac4e0475d)

3. ![Image](https://github.com/pkp/pkp-lib/assets/112014397/c3445971-a9c8-4509-ab4f-22ce7a23d5b2)

[...]

PR #9796, opened 2024-03-14
pkp/pkp-lib#9658 add get user api family name and given name #

ipula:


ewhanson (approved):
I don't have full context for if this would solve the underlying challenge in the issue referenced, but I think exposing the family and given names in the API summary are a good idea. 👍

withanage:
Thanks for looking @ewhanson

Issue #9789, opened 2024-03-11
System Information displays SMTP password as plain text.


bdcaf:
**Describe the bug**
System Information displays SMTP password as plain text - in comparison DB password is concealed. IMO this is similar sensitive and should be protected similarly.
This is not a deal breaker as the page is login protected, but does seem not too difficult to add either.

**To Reproduce**
1. Go to '/index/admin/systemInfo'
3. Scroll down to 'email section'
4. See password in plain text

**What application are you using?**
 OJS 3.4.0.5

**PRs**
#### Stable-3.3.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9805
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4217 [TEST ONLY]

#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9801
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4215 [TEST ONLY]

#### Main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9804
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4216 [TEST ONLY]

touhidurabir:
@asmecher can you review the PR at https://github.com/pkp/pkp-lib/issues/9789#issue-2178678595 . need forward porting to main and will do once test passes and confirmed.

asmecher:
Done, thanks: https://github.com/pkp/pkp-lib/pull/9801#pullrequestreview-1944003139

touhidurabir:
`pkp-lib` changes are merged . Need to do the submodule update later manually .

PR #9801, opened 2024-03-18
pkp/pkp-lib#9789 sensitive config data handle

touhidurabir:
for pkp/pkp-lib#9789

### DIFF
--- a/classes/config/Config.php
+++ b/classes/config/Config.php
@@ -47,13 +47,8 @@ class Config
 
     /**
      * Check and determine if the given section key is sensitive data or not
-     *
-     * @param string $section
-     * @param string $key
-     *
-     * @return bool
      */
-    public static function isSensitive($section, $key)
+    public static function isSensitive(string $section, string $key): bool
     {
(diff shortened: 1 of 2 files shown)


classes/config/Config.php
```diff
@@ -29,6 +29,39 @@
...
+
+    /**
+     * Check and determine if the given section key is sensitive data or not
+     *
+     * @param string $section
+     * @param string $key
+     *
+     * @return bool
+     */
+    public static function isSensitive($section, $key)
```
asmecher, line 56: For `stable-3_4_0` and `main`, use explicit typehints in the function declaration rather than PHPdoc-style comments (unless the comments add anything beyond the typehints, in which case use both). For porting back to `stable-3_3_0`, don't use typehints yet.

touhidurabir: Added. I know about this but intentionally did not add the type hint as none of the other method has it and seems odd to have only one method to have it . 

perhaps in one of other issue/PR we need to update the whole code base to have type hints to all places which possible using the [rector](https://github.com/rectorphp/rector) .

Issue #9806, opened 2024-03-19
CSS might be cached with invalid URLs


jonasraoni:
**Describe the bug**
When loading the dynamic generated CSS (e.g. `index.php/index/$$$call$$$/page/page/css`), a static cache is generated.

The cache contains URLs with domains, and this might lead to CORS issues when accessing the same page from another domain.

**To Reproduce**
1. Access the Administration menu using the domain/journal A
2. Access the Administration menu using the domain/journal B
3. See the errors on the console

**What application are you using?**
OJS 3.3

**Additional information**
![image](https://github.com/pkp/pkp-lib/assets/361921/f6219716-80b8-4278-8167-275e89b333da)

asmecher:
@jonasraoni, the PR looks OK, but before merging it, a sanity check -- isn't this a case where the same resources can be accessed via several URLs? Wouldn't it be better to e.g. adjust the rewriting rules so that only a single URL can be used for any resource?

jonasraoni:
Here's the flow:
1. `journal-a.domain` accesses the administrative panel
2. CSS is cached with the domain `journal-a.domain`
3. `journal-b.domain` acesses the administrative panel
4. Received cached CSS with fixed URLs (e.g. `journal-a.domain/stylesheet.css`)
5. Resource fails to be loaded due to CORS limitations

So, the other way would be to setup the server to provide the CORS permissions for such static resources, but I prefer to adopt the "it just works!" approach.

jonasraoni:
Cherry-picked at:
- https://github.com/pkp/pkp-lib/commit/ae4d9d509ab01efbaab3d9dba6ee265ea82be10d
- https://github.com/pkp/pkp-lib/commit/a2324e5bee6c03ebb7168f87f37fa39825bb050c

PR #9807, opened 2024-03-19
pkp/pkp-lib#9806 Included base URL to the CSS cache

jonasraoni:


classes/template/PKPTemplateManager.inc.php
```diff
@@ -453,7 +453,7 @@ public function getCachedLessFilePath($name) {
 		$cacheDirectory = CacheManager::getFileCachePath();
 		$context = $this->_request->getContext();
 		$contextId = is_a($context, 'Context') ? $context->getId() : 0;
-		return $cacheDirectory . DIRECTORY_SEPARATOR . $contextId . '-' . $name . '.css';
+		return $cacheDirectory . DIRECTORY_SEPARATOR . $contextId . '-' . $name . '-' . crc32( Application::get()->getRequest()->getBaseUrl($context)) . '.css';
```
jonasraoni, line 456: The `clearCssCache()` uses a `glob('*.css')`, so it's compatible.

asmecher: `PKPRequest::getBaseUrl` doesn't take a `Context` parameter.

jonasraoni: Thanks! That should be a true, but I'll just remove for simplicity :)

Issue #9803, opened 2024-03-18
Themes with multilingual options don't save and/or generate fatal errors


jnugent:
**Describe the bug**
This might be due to this being quite rare in themes, but we were building a theme for a client that required a theme option to be multi-lingual. The symptom was that the form was saving once, but never updating and retrieving what was in the database was not working as expected.

Creating a skeletal theme with a single option helped narrow down the problem.  If you start with an OJS installation with e.g. two languages enabled, and do this in a theme:

`
                $this->addOption('testInfo', 'FieldRichTextarea', [
                        'label' => __('plugins.themes.testPlugin.option.footerInfo.label'),
                        'description' => __('plugins.themes.testPlugin.option.footerInfo.description'),
                        'isMultilingual' => true,
                        'value' => $this->getOption('testInfo')
                ]);
`

You correctly get a multi lingual option on the Plugin settings form:

![Screenshot from 2024-03-18 14-30-54](https://github.com/pkp/pkp-lib/assets/218719/bfec8e71-d834-4e1e-865f-66b6019b5d83)

You can enter in text in both fields, no problem.  If you click save, the form data is sent as expected:

![Screenshot from 2024-03-18 14-29-07](https://github.com/pkp/pkp-lib/assets/218719/1d1d2058-6188-4984-9072-a1e089dec6b9)

And appears to be correctly saved in the database:

![Screenshot from 2024-03-18 14-41-59](https://github.com/pkp/pkp-lib/assets/218719/7f36f10d-071d-4181-8340-e3754ce01b41)

However, trying to reload the Website Settings page at this point to see if the form is populated again results in either no data in either language field, or (if you have strict warnings turned on), the following Fatal Error:

`
PHP Fatal error:  Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, string given in /var/www/html/www-33/lib/pkp/classes/components/forms/FormComponent.inc.php:296
`
Dumping the data prior to this line of code being called, It appears that the form field data is still in JSON format before it's passed in:

`
[Mon Mar 18 14:34:10.154713 2024] [php:notice] [pid 1478300] [client 127.0.0.1:54876] value is {"en_US":"<p>this is English.<\\/p>","fr_CA":"<p>c'est Francais.<\\/p>"}
`
We are pretty sure we're following the documentation for theme creation with form widgets like this, so this might be a bug as it may be relatively untested code.  I've not found another example of a theme doing this with a plugin setting.

[...]

PR #9814, opened 2024-03-21
pkp/pkp-lib#9803 Ensure fetched multilingual option is an array

jnugent:
This performs a quick sanity check to ensure that the pulled JSON string from the database is converted to an array before use.


jonasraoni:
I found out that things were a little bit more intertwined, so I've created another PR here: https://github.com/pkp/pkp-lib/pull/9847/files

PR #9843, opened 2024-03-28
pkp/pkp-lib#8365 use internal functions in ConvertLogFile that consid…

bozana:
…er disable_path_info

s. https://github.com/pkp/pkp-lib/issues/8365#id=issuecomment-2022472071


asmecher:
(good catch, thanks!)

Issue #9834, opened 2024-03-26
Statistics processing is stalled


jonasraoni:
**Describe the bug**
The issue https://github.com/pkp/pkp-lib/issues/9833, highlighted another issue (refer to the linked issue before reading the rest).

Problems:
1. The skipped log file is placed back to the staging folder, which means the system will attempt to process it again on the next run (infinite loop).
2. The presence of a single bad file affects the processing of other valid files, because of this condition:
https://github.com/pkp/pkp-lib/blob/11e269dc5bfa2188c8342e0ff15c8e287eee5670/classes/task/PKPUsageStatsLoader.php#L109-L112

If N log files are ok, those N files will be moved to the `dispatch` folder, and wait for a job to process them, but this never happens, due to the failing log file that makes the process quit.

Besides coming up with a fix, I think it's good to add extra log messages, this required some debugging.

**What application are you using?**
OJS 3.4

<hr>

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9844
- https://github.com/pkp/ojs/pull/4227 (only submodule update)
- https://github.com/pkp/omp/pull/1539 (only submodule update)
- https://github.com/pkp/ops/pull/652 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/9850

jonasraoni:
Hi @bozana! Not sure if you're too busy, if yes, I might take a look at this issue and the linked one.

bozana:
Hi @jonasraoni, yes, this is true. 
I think there is no good solution for #1 in your list above. The admin will need to see the scheduled task logs to find out the reason and solve the problem. It seems that now the log file is in the stage folder only if it is in the wrong format or if the stats for that month already exist (and only when the daily stats are not kept in the DB, so that the admin needs to re-process the whole month).
Eventually we could move that file into reject folder :thinking: Or maybe into a new/different folder to be able to differentiate those files from the other once? Hmmm... I will think about it...
And for the #2, yes, I think we can move that `if (!$processFilesResult)` at the end of the function (s. https://github.com/pkp/pkp-lib/blob/main/classes/task/PKPUsageStatsLoader.php#L136), together with `!$processingDirError`...

jonasraoni:
- **#<wbr>1:** I think it's ok, it's not a big deal to process it again. But it was difficult to understand what was happening without debugging the code. I saw the code has a lot of log messages, but I think we need to add more, to the remaining "escape" conditions.
[...]

PR #9844, opened 2024-03-28
pkp/pkp-lib9834 allow further usage log file processing for successfu…

bozana:
s. https://github.com/pkp/pkp-lib/issues/9834


jonasraoni (approved):
I think this should be enough to not block the processing of the other logs 👌

Issue #9833, opened 2024-03-26
Log files might be skipped by the statistics processing after migrating to OJS 3.4


jonasraoni:
**Describe the bug**
The `checkUsageStatsLogs()` method at the pre-flight migration is checking if there are old files by watching the modified date:
https://github.com/pkp/pkp-lib/blob/a006b3c882ede1d9ef99fca175f70fb2ef1efde7/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php#L208

But old files that were modified recently will skip the condition:
![image](https://github.com/pkp/pkp-lib/assets/361921/10de4904-1a1a-464b-adc8-a058b79b7f3f)

And create issues later on, at the migration `I8508_ConvertCurrentLogFile`, which will just attempt to convert files from today and yesterday (see snippet below), the other will be left in the folder with the old format.
https://github.com/pkp/pkp-lib/blob/a006b3c882ede1d9ef99fca175f70fb2ef1efde7/classes/migration/upgrade/v3_4_0/I8508_ConvertCurrentLogFile.php#L50-L63

**What application are you using?**
OJS 3.4

bozana:
Yes, I think <= would be correct

bozana:
Hi @jonasraoni, hmm... if ($date < $dateR5Installed) is correct, I think -- in this case the file is not valid i.e. in the old format that can not be parsed...

bozana:
@jonasraoni, can you please describe your case fully?
What log files were there when you were running the upgrade, what log files were converted, what is the counterR5StartDate, and maybe what do you see in the scheduled tasks log file for that file?

jonasraoni:
Sure!
1. The installation happened on day 19/03/2024, which is also the date that was stored on the `counterR5StartDate`.
2. A new log file was created at the same day, with the date 19/03/2024.
3. The `UsageStatsLoader` skips it, throw back to the staging folder (if I remember well), then it retry to process it again on the next run.

bozana:
Thanks @jonasraoni! Is that file still in the staging folder? What do you see in the scheduled task log file about that file?
It is probably processed on 20/03/2024 (or on the first date after that when the scheduled task is first run). So maybe you will find some information in the scheduled task log file scheduledTaskLogs/Usagestatisticsfileloadertask-...-20240320.log in the files folder. Or if it is still in the stage folder see the latest log file.
Thanks a lot!

jonasraoni:
Upgrade log:
```
[...]

PR #9846, opened 2024-03-29
pkp/pkp-lib#9833 Updated pre-flight script to extract the date from t…

jonasraoni:
…he filename instead of its modification date (can be changed if the files moved from one server to another)

### DIFF
--- a/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
+++ b/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
@@ -205,9 +205,12 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
         $usageStatsDir = StatisticsHelper::getUsageStatsDirPath();
         // check if there are usage stats log files older than yesterday
         foreach (glob($usageStatsDir . '/usageEventLogs/*') as $usageStatsLogFile) {
-            $lastModified = date('Ymd', filemtime($usageStatsLogFile));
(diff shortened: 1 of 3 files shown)


classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -205,7 +205,10 @@ protected function checkUsageStatsLogs(): void
         $usageStatsDir = StatisticsHelper::getUsageStatsDirPath();
         // check if there are usage stats log files older than yesterday
         foreach (glob($usageStatsDir . '/usageEventLogs/*') as $usageStatsLogFile) {
-            $lastModified = date('Ymd', filemtime($usageStatsLogFile));
+            if (!preg_match('/\d{8}\.log$/', $usageStatsLogFile, $lastModified)) {
+                throw new Exception("The log file \"{$usageStatsLogFile}\" doesn't follow the expected naming pattern \"usage_events_YearMonthDay.log\"");
+            }
+            [$lastModified] = $lastModified;
```
bozana, line 211: I think you need to use `'/(\d{8})\.log$/'` in preg_match above, and then `$lastModified = $lastModified[1]` right?

jonasraoni: Ops, I tagged you before reviewing and creating the PR for the tests 🙈
I'll change the name too 👌

PR #9847, opened 2024-03-29
Bugfix/stable 3 3 0/6186 support multilingual theme options

jonasraoni:


classes/plugins/ThemePlugin.inc.php
```diff
@@ -552,7 +552,7 @@ public function getOptionValues($contextId) {
 		foreach ($this->options as $optionName => $optionConfig) {
 			$value = isset($values[$optionName]) ? $values[$optionName] : null;
 			// Convert values stored in the db to the type of the default value
-			if (!is_null($optionConfig->default)) {
+			if ($value !== null && $optionConfig->default !== null) {
```
jonasraoni, line 555: The default value is used at the end of the `Field::getConfig()`, see:
https://github.com/pkp/pkp-lib/blob/d36f60ade0f546aeb570386deabb19ab4c282094/classes/components/forms/Field.inc.php#L125-L129

So, we shouldn't do coercions here, otherwise the field will have a non-null value and its default value will not be used later on.

classes/plugins/ThemePlugin.inc.php
```diff
@@ -571,6 +571,16 @@ public function getOptionValues($contextId) {
 						break;
 				}
 			}
+			// If the value isn't null and it's a multilingual field, then we must ensure it's an array
+			if ($optionConfig->isMultilingual && $value !== null && !is_array($value)) {
```
jonasraoni, line 575: This is what's missing to make the multilingual field works. Its content is stored as JSON, but only fields that have a `default` value (see `switch()` above) are being decoded.

jonasraoni:
Hi @ewhanson, as you've played around with theme, I'm tagging you for a brief review 😁

classes/plugins/ThemePlugin.inc.php
```diff
@@ -571,6 +571,16 @@ public function getOptionValues($contextId) {
...
+				try {
+					$value =  json_decode((string) $value, true, 512, JSON_THROW_ON_ERROR);
+				} catch (Exception $e) {
+					// FIXME: pkp/pkp-lib#6250 Remove after 3.3.x upgrade code is removed (see also pkp/pkp-lib#5772)
```
ewhanson, line 579: Just to confirm my understanding, this can safely be removed once upgrades can no longer go through the 3.3.x parts of the upgrade code?

jonasraoni: AFAICR, the upgrade code for 3.3 is using OJS classes to retrieve data, and that's why it's needed to keep this workaround.
From +3.4 onwards there shouldn't be `serialize()` content in the database anymore.

Issue #9813, opened 2024-03-21
Improve Editorial Activity calculations


bozana:
s. https://forum.pkp.sfu.ca/t/editorial-activity-statistics-look-strange-ojs-3-4-0-4/87021/13

- [x] Editorial Activity > Submissions In Progress checks after the old value '0' in the submission_progress column.
In 3.4 the values for submission_progress column have changed, s. https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php. 
However the code to get the number of submissions in progress for the editorial activity page still checks after the old value '0'. Consider here the new value '', but also the old value 0, because the QuickSubmit plugin still inserts old values.
- [x] Consider the new value '' in the submission_progress column here https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php#L390.
- [x] Use $this->getAcceptedDecisions() here https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/services/PKPStatsEditorialService.php#L65, in order to calculate the acceptance rate calculated correctly.

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9848
- https://github.com/pkp/ojs/pull/4241
- https://github.com/pkp/omp/pull/1551 (only submodule update)
- https://github.com/pkp/ops/pull/664 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/9907
- https://github.com/pkp/ojs/pull/4252

bozana:
@jonasraoni, could you please review the PR above? 
I am not 100% sure about using getAcceptedDecisions() instead of Decision::ACCEPT in the function PKPStatsEditorialService::getAverages(), but I think it is correct, because all the other numbers consider all the accept decisions from getAcceptedDecisions() and not only Decision::ACCEPT. Maybe you can also double think about it :-)
Thanks a lot!

bozana:
Hi @jonasraoni, could you take a look at the PR above?

jonasraoni:
This part of the code, with exceptions for the QuickSubmit, that doesn't align very well with the other types of data import (NativeXML) is a great source of confusion 😁

bozana:
Hi @jonasraoni, I changed it as said. Could you please take another look and the PRs above?
Thanks a lot!

PR #9848, opened 2024-03-29
pkp/pkp-lib#9813 fix bugs in editorial activity stats

bozana:
s. https://github.com/pkp/pkp-lib/issues/9813

### DIFF
--- a/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
+++ b/classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
@@ -387,7 +387,7 @@ abstract class PKPStatsEditorialQueryBuilder
         $q = $this->_getBaseQuery();
 
         // Exclude incomplete submissions
-        $q->where('s.submission_progress', '=', 0);
+        $q->where('s.submission_progress', '=', '');
 
         // Exclude submissions when the date_submitted is later
         // than the first date_published. This prevents imported
@@ -470,7 +470,7 @@ abstract class PKPStatsEditorialQueryBuilder
     public function countInProgress()
     {
         return $this->_getBaseQuery()
-            ->where('s.submission_progress', '<>', 0)
+            ->where('s.submission_progress', '<>', '')
             ->when($this->dateStart, fn (Builder $q) => $q->where('s.date_submitted', '>=', $this->dateStart))
             ->when($this->dateEnd, fn (Builder $q) => $q->where('s.date_submitted', '<=', $this->dateEnd))
             ->count();
--- a/classes/services/PKPStatsEditorialService.php
+++ b/classes/services/PKPStatsEditorialService.php
@@ -62,7 +62,7 @@ class PKPStatsEditorialService
             // within the date range that were accepted or declined. This
             // excludes submissions that were made within the date range but
             // have not yet been accepted or declined.
-            $acceptedForSubmissionDate = $this->countByDecisionsForSubmittedDate(Decision::ACCEPT, $args);
+            $acceptedForSubmissionDate = $this->countByDecisionsForSubmittedDate($this->getAcceptedDecisions(), $args);
             $declinedDeskForSubmissionDate = $this->countByDecisionsForSubmittedDate(Decision::INITIAL_DECLINE, $args);
             $declinedReviewForSubmissionDate = $this->countByDecisionsForSubmittedDate(Decision::DECLINE, $args);
             $totalDecidedForSubmissionDate = $acceptedForSubmissionDate + $declinedDeskForSubmissionDate + $declinedReviewForSubmissionDate;
@@ -228,7 +228,7 @@ class PKPStatsEditorialService
 
         // Editorial decisions (accepted and declined)
         $decisionsList = [
-            'submissionsAccepted' => [Decision::ACCEPT],
+            'submissionsAccepted' => $this->getAcceptedDecisions(),
             'submissionsDeclined' => [Decision::INITIAL_DECLINE, Decision::DECLINE],
             'submissionsDeclinedDeskReject' => [Decision::INITIAL_DECLINE],
             'submissionsDeclinedPostReview' => [Decision::DECLINE],


classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
```diff
@@ -387,7 +387,14 @@ protected function _getObject(): Builder
         $q = $this->_getBaseQuery();
 
         // Exclude incomplete submissions
-        $q->where('s.submission_progress', '=', 0);
+        // although the values for submission_progress column have changed
+        // (s. classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php)
+        // so that 0 should be ''
+        // still consider 0 because the QuickSubmitPlugin still inserts 0
+        $q->where(
+            fn (Builder $q) => $q->where('s.submission_progress', '=', '0')
```
jonasraoni, line 395: You can use `whereIn()` here and `whereNotIn()` on the `countInProgress()`.

classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
```diff
@@ -387,7 +387,14 @@ protected function _getObject(): Builder
         $q = $this->_getBaseQuery();
 
         // Exclude incomplete submissions
-        $q->where('s.submission_progress', '=', 0);
+        // although the values for submission_progress column have changed
+        // (s. classes/migration/upgrade/v3_4_0/I7191_SubmissionProgressType.php)
+        // so that 0 should be ''
+        // still consider 0 because the QuickSubmitPlugin still inserts 0
```
jonasraoni, line 393: I think other places that make use of the submission progress (I didn't inspect how far it goes), should be updated too.

Hmm, but instead of including workarounds due to issues on non-native plugins (e.g. the "custom" heuristic, based on the QuickSubmit plugin, that we've got to determine if a submission is imported doesn't satisfy other methods/plugins), I think it's better to find a solution with migrations (`submission_progress: 0 => '', 1 => 'start'`).

- A migration for 3.5 is fine, the plugin should be fixed by then.
- The migration for 3.4, it also doable, but as an outdated plugin might generate issues again, perhaps we can do a pre-flight check "Upgrade the QuickSubmit at least to the version X.Y.Z before upgrading" 👀

jonasraoni (changes requested):
@bozana, I've offered a suggestion to avoid adding the check for `0`.
Also I've created PRs for the QuickSubmit below, once they are merged, I think it's worth to release a new version 😁
- https://github.com/pkp/quickSubmit/pull/86
- https://github.com/pkp/quickSubmit/pull/87

classes/services/queryBuilders/PKPStatsEditorialQueryBuilder.php
bozana: @jonasraoni, do you then suggest not to consider 0 here?
Maybe actually the QuickSubmit should implement the upgrade script that will repair this, if there are no other places in the core code that needs to be fixed...

jonasraoni: Yeah, just to ensure the data is normalized.

classes/migration/upgrade/v3_4_0/I9813_QuickSubmitSubmissionProgressType.php
```diff
@@ -0,0 +1,56 @@
...
+ * @brief Fix the old submission_progress values inserted by QuickSubmit plugin,
+ *   from an int to a string to match the new step ids
+ */
+
+namespace PKP\migration\upgrade\v3_4_0;
+
+use Illuminate\Support\Facades\DB;
+use PKP\install\DowngradeNotSupportedException;
+
+class I9813_QuickSubmitSubmissionProgressType extends \PKP\migration\Migration
```
jonasraoni, line 21: @bozana, this migration is enough for the next release (3.5).
But for 3.4, I think it also needs to check if the user is using a broken `Quick Submit` plugin (version 1_0_7-1), and block the upgrade or it should at least emit an installation warning and/or a notification to the journal manager.

It's also important to release a new version of the `Quick Submit` with the latest fixes to stop the bleeding, but I don't have permissions to do that 😁

jonasraoni (changes requested):
Just one comment here.

classes/migration/upgrade/v3_4_0/I9813_QuickSubmitSubmissionProgressType.php
bozana: I think we can change plugin.xml so that 1_0_7-1 version is only available until 3.4.0.5.
Then, with the next release >= 3.4.0.6 everyone should be without this bug, using this upgrade script.
Correct?

Yes, I have already spoken with Alec -- he would do it, I just wanted to double check if anything else would need to be considered for this next release... I will check the issues and ask in dev channel...

bozana: I went through the open issues -- I do not think there is any needed to be considered immediately, so I will ask Alec to create a new release...

bozana: Hmmm... I see...
If the user updates to 3.4.0.6 it could still have and continue to use QuickSubmit plugin 1_0_7-1.
Hmmm... OK...
I will then throw an error, so that the user first needs to update the QuickSubmit plugin.
OK?

jonasraoni: It would be nice to have this as a `pre-flight check`, but the current pre-flight check won't run again at minor upgrades, I'll create a discussion to discuss this 🤔

These magic values should be linked against the database, then we wouldn't have such issues and it would be easier to understand what are the possible values (I have no idea without inspecting the code): https://github.com/pkp/pkp-lib/discussions/7981.

bozana: @jonasraoni, I have added the function that checks the plugin installed version and throws an exception.

classes/migration/upgrade/v3_4_0/I9813_QuickSubmitSubmissionProgressType.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    public function up(): void
+    {
+        $thresholdVersion = '1.0.7.1';
+        /** @var VersionDAO $versionDao */
+        $versionDao = DAORegistry::getDAO('VersionDAO');
+        $installedQuickSubmitPluginVersion = $versionDao->getCurrentVersion('plugins.importexport', 'quickSubmit');
+        if ($installedQuickSubmitPluginVersion->compare($thresholdVersion) === 0) {
```
bozana, line 31: because the script is executed only for OJS versions 3.4.0.0 - 3.4.0.5, it is OK to check for only this one plugin version, right? -- the older versions will not work with 3.4.x...

bozana:
@jonasraoni, would this be OK now?

classes/migration/upgrade/v3_4_0/I9813_QuickSubmitSubmissionProgressType.php
jonasraoni: Yes, AFAICS there's just one version of the QuickSubmit for 3.4.

jonasraoni, line 31: It's better to avoid using application classes in the migrations, but if you're comfortable with that, no objections.

jonasraoni (approved):
If Alec has no objections regarding raising an exception in a minor migration, I think everything is fine 👌

bozana:
Hi @jonasraoni, we discussed it on the dev call yesterday: it should not be an error, but just a warning that the user needs to upgrade the plugin immediately after the OJS upgrade. I have just changed that.
Because of that, I think we cannot assume that the user will do it for sure -- they could continue to use the wrong version, thus, the same upgrade script will be called for versions 3.4.0.0 - 3.4.9.9. Then I created a new issue for 3.4.1, to implement a preflight check that will check the plugin version installed, s. https://github.com/pkp/pkp-lib/issues/9889.

Do you think it is OK so now?

Thanks a lot!

jonasraoni:
So I think it's ok.

PR #9852, opened 2024-04-01
craft-oa-invitation-api Initial Commit

defstat:


api/v1/invitations/PKPInvitationController.php
```diff
@@ -0,0 +1,191 @@
...
+    
+        if (is_null($invitation)) {
+            return response()->json([
+                'error' => __('api.invitations.404.invitationNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        // Attempt to acceptHandle
+        try {
+            $invitation->acceptHandle();
```
asmecher, line 147: Are these accept/decline functions intended to replace the similar calls on the page handler?

classes/invitation/invitations/CustomInvite.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+/**
+ * @file invitation/invitations/CustomInvite.php
```
asmecher, line 4: Should be the full pathname; missing `classes/`

classes/invitation/invitations/CustomInvite.php
```diff
@@ -0,0 +1,71 @@
...
+ * @ingroup invitations
+ *
+ * @brief CustomInvite
+ */
+
+namespace PKP\invitation\invitations;
+
+use Illuminate\Mail\Mailable;
+
+class CustomInvite extends BaseInvitation
```
asmecher, line 21: Is this just an example, or will this class be used in production?

classes/invitation/invitations/BaseInvitation.php
```diff
@@ -252,4 +236,17 @@ public function getId(): int
     {
         return $this->id;
     }
+
+    /**
+     * Get constructor parameters for exclusion.
+     */
+    public function getExcludedUpdatePayloadVariables(): array
```
asmecher, line 243: Would it be cleaner to just keep the invitation subclass (what you're calling a business object) and model separate? We can implement the model code generically, and leave all invitation-specific concerns (both serialized data and functions) to the invitation subclass. Then I think we can let the functions on the subclass to have access to both, and let it get the right data from the right place. I think mixing the data together is going to lead us into confusion.

composer.lock
```diff
@@ -61,6 +61,75 @@
             ],
             "time": "2023-01-15T23:15:59+00:00"
         },
+        {
```
asmecher, line 64: Is the composer lockfile change intentional? I don't think it's needed here.

classes/invitation/invitations/CustomInvite.php
defstat: Just a showcase - not intended to include it in the final version of the PRs

classes/invitation/repositories/Invitation.php
```diff
@@ -189,4 +189,40 @@ public function addInvitation(BaseInvitation $invitationBO): int
 
         return $model->id;
     }
+
+    public function updatePayload(BaseInvitation $invitationBO, array $attributesToUpdate = []): ?BaseInvitation
```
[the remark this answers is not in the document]
defstat: Sure, I will remove it.

Issue #9849, opened 2024-03-31
Unable to create new autor's checklist-items when ALL former items are deleted


marcbria:
**Describe the bug**
If you delete all items from the author's check-list you are not able to create new items again.

**To Reproduce**
1. Go to /management/settings/workflow#submission/submissionChecklist
2. Remove all items and save.
3. Click on "New Item" to create a new one.
4. See a circle spinning for ever. 

**What application are you using?**
 OJS 3.3.0-17 (dockerized)

**Additional information**

Apache reports a 500 error:
```
[30/Mar/2024:23:30:47 +0000] "POST /myjournal/$$$call$$$/grid/settings/submission-checklist/submission-checklist-grid/update-item HTTP/1.1" 500 - "https://foo.org/myjournal/management/settings/workflow" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:124.0) Gecko/20100101 Firefox/124.0"
```

PHP reports the following at /var/log/apache2/error.log 

```
[...]

PR #9856, opened 2024-04-02
pkp/pkp-lib#9849 fixed issue determining checklist next id

touhidurabir:
for pkp/pkp-lib#9849


asmecher (approved):
@touhidurabir, I haven't tested this with an empty list (the problem reported in the issue), but it looks good to me -- please go ahead.

Issue #9857, opened 2024-04-02
Create FieldSlider component


jardakotesovec:
**Describe the feature**
For [Submission listing](https://github.com/pkp/pkp-lib/issues/7495) and [Review reminders](https://github.com/pkp/pkp-lib/issues/5885) we need form field slider with following features:

- field label
- custom labels for min/max values
- custom value label for min/max for current value
- min/max values

![image](https://github.com/pkp/pkp-lib/assets/1590681/66f6e3fa-2cac-4b8d-8271-fb6ae7d8b744)

jardakotesovec:
Added new dependency primevue. 

Landscape of headless libraries for Vue.js is evolving rapidly and hard to say whether one day there will be one which can handle all our needs. But for now bringing primevue, which now supports headless mode. 
- It has strong focus on accessibility. 
- It has small footprint. 
- Its easy to style with tailwindcss. 
- And it has very large collection of components - we might use some of them to replace some of our scattered dependencies (uploader, color field)..
- We might use date picker from primevue 

ojs: https://github.com/pkp/ojs/pull/4233
pkp-lib: https://github.com/pkp/pkp-lib/pull/9858/files
ui-library: https://github.com/pkp/ui-library/pull/343

@touhidurabir Finally have slider for you :-), could you please help with review? Thanks!

jardakotesovec:
@eirikhanssen Can I ask you for code review on this one as well? Thanks!

jardakotesovec:
merged

PR #9858, opened 2024-04-02
pkp/pkp-lib#9857 FieldSlider

jardakotesovec:


touhidurabir (changes requested):
Along with all these control option , I think we should also add the ability to define what is the range steps would be . It should not be a nullable and allow both int and float value .

touhidurabir (changes requested):
I think we should allow both int and float value here and also below for max value . There is no saying that it will always be integer so better to have that flexibility .

classes/components/forms/FieldSlider.php
```diff
@@ -0,0 +1,78 @@
...
+    public function getConfig()
+    {
+        $config = parent::getConfig();
+        $config['min'] = $this->min;
+        $config['max'] = $this->max;
+        $config['minLabel'] = $this->minLabel;
+        $config['maxLabel'] = $this->maxLabel;
+        $config['valueLabel'] = $this->valueLabel;
+        $config['valueLabelMin'] = $this->valueLabelMin;
+        $config['valueLabelMax'] = $this->valueLabelMax;
```
touhidurabir, line 73: Along with all these control option , I think we should also add the ability to define what is the range steps would be . It should not be a nullable and allow both int and float value .

classes/components/forms/FieldSlider.php
```diff
@@ -0,0 +1,78 @@
...
+class FieldSlider extends Field
+{
+    /** @copydoc Field::$component */
+    public $component = 'field-slider';
+
+
+    /**
+     * Range min value
+     */
+    public int $min;
```
touhidurabir, line 27: Along with all these control option , I think we should also add the ability to define what is the range steps would be . It should not be a nullable and allow both int and float value .

touhidurabir:
@jardakotesovec I think all ok for now as adding too much option right away make no sense unless we really need that . Also as this going into the main only, we have enough time to update/modify if necessary . Just few minor enhancements to address .

classes/components/forms/FieldSlider.php
jardakotesovec: @touhidurabir On the other hand - the improvements you are suggesting are small effort..  so not really problem include them. Will do the int/float & step property.

touhidurabir: As the value can be float and in such case steps can be float also . Right now we do no have a use case but it's not unnatural to have the slider range increase like [0.5, 1.0, 1.5, 2.0, 2.5, ...] .

jardakotesovec:
@touhidurabir Updated - can you double check it looks ok? Thanks!

touhidurabir:
@jardakotesovec all seems ok , good to merge as I see it .

PR #9862, opened 2024-04-03
Imports: Allow for a Native XML file to be imported if the primary contact ID is not provided

kennydude:
This pull request includes a fix for an edge case where importing native XML which does not contain a primary contact as it would throw a SQL integrity constraint error.


asmecher:
@defstat, could you please take a look at this one?

defstat (changes requested):
@kennydude thanks so much for your contribution. Just one minor syntax change and it is ready to be merged!

Issue #9837, opened 2024-03-27
you can access to worflow process even if a submission is incomplete


forgive38:
**Describe the bug**
If you enter the id of an incomplete submission directly into the URL: e.g. https://ojs34.testdrive.publicknowledgeproject.org/index.php/testdrive-journal/workflow/index/981/4 (here, 981 is an incomplete submission), you can access the editorial workflow and accept the submission, and so on.

So I think there should be a safety feature to prevent this.
**To Reproduce**
just change the id of the submission in the url

**What application are you using?**
 OJS 3.4 (test drive)

Thank you

**PRs**

#### Stable-3.3.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9869
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4238 [TEST ONLY]

#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10008
[ojs](https://github.com/pkp/ojs) -->  https://github.com/pkp/ojs/pull/4295 [TEST ONLY]

#### Main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10007
[ojs](https://github.com/pkp/ojs) -->  https://github.com/pkp/ojs/pull/4294 [TEST ONLY]

touhidurabir:
This issue also exists in 3.3.0 + .

touhidurabir:
@Vitaliy-1 can you take a look at the PR at https://github.com/pkp/pkp-lib/issues/9837#issue-2210378703 . If all ok, need to forward port it to 3.4.0 and main .

touhidurabir:
@Vitaliy-1 , ported to main and stable-3.4.0 . Also please see the comment https://github.com/pkp/pkp-lib/pull/9869#issuecomment-2139319509

touhidurabir:
All merged . will let the submodule get updated later .

PR #9869, opened 2024-04-08
pkp/pkp-lib#9837 Policy to restrict incomplete submission to access workflow

touhidurabir:
for pkp/pkp-lib#9837

### DIFF
--- a/classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
+++ b/classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * @file classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
+ *
+ * Copyright (c) 2014-2024 Simon Fraser University
+ * Copyright (c) 2000-2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class SubmissionCompletePolicy
+ * @ingroup security_authorization_internal
+ *
+ * @brief Class to control access to workflow only for complete submissions
+ *
+ */
+
+ import('lib.pkp.classes.security.authorization.DataObjectRequiredPolicy');
+
+class SubmissionCompletePolicy extends DataObjectRequiredPolicy {
+
+	/**
+	 * Constructor
+	 * 
+	 * @param PKPRequest 	$request 					The PKP core request object
+	 * @param array 		$args 						Request parameters
+	 * @param string 		$submissionParameterName 	The request parameter we expect the submission id in.
+	 * @param string 		$operation 					Optional list of operations for which this check takes effect. If specified, 
+	 * 													operations outside this set will not be checked against this policy
+	 */
+	function __construct($request, &$args, $submissionParameterName = 'submissionId', $operations = null) {
+		parent::__construct(
+			$request,
+			$args,
+			$submissionParameterName,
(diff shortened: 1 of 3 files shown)


classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
```diff
@@ -0,0 +1,64 @@
...
+	//
+	// Implement template methods from AuthorizationPolicy
+	//
+	/**
+	 * @see DataObjectRequiredPolicy::dataObjectEffect()
+	 */
+	public function dataObjectEffect() {
+		$submissionId = $this->getDataObjectId();
+
+		if ($submissionId === false) {
```
Vitaliy-1, line 51: I see that SubmissionRequiredPolicy exists with identical effect regarding existence of the submission ID. Also, as I recall when submission is incomplete, it already has an ID.

classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
```diff
@@ -0,0 +1,64 @@
...
+			return AUTHORIZATION_DENY;
+		}
+
+		$submissionDao = DAORegistry::getDAO("SubmissionDAO"); /** @var SubmissionDAO $submissionDao */
+		$submission = $submissionDao->getById($submissionId);
+
+		if ($submission->getData('submissionProgress') > 0) {
```
Vitaliy-1, line 58: We do need to allow access to authors, admins and journal managers to the incomplete submissions. Does it take that into account?

classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
touhidurabir: right , this check also present in the `SubmissionRequiredPolicy` . I added this as an extra safe check but we remove this as that checking belongs to `SubmissionRequiredPolicy` .

classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
touhidurabir: Yes , authors or Journal Managers/Editors can access the incomplete submission to complete it . but say an incomplete submission has id `10`. usually the workflow stages should not be accessible for that submission id . but if a user (who has access to workflow stages e.g. with JM/Editor/Admin role) put that submission id in workflow url like `http://ojs-stable-3_3_0.test/index.php/test-01/workflow/index/INCOMPLETE_SUBMISSION_ID/1` , they can initiate the workflow functionality for that incomplete submission .

Vitaliy-1: Makes sense!

Vitaliy-1 (approved):
Looks good! 
It may require additional tweaks for 3.4 and main. E.g., check if decisions like send to another stage on an incomplete submissions are prevented

classes/security/authorization/internal/SubmissionCompletePolicy.inc.php
Vitaliy-1: I meant to prevent access to submission wizard for incomplete submission for specific roles but it looks that this isn't needed

touhidurabir:
@Vitaliy-1 
> It may require additional tweaks for 3.4 and main. E.g., check if decisions like send to another stage on an incomplete submissions are prevented

Bit confused about it as if the access to the workflow stage is restricted , can there be any decisions to moved to any stage ? or I am misunderstanding something ?

PR #9878, opened 2024-04-11
Added missing strings in Weblate

germanhb:
Added six strings not used in later versions and therefore not available for translation in Weblate.


asmecher:
Thanks, @germanhb!

Issue #9879, opened 2024-04-11
Unused AuthorForm files


henriquejsfj:
**Describe the case**
The AuthorForm and its related files are marked as deprecated and unused in OJS 3.4, I'm not sure about OPS or OMS.

So, why don't we delete it? Keeping it made me spend some hours trying to access the Hooks.
(OrcidProfile Plugin keeps adding hooks to `authorform`)

Although keeping it causes no error, removing it causes no harm. (AFAIK)

asmecher:
This is required for QuickSubmit; see https://github.com/pkp/pkp-lib/pull/9880#issuecomment-2050363587. I'm eager to rewrite QuickSubmit just so we can get rid of all of this! However, closing for now.

PR #9880, opened 2024-04-11
#9879 Remove unused AuthorForm and its related files

henriquejsfj:
I added on top of branch `stable-3_4_0` since is the one I tested locally.
#9879


henriquejsfj:
If going forward with it, the AuthorForm from OJS must be removed too.

asmecher:
@henriquejsfj, I think this has been intentionally kept around for the sake of QuickSubmit, which still uses the old author grid. Removing it would break QuickSubmit.

Issue #9832, opened 2024-03-25
Enrolling an existing user as reviewer is broken


asmecher:
**Describe the bug**
From: https://forum.pkp.sfu.ca/t/php-fatal-error-when-trying-to-search-user-in-add-reviewer-enroll-an-existing-user-as-reviewer/87148

When inserting name in the search bar for Enroll an Existing User as Reviewer, php fatal error occurred and no result is display. only spinning wheel.

**Steps I took leading up to the issue**
1. Add reviewer
2. Enroll an Existing User as Reviewer
3. Search Name

What application are you using?
OJS 3.4.0.5, PHP 8.2, MariaDB 10.5

```
AH01071: Got error ‘PHP message: PHP Fatal error: Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column ‘u.user_id’ in ‘where clause’ in /home/xxxxxxx/public_html/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php:414\nStack trace:
n#0/home/xxxxxxx/public_html/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(414): PDO->prepare(‘select u.* fr…’)
n#1/home/xxxxxxx/public_html/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(753): Illuminate\Database\Connection->Illuminate\Database\{closure}(‘select u.* fr…’, Array)
n#2/home/xxxxxxx/public_html/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(720): Illuminate\Database\Connection->runQueryCallback(‘select u.* fr…’, Array, Object(Closure))
n#3/home/xxxxxxx/public_html/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(405): Illuminate\Database\Connection->run(‘select u.* fr…’…’, referer:
```

This is a regression caused by https://github.com/pkp/pkp-lib/issues/9541.

<ht>

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/9882
- https://github.com/pkp/ojs/pull/4243 (only submodule update)
- https://github.com/pkp/omp/pull/1543 (only submodule update)
- https://github.com/pkp/ops/pull/656 (only submodule update)

main:
[...]

PR #9882, opened 2024-04-12
pkp/pkp-lib#9832 Fix MariaDB SQL in enrolling an existing user as rev…

bozana:
s. https://github.com/pkp/pkp-lib/issues/9832

### DIFF
--- a/classes/user/Collector.php
+++ b/classes/user/Collector.php
@@ -444,35 +444,24 @@ class Collector implements CollectorInterface
             return $this;
         }
 
-        $subQuery = DB::table('user_user_groups as uug')
+        $userGroupsSubquery = fn (Builder $query) => $query->from('user_user_groups as uug')
             ->join('user_groups AS ug', 'uug.user_group_id', '=', 'ug.user_group_id')
             ->whereColumn('uug.user_id', '=', 'u.user_id')
-            ->when($this->userGroupIds !== null, fn (Builder $subQuery) => $subQuery->whereIn('uug.user_group_id', $this->userGroupIds))
+            ->when($this->userGroupIds !== null, fn (Builder $query) => $query->whereIn('uug.user_group_id', $this->userGroupIds))
             ->when(
                 $this->workflowStageIds !== null,
-                fn (Builder $subQuery) => $subQuery
+                fn (Builder $query) => $query
                     ->join('user_group_stage AS ugs', 'ug.user_group_id', '=', 'ugs.user_group_id')
                     ->whereIn('ugs.stage_id', $this->workflowStageIds)
             )
-            ->when($this->roleIds !== null, fn ($subQuery) => $subQuery->whereIn('ug.role_id', $this->roleIds))
-            ->when($this->contextIds !== null, fn ($subQuery) => $subQuery->whereIn('ug.context_id', $this->contextIds));
+            ->when($this->roleIds !== null, fn ($query) => $query->whereIn('ug.role_id', $this->roleIds))
+            ->when($this->contextIds !== null, fn ($query) => $query->whereIn('ug.context_id', $this->contextIds));
 
-        $joinSub = clone $subQuery;
-        $query
-            ->addWhereExistsQuery(
-                $subQuery->when(
-                    $this->excludeRoles != null,
-                    // This aggregates a column role_count, which holds an information whether user holds specified role
-                    fn (Builder $subQuery) => $subQuery->leftJoinSub(
-                        $joinSub
-                            ->select('uug.user_id')
-                            ->whereIn('ug.role_id', $this->excludeRoles),
-                        'agr',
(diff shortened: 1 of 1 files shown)


classes/user/Collector.php
```diff
@@ -458,19 +458,29 @@ protected function buildUserGroupFilter(Builder $query): self
...
-                    fn (Builder $subQuery) => $subQuery->leftJoinSub(
-                        $joinSub
-                            ->select('uug.user_id')
-                            ->whereIn('ug.role_id', $this->excludeRoles),
-                        'agr',
-                        'uug.user_id',
-                        'agr.user_id'
-                    )->whereNull('agr.user_id')
+                    fn (Builder $subQuery) =>
+                        DB::connection()->getDoctrineConnection()->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\MariaDBPlatform
```
jonasraoni, line 467: If the version for MariaDB works with MySQL, I'd vote to leave just this one and avoid checking the database version.

classes/user/Collector.php
```diff
@@ -458,19 +458,29 @@ protected function buildUserGroupFilter(Builder $query): self
...
+                        )
+                        :
+                        $subQuery->leftJoinSub(
+                            $joinSub
+                                ->select('uug.user_id')
+                                ->whereIn('ug.role_id', $this->excludeRoles),
+                            'agr',
+                            'uug.user_id',
+                            'agr.user_id'
+                        )->whereNull('agr.user_id')
```
jonasraoni, line 483: Hmm, as far as I can see, the cloned query `$joinSub`, isn't making explicit use of the outer query `$subQuery`, it's using just the `user_id` field, which is available at the main table.

So, I think we can move the `excludeRoles` stuff to the outer scope (`$query`), something like: `$query->notExists(query to select the excluded roles)`.

jonasraoni (changes requested):
Please, check if what I wrote makes sense, anyway, this problem is probably happening elsewhere. Only tests will save us 😁

classes/user/Collector.php
bozana: OK, yes, it works, the other one is just a little bit better regarding the performance, I believe...
(Eventually MariaDB will allow the outer references in derived tables one day too...)

classes/user/Collector.php
bozana: Hmmm... Yes...
So instead of having:
```
select u.* from users as u
where u.disabled = 0 and exists 
(
select * from user_user_groups as uug
inner join user_groups as ug on uug.user_group_id = ug.user_group_id 
where uug.user_id = u.user_id and ug.context_id in (1) and u.user_id not IN 
(
select uug.user_id from user_user_groups as uug
inner join user_groups as ug on uug.user_group_id = ug.user_group_id 
where uug.user_id = u.user_id and ug.context_id in (1) and ug.role_id in (4096)
)
) 
order by u.user_id asc;
```

it would be:
```
select "u".* from "users" as "u" 
where "u"."disabled" = 0 and exists 
(
select * from "user_user_groups" as "uug" 
inner join "user_groups" as "ug" on "uug"."user_group_id" = "ug"."user_group_id" 
where "uug"."user_id" = "u"."user_id" and "ug"."context_id" in (1)
) 
and not exists 
(
select * from "user_user_groups" as "uug" 
inner join "user_groups" as "ug" on "uug"."user_group_id" = "ug"."user_group_id" 
where "uug"."user_id" = "u"."user_id" and "ug"."context_id" in (1) and "ug"."role_id" in (4096)
) 
order by "u"."user_id" asc
```

Which seems to be better...

classes/user/Collector.php
```diff
@@ -457,22 +457,14 @@ protected function buildUserGroupFilter(Builder $query): self
...
-                    )->whereNull('agr.user_id')
-                )
-            );
+        $query->addWhereExistsQuery($subQuery);
+
+        if ($this->excludeRoles != null) {
+            $excludeRolesSubQuery = clone $subQuery;
+            $excludeRolesSubQuery->whereIn('ug.role_id', $this->excludeRoles);
+            $query->whereRaw("not exists({$excludeRolesSubQuery->toSql()})")
+                ->mergeBindings($excludeRolesSubQuery);
```
jonasraoni, line 466: I think you can avoid the `whereRaw()` if you move the `$subQuery` to a function.
```php
$getUserGroupsQuery = fn (Builder $q) => $q->from('user_user_groups as uug')->...;
$subQuery = $getUserGroupsQuery(DB::query());
...
$query->whereNotExists(fn (Builder $q) => $getUserGroupsQuery($q)->...)
```

bozana:
@jonasraoni, could you please take another look? -- I have changed the code to use a function, as you suggested...
Thanks a lot!

jonasraoni (approved):

Issue #9825, opened 2024-03-22
Fetch submission' title for the reviewer assignments list


Vitaliy-1:
Reviewer dashboard should include not only data related to the assignments but also submission title. Submission could be added to the review assignments schema the same way as it's done for the review/stage assignments in the submission schema, see https://github.com/pkp/pkp-lib/issues/8885

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/9908
tests: https://github.com/pkp/ojs/pull/4253

Vitaliy-1:
@jonasraoni, can you take a look at this PR, which includes fetching submissions into review assignments list in a performant way?

jardakotesovec:
Hi @Vitaliy-1 ,
just connecting this to the UI. And I realised that when we are showing submission metadata - we have slightly different fallbacking strategy there.

If meta are not available in the 'current UI language', it fallbacks to the publication locale and than it would fallback to the journal primary language. 

To be able to keep consistent with this logic - I would also need to know the publication locale - could we add that one as well?

jardakotesovec:
@Vitaliy-1 Currently getting following payload from `/publicknowledge/api/v1/_submissions/reviewerAssignments?offset=0&count=30&page=1&perPage=30`

```
{
            "_href": "http:\/\/localhost:7001\/index.php\/_\/api\/v1\/_submissions\/reviewAssignments\/6",
            "cancelled": false,
            "comments": null,
            "competingInterests": null,
            "considered": null,
            "dateAcknowledged": null,
            "dateAssigned": "2024-05-23 05:22:12",
            "dateCancelled": null,
            "dateCompleted": null,
            "dateConfirmed": null,
            "dateDue": "2024-06-20 00:00:00",
            "dateNotified": "2024-05-23 05:22:12",
            "dateRated": null,
            "dateReminded": null,
            "dateResponseDue": "2024-06-20 00:00:00",
            "declined": false,
            "id": 6,
            "lastModified": "2024-05-23 05:22:12",
            "publicationTitle": {
                "en": "Genetic transformation of forest trees",
                "fr_CA": ""
            },
            "quality": null,
            "recommendation": null,
            "reminderWasAutomatic": false,
            "requestResent": false,
            "reviewFormId": null,
            "reviewMethod": 2,
            "reviewRoundId": 4,
            "reviewerFullName": "Paul Hudson",
            "reviewerId": 8,
            "round": 1,
[...]

PR #9908, opened 2024-04-23
pkp/pkp-lib#9825 Submission data in review assignment schema

Vitaliy-1:


classes/submission/Collector.php
```diff
@@ -626,6 +638,18 @@ public function getQueryBuilder(): Builder
...
+        $q->when(
+            $this->reviewIds !== null,
+            fn (Builder $q) => $q
+                ->whereIn(
+                    's.submission_id',
+                    fn (Builder $q) => $q
+                        ->select('ra.submission_id')
+                        ->from('review_assignments AS ra')
+                        ->whereIn('ra.review_id', $this->reviewIds)
+                )
```
jonasraoni, line 650: I think it's better to use `exists()` with the outer field (`s.submission_id`).

classes/submission/reviewAssignment/maps/Schema.php
```diff
@@ -52,9 +55,21 @@ public function summarize(ReviewAssignment $item): array
     public function mapMany(Enumerable $collection): Enumerable
     {
         $this->collection = $collection;
-        return $collection->map(function ($item) {
-            return $this->map($item);
-        });
+        $this->submissions = Repo::submission()->getCollector()
+            ->filterByContextIds([$this->context->getId()])
```
jonasraoni, line 59: I didn't inspect how this code is used, but the `$this->context` is nullable, so it's better to add some protection either here or on the constructor (i.e. force it to not be null).

classes/submission/reviewAssignment/maps/Schema.php
```diff
@@ -65,19 +80,34 @@ public function mapMany(Enumerable $collection): Enumerable
...
+            fn ($item) =>
+            $this->summarize($item, $this->submissions->get($associatedSubmissions->get($item->getId())))
+        );
     }
 
     /**
      * Map schema properties of an Announcement to an assoc array
      */
-    protected function mapByProperties(array $props, ReviewAssignment $item): array
+    protected function mapByProperties(array $props, ReviewAssignment $item, Submission $submission): array
```
jonasraoni, line 103: This code isn't used anywhere? I didn't see updates to the calling sites.

classes/submission/reviewAssignment/maps/Schema.php
```diff
@@ -24,24 +25,26 @@ class Schema extends \PKP\core\maps\Schema
 
     public string $schema = PKPSchemaService::SCHEMA_REVIEW_ASSIGNMENT;
 
+    public Enumerable $submissions;
```
jonasraoni, line 28: AFAICS this property isn't shared nor reused in later calls, so I think it's better to demote it to a local variable.

jonasraoni (changes requested):
@Vitaliy-1 I just left one advice and some questions.

classes/submission/Collector.php
jonasraoni: Yeah, if the number of retrieved records is small, I don't think it's going to change much, but I think it's generally better to use the `exists()`.

classes/submission/reviewAssignment/maps/Schema.php
Vitaliy-1: Hmm, I see this pattern is used in other schemas, e.g., submission. I'm going to leave this as is - I'm not sure if we should allow context wild card here.

jonasraoni: I've just suggested to add defensive code, because the parent class (`PKP\core\maps\Schema`) allows a null context to be passed, otherwise an exception might happen.

```php
->filterByContextIds($this->context ? [$this->context->getId()] : [])
```

p.s.: There's a protection to not load data for all contexts, if that's what you mean with "context wildcard". You've got to pass `Application::CONTEXT_ID_ALL` which translates to `_` (🤨).

Vitaliy-1: I'll open another issue

Vitaliy-1: https://github.com/pkp/pkp-lib/issues/9944

Issue #9895, opened 2024-04-18
Introduce APP KEY feature of Laravel


touhidurabir:
**Describe the Proposal**
Introduce the Laravel like APP KEY feature/behaviour for next release of OJS/OMP/OPS . This will be a required details need to set which will be required at the app bootstrapping . 

**Additional Information**
With the merge of [Convert session and cookie management to Laravel](https://github.com/pkp/pkp-lib/issues/9566), we have the ability to make the cookie encrypted and that feature added . However we have limited chiper support for the cookie encryption because we do not have full featured app key behaviour implemented . By introducing the app key we can have bring in the full encryption support of laravel.

However this will not only have impact for cookie encryption but for any encryption and auth token generation process and we can leverage the native laravel built in functionality for that purpose . Also this does not impact with any hashing functionality e.g. password hashing . 

**Possible Implementation flow**
1. A config option named `app_key`
2. Command line tool to generate/update app key
3. config option to on/off cookie encryption
4. config option to set cipher settings e.g. aes-128-cbc/aes-256-cbc/aes-128-gcm/aes-256-gcm
5. keeping the app key optional of app bootstrapping process but required for encryption/decryption process . 
6. upgrade migration to set the app key in `app_key` variable present in the `config.inc.php` file, it will not throw error but log error in missing or failed . 

**Impacted systems**
OJS/OMP/OPS main(3.5.0 pre release)

**PRs **
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9918
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4257
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1575
[ops](https://github.com/touhidurabir/ops) --> https://github.com/pkp/ops/pull/690

touhidurabir:
@asmecher can you review the PRs at https://github.com/pkp/pkp-lib/issues/9895#issue-2249932770

asmecher:
Reviewed: https://github.com/pkp/ojs/pull/4257#pullrequestreview-2072664307

Once that's ready, please port to the other apps as well.

asmecher:
Thanks, @touhidurabir! I've added some reviews:
- https://github.com/pkp/pkp-lib/pull/9918#pullrequestreview-2089030304
- https://github.com/pkp/omp/pull/1575#pullrequestreview-2088983021
[...]

PR #9918, opened 2024-04-28
pkp/pkp-lib#9895 app key and encryption service provider attached to container

touhidurabir:
for pkp/pkp-lib#9895

### DIFF
--- a/classes/cliTool/UpgradeTool.php
+++ b/classes/cliTool/UpgradeTool.php
@@ -20,6 +20,7 @@ namespace PKP\cliTool;
 
 use APP\core\Application;
 use APP\install\Upgrade;
+use PKP\core\PKPAppKey;
 use PKP\site\VersionCheck;
 
 Application::upgrade();
@@ -75,6 +76,7 @@ class UpgradeTool extends \PKP\cliTool\CommandLineTool
     public function check()
     {
         $this->checkVersion(VersionCheck::getLatestVersion());
+        $this->checkForAppKey();
     }
 
     /**
@@ -207,6 +209,21 @@ class UpgradeTool extends \PKP\cliTool\CommandLineTool
         }
 
         return $compare1;
+    }
+
+    /**
+     * Check the existence of `app_key` variable in config file and print warning message if not found
+     */
+    public function checkForAppKey(): void
+    {
+        // if the app key variable `app_key` set in the config, nothing to do
+        if (PKPAppKey::hasKeyVariable()) {
+            return;
+        }
+        
+        printf("\n\e[;43mWARNING: It is noticed that there is not `app_key` variable defined in the `general` section of the config file which is necessary to cookie and other encryption purpose.\nWe suggest add the following line in the `general` section of config file.\e[0m\n\n");
+        
+        printf("\e[;44mapp_key = \e[0m\n");
     }
 
     /**
--- a/classes/cliTool/UpgradeTool.php
+++ b/classes/cliTool/UpgradeTool.php
@@ -76,7 +76,6 @@ class UpgradeTool extends \PKP\cliTool\CommandLineTool
     public function check()
     {
         $this->checkVersion(VersionCheck::getLatestVersion());
-        $this->checkForAppKey();
     }
 
     /**
@@ -209,21 +208,6 @@ class UpgradeTool extends \PKP\cliTool\CommandLineTool
         }
 
         return $compare1;
-    }
-
-    /**
-     * Check the existence of `app_key` variable in config file and print warning message if not found
-     */
-    public function checkForAppKey(): void
-    {
-        // if the app key variable `app_key` set in the config, nothing to do
-        if (PKPAppKey::hasKeyVariable()) {
-            return;
-        }
-        
-        printf("\n\e[;43mWARNING: It is noticed that there is not `app_key` variable defined in the `general` section of the config file which is necessary to cookie and other encryption purpose.\nWe suggest add the following line in the `general` section of config file.\e[0m\n\n");
-        
-        printf("\e[;44mapp_key = \e[0m\n");
     }
 
     /**
--- a/classes/config/ConfigParser.php
+++ b/classes/config/ConfigParser.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/config/ConfigParser.php
  *
- * Copyright (c) 2024 Simon Fraser University
- * Copyright (c) 2024 John Willinsky
+ * Copyright (c) 2014-2024 Simon Fraser University
+ * Copyright (c) 2014-2024 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class ConfigParser
--- a/classes/config/ConfigParser.php
+++ b/classes/config/ConfigParser.php
@@ -3,13 +3,11 @@
 /**
  * @file classes/config/ConfigParser.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class ConfigParser
- *
- * @ingroup config
  *
  * @brief Class for parsing and modifying php.ini style configuration files.
  */
@@ -18,15 +16,10 @@ namespace PKP\config;
 
 class ConfigParser
 {
-    /** @var string Contents of the config file currently being parsed */
-    public $content;
-
     /**
-     * Constructor.
+     * Contents of the config file currently being parsed
      */
-    public function __construct()
-    {
-    }
+    protected string $content;
 
     /**
      * Read a configuration file into a multidimensional array.
@@ -36,7 +29,7 @@ class ConfigParser
      *
      * @return bool|array the configuration data (same format as http://php.net/parse_ini_file)
      */
-    public static function &readConfig($file)
+    public static function &readConfig(string $file): bool|array
     {
         $configData = [];
         $currentSection = false;
@@ -129,7 +122,7 @@ class ConfigParser
      *
      * @return bool true if file could be read, false otherwise
      */
-    public function updateConfig($file, $params)
+    public function updateConfig(string $file, array $params): bool
     {
         if (!file_exists($file) || !is_readable($file)) {
             return false;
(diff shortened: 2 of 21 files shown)


classes/core/PKPEncryptionServiceProvider.php
```diff
@@ -0,0 +1,40 @@
...
+        // if application not installed, we do not check for app key as
+        // may be it's need to be generated at application installation process
+        if (!Application::isInstalled()) {
+            return;
+        }
+
+        // if no app key set, the encrypter can not be registered
+        if (!PKPAppKey::hasKey()) {
+            return;
+        }
```
touhidurabir, line 36: This is to allow the application to boot and run if `app_key` not set and we don't want to make it a must requirement for app to run . We can remove it if we decide to it make a must .

classes/install/Installer.php
```diff
@@ -1008,6 +1010,34 @@ public function checkPhpVersion()
...
+        if (PKPAppKey::hasKey()) {
+            return true;
+        }
+
+        // will set an error if app key variable not set
+        // but will not halt the process
+        if (!PKPAppKey::hasKeyVariable()) {
+            error_log("No key variable named `app_key` defined in the `general` section of config file. Please update the config file's general section and add line `app_key = `");
+            return true;
+        }
```
touhidurabir, line 1031: If `app_key` variable not defined, we are not halting the process as not making the app key a must but if we decide to make it a must, we can set error and return false to halt the process .

classes/install/Installer.php
```diff
@@ -1008,6 +1010,34 @@ public function checkPhpVersion()
...
+
+        try {
+            PKPAppKey::writeToConfig(PKPAppKey::generate());
+        } catch (Throwable $exception) {
+            error_log($exception->getMessage());
+        } finally {
+            return true;
+        }
+    }
```
touhidurabir, line 1040: If `app_key` writing to config file we are not halting the process as not making the app key a must but if we decide to make it a must, we can set error and return false to halt the process .

classes/cliTool/UpgradeTool.php
```diff
@@ -209,6 +211,21 @@ public function checkVersion($versionInfo, $displayInfo = false)
...
+    {
+        // if the app key variable `app_key` set in the config, nothing to do
+        if (PKPAppKey::hasKeyVariable()) {
+            return;
+        }
+        
+        printf("\n\e[;43mWARNING: It is noticed that there is not `app_key` variable defined in the `general` section of the config file which is necessary to cookie and other encryption purpose.\nWe suggest add the following line in the `general` section of config file.\e[0m\n\n");
+        
+        printf("\e[;44mapp_key = \e[0m\n");
+    }
```
touhidurabir, line 227: on `upgrade check` if 
1. `app_key` variable not defined in the config file, just returning but we can add another warning here . 
2. if app key not set, showing a warning . 

if we decided to make the app key a must, we can turn it into error/exception .

asmecher (changes requested):
I've suggested a change in the OJS part of this that'll affect this PR too; otherwise looks good!

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,185 @@
...
+     * The default cipher algorithms
+     *
+     * @var string
+     */
+    private static string $defaultCipher = 'aes-256-cbc';
+
+    /**
+     * Get the list of supported ciphers
+     */
+    public static function getSupportedCiphers(): array
```
asmecher, line 49: I don't think the getter function adds value; in general I'd like to decrease our use of getters/setters (and eventually [property hooks](https://wiki.php.net/rfc/property-hooks) will remove more justification for these). I don't have a strong opinion about static variables vs. constants, but we do have an existing pattern for `public const MY_CONST_NAME = abc` in classes, and the `UPPER_SNAKE_CASE` convention does make the invariability clear.

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,185 @@
...
+    {
+        return Config::getVar('security', 'cipher', self::$defaultCipher);
+    }
+
+    /**
+     * Has the app key defined in config file
+     */
+    public static function hasKey(): bool
+    {
+        return !empty(Config::getVar('general', 'app_key', ''));
```
asmecher, line 67: You can leave the 3rd parameter blank if you don't want to specify a default; also below

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Get the defined cipher
+     */
+    public static function getCipher(): string
+    {
+        return Config::getVar('security', 'cipher', self::$defaultCipher);
+    }
+
+    /**
+     * Has the app key defined in config file
```
asmecher, line 63: `Is` rather than `Has` (also below)

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        if (!$configParser->updateConfig(Config::getConfigFileName(), $configParams)) {
+            // Error reading config file
+            throw new Exception('Unable to read the config file');
+        }
+
+        if (!$configParser->writeConfig(Config::getConfigFileName())) {
+            // Error writing config file
+            throw new Exception('Unable to write the app key in the config file');
```
asmecher, line 168: Since hosts might for legitimate reasons make the config file read-only to the web server, it would be nice to provide a way for admins to manually add the new line of configuration rather than having this be a fatal error. For example, if we successfully generate a key but it can't be written to the configuration file, the upgrade script should instruct the admin what to add to the config file (including the generated key).

That message should also be localized. (It's tough to find a balance between throwing a hard-coded exception with English text vs. localizing the text, but my rule of thumb is: if it's a message that a user might encounter in the normal operation of the system, then it should be localized; if it's a system failure or coding problem, then it's OK to leave it as a hard-coded exception or error. Because admins who are running the system properly might encounter this message, it should be localized.)

classes/core/PKPEncryptionServiceProvider.php
asmecher: Because we're adding the app key automatically both on install and upgrade, I see no downsides to making the app key a requirement right now. It'll become necessary later as we adopt more of Laravel, so we may as well add the requirement now, since it's low burden to the admin. It'll keep our code simpler.

classes/migration/upgrade/v3_5_0/I9895_AddAppKeyToConfigFile.php
```diff
@@ -0,0 +1,57 @@
...
+    public function up(): void
+    {
+        // if APP KEY already exists, nothing to do
+        if (PKPAppKey::hasKey()) {
+            return;
+        }
+
+        // will set an error if app key variable not set
+        // but will not halt the process
+        if (!PKPAppKey::hasKeyVariable()) {
```
asmecher, line 36: I think this should add the variable if it's not present; there's no benefit to making the admin go through it manually.

classes/migration/upgrade/v3_5_0/I9895_AddAppKeyToConfigFile.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Reverse the downgrades
+     *
+     * @throws DowngradeNotSupportedException
+     */
+    public function down(): void
+    {
+        throw new DowngradeNotSupportedException();
```
asmecher, line 55: I suggest we leave the `down()` function empty rather than throwing an exception. There's no harm in leaving the variable in the config file, and if the upgrade is run again, it will preserve the existing value.

classes/migration/upgrade/v3_5_0/I9895_AddAppKeyToConfigFile.php
```diff
@@ -0,0 +1,57 @@
...
+
+        try {
+            PKPAppKey::writeToConfig(PKPAppKey::generate());
+        } catch (Throwable $exception) {
+            error_log($exception->getMessage());
+        }
+    }
+
+    /**
+     * Reverse the downgrades
```
asmecher, line 49: `Reverses the migration`

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,185 @@
...
+        $configParser = new ConfigParser;
+        $configParams = [
+            'general' => [
+                'app_key' => $key,
+            ],
+        ];
+
+        if (!static::hasKeyVariable()) {
+            // Error if the config key `app_key` not defined under `general` section
+            throw new Exception('Config variable named `app_key` not defined in the `general` section');
```
asmecher, line 158: I think this is redundant with https://github.com/pkp/pkp-lib/pull/9918/files#diff-8ae64e2e6552e29dc508801b292812f26856a2d5e80147614e8b39d48e29d906R36 (I have a suggestion on that line already).

asmecher (changes requested):
Some suggestions for change! The biggest shift from the last review is that I think we should pick a side on adding the app key vs. keeping it optional. We're already adding it on upgrade and new install, so I think we should make it mandatory and reduce the redundant checks and messages.

classes/core/PKPAppKey.php
touhidurabir: I Think we can keep this one as these are 2 different method (one to set/write the key variable and one to set/write the key itself) which can work independent of each other .

classes/core/PKPAppKey.php
asmecher: (minor, but still needs attention)

touhidurabir:
@asmecher all tests are passing . Ready for a final look and if all ok, good to merge .

classes/config/ConfigParser.php
```diff
@@ -3,30 +3,23 @@
 /**
  * @file classes/config/ConfigParser.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2024 Simon Fraser University
```
asmecher, line 6: Not a big deal, but don't remove old dates when they're present -- just update the latest date if it's not current.

classes/core/PKPAppKey.php
```diff
@@ -0,0 +1,237 @@
...
+            Encrypter::generateKey(static::validateCipher($cipher ?? $config['cipher']))
+        );
+    }
+
+    /**
+     * Write the `app_key` variable in the `config.inc.php` file
+     *
+     * @throws \Exception
+     */
+    public static function writeAppKeyVaribaleToConfig(): bool
```
asmecher, line 140: `Varibale` is a typo: should be `Variable`. (Here and elsewhere.)

asmecher (changes requested):

Issue #9453, opened 2023-10-23
In OJS, a reviewer cannot access his/her recommendations from a previous round.


pilasou:
As a reviewer reinvited to a new review round, I would like to be able the see the evaluation I provided in previous rounds. 

In OJS, reviewers that are being invited to different review rounds on the same submission cannot access what was their previous recmmandations.

This improvement will offer the opportunity to reviewers to access their previous review reports and thus, better determine how the author made his/her changes to the manuscript. 

The reviewer will be able to access his/her Previous round reports by clicking on a Round button on top of the reviewer process page (as shown below). 
![image](https://github.com/pkp/pkp-lib/assets/19291880/b721acf1-f5ec-4a4b-9be8-d4fff9c64e0a)

When clicking on a cycle button a new window open containing the following information:  

- Recommandation 
- Forms informations 
- Review file sent 
- Cycle dates 
- Article version reviewed during preceding round

![image](https://github.com/pkp/pkp-lib/assets/19291880/cd4b5df2-59d9-4197-bce5-e73aaf053fd9)

### PRs:

Update (2024-04-02)
[pkp/ui-library#313](https://github.com/pkp/ui-library/pull/341) (jardakotesovec repository)
[pkp/pkp-lib#9526](https://github.com/pkp/pkp-lib/pull/9526)
[pkp/ojs#4148](https://github.com/pkp/ojs/pull/4148)

Initial: 
[#9526](https://github.com/pkp/pkp-lib/pull/9526)

**OMP adaptation PRs:**
OJS: pkp/ojs#4258
OMP: pkp/omp#1556
OPS: pkp/ops#668 ***TESTS ONLY***

PKP-LIB: #9920 ***Initial Commits for PKP-LIB included***

### How to test?

1. Make a submission and send it to the Review step.
2. Invite one reviewer.
3. Connect as reviewer and complete the review (fill the form and add a file).
4. As editor, Request revision and select Revisions will be subject to a new round of peer reviews.
5. As an author, submit revision.
6. As editor, open a 2nd round of review.
7. Invite the 1st round reviewer.
8. Connect as the reviewer.
9. On top of the review page, you should see the Round 1 bouton.
10. In the new window, the reviewer can access its answers (forms and files).

asmecher:
PR: https://github.com/pkp/pkp-lib/pull/9526

@Vitaliy-1, could you review this? Thanks!

pilasou:
[...]

PR #9920, opened 2024-04-29
[PKP-LIB][main] #9453 Reviewer History

defstat:


api/v1/reviews/PKPReviewController.php
```diff
@@ -0,0 +1,225 @@
...
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+        $illuminateRequest = $args[0]; /** @var \Illuminate\Http\Request $illuminateRequest */
+        $actionName = static::getRouteActionName($illuminateRequest);
+
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
+        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+
+        if ($actionName === 'getHistory') {
```
asmecher, line 84: `getHistory` is the only function offered by this controller, right? It might be better to just remove this test entirely and always add the policy. If a later function gets added to the handler, it'll probably need the same policy -- and as it stands, the policy wouldn't be applied.

defstat: You are right. Removed.

Issue #9957, opened 2024-05-13
Author Reviewer Grid: the grid fails to be displayed.


defstat:
**Describe the bug**
In the Author's workflow->review stage of a given submission, the Reviewer Grid fails to be displayed.

**To Reproduce**
1. Go to a submission with an Open Peer Review reviewer assigned to it (the reviewer must have accepted the review assignment)
2. Login as the Author of that submission
3. Go to that submission
4. See the error.

![image](https://github.com/pkp/pkp-lib/assets/560456/32c4d774-58d2-46f4-995c-9a93b00a94f9)

**What application are you using?**
 OJS `main` branch

**Additional information**
The underling error is `Fatal error: Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given` and it is referring to the `actions` template variable of the `AuthorReviewerGridCellProvider`.

---
**PRs**

OJS: pkp/ojs#4675 **TESTS ONLY**
PKP-LIB: #10953

Vitaliy-1:
@defstat, while testing this I noticed  another problem - unable to see any files under that grid. My steps:

1. Go to the submission in the review stage
2. Assign a reviewer with an open review mode
3. Perform a review and attach a review file during 3rd step. Ensuring that it's safe in the database with a file stage `SUBMISSION_FILE_REVIEW_ATTACHMENT`
4. Finish the review
5. Log in as an author
6. Seeing that the Reviewer's Attachments grid is empty

I can reproduce the bug described here by assigning a reviewer and declining this request. But I'm confused. I don't remember this part of the code at all. So, under that grid all files at the stage `SUBMISSION_FILE_REVIEW_ATTACHMENT` should be shown if the review is open, right? But author is able to read the review and see the file through that form, so it's duplication of the functionality?

jardakotesovec:
@defstat Also run into that - and was not aware of this issue. https://github.com/pkp/pkp-lib/issues/10453

I will keep it opened as well as it offers alternative solution.

What I propose as fix that display correctly the review assignment also if its not submitted yet. As result author (for open review) basically sees exactly the same thing as editors, because from the code that seemed to be the intention.

@Vitaliy-1 @Devika008 @asmecher  Any opinion on this?

Context is that when the review is opened one - we provide similar reviewer listing to author as the editor can see.  Just with option to Read the review and no other actions.   Which means that the author see whole process of reviewers are being requested, and they responded etc.  

[...]

PR #9958, opened 2024-05-13
[PKP-LIB][main] #9957 Add isOpen filter to ReviewAssignment

defstat:


Vitaliy-1 (commented):
I don't have any other comments except those mentioned in the issue thread. This can be merged and with other stuff we can dealt later.

defstat:
Closed because the fix is not relevant any more for the `main` branch.

Issue #9968, opened 2024-05-20
Bring in Laravel's Blade Templating Engine.


touhidurabir:
R&D on a proof of concept to see if it is possible to bring in and integrate the Laravel's blade templating engine into the core system . 

### Usage Doc
 - By default all the views in pkp share lib and app level are namespaced but it will work with/without namespace . For example , `@include('VIEW_NAMESPACE::some-template')` or `just @include('some-template')` . However without specifying the namespace , the app level view will have higher priority when exists . 
 - Also by default all view components in pkp share lib and app level are namespaced but it will work with/without namespace . For example, `<x-COMPONENT_NAMESPACE::some-component />` or just `<x-some-component />`. However without specifying the namespace , the app level view will have higher priority when exists . 
 - The app level view namespace is set to `app` and shared lib pkp level view namespace is set as `pkp`. so for view it can work as `app::some-template` or `x-app::some-component`.
 - both the `.blade` and `.blade.php` extension for blade view files are supported . 
 - The default path of component classes are `classes/view/components/` and corresponding views are `classes/view/components/` for both the app and shared lib pkp level .  
 - for plugins , the default path of component classes are `classes/components/` and corresponding views are `templates/components/`.
 - Plugins views will be auto namespaced and need to work with view namespace to render the blade views . by default, for each plugin , the view namespace will be default to it's name but can be overridden by having the method `public function getTemplateViewNamespace(): string` define for each plugin . 
 - for plugins , a variable `$viewNamespace` which hold the view namespace value of that plugins will be auto injected for each blade view except the class based component views. so plugin can use that variable for include partial view like `@include("{$viewNamespace}::frontend.components.header")` . 
 - plugins must use view namespace to use view components as one of following manner 
 ```blade
{{-- Directly use the view namespace in the component usage--}}
<x-citationstylelanguageplugin::citation-style 
        :style-id="$citationStyle['id']"
        :title="$citationStyle['title']"
        :params="$citationArgs"
        :paramsJson="$citationArgsJson"
    />
```
```blade
{{-- Or use as dynamic component only where it's possible to use the `$viewNamespace` in a dynamic way--}}
<x-dynamic-component
    :component="$viewNamespace . '::citation-style'"
    :style-id="$citationStyle['id']"
    :title="$citationStyle['title']"
    :params="$citationArgs"
    :paramsJson="$citationArgsJson"
/>
```
[...]

PR #9969, opened 2024-05-20
pkp/pkp-lib#9968 blade templating engine integration R&D

touhidurabir:


classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -0,0 +1,267 @@
...
+        // use as following in a blade view
+        // "@url([
+        //     'page' => 'citationstylelanguage',
+        //     'op' => 'get',
+        //     'path' => $citationStyle['id'],
+        //     'params' => $citationArgsJson
+        // ])"
+        Blade::directive('url', function ($expression) {
+            return "<?php
+                \$parameters = $expression ? (array) ($expression) : [];
```
asmecher: The use of `$expression` twice here will result in the array being constructed twice, which is probably not what we want. (Here and below.) Since we always expect an array to be passed in, let's just remove the test and cast.

asmecher (changes requested):
Hi @touhidurabir! Thanks -- just one suggestion here.

classes/core/PKPBladeViewServiceProvider.php
touhidurabir: Not sure I understand the concern here properly, for blade directive , anything passed through it will be a string . basically when we define a custom Blade directive, the function we provide receives the exact string content that is placed inside the directive's parentheses. In this case, we cast it as array as our target method need array . we can do it one single line but seems bit hard to read and get what it does so I break it down like this .

asmecher: If the Blade template contains something like
```
@some-directive(['publicationId' => Repo::submission()->get($submissionId)->getData('currentPublicationId')])
```
...then in the compiled template you'll get (formatted for readability)...
```php
$parameters = ['publicationId' => Repo::submission()->get($submissionId)->getData('currentPublicationId')] ?
    (array) ['publicationId' => Repo::submission()->get($submissionId)->getData('currentPublicationId')] : [];
```
That is to say, the entire array will be constructed twice in the normal case where it evaluates `true`. It's wasteful, and could cause side-effects.

I suggest instead that we just treat it like an array; if it isn't, we'll have a fatal error on syntax (or type -- see below), which is appropriate. (Also, the left-hand side of the ternary operator is not bracketed, which could also cause confusion in case a non-array expression is passed in -- but let's just remove it rather than adding brackets.)

So:
```php
Blade::directive('url', function ($expression) {
    return "<?php echo \\PKP\\template\\PKPTemplateManager::getManager()->smartyUrl($expression); ?>";
});
```
For added points, add typehints to the `smartyUrl` function so that it knows to expect an `array` for `$parameters`.

touhidurabir: Right , I am only thinking about what has been passed not what will be compiled to before render .

PR #9970, opened 2024-05-20
Log reviewer response

MrRob100:


controllers/grid/users/reviewer/ReviewerGridRow.php
```diff
@@ -215,6 +216,66 @@ public function initialize($request, $template = null)
                     )
                 );
             }
+            if ($reviewAssignment->getDateConfirmed() == null) {
+                $context = $request->getContext();
+                $vueModalArgs = [
+                    'description' => $submission->getCurrentPublication()->getLocalizedTitle(null, 'html'),
+                    'logResponseForm' => [
+                        'id' => 'logResponseForm',
```
jardakotesovec, line 224: We do have classes to create form structure. Example would be https://github.com/pkp/pkp-lib/blob/main/classes/components/forms/dashboard/SubmissionFilters.php  or any other form in that forms folder.

controllers/grid/users/reviewer/ReviewerGridRow.php
```diff
@@ -215,6 +216,66 @@ public function initialize($request, $template = null)
...
+                        'method' => 'POST',
+                        'action' => $router->url($request, null, null, 'addLog', null, $actionArgs),
+                        'fields' => [
```
jardakotesovec, line 227: Is not this opportunity to create rest API endpoint - instead of using these grid legacy endpoints? 

Basically everytime we replace something with Vue.js which needs to interact with backend - we also create rest api endpoint, which is generally useful for interacting with OJS.

Issue #9946, opened 2024-05-08
Do not remove numbers from the beginning of each citation


bozana:
Remove this part of the code https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/citation/CitationListTokenizerFilter.php#L59-L62, that removes numbers from the beginning of each citation.
S. https://forum.pkp.sfu.ca/t/if-a-reference-list-item-starts-with-a-numeral-it-is-removed-3-1-2-4-and-3-3-0-15/87845.
This way the correct citations that start with a number will be correctly reprocessed and displayed. For the others, that this part of the code tries to solve: actually the editor is responsible for entering the correct citations and not for example a numbered list.

PRs:
stable-3_3_0:
- https://github.com/pkp/pkp-lib/pull/9983
- https://github.com/pkp/ojs/pull/4288 (only submodule update)
- https://github.com/pkp/omp/pull/1573 (only submodule update)
- https://github.com/pkp/ops/pull/688 (only submodule update)

stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/10002
- https://github.com/pkp/ojs/pull/4291 (only submodule update)
- https://github.com/pkp/omp/pull/1577 (only submodule update)
- https://github.com/pkp/ops/pull/692 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/10003
- https://github.com/pkp/ojs/pull/4292 (only submodule update)
- https://github.com/pkp/omp/pull/1578 (only submodule update)
- https://github.com/pkp/ops/pull/693 (only submodule update)

bozana:
Hi @jonasraoni, could you please take a look at the PR above?

PR #9980, opened 2024-05-22
Do not remove numbers from the beginning of each citation:

GaziYucel:
* https://github.com/pkp/pkp-lib/issues/9946
* https://forum.pkp.sfu.ca/t/if-a-reference-list-item-starts-with-a-numeral-it-is-removed-3-1-2-4-and-3-3-0-15

### DIFF
--- a/classes/citation/CitationListTokenizerFilter.inc.php
+++ b/classes/citation/CitationListTokenizerFilter.inc.php
@@ -47,10 +47,6 @@ class CitationListTokenizerFilter extends Filter {
 		} else {
 			$citations = explode("\n", $input);
 		}
-		// 4) Remove numbers from the beginning of each citation.
-		foreach($citations as $index => $citation) {
-			$citations[$index] = PKPString::regexp_replace('/^\s*[\[#]?[0-9]+[.)\]]?\s*/', '', $citation);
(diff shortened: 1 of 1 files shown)


bozana:
Hi @GaziYucel, 

Would you maybe like to also provide the submodule update PRs (pointing to these changes) for OJS, OMP and OPS? -- We need them in order for the tests to be run, but also for stable branches, so that the user can update properly/accordingly.
If it is too much for you, I can do this -- I would take your commit and add the submodule updates. Just let me know please.
Thanks a lot!

GaziYucel:
Hi @bozana,
Can you please do this for this PR. I have never done that before, I will find out how and will do this for my next PRs. 
Thanks.

GaziYucel:
Hi @bozana, another question. I did this now for version 3.3.0. Shall I do this also for the versions 3.4.0 and 3.5.0? There is no corresponding issue for these.

bozana:
Hi @GaziYucel, I will then cherry-pick it to the other branches now. Else, yes please, always for all higher/newer branches as well :-)
Thanks a lot!

jonasraoni:
Closing as it was cherry-picked.

Issue #9899, opened 2024-04-19
Add unit test for each queue jobs class


touhidurabir:
**Describe the problem**
Starting from `3.4.0` we have queue jobs to handle long running tasks or for tasks that need to done later . Our queue jobs implementation is an extended implementation of Laravel queue jobs.

The problem arise or will arise when any of the jobs structure will update but there is one or more pending jobs with older structure already dispatched in queue waiting to get executed . This problem get more complex with the upgrade of newer version when older structure based job still in queue as they may continue to fail and no way to resolve it after upgrade which may lead to undesirable outcome. 

**Describe the proposal**
The initial plan was to force to run all pending queue jobs before upgrade which proposed at [Add pre-flight check for pending jobs before running upgrade](https://github.com/pkp/pkp-lib/issues/8707) . This was a very simple and easy approach but may not suited for all possible cases . Also a job class structure or handle implementation may need to update based on requirement.

After some detailed discussion, we decided to add unit tests for each job class to make sure the consistence update the job class structure and handle method implementation . The idea is to try to mock each of jobs class to make sure it's on even after any update to it.

**Impacted version**
 OJS, OMP or OPS version 3.4.0 and main

**Additional information**
Details discussion at [Right approach to handle pending queue jobs at upgrade](https://github.com/pkp/pkp-lib/discussions/9868)

**Job classes to have unit tests**
##### OJS
- [x] APP\jobs\doi\DepositIssue
- [x] APP\jobs\notifications\IssuePublishedNotifyUsers
- [x] APP\jobs\notifications\OpenAccessMailUsers
- [x] APP\jobs\statistics\CompileCounterSubmissionDailyMetrics
- [x] APP\jobs\statistics\CompileCounterSubmissionInstitutionDailyMetrics
- [x] APP\jobs\statistics\CompileIssueMetrics
- [x] APP\jobs\statistics\CompileSubmissionGeoDailyMetrics
- [x] APP\jobs\statistics\CompileUniqueInvestigations
- [x] APP\jobs\statistics\CompileUniqueRequests
- [ ] APP\jobs\statistics\CompileUsageStatsFromTemporaryRecords [deprecated 3.4.0.5]
- [x] APP\jobs\statistics\DeleteUsageStatsTemporaryRecords
- [x] APP\jobs\statistics\ProcessUsageStatsLogFile
##### OMP
- [x] APP\jobs\statistics\CompileCounterSubmissionDailyMetrics
[...]

PR #9981, opened 2024-05-23
pkp/pkp-lib#9899 proper class path added for Core class

touhidurabir:


tests/jobs/submissions/UpdateSubmissionSearchJobTest.php
asmecher, line 1: It looks like there's still this one file classified as a binary.

asmecher: It looks like the file was emptied but not removed, at a glance.

touhidurabir: @asmecher I emptied out this one to see how it behaves . Also deleting this file and adding a new one with same name marked that binary again , seems like reinitiate the deleted file . Having a different file name does resolve this . Checking the file encoding from terminal shows as `text/x-php; charset=us-ascii`, does not make any sense why git still marking it as binary .

asmecher: @touhidurabir, I'm happy to write this off as a Github quirk. I don't see anything unusual about the file, at a glance.

PR #9983, opened 2024-05-24
 Do not remove numbers from the beginning of each citation:

bozana:
s. https://github.com/pkp/pkp-lib/issues/9946

s. https://github.com/pkp/pkp-lib/pull/9980


bozana:
@jonasraoni, could you please take a look at this PR?
Thanks a lot!

jonasraoni (approved):
@bozana I agree with the changes, we shouldn't drop information behind the curtains.

If we decide to process the citations in a better way, given that some formats (e.g. JATS?!) require more specific information, then we should have a couple of fields and a maybe a parser to setup those fields.

bozana:
Thanks a lot @jonasraoni! TIB will be working on a better citations, their parsing and formatting...

PR #10008, opened 2024-06-03
pkp/pkp-lib#9837 Policy to restrict incomplete submission to access workflow

touhidurabir:
pkp/pkp-lib#9837


touhidurabir:
@Vitaliy-1 please check the PR . It consider the API decisions also for incomplete submission .

PR #10022, opened 2024-06-06
Check phpunit version for RouterTest

withanage:


withanage:
Hi @asmecher ,

This PR chcecks for phpunit version for  version 3.3 , php errors

withanage:
#### Related
- https://github.com/pkp/ops/pull/695
- https://github.com/pkp/ops/pull/695

asmecher (commented):
Since we specify the phpunit version in Composer, I don't think we need to support multiple versions.

withanage:
@asmecher  thanks! closing without merged,  cause it can be solved  by setting php ini values explicitly for phpunit

PR #10023, opened 2024-06-07
Feature main 8333 add remaining fks

jonasraoni:


classes/core/PKPPageRouter.php
```diff
@@ -31,8 +31,8 @@ class PKPPageRouter extends PKPRouter
     /** @var array pages that don't need an installed system to be displayed */
     public $_installationPages = ['install', 'help', 'header', 'sidebar'];
 
-    public const ROUTER_DEFAULT_PAGE = './pages/index/index.php';
-    public const ROUTER_DEFAULT_OP = 'index';
+    public const ROUTER_DEFAULT_PAGE = './pages/' . Application::SITE_CONTEXT_PATH . '/index.php';
```
asmecher, line 34: I think the `SITE_CONTEXT_PATH` constant is at risk of being used for too many different things. At the least, the use of URLs and file paths shouldn't be mixed, and in cases like this I sometimes find that repeating the string is more readable/maintainable than burying it into a pile of pseudo-globals in `PKPApplication`.

If the purpose of the constant is to reserve the `index` context name, then it shouldn't be mixed with the default op (below) or the file path (this line).

classes/log/EmailLogEntry.php
```diff
@@ -127,39 +126,29 @@ public function setAssocId($assocId)
...
-            ? Repo::user()->get($this->getSenderId(), true)
-            : null;
+        return $this->sender ??= $this->getSenderId() ? Repo::user()->get($this->getSenderId(), true)?->getFullName() : null;
+    }
 
-        return $sender ? $sender->getFullName() : '';
+    /**
+     * Return the full name of the sender (not necessarily the same as the from address).
+     */
+    public function getSenderFullName(): string
```
asmecher, line 139: In cases like this, I'd rather permit a `null` return to indicate that the data is missing. That means data presence/absence stays type-based (`string` vs `null`) rather than value-based (`'something'` vs `''`), which is conceptually more clear.

If it should always be the case that a value is returned, then having this function throw a `TypeError` or similar is probably not a bad thing.

classes/log/SubmissionEmailLogDAO.php
```diff
@@ -69,27 +69,26 @@ public function getBySubmissionId($submissionId)
...
      * Get submission email log entries by submission ID, event type and sender ID
      *
-     * @param int $submissionId
      * @param int $eventType SubmissionEmailLogEntry::SUBMISSION_EMAIL_*
      * @param int $senderId Return only emails sent by this user.
      *
      * @return DAOResultFactory<SubmissionEmailLogEntry>
      */
-    public function getBySenderId($submissionId, $eventType, $senderId)
+    public function getBySenderId(int $submissionId, int $eventType, int $senderId): DAOResultFactory
```
asmecher, line 77: These changes are probably going to conflict with @taslangraham's work over on https://github.com/pkp/pkp-lib/issues/10129. I'd suggest leaving the cleanups on `SubmissionEmailLogDAO` et al out of this PR.

classes/middleware/SetupContextBasedOnRequestUrl.php
```diff
@@ -29,12 +29,9 @@ class SetupContextBasedOnRequestUrl
      *
      * @example
      *      1. admin section (failed job list) : index.php/index/api/v1/jobs/all
-     *      2. admin section (create context) : index.php/_/api/v1/contexts
```
asmecher, line 32: Are you sure the `_` in URLs is appropriately removed/changed? I don't recall this part in detail, but I do recall there being special consideration for `_` in the URL distinct from the context path. It's probably not a good idea to make a breaking change in API URLs without a lot of thought/consideration first!

classes/migration/upgrade/v3_5_0/I8333_AddMissingForeignKeys.php
```diff
@@ -0,0 +1,253 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_5_0/I8333_AddMissingForeignKeys.php
+ *
+ * Copyright (c) 2023 Simon Fraser University
```
asmecher, line 6: `2024`

classes/migration/upgrade/v3_5_0/I8333_AddMissingForeignKeys.php
```diff
@@ -0,0 +1,253 @@
...
+            ->where('parent_id', '=', 0)
+            ->update(['parent_id' => null]);
+        DB::table('email_log')
+            ->where('sender_id', '=', 0)
+            ->update(['sender_id' => null]);
+        // Only the administrator user group (role_id = 1) is allowed to have a null context_id
+        DB::table('user_groups')
+            ->where('role_id', '=', 1)
+            ->where('context_id', '=', 0)
+            ->update(['context_id' => null]);
```
asmecher, line 66: I would also suggest deleting any `user_groups` entries where `context_id > 0 AND role_id=1`. Entries like that might be an outcome of https://github.com/pkp/pkp-lib/issues/10180.

classes/navigationMenu/NavigationMenuDAO.php
```diff
@@ -197,7 +198,7 @@ public function installSettings(int $contextId, string $filename): bool
 
         foreach ($tree->getChildren() as $navigationMenuNode) {
             $site = $navigationMenuNode->getAttribute('site');
-            if ($contextId == \PKP\core\PKPApplication::CONTEXT_ID_NONE && !$site) {
+            if ($contextId === Application::SITE_CONTEXT_ID && !$site) {
```
asmecher, line 201: I don't suggest adding the strict checks here unless you're sure they won't run afoul of something like https://github.com/pkp/pkp-lib/issues/10034.

classes/navigationMenu/NavigationMenuItemDAO.php
```diff
@@ -111,15 +112,15 @@ public function getByTypeAndTitleLocaleKey(int $contextId, string $menuItemType,
      *
      * @return DAOResultFactory<NavigationMenuItem>
      */
-    public function getByType(string $type, ?int $contextId = null): DAOResultFactory
+    public function getByType(string $type, ?int $contextId = Application::SITE_CONTEXT_ID_ALL)
```
asmecher, line 115: This removes the return typehint; is that intentional?

classes/security/Validation.php
```diff
@@ -195,21 +195,13 @@ public static function checkCredentials($username, $password)
...
      * @return bool
      */
-    public static function isAuthorized($roleId, $contextId = 0)
+    public static function isAuthorized($roleId, ?int $contextId = Application::SITE_CONTEXT_ID)
     {
         if (!self::isLoggedIn()) {
             return false;
         }
 
-        if ($contextId === -1) {
```
asmecher, line 204: I take it this case was no longer in any calling code?

classes/services/PKPStatsGeoService.php
```diff
@@ -88,10 +88,6 @@ public function getDefaultArgs(): array
         return [
             'dateStart' => StatisticsHelper::STATISTICS_EARLIEST_DATE,
             'dateEnd' => date('Y-m-d', strtotime('yesterday')),
-
-            // Require a context to be specified to prevent unwanted data leakage
-            // if someone forgets to specify the context.
-            'contextIds' => [\PKP\core\PKPApplication::CONTEXT_ID_NONE],
```
asmecher, line 94: Are you sure about removing this safety (here and below)?

asmecher (changes requested):
Thanks, @jonasraoni! This does play a little fast and loose with the distinction between `0` and `null`, both in the database (`COALESCE` used quite heavily) and PHP-land (e.g. `SITE_CONTEXT_ID|intval`) but I think this does take huge steps in the right direction by adding referential integrity and a lot of clearer typehinting.

Have a look through the comments and when they're wrapped up please go ahead and port to OMP and OPS!

classes/security/Validation.php
jonasraoni: Yep, I didn't find anything using it.

classes/navigationMenu/NavigationMenuItemDAO.php
jonasraoni: I didn't notice this one while rebasing/merging 😁

classes/navigationMenu/NavigationMenuDAO.php
jonasraoni: Yeah, I'll drop the strict check 😪

jonasraoni: Even though in this case it might be ok (I guess null values will be retrieved as null instead of an empty string).

classes/migration/upgrade/v3_5_0/I8333_AddMissingForeignKeys.php
jonasraoni: I'll do it, it makes sense.

classes/migration/upgrade/v3_5_0/I8333_AddMissingForeignKeys.php
jonasraoni: This code has aged, but I'll update haha.

classes/core/PKPPageRouter.php
jonasraoni: Yeah, my intention was to reserve the "index", if someone decides to rename it to something else, it would be an easier task.

The default op was an overlook (the other `index` usages were left untouched) and the default page a mistake (my automatic pilot told me it was a virtual path and not a physical one), I'll update both :)

classes/log/SubmissionEmailLogDAO.php
jonasraoni: The PR was merged, so I'll just do a rebase later.

classes/middleware/SetupContextBasedOnRequestUrl.php
jonasraoni: Yep, this is another thing that I'm going to review before merging (I'll do self-review and write some final comments later). I found it curious to have both :)

classes/log/EmailLogEntry.php
jonasraoni: In general, I prefer to avoid nulls, unless it's really needed (price: 0 vs null). Otherwise we need to be careful when passing it around (e.g. `strlen($name ?? '')`).

classes/middleware/SetupContextBasedOnRequestUrl.php
jonasraoni: @asmecher I didn't flag anything special regarding the `_`, there are such checks in place: `($contextPath === 'index' || !$contextPath || $contextPath === Application::CONTEXT_ID_ALL)`

The origin of the `_` is here (previously `*` was being used, then it caused problems and got replaced by `_`): https://github.com/pkp/pkp-lib/issues/4414#issuecomment-764669427

The only drawback is that it's currently possible to use `index` as a journal path (there are 31 journals using such name on the Beacon).

jonasraoni: I think it's interesting to reserve the word `index` and force these journals to change, the name conflicts with the site-wide frontend (in a multi-journal installation, an `index` journal will be shadowed) and might pose other unexpected situations (regarding those checks that I've shared above).

If you agree, I'll leave the code this way and create an extra issue to complete the handling (pre-flight migration to force users to change the path).

asmecher: I agree that the use of `_` is an outlier and if we were introducing it afresh I'd prefer to stick with `index` for consistency -- but this does apply to the public API and would break existing integrations that use the `contexts` endpoint and probably others. See the REST API documentation: https://docs.pkp.sfu.ca/dev/api/ojs/3.4#tag/Access

jonasraoni: Hmm, given the documentation doesn't talk much about it, I think few people might be using this route 🤔

So....:
- I'll keep the updates for the `index`
- Keep this route functional (it's already this way, `_` might be replaced by `index`), but deprecated.
- Create another issue to make the `index` a reserved path.

jonasraoni: I've created this issue https://github.com/pkp/pkp-lib/issues/10241 and I'll push a PR to update the documentation.

Issue #9979, opened 2024-05-22
Scheduled task cannot open log file


ublukas:
**Describe the bug**
The scheduled task for `mailable.editorialReminder` fails to open its logfile when using the german locale.
It uses the translation as a log file name (`EineautomatisierteE-Mail,dieanRedakteur/innenmitaussstehendenAufgabengeschicktwird-664961d155097-20240519.log`) which cannot be created since it contains a `/`.

**Proposed Solution**
Remove or escape all special characters in the file name.

**What application are you using?**
 OMP version 3.4.0-5

**Additional information**
stacktrace:
```
PHP Warning:  fopen(/data/omp-files/scheduledTaskLogs/EineautomatisierteE-Mail,dieanRedakteur/innenmitaussstehendenAufgabengeschicktwird-664961d155097-20240519.log): Failed to open stream: No such file or directory in /data/www/omp-test/lib/pkp/classes/scheduledTask/ScheduledTask.php on line 127
PHP Fatal error:  Uncaught TypeError: flock(): Argument #1 ($stream) must be of type resource, bool given in /data/www/omp-test/lib/pkp/classes/scheduledTask/ScheduledTask.php:128
Stack trace:
#0 /data/www/omp-test/lib/pkp/classes/scheduledTask/ScheduledTask.php(128): flock()
#1 /data/www/omp-test/lib/pkp/classes/scheduledTask/ScheduledTask.php(162): PKP\scheduledTask\ScheduledTask->addExecutionLogEntry()
#2 /data/www/omp-test/lib/pkp/classes/cliTool/ScheduledTaskTool.php(134): PKP\scheduledTask\ScheduledTask->execute()
#3 /data/www/omp-test/lib/pkp/classes/cliTool/ScheduledTaskTool.php(111): PKP\cliTool\ScheduledTaskTool->executeTask()
#4 /data/www/omp-test/lib/pkp/classes/cliTool/ScheduledTaskTool.php(81): PKP\cliTool\ScheduledTaskTool->parseTasks()
#5 /data/www/omp-test/tools/runScheduledTasks.php(26): PKP\cliTool\ScheduledTaskTool->execute()
#6 {main}
  thrown in /data/www/omp-test/lib/pkp/classes/scheduledTask/ScheduledTask.php on line 128
```

<hr>

PRs:
stable-3_3_0:
- https://github.com/pkp/pkp-lib/pull/10043
- https://github.com/pkp/ojs/pull/4310 (only submodule update)
[...]

PR #10037, opened 2024-06-11
pkp/pkp-lib#9979 use class name for scheduled task log file name

bozana:
s. https://github.com/pkp/pkp-lib/issues/9979

### DIFF
--- a/classes/scheduledTask/ScheduledTask.php
+++ b/classes/scheduledTask/ScheduledTask.php
@@ -52,7 +52,8 @@ abstract class ScheduledTask
         $fileMgr = new PrivateFileManager();
 
         $scheduledTaskFilesPath = realpath($fileMgr->getBasePath()) . '/' . ScheduledTaskHelper::SCHEDULED_TASK_EXECUTION_LOG_DIR;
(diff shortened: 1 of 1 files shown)


classes/scheduledTask/ScheduledTask.php
```diff
@@ -52,7 +52,8 @@ public function __construct($args = [])
         $fileMgr = new PrivateFileManager();
 
         $scheduledTaskFilesPath = realpath($fileMgr->getBasePath()) . '/' . ScheduledTaskHelper::SCHEDULED_TASK_EXECUTION_LOG_DIR;
-        $this->_executionLogFile = "{$scheduledTaskFilesPath}/" . str_replace(' ', '', $this->getName()) .
+        $classNameParts = explode('\\', get_class($this)); // Separate namespace info from class name
```
jonasraoni, line 55: Not important, but you can use `$this::class`.

jonasraoni (approved):

classes/scheduledTask/ScheduledTask.php
bozana: Thanks a lot @jonasraoni! I will leave it so here, but change it for main...

Issue #10038, opened 2024-06-11
Not possible to add an institution


bozana:
PR: 
https://github.com/pkp/pkp-lib/pull/10039
https://github.com/pkp/ojs/pull/4307 (only test i.e. submodule udpate)

bozana:
@touhidurabir, could you please take a look at the PR above?
If OK, I would merge only PKP-LIB PR and no submodule udpates...

PR #10039, opened 2024-06-11
pkp/pkp-lib#10038 fix for adding an institution

bozana:
s. https://github.com/pkp/pkp-lib/issues/10038


touhidurabir:
@bozana all ok . good to merge .

Issue #9191, opened 2023-07-27
Inconsistent seq values in database affecting journals ordering 


jardakotesovec:
**Describe the bug**
When new Journal is added, ordering gets recalculated incorrectly.

**To Reproduce**
If I create consequently 4 journals. They will have following values in `seq` column in the `journals` table:
```
1
1
3
3
```

Expected is
```
1
2
3
4
```

After reordering - it creates correct seq values. But goes wrong again after adding new journal.

**What application are you using?**
OJS 3.4

asmecher:
(Added to the 3.3.0-18 milestone, but the issue hasn't yet been replicated there.)

taslangraham:
I've reproduced the inconsistent ordering on OJS.3.3, with a slight difference in ordering. After creating 4 consecutive records, they have the following values in the `seq` column in the `journals` table:
3
3
1
1

jonasraoni:
Another relevant comment is that such type operation should be atomic, especially now that we support APIs and people might call them in parallel. We can't grab a "last ID", then do further operations assuming that the database still has the same state.

asmecher:
@jonasraoni, true, but note that this isn't an ID column -- it's a sequence column. We never get values by sequence, we only use it for ordering, so the impact of a collision would just be that two items will jostle for position until a resequencing happens.

@taslangraham, not that it's any better, but we do have an existing pattern you could follow in order to stay consistent; see:

https://github.com/pkp/pkp-lib/blob/87db0786c680408d0dab0ccc56e6e750f159e748/controllers/grid/settings/category/form/CategoryForm.inc.php#L244..L246

I'd suggest following that pattern for `stable-3_3_0` (and probably `stable-3_4_0`); if you'd like to follow a different pattern for `main`, feel free to propose something -- but be prepared to have to apply it to a few spots in the codebase :)

asmecher:
(In fact, there already is a `resequence` function for this purpose in `lib/pkp/classes/context/ContextDAO.php` in `ojs-main`, and `.inc.php` in `stable-3_3_0`.)

jonasraoni:
No worries, just offering some ideas to improve, it's also possible to avoid the resequencing (I've shared a link with Erik) 😁

jardakotesovec:
Did not expect this issue to get so much attention :-D

asmecher:
@jonasraoni, I am open to new patterns on this -- what's the link you shared with Erik?

taslangraham:
[...]

PR #10047, opened 2024-06-12
pkp/pkp-lib/issues/9191 Fix Context seq ordering

taslangraham:
For #9191

### DIFF
--- a/classes/context/ContextDAO.inc.php
+++ b/classes/context/ContextDAO.inc.php
@@ -143,22 +143,9 @@ abstract class ContextDAO extends SchemaDAO {
 	 */
 	function resequence() {
 		$result = $this->retrieve('SELECT ' . $this->primaryKeyColumn . ' AS context_id FROM ' . $this->tableName . ' ORDER BY seq');
-		$i=1;
-		for ($i=1; $row = (array) $result->current(); $i+=2 && $result->next()) {
-			$this->update('UPDATE ' . $this->tableName . ' SET seq = ? WHERE ' . $this->primaryKeyColumn . ' = ?', [$i, $row['context_id']]);
-			$result->next();
-			$i+=2;
+
+		foreach ($result as $key => $value) {
+			$this->update('UPDATE ' . $this->tableName . ' SET seq = ? WHERE ' . $this->primaryKeyColumn . ' = ?', [$key + 1, $value->context_id]);
 		}
-	}
-
-	/**
-	 * Calculates the seq value to be assigned to a new Context entry.
-	 * @return int The seq value for a new Context entry.
-	 */
-	public function getNextSeqValue()
-	{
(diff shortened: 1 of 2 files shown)


classes/context/ContextDAO.inc.php
```diff
@@ -144,10 +144,9 @@ function getBySetting($settingName, $settingValue, $contextId = null) {
 	function resequence() {
 		$result = $this->retrieve('SELECT ' . $this->primaryKeyColumn . ' AS context_id FROM ' . $this->tableName . ' ORDER BY seq');
 		$i=1;
-		for ($i=1; $row = (array) $result->current(); $i+=2 && $result->next()) {
+		for ($i = 1; $row = (array) $result->current(); $i += 1 && $result->next()) {
 			$this->update('UPDATE ' . $this->tableName . ' SET seq = ? WHERE ' . $this->primaryKeyColumn . ' = ?', [$i, $row['context_id']]);
-			$result->next();
-			$i+=2;
+			$i += 1;
```
asmecher, line 149: I think `$i` is already being incremented in the `for` statement, so could this not be removed?

classes/context/ContextDAO.inc.php
```diff
@@ -144,10 +144,9 @@ function getBySetting($settingName, $settingValue, $contextId = null) {
 	function resequence() {
 		$result = $this->retrieve('SELECT ' . $this->primaryKeyColumn . ' AS context_id FROM ' . $this->tableName . ' ORDER BY seq');
 		$i=1;
-		for ($i=1; $row = (array) $result->current(); $i+=2 && $result->next()) {
+		for ($i = 1; $row = (array) $result->current(); $i += 1 && $result->next()) {
```
jonasraoni, line 147: As you're touching this old code:
- I don't see a reason to convert the object into an array.
- The `1 && $result->next()` turns into a `boolean`, I think it would be better to separate the statements.
  - With a comma: `for ($i = 1; $row = $result->current(); ++$i, $result->next()) { blabla; }`
  - Or to move it to another line : `for ($i = 1; $row = $result->current(); ++$i) { blabla; $result->next(); }

taslangraham: Updated to use `foreach`

classes/context/ContextDAO.inc.php
taslangraham: updated - replaced the `for` loop with a `foreach`

jonasraoni (commented):
It's already merged, but as a review was requested, the last comment would be to use the `DB::table('table')->update()` syntax.
Other possibilities:
- Decrease the number of database operations (use a single update to handle all records), it's possible.
- Or lock the rows during the update to avoid side requests from modifying the sequence.
- Find the motivation/origin of the `resequence()` and improve it (only modify the needed records).

PR #10052, opened 2024-06-13
pkp/pkp-lib#10051 Fixed HTML entity code &amp; in OJS email subject

### LINKED ISSUE #10051: Incorrect handling of HTML entity code &amp; in OJS email subject
**Describe the bug**
When sending emails from OJS, the HTML entity `&amp;` is displayed as `&amp;` in the email subject instead of `&`.

**To Reproduce**
1. Configure OJS to send an email with a subject containing `&`.
2. Observe that the subject in the received email contains `&amp;` instead of `&`.

  OJS 3.3.0.13

**Additional information**
Reference forum post: [HTML entity code &amp; in OJS email subject](https://forum.pkp.sfu.ca/t/html-entity-code-amp-in-ojs-email-subject/88673)

Hafsa-Naeem:
Fixes #10051 

### Changes Made
- Updated the `setSubject` function to sanitize and decode the email subject correctly.
- Added a `sanitizeSubject` function to remove HTML tags from the email subject for security.

### DIFF
--- a/classes/mail/Mail.inc.php
+++ b/classes/mail/Mail.inc.php
@@ -362,11 +362,24 @@ class Mail extends DataObject {
 	}
 
 	/**
-	 * Set the subject of the message.
-	 * @param $subject string
-	 */
+	* Set the subject of the message.
+	* @param $subject string
+	*/
 	function setSubject($subject) {
-		$this->setData('subject', $subject);
+		$sanitizedSubject = $this->sanitizeSubject($subject);
+		$decodedSubject = htmlspecialchars_decode($sanitizedSubject, ENT_QUOTES | ENT_HTML5);
+		$this->setData('subject', $decodedSubject);
+	}
+
+	/**
+	 * Sanitize subject to remove any harmful content.
+	 * @param $subject string
+	 * @return string
+	 */
+	function sanitizeSubject($subject) {
+		$sanitizedSubject = strip_tags($subject);
+		$sanitizedSubject = preg_replace('/[\x00-\x1F\x7F<>]/', '', $sanitizedSubject);
+		return $sanitizedSubject;
 	}
 
 	/**
--- a/classes/mail/Mail.inc.php
+++ b/classes/mail/Mail.inc.php
@@ -366,20 +366,7 @@ class Mail extends DataObject {
 	* @param $subject string
 	*/
 	function setSubject($subject) {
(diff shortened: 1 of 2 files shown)


classes/mail/Mail.inc.php
```diff
@@ -362,11 +362,24 @@ function getReplyToString($send = false) {
 	}
 
 	/**
-	 * Set the subject of the message.
+	* Set the subject of the message.
+	* @param $subject string
+	*/
+	function setSubject($subject) {
```
asmecher, line 368: `setSubject` is just a setter function, so I'd rather we kept anything more complicated than that outside of it.

We have a convention to avoid XSS flaws in emails that looks a little like this:
- In email templates, you can use replacement variables like `{$contextName}` so that the templates are personalized before sending.
- Just about all of those variables appear in the email body -- but a few do appear in the subject, like in this case. The same variables and values are available to both places. (You can also look at the Announcements tool for another example where this problem shows up; if you put a `&` in the announcement title it'll also come out as `&amp;` when sent.)
- Email bodies are in HTML format, and XSS escaping is expected to be done when assigning variables to the template, which is when we know how to escape them (based on whether they are HTML or plaintext values, for example). [Example](https://github.com/pkp/pkp-lib/blob/87db0786c680408d0dab0ccc56e6e750f159e748/classes/mail/MailTemplate.inc.php#L151)
- So you can see we have a mismatch: variables are provided to the template in HTML format, and then applied to both the body and the subject as required. However, subjects expect plaintext, and bodies expect HTML.
- You can see where variables are replaced in both the body and the subject in the [Mail::send](https://github.com/pkp/pkp-lib/blob/87db0786c680408d0dab0ccc56e6e750f159e748/classes/mail/MailTemplate.inc.php#L222) function (for `stable-3_3_0`; this has changed in `stable-3_4_0` and `main`). Here, the same approach is applied to both body and subject -- I think that's the heart of the bug.

So I'd suggest fixing the escaping of the subject in the `MailTemplate::replaceParams` function.

classes/mail/Mail.inc.php
```diff
@@ -362,9 +362,9 @@ function getReplyToString($send = false) {
 	}
 
 	/**
-	 * Set the subject of the message.
-	 * @param $subject string
-	 */
+	* Set the subject of the message.
```
asmecher, line 365: Looks like just a couple of inadvertent spacing changes left behind here.

Issue #10053, opened 2024-06-14
Introduce Author::add::before hook


bozana:
Introduce the new Author::add::before before inserting the author https://github.com/pkp/pkp-lib/blob/4c027a986761e8751df91e85dabeb7e1e498573d/classes/author/Repository.php#L162

PR:
- ~~https://github.com/pkp/pkp-lib/pull/10054~~ https://github.com/pkp/pkp-lib/pull/9726
- ~~https://github.com/pkp/ojs/pull/4315 (only submodule update)~~

bozana:
Hi @jonasraoni, could you please take a look at this PR? Thanks a lot!

forgive38:
Hello,
il this PR is validated, you can close  #
 #9726

bozana:
Ah, I was not aware (and have not searched) of that PR... :-\
Thanks @forgive38!

bozana:
Thanks a lot!
I merged the PR from @forgive38, and closed my PRs...
Will close the issue...

PR #10054, opened 2024-06-14
pkp/pkp-lib#10053 move Author::add hook before insert

bozana:
s. https://github.com/pkp/pkp-lib/issues/10053

### DIFF
--- a/classes/author/Repository.php
+++ b/classes/author/Repository.php
@@ -149,6 +149,7 @@ class Repository
      * @copydoc \PKP\services\entityProperties\EntityWriteInterface::add()
      *
      * @hook Author::add [[$author]]
+     * @hook Author::add::before [[$author]]
      */
     public function add(Author $author): int
     {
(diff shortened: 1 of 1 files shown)


jonasraoni (changes requested):
I found a bunch of `::add` hooks, which are called after adding the entity. So I don't agree with the change, unless we modify it everywhere.

If you need a hook at this point, there's another pattern in place, which is better, see the:
https://github.com/pkp/pkp-lib/blob/188bba290b9e8e2b158891e71cd262d484d944af/classes/author/Repository.php#L183-L191

I've wrote about it before, there should be a standard, like database triggers... Before/after for all the add/edit/delete operations. And I'd rather use events instead of hooks for everything, hard coded strings are evil 😁

bozana:
Ah, OK, your are right... I would then add Hook::call(' ', [$author]);  for now -- the hook is needed by CRAFT OA...

bozana:
@jonasraoni, now OK?

jonasraoni (approved):

PR #10058, opened 2024-06-15
pkp/pkp-lib#10051 Fixed HTML entity code &amp; in OJS email subject

### LINKED ISSUE #10051: Incorrect handling of HTML entity code &amp; in OJS email subject

Hafsa-Naeem:
Fixes #10051

### DIFF
--- a/classes/mail/Mailable.php
+++ b/classes/mail/Mailable.php
@@ -353,7 +353,10 @@ class Mailable extends IlluminateMailable
      */
     protected function buildSubject($message): self
     {
-        $this->subject ??= ''; // Allow email with empty subject if not set
+        if ($this->subject === null) {
+            $this->subject = ''; // Allow email with empty subject if not set
+        }
+
         $withoutTagViewData = collect($this->viewData)
             ->map(fn(mixed $viewableData) => is_string($viewableData) ? strip_tags($viewableData) : $viewableData)
             ->toArray();
@@ -364,7 +367,7 @@ class Mailable extends IlluminateMailable
 
         if (empty($subject)) {
             trigger_error(
-                'You are sending ' . static::getName() ?? static::class . ' email with empty subject',
+                'You are sending ' . (static::getName() ?? static::class) . ' email with empty subject',
                 E_USER_WARNING
             );
         }
--- a/classes/mail/Mailable.php
+++ b/classes/mail/Mailable.php
@@ -353,10 +353,7 @@ class Mailable extends IlluminateMailable
      */
     protected function buildSubject($message): self
     {
-        if ($this->subject === null) {
-            $this->subject = ''; // Allow email with empty subject if not set
-        }
-
+        $this->subject ??= ''; // Allow email with empty subject if not set
         $withoutTagViewData = collect($this->viewData)
             ->map(fn(mixed $viewableData) => is_string($viewableData) ? strip_tags($viewableData) : $viewableData)
             ->toArray();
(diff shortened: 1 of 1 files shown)


classes/mail/Mailable.php
```diff
@@ -359,6 +359,9 @@ protected function buildSubject($message): self
             ->toArray();
 
         $subject = app('mailer')->compileParams($this->subject, $withoutTagViewData);
+        // decode HTML entities for the subject
+        $subject = htmlspecialchars_decode($subject, ENT_QUOTES | ENT_HTML5);
+
         if (empty($subject)) {
             trigger_error(
                 'You are sending ' . static::getName() ?? static::class . ' email with empty subject',
```
jonasraoni, line 367: @Hafsa-Naeem as you're updating this part of code, could please fix this issue here?

The second part of the `??` will never be executed.

asmecher: @Hafsa-Naeem, I see you've changed this line, but I don't think the change addresses @jonasraoni's comment. I think he was pointing out that `Mailable::getName()` will always return a `string` (see [the function declaration](https://github.com/pkp/pkp-lib/blob/1d2e1eeedde89ef2a42d687e1a3cd69ad61db39c/classes/mail/Mailable.php#L134)), so the `?? static::class` will never be executed. You can remove `?? static::class` entirely.

asmecher: @jonasraoni and @Hafsa-Naeem, it looks like you two collided on this change:
- @jonasraoni made this change: https://github.com/pkp/pkp-lib/commit/49ad7a8a02ad34dbc06ad148224bd8ac5d42ccfd
- @Hafsa-Naeem made this change: https://github.com/pkp/pkp-lib/commit/70278465c9c0137a79bc4092ee064e67c64ffe2d

@jonasraoni, what was the intention of your change?

jonasraoni: Hi! I went ahead and applied the blind fix yesterday, but after watching the code...

I think we should use just the `static::class`. The `getName()` uses translated content, but logs are for developers, we don't need it, it just makes things more difficult to understand (e.g. user might change the translation). If the locale is important for the error, it should be added properly.

But if we want to keep it this way... Then as the `getName()` returns a string, and possibly an empty one, it would be better to use `(static::getName() ?: static::class)`.

And a last comment: The `trigger_error()` isn't very helpful, I'd rather use an `Exception` (`error_log(new Exception('error'))`) to display the stack trace, until we enable the Laravel logger https://github.com/pkp/pkp-lib/discussions/7966.

Hafsa-Naeem: @asmecher @jonasraoni Thanks for the feedbacks. based on both your reviews, the solution might be:

- Retaining the initial check for null and set this->subject to an empty string if needed.
- Use static::class for logging since getName may return translated content which might not be useful in logs.

- Also replace trigger_error with error_log(new Exception(...)) for better debugging?

```
if (empty($subject)) {
    error_log(new Exception(
        'You are sending ' . static::class . ' email with empty subject'
    ));
}

```

asmecher: We're bikeshedding at this point :)

This code is only executed in an unusual circumstance, and we just need something that'll allow us to help debug when someone in the wild encounters it e.g. with custom code or an out-of-spec database. I'm perfectly happy with the approach that's currently in the PR, @Hafsa-Naeem.

However, we do have a merge conflict. I'll revert https://github.com/pkp/pkp-lib/commit/49ad7a8a02ad34dbc06ad148224bd8ac5d42ccfd and merge this PR as-is.

Issue #10061, opened 2024-06-15
Highlights feature shouldn't be displayed if it's not enabled


jonasraoni:
**Describe the bug**
The highlights it's feature flagged, therefore it shouldn't be displayed on the site settings.

The fix is needed only for OJS 3.4, on main it's supposed to not be behind a flag.

**To Reproduce**
1. Ensure the journal has at least 2 journals
2. Click on Administration > Site Settings
3. See the feature is available

**PRs**
- https://github.com/pkp/pkp-lib/pull/10063
- https://github.com/pkp/ojs/pull/4317

**What application are you using?**
OJS 3.4

jonasraoni:
@bozana could you review the linked PRs?

PR #10063, opened 2024-06-15
pkp/pkp-lib#10061 Updated highlights tab to respect the feature flag

jonasraoni:


pages/admin/AdminHandler.php
```diff
@@ -274,27 +265,20 @@ private function siteSettingsAvailability($request)
...
+        $tabs['announcements'] = (bool) Config::getVar('features', 'site_announcements');
+        $tabs['highlights'] = (bool) Config::getVar('features', 'highlights');
 
-        foreach ($tabs as $tab) {
-            $tabsAvailability[$tab] = true;
-            if ($singleContextSite && !in_array($tab, $tabsSingleContextAvailability)) {
-                $tabsAvailability[$tab] = false;
+        $isSingleContextSite = Services::get('context')->getCount() < 2;
+        if ($isSingleContextSite) {
+            $multipleContextTabs = ['siteSetup', 'languages', 'bulkEmails', 'statistics'];
```
bozana, line 275: I think this is wrong -- those are single context tabs

bozana (commented):
Hi @jonasraoni, I left 2 comments...
Thanks a lot!

pages/admin/AdminHandler.php
jonasraoni: Yes! Looks like I had a brief moment of dyslexia and inverted things 😁

pages/admin/AdminHandler.php
```diff
@@ -274,27 +265,20 @@ private function siteSettingsAvailability($request)
...
 
-        $tabsAvailability = [];
+        $tabs['announcements'] = (bool) Config::getVar('features', 'site_announcements');
+        $tabs['highlights'] = (bool) Config::getVar('features', 'highlights');
 
-        foreach ($tabs as $tab) {
-            $tabsAvailability[$tab] = true;
-            if ($singleContextSite && !in_array($tab, $tabsSingleContextAvailability)) {
-                $tabsAvailability[$tab] = false;
+        $isSingleContextSite = Services::get('context')->getCount() < 2;
```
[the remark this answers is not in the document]
jonasraoni: I think the previous behavior would probably display all tabs, I'll remove the contexts and do a small test.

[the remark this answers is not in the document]
jonasraoni: FYI I saw that it makes sense to keep the `== 1`. When there's no journal, the "site UI" is displayed. So, I've updated and left a comment :)

pages/admin/AdminHandler.php
```diff
@@ -247,54 +247,27 @@ public function settings($args, $request)
...
-            'languages',
-            'navigationMenus',
-            'highlights',
-            'bulkEmails',
-            'siteTheme',
-            'siteAppearanceSetup',
-            'statistics',
-            'announcements',
+        // The multi context UI is also displayed when the journal has no contexts
+        $isMultiContextSite = Services::get('context')->getCount() === 1;
```
bozana, line 255: It should be `!==`, right? :sweat_smile:

jonasraoni: Definitely 😆
I'll use the opportunity to debug why the upload at the `NativeXmlImportExportSubmission` is failing.

PR #10064, opened 2024-06-15
Bugfix stable 3 4 0 10060 fix announcement type removal

jonasraoni:


classes/announcement/Announcement.php
```diff
@@ -35,92 +35,76 @@ class Announcement extends \PKP\core\DataObject
...
-    public function getAnnouncementTypeName()
+    public function getAnnouncementTypeName(): ?string
     {
+        if (!$this->getData('typeId')) {
+            return null;
+        }
         $announcementTypeDao = DAORegistry::getDAO('AnnouncementTypeDAO'); /** @var AnnouncementTypeDAO $announcementTypeDao */
-        $announcementType = $announcementTypeDao->getById($this->getData('typeId'));
-        return $announcementType ? $announcementType->getLocalizedTypeName() : null;
+        return $announcementTypeDao->getById($this->getData('typeId'))->getLocalizedTypeName();
```
bozana, line 93: `$announcementTypeDao->getById($this->getData('typeId'))` could theoretically be null.

bozana (commented):
Hi @jonasraoni,
I will post my comments now -- not to lose them... :-)
We talked about some (having always ann title and datePosted, and annType name + no need to delete announcements when announcement type is deleted), but here are a few other smaller comments/questions...

classes/announcement/Collector.php
```diff
@@ -165,53 +169,39 @@ public function orderBy(?string $sorter, string $direction = self::ORDER_DIR_DES
...
+            fn (Builder $qb) => $qb->where(fn (Builder $qb) => $qb->where('a.date_expire', '>', $this->isActive)->orWhereNull('a.date_expire'))
+        );
+
+        $searchPhrase = trim($this->searchPhrase ?? '');
+        if (strlen($searchPhrase)) {
+            $words = PKPString::regexp_split('/\s+/', $searchPhrase);
+            $qb->whereIn('a.announcement_id', function (Builder $q) use ($words) {
+                $q->select('announcement_id')->from($this->dao->settingsTable);
+                foreach ($words as $word) {
+                    $word = strtolower(addcslashes($word, '%_'));
```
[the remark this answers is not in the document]
jonasraoni: It's better, I trust more in the database than PHP when handling UTF-8 data, I just copied this from another place, so I'll probably spread the changes.

classes/announcement/Announcement.php
jonasraoni: Yep! I remember about changing this, it should be `?->`

classes/announcement/Announcement.php
```diff
@@ -35,92 +35,76 @@ class Announcement extends \PKP\core\DataObject
...
     }
 
     /**
      * Get localized announcement title
-     *
-     * @return string
      */
-    public function getLocalizedTitle()
+    public function getLocalizedTitle(): ?string
```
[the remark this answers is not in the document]
jonasraoni: It's not possible to trust in the content of our `*_settings` tables, besides the lack of data integrity, maybe in the past it was allowed to have nulled data here, which will raise an Exception unless I do coercions.

I think we should either coerce to empty strings (not null data is easier to deal with and consumes less space in the database) or allow nulls.

I decided to allow nulls, due to the existence of generic methods like the `getData('field')`, which will bring data as it was stored, perhaps with nulls.

classes/announcement/Announcement.php
```diff
@@ -132,149 +116,124 @@ public function getLocalizedTitleFull()
...
     }
 
     /**
      * Get announcement posted date.
      *
-     * @return string (YYYY-MM-DD)
+     * @return ?string Format YYYY-MM-DD HH:MM:SS
      */
-    public function getDatePosted()
+    public function getDatePosted(): ?string
```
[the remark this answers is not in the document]
jonasraoni: Yep, it's definitely ok to remove the nullability here, the field is not stored in the settings table and not null.

classes/announcement/Collector.php
```diff
@@ -165,53 +169,39 @@ public function orderBy(?string $sorter, string $direction = self::ORDER_DIR_DES
...
-            $qb->whereIn('a.assoc_id', $this->contextIds);
-            if ($this->includeSite === self::SITE_AND_CONTEXTS) {
-                $qb->orWhereNull('a.assoc_id');
-            }
-        } elseif ($this->includeSite === self::SITE_ONLY) {
-            $qb->where('a.assoc_type', Application::get()->getContextAssocType());
-            $qb->whereNull('a.assoc_id');
+        $qb = DB::table($this->dao->table, 'a')
+            ->select(['a.*'])
+            ->where('a.assoc_type', Application::get()->getContextAssocType());
```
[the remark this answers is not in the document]
jonasraoni: Yes, nobody can understand what this field accepts without inspecting the code. I think I'll do this update for the main branch and use a `context_id` to help getting rid of the ASSOC_* constants.

classes/announcement/Announcement.php
```diff
@@ -132,149 +116,124 @@ public function getLocalizedTitleFull()
...
-     *
-     * @param string $locale
-     *
-     * @return string
      */
-    public function getTitle($locale)
+    public function getTitle(?string $locale): array|string|null
     {
-        return $this->getData('title', $locale);
+        return $this->getData('title', $locale) ?? '';
```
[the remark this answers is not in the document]
jonasraoni: I'll remove it, just curious why I've added it 🤔

classes/announcement/Announcement.php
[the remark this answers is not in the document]
jonasraoni: I'll take the same approach for all the null comments, this should be migrated to a Laravel model later, then we can rethink how things are done.

classes/announcement/Collector.php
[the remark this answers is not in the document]
asmecher: Laravel does have an equivalent in the form of [`morphs`](https://laravel.com/docs/10.x/migrations#column-method-morphs) relationships. It's exactly the same pattern (a paired ID and type column). In cases where we do want to preserve the flexibility that `assoc_type` / `assoc_id` gives us, I suggest we consider eventually moving fully to Laravel `morphs`. So I don't think we should declare war on `assoc_type` / `assoc_id` -- it's not a 100% PKP invention.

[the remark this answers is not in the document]
jonasraoni: Yeah, it's an old idea, I've used this to generate dynamic links to linkable parts of the system (e.g. link to a category, product, blog post, etc.) in a CMS.
If we don't have a really dynamic need, I'd rather add extra columns (`context_id`, `user_id`, `*_id`). The lack of data integrity is even more important for me than the cognitive issue I've mentioned.

classes/announcement/Announcement.php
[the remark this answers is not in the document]
asmecher: To answer the question directly, though: there should always be an announcement title, since it's a required field for announcements. Even in an obscure case like a journal that has changed primary locales, `DataObject::getLocalizedData` will fall back on the first available piece of data if the locale precedence list doesn't provide something.

classes/announcement/Announcement.php
jonasraoni: I just saw there's a `if (!$this->getData('typeId'))` before this line, so it should be ok for our standards.

Anyway, I'll add the extra check (for the theoretical case where the type is removed a couple of milliseconds after the announcement is retrieved).

classes/announcement/AnnouncementType.php
```diff
@@ -20,58 +20,42 @@
...
         $this->setData('contextId', $contextId);
     }
 
     /**
      * Get the type of the announcement type.
-     *
-     * @return string
      */
-    public function getLocalizedTypeName()
+    public function getLocalizedTypeName(): ?string
```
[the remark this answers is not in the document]
jonasraoni: I've coerced to an empty string, which is safer (e.g. `(new Announcement())->getLocalizedTypeName()` doesn't depend on the database and would fail)

classes/announcement/Announcement.php
```diff
@@ -132,149 +116,124 @@ public function getLocalizedTitleFull()
 
     /**
      * Get announcement title.
-     *
-     * @param string $locale
-     *
-     * @return string
      */
-    public function getTitle($locale)
+    public function getTitle(?string $locale): array|string|null
```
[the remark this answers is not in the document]
jonasraoni: These methods that forward arguments to `getData()`/`setData()` are risky for me, the return/side-effects are variable and I don't trust in the contents of the `*_settings` tables, so I'll leave it as a debt, for the time when we move this to a Laravel model.

classes/announcement/Announcement.php
```diff
@@ -132,149 +116,124 @@ public function getLocalizedTitleFull()
...
     }
 
     /**
      * Set announcement title.
-     *
-     * @param string $title
-     * @param string $locale
      */
-    public function setTitle($title, $locale)
+    public function setTitle(array|string|null $title, ?string $locale): void
```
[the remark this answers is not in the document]
jonasraoni: These methods that forward arguments to `getData()`/`setData()` are risky for me, the return/side-effects are variable and I don't trust in the contents of the `*_settings` tables, so I'll leave it as a debt, for the time when we move this to a Laravel model.

classes/announcement/Announcement.php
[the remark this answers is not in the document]
asmecher: I'm not sure it's a bad thing for OJS to raise an error on out-of-spec data. But I don't have a strong opinion on that at the moment, so I'm OK either way.

classes/announcement/AnnouncementTypeDAO.php
```diff
@@ -98,108 +80,75 @@ public function updateLocaleFields($announcementType)
...
-        return $this->deleteById($announcementType->getId());
+        $this->deleteById($announcementType->getId(), $deleteAnnouncements);
     }
 
     /**
-     * Delete an announcement type by announcement type ID. Note that all announcements with
-     * this type ID are also deleted.
-     *
-     * @param int $typeId
+     * Delete an announcement type by announcement type ID. Note that all announcements with this type ID are also deleted.
```
bozana, line 115: Maybe to remove the second sentence?

classes/announcement/AnnouncementTypeDAO.php
```diff
@@ -98,108 +80,75 @@ public function updateLocaleFields($announcementType)
...
+        if ($deleteAnnouncements) {
+            $collector = Repo::announcement()->getCollector()->filterByTypeIds([$typeId]);
+            Repo::announcement()->deleteMany($collector);
+        }
 
-        $collector = Repo::announcement()->getCollector()->filterByTypeIds([(int) $typeId]);
-        Repo::announcement()->deleteMany($collector);
+        DB::table('announcement_types')
+            ->where('type_id', '=', $typeId)
+            ->delete();
```
bozana, line 127: Also the entries in the DB table announcement_type_settings would need to be deleted here, no?

classes/announcement/AnnouncementTypeDAO.php
```diff
@@ -98,108 +80,75 @@ public function updateLocaleFields($announcementType)
...
     }
 
     /**
      * Delete announcement types by context ID.
-     *
-     * @param int $contextId
      */
-    public function deleteByContextId($contextId)
+    public function deleteByContextId(?int $contextId): void
```
bozana, line 133: Maybe to note that if null is given as parameter, the site announcements (context_id = 0) is assumed?
Maybe we can also set the default value?
(The same for the getByContextId())

classes/announcement/Collector.php
```diff
@@ -165,53 +169,39 @@ public function orderBy(?string $sorter, string $direction = self::ORDER_DIR_DES
...
+        $searchPhrase = trim($this->searchPhrase ?? '');
+        if (strlen($searchPhrase)) {
+            $words = PKPString::regexp_split('/\s+/', $searchPhrase);
+            $qb->whereIn('a.announcement_id', function (Builder $q) use ($words) {
+                $q->select('announcement_id')->from($this->dao->settingsTable);
+                foreach ($words as $word) {
+                    $word = addcslashes($word, '%_');
+                    $q->where(
+                        fn (Builder $q) => $q
+                            ->where(fn (Builder $q) => $q->where('setting_name', 'title')->whereRaw('LOWER(setting_value) LIKE ?', ["%{$word}%"]))
```
bozana, line 199: I think here and below `LOWER(?) `should be, no?

schemas/user.json
```diff
@@ -117,7 +117,10 @@
 		},
 		"givenName": {
 			"type": "string",
-			"multilingual": true
+			"multilingual": true,
+			"validation": [
+				"nullable"
+			]
```
bozana, line 123: I am wondering what is the difference between now and earlier (without nullable) :thinking:

bozana (commented):
Hi @jonasraoni 
Just a few comments...

I think we can assume that required fields are there -- this is how we do it everywhere, I think, but I will leave the decision to you.
The getLocalized... functions behave here differently that the others -- they return '' instead of null, if nothing has been found. This maybe makes it more difficult for developers -- they need to check the results differently...
And, maybe to return null (instead of '') for getDatePosted(), if there is no datePosted (even if it should be there/it should exist) -- to behave the same as everywhere?

Thanks a lot!

schemas/user.json
jonasraoni: I've tested this using the context UI (it uses the API under the hood, so perhaps, different validations than different parts of the system) and a multilingual field. This is the behavior that I got:
- Required without nullable: All locales must be filled.
- Required with nullable: Only the main locale is required.

Basically, all fields in the system which are multilingual, also have this nullable validation, so I've just added to the ones that were missing.

classes/announcement/Collector.php
jonasraoni: YES, this was a half-baked update 😂

classes/announcement/AnnouncementTypeDAO.php
jonasraoni: I think it's better to not add a default value, it should be explicit.

p.s.: I'm working on a parallel PR to add the remaining foreign keys (and get rid of the `0` on the database), so updates related to it will be done there, for now I'm using `null`, another possibility that sounds interesting to my brain, but will generate more work is to accept multiple types (e.g. `int|SiteContext $contextId`) when the function is supposed to work with the site context.

classes/announcement/Announcement.php
```diff
@@ -132,149 +110,124 @@ public function getLocalizedTitleFull()
...
+    public function getDateExpire(): ?string
     {
         return $this->getData('dateExpire');
     }
 
     /**
      * Set announcement expiration date.
      *
-     * @param string $dateExpire (YYYY-MM-DD)
+     * @param ?string $dateExpire Format YYYY-MM-DD HH:MM:SS
```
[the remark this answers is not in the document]
jonasraoni: It's a typo, but this function should be smarter (drop the time).

classes/announcement/AnnouncementTypeDAO.php
jonasraoni: This should be handled by the database (there's an `on delete cascade`), and the only reason to delete manually the announcements above, is because there might be physical files on the disk, so an improvement would be to delete manually only the ones that have files.

jonasraoni:
This is my opinion about `nulls`:
In general, I think it just makes everything more complex, you need to do extra checks on the code, and also on the database queries... So, unless it's really required, I think all fields should be not null. Therefore, I'll leave the code this way, in my perception, moving away from nulls is an enhancement.

PHP has strangely deprecated passing null to functions like `strlen()`, so it will create noise on the log (and later fatal errors), as we're definitely not checking everywhere if `getPossibleStringOrNull()` is null before using it, so coercing to an empty string is more developer friendly/less error prone.

About the `getDatePosted()`, I agree, but when dealing with dates I think we should work with the `DateTime`/`DateTimeImmutable`, so I'll do an extra update, and the null will be a requirement.

asmecher:
Just a heads-up that we should be saving rewrites/clean-ups for `main`, not `stable-3_4_0` -- there is a fair chance there will be regressions introduced here without providing much benefit to the end user.

bozana:
Hi @jonasraoni,
What do you mean with 

> and the null will be a requirement

?
Or, I will probably see once you change it :-)

And regarding returning empty string in the getLocalized... functions here:
Please check then their use in the code, e.g. the things like this https://github.com/pkp/pkp-lib/blob/stable-3_4_0/templates/frontend/objects/announcement_full.tpl#L22-L26.

Thanks a lot!

jonasraoni:
@asmecher I agree, breaking changes should be left for main. I start with a little change, then things get out of control 😵

schemas/user.json
bozana: Yes... I also understood it so from the code... but could not test it with user form... Maybe because it is working differently...
:+1:

classes/announcement/AnnouncementTypeDAO.php
bozana: And what about having only `int` and not `?int` ?

classes/announcement/AnnouncementTypeDAO.php
bozana: Ah, right, I forgot about on delete cascade here... :-)

classes/announcement/AnnouncementTypeDAO.php
jonasraoni: As the function was supporting `?int` before, I'll leave it this way to avoid extra changes.
I just can say that different parts of the codebase have a different handling for the "SITE" context, some places accept null, others 0, I'll standardize this on https://github.com/pkp/pkp-lib/issues/8333.

jonasraoni:
I've checked the usage, and everything seems to be fine, the string fields will work better without the null possibility.
I'll change the type of the datePosted to a class just on the `main` branch, I'll wait you to approve before creating new PRs.
And probably a couple of fixes will have to be moved to 3.3 as well.

jonasraoni:
Yep, it should work fine. The only problem would be a strict comparison with null (e.g. `$announcement->getLocalizedDescription() === null`).

Issue #9914, opened 2024-04-25
Editorial History


bozana:
The editorial history page should be under context's about.
It should list all ended editorial masthead services i.e. all masthead roles as well as uses that were active and are not active any more in those masthead roles and that have accepted to be listed on masthead.
Users are listed with following information: full name, affiliation, start date - end date, and orcidIcon (if ORCID authorized).
Reviewer role is treated in the same way, differently to the display on the active masthead page.

JII plugin should be able to overwrite the masthead users -- generally, a user can opt-out of being displayed on masthead, but this does not guilt for JII users -- according to JII all users in the masthead roles need to be displayed (and cannot opt-out).

At the end of the editorial history page, the content of the new free text field (to be found under Settings > Journal > Masthead > Editorial History) should be displayed. The new free text field allows the JM to describe/list the editorial history before they had started to use OJS/OMP/OPS.

bozana:
@asmecher, could you please review the PRs above?

bozana:
Hi @asmecher, I merged the issue https://github.com/pkp/pkp-lib/issues/9937 with this one, because they were closely connected. Thus, could you please review these PRs?
Thanks a lot!

asmecher:
Thanks, @bozana! I approved the app PRs, but added a few suggestions to the pkp-lib PR [here](https://github.com/pkp/pkp-lib/pull/9941#pullrequestreview-2062049776).

bozana:
Some design changes are needed:
- The text that links to the Editorial History from the Editorial Masthead page should be: `View Editorial History for this journal`
- The title heading on the Editorial History page should be `Editorial History Page`, followed by this explanation text: `Here, you will find a comprehensive record of all past contributors associated with this journal. This page highlights our commitment to transparency and showcases our editorial evolution.`
- Editorial History page can contain/list several services for one and the same user and role -- for example if an user was an editor from 2019-2020, and then again from 2022-2023. Thus the services should be listed by start date desc, and user name and affiliation should be listed only once (and not for each service). I.e. like this:
```
Editors

<Name>
Affiliation
Start Date - Last Date
Start Date - Last Date

<Name>
Affiliation
Start Date - Last Date

Section Editor

<Name>
Affiliation
Start Date - Last Date

<Name>
Affiliation
Start Date - Last Date
```

bozana:
Hi @asmecher, could you please take a look at these PRs (also listed above), that contain some design changes discussed with Devika:
[...]

PR #10079, opened 2024-06-18
pkp/pkp-lib#9914 sort services by start date and display user name an…

bozana:
…d affiliation only once on editorial history page, use different text to link to the editorial history form the editorial masthead page

This PR contains some design changes:

- The text that links to the Editorial History from the Editorial Masthead page should be: `View Editorial History for this journal`
- The title heading on the Editorial History page should be `Editorial History Page`, followed by this explanation text: `Here, you will find a comprehensive record of all past contributors associated with this journal. This page highlights our commitment to transparency and showcases our editorial evolution.`
- Editorial History page can contain/list several services for one and the same user and role -- for example if an user was an editor from 2019-2020, and then again from 2022-2023. Thus the services should be listed by start date desc, and user name and affiliation should be listed only once (and not for each service). I.e. like this:

```
Editors

<Name>
Affiliation
Start Date - Last Date
Start Date - Last Date

<Name>
Affiliation
Start Date - Last Date

Section Editor

<Name>
Affiliation
Start Date - Last Date

<Name>
Affiliation
Start Date - Last Date


```

### DIFF
--- a/locale/en/common.po
+++ b/locale/en/common.po
@@ -45,6 +45,9 @@ msgstr "Principal Contact"
 msgid "about.contact.supportContact"
 msgstr "Support Contact"
 
+msgid "about.editorialMasthead.linkToEditorialHistory"
+msgstr "View <a href=\"{$url}\">Editorial History</a>"
+
 msgid "about.other"
 msgstr "Other"
 
@@ -159,6 +162,12 @@ msgstr "The editors express their appreciation of the reviewers for {$year} list
 
 msgid "common.editorialHistory"
 msgstr "Editorial History"
+
+msgid "common.editorialHistory.page"
+msgstr "Editorial History Page"
+
+msgid "common.editorialHistory.page.description"
+msgstr "This section lists past contributors."
 
 msgid "common.name"
 msgstr "Name"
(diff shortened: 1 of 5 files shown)


locale/en/common.po
```diff
@@ -160,6 +160,12 @@ msgstr "The editors express their appreciation of the reviewers for {$year} list
 msgid "common.editorialHistory"
 msgstr "Editorial History"
 
+msgid "common.editorialHistory.page"
+msgstr "Editorial History Page"
+
+msgid "common.editorialHistory.page.description"
+msgstr "Here, you will find a comprehensive record of all past contributors associated with this journal. This page highlights our commitment to transparency and showcases our editorial evolution."
```
asmecher, line 167: Locale keys in `pkp-lib` shouldn't reference OJS-specific language, and I think the style is a little too advertorial for a platform. I think it's OK for us to be activist in the PFL, but this is built into OJS/OMP/OPS. I suggest we either use something neutral:

> This area provides a comprehensive record of past contributors.

...or if that's not helpful to OJS/OMP/OPS users in general, that we don't include a description.

bozana: OK, I would change it like that, if @Devika008 has nothing against it?

bozana: If it would be me, I would not have any description on this page -- I am a minimalist regarding descriptions/explanations -- only just really needed :sweat_smile:

Devika008: We can make it really simple with

Heading: View Editorial History
Description Text: This section lists past contributors.

My intention of including a hint text below the description is to set the intention of the page for the reader since these are new additions. If you both feel that its not necessary we can really get away with it.

PR #10100, opened 2024-06-19
Fix pkp/pkp-lib/10020 - Long publication title stops submit/review process

taslangraham:
For pkp#10020

### DIFF
--- a/classes/log/EmailLogDAO.php
+++ b/classes/log/EmailLogDAO.php
@@ -22,11 +22,19 @@ use APP\facades\Repo;
 use Illuminate\Database\Query\Builder;
 use Illuminate\Database\Query\JoinClause;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
 use PKP\db\DAOResultFactory;
 use PKP\plugins\Hook;
 
 class EmailLogDAO extends \PKP\db\DAO
 {
+    /**
+     * The maximum length for the email subject.
+     *
+     * This value should match the length of the `subject` column in the `email_log` table, defined in LogMigration.php.
+     */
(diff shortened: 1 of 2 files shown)


classes/log/EmailLogDAO.php
```diff
@@ -157,7 +159,7 @@ public function insertObject($entry)
                 $entry->getRecipients(),
                 $entry->getCcs(),
                 $entry->getBccs(),
-                $entry->getSubject(),
+                $this->_truncateEmailSubjectLine($entry->getSubject()),
```
asmecher, line 162: I'd suggest using [`Str::limit`](https://laravel.com/docs/11.x/strings#method-str-limit) rather than writing a custom function. (This will also solve the problem below that you're using the constant in some parts, but still have a hard-coded `252` in others.)

classes/log/EmailLogDAO.php
```diff
@@ -27,6 +27,8 @@
 
 class EmailLogDAO extends \PKP\db\DAO
 {
+    private int $MAX_SUBJECT_LENGTH = 255;
```
asmecher, line 30: Make this a `const` rather than a `private int`; also add a comment that this should match the `email_log` table's `subject` column length. Likewise, add a comment to `lib/pkp/classes/migration/install/LogMigration.php` to mention that the length should be kept the same as `EmailLogDAO::MAX_SUBJECT_LENGTH`.

asmecher (changes requested):
A couple changes to suggest, then please go ahead and forward-port as needed!

classes/log/EmailLogDAO.php
taslangraham: updated

PR #10109, opened 2024-06-22
Replace homebrew caching with Laravel tools

asmecher:


classes/navigationMenu/NavigationMenuDAO.php
```diff
@@ -64,40 +53,26 @@ public function getById($navigationMenuId, $contextId = null)
 
     /**
      * Retrieve a navigation menu by context Id.
-     *
-     * @param int $contextId Context Id
-     *
-     * @return DAOResultFactory<NavigationMenu>
```
jonasraoni, line 70: @asmecher It's better to leave these specialized returns (`DAOResultFactory<NavigationMenu>`) in place, on the calling site a good editor will be able to understand that each item of the collection is a `NavigationMenu`. I could even augment the "*DAO" classes with such annotations, but as they are deprecated, I didn't do it.

And when leaving the doctype, it's needed to take care with this:
image

asmecher: It's probably a good idea to talk new standards for docblocks.

I don't use an IDE that guides with typehints, but I have been seeing a general shift away from redundancy in typehints. See e.g. [optional `@param` types in phpstan](https://github.com/phpstan/phpdoc-parser/issues/77#issuecomment-1151075989), which is paired with phpDocumentor. By my read, phpDocumentor currently considers...

- The type in `@param` is optional
- The type in `@return` is mandatory ([reference](https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/return.html#param))
- The type in `@var` is mandatory ([reference](https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/var.html#param)

My goal in recent changes is to keep phpdoc block short and sweet. We had a lot of docblocks that added nothing, e.g.:
```php
/**
 * Get the xyz by context id.
 * @param int $contextId Context ID
 * @return DAOResultFactory
 */
function getXyzByContextId($contextId) {
   ...
}
```
I would much rather see...
```php
/**
 * Get the xyz by context id.
 */
function getXyzByContextId(int $contextId) : DAOResultFactory
{
   ...
}
```
This has no redundancy (=less opportunity for mismatches) and no unhelpful content.

If I had my druthers :) we would not even be forced to add types to `@return` or `@var` -- but I see they're mandatory, probably because it's impossible to distinguish the first word of the description from the type without a `$varName` to look for as separator. So we have to keep those.

I agree that e.g. `DAOResultFactory<NavigationMenu>` does enrich things, so I'll stop removing those.

jonasraoni: Yeah! I agree with removing trivial comments. I'm trying to do the same, I'm just leaving superfluous method descriptions "Get the ID" for `function getId(){}` 😁

About the `@param`, I think it's better to keep it: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#510-param, my code editor is following the standard.

PR #10122, opened 2024-06-25
Add masthead for userGroups.dtd

withanage:


withanage:
Hi @bozana , 
i have seen you fixed this for main. Just backported for 3.3 to fix the github actions. Can you have a look

bozana:
@withanage, I think it is not needed elsewhere because masthead was introduced recently and only in the main branch...

Issue #10135, opened 2024-06-27
Add HTML decoding in getLocalizedTitle and getLocalizedSubTitle in PKPPublication


bozana:
When required/wished format of the publication title and subtitle is 'text' the string also needs to be HTML decoded (htmlspecialchars_decode).
See https://github.com/pkp/citationStyleLanguage/pull/122#discussion_r1657199912

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/10136
- https://github.com/pkp/ojs/pull/4344
- https://github.com/pkp/omp/pull/1617 (only submodule update)
- https://github.com/pkp/ops/pull/721 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/10198
- https://github.com/pkp/ojs/pull/4366
- https://github.com/pkp/omp/pull/1635 (only submodule update)
- https://github.com/pkp/ops/pull/730 (only submodule update)

bozana:
@touhidurabir, could you review the pkp-lib PR above? If OK, I would merge (so that I can fix another bug in citationStyleLanguage plugin first) and then I would take another look at the code to see if we can remove PKPString::html2text used together with publication title and subtitle in text format.
Thanks a lot!

PR #10136, opened 2024-06-27
pkp/pkp-lib#10135 add htmlspecialchars_decode for publication title a…

bozana:
s. https://github.com/pkp/pkp-lib/issues/10135

### DIFF
--- a/classes/publication/PKPPublication.php
+++ b/classes/publication/PKPPublication.php
@@ -26,11 +26,11 @@ use PKP\facades\Locale;
 use PKP\userGroup\UserGroup;
 
 class PKPPublication extends \PKP\core\DataObject
-{    
+{
     /**
      * Get the default/fall back locale the values should exist for
      */
-    public function getDefaultLocale(): ?string 
+    public function getDefaultLocale(): ?string
     {
         return $this->getData('locale');
     }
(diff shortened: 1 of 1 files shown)


classes/publication/PKPPublication.php
```diff
@@ -98,7 +98,7 @@ public function getLocalizedTitle($preferredLocale = null, string $format = 'tex
                 break;
             case 'text':
                 // Title is in HTML, prefix is already in text. Convert title.
-                $title = strip_tags($title);
+                $title = htmlspecialchars_decode(strip_tags($title));
```
touhidurabir, line 101: @bozana I understand the intension and purpose of this change and I think it's ok . However if I am not mistaken and remember correctly, we probably omit the use of `htmlspecialchars_decode` intentionally so that even text version slips into somewhere without the use of `htmlspecialchars` or `strip_unsafe_html`, it will not be abused.  . So developer will be forced to use the `htmlspecialchars_decode` manually as necessary and knowingly . But allowing text version as decoded html special chars, we need to be careful to use `htmlspecialchars` and/or `strip_unsafe_html` if used in template directly . 

But I think it's ok to add this and just need to extra careful of the usage of text version .

bozana: Hmmm... that sounds reasonable...
@asmecher and @jonasraoni, what do you think?

asmecher: If someone uses a `text` form of the title in the context of an HTML presentation, they'll have to use `escape` (Smarty) or `htmlspecialchars` (PHP)`. But they should use the `html` form instead in that situation. In any case, this is consistent with our plaintext escaping needs/strategy elsewhere in the system, so I'm comfortable with it. :+1:

Issue #10145, opened 2024-06-28
Consider TSV as API response


bozana:
Within the issue https://github.com/pkp/pkp-lib/issues/9666 -- that is the migration of the issue https://github.com/pkp/pkp-lib/issues/8248 to the main -- the TSV as API response needs to be implemented.

S. also: https://github.com/pkp/pkp-lib/commit/2f9966308e3e69b9f5b125a9f20f5c1846dbf53f

PRs:
- https://github.com/pkp/pkp-lib/pull/10146
- https://github.com/pkp/ojs/pull/4345
- https://github.com/pkp/omp/pull/1621 (only submodule update)
- https://github.com/pkp/ops/pull/722 (only submodule update)

bozana:
@touhidurabir, could you please review the PR above? I can gladly solve it differently -- e.g. put it at another place -- if needed.
Thanks a lot!

bozana:
@touhidurabir, I have now changed it accordingly. Could you please take another look?
Thanks a lot!

touhidurabir:
@bozana looks great . should be good to merge .

PR #10146, opened 2024-06-28
pkp/pkp-lib#10145 consider TSV as API response

bozana:
s. https://github.com/pkp/pkp-lib/issues/10145


touhidurabir:
@bozana this look good . But I think what the macro method is doing more than what it specified as by name . I am wondering if we should normalize the method behaviour with update of name like `withFile` and have the const defined as 
```php
public const RESPONSE_CSV  = [
'mime' => ''text/csv'',
'extension' => '.csv',
'separator' => ','
]
```

what you think ? 

But if you think it's too much work, it's fine as you have modified .

bozana:
Hi @touhidurabir, I think your suggestions are good. I will do it that way...
Thanks a lot!

Issue #10129, opened 2024-06-25
Port EmailLogDao and EmailLogEntry to Eloquent Model


taslangraham:
**Describe the change**
Right now the EmailLogEntry entity is a DataObject and it follows the DAO pattern.

We will port EmailLogEntry entity and its respective DAO class, and related sub-classes to Laravel's Eloquent Model..

**What application will be affected?**
OJS, OMP or OPS main branch

### PRs:
**pkp-lib main** -  https://github.com/pkp/pkp-lib/pull/10150/files
- Introduced [eloquence](https://github.com/kirkbushell/eloquence) package to allow easy camelCase access to Laravel eloquent model attributes
- Converted EmailLogEntry class to Laravel eloquent model
- Replaced DAOs with a repository
- Replaced child classes of `EmailLogEntry` with enums that contain app-specific values

**OMP main** - https://github.com/pkp/omp/pull/1632/files
**OJS main** - https://github.com/pkp/ojs/pull/4369/files
**OPS main** - https://github.com/pkp/ops/pull/732/files

taslangraham:
Ready for review @Vitaliy-1 @asmecher 
pkp-lib main - https://github.com/pkp/pkp-lib/pull/10150/files
OMP main - https://github.com/pkp/omp/pull/1632/files

asmecher:
Thanks, @taslangraham, I've added [a review here](https://github.com/pkp/pkp-lib/pull/10150#pullrequestreview-2175998033)!

asmecher:
Added a comment over on https://github.com/pkp/pkp-lib/pull/10150#issuecomment-2232030186, thanks, @taslangraham!

taslangraham:
All updated @asmecher 
- https://github.com/pkp/omp/pull/1632/commits/914c87ce0cbae3f478de0f38439f340b2c743a10#diff-d6dbec6efe4425466bdbd9bd101a861b698709f6fd33f1a3a2d1c3d86dde3c60
- https://github.com/pkp/pkp-lib/pull/10150/commits/e6f110bcd5151b3f9fdc67cbbe86d7f9df667336#diff-0b46927ec2829a68ba70e80024a7073623a16c2ea4ddea6abf7ad5bb0f8113a4

asmecher:
Merged all -- thanks, @taslangraham! I made two additional changes, not wanting to hold up the merge:
[...]

PR #10150, opened 2024-07-01
Port EmailLogDao and EmailLogEntry to Eloquent Mode

taslangraham:
For #10129


classes/log/EmailLogEntry.php
```diff
@@ -133,17 +171,18 @@ public function setAssocId($assocId)
      */
     public function getSenderFullName()
```
Vitaliy-1, line 172: How do you think, would implementing accessor would be better for cases like this? E.g., something like:
```php
public function senderFullName()
    {
        return new Attribute(
            get: fn () => ...
        );  
    }
```

Vitaliy-1 (changes requested):
Thanks, @taslangraham!

I left a couple comments to think about regarding the concept in general. They don't have examples in our codebase. Let me know if you'll need some help or more clarification on this.

classes/log/SubmissionEmailLogEntry.php
```diff
@@ -18,6 +18,18 @@
...
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Support\Facades\Mail;
+use PKP\core\Core;
+use PKP\db\DAOResultFactory;
+use PKP\facades\Locale;
+use PKP\facades\Repo;
+use PKP\mail\Mailable;
+use PKP\user\User;
+
 class SubmissionEmailLogEntry extends EmailLogEntry
```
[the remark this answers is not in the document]
taslangraham: Based on the current structure, a single model could work. I'm not entirely sure what the original vision behind having separate classes like `SubmissionEmailLog` or `MonographFileEmailLog` was. An advantage I see is that each class maintains its own `assocType` (`assoc_type` in the database table) value. This helps when performing queries, as we can get the value of the assoc type directly from the class. However, passing that value to the methods performing the queries (when using a single class) doesn't seem like it would be a major issue.

classes/log/EmailLogEntry.php
```diff
@@ -18,112 +18,150 @@
...
+        );
     }
 
-    /**
-     * Get date email was sent.
-     *
-     * @return string
-     */
-    public function getDateSent()
+    protected function senderId(): Attribute
```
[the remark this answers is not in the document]
taslangraham: I don't have a clear idea in mind, but I can do some research and experiment.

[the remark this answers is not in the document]
Vitaliy-1: E.g., as a concept:
```php
public function __get($key)
{
    $attributes = $this->toArray();
    $snakedKey = Str::snake($key);
    if (key_exists($snakedKey, $attributes)) {
        return $this->getAttribute($snakedKey);
    }
    return parent::__get($key);
}
```
But I'm curious why toArray() doesn't return camelcase attributes even if `$snakeAttributes` Model property is set to false...

classes/log/SubmissionEmailLogEntry.php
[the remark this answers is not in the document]
Vitaliy-1: Yeah, and we can transform classes that hold app specific log-related constants into enums and use just for it. @asmecher knows historical reasons of keeping separate email log classes better, what do you think about their unification into a single model?

classes/log/EmailLogEntry.php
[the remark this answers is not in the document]
Vitaliy-1: I found a library that does a pretty good job on using camel case with Eloquent attributes: https://github.com/kirkbushell/eloquence/blob/master/src/Behaviours/HasCamelCasing.php. It overrides all sensitives places and looks like a more elegant solution to our current approach.

classes/log/SubmissionEmailLogEntry.php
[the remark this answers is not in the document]
asmecher: I'd support reconciling things into a single class. The abstraction isn't worth defending if it's going to make things awkward/verbose with Eloquent.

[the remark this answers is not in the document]
taslangraham: I'll proceed with refactoring the code to use a single class, as it simplifies some things.

Here’s an example of how I was managing class-specific values:

Initialize the value in the constructor of the child class:
https://github.com/pkp/pkp-lib/blob/11f2069f6fa6c08192891d80a01bb66cdb936e17/classes/log/SubmissionEmailLogEntry.php#L88-L91

Add the methods that used such value as a static method on the parent class:
https://github.com/pkp/pkp-lib/blob/11f2069f6fa6c08192891d80a01bb66cdb936e17/classes/log/EmailLogEntry.php#L252-L255
Static method accesses the value like ```
$instance->assocType```

Use method via child class like ```SubmissionEmailLogEntry::getByEventType(...args).```

classes/log/EmailLogEntry.php
taslangraham: updated

classes/log/EmailLogEntry.php
```diff
@@ -11,230 +11,143 @@
...
-    /**
-     * Set associated type.
-     *
-     * @param int $assocType
-     */
-    public function setAssocType($assocType)
+    //
+    // Accessors / Mutators
+    //
+    protected function id(): Attribute
```
taslangraham, line 69: I kept this accessor/mutator for the primary key `log_id` because `$email->id` seems more standard and intuitive than `$email->logId`.

classes/log/EmailLogEntry.php
```diff
@@ -11,230 +11,143 @@
...
-
-    public function setSubject($subject)
+    public function scopeWithAssocId(Builder $query, int $submissionId = null): Builder
     {
-        $this->setData('subject', $subject);
+        return $query->where('assoc_id', (int)$submissionId);
     }
 
-    public function getBody()
+    public function scopeWithSenderId(Builder $query, $senderId): Builder
```
asmecher, line 132: `$senderId` should be typehinted `?int`.

classes/log/EmailLogEntry.php
```diff
@@ -11,230 +11,143 @@
...
-        $this->setData('bccs', $bccs);
-    }
-
-    public function getSubject()
-    {
-        return $this->getData('subject');
-    }
-
-    public function setSubject($subject)
+    public function scopeWithAssocId(Builder $query, int $submissionId = null): Builder
```
asmecher, line 127: Use `$assocId` rather than `$submissionId` -- we don't know for sure here that an assoc ID will be a submission ID. Also, I'm not 100% sure this won't have knock-on effects, but I'd suggest using a strict `int` rather than `?int`. Elsewhere, `null`s in filter functions are intended to mean "do not filter on this column", and allowing `null` here to get silently cast into an `int` might lead to confusion later. (Just FYI, when we do explicitly want to filter by a `null` value, we typically use functions that take arrays -- in this case it would be `scopeWithAssocIds` -- and accept a `[null]` filter value to explicitly mean "I want to select by null values".)

classes/log/EmailLogEntry.php
```diff
@@ -11,230 +11,143 @@
...
+        });
     }
 
-    /**
-     * Returns the subject of the message with a prefix explaining the event type
-     *
-     * @return string Prefixed subject
-     */
-    public function getPrefixedSubject()
+    public function scopeWithAssocType(Builder $query, $assocType): Builder
```
asmecher, line 146: Use typehinted `int` for `assocType`.

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+use PKP\log\core\EmailLogEventType;
+use PKP\mail\Mailable;
+use PKP\user\User;
+use Illuminate\Support\Facades\DB;
+
+class Repository
+{
+    private EmailLogEntry $model;
+
+    public function __construct(EmailLogEntry $model)
```
asmecher, line 33: (feel free to use constructor promotion if you like)

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param int $assocType
+     * @param int $assocId
+     * @return int The number of affected rows.
+     */
+    public function deleteByAssoc(int $assocType, int $assocId): int
+    {
+        return $this->model
+            ->newQuery()
+            ->where('assoc_type', (int)$assocType)
```
asmecher, line 95: Casts are redundant here now that we have typehinting.

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+                    ['email_log_id' => $entry->id, 'user_id' => $user->getId()]
+                );
+            }
+        }
+    }
+
+    /**
+     * Delete all log entries for an object.
+     *
+     * @param int $assocType
```
asmecher, line 87: These two `@param` phpdocs add nothing that isn't already in the typehint; remove them. (Also elsewhere.)

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param int $oldUserId
+     * @param int $newUserId
+     */
+    public function changeUser($oldUserId, $newUserId)
+    {
+        return [
+            DB::update(
+                'UPDATE email_log SET sender_id = ? WHERE sender_id = ?',
+                [(int) $newUserId, (int) $oldUserId]
```
asmecher, line 111: Make the two params typehinted and remove the casts. (Also elsewhere)

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Transfer all log and log users entries to another user.
+     *
+     * @param int $oldUserId
+     * @param int $newUserId
+     */
+    public function changeUser($oldUserId, $newUserId)
+    {
+        return [
+            DB::update(
```
asmecher, line 109: New code should be converted to Laravel query construction.

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+        $this->model->bccRecipients = $this->getContactString($clonedMailable->bcc);
+        $this->model->body = $clonedMailable->render();
+        $this->model->assocType = Application::ASSOC_TYPE_SUBMISSION;
+        $this->model->subject = Mail::compileParams(
+            $clonedMailable->subject,
+            $clonedMailable->getData(Locale::getLocale())
+        );
+
+        $this->model->save();
+        $this->insertLogUserIds( $this->model);
```
asmecher, line 156: The space after the `(` suggests the automatic code formatter isn't configured. This should be running before commits are allowed. Check out the documentation for Captain Hook, which is used to kick this off in pkp-lib.

classes/log/Repository.php
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Get email log entries by assoc ID, event type and assoc type
+     *
+     * @param int $assocId
+     * @param EmailLogEventType $eventType Type of email event
+     * @param int $assocType
+     * @param ?int $userId optional Return only emails sent to this user.
+     */
+    function getByEventType(int $assocId, EmailLogEventType $eventType, int $assocType, ?int $userId = null)
```
asmecher, line 169: Is it possible to remove this function and rely instead on Eloquent operations?

classes/log/SubmissionEmailLogEntry.php
```diff
@@ -7,131 +7,69 @@
...
- *
- * @brief Base class for describing an entry in the submission email log.
+ * @brief Enumeration for submission email log event types.
  */
 
 namespace PKP\log;
+use PKP\log\core\EmailLogEventType;
 
-class SubmissionEmailLogEntry extends EmailLogEntry
+enum SubmissionEmailLogEntry: int implements EmailLogEventType
```
asmecher, line 20: I'd suggest renaming these enums to reflect their changed relationship with the `EmailLogEntry` model. This should be something like `SubmissionEmailLogEntryType`.

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -235,10 +234,10 @@ public function sendMessage(int $userId, Submission $submission, Request $reques
         );
 
         $mailable->allowUnsubscribe($notification);
-        $logDao = null;
+        $log = null;
         try {
             Mail::send($mailable);
-            $logDao = DAORegistry::getDAO('SubmissionEmailLogDAO'); /** @var SubmissionEmailLogDAO $logDao */
+            $log =  Repo::emailLogEntry();
```
asmecher, line 240: Call this `$logRepository` rather than `$log`; at first glance I thought this was a log entry.

classes/log/EmailLogEntry.php
asmecher: (Further to this -- you'll note below that the `deleteByAssoc` function requires `int` parameters, not `?int`. It would be risky in that case to allow `?int` parameters and either cast them to zero, or [use conditional queries](https://github.com/pkp/pkp-lib/pull/10150/files#diff-fb677884b3042d12be73a0449956cc5ea712654c3ba980d1242728b288db663eR141), because user-supplied inputs might pass `null` through and suddenly affect a whole range of entries rather than an intended one. Let's try to stick consistent with that convention, which is safer.)

asmecher (changes requested):
Lots of relatively small comments and suggestions -- thanks, @taslangraham!

classes/log/Repository.php
taslangraham: updated

classes/log/Repository.php
taslangraham: p.s: after an earlier discussion, we've decided to keep this function for now.

taslangraham:
@asmecher I've made the requested updates.
Changes:
- https://github.com/pkp/pkp-lib/pull/10150/commits/92624847e040e4fe49c41f0c369aa8fd20491ead#diff-fb677884b3042d12be73a0449956cc5ea712654c3ba980d1242728b288db663e
- https://github.com/pkp/pkp-lib/pull/10150/commits/92624847e040e4fe49c41f0c369aa8fd20491ead#diff-0669bbe1e9180ffab8f8706eb5a75703bc3a8dc3b8910ee0265ef2b5be11fc41
- https://github.com/pkp/pkp-lib/pull/10150/commits/92624847e040e4fe49c41f0c369aa8fd20491ead#diff-5ccff4c4028da3572134aec0125d92ac1924c24c93dd124a9ce1049eebc04fd3

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
taslangraham: update

classes/log/SubmissionEmailLogEntry.php
taslangraham: updated

taslangraham:
Tests are failing on the corresponding OMP PR. I'm looking into it.

taslangraham:
@asmecher I found and fixed a few issues after the initial review. You can check out the changes in this [commit](https://github.com/pkp/pkp-lib/pull/10150/commits/a434ffc305e451e86bb3a6dca59ed433da7249b9).

asmecher:
@taslangraham, that looks great to me -- ready to go from my perspective. This still needs porting to the other apps, right? Please rebase this to resolve the merge conflict on the composer dependencies, then I'd say it's ready for porting and merge.

taslangraham:
@asmecher I've resolved all conflicts and created PRs for submodule updates for each app. Please see https://github.com/pkp/pkp-lib/issues/10129 for PRs

Issue #9810, opened 2024-03-20
Error while upgrading to OJS 3.4 if the journal's URL/domain has been changed


jonasraoni:
**Describe the bug**
At the `I8508_ConvertCurrentLogFile` migration, the error below happens if the `config.inc.php` isn't using the production URL.

`Error: File /files/usageStats/usageEventLogs/usage_events_20240319.log could not be successfully converted.`

1. As a sysadmin, I might decide to upgrade the system under a temporary host, and this error is unexpected to me and tough to reason about.
  Just a warning like `ensure you've got the right base_url before migrating` might not be enough, so I think we can be more pro-active and attempt to convert the file at this point (at least a couple of lines to ensure that it's ok):
  https://github.com/pkp/pkp-lib/blob/3db74ed94aff4ccc767c5396c0ccfe8f2c38bfba/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php#L206

2. As the code has a couple of `continue`s, I think it's better to ensure that **at least one entry really failed** before assuming that the file couldn't be processed (e.g. if the file just has one entry, an [internal apache request](https://github.com/pkp/pkp-lib/blob/3db74ed94aff4ccc767c5396c0ccfe8f2c38bfba/classes/cliTool/traits/ConvertLogFile.php#L148-L150) then the tool will not process anything and mistakenly assume that it failed).

**To Reproduce**
1. Setup a different `base_url`
2. Attempt to upgrade with a pending log file at the `{$files_dir}/usageStats/usageEventLogs`
3. See error

**What application are you using?**
OJS 3.4

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/10152
- https://github.com/pkp/ojs/pull/4348 (only submodule update)
- https://github.com/pkp/omp/pull/1623 (only submodule update)
- https://github.com/pkp/ops/pull/723 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/10182

jonasraoni:
@bozana check if it makes sense what I wrote.

bozana:
Hi @jonasraoni, 
1. Yes, the right URL is needed, because we extract the context path, page, operation, args from the URL. For this we first remove the base URL (s. https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/cliTool/traits/ConvertLogFile.php#L320). 
And, yes, maybe we can improve the preflight check. It will need some code, but...
[...]

PR #10152, opened 2024-07-02
pkp/pkp-lib#9810 check if the URL in the usage stats log file corresp…

bozana:
…onds to the configured base_url

s. https://github.com/pkp/pkp-lib/issues/9810

### DIFF
--- a/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
+++ b/classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
@@ -17,6 +17,7 @@ namespace PKP\migration\upgrade\v3_4_0;
 use APP\core\Application;
 use APP\migration\upgrade\v3_4_0\MergeLocalesMigration;
 use APP\statistics\StatisticsHelper;
+use DateTime;
 use Exception;
 use Illuminate\Database\MySqlConnection;
 use Illuminate\Database\PostgresConnection;
@@ -25,6 +26,7 @@ use Illuminate\Database\Query\JoinClause;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
+use PKP\config\Config;
 use PKP\db\DAORegistry;
 use Throwable;
 
@@ -94,6 +96,7 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
 
     /**
      * Check the contexts' contact details before upgrade
+     *
      * @see https://github.com/pkp/pkp-lib/issues/8183
      *
      * @throws Exception
@@ -128,6 +131,7 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
 
     /**
      * Ensures locale conflicts won't happen at a later stage of the migration
+     *
      * @see https://github.com/pkp/pkp-lib/issues/8598
      *
      * @throws Exception
@@ -204,6 +208,7 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
     protected function checkUsageStatsLogs(): void
     {
         $usageStatsDir = StatisticsHelper::getUsageStatsDirPath();
+        $filePathToCheck = '';
         // check if there are usage stats log files older than yesterday
         foreach (glob($usageStatsDir . '/usageEventLogs/*') as $usageStatsLogFile) {
             if (!preg_match('/(\d{8})\.log$/', $usageStatsLogFile, $logFileDate)) {
@@ -214,6 +219,7 @@ abstract class PreflightCheckMigration extends \PKP\migration\Migration
             if ($yesterday > $logFileDate) {
                 throw new Exception("There are unprocessed log files from more than 1 day ago in the directory {$usageStatsDir}/usageEventLogs/. This happens when the scheduled task to process usage stats logs is not being run daily. All logs in this directory older than {$yesterday} must be processed or removed before the upgrade can continue.");
             }
+            $filePathToCheck = $usageStatsLogFile;
         }
         // check if there are old usage stats log files there that were not successfully processed
         if (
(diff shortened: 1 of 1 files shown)


classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -368,7 +449,7 @@ protected function checkForeignKeySupport(): void
         );
 
         if (count($result) > 0) {
-            error_log(print_r($result,true));
+            error_log(print_r($result, true));
```
jonasraoni, line 452: I've seen this before, and I forgot to remove it. I think it shouldn't be here.

bozana: I can do it.. What exactly should or should not happen?

bozana: Or to maybe open a new issue for that?

jonasraoni: I think this can be safely erased, someone was probably just debugging and forgot it here, the `$result` is used/displayed to the user below.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -223,10 +229,80 @@ protected function checkUsageStatsLogs(): void
...
+
+    /**
+     * Check if the domains in the log file URLs are correct i.e. the same as the base_url in the config.inc.php.
+     * Only the first URL in the log file is checked (and the same assumed for all URLs in the log file).
+     *
+     * @throws Exception
+     */
+    protected function checkUsageStatsLogFileUrl(string $filePathToCheck): void
+    {
+        $fhandle = fopen($filePathToCheck, 'r');
```
jonasraoni, line 244: In general, I think it's better to use the class based `fopen` (`SplFileObject`), we probably also have some libraries to deal with files, but to initialize them doesn't seem to be worth, see: https://github.com/pkp/pkp-lib/blob/9764e3f2538df94211b30a1bf1a3110bcef24caa/classes/migration/upgrade/v3_4_0/I8073_RemoveNotesWithoutQueriesAndRelatedObjects.php#L38-L54).

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -223,10 +229,80 @@ protected function checkUsageStatsLogs(): void
...
+            $baseUrls = array_merge(Config::getContextBaseUrls(), [$configBaseUrl]);
+            foreach ($baseUrls as $baseUrl) {
+                if (str_contains($entryData['url'], $baseUrl)) {
+                    return;
+                }
+            }
+            break;
+        }
+        fclose($fhandle);
+        throw new Exception('The base_url in config.inc.php should be the same as in URLs in the usage stats log file.');
```
jonasraoni, line 279: I think this exception should live inside the loop, on the place of the `break`. Otherwise, an empty file or a file with invalid entries (e.g. '*', bad status code, ...) will emit the exception.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -223,10 +229,80 @@ protected function checkUsageStatsLogs(): void
...
+    /**
+     * Get data from the passed log entry.
+     */
+    protected function getDataFromLogEntry(string $entry): array
+    {
+        $entryData = [];
+        // The default regex that can parse the usageStats plugin's log files.
+        $parseRegex = '/^(?P<ip>\S+) \S+ \S+ "(?P<date>.*?)" (?P<url>\S+) (?P<returnCode>\S+) "(?P<userAgent>.*?)"/';
+        if (preg_match($parseRegex, $entry, $m)) {
+            $associative = count(array_filter(array_keys($m), 'is_string')) > 0;
```
jonasraoni, line 291: This can be removed. Our version of PHP will always provide the named groups. And if it didn't, it would be better to use the integer indexed anyway.

jonasraoni (changes requested):
@bozana I left some comments, but the only thing that I consider important is moving the Exception inside the loop.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
bozana: I will then use SplFileObject

bozana:
Hi @jonasraoni, I have considered all the comments. Could you please take another look -- you can see the second commit for changes.
Thanks a lot!

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -288,14 +290,13 @@ protected function getDataFromLogEntry(string $entry): array
         // The default regex that can parse the usageStats plugin's log files.
         $parseRegex = '/^(?P<ip>\S+) \S+ \S+ "(?P<date>.*?)" (?P<url>\S+) (?P<returnCode>\S+) "(?P<userAgent>.*?)"/';
         if (preg_match($parseRegex, $entry, $m)) {
-            $associative = count(array_filter(array_keys($m), 'is_string')) > 0;
-            $entryData['ip'] = $associative ? $m['ip'] : $m[1];
-            $time = $associative ? $m['date'] : $m[2];
+            $entryData['ip'] = $m[1];
```
jonasraoni, line 293: Oh, I personally like the named groups (you don't need to update indexes if something changes), as you're not using them, it's possible to remove the `?P<identifier>` from the regexp. But I personally would leave (it also works as a kind of documentation).

jonasraoni (approved):

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
bozana: Ah, I understood it the other way around -- that numbered indexes are better :sweat_smile:

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -288,14 +290,13 @@ protected function getDataFromLogEntry(string $entry): array
         // The default regex that can parse the usageStats plugin's log files.
         $parseRegex = '/^(?P<ip>\S+) \S+ \S+ "(?P<date>.*?)" (?P<url>\S+) (?P<returnCode>\S+) "(?P<userAgent>.*?)"/';
         if (preg_match($parseRegex, $entry, $m)) {
-            $associative = count(array_filter(array_keys($m), 'is_string')) > 0;
-            $entryData['ip'] = $associative ? $m['ip'] : $m[1];
-            $time = $associative ? $m['date'] : $m[2];
+            $entryData['ip'] = $m['ip'];
```
bozana, line 293: @jonasraoni, OK this way?

Issue #10133, opened 2024-06-26
Port Genre and GenreDAO to Use Eloquent Model


Hafsa-Naeem:
## Issue Description
Currently, `Genre` and `GenreDAO` utilize a traditional DAO framework. These should be ported to use Laravel's Eloquent model.

## What application are you using?
 OJS, OMP or OPS `main` branch

## Additional information
 [Laravel Eloquent Documentation](https://laravel.com/docs/11.x/eloquent)

Hafsa-Naeem:
Ready for review @asmecher
**PR**
OJS main: [Port Genre and GenreDAO to Use Eloquent Model #10133](https://github.com/pkp/pkp-lib/issues/10133
)

Hafsa-Naeem:
Ready for review @asmecher @Vitaliy-1 
ojs_main [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #4910](https://github.com/pkp/ojs/pull/4910)
pkp-lib [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #11441](https://github.com/pkp/pkp-lib/pull/11441)
oaiJats [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #77](https://github.com/pkp/oaiJats/pull/77)
crossref-ojs [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #69](https://github.com/pkp/crossref-ojs/pull/69)

asmecher:
Thanks, @Hafsa-Naeem! I'll let @Vitaliy-1 review this first -- but I do see that the OJS tests are not currently passing.

Hafsa-Naeem:
Ready for review @Vitaliy-1 
ojs_main [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #4910](https://github.com/pkp/ojs/pull/4910)
pkp-lib [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #11441](https://github.com/pkp/pkp-lib/pull/11441)
oaiJats [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #77](https://github.com/pkp/oaiJats/pull/77)
crossref-ojs [pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model #69](https://github.com/pkp/crossref-ojs/pull/69)

[...]

PR #10194, opened 2024-07-12
pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model

Hafsa-Naeem:
for #10133


classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,93 @@
...
+
+    protected $fillable = [
+        'context_id', 'seq', 'enabled', 'category', 'dependent', 
+        'supplementary', 'required', 'entry_key'
+    ];
+
+    // Relationships
+    public function submissionFiles(): HasMany
+    {
+        return $this->hasMany(SubmissionFile::class, 'genre_id');
```
asmecher, line 31: Are you sure this should be `genre_id`?

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+
+    // Accessors and Mutators
+    protected function entryKey(): Attribute
+    {
+        return Attribute::make(
```
asmecher, line 38: Was there an equivalent to this in the previous implementation?

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,93 @@
...
+            get: fn($value) => strtoupper($value),
+            set: fn($value) => strtolower($value)
+        );
+    }
+
+    protected function enabled(): Attribute
```
asmecher, line 44: If we're just looking to cast a PHP boolean into a database int...
- It might be better to use [attribute casting](https://laravel.com/docs/11.x/eloquent-mutators#attribute-casting), or
- better still, just leave it an integer and at some point we'll declare boolean columns in the DB with the [`boolean` type](https://laravel.com/docs/11.x/migrations#column-method-boolean) so that everything matches nicely without needing casting.

classes/submission/genre/GenreDAO.php
asmecher, line 1: Will this class still be required after the move to Eloquent is completed?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,192 @@
...
+use PKP\db\XMLDAO;
+
+
+class Repository
+{
+
+    /**
+     * Get enabled genres by context ID.
+     *
+     * @param int $contextId
```
asmecher, line 16: The `@param` phpdocs don't add anything that's not already in the function declaration, so please remove them. (Here and below.)

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,192 @@
...
+        $query = Genre::where('context_id', $contextId)
+                      ->where('enabled', true)
+                      ->orderBy('seq');
+    
+        if ($rangeInfo) {
+            return $query->paginate($rangeInfo->perPage, ['*'], 'page', $rangeInfo->currentPage);
+        }
+    
+        return $query->get();
+    }
```
asmecher, line 30: There should be a blank space between function declarations.

asmecher (changes requested):
Thanks, @Hafsa-Naeem, I've added some comments but I think the biggest thing is to agree on some conventions WRT the settings table and collectors -- feel free to clean up some details, then let's make sure we've got those conventions figured out before continuing!

classes/submission/genre/GenreDAO.php
Hafsa-Naeem: @asmecher  No. it's temporary until the settings table handling is sorted out.

classes/submission/genre/Genre.php
Hafsa-Naeem: I  used this  relationship definition because in the database schema, `genre_id` is used as the foreign key in `submission_files`.

asmecher: Can you point me to the code you were looking at? I don't think it's correct.

asmecher: :+1: I misread this, thanks!

classes/submission/genre/Genre.php
asmecher: It's the uppercase/lowercase conversion I was wondering about; if it didn't exist in the previous code, best not to add it here.

Issue #9892, opened 2024-04-17
Upgrade to Laravel version 11.x


touhidurabir:
**Describe the Proposal**
With the intension to make the release of `3.5.0` a `LTS`, we should aim to have all the dependencies updated as much as possible . The current main(pre release of 3.5.0) now have Laravel 10.x as the dependency which will be get bug fix till `August 6th, 2024` .  Also by the time the 3.5.0 get to release it will be an older version of laravel . Better to upgrade the laravel version to `11.x` .

**Additional Information**
With the release of stable laravel version `11.x` at `March 12th, 2024` which seems like a more of better candidate for next LTS release as it will have a longer support period, more details at https://laravel.com/docs/11.x/releases#support-policy . The most important requirement for it is `PHP 8.2+` which is also seems like a good opportunity to ship the next LTS with more actively supported PHP version  as `PHP 8.1` will be at the end of active support by next release, more details at https://www.php.net/supported-versions.php . 

The upgrade process and details does not seems to contains many high impact changes and most of those seems to be resolvable with less effort, more details at https://laravel.com/docs/11.x/upgrade . 

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10211
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4367
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1669
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/748

asmecher:
@touhidurabir, I checked with the Technical Committee, and they didn't raise any problems with bumping our PHP base to 8.2 for OJS/OMP/OPS 3.5. They're primarily familiar with RHEL (EPEL) and Debian/Ubuntu.

touhidurabir:
@asmecher can you review it at https://github.com/pkp/pkp-lib/issues/9892#issue-2248712326 . Unit tests are still failing as we need to update the github workflow main branch to make it phpunit test command compatible with version 11 .

asmecher:
@touhidurabir, I think you're probably referring to [this change](https://github.com/pkp/pkp-github-actions/issues/60), which got merged last week; did it resolve the problem?

asmecher:
[...]

PR #10211, opened 2024-07-16
pkp/pkp-lib#9892 upgrade to laravel 11 and PHP 8.2+

touhidurabir:
pkp/pkp-lib#9892


classes/install/form/InstallForm.php
```diff
@@ -44,7 +44,8 @@ class InstallForm extends MaintenanceForm
         // <driver> => array(<php-module>, <name>)
         'mysqli' => ['mysqli', 'MySQLi'],
         'postgres9' => ['pgsql', 'PostgreSQL'],
-        'mysql' => ['mysql', 'MySQL']
+        'mysql' => ['mysql', 'MySQL'],
+        'mariadb' => ['mysql', 'MariaDB'],
```
asmecher, line 48: Should this be `mysql` inside the array, or `mysqli`?

classes/migration/install/CommonMigration.php
```diff
@@ -96,6 +96,7 @@ public function up(): void
 
         switch (DB::getDriverName()) {
             case 'mysql':
+            case 'mariadb':
```
asmecher, line 99: In situations where there's a `case` statement running through all the possible drivers, please make sure there is also a:
```php
   default: throw new \Exception('Unexpected database driver!');
```
Also elsewhere.

classes/user/Collector.php
```diff
@@ -654,9 +656,16 @@ protected function buildReviewerStatistics(Builder $query): self
...
-            ? "DATEDIFF({$dateA}, {$dateB})"
-            : "DATE_PART('day', {$dateA} - {$dateB})";
+
+        $dateDiff = fn (string $dateA, string $dateB): string => match (true) {
+            DB::connection() instanceof MySqlConnection,
+            DB::connection() instanceof MariaDbConnection
+                => "DATEDIFF({$dateA}, {$dateB})",
+            DB::connection() instanceof PostgresConnection
+                => "DATE_PART('day', {$dateA} - {$dateB})",
+            default
```
asmecher, line 666: `match` will already take care of throwing an exception when there's no matched condition and no default is specified; I'd suggest letting that take care of it. https://php.watch/versions/8.0/match-expression#must-match

asmecher (changes requested):
Thanks, I've suggested a few changes!

classes/install/form/InstallForm.php
touhidurabir: > Should this be mysql inside the array, or mysqli

should be `mysqli` .

composer.json
```diff
@@ -1,40 +1,40 @@
 {
 	"require": {
+		"php": "^8.2",
 		"components/jquery": "^3.5",
 		"components/jqueryui": "1.*",
-		"composer/semver": "^3.3",
+		"composer/semver": "^3.4",
 		"cweagans/composer-patches": "^1.7",
 		"dflydev/base32-crockford": "^1.0",
-		"doctrine/dbal": "^3.5",
```
touhidurabir, line 8: see https://laravel.com/docs/11.x/upgrade#doctrine-dbal-removal

tests/classes/config/ConfigTest.php
```diff
@@ -21,7 +21,16 @@
...
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\CoversMethod;
 
+#[CoversMethod(Config::class, 'getConfigFileName')]
+#[CoversMethod(Config::class, 'setConfigFileName')]
+#[CoversMethod(Config::class, 'reloadData')]
+#[CoversMethod(Config::class, 'getVar')]
+#[CoversMethod(Config::class, 'getData')]
+#[CoversMethod(Config::class, 'hasVar')]
+#[CoversMethod(Config::class, 'isSensitive')]
```
touhidurabir, line 33: probably we can replace it with `CoverClass` .

classes/core/PKPUserProvider.php
```diff
@@ -150,6 +152,26 @@ public function validateCredentials(UserContract $user, array $credentials)
...
+        if (!$this->hasher->needsRehash($user->getAuthPassword()) && !$force) {
+            return;
+        }
+
+        $rehash = Validation::encryptCredentials($user->getUsername(), $credentials['password']);
+        $user->setPassword($rehash);
+
+        Repo::user()->edit($user);
+    }
+
```
touhidurabir, line 174: see https://laravel.com/docs/11.x/upgrade#the-user-provider-contract

classes/core/PKPUserProvider.php
```diff
@@ -150,6 +152,26 @@ public function validateCredentials(UserContract $user, array $credentials)
...
 
+    /**
+     * Rehash the user's password if required and supported.
+     *
+     * @param  \Illuminate\Contracts\Auth\Authenticatable|\PKP\user\User  $user
+     * @param  array  $credentials
+     * @param  bool  $force
+     * @return void
+     */
+    public function rehashPasswordIfRequired(UserContract $user, #[\SensitiveParameter] array $credentials, bool $force = false)
```
asmecher, line 163: Cool, I wasn't aware of `#[\SensitiveParameter]`.

classes/core/traits/HasForeignKey.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @brief A trait to determine the existence foreign key in a table
+ *
+ */
+
+namespace PKP\core\traits;
+
+use Illuminate\Support\Facades\Schema;
+
+trait HasForeignKey
```
asmecher, line 20: I'm not sure a trait is the right place for this. It implies that the class that uses the trait has a foreign key, when actually this adds a way to check whether a database does. And in general I'm hesitant to add support code that's only called from the migrations, since migrations are throw-away (meaning they will eventually be removed when the upgrade is deemed too old to support), and when that happens we'll be left with unused traits lying around. (This is particularly a risk with traits, since they don't need any particular inheritance heirarchy and will pile up in the general purpose `traits` directory.)

I can think of two approaches:
 - Extend the Schema facade to add a general purpose `hasForeignKey` function, then call it there. If we needed them, an extended Schema facade would also be a more sensible home for some of the `Installer` functions that you've also adapted from Doctrine DBAL; they aren't currently used in `main` but have in the past (and may be again in the future) in the `upgrade.xml` script. See e.g. the `condition` attribute in this upgrade line for 3.3.0-x: https://github.com/pkp/ojs/blob/stable-3_3_0/dbscripts/xml/upgrade.xml#L24C1-L24C11. However, see my comments on those later, though -- it looks like those can just be removed entirely.
 *...or...*
 - Just continue ot use `Schema::getForeignKeys` directly, even though it'll mean a little bit of duplicated `->pluck(...)->contains(...)` code in the migrations. It's not much.

classes/install/Installer.php
```diff
@@ -854,17 +854,12 @@ public function installFilterConfig($filterConfigFile)
...
      * Check to see whether a column exists.
      * Used in installer XML in conditional checks on <data> nodes.
-     *
-     * @param string $tableName
-     * @param string $columnName
-     *
-     * @return bool
      */
-    public function columnExists($tableName, $columnName)
+    public function columnExists(string $tableName, string $columnName): bool
```
asmecher, line 858: I think `Schema::hasColumn` could be used directly in cases where this is required, so this can be removed.

classes/install/Installer.php
```diff
@@ -875,15 +870,12 @@ public function columnExists($tableName, $columnName)
     /**
      * Check to see whether a table exists.
      * Used in installer XML in conditional checks on <data> nodes.
-     *
-     * @param string $tableName
-     *
-     * @return bool
      */
-    public function tableExists($tableName)
+    public function tableExists(string $tableName): bool
```
asmecher, line 874: This is only used in `classes/install/Upgrade.php` in `migrateStaticPagesToNavigationMenuItems`, which can be modified to use `Schema::hasTable` directly -- so I think this function can be removed.

asmecher (changes requested):
Just a few suggestions!

classes/core/traits/HasForeignKey.php
touhidurabir: @asmecher extending the schema facade just for this seems a bit overkill to me . That will make to register it and maintaining it. I am wondering if this is something we can put in our core migration , that is `PKP\migration\Migration` . One file and we can remove it later if we need to . Placing following code 
```php
collect(Schema::getForeignKeys($tableName))
            ->pluck('name')
            ->contains($keyName)
```
in 10/15 places also seems too much duplication . What you say ?

PR #10215, opened 2024-07-16
invitation-api-userRoleAssignment Initial Commit

defstat:


classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
+        $emailUsed = null;
+        if (isset($userId)) {
+            $userIdUsed = $userId;
+        } else {
+            $emailUsed = $email;
         }
 
-        $this->invitationModel->userId = $userId;
+        InvitationModel::byStatus(InvitationStatus::INITIALIZED)
+            ->when($userIdUsed !== null, function ($query) use ($userIdUsed) {
```
asmecher, line 118: I prefer shorthand `fn()` syntax, and typehinted parameters. This would be e.g.:
```php
 ->when($userIdUsed !== null, fn (Builder $q) => $q->byUserId($userIdUsed))
```

classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
+    public function initialize(?int $userId = null, ?int $contextId = null, ?string $email = null, ?int $inviterId = null): void
     {
         if (!isset($userId) && !isset($email)) {
-            throw new Exception("Invitation should contain at least one user id or an invited email')");
+            throw new Exception("Invitation should contain the user id or an invited email.')");
+        }
+
+        $userIdUsed = null;
+        $emailUsed = null;
+        if (isset($userId)) {
```
asmecher, line 111: I think you could get the same effect a little more clearly with:
```php
// If a user ID is specified, give it preference.
if (isset($userId)) unset($email);
```
...and then just use `$userId` / `$email` below instead of `$userIdUsed` and `$emailUsed`.

classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
+     * fromArray function of the given Object correlation
+     */
+    protected function fillFromPayload(): void
+    {
+        if ($this->invitationModel->payload) {
+            foreach ($this->invitationModel->payload as $key => $value) {
+                if (property_exists($this, $key)) {
+                    if (property_exists($this, 'propertyType') && !empty($this->propertyType) && array_key_exists($key, $this->propertyType)) {
+                        if (is_array($value)) {
+                            $this->{$key} = array_map(function ($item) use ($key) {
```
asmecher, line 152:
```php
$this->{$key} = array_map(fn ($item) => $this->propertyType[$key]::fromArray($item), $value);
```
...but see my comment about payload data mixing into this class first.

classes/invitation/core/Invitation.php
```diff
@@ -33,18 +38,53 @@ abstract class Invitation
...
+     * Only the properties of the invitation that are added here will be considered
+     * for the payload
+     */
     protected array $payloadAccessibleProperties = [];
 
+    /**
+     * This is filled with correlation between an invitation property
+     * and the Object that this property corresponds to.
+     */
+    protected array $propertyType = [];
```
asmecher, line 67: Maybe I'm missing it, but is there an example where this isn't empty?

classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
 
-    public function fillFromArgs(array $args)
+    /**
+     * Used to fill the invitation's properties from the model's payload values.
+     * if the $propertyType has values in, then the property is filled by the 
+     * fromArray function of the given Object correlation
+     */
+    protected function fillFromPayload(): void
+    {
+        if ($this->invitationModel->payload) {
```
asmecher, line 147: I think this check is unnecessary.
...but see my comment about payload data mixing into this class first.

classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
+            foreach ($this->invitationModel->payload as $key => $value) {
+                if (property_exists($this, $key)) {
```
asmecher, line 149: Are there cases where something will be in the payload and *shouldn't* be here as a property? If not, I wonder whether it wouldn't be better to...
```php
if (!property_exists($this, $key)) {
    error_log("Payload property {$key} does not exist on object " . get_class($this));
    continue;
}
```
...but see my comment about payload data mixing into this class first.

classes/invitation/core/Invitation.php
```diff
@@ -54,53 +94,105 @@ public function __construct(?InvitationModel $invitationModel = null)
...
         $this->invitationModel->save();
     }
 
-    public function fillFromArgs(array $args)
+    /**
+     * Used to fill the invitation's properties from the model's payload values.
+     * if the $propertyType has values in, then the property is filled by the 
+     * fromArray function of the given Object correlation
+     */
+    protected function fillFromPayload(): void
```
asmecher, line 145: There's a lot of semi-magic code here to mix data from the payload into the current object, and we need to be careful about the relationship between the attributes that are already here vs. the ones that should be imported from the payload -- but we also have a class representing the payload in [UserGroupPayload](https://github.com/pkp/pkp-lib/pull/10215/files#diff-4eb6cb4ac67e37f6d58e344a20141dbf85d3b142afd1665544789390297f30f9). Might it be better to make use of the UserGroupPayload and make it a property of this class, rather than mixing data from it right into this object?

classes/invitation/core/Invitation.php
```diff
@@ -284,4 +468,30 @@ protected function getExpiryDays(): int
...
+     * so that properties of the invitation are filled before returning 
+     * the invitation object to the code.
+     * Used in InvitationFactory::getExisting.
+     */
+    public function fillCustomProperties(): void
+    {
+        return;
+    }
+
+    public function getUserId(): ?int
```
asmecher, line 483: Since the `invitationModel` is a public attribute anyway, I'd rather we reduced the number of getters in favour of just accessing that directly.

classes/invitation/invitations/changeProfileEmail/ChangeProfileEmailInvite.php
```diff
@@ -39,59 +39,68 @@ class ChangeProfileEmailInvite extends Invitation implements IBackofficeHandleab
...
-        $request = Application::get()->getRequest();
-        $site = $request->getSite();
-        $sitePrimaryLocale = $site->getPrimaryLocale();
+        $mailable = new ChangeProfileEmailInvitationNotify();
+        $mailable->recipients([$receiver]);
+        $mailable->sender($request->getUser());
+
+        $context = $request->getContext();
+
+        $contextId = 1;
```
asmecher, line 68: Test code left behind?

classes/invitation/invitations/userRoleAssignment/UserRoleAssignmentInvite.php
```diff
@@ -0,0 +1,411 @@
...
+                function ($attribute, $value, $fail) {
+                    if (empty($this->invitation->userGroupsToAdd) && empty($this->invitation->userGroupsToRemove)) {
+                        $fail(__('invitation.userRoleAssignment.validation.error.noUserGroupChanges'));
+                    }
+                }
+            ]
+        ];
+    }
+
+    public function getFinaliseValidationRules(): array
```
asmecher, line 181: `Finalise` has been changed to `Finalize` in some places, but not e.g. here.

classes/invitation/invitations/userRoleAssignment/UserRoleAssignmentInvite.php
```diff
@@ -0,0 +1,411 @@
...
+                                ])
+                            );
+                        }
+                    }
+                }
+            ]
+        ];
+    }
+
+    public function getValidationRules(): array
```
asmecher, line 214: Rather than a function return a list of validation rules, why not a `function validate()` that applies them directly when called? (It's often a lot easier to work with immediate functions when there's not a reason for the indirection. Thinking e.g. of authorization rulesets, which are a pain to debug when they aren't working as expected.)

There's also a confusing list of functions here related to validation -- `validate` (which always returns `true`), `getValidationRules`, `getInviteValidationRules`, etc. We should have a single idiom for validation, and it should be clear what rules are applied when.

classes/invitation/invitations/userRoleAssignment/UserRoleAssignmentInvite.php
```diff
@@ -0,0 +1,411 @@
...
+    public ?string $country = null;
+
+    public ?string $username = null;
+    public ?string $password = null;
+
+    public ?string $emailSubject = null;
+    public ?string $emailBody = null;
+    public ?bool $existingUser = null;
+
+    public ?array $userGroupsToAdd = null;
```
asmecher, line 69: It looks like `$userGroupsToAdd` / `$userGroupsToRemove` are sometimes arrays of `UserGroup` objects, and other times arrays of `UserGroupPayload` objects. They should be one consistent type.

asmecher (changes requested):
Hi @defstat! Just an early review on the work in progress. It's coming along well, but could use some clarity on the object model, which is a little murky.

classes/invitation/core/Invitation.php
defstat: Changed.

classes/invitation/core/Invitation.php
defstat: Agreed - changed

classes/invitation/core/Invitation.php
defstat: Removed entirely along with the `$propertyType`

locale/en/emails.po
```diff
@@ -547,3 +547,108 @@ msgstr "orcidRequestAuthorAuthorization"
...
+msgid "emails.userRoleAssignmentInvitationNotify.body"
+msgstr ""
+"<div class='email-container'>"
+"    <div class='email-header'>"
+"        <h2>Invitation to New Role</h2>"
+"    </div>"
+"    <div class='email-content'>"
+"        <p>Dear {$recipientName},</p>"
+"        <p>In light of your expertise, you have been invited by {$inviterName} to take on new roles at {$journalName}</p>"
+"        <p>At {$journalName}, we value your privacy. As such, we have taken steps to ensure that we are fully GDPR compliant. These steps include you being accountable to enter your own data and choosing who can see what information. or additional information on how we handled your data, please refer to our Privacy Policy.</p>"
```
asmecher, line 569: OJS-specific language should not be in pkp-lib.

asmecher (changes requested):
Thanks, @defstat -- just one comment needing resolution before it's ready to merge!

defstat:
Closing in favor of #10472

Issue #9987, opened 2024-05-24
Fatal exception when setting a notification as not read


jonasraoni:
**Describe the bug**
It's not possible to mark notifications as not read.

Updating the NotificationDAO::setDateRead() to accept null dates, should do it.

**To Reproduce**
1. Select a notification at the top menu
2. Click on "Mark New"
3. See error

**What application are you using?**
OJS 3.4

jonasraoni:
@taslangraham I've merged the PRs for the `pkp-lib` and closed the others (there are some conflicts).

If our policy hasn't changed (remind me to confirm this on the next call), we're not merging very often PRs that just update the submodule, to avoid creating such merge conflicts.

PR #10225, opened 2024-07-19
pkp#9987 fix issue when marking notification as unread

taslangraham:
For https://github.com/pkp/pkp-lib/issues/9987


jonasraoni (approved):
I'd just replace `string|null` by `?string`, but not a big deal 😁

Issue #9799, opened 2024-03-18
Mark review assignment when it's opened by an editor


Vitaliy-1:
After review is complete, editor is able to view review details. This event isn't reflected anywhere in the system.
`considered` review assignment property [holds 4 values](https://github.com/pkp/pkp-lib/blob/a95b221969aa35aa25540c23f8874a9226fcc49e/classes/submission/reviewAssignment/ReviewAssignment.php#L40-L43): `REVIEW_ASSIGNMENT_NEW`, `REVIEW_ASSIGNMENT_CONSIDERED`, `REVIEW_ASSIGNMENT_UNCONSIDERED`, `REVIEW_ASSIGNMENT_RECONSIDERED`. Add a new value to reflect that the review was viewed but not considered. This will help submission status tracking on the review stage.

pkp-lib: https://github.com/pkp/pkp-lib/pull/10233
tests OJS: https://github.com/pkp/ojs/pull/4682

jardakotesovec:
@Vitaliy-1 Would that also introduce new review assignment status? `REVIEW_ASSIGNMENT_STATUS_READ` or `REVIEW_ASSIGNMENT_STATUS_VIEWED`, something like that?

Vitaliy-1:
Yes, was thinking about `REVIEW_ASSIGNMENT_STATUS_VIEWED`. Another way could be just highlight on the new dashboard review assignments completed by the reviewer but without a a considered status from an editor for a certain amount of time.

Vitaliy-1:
Filed a bug that is related: https://github.com/pkp/pkp-lib/issues/10232

Vitaliy-1:
PR pkp-lib: https://github.com/pkp/pkp-lib/pull/10233

Vitaliy-1:
@jardakotesovec, do we need also to record a date when review assignment was first opened by the editor?

jardakotesovec:
@Vitaliy-1 From current UI plans that we have for review assignments popups - we don't need it. 

But what about review history? It would probably make sense to include it there, but maybe thats coming from different logging table?

<img width="851" alt="Screenshot 2024-08-06 at 15 40 12" src="https://github.com/user-attachments/assets/82be040b-f526-4e87-bdf4-4b8c1eb399e0">

Vitaliy-1:
[...]

PR #10233, opened 2024-07-22
pkp/pkp-lib#9799 Record review assignment view action

Vitaliy-1:


classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -41,6 +41,7 @@ class ReviewAssignment extends \PKP\core\DataObject
     public const REVIEW_ASSIGNMENT_CONSIDERED = 3; // Has been marked considered by an editor
     public const REVIEW_ASSIGNMENT_UNCONSIDERED = 1; // Considered status has been revoked by an editor and is awaiting re-confirmation by an editor
     public const REVIEW_ASSIGNMENT_RECONSIDERED = 2; // Considered status has been granted again by an editor
+    public const REVIEW_ASSIGNMENT_VIEWED = 4; // Editor opened a review assignment but hasn't marked it as considered yet
```
jardakotesovec, line 45: Could we also add REVIEW_ASSIGNMENT_STATUS_VIEWED?

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -682,12 +685,18 @@ public function getStatus()
...
-            if ($this->getConsidered() != self::REVIEW_ASSIGNMENT_UNCONSIDERED && $this->isRead()) { // ...and confirmed by an editor
-                return self::REVIEW_ASSIGNMENT_STATUS_COMPLETE;
+            if ($this->getConsidered() != self::REVIEW_ASSIGNMENT_UNCONSIDERED) {
+                // if it was read and confirmed by the editor
+                if ($this->isRead()) {
+                    return self::REVIEW_ASSIGNMENT_STATUS_COMPLETE;
+                    // If it was viewed by the editor but not confirmed
+                } else {
+                    return self::REVIEW_ASSIGNMENT_STATUS_VIEWED;
+                }
```
jardakotesovec, line 699: I would expect to return REVIEW_ASSIGNMENT_STATUS_VIEWED, when $this->getConsidered() === REVIEW_ASSIGNMENT_VIEWED.

But it has been long day...

Vitaliy-1: The logic here is pretty the same but I can narrow the results with `else if` instead of `else` just in case.

Issue #9812, opened 2024-03-21
The plugin gallery isn't visible at the site settings


jonasraoni:
**Describe the bug**
At the site settings, it's possible to view, enable and also install plugins from a package, but I'm unable to access the plugin gallery.

**To Reproduce**
1. Access the menu Administration > Site Settings

**What application are you using?**
OJS 3.3

**Additional information**
Site Settings
![image](https://github.com/pkp/pkp-lib/assets/361921/338b2561-0390-4d9d-9680-db6fe2858aa0)

Website Settings
![image](https://github.com/pkp/pkp-lib/assets/361921/a462aa08-21c0-4515-ab0d-f3f6ca6af8c6)

kaitlinnewson:
PR for review @asmecher and will forward-port once approved:

- pkp-lib 3_3_0: https://github.com/pkp/pkp-lib/pull/10240

This PR adds the plugin gallery to `/index/admin/settings#plugins`. 

As I was testing I noticed that the plugin gallery in the settings wizard (e.g. `/index/admin/wizard/1#plugins`) redirects the user to the website settings after installing a plugin. I left this behaviour as it is but can file an issue for this unless it's intentional.

asmecher:
Thanks, I added a review! https://github.com/pkp/pkp-lib/pull/10240#pullrequestreview-2194890597

asmecher:
Thanks! This took a lot less code than I was expecting.

PR #10240, opened 2024-07-23
pkp/pkp-lib#9812 enable plugin gallery in admin site settings

kaitlinnewson:


controllers/grid/plugins/PluginGalleryGridHandler.inc.php
```diff
@@ -256,7 +256,15 @@ function upgradePlugin($args, $request) {
 	 * @param $isUpgrade boolean
 	 */
 	function installPlugin($args, $request, $isUpgrade = false) {
-		$redirectUrl = $request->getDispatcher()->url($request, ROUTE_PAGE, null, 'management', 'settings', array('website'), array('r' => uniqid()), 'plugins');
+		if ($request->getContext() == '') {
```
asmecher, line 259: `$request->getContext()` will return a `null` if you're in the administration area -- the empty string cast/compare is a little confusing. And I think it's clearer to read if the URLs are built for each area independently, rather than using `$handler` / `$path`. Add in `[]` for array syntax, and I'd recommend...

```php
if ($request->getContext()) {
    $redirectUrl = $request->getDispatcher()->url($request, ROUTE_PAGE, null, 'management, 'settings', ['website'], array('r' => uniqid()), 'plugins');
} else {
    $redirectUrl = $request->getDispatcher()->url($request, ROUTE_PAGE, null, 'admin', 'settings', null, array('r' => uniqid()), 'plugins');
}
```

asmecher (changes requested):
Just one small tweak to suggest -- this is nice and compact. Thanks, @kaitlinnewson!

kaitlinnewson:
Thanks @asmecher! I've addressed the comment and will work on forward-porting.

Issue #10158, opened 2024-07-03
Move publication to category relationship out of publication_settings


asmecher:
**Describe the bug**
Raised in https://github.com/pkp/pkp-lib/discussions/7975 and elsewhere:

Storing the category IDs for the publication in the `publication_settings` table is problematic for several reasons:
- It doesn't support relational integrity
- It performs poorly due to `setting_value` essentially being a blob column

For these reasons, the category IDs should instead be stored in a separate table. I suggest using an Eloquent model as discussed in https://github.com/pkp/pkp-lib/discussions/8103 and as implemented in e.g. the `UserUserGroup` model.

Hafsa-Naeem:
@asmecher  I've submitted a Draft PR which is currently under progress.
[pkp/pkp-lib#10158 Move categoryids out of of publication_settings](https://github.com/pkp/pkp-lib/pull/10247)

Hafsa-Naeem:
@asmecher Ready for review
OJS_3_4_0 [pkp/pkp-lib#10158 Move categoryids out of of publication_settings](https://github.com/pkp/pkp-lib/pull/10247)

asmecher:
Reviewed, thanks, @Hafsa-Naeem! https://github.com/pkp/pkp-lib/pull/10247#pullrequestreview-2208482792

Hafsa-Naeem:
@asmecher Ready for review. Based on the code review comments, I've added thechanges in the `main` branch.
**PR:**
OJS_main: [pkp/pkp-lib#10158 Move categoryids out of of publication_settings](https://github.com/pkp/pkp-lib/pull/10265)

Hafsa-Naeem:
Ready for review again @asmecher and @jonasraoni I've also added comments that could use your insights.
OJS_main: [pkp/pkp-lib#10158 Move categoryids out of of publication_settings #10265](https://github.com/pkp/pkp-lib/pull/10265)

Hafsa-Naeem:
@asmecher Ready for review again
OJS_main: [pkp/pkp-lib#10158 Move categoryids out of of publication_settings #10265](https://github.com/pkp/pkp-lib/pull/10265)

Hafsa-Naeem:
@asmecher @jonasraoni I've addressed this missing [comment ](https://github.com/pkp/pkp-lib/pull/10265#discussion_r1706790758)

jonasraoni:
Closing 🙃

PR #10247, opened 2024-07-24
pkp/pkp-lib#10158 Move  categoryids out of of publication_settings

Hafsa-Naeem:
for #10158

### DIFF
--- a/classes/publication/DAO.php
+++ b/classes/publication/DAO.php
@@ -489,13 +489,8 @@ class DAO extends EntityDAO
      */
     protected function setCategories(Publication $publication)
     {
-        $publication->setData(
-            'categoryIds',
-            Repo::category()->getCollector()
-                ->filterByPublicationIds([$publication->getId()])
-                ->getIds()
-                ->toArray()
-        );
+        $categoryIds = PublicationCategory::getCategoriesByPublicationId($publication->getId());
+        $publication->setData('categoryIds', $categoryIds);
     }
 
     /**
@@ -503,12 +498,8 @@ class DAO extends EntityDAO
      */
     protected function saveCategories(Publication $publication)
     {
-        Repo::category()->dao->deletePublicationAssignments($publication->getId());
-        if (!empty($publication->getData('categoryIds'))) {
-            foreach ($publication->getData('categoryIds') as $categoryId) {
-                Repo::category()->dao->insertPublicationAssignment($categoryId, $publication->getId());
-            }
-        }
+        $categoryIds = $publication->getData('categoryIds');
+        PublicationCategory::assignCategoriesToPublication($publication->getId(), $categoryIds);
     }
 
     /**
@@ -516,7 +507,7 @@ class DAO extends EntityDAO
      */
     protected function deleteCategories(int $publicationId)
     {
-        Repo::category()->dao->deletePublicationAssignments($publicationId);
+        PublicationCategory::where('publication_id', $publicationId)->delete();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


classes/publication/DAO.php
```diff
@@ -489,25 +490,27 @@ protected function deleteControlledVocab(int $publicationId)
...
     protected function setCategories(Publication $publication)
     {
-        $publication->setData(
-            'categoryIds',
-            Repo::category()->getCollector()
-                ->filterByPublicationIds([$publication->getId()])
-                ->getIds()
-                ->toArray()
-        );
+        try {
```
asmecher, line 493: I'd rather leave the exception to bubble upwards if necessary, rather than wrapping it in a `try` / `catch` / `error_log`. (Also below.) As it stands, this prevents calling code from being able to catch errors.

classes/publication/PublicationCategory.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace PKP\publication;
+
+use Illuminate\Database\Eloquent\Model;
+
+class PublicationCategory extends Model
+{
```
asmecher, line 8: Use the [`HasCamelCasing`](https://github.com/pkp/pkp-lib/blob/main/classes/stageAssignment/StageAssignment.php#L34) trait. ([Documentation](https://github.com/kirkbushell/eloquence?tab=readme-ov-file#camel-case-all-the-things))

classes/publication/PublicationCategory.php
```diff
@@ -0,0 +1,46 @@
...
+
+    protected $fillable = [
+        'publication_id', 'category_id'
+    ];
+
+    /**
+     * Get the categories associated with a publication.
+     *
+     */
+    public static function getCategoriesByPublicationId(int $publicationId)
```
asmecher, line 22: Use local [scope functions](https://laravel.com/docs/10.x/eloquent#local-scopes), allowing chaining of constraints. [Example](https://github.com/pkp/pkp-lib/blob/main/classes/stageAssignment/StageAssignment.php#L100)

classes/publication/PublicationCategory.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return self::where('publication_id', $publicationId)->pluck('category_id');
+    }
+
+    /**
+     * Assign categories to a publication.
+     *
+     */
+    public static function assignCategoriesToPublication(int $publicationId, array $categoryIds)
```
asmecher, line 31: This could go into a `Repository` class. [Example](https://github.com/pkp/pkp-lib/blob/main/classes/stageAssignment/Repository.php)

asmecher: Also, don't forget return typehints, even if it's just `void` for functions that don't return values.

classes/submission/Collector.php
```diff
@@ -510,8 +512,10 @@ public function getQueryBuilder(): Builder
         }
 
         if (isset($this->categoryIds)) {
-            $q->join('publication_categories as pc', 's.current_publication_id', '=', 'pc.publication_id')
-                ->whereIn('pc.category_id', $this->categoryIds);
+            $publicationIds = PublicationCategory::whereIn('category_id', $this->categoryIds)
```
asmecher, line 515: For code external to the `PublicationCategory` class, better to introduce local scope functions and call them from outside.

asmecher (changes requested):
Thanks, @Hafsa-Naeem! I made some suggestions.

classes/publication/PublicationCategory.php
Hafsa-Naeem: @asmecher I noticed that the HasCamelCasing trait isn’t present in stable-3_4_0 (only in main). Could you please clarify it?

Issue #9744, opened 2024-02-19
Migrate older Tables to newer Table API


jardakotesovec:
**Describe the issue**
With new [submissions listing](https://github.com/pkp/pkp-lib/issues/7495) it was necessary to introduce new Table component with more flexible API.  New table is now covered in storybook.  But there is couple of use cases using previous Table component.

**Solution**
Refactor older tables to use new API, should be relatively straightforward.

Pull requests:
pkp-lib/main: https://github.com/pkp/pkp-lib/pull/10255 (_merged_)
ui-library/main: https://github.com/pkp/ui-library/pull/389 (_merged_)
ojs/main: https://github.com/pkp/ojs/pull/4378 (_merged_)
omp/main: https://github.com/pkp/omp/pull/1656 (_merged_)
ops/main: https://github.com/pkp/ops/pull/737 (_merged_)

Cleanup PRs:
pkp-lib/main: https://github.com/pkp/pkp-lib/pull/10369 (_merged_)
ui-library/main: https://github.com/pkp/ui-library/pull/407 (_merged_)
ojs/main: https://github.com/pkp/ojs/pull/4427 _(submodule updates only, closed)_
omp/main: https://github.com/pkp/omp/pull/1691 _(submodule updates only, closed)_
ops/main: https://github.com/pkp/ops/pull/764 _(submodule updates only, closed)_

blesildaramirez:
Pull requests:
pkp-lib/main: https://github.com/pkp/pkp-lib/pull/10255
ui-library/main: https://github.com/pkp/ui-library/pull/389
ojs/main: https://github.com/pkp/ojs/pull/4378

Submodule updates only:
omp/main: https://github.com/pkp/omp/pull/1656
ops/main: https://github.com/pkp/ops/pull/737

blesildaramirez:
Raised a separate issue to address one more table to be migrated to use new table component: https://github.com/pkp/pkp-lib/issues/10443 [URNPubIdPlugin]

PR #10255, opened 2024-07-26
pkp/pkp-lib#9744 Migrate older Tables to newer Table API

blesildaramirez:


templates/admin/failedJobDetails.tpl
```diff
@@ -14,26 +14,28 @@
...
-						<span>
-							<pre>{{ row.value }}</pre>
-						</span>
-					</template>
-				</table-cell>
-			</template>
+		<div class="bg-default p-4">
+			<caption v-if="label">
+				<div class="pkpTable__label">
+					<slot name="label">{{ label }}</slot>
```
jardakotesovec, line 20: One thing to be super careful about is using classes that are defined in some particular component. These classes should be just used inside that component, not anywhere else. Components should be self-contained as possible. So if we for example refactor some component - it should not impact some UI outside of this component, which is very difficult to notice.

In this case its even class coming from the older Table component, which we intend to remove. So this is good example where this would go wrong.

templates/admin/failedJobDetails.tpl
```diff
@@ -14,26 +14,28 @@
...
-					<template #default v-if="column.name === 'value' && isValidJson(row.value)">
-						<span>
-							<pre>{{ row.value }}</pre>
-						</span>
-					</template>
-				</table-cell>
-			</template>
+		<div class="bg-default p-4">
+			<caption v-if="label">
+				<div class="pkpTable__label">
```
jardakotesovec, line 19: I believe that caption tag is intended to use inside table tag to be correctly associated with that table?

Our new table don't really have capability to define heading/description yet. Reason is really that it was initially created for new submission listing which does not really title/description embeded with table.

Screenshot 2024-08-07 at 13 21 24

But actually most other tables have quite simple pattern of having heading and description. So I think it make sense to extend the new table component to support that pretty much the same as we had in previous Table.

As result it would be possible to either provide label/description, or also possible to provide the aria variants if we just need to connect with some custom heading etc - aria-labelledBy and aria-describedBy like we will want to do in submission listing for example.

For label/description we can follow some design examples from submission listing:


Screenshot 2024-08-07 at 13 25 50

templates/admin/failedJobDetails.tpl
```diff
@@ -14,26 +14,28 @@
...
+		<pkp-table :aria-label="label">
+			<table-header>
+				<table-column v-for="column in columns" :key="column.name" :id="column.name">
+					{{ column.label }}
+				</table-column>
+			</table-header>
+			<table-body>
+				<table-row v-for="(row) in rows" :key="row.key">
+					<table-cell v-html="row.attribute"></span>
+					</table-cell>
```
jardakotesovec, line 33: Seeing some extra span here :-)

templates/admin/failedJobs.tpl
```diff
@@ -14,34 +14,49 @@
 		{translate key=$pageTitle}
 	</h1>
```
jardakotesovec, line 15: For these simple pages I would consider to actually migrating from smarty to vue. Obviously can be done separately if you prefer. But we will do it at some point anyway

But there are some good benefits, because once you move thing over to Vue file from the smarty (tpl) file it will help you with linting and syntax to make sure that the template is solid.

If you think its good idea to do it as part of this job for simple pages I can provide couple more details what to watch out.

templates/stats/context.tpl
```diff
@@ -110,31 +110,26 @@
...
-							<table-cell
-								v-for="(column, columnIndex) in tableColumns"
-								:key="column.name"
-								:column="column"
-								:row="row"
-								:tabindex="!rowIndex && !columnIndex ? 0 : -1"
-							>
-								<template #default v-if="column.name === 'title'">
+					<pkp-table aria-label="{translate key="stats.views"}">
+						<table-header>
```
jardakotesovec, line 114: So in this case I think we would use the aria-labelled-by..  which again is not in new table yet, but if you can add it.

templates/admin/failedJobDetails.tpl
blesildaramirez: Update here, I just added a `class="!max-w-[8rem] truncate"` directly on my use of TableColumn. It's not common to have long column names in the tables I migrated so I think this would be enough.

js/load.js
```diff
@@ -159,6 +163,10 @@ VueRegistry.registerComponent('PkpProgressBar', ProgressBar);
 VueRegistry.registerComponent('PkpSearch', Search);
 VueRegistry.registerComponent('PkpTable', Table);
 VueRegistry.registerComponent('PkpTableCell', TableCell);
+VueRegistry.registerComponent('TableColumn', TableColumn);
+VueRegistry.registerComponent('TableHeader', TableHeader);
```
jardakotesovec, line 167: For globally exposed components we want all of them to be prefixed with pkp.

templates/stats/publications.tpl
```diff
@@ -184,30 +184,33 @@
 					<pkp-table
 						labelled-by="publicationDetailTableLabel"
 						:class="tableClasses"
```
jardakotesovec, line 186: Oki.. I guess the :attr.classes it was because of this one?

This seems to just change opacity on loading..  I don't think its necessary. It already shows loading clearly enough.

I would suggest just to remove that.

templates/stats/publications.tpl
```diff
@@ -184,30 +184,33 @@
...
-						:columns="tableColumns"
-						:rows="items"
-						:order-by="orderBy"
-						:order-direction="orderDirection"
-						@order-by="setOrderBy"
+						@sort="setOrderBy"
 					>
```
jardakotesovec, line 188: Have you tested that the ordering works? I don't have any stats generated so can't test it easily..  and thats something that might work quite differently to before.. so best to check that.

jardakotesovec (changes requested):
Just couple small things.. its very close to being merged :-)

templates/stats/publications.tpl
blesildaramirez: Hi Jarda, yes the sorting still works. You might get some data in your local if you update the date range.

jardakotesovec (approved):

PR #10258, opened 2024-07-29
Allow bulk deletion of Incomplete Submissions

taslangraham:
For #6528

### DIFF
--- a/api/v1/_submissions/PKPBackendSubmissionsController.php
+++ b/api/v1/_submissions/PKPBackendSubmissionsController.php
@@ -20,6 +20,7 @@ namespace PKP\API\v1\_submissions;
 use APP\core\Application;
 use APP\facades\Repo;
 use APP\submission\Collector;
+use APP\submission\Submission;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
@@ -85,6 +86,14 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
                             Role::ROLE_ID_ASSISTANT,
                         ]
                 ),
+            ]);
+
+        Route::delete('', $this->bulkDeleteIncompleteSubmissions(...))
+            ->name('_submission.incomplete.delete')
+            ->middleware([
+                self::roleAuthorizer([
+                    Role::ROLE_ID_SITE_ADMIN,
+                ]),
             ]);
 
         Route::delete('{submissionId}', $this->delete(...))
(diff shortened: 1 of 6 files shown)


api/v1/_submissions/PKPBackendSubmissionsController.php
```diff
@@ -87,6 +87,15 @@ public function getGroupRoutes(): void
                 ),
             ]);
 
+        Route::delete('', $this->bulkDeleteIncompleteSubmissions(...))
+            ->name('_submission.incomplete.delete')
+            ->middleware([
+                self::roleAuthorizer([
+                    Role::ROLE_ID_SITE_ADMIN,
+                    Role::ROLE_ID_MANAGER,
```
taslangraham, line 95: Without the `Role:: ROLE_ID_MANAGER` entry for the middleware, requests with only a `Role::ROLE_ID_SITE_ADMIN` fails with a 401 error -  {"error":"You are not authorized to access the requested resource."}.


There is also a previous delete route, to delete a declined submission, that fails with the same error if I update the middleware to only use `Role::ROLE_ID_SITE_ADMIN` and attempt to delete a submission as an admin

https://github.com/pkp/pkp-lib/blob/3704af0dc5ce4454c3f0dbddde2a9734cb1351c5/api/v1/_submissions/PKPBackendSubmissionsController.php#L90-L99

asmecher: @taslangraham and @touhidurabir, I think there was probably an issue introduced with https://github.com/pkp/pkp-lib/issues/7698.

Before #7698 was introduced, everywhere that the `ROLE_ID_MANAGER` role was checked, `ROLE_ID_SITE_ADMIN` was checked too -- either was sufficient to grant access. (This was introduced with [#7392](https://github.com/pkp/pkp-lib/issues/7392#issuecomment-1119184605.)) But once [`HasRoles`](https://github.com/pkp/pkp-lib/blob/main/classes/middleware/HasRoles.php) replaced calls to [`PKPBaseController::addRoleAssignment`](https://github.com/pkp/pkp-lib/issues/7698), `ROLE_ID_SITE_ADMIN` is never getting added to the role list because [`$user->getRoles`](https://github.com/pkp/pkp-lib/blob/22af741d448e394dcbaeb8cc652530f5705666e1/classes/middleware/HasRoles.php#L50) will never return it when there is a context. (Enrolments with `ROLE_ID_SITE_ADMIN` will always have a context ID of `null`.)

@touhidurabir, could you look into adjusting `HasRoles` to include site admins?

jonasraoni (commented):
@taslangraham I just saw it's a work in progress, but sharing some nitpicks.

I've used the GitHub suggestions, so it's easier to pick the changes and do a single commit 😁
image

api/v1/_submissions/PKPBackendSubmissionsController.php
taslangraham: @touhidurabir @asmecher I've opened an issue and PR for this.
PR - https://github.com/pkp/pkp-lib/pull/10267

taslangraham:
@jonasraoni Thanks for the early feedback!

api/v1/_submissions/PKPBackendSubmissionsController.php
taslangraham: The fix from the PR above has been merged

classes/submission/Collector.php
```diff
@@ -53,6 +53,7 @@ abstract class Collector implements CollectorInterface, ViewsCount
     public DAO $dao;
     public ?array $categoryIds = null;
     public ?array $contextIds = null;
+    public ?array $submissionIds = null;
```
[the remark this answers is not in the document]
taslangraham: maybe we could explore this separately to avoid having this PR touch too many unrelated lines of code?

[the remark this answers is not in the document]
jonasraoni: Yep, it can be left for another time :)

classes/submission/Collector.php
```diff
@@ -250,6 +251,17 @@ public function filterByRevisionsSubmitted(?bool $revisionsSubmitted): AppCollec
         return $this;
     }
 
+    /**
+     * Limit results to only submissions with the specified IDs
+     *
+     * @param int[] $submissionIds Submission IDs
+     */
+    public function filterBySubmissionIds(array $submissionIds): static
```
jonasraoni, line 259: I think `null` should be allowed to disable the filter.

```suggestion
     * @param ?int[] $submissionIds Submission IDs
     */
    public function filterBySubmissionIds(?array $submissionIds): static
```

classes/security/authorization/SubmissionAccessPolicy.php
```diff
@@ -34,7 +34,7 @@ class SubmissionAccessPolicy extends ContextPolicy
      * @param array $roleAssignments
      * @param string $submissionParameterName the request parameter we
      *  expect the submission id in.
-     * @param bool $permitDeclined True iff declined reviews are permitted for viewing by reviewers
+     * @param bool $permitDeclined True if declined reviews are permitted for viewing by reviewers
```
jonasraoni, line 37: I thought it was a typo as well, but it means "if and only if" 😁

jonasraoni (approved):

classes/submission/Collector.php
taslangraham: updated

api/v1/_submissions/PKPBackendSubmissionsController.php
```diff
@@ -430,6 +439,70 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+                    'error' => __('api.submissions.403.unauthorizedDeleteSubmission'),
+                ], Response::HTTP_FORBIDDEN);
+            }
+        }
+
+        foreach ($submissions as $submission) {
+            Repo::submission()->delete($submission);
+        }
+
+        return response()->json([], Response::HTTP_OK);
```
jonasraoni, line 503: Perhaps the frontend won't be very friendly with that, but the `response()->noContent()` would fit better :)

beatrizmilz:
Please, this feature would be great!! 
https://github.com/pkp/pkp-lib/issues/5709

locale/en/common.po
```diff
@@ -370,6 +370,9 @@ msgstr "Delete Selection"
 msgid "common.deselect"
 msgstr "Deselect"
 
+msgid "common.deselectAll"
+msgstr "Deselect All"
```
taslangraham, line 374: @jardakotesovec I'm not sure if your UI implementation uses this and the entry below. If not I'll remove them

ewhanson (approved):

Issue #10266, opened 2024-07-31
Update HasRoles middleware to apply site admin role when applicable


taslangraham:
## Issue Description
See the following comments for context:
https://github.com/pkp/pkp-lib/pull/10258#discussion_r1695897652
https://github.com/pkp/pkp-lib/pull/10258#discussion_r1697679498

PR - https://github.com/pkp/pkp-lib/pull/10267

taslangraham:
@asmecher @touhidurabir I've created a PR for this. Please take a look.
PR - https://github.com/pkp/pkp-lib/pull/10267

asmecher:
Thanks, @taslangraham! @touhidurabir, could you take a first review?

PR #10267, opened 2024-07-31
pkp#10266 update HasRoles middleware to apply admin role for site wide admins

taslangraham:
For #10266

### DIFF
--- a/classes/middleware/HasRoles.php
+++ b/classes/middleware/HasRoles.php
@@ -18,6 +18,7 @@ namespace PKP\middleware;
 
 use APP\core\Application;
 use Closure;
+use Exception;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Str;
@@ -48,32 +49,24 @@ class HasRoles
     {
         $user = $request->user(); /** @var \PKP\user\User $user */
 
+        if (!$user) {
+            throw new Exception("No user found in `HasRole` middleware. Ensure that `HasUser` middleware is applied before `HasRole`.");
+        }
+
         // Get all user roles.
-        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var \PKP\security\RoleDAO $roleDao */
-        $userRoles = collect($roleDao->getByUserIdGroupedByContext($user->getId()));
-        $context = $request->attributes->get('context'); /** @var \PKP\context\Context $context */
+        $context = $request->attributes->get('context'); /** @var ?\PKP\context\Context $context */
 
         $matchableRoles = Str::of($matchableRoles)
             ->explode('|')
-            ->map(fn($role) => (int)$role)
-            ->sort();
+            ->map(fn($role) => (int)$role);
 
-        $matchedRoles = collect([]);
+        $matcher = fn(int $roleId) => $user->hasRole($roleId, $roleId === Role::ROLE_ID_SITE_ADMIN ? Application::SITE_CONTEXT_ID : $context->getId());
+        $isAuthorized = match ($rolesMatchingCriteria) {
+            static::ROLES_MATCH_LOOSE => $matchableRoles->some($matcher),
+            static::ROLES_MATCH_STRICT => $matchableRoles->every($matcher)
+        };
 
-        // Check for a match amongst Context specific roles
-        if ($context && $userRoles->has($context->getId())) {
-            $userContextRoles = collect(array_map(fn(Role $role) => $role->getId(), $userRoles->get($context->getId())));
-            $matchedRoles = $userContextRoles->intersect($matchableRoles);
-        }
-
-        // Apply site wide Admin role if user is Admin and matchableRoles contains value of Role::ROLE_ID_SITE_ADMIN.
-        if (array_key_exists(Role::ROLE_ID_SITE_ADMIN, $userRoles->get((int)Application::SITE_CONTEXT_ID) ?? []) && $matchableRoles->contains(Role::ROLE_ID_SITE_ADMIN)) {
-            $matchedRoles->add(Role::ROLE_ID_SITE_ADMIN);
-        }
-
-        // if no roles matched
-        // Or if role matching set to strict and not all roles matched
-        if ($matchedRoles->isEmpty() || ($rolesMatchingCriteria === self::ROLES_MATCH_STRICT && !$matchableRoles->diff($matchedRoles)->isEmpty())) {
+        if (!$isAuthorized) {
             return response()->json([
                 'error' => __('api.403.unauthorized'),
             ], Response::HTTP_UNAUTHORIZED);
--- a/classes/middleware/HasRoles.php
+++ b/classes/middleware/HasRoles.php
@@ -21,6 +21,8 @@ use Closure;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Str;
+use PKP\security\Role;
+use PKP\db\DAORegistry;
 
 class HasRoles
 {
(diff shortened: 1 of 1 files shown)


classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,31 @@ class HasRoles
...
         $matchableRoles = Str::of($matchableRoles)
             ->explode('|')
             ->map(fn ($role) => (int)$role)
             ->sort();
 
-        $matchedRoles = $userRoles->intersect($matchableRoles);
+        $matchedRoles = collect([]);
+
+        // Check for a match amongst Context specific roles
+        if ($context->getId() && $userRoles->has($context->getId())) {
```
touhidurabir, line 64: calling `getId()` on non-contextual request will cause exception like `Error: Call to a member function getId() on null`  as context not available for such request . For example, try visiting the `View Jobs` or `View Failed Jobs` from `Administration` section .

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,31 @@ class HasRoles
...
+            $userContextRoles = collect(array_map(fn($role) => $role->getId(), $userRoles->get($context->getId())));
+            $matchedRoles = $userContextRoles->intersect($matchableRoles);
+        }
+
+        // Apply site wide Admin role if user is Admin and matchableRoles contains value of Role::ROLE_ID_SITE_ADMIN.
+        if ($userRoles->has((int)Application::SITE_CONTEXT_ID) &&
+            array_key_exists(Role::ROLE_ID_SITE_ADMIN, $userRoles->get((int)Application::SITE_CONTEXT_ID)) &&
```
touhidurabir, line 71: perhaps we can write these 2 lines as a single check like 
```php
array_key_exists(Role::ROLE_ID_SITE_ADMIN, $userRoles->get((int)Application::SITE_CONTEXT_ID) ?? [])
```

touhidurabir (changes requested):
@taslangraham sorry for delay . I have added few comments, please check those .

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,31 @@ class HasRoles
...
             ->explode('|')
             ->map(fn ($role) => (int)$role)
             ->sort();
 
-        $matchedRoles = $userRoles->intersect($matchableRoles);
+        $matchedRoles = collect([]);
+
+        // Check for a match amongst Context specific roles
+        if ($context->getId() && $userRoles->has($context->getId())) {
+            $userContextRoles = collect(array_map(fn($role) => $role->getId(), $userRoles->get($context->getId())));
```
jonasraoni, line 65:
```suggestion
            $userContextRoles = collect(array_map(fn (Role $role) => $role->getId(), $userRoles->get($context->getId())));
```

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,31 @@ class HasRoles
...
     {
         $user = $request->user(); /** @var \PKP\user\User $user */
-        $context = $request->attributes->get('context'); /** @var \PKP\context\Context $context */
 
-        $userRoles = collect($user->getRoles($context?->getId() ?? Application::SITE_CONTEXT_ID))
-            ->map(fn ($role) => $role->getId())
-            ->sort();
+        // Get all user roles.
+        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var RoleDAO $roleDao */
+        $userRoles = collect( $roleDao->getByUserIdGroupedByContext($user->getId()));
```
jonasraoni, line 53:
```suggestion
        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var PKP\security\RoleDAO $roleDao */
        $userRoles = collect($roleDao->getByUserIdGroupedByContext($user->getId()));
```

taslangraham:
Thanks for the review @touhidurabir and @jonasraoni. I've applied the suggested changes.

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,29 @@ class HasRoles
     public function handle(Request $request, Closure $next, string $matchableRoles, int $rolesMatchingCriteria = HasRoles::ROLES_MATCH_LOOSE)
     {
         $user = $request->user(); /** @var \PKP\user\User $user */
```
jonasraoni, line 49: I didn't check how this middleware is handled, but the request might not have a user, so it sounds reasonable to include an extra check.

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,29 @@ class HasRoles
...
-        $context = $request->attributes->get('context'); /** @var \PKP\context\Context $context */
 
-        $userRoles = collect($user->getRoles($context?->getId() ?? Application::SITE_CONTEXT_ID))
-            ->map(fn ($role) => $role->getId())
-            ->sort();
+        // Get all user roles.
+        $roleDao = DAORegistry::getDAO('RoleDAO'); /** @var \PKP\security\RoleDAO $roleDao */
+        $userRoles = collect($roleDao->getByUserIdGroupedByContext($user->getId()));
+        $context = $request->attributes->get('context'); /** @var \PKP\context\Context $context */
```
jonasraoni, line 54: Nitpick

```suggestion
        $context = $request->attributes->get('context'); /** @var ?\PKP\context\Context $context */
```

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,29 @@ class HasRoles
...
 
         $matchableRoles = Str::of($matchableRoles)
             ->explode('|')
-            ->map(fn ($role) => (int)$role)
+            ->map(fn($role) => (int)$role)
             ->sort();
```
jonasraoni, line 59: I think it's not needed to sort 🤔

classes/middleware/HasRoles.php
```diff
@@ -45,18 +47,29 @@ class HasRoles
...
+        // Check for a match amongst Context specific roles
+        if ($context && $userRoles->has($context->getId())) {
+            $userContextRoles = collect(array_map(fn(Role $role) => $role->getId(), $userRoles->get($context->getId())));
+            $matchedRoles = $userContextRoles->intersect($matchableRoles);
+        }
+
+        // Apply site wide Admin role if user is Admin and matchableRoles contains value of Role::ROLE_ID_SITE_ADMIN.
+        if (array_key_exists(Role::ROLE_ID_SITE_ADMIN, $userRoles->get((int)Application::SITE_CONTEXT_ID) ?? []) && $matchableRoles->contains(Role::ROLE_ID_SITE_ADMIN)) {
+            $matchedRoles->add(Role::ROLE_ID_SITE_ADMIN);
+        }
```
jonasraoni, line 72: I think this code (including the next `if`, which I can't include on this comment) can be simplified to:

```suggestion
        $matcher = fn (int $roleId) => $user->hasRole($roleId, $roleId === Role::ROLE_ID_SITE_ADMIN ? Application::SITE_CONTEXT_ID : $context->getId());
        $isAuthorized = match ($rolesMatchingCriteria) {
            static::ROLES_MATCH_LOOSE => $matchableRoles->some($matcher),
            static::ROLES_MATCH_STRICT => $matchableRoles->every($matcher)
        };
```

jonasraoni (commented):
@taslangraham I've left more comments and a possible improvement.

touhidurabir (approved):

touhidurabir:
@taslangraham I have approved it . However there is still some suggestions left by @jonasraoni unresolved and maybe you can resolve those before merging it .

classes/middleware/HasRoles.php
touhidurabir: Probably not a bad idea, however the `HasRole` middleware should always followed by a `HasUser` middleware . But as the HasUser middleware may use as stand alone and there is no such easy way to define that `HasRole` must follow the `HasUser`, it does leave a very small chance misuse that will cause exception . 

My suggestion is that rather than a check, probably an exception is better if there is no user . Basically without a user `HasRole` does not seems to have any meaning .

taslangraham: Should I add such exception to this PR?

jonasraoni: If there's a chance of misuse, then go ahead with a good exception message regarding what Touhidur wrote 😁

taslangraham: added exception https://github.com/pkp/pkp-lib/pull/10267/commits/11013232e58314d1957d4f2d2991feaf0ac59d35

classes/middleware/HasRoles.php
jonasraoni: @taslangraham I think this is the last comment, check if makes sense, then I think we can merge this 😁

taslangraham: This is a good suggestion - I've made the update

jonasraoni (approved):
@taslangraham I don't know if you have write permissions here, so I'll get it merged 😁

Issue #10188, opened 2024-07-10
Convert Notes to an Eloquent model


asmecher:
The Note class (https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/note/Note.php) and NoteDAO (https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/note/NoteDAO.php) are implemented using the old-style DAO pattern. Migrate to an Eloquent model.

kaitlinnewson:
@asmecher This is ready for a review: 

- pkp-lib main: https://github.com/pkp/pkp-lib/pull/10268

I did a PR to OJS main for tests and they are failing, but I'm not sure that it's related to my changes - still reviewing the logs:

- OJS main (tests only): https://github.com/pkp/ojs/pull/4385

asmecher:
Thanks, @kaitlinnewson! A few of us have added some suggestions/comments -- this is getting very close to ready.

Looking at the log failures, I picked out the error log artifact from the end of one of the runs. It contains a bunch of repetitions of this:
```
PHP Fatal error:  Class APP\core\Services contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (PKP\core\PKPServices::init) in /home/runner/ojs/classes/core/Services.php on line 20
[Thu Aug  1 17:27:51 2024] 127.0.0.1:60172 [500]: GET / - Class APP\core\Services contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (PKP\core\PKPServices::init) in /home/runner/ojs/classes/core/Services.php on line 20
```
The abstract `init` function was removed from `PKPServices` [just today](https://github.com/pkp/pkp-lib/commit/4c31a124038bb79d0dfb4fdd0755f17c891c3d28#diff-06f399b2adf99d15ebf0423883a17c9cb1996d0139e5837f0e9a479c8b47aff1L43) -- I think you can resolve this error by...
 - Rebasing the `10188-main` branch of your `pkp-lib` repository on the latest `main`,
 - Force-pushing your newly-rebased `10188-main` branch of `pkp-lib` up to your remote,
 - Rebasing the `10188-main` branch of your `ojs` repository on the latest `main`, making sure to update the `lib/pkp` submodule to the new latest, and finally
 - Force-pushing the newly-rebased `101188-main` branch of OJS up to your remote.

kaitlinnewson:
[...]

PR #10268, opened 2024-08-01
pkp/pkp-lib#10188 Convert Notes to an Eloquent model

kaitlinnewson:


classes/note/Repository.php
```diff
@@ -0,0 +1,46 @@
...
+                'dateModified' => Carbon::now(),
+            ]));
+    }
+}
+
+if (!PKP_STRICT_MODE) {
+    class_alias('\PKP\note\Note', '\Note');
+    define('NOTE_ORDER_DATE_CREATED', Note::NOTE_ORDER_DATE_CREATED);
+    define('NOTE_ORDER_ID', Note::NOTE_ORDER_ID);
+}
```
taslangraham, line 46: don't think this should be here

classes/note/Note.php
```diff
@@ -3,173 +3,145 @@
...
 
     /**
-     * set date note was modified
+     * Return the user of the note's author.
      *
-     * @param string $dateModified (YYYY-MM-DD HH:MM:SS)
+     * @return \PKP\user\User
      */
-    public function setDateModified($dateModified)
+    public function getUser()
```
taslangraham, line 68: we could make this a model accessor too. Like
```
    protected function user(): Attribute
    {
        return Attribute::make(
            get: function () {
                return Repo::user()->get($this->userId, true);
            },
        );
    }
```

taslangraham (changes requested):
@kaitlinnewson just few things. Otherwise this looks good

classes/note/Note.php
```diff
@@ -3,173 +3,145 @@
...
 
     /**
-     * get note contents
+     * Compatibility function for including note IDs in grids.
      *
-     * @return string
+     * @deprecated
      */
-    public function getContents()
+    public function getId()
```
taslangraham, line 78: is this needed given that you can call the `id` accessor directly?

asmecher: It's necessary for some of the legacy code, e.g. GridHandler, which expects to deal with `DataObject` classes rather than Eloquent `Model`s. I had to add one for `Notification` models [here](https://github.com/pkp/pkp-lib/blob/4c31a124038bb79d0dfb4fdd0755f17c891c3d28/classes/notification/Notification.php#L135) too.

classes/note/Repository.php
asmecher: In fact, since we're breaking backwards compatibility, now is a good time to remove the `define` / `class_alias` calls entirely. The remaining ones will go away with 3.6, but we're removing them opportunistically when we introduce a breaking change, like shifting over to Eloquent.

classes/note/Repository.php
```diff
@@ -0,0 +1,46 @@
...
+namespace PKP\note;
+
+use Carbon\Carbon;
+
+class Repository
+{
+    /**
+     * Fetch a note by symbolic info, building it if needed.
+     */
+    public function build(int $assocType, int $assocId, ?int $userId, ?string $contents, ?string $title): Note
```
asmecher, line 26: `user_id` is not nullable in the database, so it should not be optional here.

classes/query/Query.php
```diff
@@ -133,22 +133,23 @@ public function setIsClosed($isClosed)
...
      * @param int $sortOrder Optional DAO::SORT_DIRECTION_...
      *
      */
-    public function getReplies(?int $userId = null, int $sortBy = NoteDAO::NOTE_ORDER_ID, int $sortOrder = \PKP\db\DAO::SORT_DIRECTION_ASC): LazyCollection
+    public function getReplies(?int $userId = null, int $sortBy = Note::NOTE_ORDER_ID, int $sortOrder = DAO::SORT_DIRECTION_ASC): LazyCollection
     {
-        $noteDao = DAORegistry::getDAO('NoteDAO'); /** @var NoteDAO $noteDao */
-        return $noteDao->getByAssoc(PKPApplication::ASSOC_TYPE_QUERY, $this->getId(), null, $sortBy, $sortOrder);
+        return Note::withAssoc(PKPApplication::ASSOC_TYPE_QUERY, $this->getId())
+                    ->when($userId, fn (Builder $query, string $userId) => $query->withUserId($userId))
```
asmecher, line 150: `string` should be `int` for `$userId`.

classes/user/Repository.php
```diff
@@ -330,12 +330,7 @@ public function mergeUsers(int $oldUserId, int $newUserId)
             Repo::submissionFile()->edit($submissionFile, ['uploaderUserId' => $newUserId]);
         }
 
-        $noteDao = DAORegistry::getDAO('NoteDAO'); /** @var NoteDAO $noteDao */
-        $notes = $noteDao->getByUserId($oldUserId);
-        while ($note = $notes->next()) {
-            $note->setUserId($newUserId);
-            $noteDao->updateObject($note);
-        }
+        Note::withUserId($oldUserId)->each()->update(['userId' => $newUserId]);
```
asmecher, line 333: This will materialize each `Note` object individually, which can be an expensive operation. I'd suggest adding a function to the `Repository` class to handle this case -- see `lib/pkp/classes/notification/Repository.php` for an example in the `transfer()` function. (It would be possible to remove the `each()` to do it here -- but that would require us to add `snake_case` column names outside of the model code, which we'd like to avoid.)

controllers/grid/queries/QueriesGridCellProvider.php
```diff
@@ -69,24 +71,24 @@ public function getTemplateVarsFromRowColumn($row, $column)
...
+        $notes = $element->getReplies(null, Note::NOTE_ORDER_ID, \PKP\db\DAO::SORT_DIRECTION_DESC);
         $context = Application::get()->getRequest()->getContext();
         $datetimeFormatShort = PKPString::convertStrftimeFormat($context->getLocalizedDateTimeFormatShort());
 
         switch ($columnId) {
             case 'replies':
                 return ['label' => max(0, $notes->count() - 1)];
             case 'from':
-                return ['label' => ($user ? $user->getUsername() : '&mdash;') . '<br />' . ($headNote ? date($datetimeFormatShort, strtotime($headNote->getDateCreated())) : '')];
+                return ['label' => ($user ? $user->getUsername() : '&mdash;') . '<br />' . ($headNote ? date($datetimeFormatShort, strtotime($headNote->dateCreated)) : '')];
```
asmecher, line 86: `$headNote->dateCreated` should be a `Carbon` object now, so `strtotime` is no longer appropriate here (and below).

controllers/grid/queries/QueryNotesGridHandler.php
```diff
@@ -275,8 +273,8 @@ public function deleteNote($args, $request)
             return new JSONMessage(false);
         }
 
-        $noteDao->deleteObject($note);
-        return \PKP\db\DAO::getDataChangedEvent($note->getId());
+        $note->delete();
+        return \PKP\db\DAO::getDataChangedEvent($note->id);
```
asmecher, line 277: I would guess it's probably safe, but I haven't tried working with Eloquent model attribute data after calling a `delete()` operation. It might be worth specifically testing this to see if `$note->id` is still set after the `delete()`.

classes/note/Note.php
taslangraham: Okay, thanks for clarifying

controllers/grid/queries/QueriesGridCellProvider.php
```diff
@@ -69,24 +71,24 @@ public function getTemplateVarsFromRowColumn($row, $column)
...
             case 'lastReply':
                 $latestReply = $notes->first();
-                if ($latestReply && $latestReply->getId() != $headNote->getId()) {
+                if ($latestReply && $latestReply->getId() != $headNote->id) {
                     $repliedUser = $latestReply->getUser();
-                    return ['label' => ($repliedUser ? $repliedUser->getUsername() : '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($latestReply->getDateCreated()))];
+                    return ['label' => ($repliedUser ? $repliedUser->getUsername() : '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($latestReply->dateCreated))];
```
jonasraoni, line 91:
```suggestion
                    return ['label' => ($repliedUser?->getUsername() ?? '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($latestReply->dateCreated))];
```

controllers/grid/queries/QueriesGridCellProvider.php
jonasraoni, line 86: Perhaps this label should be moved to a locale file, it's always curious to see HTML code mixed with PHP.

controllers/grid/queries/QueryNotesGridCellProvider.php
```diff
@@ -59,14 +61,14 @@ public function getTemplateVarsFromRowColumn($row, $column)
...
-        assert($element instanceof \PKP\core\DataObject && !empty($columnId));
+        assert(($element instanceof DataObject || $element instanceof Model) && !empty($columnId));
         /** @var Note $element */
         $user = $element->getUser();
         $datetimeFormatShort = PKPString::convertStrftimeFormat(Application::get()->getRequest()->getContext()->getLocalizedDateTimeFormatShort());
 
         switch ($columnId) {
             case 'from':
-                return ['label' => ($user ? $user->getUsername() : '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($element->getDateCreated()))];
+                return ['label' => ($user ? $user->getUsername() : '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($element->dateCreated))];
```
jonasraoni, line 71:
```suggestion
                return ['label' => ($user?->getUsername() ?? '&mdash;') . '<br />' . date($datetimeFormatShort, strtotime($element->dateCreated))];
```

controllers/grid/queries/QueryNotesGridHandler.php
```diff
@@ -186,10 +185,10 @@ public function getRequestArgs()
...
     {
         return $this->getQuery()
-            ->getReplies(null, NoteDAO::NOTE_ORDER_DATE_CREATED, \PKP\db\DAO::SORT_DIRECTION_ASC)
+            ->getReplies(null, Note::NOTE_ORDER_DATE_CREATED, \PKP\db\DAO::SORT_DIRECTION_ASC)
             ->filter(function (Note $note) use ($request) {
-                return (bool) $note->getContents() || (
-                    $note->getUserId() === $request->getUser()->getId()
+                return (bool) $note->contents || (
+                    $note->userId === $request->getUser()->getId()
                 );
```
jonasraoni, line 192:
```suggestion
                return $note->contents || $note->userId === $request->getUser()->getId();
```

controllers/grid/queries/QueryNotesGridHandler.php
```diff
@@ -246,7 +245,7 @@ public function getCanManage($note)
         if ($note === null) {
             return $isAdmin;
         } else {
-            return ($note->getUserId() == $this->_user->getId() || $isAdmin);
+            return ($note->userId == $this->_user->getId() || $isAdmin);
         }
```
jonasraoni, line 249: Hmm, the UI of GitHub is bugged here, but the if/else can be shortened into:

```suggestion
return $note?->userId == $this->_user->getId() || $isAdmin;
```

controllers/grid/queries/QueryNotesGridHandler.php
```diff
@@ -261,11 +260,10 @@ public function getCanManage($note)
     public function deleteNote($args, $request)
     {
         $query = $this->getQuery();
-        $noteDao = DAORegistry::getDAO('NoteDAO'); /** @var NoteDAO $noteDao */
-        $note = $noteDao->getById($request->getUserVar('noteId'));
+        $note = Note::find($request->getUserVar('noteId'));
```
jonasraoni, line 263:
```suggestion
        $note = Note::find((int) $request->getUserVar('noteId'));
```

controllers/grid/queries/QueryNotesGridHandler.php
```diff
@@ -261,11 +260,10 @@ public function getCanManage($note)
...
         $user = $request->getUser();
 
-        if (!$request->checkCSRF() || !$note || $note->getAssocType() != Application::ASSOC_TYPE_QUERY || $note->getAssocId() != $query->getId()) {
+        if (!$request->checkCSRF() || !$note || $note->assocType != Application::ASSOC_TYPE_QUERY || $note->assocId != $query->getId()) {
```
jonasraoni, line 266:
```suggestion
        if (!$request->checkCSRF() || $note?->assocType != Application::ASSOC_TYPE_QUERY || $note?->assocId != $query->getId()) {
```

controllers/grid/queries/QueryNotesGridRow.php
```diff
@@ -64,7 +64,7 @@ public function initialize($request, $template = null)
         // Is this a new row or an existing row?
         $rowId = $this->getId();
         $headNote = $this->getQuery()->getHeadNote();
-        if (!empty($rowId) && is_numeric($rowId) && (!$headNote || $headNote->getId() != $rowId)) {
+        if (!empty($rowId) && is_numeric($rowId) && (!$headNote || $headNote->id != $rowId)) {
```
jonasraoni, line 67:
```suggestion
        $rowId = abs($this->getId());
        $headNote = $this->getQuery()->getHeadNote();
        if ($rowId > 0 && $headNote?->id != $rowId)) {
```

controllers/grid/queries/QueryTitleGridColumn.php
```diff
@@ -93,10 +93,10 @@ public function getCellActions($request, $row, $position = GridHandler::GRID_ACT
                     'readQuery',
                     new AjaxModal(
                         $router->url($request, null, null, 'readQuery', null, $actionArgs),
-                        $headNote ? htmlspecialchars($headNote->getTitle()) : '&mdash;',
+                        $headNote ? htmlspecialchars($headNote->title) : '&mdash;',
                         'modal_edit'
                     ),
-                    ($headNote && $headNote->getTitle() != '') ? htmlspecialchars($headNote->getTitle()) : '&mdash;',
+                    ($headNote && $headNote->title != '') ? htmlspecialchars($headNote->title) : '&mdash;',
```
jonasraoni, line 99:
```suggestion
                    ($headNote?->title != '') ? htmlspecialchars($headNote->title) : '&mdash;',
```

controllers/grid/queries/form/QueryForm.php
```diff
@@ -392,7 +392,7 @@ public function fetch($request, $template = null, $display = false, $actionArgs
         $allowedEditTimeNotice = ['show' => false, 'limit' => 60];
         if (array_intersect($assignedRoles, [Role::ROLE_ID_ASSISTANT, Role::ROLE_ID_AUTHOR, Role::ROLE_ID_REVIEWER])) {
             $allowedEditTimeNotice['show'] = true;
-            $allowedEditTimeNotice['limit'] = (int) ($allowedEditTimeNotice['limit'] - (time() - strtotime($headNote->getDateCreated())) / 60);
+            $allowedEditTimeNotice['limit'] = (int) ($allowedEditTimeNotice['limit'] - (time() - strtotime($headNote->dateCreated)) / 60);
```
jonasraoni, line 395: If possible, better to use the `Carbon` package to make it more understandable :)

controllers/informationCenter/InformationCenterHandler.php
```diff
@@ -129,13 +129,13 @@ public function deleteNote($args, $request)
         $this->setupTemplate($request);
 
         $noteId = (int) $request->getUserVar('noteId');
-        $noteDao = DAORegistry::getDAO('NoteDAO'); /** @var NoteDAO $noteDao */
-        $note = $noteDao->getById($noteId);
+        $note = Note::find($noteId);
 
-        if (!$request->checkCSRF() || !$note || $note->getAssocType() != $this->_getAssocType() || $note->getAssocId() != $this->_getAssocId()) {
+        if (!$request->checkCSRF() || !$note || $note->assocType != $this->_getAssocType() || $note->assocId != $this->_getAssocId()) {
```
jonasraoni, line 134:
```suggestion
        if (!$request->checkCSRF() || $note?->assocType != $this->_getAssocType() || $note?->assocId != $this->_getAssocId()) {
```

classes/note/Note.php
```diff
@@ -3,173 +3,145 @@
 /**
  * @file classes/note/Note.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
```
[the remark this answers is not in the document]
kaitlinnewson: @asmecher Can you confirm the practice on this? I noticed in the `Notification.php` changes you removed the initial year.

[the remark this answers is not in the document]
asmecher: I don't think it matters at ton :) but I try to bump the end year to the current year when I modify something.
In `Notification.php`, I removed the start year and just left `2024` because it was a complete rewrite of the file to a new implementation. But I almost never do that.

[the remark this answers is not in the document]
kaitlinnewson: Thanks for clarifying - since this one is also a rewrite I think it makes sense to leave it as `2024`.

controllers/grid/queries/QueryNotesGridHandler.php
kaitlinnewson: Tested and confirmed that `$note->id` is still set.

classes/controllers/grid/DataObjectGridCellProvider.php
```diff
@@ -74,7 +76,7 @@ public function getTemplateVarsFromRowColumn($row, $column)
     {
         $element = $row->getData();
         $columnId = $column->getId();
-        assert($element instanceof \PKP\core\DataObject && !empty($columnId));
+        assert(($element instanceof DataObject || $element instanceof Model) && !empty($columnId));
```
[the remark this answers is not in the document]
kaitlinnewson: I see quite a few of these `assert`s in the grid classes - could we file this as a separate issue to review rather than including it in this issue?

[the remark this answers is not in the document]
jonasraoni: Yep, there must be a lot. I don't think it's important, I'm just suggesting the update because you've touched the lines.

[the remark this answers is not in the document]
kaitlinnewson: I've filed an issue for this here: https://github.com/pkp/pkp-lib/issues/10285

classes/note/Note.php
```diff
@@ -3,176 +3,148 @@
...
     }
 
     /**
-     * set note assoc id
-     *
-     * @param int $assocId
+     * Scope a query to a specific sort order.
      */
-    public function setAssocId($assocId)
+    public function scopeWithSort(Builder $query, int $orderBy = NOTE_ORDER_DATE_CREATED, int $sortDirection = DAO::SORT_DIRECTION_DESC): Builder
```
asmecher, line 135:
```suggestion
    public function scopeWithSort(Builder $query, int $orderBy = static::NOTE_ORDER_DATE_CREATED, int $sortDirection = DAO::SORT_DIRECTION_DESC): Builder
```

classes/note/Note.php
```diff
@@ -3,176 +3,148 @@
...
     {
-        $this->setData('assocId', $assocId);
+        // Sanitize sort ordering
+        $orderSanitized = match ($orderBy) {
+            self::NOTE_ORDER_ID => 'note_id',
+            default => 'date_created',
```
asmecher, line 140:
```suggestion
            self::NOTE_ORDER_DATE_CREATED => 'date_created',
```
This way unsupported incoming parameters will result in an exception rather than silently ignoring it.

classes/note/Note.php
```diff
@@ -3,176 +3,148 @@
...
+        };
+
+        $directionSanitized = match ($sortDirection) {
+            DAO::SORT_DIRECTION_ASC => 'ASC',
+            default => 'DESC',
```
asmecher, line 145:
```suggestion
            DAO::SORT_DIRECTION_DESC => 'DESC',
```
...as above.

classes/query/Query.php
```diff
@@ -133,22 +133,23 @@ public function setIsClosed($isClosed)
...
      * Get all notes on a query.
      *
      * @param int $userId Optional user ID
-     * @param int $sortBy Optional NoteDAO::NOTE_ORDER_...
+     * @param int $sortBy Optional Note::NOTE_ORDER_...
      * @param int $sortOrder Optional DAO::SORT_DIRECTION_...
      *
      */
-    public function getReplies(?int $userId = null, int $sortBy = NoteDAO::NOTE_ORDER_ID, int $sortOrder = \PKP\db\DAO::SORT_DIRECTION_ASC): LazyCollection
+    public function getReplies(?int $userId = null, int $sortBy = Note::NOTE_ORDER_ID, int $sortOrder = DAO::SORT_DIRECTION_ASC): LazyCollection
```
asmecher, line 147: Hmm, I'm not 100% opposed to the current approach, but I think now that we're using Laravel scope functions for accessing notes we can get rid of `getReplies` and rely on `Note::withAssoc(...)->...`. Some callers want `withUser` and others not; some will benefit from lazy loading and others will just want the first note and that's all; etc. This function kind of emulates the old DAO-based functions we used to have with a list of fairly random parameters, and I'd rather we moved away from that.

controllers/grid/queries/QueriesAccessHelper.php
```diff
@@ -128,7 +129,7 @@ public function getCanEdit($queryId)
         // Assistants, authors and reviewers are allowed, if they created the query less than x seconds ago
         if ($this->hasStageRole($query->getStageId(), [Role::ROLE_ID_ASSISTANT, Role::ROLE_ID_AUTHOR, Role::ROLE_ID_REVIEWER])) {
             $headNote = $query->getHeadNote();
-            if ($headNote->getUserId() == $this->_user->getId() && (time() - strtotime($headNote->getDateCreated()) < 3600)) {
+            if ($headNote->userId == $this->_user->getId() && (Carbon::now()->diffInSeconds($headNote->dateCreated)) < 3600) {
```
asmecher, line 132:
```suggestion
            if ($headNote->userId == $this->_user->getId() && ($headNote->dateCreated->diffInHours() < 1)) {
```

controllers/grid/queries/QueriesGridCellProvider.php
```diff
@@ -69,24 +72,24 @@ public function getTemplateVarsFromRowColumn($row, $column)
...
+        $notes = $element->getReplies(null, Note::NOTE_ORDER_ID, DAO::SORT_DIRECTION_DESC);
         $context = Application::get()->getRequest()->getContext();
         $datetimeFormatShort = PKPString::convertStrftimeFormat($context->getLocalizedDateTimeFormatShort());
 
         switch ($columnId) {
             case 'replies':
                 return ['label' => max(0, $notes->count() - 1)];
             case 'from':
-                return ['label' => ($user ? $user->getUsername() : '&mdash;') . '<br />' . ($headNote ? date($datetimeFormatShort, strtotime($headNote->getDateCreated())) : '')];
+                return ['label' => ($user?->getUsername() ?? '&mdash;') . '<br />' . ($headNote ? $headNote->dateCreated->format($datetimeFormatShort) : '')];
```
asmecher, line 87:
```suggestion
                return ['label' => ($user?->getUsername() ?? '&mdash;') . '<br />' . $headNote?->dateCreated->format($datetimeFormatShort)];
```

controllers/grid/queries/form/QueryForm.php
```diff
@@ -392,7 +390,7 @@ public function fetch($request, $template = null, $display = false, $actionArgs
         $allowedEditTimeNotice = ['show' => false, 'limit' => 60];
         if (array_intersect($assignedRoles, [Role::ROLE_ID_ASSISTANT, Role::ROLE_ID_AUTHOR, Role::ROLE_ID_REVIEWER])) {
             $allowedEditTimeNotice['show'] = true;
-            $allowedEditTimeNotice['limit'] = (int) ($allowedEditTimeNotice['limit'] - (time() - strtotime($headNote->getDateCreated())) / 60);
+            $allowedEditTimeNotice['limit'] = (int) ($allowedEditTimeNotice['limit'] - Carbon::now()->diffInMinutes($headNote->dateCreated));
```
asmecher, line 393:
```suggestion
            $allowedEditTimeNotice['limit'] = (int) ($allowedEditTimeNotice['limit'] - $headNote->dateCreated->diffInMinutes());
```

classes/query/Query.php
kaitlinnewson: I've gone ahead and removed this and replaced the calls as suggested

classes/note/Note.php
kaitlinnewson: Setting this to static results in `static::" is not allowed in compile-time constants` errors so I've left this one out.

asmecher: Sorry, you're right that `static::` won't work -- but shouldn't it be `self::`? It's now a class constant, and I think you removed the old backwards compatibility defines for these constants, so just referring to `NOTE_ORDER_DATE_CREATED` without qualification should result in an `Uncaught Error: Undefined constant` error.

kaitlinnewson: Yes, I forgot to note that I've set it to `self::` instead

Issue #10279, opened 2024-08-05
API request with API Token requires CSRF token


taslangraham:
**Describe the bug**
The [CSRF Token section](https://docs.pkp.sfu.ca/dev/documentation/en/architecture-authentication#csrf-tokens) of the PKP documentation states that 

> CSRF tokens are not required for requests to the API when the [API Token](https://docs.pkp.sfu.ca/dev/api/#api-token) is used.

However, the following CSRF error response is returned when requests (`PUT`, `DELETE`, `POST`) are made to the API with the API Token:
```"error": "The form could not be submitted. You may have been logged out. Please reload the page and try again."```

**To Reproduce**
To reproduce using Postman or a similar
1. Log in
2. Navigate to the User Profile section and select the API Key tab
3. Click the `Create API Key` button to create an API key
4. Follow steps [here](https://docs.pkp.sfu.ca/dev/documentation/en/resources#postman-api) to setup PKP collection in postman with API Token
5. Make a  `PUT`, `DELETE`, or  `POST` request.
6. Observe response

**What application are you using?**
 OJS (main)

**Additional information**
Please add any screenshots, logs or other information we can use to investigate this bug report.

taslangraham:
@touhidurabir  This issue was introduced via the work done on https://github.com/pkp/pkp-lib/issues/7698.

The [`isApiRequest`](https://github.com/pkp/pkp-lib/blob/a1e4461b4fc32bbe8402ec1610e6beb6cf9d486c/classes/middleware/ValidateCsrfToken.php#L52-L55) method checks for the `apiToken` in the request's query params, however, the token exists in the authorization headers.

I'll open a PR with a fix

taslangraham:
PR - https://github.com/pkp/pkp-lib/pull/10286/files @touhidurabir

touhidurabir:
@taslangraham I added a suggestion at https://github.com/pkp/pkp-lib/pull/10286/files#r1708717742 . Also do a sub module commit and app level to PR so that we can confirm all tests are passing before merging it .

taslangraham:
@touhidurabir I've made the suggested changes. I've also included a comment indicating that the use of API Token in query param was [deprecated](https://docs.pkp.sfu.ca/dev/api/ojs/3.4.html) in 3.4.

Additionally, I created tests PRs for each app:
ojs - https://github.com/pkp/ojs/pull/4391
omp - https://github.com/pkp/omp/pull/1667 
[...]

PR #10286, opened 2024-08-07
pkp#10279 fix incorrect check for API Token

taslangraham:
for https://github.com/pkp/pkp-lib/issues/10279

### DIFF
--- a/classes/middleware/ValidateCsrfToken.php
+++ b/classes/middleware/ValidateCsrfToken.php
@@ -47,11 +47,21 @@ class ValidateCsrfToken
     }
 
     /**
-     * Check if this is a API request
+     * Check if this is an API request
      */
     protected function isApiRequest(Request $request): bool
     {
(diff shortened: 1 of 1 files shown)


classes/middleware/ValidateCsrfToken.php
```diff
@@ -51,7 +51,7 @@ public function handle(Request $request, Closure $next)
      */
     protected function isApiRequest(Request $request): bool
     {
-        return $request->query('apiToken', null) ? true : false;
+        return !!$request->header('authorization');
```
touhidurabir, line 54: This is ok but I think we need to support both case that is as `query param` and `authorization header` . I suggest to update it as 
```php
// API Token may be sent as request query param
if ($request->query('apiToken', null)) {
    return true;
}

// API Token may set as request Authorization header
if ($request->header('authorization', null)) {
    return true;
}

return false;
```

Issue #10287, opened 2024-08-08
PHP fatal error when editing Publisher Library file


jyhein:
**Describe the bug**
PHP fatal error when editing Publisher Library file.

**To Reproduce**
In Settings > Workflow > Publisher Library: 
1. add file: Type any, upload file, and  check 'Public Access' -> OK
2. Try to 'Edit' the file -> Error (pic)
![error](https://github.com/user-attachments/assets/6e52380a-be54-49eb-94b5-e2a9083ffa05)

PHP Error Log:
PHP Fatal error:  Type of PKP\controllers\grid\settings\library\form\EditLibraryFileForm::$contextId must be int (as in class PKP\controllers\grid\files\form\LibraryFileForm) in ojs/lib/pkp/controllers/grid/settings/library/form/EditLibraryFileForm.php on line 27

**What application are you using?**
 OJS version 3.5

taslangraham:
I opened a PR for this @asmecher. It also includes a few other minor fixes in the file. 
PR - https://github.com/pkp/pkp-lib/pull/10291
Test(ojs) - https://github.com/pkp/ojs/pull/4392

asmecher:
Added [a comment](https://github.com/pkp/pkp-lib/pull/10291/files#r1715825760), thanks, @taslangraham!

taslangraham:
thanks for the review @asmecher, I've made the suggested change

PR #10291, opened 2024-08-08
PKP#10287 fix types

taslangraham:
For https://github.com/pkp/pkp-lib/issues/10287

### DIFF
--- a/controllers/grid/settings/library/form/EditLibraryFileForm.php
+++ b/controllers/grid/settings/library/form/EditLibraryFileForm.php
@@ -26,16 +26,12 @@ use PKP\file\TemporaryFileManager;
 
 class EditLibraryFileForm extends LibraryFileForm
 {
-    /** @var LibraryFile the file being edited, or null for new */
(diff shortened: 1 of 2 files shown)


controllers/grid/settings/library/form/EditLibraryFileForm.php
```diff
@@ -26,16 +26,15 @@
...
 {
-    /** @var LibraryFile the file being edited, or null for new */
-    public $libraryFile;
+    /** The file being edited, or null for new */
+    public ?LibraryFile $libraryFile;
 
-    /** @var int the id of the context this library file is attached to */
-    public $contextId;
+    /** The id of the context this library file is attached to */
+    public int $contextId;
```
asmecher, line 33: Since this attribute is already present in the parent class through constructor promotion, it shouldn't be re-declared here. The declaration and comment can be removed from this class.

taslangraham: good catch! I've updated it

PR #10295, opened 2024-08-10
pkp#6055 Bug fix: address missing category selection after failed submission form validation

taslangraham:
For #6055


jonasraoni (approved):
I've just skimmed through, but I think it's better to keep the old variable name and just feed the missing `$assignedCategories`

PR #10298, opened 2024-08-12
pkp/pkp-lib#10157 Moved issueIds out of publication_settings

### LINKED ISSUE #10157: Move publication to issue_id relationship out of publication_settings table
**Describe the bug**
Raised in https://github.com/pkp/pkp-lib/discussions/7975 and elsewhere:

Storing the issue ID for the publication in the `publication_settings` table is problematic for several reasons:
- It doesn't support relational integrity
- It performs poorly due to `setting_value` essentially being a blob column

For these reasons, the issue ID should instead be stored as an `issue_id` column in `publications`.

Note that this column will apply to OJS only, as OMP and OPS don't have issue-based publication!

Hafsa-Naeem:
fix #10157

### DIFF
--- a/classes/galley/DAO.php
+++ b/classes/galley/DAO.php
@@ -260,7 +260,6 @@ class DAO extends EntityDAO implements RepresentationDAOInterface
 
         $q = DB::table('publication_galleys', 'g')
             ->leftJoin('publications AS p', 'p.publication_id', '=', 'g.publication_id')
-            ->leftJoin('publication_settings AS ps', 'ps.publication_id', '=', 'p.publication_id')
             ->leftJoin('submissions AS s', 's.submission_id', '=', 'p.submission_id')
             ->leftJoin('submission_files AS sf', 'g.submission_file_id', '=', 'sf.submission_file_id')
             ->when($pubIdType != null, fn (Builder $q) => $q->leftJoin('publication_galley_settings AS gs', 'g.galley_id', '=', 'gs.galley_id'))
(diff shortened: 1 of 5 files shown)


classes/publication/Collector.php
```diff
@@ -79,6 +81,17 @@ public function filterBySubmissionIds(?array $submissionIds): self
         return $this;
     }
 
+    /**
+     * Set the filter for issue IDs to limit the publications retrieved.
+     *
+     * @param array|null $issueIds An array of issue IDs to filter by, or null to remove the filter.
```
jonasraoni, line 87: When possible, I think the shorter syntax is preferred:

```suggestion
     * @param ?int[] $issueIds An array of issue IDs to filter by, or null to remove the filter.
```

classes/galley/DAO.php
```diff
@@ -276,7 +275,8 @@ public function getExportable(int $contextId, $pubIdType = null, $title = null,
             ->when($pubIdType != null, fn (Builder $q) => $q->where('gs.setting_name', '=', "pub-id::{$pubIdType}")->whereNotNull('gs.setting_value'))
             ->when($title != null, fn (Builder $q) => $q->where('pst.setting_name', '=', 'title')->where('pst.setting_value', 'LIKE', "%{$title}%"))
             ->when($author != null, fn (Builder $q) => $q->whereRaw("CONCAT(COALESCE(asgs.setting_value, ''), ' ', COALESCE(asfs.setting_value, '')) LIKE ?", ["%{$author}%"]))
-            ->when($issueId != null, fn (Builder $q) => $q->where('ps.setting_name', '=', 'issueId')->where('ps.setting_value', '=', $issueId)->where('ps.locale', '=', ''))
+            ->when($issueId != null, fn (Builder $q) => $q->where('p.issue_id', '=', $issueId))
```
jonasraoni, line 278:
```suggestion
            ->when($issueId !== null, fn (Builder $q) => $q->where('p.issue_id', '=', $issueId))
```

asmecher:
@Hafsa-Naeem, the `issue_id` / `publication_id` relationship is only applicable to OJS, but we're adding it here to `pkp-lib` which will also affect OMP and OPS. Could you see if you can isolate the changes to OJS only? The best case would be for `issue_id` code only to appear in the OJS repo, but if it is necessary for some of those concerns to move into pkp-lib, then it would need to be an optional relationship.

asmecher (changes requested):
https://github.com/pkp/pkp-lib/pull/10298#issuecomment-2313405065

Issue #9999, opened 2024-05-29
Track Editor Workload While Assigning Participants


Devika008:
### Describe the problem you would like to solve
Currently, there isn't an efficient method to track the workload of an editor and make informed decisions when assigning users to submissions.

### Describe the solution you'd like
We need to enhance the editor assignment process by providing more detailed information. This way, the assigning editor can view useful data to guide their assignments, such as the workload of each editor and which editor possesses the most appropriate subject knowledge.

Here is my design proposition for the same:

![Image](https://github.com/pkp/pkp-lib/assets/112014397/f666e0b2-9365-4890-be5e-f9ed1697e7a2)

Design Considerations
a. I've implemented the new table design, which is also featured in the storybook and utilized for user invitations and submissions lists. This table is highly accessible, with no layers of information, making it easier to translate this screen for users with different abilities and in different languages as well.
b. The layout of the information helps reduce cognitive load.

Some considerations as listed by Tim that I fully agree with
a. The Active Submission Column will show all assignments currently from Submission until they are Published. Submissions that have been Declined or Published will not be counted
b. The Active Submission Column will only take the selected Role into account - e.g. if the Journal Editor role is selected then it should only show the active assignments that the user has as a Journal Editor. If that editor also has an assignment as an alternative role (e.g. Author) then this would not count unless the assignment role is changed to that role

Who is asking for this feature?
Tim Wakeford from Ubiquity 

PR pkp-lib: https://github.com/pkp/pkp-lib/pull/10312

twakeford:
We're happy with it from our side.

The only comment from us is that the 'Same institution as author' function raised quite a lot of back and forth when implemented for the reviewer invitation function, as it is a match based on the affiliation string, which has room for missing valid matches due to the entry being slightly different (e.g. a typo in the affiliation). We still think that this is the most viable option though until ROR functionality is expanded, and the function already implemented for reviewer invitations has been very well received by our editors. Should the exact string match not be optimal then we suggest that a partial match or possible match is used, so that a flag is raised based on similarity rather than exact match. i.e. If the affiliation is 'University of Liverpool', 'Uni of Liverpool', or 'Liverpool University' then the table would still bring up a flag, but perhaps we alter the text presented on the screen so it isn't as definite as 'Same institution as author' in case of mismatches. Ultimately, it is just a flag to tell the editor to look into it prior to assignment.

asmecher:
[...]

PR #10312, opened 2024-08-14
pkp#9999 feature:Track Editor Workload While Assigning Participants

salmanm2003:
Hi @asmecher,
Regarding this issue [https://github.com/pkp/pkp-lib/issues/9999](https://github.com/pkp/pkp-lib/issues/9999)
 I’ve made the changes to display the details mentioned in the conversation/screenshot. However, I couldn’t implement the affiliation match with ROR data because the plugin isn’t compatible with OJS 3.5. Please let me know when a compatible ROR version for 3.5 is available so I can complete that as well.

### DIFF
--- a/classes/user/Collector.php
+++ b/classes/user/Collector.php
@@ -30,6 +30,7 @@ use PKP\core\interfaces\CollectorInterface;
 use PKP\facades\Locale;
 use PKP\identity\Identity;
 use PKP\plugins\Hook;
+use PKP\submission\PKPSubmission;
 use PKP\userGroup\relationships\enums\UserUserGroupMastheadStatus;
 use PKP\userGroup\relationships\enums\UserUserGroupStatus;
 
@@ -65,6 +66,7 @@ class Collector implements CollectorInterface
     public ?string $registeredAfter = null;
     public ?string $status = self::STATUS_ACTIVE;
     public bool $includeReviewerData = false;
+    public ?array $includeActiveSubmissions = null;
     public ?array $assignedSectionIds = null;
     public ?array $assignedCategoryIds = null;
     public ?array $settings = null;
@@ -243,6 +245,17 @@ class Collector implements CollectorInterface
     public function includeReviewerData(bool $includeReviewerData = true): self
     {
         $this->includeReviewerData = $includeReviewerData;
+        return $this;
+    }
+
+    /**
+     * Defines whether active submissions count should be included
+     */
+    public function includeActiveSubmissions(int $userGroupId): self
+    {
+        $this->includeActiveSubmissions = [
+            'user_group_id' => $userGroupId,
+        ];
         return $this;
     }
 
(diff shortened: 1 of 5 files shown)


classes/user/Collector.php
```diff
@@ -246,6 +248,17 @@ public function includeReviewerData(bool $includeReviewerData = true): self
         return $this;
     }
 
+    /**
+     * Defines whether active submissions count should be included
+     */
+    public function includeActiveSubmissions(int $userGroupId): self
```
Vitaliy-1, line 254: What about naming this `includeActiveSubmissionsByUserGroupIds`? And passing the array of IDs instead of int?

classes/user/Collector.php
```diff
@@ -246,6 +248,17 @@ public function includeReviewerData(bool $includeReviewerData = true): self
...
+    {
+        $this->includeActiveSubmissions = [
+            'user_group_id' => $userGroupId,
```
Vitaliy-1, line 257: Do we need assigning associated array if filtering only by 1 parameter?

classes/user/Collector.php
```diff
@@ -425,6 +438,15 @@ public function getQueryBuilder(): Builder
 
         $query = DB::table('users', 'u')
             ->select('u.*')
+            // Adds active submissions that are assigned to each user
+            ->when($this->includeActiveSubmissions != null, fn (Builder $query) => $query->selectSub(function ($query) {
+                $query->from('stage_assignments as sa')
+                    ->join('submissions as s', 'sa.submission_id', '=', 's.submission_id')
+                    ->whereColumn('sa.user_id', 'u.user_id')
+                    ->where('sa.user_group_id', '=', $this->includeActiveSubmissions['user_group_id'])
+                    ->whereNotIn('s.status', [PKPSubmission::STATUS_PUBLISHED, PKPSubmission::STATUS_DECLINED])
```
Vitaliy-1, line 447: I think we should also exclude incomplete submissions

classes/user/Collector.php
Vitaliy-1: But that's on your opinion

Vitaliy-1 (commented):
Thanks, this is looking great! just very minor comments

classes/user/Collector.php
salmanm2003: I have renamed the Function name, however the I kept the param as int as it is going to be always one group id when filtering the users?

classes/user/Collector.php
salmanm2003: The incomplete submissions have been excluded as well.

classes/user/Collector.php
salmanm2003: I have tried that, but oddly when passing it to the query, it doesn't retrieve the active submissions number correctly for some reason. so I had to keep it as it is.

controllers/grid/users/userSelect/UserSelectGridCellProvider.php
```diff
@@ -51,12 +52,24 @@ public function getTemplateVarsFromRowColumn($row, $column)
...
+        $interests = $interestManager->getInterestsForUser($element);
+        $interests = ($interests) ? implode(', ', $interests) : '';
         switch ($column->getId()) {
             case 'select': // Displays the radio option
                 return ['rowId' => $row->getId(), 'userId' => $this->_userId];
 
             case 'name': // User's name
                 return ['label' => $element->getFullName()];
+
+            case 'activeSubmissions': // User's name
```
asmecher, line 65: Comments are wrong for these 3 `case` statements.

Issue #10292, opened 2024-08-08
Convert controlled vocabulary classes to Eloquent models


asmecher:
There are several classes that use the old DAO storage model to handle controlled vocabularies and related code:
- controlled_vocab_entries and controlled_vocab_entry_settings (these will require a trait for settings tables, which @Vitaliy-1 is working on)
- controlled_vocabs
- user_interests

In preparation for extensions to the controlled vocabulary toolset in support of vocabulary plugins that implement specific vocabs (e.g. LCSH etc), the storage model for these tables should be migrated to Eloquent.

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10324
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4403
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1722
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/788
[citationStyleLanguage](https://github.com/pkp/citationStyleLanguage) --> https://github.com/pkp/citationStyleLanguage/pull/127
[jatsTemplate](https://github.com/pkp/jatsTemplate) --> https://github.com/pkp/jatsTemplate/pull/54
[oaiJats](https://github.com/pkp/oaiJats) --> https://github.com/pkp/oaiJats/pull/51
[reviewReport](https://github.com/pkp/reviewReport) --> https://github.com/pkp/reviewReport/pull/52

touhidurabir:
@asmecher please review the PRs at  https://github.com/pkp/pkp-lib/issues/10292#issue-2456328759 . There is one final [TODO](https://github.com/pkp/pkp-lib/pull/10324/files#diff-62f11bae292e7885dc81f6f5033a3c04607b3ff6c24337fec1fb769a0a639f7bR41-R47) to resolve, but it's ready for review .

touhidurabir:
@asmecher Tests are all passing , Please review the PRs at https://github.com/pkp/pkp-lib/issues/10292#issue-2456328759

mreiko:
hi @asmecher Can this be considered 'done'?

asmecher:
No, I still have to review and merge first!

asmecher:
@touhidurabir, I've added some initial comments on the pkp-lib PR: https://github.com/pkp/pkp-lib/pull/10324#pullrequestreview-2449424243

[...]

PR #10324, opened 2024-08-16
pkp/pkp-lib#10292 Controlled Vocab DAO to Eloquent Model

touhidurabir:
for pkp/pkp-lib#10292


classes/submission/SubmissionAgencyVocab.php
```diff
@@ -0,0 +1,90 @@
...
+        return ['submissionAgency'];
+    }
+
+    /**
+     * Get agencies for a specified submission ID.
+     *
+     * @param int $assocType DO NOT USE: For <3.1 to 3.x migration pkp/pkp-lib#6213
+     * 
+     */
+    public function scopeGetAgencies(
```
touhidurabir, line 41: I feel like using the scope methods for these in these extended model classes does not make sense . It does not utilise the scope's methods capabilities and here we had to pass the `$builder` which has no use . Perhaps a simple static method is good enough .

classes/submission/SubmissionAgencyVocab.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    public const CONTROLLED_VOCAB_SUBMISSION_AGENCY = 'submissionAgency';
+
+    /**
+     * Get the list of localized additional fields to store.
+     */
+    public function scopeGetLocaleFieldNames(Builder $query): array
+    {
+        return ['submissionAgency'];
+    }
```
touhidurabir, line 33: Not sure if these methods have any use A

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,146 @@
...
 
     /**
-     * get assoc id
+     * Indicates if the model should be timestamped.
      *
-     * @return int
+     * @var bool
      */
-    public function getAssocId()
+    public $timestamps = false;
```
touhidurabir, line 41: Wondering is there is specific reason we are not adding timestamp columns to the table when we are updating those to use the eloquent model ? It does not have any effect right away but in long run we may start using those .

classes/controlledVocab/Repository.php
```diff
@@ -0,0 +1,199 @@
...
+                }
+
+                UserInterest::create([
+                    'userId' => $userId,
+                    'controlledVocabEntryId' => $interestEntry->getId(),
+                ]);
+            });
+        
+        DB::commit();
+    }
```
touhidurabir, line 198: perhaps this one can be moved to `UserInterest` model as a static method but seems kind of too much to have for a model . Then again this also seems kind of out of place in the vocab repo but then again it does handle a specific type of vocab .

classes/user/UserInterest.php
```diff
@@ -0,0 +1,161 @@
...
+
+        // Sort by name.
+        $interests = $iterator->toArray();
+        usort($interests, function ($s1, $s2) {
+            return strcmp($s1->getInterest(), $s2->getInterest());
+        });
+
+        // Turn back into an iterator.
+        return new ArrayItemIterator($interests);
+    }
```
touhidurabir, line 160: this needed to be refactored but depends on how we refactor the all `......EntryDAO` classes along with the settings table .

classes/user/UserInterest.php
```diff
@@ -0,0 +1,161 @@
...
+    public function scopeWithUserId(Builder $query, int $userId): Builder
+    {
+        return $query->where('user_id', $userId);
+    }
+
+    /**
+     * Get a list of controlled vocabulary entry IDs (corresponding to interest keywords) 
+     * attributed to a user
+     */
+    public static function getUserInterestIds(int $userId): array
```
touhidurabir, line 100: It is possible to centralise these (this one and below other static method) in the controlled vocab repo but these methods are only required for user interest . But they seems like out of place methods for a model . At the same time they are more like some generic method that should be accessible from a repo . Any thoughts ?

classes/controlledVocab/ControlledVocab.php
asmecher: I have no objection to adding these -- but as you say, we're not using them yet, so it hasn't been a priority.

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,146 @@
...
-     * Get symbolic name.
-     *
-     * @return string
+     * Scope a query to only include notes with a specific user ID.
      */
-    public function getSymbolic()
+    public function scopeWithSymbolic(Builder $query, string $symbolic): Builder
     {
-        return $this->getData('symbolic');
+        return $query->where(DB::raw('LOWER(symbolic)'), strtolower($symbolic));
```
asmecher, line 88: Is the case insensitivity a new introduction? I'd rather not add this unless necessary. It looks like it's supported some places and not others.

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,146 @@
...
 
     /**
-     * Set associated type.
+     * Compatibility function for including notes in grids.
      *
-     * @param int $assocType
+     * @deprecated 3.5. Use $model or $model->$field instead. Can be removed once the DataObject pattern is removed.
      */
-    public function setAssocType($assocType)
+    public function getData(?string $field): mixed
```
asmecher, line 78: Is `getData(null)` ever called in practice? If not, I'd rather forestall that calling convention.

classes/controlledVocab/Repository.php
asmecher: Agreed, it shouldn't be here. I'd prefer `UserInterest` as a static method, or perhaps better a `Repository` class alongside `UserInterest`.

classes/publication/DAO.php
```diff
@@ -370,10 +348,10 @@ protected function deleteAuthors(int $publicationId)
      */
     protected function setControlledVocab(Publication $publication)
     {
-        $publication->setData('keywords', $this->submissionKeywordDao->getKeywords($publication->getId()));
-        $publication->setData('subjects', $this->submissionSubjectDao->getSubjects($publication->getId()));
-        $publication->setData('disciplines', $this->submissionDisciplineDao->getDisciplines($publication->getId()));
-        $publication->setData('supportingAgencies', $this->submissionAgencyDao->getAgencies($publication->getId()));
+        $publication->setData('keywords', SubmissionKeywordVocab::getKeywords($publication->getId()));
```
asmecher, line 351: Something to work on: we're trying to avoid relying on static methods, as they're hard to mock or extend. This might be better placed in a Repository class.

classes/submission/SubmissionAgencyVocab.php
asmecher: Just looking at the `DO NOT USE` note -- is this code that got left over after the 3.0.x to 3.1.x upgrade got removed? (Also more below.)

classes/submission/SubmissionAgencyVocab.php
```diff
@@ -0,0 +1,90 @@
...
+
+namespace PKP\submission;
+
+use APP\facades\Repo;
+use Illuminate\Database\Eloquent\Builder;
+use PKP\controlledVocab\ControlledVocab;
+use PKP\core\PKPApplication;
+
+
+class SubmissionAgencyVocab extends ControlledVocab
```
asmecher, line 23: Do you think we could get rid of these classes entirely? If all we're doing is providing a convenience `getAgencies` function, rather than `withSymbolic(SOME_AGENCIES_CONSTANT, ...)`, then I don't think it's worth the extra classes.

classes/user/UserInterest.php
asmecher: I think this one is a bit of a unique case at the moment, since we don't have other relationship tables mapping `controlled_vocab_interest_id` to another entity. We might need them more in the future, but I don't mind if we don't refactor this to be reusable until that's needed.

asmecher (changes requested):
Thanks, I've added some early review comments!

classes/metadata/MetadataProperty.php
```diff
@@ -423,8 +423,16 @@ public function isValid($value, $locale = null)
...
                                 // Try to translate the string value into a controlled vocab entry
-                                $controlledVocabEntryDao = DAORegistry::getDAO('ControlledVocabEntryDAO'); /** @var ControlledVocabEntryDAO $controlledVocabEntryDao */
-                                if (!is_null($controlledVocabEntryDao->getBySetting($value, $symbolic, $assocType, $assocId, 'name', $locale))) {
+                                $entry = ControlledVocabEntry::query()
+                                    ->whereHas(
+                                        'controlledVocab',
+                                        fn ($query) => $query->withSymbolic($symbolic)->withAssoc($assocType, $assocId)
+                                    )
+                                    ->withLocale($locale)
+                                    ->withSetting($symbolic, $value)
```
touhidurabir, line 432: here previously setting name passed as `name` but is that correct as for controlled vocabs we have specific settings ?

classes/migration/upgrade/v3_5_0/I10292_RemoveControlledVocabEntrySettingType.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Reverse the migration.
+     */
+    public function down(): void
+    {
+        Schema::table('controlled_vocab_entry_settings', function (Blueprint $table) {
+            $table->string('setting_type', 6);
+        });
+    }
```
touhidurabir, line 41: dropping the column `setting_type` seems reasonable as it seems to have no use and our newer eloquent based implementation for setting tables does not have support for it either .

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,201 @@
...
-     *
      * @return array $controlledVocabEntryId => name
      */
-    public function enumerate($settingName = 'name')
-    {
-        $controlledVocabDao = DAORegistry::getDAO('ControlledVocabDAO'); /** @var ControlledVocabDAO $controlledVocabDao */
-        return $controlledVocabDao->enumerate($this->getId(), $settingName);
-    }
-}
+    public function enumerate(?string $settingName = null): array
```
touhidurabir, line 163: Here `$settingName` previously have default value to `name` but that seems not right as for controlled vocab, we have a specific sets to settings only . Or I am missing something ?

classes/user/interest/Repository.php
```diff
@@ -0,0 +1,144 @@
...
+                );
+            
+            collect($currentInterests->pluck('id'))
+                ->merge($newInterestIds)
+                ->each(fn ($interestId) => UserInterest::create([
+                    'userId' => $user->getId(),
+                    'controlledVocabEntryId' => $interestId,
+                ]));
+            
+            Repo::controlledVocab()->resequence($controlledVocab->id);
```
touhidurabir, line 133: do we need to resequence the vocabs for user interest, don't see any benefit or usage of it ?

tests/classes/form/validation/FormValidatorControlledVocabTest.php
```diff
@@ -3,86 +3,83 @@
...
-            ->getMock();
-        $mockControlledVocab->setId(1);
-        $mockControlledVocab->setAssocType(Application::ASSOC_TYPE_CITATION);
-        $mockControlledVocab->setAssocId(333);
-        $mockControlledVocab->setSymbolic('testVocab');
+        $assocId = (DB::table("publications")
+            ->select("publication_id as id")
+            ->orderBy("publication_id", "desc")
+            ->first()
+            ->id ?? 0) + 100;
```
touhidurabir, line 40: rather than setting a fixed value like `333`, it's better to build one based on what we have in DB and add a const value (like here `100`) to just to be safe .

tests/classes/form/validation/FormValidatorControlledVocabTest.php
```diff
@@ -3,86 +3,83 @@
...
-        $form->setData('testData', '2');
+        $form->setData('testData', $controlledVocabEntryId2);
         self::assertTrue($validator->isValid());
 
-        $form->setData('testData', '3');
+        $form->setData('testData', 3);
         self::assertFalse($validator->isValid());
+
+        // Delete the test vocab along with entries
+        ControlledVocab::find($testControlledVocab->id)->delete();
```
touhidurabir, line 83: With implementation of setting table based eloquent, we can not add any arbitrary setting data so only options seems to create data and delete at the end of test . Another option can be is to backup/restore specific tables or full DB but that make the test time high.

classes/submission/SubmissionAgencyVocab.php
touhidurabir: All of there classes have been removed . By removing these extra classes and few related others , our code also reduced significantly .

classes/core/SettingsBuilder.php
```diff
@@ -75,6 +76,42 @@ public function update(array $values)
...
+                        'setting_name' => $settingName,
+                        'setting_value' => $settingValue,
+                    ];
+                }
+            }
+        });
+
+        if (count($rows) > 0) {
+            DB::table($this->model->getSettingsTable())->insert($rows);
+        }
```
touhidurabir, line 114: @Vitaliy-1 this particular change is seems required as when updating a model instance with new settings/multilingual column, it does not add the data in the settings table . for example , a announcement model data may not have `descriptionShort` or may have `descriptionShort` for only `en` . But if we try to update it as following 
```php
$announcement->update([
    "descriptionShort" => [
        "en" => "<p>Test Announcement 2025 nothing</p>"
        "fr_CA" => "<p>something new</p>"
    ],
]);
```
it will not add a new row in the settings table for `fr_CA` which is expected . I think this will cause issue when updating a multilingual or normal setting field as it will never get inserted . Please take a look of above implementation .

classes/core/SettingsBuilder.php
```diff
@@ -288,8 +331,19 @@ protected function filterRow(stdClass $row, string|array $columns = ['*']): stdC
...
+        $columns = collect($columns)
+            ->map(
+                fn (string $column): string => in_array($column, $settingColumns)
+                    ? $column
+                    : Str::snake($column)
+            )
+            ->toArray();
+
+        foreach ($row as $property => $value) {
+            if (!in_array($property, $columns) && isset($row->{$property})) {
```
touhidurabir, line 346: @Vitaliy-1 this change seems required as with current implementation , we do not get the desired outcome with specifying the columns as such 
```php
$announcement = Announcement::find(31, ["title"]);
```
Also throws a lot of warnings . Please take a look .

classes/core/traits/ModelWithSettings.php
```diff
@@ -65,8 +65,13 @@ abstract protected function ensureCastsAreStringValues($casts);
     public function __construct(array $attributes = [])
     {
         parent::__construct($attributes);
+        
         if (static::getSchemaName()) {
             $this->setSchemaData();
+        } else {
+            if (!empty($this->fillable)) {
+                $this->mergeFillable(array_merge($this->getSettings(), $this->getMultilingualProps()));
+            }
```
touhidurabir, line 74: @Vitaliy-1 this changes seems required to populate the proper `fillable` as in the `if` condition, we are only considering entities with schema maps but also need to consider for entities with settings with no schema map .

classes/core/traits/ModelWithSettings.php
```diff
@@ -183,4 +197,35 @@ protected function id(): Attribute
...
+        $settingsWithMultilingual = array_merge($this->getSettings(), $this->getMultilingualProps());
+        $camelKey = Str::camel($key);
+        
+        // Check if this column included in setting and multilingula props and not set to guarded
+        if (in_array($camelKey, $settingsWithMultilingual) && !in_array($camelKey, $this->getGuarded())) {
+            return true;
+        }
+        
+        return in_array($key, (array)static::$guardableColumns[get_class($this)]);
+    }
```
touhidurabir, line 230: @Vitaliy-1 this override of `\Illuminate\Database\Eloquent\Concerns\GuardsAttributes::isGuardableColumn` seems required when defining any guarded attributes for entities with no schema map as for schema mapped entities, we always generate the `fillable` .

classes/core/casts/MultilingualSettingAttribute.php
```diff
@@ -0,0 +1,55 @@
...
+            );
+        }
+        
+        if (is_string($value)) {
+            return [$key =>  [Locale::getLocale() => $value]];
+        }
+
+        return [$key => array_filter($value)];
+    }
+}
```
touhidurabir, line 55: related to https://github.com/pkp/pkp-lib/issues/10476 .

classes/core/traits/ModelWithSettings.php
```diff
@@ -46,27 +53,43 @@ abstract public function getTable();
...
         if (static::getSchemaName()) {
             $this->setSchemaData();
+        } else {
+            $this->generateAttributeCast(
+                collect($this->getMultilingualProps())
+                    ->flatMap(
+                        fn (string $attribute): array => [$attribute => MultilingualSettingAttribute::class]
+                    )
+                    ->toArray()
+            );
```
touhidurabir, line 88: related to https://github.com/pkp/pkp-lib/issues/10476

classes/core/SettingsBuilder.php
```diff
@@ -75,11 +76,63 @@ public function update(array $values)
...
+                        'setting_name' => $settingName,
+                        'setting_value' => $settingValue,
+                    ];
+                }
+            }
+        });
+
+        if (count($rows) > 0) {
+            DB::table($this->model->getSettingsTable())->insert($rows);
+        }
```
touhidurabir, line 129: @Vitaliy-1  Handle the update with multilingual prop .

classes/core/SettingsBuilder.php
Vitaliy-1: Good spot! It seems that my idea to update settings in 1 query doesn't work if the value wasn't set. Considering the example query:
```sql
UPDATE announcement_settings SET setting_value = 
CASE WHEN setting_name='title' AND locale='en' THEN 'English Title' WHEN setting_name='title' AND locale='fr_CA' THEN 'French TItle' ELSE setting_value END
WHERE announcement_id = 16;
```
I really don't now SQL well enough to say if we can modify it to create a new row if it doesn't exist. 
On the other hand, we can just run a series of conventional sql queries to update existing rows or insert new ones. This is a quick example: https://github.com/pkp/pkp-lib/pull/10542/files
Let me know if you want me to resolve this bug

touhidurabir: @Vitaliy-1 I am not sure also how to update and insert in a single query and the query may become too complex to work in future too . However I think we need to consider 3 cases here 
1. Updating the existing ones in table in passed through update request .
2. Adding new ones that do not exists in table passed through update request
3. deleting what is not passed in the update request . 

for for example , if an `announcement` model instance have 4 settings row for locales `en`, `fr_CA`, `fr_FR`, `pt_BR`, for `title` prop and we make a request like 
```php
$announcement->update([
    "title" => [
        "en" => "Title in EN"
        "fr_CA" => "Title in fr_CA",
        "fr_FR" => "Title in fr_FR",
        "el" => "Title in el",
    ],
]);
```
So in the above example , we updating `en`, `fr_CA`, `fr_FR` , adding new `el` and no details about `pt_BR` which need to be removed . I have added an implementation at https://github.com/pkp/pkp-lib/pull/10324/files#diff-4dcc09fe73d537128c17c4bbb260ad4ba24ccc0ecb8cdc7d0b47ebc7986e2cedR79-R129 but that run 3 queries(update, insert and delete) . 

Maybe we can optimize in 2 queries by deleting all for that given props and inserting again. What you think ?

Vitaliy-1: In general, on the object level, all looks OK. All three conditions are met. So in theory we should have all needed info to update data once per row. Looking into it

Vitaliy-1: Aha, we have `PKPSchemaService::addMissingMultilingualValues()` for API calls that ensures consistency. And we don't have the same logic for Models.

Vitaliy-1 (approved):
Thanks, @touhidurabir. My only concert is the conflict with the `HasCamelCasing` trait. Otherwise, I like the approach, looks great!

asmecher (changes requested):
Hi Touidur! Some initial comments.

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -3,101 +3,189 @@
...
+    {
+        return $query->where('controlled_vocab_id', $controlledVocabId);
+    }
+
+    /**
+     * Scope a query to only include entries for a specific locale/s
+     */
+    public function scopeWithLocale(Builder $query, string|array $locale): Builder
+    {
+        if (is_array($locale)) {
```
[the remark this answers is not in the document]
touhidurabir: As per our discussion, we agree that this works as we intend to, so keeping this as it is .

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,201 @@
...
+            ->where(
+                fn ($query) => $query
+                    ->select('context_id')
+                    ->from('submissions')
+                    ->whereColumn(
+                        DB::raw(
+                            "(SELECT publications.submission_id 
+                            FROM publications 
+                            INNER JOIN {$this->table} 
+                            ON publications.publication_id = {$this->table}.assoc_id
```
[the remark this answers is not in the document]
touhidurabir: @asmecher I initially thought there can be a easier alternative but it will be bit extra work . First of all, this is how we were retrieving all the vocabs for a given context and the same concept (from controlled_vocabs -> publications -> submissions to get the context id ) used previous code base , just updated newer query builder mechanism . 

However if I am correct, we would like to allow different types of assoc , not just the publication one as we are doing it now . in that case, as still sometimes need to figure out vocabs for a given context id , it will become too difficult if there are multiple type assoc get associated with for a multi context system . 

My proposal is to allow the `context_id` column to `controlled_vocabs` table and that will solve this case with a lot easier approach .

classes/core/traits/EntityUpdate.php
```diff
@@ -74,7 +74,15 @@ public function updateSettings(array $props, int $modelId, $schema = null): void
...
             }
+
             if (!empty($propSchema->multilingual)) {
+                // first we will delete the settings entries for the locale keys which are not present in update query
+                DB::table($this->getSettingsTable())
+                    ->where($this->getPrimaryKeyName(), '=', $modelId)
+                    ->where('setting_name', '=', $propName)
+                    ->whereNotIn('locale', array_keys($props[$propName]))
+                    ->delete();
+
```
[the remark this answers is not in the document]
touhidurabir: @Vitaliy-1 can you please take a look at this modification .

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -3,101 +3,199 @@
...
+                    '=',
+                    $settingName
+                )
+                ->where(
+                    "{$this->getSettingsTable()}.setting_value",
+                    $match->operator(),
+                    $match->searchKeyword($settingValue)
+                )
+        );
+    }
```
touhidurabir, line 200: @asmecher as per your suggestion at https://github.com/pkp/pkp-lib/pull/10324#discussion_r1850814762, I have updated it handle string and array but that lead to 2 different implementation as we need to do a `IN` and `LIKE` based query . Wondering may be it's better to have one single `string|array` like implementation just for this case ?

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -3,101 +3,199 @@
...
-     * @return string
+     * @copydoc \PKP\core\traits\ModelWithSettings::getSettings
      */
-    public function getLocalizedName()
+    public function getSettings(): array
     {
-        return $this->getLocalizedData('name');
+        return array_merge(
+            $this->settings,
+            ControlledVocab::getDefinedVocabSymbolic(),
```
asmecher, line 100: As above -- mixing symbolic names and setting names?

classes/publication/DAO.php
```diff
@@ -432,10 +431,10 @@ protected function saveControlledVocab(array $values, int $publicationId)
      */
     protected function deleteControlledVocab(int $publicationId)
     {
-        $this->submissionKeywordDao->insertKeywords([], $publicationId);
-        $this->submissionSubjectDao->insertSubjects([], $publicationId);
-        $this->submissionDisciplineDao->insertDisciplines([], $publicationId);
-        $this->submissionAgencyDao->insertAgencies([], $publicationId);
+        ControlledVocab::query()
+            ->withSymbolics(ControlledVocab::getDefinedVocabSymbolic())
```
asmecher, line 435: Mixing symbolic names and setting names?

classes/controlledVocab/ControlledVocab.php
```diff
@@ -1,108 +1,199 @@
...
     }
 
     /**
-     * Get associated type.
-     *
-     * @return int
+     * Get the list of pre defined vocab symbolics
      */
-    public function getAssocType()
+    public static function getDefinedVocabSymbolic(): array
```
asmecher, line 88: Once you've cleaned up the symbolic name vs. setting name confusion, I think you can get rid of this function. I'd like to avoid hard-coding a list of all "acceptable" controlled vocabs, because third party code (e.g. plugins) might well want to extend the list.

classes/controlledVocab/ControlledVocab.php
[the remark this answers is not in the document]
asmecher: I think the only place this function is used is in the vocabs API controller, which is where the front end gets autosuggest possibilities. I think having a single endpoint for all vocab suggestions is actually the mistake -- it leads to the need for a single point of access for vocabs that might be attached to many different things, so we have to hard-code all the checks to make sure the list is allowed into a single place.

Instead, I suggest we break the endpoints where autosuggest responses are built into the appropriate places in the API. Autosuggestions for reviewer interests, which are global, should not be attached to the same endpoint as autosuggest results for article keywords, which are (probably) journal-specific.

@ewhanson, can we revisit the controlled vocab API handler? I don't think a single endpoint is the right design.

asmecher (changes requested):

classes/publication/DAO.php
touhidurabir: Not exactly, here we just need to delete all the controlled vocabs for the given publication but we can do it without taking into the consideration of symbolics as we only need need assoc type and assoc id . Updated .

classes/user/interest/UserInterest.php
```diff
@@ -0,0 +1,111 @@
...
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Builder;
+
+class UserInterest extends Model
+{
+    use HasCamelCasing;
+
+    public const CONTROLLED_VOCAB_INTEREST = 'interest';
+    public const CONTROLLED_VOCAB_INTEREST_ASSOC_TYPE = 0;
```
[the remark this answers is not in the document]
touhidurabir: @asmecher this is fine but the reason I did not go with this approach initially because our site_id with latest implementation is set to `null` as defined by const `SITE_CONTEXT_ID` . But in `controlled_vocabs` table , columns `assoc_id` expect an `int` . so either we have to always do an int casting when storing/updating/retrieving or update the `assoc_id` to be `nullable` which kind of feel weird . What you think ?

classes/core/SettingsBuilder.php
```diff
@@ -308,8 +344,19 @@ protected function filterRow(stdClass $row, string|array $columns = ['*']): stdC
         }
 
         $columns = Arr::wrap($columns);
-        foreach ($row as $property) {
-            if (!in_array($property, $columns)) {
+
+        // TODO : Investigate how to handle the camel to snake case issue. related to pkp/pkp-lib#10485
```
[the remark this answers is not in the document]
touhidurabir: This need to be resolved as part of another issue (pkp/pkp-lib#10485), we will work on it once this get merged . it's just there as a reference someone will work on that .

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -3,101 +3,199 @@
...
+                    "{$this->getTable()}.{$this->primaryKey}"
+                )
+                ->whereIn(DB::raw("{$this->getSettingsTable()}.locale"), $locales)
+        );
+    }
+
+    /**
+     * Scope a query to only include entries for a specific setting name and values
+     */
+    public function scopeWithSettings(Builder $query, string $settingName, array $settingValue): Builder
```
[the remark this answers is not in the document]
touhidurabir: unfortunately no, while the `scopeWithSetting` work with a single settings value and try to retrieve based on either exact or partial match , `scopeWithSettings` works with array of settings value so these 2 have different query implementation .

classes/controlledVocab/ControlledVocabEntry.php
touhidurabir: both are in `PKP\controlledVocab\ControlledVocabEntry`  also please see https://github.com/pkp/pkp-lib/pull/10324#discussion_r1888367467 .

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -3,101 +3,199 @@
...
-     * @param float $sequence
+     * @copydoc \PKP\core\traits\ModelWithSettings::getMultilingualProps
      */
-    public function setSequence($sequence)
+    public function getMultilingualProps(): array
     {
-        $this->setData('sequence', $sequence);
+        return array_merge(
+            $this->multilingualProps,
+            ControlledVocab::getDefinedVocabSymbolic(),
```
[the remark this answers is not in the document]
touhidurabir: @asmecher it does but we do need something similar like this . With our current design of eloquent model that works with settings table , we do need to define explicitly the settings and multilingual props when there is no schema define for entity like in this case for controlled vocabs. Now , for controlled vocab's case, we have following symbolics 
```
interest
submissionAgency
submissionDiscipline
submissionKeyword
submissionLanguage
submissionSubject
```
And these are the one also that get stored in `controlled_vocab_entry_settings` table for column `setting_name` as both a `setting and multilingual props` . 

Also another important factor is that currently we do not have any way for a third party (I guess mostly a plugin) to interact with these `eloquent models without schema` to add more settings or multilingual props . This probably need to be handled in a different issue .

classes/controlledVocab/ControlledVocabEntry.php
touhidurabir: please see https://github.com/pkp/pkp-lib/pull/10324#discussion_r1888497825 .

classes/controlledVocab/ControlledVocab.php
touhidurabir: @asmecher I can get rid of this function however in that case, we need to define each of these const for the settings and multilingual definition for `non schema based entity` , more details at https://github.com/pkp/pkp-lib/pull/10324#discussion_r1888497825 .

ewhanson:
@asmecher, re: revisiting the controlled vocab API handler comment above, I agree that splitting them into more single-purpose endpoints makes the most sense and will make their use cases more explicit. Were you hoping to get this type of change into 3.5 or would keeping this as proposed here for now be okay and updating this query afterwards with some API changes?

asmecher:
@ewhanson (and @touhidurabir), just given that we're so close to 3.5rc1, I think it's probably best left until post-3.5.0 -- but I'm open to proposals if anyone feels otherwise.

classes/core/traits/EntityUpdate.php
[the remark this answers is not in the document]
touhidurabir: > Do you mean that without this modification the locale value can't be updated to null in usual way? E.g., if we updating the description of the Announcement through the form by removing one locale data, it won't be removed from the database?

that's not the case . it works that way properly . 

> I'm thinking about situation when locale has been turned of globally. This means that values for the old locale will be deleted. And this removal will be inconsistent, per update request for each entity. But I prefer to preserve old locale values in case they would be needed again - when the user turns on the old locale again

considering this case (which I did not consider), we should not remove data for every other local . I am reverting this change .

ewhanson:
> @ewhanson (and @touhidurabir), just given that we're so close to 3.5rc1, I think it's probably best left until post-3.5.0 -- but I'm open to proposals if anyone feels otherwise.

@asmecher, I agree it would be best to put this until post-3.5. Think if it's working now as-is, best not to muck about too much right before release with refactors. I'll get an issue tagged against 3.6 created for it.

classes/user/form/RolesForm.php
```diff
@@ -43,7 +43,11 @@ public function fetch($request, $template = null, $display = false)
     {
         $templateMgr = TemplateManager::getManager($request);
 
-        $userGroupIds = UserGroup::getIdsByUserId($request->getUser()->getId());
+        $userGroupIds = UserGroup::query()
+            ->withUserIds([$request->getUser()->getId()])
+            ->get()
+            ->pluck('id')
+            ->toArray();
```
touhidurabir, line 50: @Vitaliy-1 without change, the profile --> roles tab is broken . and I dont see any static method `getIdsByUserId` defined for class `UserGroup` . Is this ok ?

classes/user/interest/UserInterest.php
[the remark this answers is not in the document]
touhidurabir: @As per our discussion, `controlled_vocabs` table's `assoc_id` set to nullable .

asmecher (approved):

Issue #10214, opened 2024-07-16
Outstanding editorial task automatic email content going wrong


yoruyenturk:
**Describe the bug**
Dear @asmecher, @touhidurabir 

The Editorial Reminder task email does not return the correct list. Is it possible to update this using the determineStatus function?

**To Reproduce**
1. EditorialReminder.php prepares the task content using the review round status, but prepares it incorrectly.
2. In the article shown in Picture 1, 1 reviewer has completed the evaluation and 1 reviewer is overdue. However, the editor of the article does not receive an overdue message in the Outstanding e-mail. Review Round table status 7 appears in the database (see Picture 2).
3. Similarly, there is an article with round status 8 in the database, but it is overdue in the backend but not in the mail (see Picture 3). You can see the database image from Picture 4.
4. The status statuses that appear in the system backend are not sent in the outstanding task e-mail. While the system shows notification using REVIEW_ASSIGNMENT_STATUS_XXX, the outstanding task email is sent using REVIEW_ROUND_STATUS_XXX.
5. It would be very helpful if jobs/email/EditorialReminder.php could be updated to provide a correct list.

**What application are you using?**
 OJS version 3.4.0-5

**Additional information**
Picture 1.
![image](https://github.com/user-attachments/assets/f1fbbe63-e86b-4a6a-808a-3d02bf0d945f)

Picture 2.
![image](https://github.com/user-attachments/assets/3f9f6103-1ed4-4b7f-b1cf-7545843c8919)

Picture 3.
![image](https://github.com/user-attachments/assets/a402710b-06b9-4d75-9a09-4b6150d11b58)

Picture 4.
![image](https://github.com/user-attachments/assets/bd27ad6b-7d8b-41d7-9a59-1742b5dd04f5)

**PRs**
#### Stable 3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10332
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4407 [TEST ONLY]

#### stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11083
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4742 [TEST ONLY]

#### main 
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10753
[...]

PR #10332, opened 2024-08-22
pkp/pkp-lib#10214 fixed review round outstanding status determining process

touhidurabir:
for pkp/pkp-lib#10214

### DIFF
--- a/classes/submission/reviewRound/ReviewRound.php
+++ b/classes/submission/reviewRound/ReviewRound.php
@@ -253,6 +253,7 @@ class ReviewRound extends \PKP\core\DataObject
                     break;
 
                 case ReviewAssignment::REVIEW_ASSIGNMENT_STATUS_AWAITING_RESPONSE:
+                case ReviewAssignment::REVIEW_ASSIGNMENT_STATUS_REQUEST_RESEND:
                 case ReviewAssignment::REVIEW_ASSIGNMENT_STATUS_ACCEPTED:
                     $anyIncompletedReview = true;
                     break;
--- a/jobs/email/EditorialReminder.php
+++ b/jobs/email/EditorialReminder.php
@@ -160,8 +160,11 @@ class EditorialReminder extends BaseJob
             ->from($context->getContactEmail(), $context->getLocalizedName(Locale::getLocale()))
             ->recipients([$editor])
             ->subject($emailTemplate->getLocalizedData('subject'))
-            ->body($emailTemplate->getLocalizedData('body'))
-            ->allowUnsubscribe($notification);
+            ->body($emailTemplate->getLocalizedData('body'));
+        
+        if ($notification) {
+            $mailable->allowUnsubscribe($notification);
+        }
 
         Mail::send($mailable);
 
--- a/jobs/email/EditorialReminder.php
+++ b/jobs/email/EditorialReminder.php
@@ -22,6 +22,7 @@ use APP\facades\Repo;
 use APP\notification\Notification;
 use APP\notification\NotificationManager;
 use APP\submission\Submission;
+use Carbon\Carbon;
 use Illuminate\Support\Facades\Mail;
 use PKP\context\Context;
 use PKP\db\DAORegistry;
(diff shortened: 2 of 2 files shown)


touhidurabir:
@Vitaliy-1 can you please review this PR ?

jobs/email/EditorialReminder.php
```diff
@@ -92,30 +93,31 @@ public function handle(): void
...
                     continue;
                 }
             }
 
             if (in_array($submission->getData('stageId'), [WORKFLOW_STAGE_ID_EDITING, WORKFLOW_STAGE_ID_PRODUCTION])) {
-                $lastActivityTimestamp = strtotime($submission->getData('dateLastActivity'));
-                if ($lastActivityTimestamp < strtotime('+30 days')) {
+                $lastActivityTimestamp = Carbon::parse($submission->getData('dateLastActivity'))->endOfDay();
+                $comparingTimestamp = Carbon::today()->endOfDay()->subDays(30);
+                if ($comparingTimestamp->gt($lastActivityTimestamp)) {
```
touhidurabir, line 120: This should be `-30 days` as we like to know if the last activity is older than 30 days .

jobs/email/EditorialReminder.php
```diff
@@ -92,30 +93,31 @@ public function handle(): void
                 $reviewRoundDao = DAORegistry::getDAO('ReviewRoundDAO');
                 $reviewRound = $reviewRoundDao->getLastReviewRoundBySubmissionId($submission->getId(), $submission->getData('stageId'));
 
-                if ($reviewRound->getStatus() === ReviewRound::REVIEW_ROUND_STATUS_PENDING_REVIEWERS) {
+                if ($reviewRound->determineStatus() === ReviewRound::REVIEW_ROUND_STATUS_PENDING_REVIEWERS) {
```
touhidurabir, line 96: using the `determineStatus` seems easier way rather than retrieving the last review assignment and making decisions based on that as the `determineStatus` does that internally .

Vitaliy-1: Shouldn't correct status be already determined? If we determine status and it's different from what is stored in the database, it looks wrong to me. What do you think?

classes/submission/reviewRound/ReviewRound.php
```diff
@@ -250,6 +250,7 @@ public function determineStatus()
                     break;
 
                 case ReviewAssignment::REVIEW_ASSIGNMENT_STATUS_AWAITING_RESPONSE:
+                case ReviewAssignment::REVIEW_ASSIGNMENT_STATUS_REQUEST_RESEND:
```
touhidurabir, line 253: If a review round have all reviews sorted out (submitted and confirmed) and then a new reviewer added who has declined and then again re-requested, The review round show status as `ReviewRound::REVIEW_ROUND_STATUS_REVIEWS_COMPLETED ==> 9` but it show as `ReviewRound::REVIEW_ROUND_STATUS_PENDING_REVIEWS ==> 7` as the review has not yet responded to re-request .

jobs/email/EditorialReminder.php
touhidurabir: > If we determine status and it's different from what is stored in the database, it looks wrong to me

 That's 100% right and I am still trying to figure out how such as case, but still unable to . However found one issue , see https://github.com/pkp/pkp-lib/pull/10332/files#r1766427232 . I am running bit more tests to see if I can reproduce a case where `getStatus` and `determineStatus` provide different result .

touhidurabir: see this https://github.com/pkp/pkp-lib/issues/10214#issuecomment-2372799562

jobs/email/EditorialReminder.php
```diff
@@ -158,8 +160,11 @@ public function handle(): void
             ->from($context->getContactEmail(), $context->getLocalizedName(Locale::getLocale()))
             ->recipients([$editor])
             ->subject($emailTemplate->getLocalizedData('subject'))
-            ->body($emailTemplate->getLocalizedData('body'))
-            ->allowUnsubscribe($notification);
+            ->body($emailTemplate->getLocalizedData('body'));
+        
+        if ($notification) {
+            $mailable->allowUnsubscribe($notification);
+        }
```
touhidurabir, line 167: `$notification` can be null , see https://github.com/pkp/pkp-lib/issues/10496

Vitaliy-1 (changes requested):
Thanks, @touhidurabir. My main concerns regarding the issue are:

1. The review round status in the database should be the only source of truth for such checks. Otherwise, do we need the column `status` at all? Is there a source problem where the status isn't updated accordingly?
2. The call to the `ReviewRound::determineStatus()` is expensive. If we don't know the cause of the issue and want to apply temporary fix, let's call it once.

touhidurabir:
> Thanks, @touhidurabir. My main concerns regarding the issue are:
> 
> 1. The review round status in the database should be the only source of truth for such checks. Otherwise, do we need the column `status` at all? Is there a source problem where the status isn't updated accordingly?
> 2. The call to the `ReviewRound::determineStatus()` is expensive. If we don't know the cause of the issue and want to apply temporary fix, let's call it once.

@Vitaliy-1 I agree . And I also do not plan to go through the approach of using the `ReviewRound::determineStatus()` anymore . I think I have got the core reason of this mismatch , please see comment at https://github.com/pkp/pkp-lib/issues/10214#issuecomment-2372799562 . By considering 2 more status case of review round , perhaps we can solve it without making call to `ReviewRound::determineStatus()` .

jobs/email/EditorialReminder.php
Vitaliy-1: Yeah, in this case I agree to leave it here. But can we perform calculations ones: 
```php
$status = $reviewRound->determineStatus();
if ($status === ReviewRound::REVIEW_ROUND_STATUS_PENDING_REVIEWERS) {
...
} 
```

Vitaliy-1 (changes requested):
Thanks! I left a couple of comments

jobs/email/EditorialReminder.php
touhidurabir: > as it's a job, we might need to check if the editor user exists and still has the editor role.

Right , that was missing and I think that missing check is causing the issue https://github.com/pkp/pkp-lib/issues/10496 . Also I have adjusted the checking to make it better performant .

bozana:
Hi @touhidurabir and @Vitaliy-1, I reviewed the changes as good as I could :-) and it all looks good to me.
I would however have a question:
What about these statuses:
REVIEW_ROUND_STATUS_RESUBMIT_FOR_REVIEW_SUBMITTED
REVIEW_ROUND_STATUS_RECOMMENDATIONS_READY
REVIEW_ROUND_STATUS_RECOMMENDATIONS_COMPLETED
eventually also REVIEW_ROUND_STATUS_RETURNED_TO_REVIEW
Should they not be considered in/for editor reminder?

Else, it can be merged...

Issue #10318, opened 2024-08-14
Deleting authenticated ORCIDs should revoke tokens


ewhanson:
**Describe the issue**
When a user authenticates their ORCID and links it to OJS, an access and refresh token are created and stored in the database. Currently, when an ORCID is deleted, the ORCID along with the tokens are deleted from the database, but they are not revoked from ORCID. Whenever an authenticated ORCID is removed from OJS, the corresponding tokens should be removed as well as described in the [ORCID documentaiton](https://info.orcid.org/ufaqs/how-can-i-revoke-tokens/#).

For the `FieldOrcid` Vue component, this revocation can be done as part of the OJS API call here: https://github.com/pkp/pkp-lib/blob/0fb17c21f20212ac944b5c15b19707a5e8fae83a/api/v1/orcid/OrcidController.php#L129

This will need to be handled separately for the user profile ORCID connection workflow.

**What application are you using?**
 OJS, OMP, and OPS version 3.5, `main` branch

**Additional information**
This will apply only to the `main` branch.

taslangraham:
Ready for review @ewhanson 

### Description
- Revoke ORCID tokens from ORCID when they are removed from a user’s profile or author details.
- Implemented a job to make HTTP call to the ORCID API, revoking a user's access tokens for the application.
- Revoking an ORCID linked to a user’s profile does not revoke tokens linked to their author details for individual submissions, and vice versa.

### PRs:
pkp lib - https://github.com/pkp/pkp-lib/pull/10333/files

**submodule updates**
- ojs: https://github.com/pkp/ojs/pull/4409/files
- ops: https://github.com/pkp/ops/pull/754/files
- omp: https://github.com/pkp/omp/pull/1678/files

ewhanson:
Thanks @taslangraham. Looks good! Just a few comments in the PR.

I also had a question about the placement of the delete button:

<img width="974" alt="Screenshot of user profile with ORCID" src="https://github.com/user-attachments/assets/953eb024-ba71-49d4-93cd-2e7aa1120931">

I'm not sure how easy it would be, but if possible, I think it would be better if it were in-line with the ORCID URL, same as the Vue component.

taslangraham:
@ewhanson thanks for the review. 

In response to your suggestion, 
> I'm not sure how easy it would be, but if possible, I think it would be better if it were in-line with the ORCID URL, same as the Vue component.

[...]

PR #10333, opened 2024-08-22
pkp/pkp-lib#10318 Revoke ORCID tokens when deleting authenticated ORCIDs

taslangraham:
For pkp/pkp-lib#10318

### DIFF
--- a/classes/orcid/OrcidManager.php
+++ b/classes/orcid/OrcidManager.php
@@ -20,6 +20,8 @@ use APP\facades\Repo;
 use PKP\config\Config;
 use PKP\context\Context;
 use PKP\core\Core;
+use pkp\jobs\orcid\RevokeOrcidToken;
+use PKP\user\User;
 
 class OrcidManager
 {
@@ -99,14 +101,7 @@ class OrcidManager
      */
     public static function getOrcidUrl(?Context $context = null): string
     {
-        if (self::isGloballyConfigured()) {
-            $apiType = Application::get()->getRequest()->getSite()->getData(self::API_TYPE);
-        } else {
-            if ($context === null) {
-                $context = Application::get()->getRequest()->getContext();
-            }
-            $apiType = $context->getData(self::API_TYPE);
-        }
+        $apiType = self::getApiType();
         return in_array($apiType, [self::API_PUBLIC_PRODUCTION, self::API_MEMBER_PRODUCTION]) ? self::ORCID_URL : self::ORCID_URL_SANDBOX;
     }
 
@@ -117,14 +112,7 @@ class OrcidManager
      */
     public static function getApiPath(?Context $context = null): string
     {
-        if (self::isGloballyConfigured()) {
-            $apiType = Application::get()->getRequest()->getSite()->getData(self::API_TYPE);
-        } else {
-            if ($context === null) {
-                $context = Application::get()->getRequest()->getContext();
-            }
-            $apiType = $context->getData(self::API_TYPE);
-        }
+        $apiType = self::getApiType();
 
         return match ($apiType) {
             self::API_PUBLIC_SANDBOX => self::ORCID_API_URL_PUBLIC_SANDBOX,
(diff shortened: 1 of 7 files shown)


api/v1/orcid/OrcidController.php
```diff
@@ -168,9 +169,12 @@ function (Role $role) {
             }
         }
 
+        $clonedAuthor = clone $author;
         $author->setOrcid(null);
         $author->setOrcidVerified(false);
+
         OrcidManager::removeOrcidAccessToken($author);
+        dispatch(new RevokeOrcidToken($context, $clonedAuthor));
```
ewhanson, line 177: You could dispatch the job within `OrcidManager::removeOrcidAccessToken()`, that way you wouldn't have to duplicate/remember to additionally dispatch the job and you wouldn't need to clone the author/user either.

classes/orcid/OrcidManager.php
```diff
@@ -363,4 +334,29 @@ private static function writeLog(string $message, string $level): void
         $logFilePath = Config::getVar('files', 'files_dir') . '/orcid.log';
         error_log("{$fineStamp} {$level} {$message}\n", 3, $logFilePath);
     }
+
+    /**
+     * Gets the ORCID API endpoint to revoke an access token
+     */
+    public static function getTokenRevocationUrl(?Context $context = null): string
```
ewhanson, line 341: `$context` isn't used so it can be removed from the function arguments.

classes/orcid/OrcidManager.php
```diff
@@ -363,4 +334,29 @@ private static function writeLog(string $message, string $level): void
...
+    {
+        return self::getOauthPath() . 'revoke';
+    }
+
+    /**
+     * Helper method get the ORCID Api Type.
+     */
+    private static function getApiType(?Context $context = null): string
```
ewhanson, line 349: Good idea to consolidate this!

classes/user/form/IdentityForm.php
```diff
@@ -121,9 +122,20 @@ public function execute(...$functionArgs)
...
-        $user->setPreferredPublicName($this->getData('preferredPublicName'), null);
+
+        // Request to delete ORCID token is handle separately from other form field updates
+        if($this->getData('removeOrcidId') === 'true') {
+            $clonedUser = clone $user;
+            $user->setOrcid(null);
+            $user->setOrcidVerified(false);
+
+            OrcidManager::removeOrcidAccessToken($user);
+            dispatch(new RevokeOrcidToken($request->getContext(), $clonedUser));
```
ewhanson, line 133: Same as comment above about moving the job dispatching into `OrcidManager::removeOrcidAccessToken()`.

ewhanson (changes requested):

api/v1/orcid/OrcidController.php
taslangraham: good suggestion, I've made the update.

classes/orcid/OrcidManager.php
taslangraham: updated

jobs/orcid/RevokeOrcidToken.php
```diff
@@ -0,0 +1,68 @@
...
+            $httpClient->request(
+                'POST',
+                OrcidManager::getTokenRevocationUrl(),
+                [
+                    'headers' => $headers,
+                    'form_params' => $postData,
+                ],
+            );
+
+            OrcidManager::logInfo('Token revoked for user, with ID: ' . $this->user->getId());
```
ewhanson, line 60: @taslangraham, everything else looks good, but I just noticed this. It could be either an author or a user, so it would be helpful to have the log reflect whichever type it is.

taslangraham: updated

jobs/orcid/RevokeOrcidToken.php
```diff
@@ -0,0 +1,70 @@
...
+                [
+                    'headers' => $headers,
+                    'form_params' => $postData,
+                ],
+            );
+
+            $identityTypeName = $this->identity instanceof User ? 'User' : 'Author';
+            OrcidManager::logInfo("Token revoked for {$identityTypeName}, with ID: " . $this->identity->getId());
+        } catch (ClientException $exception) {
+            $this->fail($exception);
```
ewhanson, line 64: Hi @taslangraham, one final thing I missed earlier. `$this->fail()` should be at the end of the `catch` block as it will return or throw an exception.

taslangraham: Good catch, I've updated it. Thanks!

ewhanson (approved):

PR #10335, opened 2024-08-26
pkp/pkp-lib#10263 edit metadata post publication

### LINKED ISSUE #10263: Relax editing metadata on published/posted materials
**Describe the bug**
Currently, when content is published/posted in OJS/OMP/OPS, editors are restricted from making further changes:
image

In practice, however, editors want the ability to tweak/edit metadata in published content. This has led to anti-patterns like unpublish/edit/republish cycles.

In OMP, there is an additional problem, where certain data on Publication Formats can't be accessed/reviewed because the edit links disappear once the content is published.

Accepting that editors are going to want to tweak metadata after publication, we should keep the warning message present, but stop restricting edits. We should trust the editors to understand the implications of making edits, but not try to prevent it.

Hafsa-Naeem:
fixes #10263

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -1311,11 +1311,6 @@ class PKPSubmissionController extends PKPBaseController
             ], Response::HTTP_FORBIDDEN);
         }
 
-        if ($publication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
-            return response()->json([
-                'error' => __('api.publication.403.alreadyPublished'),
-            ], Response::HTTP_FORBIDDEN);
-        }
 
         $submissionContext = $request->getContext();
         if (!$submissionContext || $submissionContext->getId() !== $submission->getData('contextId')) {
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -71,6 +71,8 @@ use PKP\submission\PKPSubmission;
 use PKP\submission\reviewAssignment\ReviewAssignment;
 use PKP\submissionFile\SubmissionFile;
 use PKP\userGroup\UserGroup;
+use PKP\observers\events\MetadataChanged;
+
 
 class PKPSubmissionController extends PKPBaseController
 {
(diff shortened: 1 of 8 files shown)


cypress/support/commands_new_workflow.js
```diff
@@ -921,7 +921,7 @@ Cypress.Commands.add('confirmEmail', user => {
 	// Current email server is sendra https://github.com/msztolcman/sendria
 	// Intentionally trying 127.0.0.1 instead of localhost as at least on mac localhost
 	// did not work - https://github.com/cypress-io/cypress/issues/26154#issuecomment-1755904773
-	let emailServer = 'http://127.0.0.1:1080/';
```
asmecher, line 924: Was this change needed for your local setup?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1237,13 +1239,6 @@ public function editPublication(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_FORBIDDEN);
         }
 
-        // Publications can not be edited when they are published
```
asmecher, line 1240: @Hafsa-Naeem, I haven't looked into this, but I'm worried that these checks also rightfully blocked authors from editing published content. That restriction should definitely remain, even if we're allowing editors to edit published content. Can you double-check to see if that was taken into account with these changes?

asmecher (commented):
Thanks, @Hafsa-Naeem, a couple more comments here.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1314,11 +1324,6 @@ public function publishPublication(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_FORBIDDEN);
         }
 
-        if ($publication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
```
asmecher, line 1317: This doesn't look right to me -- it should never be necessary to call `publishPublication` on already-published content.

api/v1/submissions/PKPSubmissionController.php
asmecher: @Hafsa-Naeem, looking into it further, authors are currently (before this change) blocked from editing publication metadata on published content because by default their `stage_assignment`'s `can_change_metadata` column gets set to `0` at the end of the author's submission process.

However, the editor can come in later and edit the author's assignment to permit metadata editing. Thereafter, without [your recent addition](https://github.com/pkp/pkp-lib/pull/10335/commits/1f8b7f79b75524cc310b8a70af2f19412636e0e6), authors would have been allowed to edit metadata on published content.

Now that you've added the latest commit, it's now fixed for `editPublication`. However, you also removed the `STATUS_PUBLISHED` check from other functions in this handler, and didn't check there for authors.

Rather than replicating this code through the class, and confusing the use of `can_change_metadata` with later checks in the PHP side, it might be better to cause the publish action to set the `can_change_metadata` column to `0` for all stage assignments. That way we can still support workflows where editors want to grant edit permission post-publication, but we won't accidentally weaken this check in the process.

locale/pl/submission.po
```diff
@@ -2735,6 +2735,7 @@ msgstr "Wersja {$wersja}"
 
 #~ msgid "publication.editDisabled"
 #~ msgstr ""
+
```
asmecher, line 2738: (This change should be removed.)

asmecher (changes requested):

Issue #10306, opened 2024-08-13
Add queue job unit tests for 3.5


touhidurabir:
This is a continuation of issue https://github.com/pkp/pkp-lib/issues/9899 for 3.5 

**Job classes to have unit tests**

##### OJS
- [x] APP\jobs\doi\DepositIssue
- [x] APP\jobs\notifications\IssuePublishedNotifyUsers
- [x] APP\jobs\notifications\OpenAccessMailUsers
- [ ] APP\jobs\orcid\DepositOrcidReview
- [x] APP\jobs\statistics\CompileCounterSubmissionDailyMetrics
- [x] APP\jobs\statistics\CompileCounterSubmissionInstitutionDailyMetrics
- [x] APP\jobs\statistics\CompileIssueMetrics
- [x] APP\jobs\statistics\CompileSubmissionGeoDailyMetrics
- [x] APP\jobs\statistics\CompileUniqueInvestigations
- [x] APP\jobs\statistics\CompileUniqueRequests
- [ ] APP\jobs\statistics\CompileUsageStatsFromTemporaryRecords [deprecated 3.4.0.5]
- [x] APP\jobs\statistics\DeleteUsageStatsTemporaryRecords
- [x] APP\jobs\statistics\ProcessUsageStatsLogFile
##### OMP
- [x] APP\jobs\statistics\CompileCounterSubmissionDailyMetrics
- [x] APP\jobs\statistics\CompileCounterSubmissionInstitutionDailyMetrics
- [x] APP\jobs\statistics\CompileSeriesMetrics
- [x] APP\jobs\statistics\CompileSubmissionGeoDailyMetrics
- [x] APP\jobs\statistics\CompileUniqueInvestigations
- [x] APP\jobs\statistics\CompileUniqueRequests
- [ ] APP\jobs\statistics\CompileUsageStatsFromTemporaryRecords [deprecated 3.4.0.5]
- [x] APP\jobs\statistics\DeleteUsageStatsTemporaryRecords
- [x] APP\jobs\statistics\ProcessUsageStatsLogFile
##### OPS
- [x] APP\jobs\statistics\CompileCounterSubmissionDailyMetrics
- [x] APP\jobs\statistics\CompileCounterSubmissionInstitutionDailyMetrics
- [x] APP\jobs\statistics\CompileSubmissionGeoDailyMetrics
- [x] APP\jobs\statistics\CompileUniqueInvestigations
[...]

Issue #10328, opened 2024-08-20
Refactor announcements as an Eloquent Model


Vitaliy-1:
Announcement is an entity that utilises settings table. This makes its refactoring to the Eloquent model a non-trivial task.
As a part of this issue
- [x] Adapt Eloquent model to work with data from a settings table as usual properties/attributes
- [x] Adapt Eloquent to work with localized properties
- [x] Refactor Announcement entity

This will serve as an example for refactoring of other entities, such as: [genres](https://github.com/pkp/pkp-lib/issues/10133) and [controlled vocabulary](https://github.com/pkp/pkp-lib/issues/10292).

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/10382
OJS: https://github.com/pkp/ojs/pull/4433

Vitaliy-1:
@touhidurabir, can you review this? 
@Hafsa-Naeem, I see that you are also working on refactoring of entity which deals with settings table. Feel free to make comments if you'd like.

touhidurabir:
@Vitaliy-1 I have added reviews at https://github.com/pkp/pkp-lib/pull/10382#pullrequestreview-2311493712

Issue #9566, opened 2023-12-08
Convert session and cookie management to Laravel


asmecher:
**Describe the bug**
In preparation for work on OAuth/SSO, convert the underlying aspects of OJS/OMP/OPS from bespoke implementations to Laravel's toolset.

It is unclear how much of this can be done without needing to convert major aspects of the stack, e.g. authorization and request handling; this work should stop short of a major intervention for the moment.

Opportunities:
- Replace `SessionManager` with `SessionServiceProvider` / `session()`
- Replace `Session`/`SessionDAO` with Eloquent-based implementation (shipped with Laravel, I think)
- Replace cookie implementation with `CookieManager`
- Add a `UserProvider` implementation serving users to Laravel from current `EntityDAO` user implementation
- Move CSRF token generation to Laravel (replacing homebrew implementation)

Pitfalls: 
- May need to tangle with `lib/vendor/laravel/framework/src/Illuminate/Session/Middleware/*` to get sessions to work properly. Not sure how deep that rabbit-hole goes yet.

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9596
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4132
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/646
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1531
[orcidProfile](https://github.com/pkp/orcidProfile) --> https://github.com/pkp/orcidProfile/pull/313

asmecher:
I've experimented with this but think the next step is to work with the session middleware. Calling in @touhidurabir for a look!

Draft PRs (very much WIP):
- https://github.com/pkp/pkp-lib/pull/9567
- https://github.com/pkp/ojs/pull/4121

To even start getting these working it's necessary to replace the `sessions` table with the new schema by dropping the current table and running the new migration (`lib/pkp/classes/migration/install/SessionsMigration.php`).

touhidurabir:
@asmecher Draft PR
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/9596
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4132

[...]

PR #10340, opened 2024-08-27
pkp/pkp-lib#10306 unit tests for queue jobs

touhidurabir:
for 10306


classes/core/PKPApplication.php
```diff
@@ -319,6 +319,13 @@ public function getUUID(): string
      */
     public function getHttpClient()
     {
+        if (PKPContainer::getInstance()->runningUnitTests()) {
+            $client = Registry::get(\PKP\tests\PKPTestCase::MOCKED_GUZZLE_CLIENT_NAME);
+            if ($client) {
+                return $client;
+            }
+        }
+
```
touhidurabir, line 328: part of functionality to mocking ability for guzzel reqeust .

classes/notification/NotificationManagerDelegate.php
```diff
@@ -56,7 +56,7 @@ public function updateNotification(PKPRequest $request, ?array $userIds, int $as
      *
      * @copydoc PKPNotificationOperationManager::createNotification()
      */
-    public function createNotification(PKPRequest $request, ?int $userId = null, ?int $notificationType = null, ?int $contextId = Application::SITE_CONTEXT_ID, ?int $assocType = null, ?int $assocId = null, int $level = Notification::NOTIFICATION_LEVEL_NORMAL, ?array $params = null): ?Notification
+    public function createNotification(?PKPRequest $request = null, ?int $userId = null, ?int $notificationType = null, ?int $contextId = Application::SITE_CONTEXT_ID, ?int $assocType = null, ?int $assocId = null, int $level = Notification::NOTIFICATION_LEVEL_NORMAL, ?array $params = null): ?Notification
```
touhidurabir, line 59: do we have any use of `PKPRequest $request` here as it seems this data not used ?

tests/PKPTestCase.php
```diff
@@ -205,8 +220,7 @@ protected function mockRequest($path = 'index/test-page/test-op', $userId = null
         $request->setRouter($router);
 
         // Test user.
-        $session = $request->getSession();
-        $session->setUserId($userId);
+        $request->getSessionGuard()->setUserId($userId);
```
touhidurabir, line 223: a leftover missing change from https://github.com/pkp/pkp-lib/issues/9566

tests/jobs/bulk/BulkEmailSenderTest.php
```diff
@@ -0,0 +1,77 @@
...
+            ->shouldReceive('getCollector')
+            ->withAnyArgs()
+            ->andReturn($userCollectorMock)
+            ->getMock();
+        
+        app()->instance(UserRepository::class, $userRepoMock);
+
+        $bulkEmailSenderJob->handle();
+
+        $this->expectNotToPerformAssertions();
```
touhidurabir, line 75: Our previous asserting was like `$this->assertNull($bulkEmailSenderJob->handle())` . However the `handle` method return type set as `void` which does turn in null and older PHPUnit version did not have a arg type defined . But the latest PHPUnit version has arg type define as `mixed` which in turn identified as error as return type `void` does not match with arg time `mixed` . Though this does not cause exception but modern IDEs identify it as error . So probably better to go with approach

classes/core/PKPContainer.php
```diff
@@ -590,6 +586,31 @@ public function environment(string ...$environments): string|bool
 
         return $this->get('config')['app']['env'];
     }
+
+    /**
+     * Override Laravel method; always false.
+     * Prevents the undefined method error when the Log Manager tries to determine the driver
+     */
+    public function runningUnitTests(): bool
```
asmecher, line 594: This could be removed once we add and configure the Laravel logging service provider, correct? I think this is related to what you're talking about with Erik. It might be worth filing in Github, noting this function in the issue, and adding a FIXME here.

classes/notification/NotificationManagerDelegate.php
asmecher: You're right -- this parameter can be removed.

tests/PKPTestCase.php
```diff
@@ -238,6 +253,49 @@ protected function localeToRegExp(string $translation): string
         $escapedPieces = array_map(fn ($piece) => preg_quote($piece, '/'), $pieces);
         return '/^' . implode('.*?', $escapedPieces) . '$/u';
     }
+
+    /**
+     * Mock the mail facade
+     * @see https://laravel.com/docs/10.x/mocking
```
asmecher, line 259: Better to link to 11, since that's what we're shipping.

tests/jobs/doi/DepositSubmissionTest.php
```diff
@@ -0,0 +1,176 @@
...
...nal for consideration.</li><li>All references have been checked for accuracy and completeness.</li><li>All tables and figures have been numbered and labeled.</li><li>Permission has been obtained to publish all photos, datasets and other material provided with this submission.</li></ul>";s:5:"fr_CA";s:37:"##default.contextSettings.checklist##";}s:15:"uploadFilesHelp";a:2:{s:2:"en";s:249:"<p>Provide any files our editorial team may need to evaluate your submission. In addition to the main work, you may wish to submit data sets, conflict of interest statements, or other supplementary files if these will be helpful for our editors.</p>";s:5:"fr_CA";s:39:"##default.submission.step.uploadFiles##";}}s:20:"_hasLoadableAdapters";b:0;s:27:"_metadataExtractionAdapters";a:0:{}s:25:"_extractionAdaptersLoaded";b:0;s:26:"_metadataInjectionAdapters";a:0:{}s:24:"_injectionAdaptersLoaded";b:0;}s:9:"\0*\0agency";O:43:"APP\\plugins\\generic\\datacite\\DatacitePlugin":4:{s:10:"pluginPath";s:24:"plugins/generic/datacite";s:14:"pluginCategory";s:7:"generic";s:7:"request";N;s:58:"\0APP\\plugins\\generic\\datacite\\DatacitePlugin\0_exportPlugin";N;}s:10:"connection";s:8:"database";s:5:"queue";s:5:"queue";}
```
asmecher, line 37: This task in particular has quite a bit of localized content, including some `##untranslated.locale.keys##` (not sure why that's the case -- `default.contextSettings.checklist` is showing untranslated for `fr_CA`, for example, but it does appear in the locale files for OJS, OMP, and OPS). How do you think we're most likely to maintain and regenerate these as needed? If manually, it might be worth trimming out the extraneous localized text into something more minimal.

tests/jobs/notifications/NewAnnouncementNotifyUsersTest.php
```diff
@@ -0,0 +1,128 @@
...
+    public function testRunSerializedJob(): void
+    {
+        $this->mockMail();
+        
+        $this->mockRequest();
+
+        /** @var NewAnnouncementNotifyUsers $newAnnouncementNotifyUsersJob */
+        $newAnnouncementNotifyUsersJob = unserialize($this->serializedJobData);
+
+        $announcementMock = Mockery::mock(\PKP\announcement\Announcement::class)
```
asmecher, line 60: This is going to require some work once https://github.com/pkp/pkp-lib/issues/10328 is merged.

tests/jobs/notifications/NewAnnouncementNotifyUsersTest.php
```diff
@@ -0,0 +1,128 @@
...
+            ->getMock();
+
+        $contextDaoMock = Mockery::mock($contextDaoClass)
+            ->makePartial()
+            ->shouldReceive('getById')
+            ->withAnyArgs()
+            ->andReturn($contextMock)
+            ->getMock();
+
+        DAORegistry::registerDAO(substr(strrchr($contextDaoClass, '\\'), 1), $contextDaoMock);
```
asmecher, line 95: I think I'd prefer a match statement on e.g. `Application::getName` rather than string parsing and assembly. At some point we'll get rid of `JournalDAO`, for example, and being able to grep for it will be helpful.

asmecher (changes requested):
Just a few minor suggestions -- thanks, @touhidurabir!

classes/core/PKPContainer.php
touhidurabir: Not exactly , this has there before this PR and probably will stay . Our Container class extends the laravel core container but We have our very own Application class . Once we able to handle the bootstrapping process to Laravel's foundation Application, we can remove this (as it's reside there) . As few core class required this information, so we have copied it from Laravel's foundation application class to our PKPContainer just to facilitate the dependency.

tests/jobs/doi/DepositSubmissionTest.php
touhidurabir: probably we do not need to worry much . First of all, we need set some rules when writing tests for `handle` method of a job . I believe we need to make sure when testing job handle, none of the calls make to real object but mock object only e.g. we need to make sure we have proper mocks for each of the call to classes in `handle` method . 

That saying we can trim down the localized content but do we need to ?

asmecher: Let's not resolve a problem until it becomes one :)

Issue #10342, opened 2024-08-28
Allow except or/and depending middleware requirement for routing middleware


touhidurabir:
**Describe the problem**
Some middleware may require the presence of one or more other middleware to run/function properly . One such example is the `DecodeApiTokenWithValidation` and `ValidateCsrfToken` middleware that require `HasUser` middleware . 

Another case is some middleware may new pass some routes to pass without verifying it's own passing logic (what requirements defined in the middleware `handle` function) .

**What problem it will solve**
One example is the invitation API where when accepting an invitation , there will/can be no API KEY or user associated with the request . This scenario make it impossible to access api routes because of global middleware such as `DecodeApiTokenWithValidation` and `ValidateCsrfToken` look for API KEY or logged in user session . 

**What application are you using?**
 OJS, OMP or OPS version `main` (upcoming `3.5.0`)

**Additional information**
see https://github.com/pkp/pkp-lib/issues/10311 and possible proposed solution at https://github.com/pkp/pkp-lib/issues/10311#issuecomment-2309574338

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10343
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4412 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1686 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/759 [TEST ONLY]

touhidurabir:
Merged the `pkp-lib` PR at https://github.com/pkp/pkp-lib/pull/10343, will let the submodule get updated later manually .

defstat:
@touhidurabir those changes should effect already existing endpoints that retrieve the user from the bearer authorization token? Should, for example, endpoints that have `$controller->addPolicy(new UserRolesRequiredPolicy($request), true);` defined in the `authorize` function be affected by this change?

touhidurabir:
> should those changes affect the existing endpoints that retrieve the user from the bearer authorization token?

It should . What is does is that if a route does not have `HasUser` middleware attached, it will bypass the global middleware `DecodeApiTokenWithValidation` and `ValidateCsrfToken`. 

> will endpoints with `$controller->addPolicy(new UserRolesRequiredPolicy($request), true);` defined in the `authorize` function be affected by this change?

[...]

PR #10343, opened 2024-08-28
pkp/pkp-lib#10342 Required depending middleware attachment check

touhidurabir:
for pkp/pkp-lib#10342


asmecher (approved):
Thanks, @touhidurabir, no objections here!

PR #10352, opened 2024-08-29
Override Smarty date formatter to support multilingualism

asmecher:


defstat:
@asmecher thanks!

I have just one comment: There’s a particular code line that might be relevant here. The line is [EditorAction.php line 199](https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/submission/action/EditorAction.php#L199), and I have a few observations:

1. It seems that this event insertion never actually happens. The `$logEntry` parameter is `false` by default, and I couldn’t find any `setDueDates` references that call this with `$logEntry` set to true.
2. This is necessary for the message `log.review.reviewDueDateSet` to appear in the event viewer grid. However, there’s a typo in the translation key: it expects the parameter `{$dueDate}` instead of `{$reviewDueDate}`. As a result, the message currently displays as `The due date for the round 1 review of submission 12 by Adela Gallego has been set to **{$dueDate}**.`
3. Even if the message displayed correctly, the translation would occur when the event is created. Therefore, any later changes to the locale wouldn’t be reflected.

Given all of the above, I’m not sure if we need to make any date-related changes. What do you think?

defstat (commented):
@asmecher thanks!

I have just one comment: There’s a particular code line that might be relevant here. The line is [EditorAction.php line 199](https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/submission/action/EditorAction.php#L199), and I have a few observations:

It seems that this event insertion never actually happens. The $logEntry parameter is false by default, and I couldn’t find any setDueDates references that call this with $logEntry set to true.
This is necessary for the message log.review.reviewDueDateSet to appear in the event viewer grid. However, there’s a typo in the translation key: it expects the parameter {$dueDate} instead of {$reviewDueDate}. As a result, the message currently displays as The due date for the round 1 review of submission 12 by Adela Gallego has been set to **{$dueDate}**.
Even if the message displayed correctly, the translation would occur when the event is created. Therefore, any later changes to the locale wouldn’t be reflected.
Given all of the above, I’m not sure if we need to make any date-related changes. What do you think?

asmecher:
@defstat, hmm, you're right -- that looks like dead code. I'm going to remove it from the `main` branch but will leave it here, as it's harmless. Thanks!

bozana:
Hi @asmecher, one question because I am curious (because of our changes to the Weblate locales): What locales one can use here for? Do you maybe know how does Carbon deal with locales, for example when a locale together with the country and script is used -- what locales does it understands and what if it does not understand a locale -- for example if it then only uses the language part of the locale? -- I can also try to figure it out, but maybe you already know...
(In the Carbon Github I see a list of languages that is pretty similar to Weblate -- e.g. there are _Latn and _Cyril (instead of @cyrillic) and it looks comprehensive... but... what about the cases that differ... )

asmecher:
@bozana, Carbon ships with a lot of languages included (rather than rely on the libc locale set); there's [a large list of included languages](https://carbon.nesbot.com/docs/#supported-locales), plus the [translation tools they use](https://carbon.nesbot.com/contribute/translate/) appear to rely on ISO639-1, including e.g. the `_Latn` suffix.

Issue #10357, opened 2024-09-02
Site level mailables can't be accessed with ContextId = null


defstat:
**Describe the bug**
A recent change related to the DB foreign keys changed the code that was using `0` as the "placeholder" for the site-related contextId. Now this is set to `null`. 

This affected the Mailables retrieval. All the related functions `getByKey`, `getCollector`, etc of the `emailTemplate` Repository are using `int $contextId` as a mandatory, not nullable parameter. In case that we may need the site level mailable (like the registration or login processes) the code returns a fatal error, because it tries to pass `null` in the parameter that this is not allowed.   

**To Reproduce**
1. Try to register a new user with the `[email] => require_validation = On` switch in the configuration file

**What application are you using?**
 OJS, OMP or OPS main branch
 
 ---
 **PRs**
 
 PKP-LIB: #10358

defstat:
@Vitaliy-1, I have made a PR for that. This one allows a null contextId in the existing repository. I'm not sure if we want to go in that direction or use a different logic for site-level mailables to avoid mixing them with the existing context-related templates. However, this seems to work nonetheless.

PR #10358, opened 2024-09-02
[pkp-lib][main] #10357 Allow null contextId for emailTemplate Repository

defstat:


classes/emailTemplate/Collector.php
```diff
@@ -162,16 +162,19 @@ protected function getDefaultQueryBuilder(): Builder
...
-            ->selectRaw('NULL as alternate_to')
-
-            ->whereNotIn('etddata.email_key', function (Builder $q) {
-                $q->select('et.email_key')
-                    ->from('email_templates as et')
-                    ->where('et.context_id', $this->contextId);
-            })
-
-            ->when(!is_null($this->keys), function (Builder $q) {
+            ->selectRaw($this->contextId !== null ? $this->contextId . ' as context_id' : 'NULL as context_id')
```
asmecher, line 165: I'd rather avoid string concatenation of queries as much as possible; Jonas had an approach that we should also follow for consistency. See: https://github.com/pkp/pkp-lib/blob/1b6f1b50d7da36d589203c17bb18360cc74604fa/classes/userGroup/Collector.php#L302

We risk confusing `null` (constraint not specified to the collector) with `SITE_CONTEXT_ID` (used to be `0`, now `null`). In cases like this it's better to have the caller provide a list of context IDs, with the `SITE_ID_CONTEXT` constant being one of the available options. Then, per the link above, map the array to `intval` when the query is being built and use `COALESCE(context_id, 0)` in the SQL in order to map `NULL`s to `0` there.

The idea is that calling code can use:

```
$collector->filterByContextIds([$journal->getId()]); // Filter by a single journal ID -- note the array syntax
```
...or...
```
$collector->filterByContextIds(null); // Do not filter by context IDs -- clear any pre-existing filter
```
...or...
```
$collector->filterByContextIds([SITE_ID_CONTEXT]); // Filter for templates that are site-wide -- note the array syntax
```
The need to specify the context ID in the constructor in this class is non-standard. I'd suggest not requiring any filters in the constructor, and instead using fluent functions to chain constraints like we do elsewhere.

asmecher (changes requested):
Added a comment with some suggestions to standardize this against other similar cases!

Vitaliy-1 (approved):
Looks good! Don't have anything to add here, thanks

classes/emailTemplate/Collector.php
defstat: @asmecher, thank you for the feedback. I understand the importance of avoiding query string concatenation and aligning with Jonas’s approach for consistency. Here’s my take on handling `contextId` in email templates:

I believe that when we’re dealing with `emailTemplates`, we are specifically referring to context-level email templates. The user, as far as I know, cannot modify the default email template at the Site level. This means that the `email_templates` table shouldn't typically contain rows with `null` in the `context_id` column.

Given this, we seem to have two primary cases to consider:

1. **When `contextId` is provided:** In this case, we would join the `email_templates` table within the `Collector`'s query to retrieve data specific to that context.
   
2. **When `contextId` is not provided:** Here, we would use `Application::SITE_CONTEXT_ID` as the `contextId`. In this scenario, we retrieve default data from `email_templates_default_data`, bypassing any need for `contextId` in the `Collector`'s query.

The root of this issue lies in the change of `Application::SITE_CONTEXT_ID` from `0` to `null` to align with updated database standards and ensure consistency with other cases. This shift led to back-office issues, as `contextId` was previously non-nullable in both `Repo` and `Collector`. Consequently, we needed to adjust the code to handle `null` values correctly. 

However, while we could potentially introduce a specific `Repo` function to handle Site-level email templates, I believe that would be more appropriate as part of a broader future task rather than within the scope of this issue.

asmecher: I backed off my suggestion here :)

Issue #10373, opened 2024-09-06
"Enroll existing user" as reviewer no longer finds users outside the current journal


asmecher:
**Describe the bug**
As an Editor, when attempting to enroll an existing user as a reviewer, it's no longer possible to find users who are not active in the current journal. This used to be possible in OJS 3.3.0-x.

**To Reproduce**
Steps to reproduce the behavior, starting from the test dataset:
1. As an administrator, create a second journal
2. Register a new account in the second journal as a reviewer
3. Log into Journal Manager in the Journal of Public Knowledge (first journal)
4. Go to the review workflow for a submission
5. Click "Add Reviewer"
6. Click "Enroll existing user"
7. Attempt to search for the new account; it will not be findable.

**What application are you using?**
OJS 3.4.0

**Additional information**
Please add any screenshots, logs or other information we can use to investigate this bug report.

asmecher:
PRs:
 - pkp-lib stable-3_4_0: https://github.com/pkp/pkp-lib/pull/10374
 - ojs stable-3_4_0 (tests only): https://github.com/pkp/ojs/pull/4429

(Needs forward-porting to `main` after review)

asmecher:
@defstat, could you review the PR above?

asmecher:
Also forward-ported to main. Merged!

PR #10374, opened 2024-09-06
pkp/pkp-lib#10373 Fix enroll existing user as reviewer feature

asmecher:


defstat (approved):
@asmecher thanks, I have no comments.

PR #10380, opened 2024-09-09
pkp/pkp-lib#5504 Control access to settings by user group

asmecher:


bozana (commented):
Hi @asmecher, just one comment, I think.
Maybe also a question, to double check, if we should consider CanAccessSettingsPolicy in: 
OJS SubscriptionsGridHandler
LanguageGridHandler
SubmissionLanguageGridHandler

What about other API Controllers, e.g. institution, highlights, emailTemplates, invitations?

(And, nothing for this PR -- I see the other/similar options are handled in the same way -- but I am wondering if some login from the UserGroupFormHandler.js should actually go somewhere else... :thinking: )

Thanks a lot!

Issue #10163, opened 2024-07-04
Add support for catching registrants to removed hooks


asmecher:
**Describe the bug**
If a hook is removed from OJS/OMP/OPS, it will no longer be called, but hook registrants may still be registered to it. This will result in hard-to-debug situations when forward-porting plugins, where plugin callbacks will never be called.

It would be helpful to developers if attempts to register against hooks that are no longer called can result in errors.

PR #10382, opened 2024-09-10
pkp/pkp-lib#10328 Refactor Announcements

Vitaliy-1:


classes/announcement/Announcement.php
```diff
@@ -1,345 +1,430 @@
...
      */
-    public function getLocalizedTitleFull()
+    public function delete()
     {
-        $typeName = $this->getAnnouncementTypeName();
-        if (!empty($typeName)) {
-            return $typeName . ': ' . $this->getLocalizedTitle();
-        } else {
-            return $this->getLocalizedTitle();
+        Hook::call('Announcement::delete::before', [$this]);
```
asmecher, line 100: We use to have two relevant hooks here: `Announcement::delete::before` (still here, but will require callers to change from expecting a `DataObject` to expecting an Eloquent model) and `Announcement::delete' (now removed).

Big picture: I'd suggest we rely on Eloquent lifecycle events for entities that use Eloquent. This will reduce the ambiguity over whether to use these vs. hooks, and will standardize us towards Laravel more generally.

Smaller picture: Notwithstanding the above, when we change hook conventions...
- If a hook is removed, make sure to add a call to `Hook::addUnsupportedHooks` e.g. in `PKPApplication` so that registrants are prevented from trying to use it. (See [#10163](https://github.com/pkp/pkp-lib/issues/10163))
- If a hook is kept but substantially changed, consider [adopting the newer [`Hook::run` conventions](https://github.com/pkp/pkp-lib/discussions/8089) since we're already introducing a breaking change, and noting it in the list of changes for the [3.5 release notebook](https://github.com/pkp/pkp-lib/issues/9276). In particular, `Hook::run` allows for us to use typehinting and named parameters instead of the older clumsy array-based parameter passing.

pages/announcement/AnnouncementHandler.php
```diff
@@ -78,13 +72,13 @@ public function view($args, $request)
...
+        $announcement = Announcement::find($announcementId);
         if (
             $announcement
-            && $announcement->getAssocType() == Application::getContextAssocType()
-            && $announcement->getAssocId() == $request->getContext()?->getId()
+            && $announcement->getAttribute('assocType') == Application::getContextAssocType()
+            && $announcement->getAttribute('assocId') == $request->getContext()?->getId()
             && (
-                $announcement->getDateExpire() == null || strtotime($announcement->getDateExpire()) > time()
+                $announcement->getAttribute('dateExpire') == null || strtotime($announcement->getAttribute('dateExpire')) > time()
```
asmecher, line 81: It should be possible to use [Eloquent date casts](https://laravel.com/docs/11.x/eloquent-mutators#date-casting) to get columns to come back automagically as Carbon objects; it shouldn't be necessary/desirable to use `strtotime` anymore.

schemas/announcement.json
```diff
@@ -8,13 +8,15 @@
 	"properties": {
 		"_href": {
 			"type": "string",
+			"origin": "composed",
```
asmecher, line 11: :+1: Like this!

templates/frontend/objects/announcement_full.tpl
```diff
@@ -13,23 +13,23 @@
...
-		{$announcement->getDatePosted()|date_format:$dateFormatShort}
+		{$announcement->getAttribute('datePosted')|date_format:$dateFormatShort}
 	</div>
-	{if $announcement->getImage()}
+	{if $announcement->getData('image')}
 		<img
 			class="obj_announcement_full_image"
-			src="{$announcement->getImageUrl()}"
-			alt="{$announcement->getImageAltText()}"
+			src="{$announcement->getAttribute('imageUrl')}"
```
asmecher, line 24: Is it possible to just access `$announcment->imageUrl` rather than using `getAttribute`? That seems to be the more common Eloquent formulation, and what we've used elsewhere. (Is it necessary for the composed attributes or something?)

Vitaliy-1: It's possible. I simple don't like getting those magically. But that's what Laravel uses in their examples and don't mind adopting it.

classes/core/SettingsBuilder.php
```diff
@@ -0,0 +1,341 @@
...
+
+        $newQuery = clone $this->query;
+
+        if ($primaryValues->isNotEmpty()) {
+            $count = parent::update($primaryValues->toArray());
+        }
+
+        // TODO Eloquent transforms attributes to snake case, find and override instead of transforming here
+        $settingValues = $settingValues->mapWithKeys(
+            fn (mixed $value, string $key) => [Str::camel($key) => $value]
```
Vitaliy-1, line 72: It seams there is additional conversion of attributes between camel case and snake case happening somewhere in the Eloquent core. Not critical, can be handled later

classes/core/SettingsBuilder.php
```diff
@@ -0,0 +1,341 @@
...
+        $primaryKey = $this->model->getKeyName();
+
+
+        $sql = $this->buildUpdateSql($settingValues, $us, $newQuery);
+
+        // Build a query for update
+        $settingCount = DB::table($us)->whereIn($us . '.' . $primaryKey, $newQuery->select($primaryKey))
+            ->update([$us . '.setting_value' => DB::raw($sql)]);
+
+        return $count ? $count + $settingCount : $settingCount; // TODO Return the count of updated setting rows?
```
Vitaliy-1, line 85: I don't know what is better to return here since we updating to tables at once?

classes/core/traits/ModelWithSettings.php
```diff
@@ -0,0 +1,164 @@
...
+        if (!$multilingualProp) {
+            throw new Exception('Attribute ' . $data . ' doesn\'t exist in the ' . static::class . ' model');
+        }
+
+        if (!in_array($data, $this->multilingualProps)) {
+            throw new Exception('Trying to retrieve localized data from a non-multilingual attribute ' . $data);
+        }
+
+        // TODO What should the default behaviour be if localized value doesn't exist?
+        return $multilingualProp[$locale] ?? null;
```
Vitaliy-1, line 120: Think if multilingual value doesn't exists it's better to just return null

classes/services/PKPContextService.php
```diff
@@ -627,11 +628,8 @@ public function delete($context)
         $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
         $genreDao->deleteByContextId($context->getId());
 
-        Repo::announcement()->deleteMany(
-            Repo::announcement()
-                ->getCollector()
-                ->filterByContextIds([$context->getId()])
-        );
+        // TODO is it OK to delete without listening Model's delete-associated events (not loading each Model)?
+        Announcement::withContextIds([$context->getId])->delete();
```
Vitaliy-1, line 632: Deleting without models instantiating, it's much more performant

pages/announcement/AnnouncementHandler.php
touhidurabir: The case here is bit different here than the usual . we can have a cast define but as our types are coming from schema which will over write it in the `ModelWithSettings` trait's `convertSchemaToCasts` method where we are calling the `mergeCasts` which merge the provided casts with pre defined ones . I think if we want to apply some casts like 
```php
/**
 * Get the attributes that should be cast.
 *
 * @return array<string, string>
 */
protected function casts(): array
{
    return [
        'dateExpire' => 'datetime',
        'datePosted' => 'datetime',
    ];
}
```
we need to override the `Illuminate\Database\Eloquent\Concerns\HasAttributes::mergeCasts` to defined how it will handle and determine the precedence of provided casts over defined ones .

touhidurabir (commented):
@Vitaliy-1 added reviews . These are more like thoughts. suggestions and questions than reviews . Feel free to adopt if you feel like those fit the use cases . 

BTW, I must say, this is very promising and will help us adopting eloquent more generally.

classes/core/SettingsBuilder.php
Hafsa-Naeem: @Vitaliy-1  Instead of returning a single summed count, could we return an associative array with separate counts for the primary table and the settings table? 
we can also use database transaction ('DB::transaction') to ensure that both updates either succeed or fail together to maintain data integrity.

pages/announcement/AnnouncementHandler.php
touhidurabir: Thinking about this a bit more, I think we should override the method `mergeCasts` inside the trait `ModelWithSettings` to allow model defined casts over write any casts coming form schema if schema available like 
```php
/**
     * Merge new casts with existing casts on the model.
     *
     * @param  array  $casts
     * @return $this
     */
    public function mergeCasts($casts)
    {
        $casts = $this->ensureCastsAreStringValues($casts);
        dump($this->casts);
        $this->casts = array_merge($casts, $this->casts);

        return $this;
    }
```
basically we are just altering how the array_merge getting done . 

As trait `ModelWithSettings` will defined directly in entity model classes, it's `mergeCasts` should have higher precedence of `HasAttributes::mergeCasts` used in `Illuminate\Database\Eloquent\Model` .

classes/announcement/Announcement.php
```diff
@@ -1,345 +1,428 @@
...
+     * The attributes that are mass assignable.
      *
-     * @param int $assocType
+     * @var array
      */
-    public function setAssocType($assocType)
-    {
-        $this->setData('assocType', $assocType);
-    }
+    protected $fillable = ['assocType', 'assocId', 'typeId', 'title', 'image', 'description', 'descriptionShort'];
```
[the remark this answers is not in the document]
touhidurabir: Given that few entities don't have schema, probably best to allow the generation of `fillable` not depending on the `schema` . That is something like 
```php
public function __construct(array $attributes = [])
    {
        parent::__construct($attributes);

        if (static::getSchemaName()) {
            $this->setSchemaData();
        }

        $this->constructFillable();
    }
```
```php
protected function constructFillable(): void
    {
        $primaryColumns = [];

        if (static::getSchemaName()) {
            $schemaService = app()->get('schema'); /** @var PKPSchemaService $schemaService */
            $primaryColumns = $schemaService->groupPropsByOrigin(static::getSchemaName())[Schema::ATTRIBUTE_ORIGIN_MAIN];
        }
        
        $this->fillable = array_merge($this->fillable, $primaryColumns, $this->settings);
    }
```

api/v1/announcements/PKPAnnouncementController.php
```diff
@@ -209,27 +205,22 @@ public function add(Request $illuminateRequest): JsonResponse
...
             ], Response::HTTP_BAD_REQUEST);
         }
 
-        $announcement = Repo::announcement()->get($announcementId);
-
         $sendEmail = (bool) filter_var($params['sendEmail'], FILTER_VALIDATE_BOOLEAN);
 
         if ($context) {
-            $this->notifyUsers($request, $context, $announcementId, $sendEmail);
+            $this->notifyUsers($request, $context, $announcement->getKey(), $sendEmail);
```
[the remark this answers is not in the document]
Vitaliy-1: OK, let's try. There is a possibility that it might interact with schema's id attribute, I don't see anything right now but `id` is a primary attribute and the logic I've put to the new schema setting `origin` defines Mutators and `composed` attributes. Meaning in `announcement.json`:
```json
"id": {
	"type": "integer",
	"origin": "primary",
	"readOnly": true,
	"apiSummary": true
},
```

classes/announcement/Announcement.php
```diff
@@ -1,345 +1,428 @@
...
-    public function getDateExpire()
+    protected function imageUrl(bool $withTimestamp = true): Attribute
     {
-        return $this->getData('dateExpire');
+        return Attribute::make(
+            get: function () use ($withTimestamp) {
+                if (!$this->hasAttribute('image')) {
+                    return '';
+                }
+                $image = $this->getAttribute('image');
```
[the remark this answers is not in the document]
Vitaliy-1: Yes, it returns empty string

classes/core/SettingsBuilder.php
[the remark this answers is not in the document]
Vitaliy-1: Ahh, it should be `isset($count) ? ...` but your suggestion is shorter!

classes/core/traits/ModelWithSettings.php
```diff
@@ -0,0 +1,164 @@
...
+
+    // The list of multilingual attributes
+    protected array $multilingualProps = [];
+
+    /**
+     * Get main table name
+     *
+     * @return string
+     */
+    abstract public function getTable();
```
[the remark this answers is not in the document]
Vitaliy-1: Primarily it's for type hinting reasons. E.g., if I'm using `$this->getTable()` inside this trait, IDE and other developers know what is expected to be returned (without the need to look at the Eloquent Model class and other traits). PHP designed traits to be compatible with any classes, I can't narrow that to just Eloquent Model, thus I'm trying to be as specific as possible.

classes/core/traits/ModelWithSettings.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * See Illuminate\Database\Eloquent\Concerns\HasAttributes::mergeCasts()
+     */
+    abstract public function mergeCasts(array $casts);
+
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct($attributes);
+        if ($this->getSchemaName()) {
```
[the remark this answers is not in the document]
Vitaliy-1: Yes, my bad

classes/core/SettingsBuilder.php
```diff
@@ -0,0 +1,341 @@
...
+
+        if ($settingValues->isEmpty()) {
+            return $id;
+        }
+
+        $rows = [];
+        $settingValues->each(function (mixed $settingValue, string $settingName) use ($id, &$rows) {
+            $settingName = Str::camel($settingName);
+            if ($this->isMultilingual($settingName)) {
+                foreach ($settingValue as $locale => $localizedValue) {
```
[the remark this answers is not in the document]
Vitaliy-1: There is also related problem with casts, which should be addressed as part of this, I've opened the issue: https://github.com/pkp/pkp-lib/issues/10476

classes/core/SettingsBuilder.php
Vitaliy-1: Opened an issue: https://github.com/pkp/pkp-lib/issues/10487

Issue #10189, opened 2024-07-10
Convert Queries to an Eloquent model


asmecher:
The Query class (https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/query/Query.php) and QueryDAO (https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/query/QueryDAO.php) are implemented using the old-style DAO pattern. Migrate to an Eloquent model.

kaitlinnewson:
Hi @asmecher, could you give this PR a review? I'm hitting a wall with some errors I'm seeing in testing and could use a second look!

- Draft PR: https://github.com/pkp/pkp-lib/pull/10384

asmecher:
@kaitlinnewson, what errors are you getting? (Do you have a branch on OJS that's getting github actions kicked off?)

kaitlinnewson:
@asmecher Errors are in manual testing, but I'll also kick off the cypress tests - one I'm running into is when creating a new discussion in some cases, but it's not cropping up consistently:

```
[Thu Sep 12 16:12:14 2024] PHP Fatal error:  Uncaught TypeError: get_class(): Argument #1 ($object) must be of type object, null given in /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/mail/Mailer.php:79
Stack trace:
#0 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/mail/Mailer.php(79): get_class(NULL)
#1 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Mail/Mailer.php(257): PKP\mail\Mailer->renderView(NULL, Array)
#2 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(282): Illuminate\Mail\Mailer->render('', Array)
#3 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Support/Traits/Localizable.php(19): Illuminate\Mail\Mailable->Illuminate\Mail\{closure}()
#4 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(279): Illuminate\Mail\Mailable->withLocale(NULL, Object(Closure))
[...]

PR #10384, opened 2024-09-10
#10189 Convert Queries to an Eloquent Model

kaitlinnewson:


classes/query/Repository.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Retrieve a count of all open queries totalled by stage
+     *
+     * @param int[] $participantIds Only include queries with these participants
+     *
+     * @return array<int,int> [int $stageId => int $count]
+     */
+    public function countOpenPerStage(int $submissionId, ?array $participantIds = null)
```
kaitlinnewson, line 60: Still in progress - this is only currently used in `lib/pkp/classes/submission/maps/Schema.php`

classes/query/Repository.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Start a query
+     *
+     * Inserts the query, assigns participants, and creates the head note
+     *
+     * @return int The new query id
+     */
+    public function addQuery(int $submissionId, int $stageId, string $title, string $content, User $fromUser, array $participantUserIds, int $contextId, bool $sendEmail = true): int
```
kaitlinnewson, line 106: In progress - is only used in this file (in `addCommentsforEditorsQuery` below).

classes/query/Repository.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Create a query with a submission's comments for the editors
+     *
+     * Creates the query and assigns all participants
+     *
+     * @return int new query id
+     */
+    public function addCommentsForEditorsQuery(Submission $submission): int
```
kaitlinnewson, line 183: In progress - only current usage is in `lib/pkp/classes/submission/Repository.php`.

controllers/grid/queries/QueriesGridHandler.php
```diff
@@ -528,22 +525,20 @@ public function readQuery($args, $request)
...
 
         $participants = [];
-        foreach ($queryDao->getParticipantIds($query->getId()) as $userId) {
-            $user = Repo::user()->get($userId);
-            if ($user) {
-                $participants[] = $user;
+        $queryParticipants = $query->queryParticipants;
+        foreach ($queryParticipants as $participant) {
+            if ($participant->user) {
+                $participants [] = $participant->user;
```
taslangraham, line 534:
```suggestion
                $participants[] = $participant->user;
```

controllers/grid/queries/form/QueryForm.php
```diff
@@ -280,8 +288,10 @@ public function fetch($request, $template = null, $display = false, $actionArgs
         $templateMgr->assign('templates', $templateKeySubjectPairs);
 
         // Get currently selected participants in the query
-        $queryDao = DAORegistry::getDAO('QueryDAO'); /** @var QueryDAO $queryDao */
-        $assignedParticipants = $query->getId() ? $queryDao->getParticipantIds($query->getId()) : [];
+        $queryParticipants = QueryParticipant::withQueryId($query->id)
+            ->pluck('user_id')
+            ->all();
+        $assignedParticipants = $query->id ? $queryParticipants : [];
```
taslangraham, line 294:
```suggestion
        
        $assignedParticipants = $query->id ?  QueryParticipant::withQueryId($query->id)->pluck('user_id')->all() : [];
```

controllers/grid/queries/form/QueryForm.php
```diff
@@ -509,37 +519,41 @@ public function validate($callHooks = true)
...
-        $queryDao->removeAllParticipants($query->getId());
+        QueryParticipant::withQueryId($query->id)
+            ->delete();
         foreach ($newParticipantIds as $userId) {
-            $queryDao->insertParticipant($query->getId(), $userId);
+            QueryParticipant::create([
+                'queryId' => $query->id,
+                'userId' => $userId
+            ]);
         }
```
taslangraham, line 544: Is it possible to do a bulk insert?

templates/controllers/grid/queries/form/queryForm.tpl
```diff
@@ -32,7 +32,7 @@
 
 		{fbvFormSection list=true title="editor.submission.stageParticipants"}
 			{foreach from=$allParticipants item="participant" key="id"}
-				{fbvElement type="checkbox" id="users[]" value=$id checked=in_array($id, $assignedParticipants) label=$participant|escape translate=false}
+                {fbvElement type="checkbox" id="users[]" value=$id checked=in_array($id, $assignedParticipants) label=$participant|escape translate=false}
```
taslangraham, line 35: This should be indented

taslangraham (changes requested):
Mostly looks good @kaitlinnewson. I just had a few questions and suggestions

classes/query/QueryParticipant.php
```diff
@@ -0,0 +1,88 @@
...
+        'queryId', 'userId'
+    ];
+
+    protected function casts(): array
+    {
+        return [
+            'queryId' => 'int',
+            'userId' => 'int'
+        ];
+    }
```
Hafsa-Naeem, line 42: In eloquent models, attribute casting should be declared by using $casts property, not a method.
```
protected $casts = [
    'queryId' => 'int',
    'userId' => 'int',
];
```

classes/query/QueryParticipant.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Accessor and Mutator for primary key => id
+     */
+    protected function id(): Attribute
+    {
+        return Attribute::make(
+            get: fn($value, $attributes) => $attributes[$this->primaryKey] ?? null,
+            set: fn($value) => [$this->primaryKey => $value],
+        );
+    }
```
Hafsa-Naeem, line 57: here the accessor for id may not be necessary as eloquent handles the primary key automatically. since it uses `$primaryKey` property to manage the primary key. so `$queryParticipant->id` should work without any issue

classes/query/QueryParticipant.php
```diff
@@ -0,0 +1,88 @@
...
+    ];
+
+    protected function casts(): array
+    {
+        return [
+            'queryId' => 'int',
+            'userId' => 'int'
+        ];
+    }
+    public function toQuery(): BelongsTo
```
Hafsa-Naeem, line 43: suggestion: you can rename  `toQuery()` to `query()` to align with Laravel's naming conventions for relationships.

classes/query/Query.php
```diff
@@ -3,142 +3,138 @@
...
+
+	protected $table = 'queries';
+	protected $primaryKey = 'query_id';
+
+    protected $fillable = [
+        'assocType', 'assocId', 'stageId', 'seq',
+        'datePosted', 'dateModified', 'closed'
+    ];
+
+    protected function casts(): array
```
Hafsa-Naeem, line 39: In eloquent models, attribute casting should be declared by using $casts property, not a method.
```
protected $casts = [
    'assocType' => 'int',
    'assocId' => 'int',
    'stageId' => 'int',
    'seq' => 'float',
    'datePosted' => 'datetime',
    'dateModified' => 'datetime',
    'closed' => 'boolean',
];
```

classes/query/Query.php
```diff
@@ -3,142 +3,138 @@
...
     }
 
     /**
-     * Get query assoc ID
-     *
-     * @return int
+     * Accessor and Mutator for primary key => id
      */
-    public function getAssocId()
+    protected function id(): Attribute
```
Hafsa-Naeem, line 55: the accessor for id may not be necessary as eloquent handles the primary key automatically.

classes/query/Query.php
```diff
@@ -3,142 +3,138 @@
...
-     * @param int $assocId
+     * Accessor for user. Can be replaced with relationship once User is converted to an Eloquent Model.
      */
-    public function setAssocId($assocId)
+    protected function user(): Attribute
     {
-        $this->setData('assocId', $assocId);
+        return Attribute::make(
+            get: function () {
+                return Repo::user()->get($this->queryParticipant->userId, true);
```
Hafsa-Naeem, line 70: the `user()` accessor is referencing `$this->queryParticipant` which doesn't exist in the `Query` model. since `Query` has a `queryParticipants()` relationship that returns a collection of participants, you could adjust the accessor to retrieve participants correctly.

Hafsa-Naeem:
@kaitlinnewson the PR looks good to me. :+1:  I've added a few comments and suggestions.

classes/query/QueryParticipant.php
taslangraham: Our laravel version was recently [upgraded](https://github.com/pkp/pkp-lib/pull/10211) to v11 which defines [casts](https://laravel.com/docs/11.x/eloquent-mutators#attribute-casting) via a method.

classes/query/QueryParticipant.php
kaitlinnewson: I didn't follow the convention in this case because `query()` [is already a method in the `Model` class](https://laravel.com/api/11.x/Illuminate/Database/Eloquent/Model.html#method_query).

classes/query/Repository.php
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public function resequence($assocType, $assocId): void
+    {
+        $result = Query::withAssoc($assocType, $assocId)
+            ->orderBy('seq')
+            ->get();
+
+        $result->each(function ($item, $key = 1) {
+            Query::find($item->queryId)
+                ->update(['seq' => $key]);
```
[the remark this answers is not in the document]
kaitlinnewson: I'm giving this a try, but run into issues with `upsert` here, e.g.:  `Column not found: 1054 Unknown column 'assocId' in 'field list'`.

This may be due to the HasCamelCasing trait, where `upsert` is expecting the exact field name.

classes/query/QueryParticipant.php
kaitlinnewson: I think this is needed so we can still use the `id` attribute since our primary key has a custom name (e.g. `query_id` in this case, rather than the default `id`).

classes/query/Repository.php
```diff
@@ -0,0 +1,201 @@
...
```
touhidurabir, line 74: `upsert` will not work here as it will require exact field name but we can do something like 
```php
$result->each(function (Query $item, int $key = 1) {
    $item->update(['seq' => $key]);
});
```

here calling `find` again need to retrieve the row and then update it which seems double work to me as we already have those when we call `get` as Database Collection . Or I am missing something ?

touhidurabir (commented):
@kaitlinnewson I have added few comments, mostly suggestions/questions . Take a look and feel free to address those if you think it make sense . Otherwise this look good to me .

classes/query/Repository.php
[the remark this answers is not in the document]
Hafsa-Naeem: @kaitlinnewson thanks for giving the upsert approach a try. I see the issue with `upsert` and the `HasCamelCasing`. In this case @touhidurabir suggestion will work since we already have the collection from `get()`.  you can update the sequence directly within the` each()` loop without needing to use `find()` again. this will avoid the extra lookup. 😊

classes/query/Query.php
```diff
@@ -3,142 +3,141 @@
...
      *
-     * @param int $stageId
+     * @deprecated 3.5 Use $model->id instead. Can be removed once the DataObject pattern is removed.
      */
-    public function setStageId($stageId)
+    public function getId(): int
     {
-        return $this->setData('stageId', $stageId);
+        return $this->id;
     }
```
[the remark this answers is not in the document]
kaitlinnewson: There were a couple of cases in other Model conversions where we needed the `getId()` for the grid handlers, but I think all of the cases in this one were refactored so I've removed this.

classes/notification/managerDelegate/QueryNotificationManager.php
```diff
@@ -58,14 +54,14 @@ public function getNotificationMessage(PKPRequest $request, Notification $notifi
                     'noteTitle' => Str::limit($headNote->title, 200),
                 ]);
             case Notification::NOTIFICATION_TYPE_QUERY_ACTIVITY:
-                $latestNote = Note::withAssoc(PKPApplication::ASSOC_TYPE_QUERY, $query->getAssocId())
```
asmecher, line 61: Good catch here.

Issue #9890, opened 2024-04-16
Implement side menu navigation component


jardakotesovec:
**Goal**
Introduce new Vue.js component that would handle a sidebar navigation with expandable submenu items and integrate it to ojs&ops&omp.

Latest mockups available: https://www.figma.com/file/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?type=design&node-id=4369-21479&mode=design&t=R90FLzq8vXy4pSVd-4

**Notes**
- Use existing Icon and Badge components
- Ensure correct accessibility
- Add new menu to storybook

~**Following will be handled in separate issue(s)**~
~- Moving functionality from top navigation to primary navigation (user account, language switching,)~
~- Secondary navigation for submission listing (thats already implemented separately)~
~- Secondary navigation for other cases, where the navigation is currently handled via tabs is also outside of the scope of this one~
_Update: The design was revised, and we will no longer implement a secondary navigation, Instead, we will incorporate these as child menus._

Pull requests:
ui-library: https://github.com/pkp/ui-library/pull/408
pkp-lib: https://github.com/pkp/pkp-lib/pull/10390
customBlockManager: https://github.com/pkp/customBlockManager/pull/87 (_plugin_)
staticPages: https://github.com/pkp/staticPages/pull/85 (_plugin_)
webFeed: https://github.com/pkp/webFeed/pull/7 (_plugin_)

ojs: https://github.com/pkp/ojs/pull/4435
omp: https://github.com/pkp/omp/pull/1699
ops: https://github.com/pkp/ops/pull/773

Other ui-library PRs (_already merged_)
[pkp/pkp-lib#9890 Implement Collapsible Primary navigation](https://github.com/pkp/ui-library/pull/391)
[pkp/pkp-lib#9890 Add actionArgs parameter for sidemenu component](https://github.com/pkp/ui-library/pull/392)
[pkp/pkp-lib#9890 Add useSideMenu composable](https://github.com/pkp/ui-library/pull/393)
[pkp/pkp-lib#9890 Handle SideMenu state through useSideMenu api](https://github.com/pkp/ui-library/pull/394)
[pkp/pkp-lib#9890 Set item to active when items don't have child items](https://github.com/pkp/ui-library/pull/399)
[pkp/pkp-lib#9890 Add 'selectedItem' to useSideMenu api](https://github.com/pkp/ui-library/pull/401)

PR #10390, opened 2024-09-11
pkp/pkp-lib#9890 Implement new sidebar navigation component to OJS/OMP/OPS

blesildaramirez:


templates/stats/context.tpl
```diff
@@ -71,21 +71,23 @@
...
-						<tbody>
-							<tr	v-for="segment in timeline" :key="segment.date">
-								<th scope="row">{{ segment.label }}</th>
-								<td>{{ segment.value }}</td>
-							</tr>
-						</tbody>
-					</table>
+					<div class="sr-only">
+						<table class="-screenReader" role="region" aria-live="polite">
+							<caption>{translate key="stats.views.timelineInterval"}</caption>
```
jardakotesovec, line 76: This is because of the issue with wrong page height, right?

jardakotesovec (changes requested):
Looking great, please just refine the new selectors in cypress to be less specific (https://docs.cypress.io/guides/references/best-practices#Selecting-Elements).

Example:
Instead of 
```
        cy.get('nav div[data-pc-section="header"] a span').contains('Settings').click();

```

use something like
```
cy.get('nav').contains('Settings') 
```

Issue #10090, opened 2024-06-19
Reviewer's discussion brakes review anonymity


Vitaliy-1:
**Describe the bug**
In case of double blind review, Reviewer can open a discussion where author names are exposed 

**To Reproduce**
1. At a review stage, assign a reviewer with a double anonymous review mode. 
2. Log in as a reviewer and start a review
3. Proceed to the step 3 - Download & review
4. Click on add discussion
5. See author name

**What application are you using?**
 OJS main branch only

**Additional information**
![discussion](https://github.com/pkp/pkp-lib/assets/20665587/1df24cae-0733-4206-a567-459bbfa1d81f)

Hafsa-Naeem:
@asmecher @Vitaliy-1 Ready for review
pkp/pkp-lib: [pkp/pkp-lib#10090 Ensure anonymity in discussions for double blind reviews](https://github.com/pkp/pkp-lib/pull/10391)

Hafsa-Naeem:
@Vitaliy-1 Ready for review again
pkp/pkp-lib: [pkp/pkp-lib#10090 Ensure anonymity in discussions for double blind reviews](https://github.com/pkp/pkp-lib/pull/10391)

Tribunal33:
@Vitaliy-1 Does this need to be rebased or tested?

Vitaliy-1:
@Hafsa-Naeem, did you resolve comments Alec and I left during the review?

Hafsa-Naeem:
@asmecher @Vitaliy-1  updated and rebased for review:
pkp/pkp-lib: [pkp/pkp-lib#10090 Ensure anonymity in discussions for double blind reviews](https://github.com/pkp/pkp-lib/pull/10391)
ojs: [Submodule update ##Hafsa-Naeem/reviewer_anonimity_bug](https://github.com/pkp/ojs/pull/4767/commits)

Tribunal33:
Currently this is working correctly as far as I can tell?  Does this need a fix?  I don't see the author, as a reviewer using the discussion from step 3.  

![Image](https://github.com/user-attachments/assets/a37803ab-8979-47b9-9bcd-1f5e5abef2ad)

Vitaliy-1:
Thanks, this also requires porting to stable 3.5.0

Vitaliy-1:
In case of double blind review, nor author, nor reviewer should now each others personal information. In case of single blind review - reviewer is aware about the author

Vitaliy-1:
[...]

PR #10391, opened 2024-09-12
pkp/pkp-lib#10090 Ensure anonymity in discussions for double blind reviews

Hafsa-Naeem:
for [pkp/pkp-lib#10090](https://github.com/pkp/pkp-lib/issues/10090)

### DIFF
--- a/classes/stageAssignment/StageAssignment.php
+++ b/classes/stageAssignment/StageAssignment.php
@@ -152,4 +152,15 @@ class StageAssignment extends Model
                 ->where('submissions.context_id', $contextId);
         });
     }
+
+
+    /**
+     * Accessor for the userId attribute.
+     *
+     * @return int The user_id as an integer.
+     */
+    public function getUserIdAttribute(): int
+    {
+        return (int) $this->attributes['user_id'];
+    }
 }
--- a/classes/stageAssignment/StageAssignment.php
+++ b/classes/stageAssignment/StageAssignment.php
@@ -153,14 +153,4 @@ class StageAssignment extends Model
         });
     }
 
-
-    /**
-     * Accessor for the userId attribute.
-     *
-     * @return int The user_id as an integer.
-     */
-    public function getUserIdAttribute(): int
-    {
-        return (int) $this->attributes['user_id'];
-    }
 }
--- a/controllers/grid/queries/form/QueryForm.php
+++ b/controllers/grid/queries/form/QueryForm.php
@@ -342,7 +342,7 @@ class QueryForm extends Form
             // if current user is an anonymous reviewer, exclude authors
             foreach ($reviewAssignments as $reviewAssignment) {
                 if ($reviewAssignment->getReviewerId() == $user->getId()) {
-                    if ($reviewAssignment->getReviewMethod() != ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN) {
+                    if ($reviewAssignment->getReviewMethod() == ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS) {
                         $excludeUsers = StageAssignment::withSubmissionIds([$query->assocId])
                             ->withRoleIds([Role::ROLE_ID_AUTHOR])
                             ->get()


controllers/grid/queries/form/QueryForm.php
```diff
@@ -319,13 +319,11 @@ public function fetch($request, $template = null, $display = false, $actionArgs
...
-                    if ($reviewAssignment->getReviewMethod() != ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN) {
-                        // Replaces StageAssignmentDAO::getBySubmissionAndRoleId
+                    if ($reviewAssignment->getReviewMethod() == ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS ) {
+                        // exclude authors from the list of participants to ensure anonymity
                         $excludeUsers = StageAssignment::withSubmissionIds([$query->getAssocId()])
                             ->withRoleIds([Role::ROLE_ID_AUTHOR])
-                            ->withUserId($user->getId())
-                            ->get()
-                            ->pluck('userId')
+                            ->pluck('user_id')
```
Vitaliy-1, line 326: We avoid usage of column names outside of a query builder. I see that UseCamelCasing doesn't convert when not calling the model directly. If it's impossible to avoid, we can just create a scope for a Model. Here we can just get the collection first and then call pluck on the collection, how it was implemented before.

Vitaliy-1:
@Hafsa-Naeem, is the PR ready? If yes, can you update it?

Hafsa-Naeem:
@Vitaliy-1 Yes, I've made the necessary changes based on your feedback. The PR is now updated and ready for your review, let me know if anything else is needed.

Vitaliy-1:
It still has conflicts with the QueryForm, can you rebase it locally on top of the main, and ping me to merge?
E.g. (depending on your setup):
```
git checkout main
git pull upstream main
```
depending on what main you are tracking, e.g., you can also update your own own fork's main in the Github interface
```
git checkout main
git pull origin main
```
You can always check what branch is tracked by:
```
git status
git remote -v
```
And then do rebase and deal with conflicts:
```
git checkout reviewer_anonimity_bug
git rebase -i main
git push --force 
```

Hafsa-Naeem:
@Vitaliy-1 Thank you. I have resolved the conflicts and rebased the branch on top of the main. Please review.

Vitaliy-1:
I don't see any changes comparing with the current main. Probably something got lost during rebase. Can you double check please?

Hafsa-Naeem:
@Vitaliy-1 I’ve fixed the rebase issues and the changes. Please review. Thanks!

controllers/grid/queries/form/QueryForm.php
```diff
@@ -327,14 +327,11 @@ public function fetch($request, $template = null, $display = false, $actionArgs
...
-                        $excludeUsers = StageAssignment::withSubmissionIds([$query->assocId])
+                    if ($reviewAssignment->getReviewMethod() == ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS) {
+                        $stageAssignments = StageAssignment::withSubmissionIds([$query->assocId])
                             ->withRoleIds([Role::ROLE_ID_AUTHOR])
-                            ->withUserId($user->getId())
-                            ->get()
-                            ->pluck('userId')
-                            ->all();
+                            ->get();
+                        $excludeUsers = $stageAssignments->pluck('userId')->all();
```
Vitaliy-1, line 334: Can we retrieve excluded users without intermediate variables? Like:
```php
$excludeUsers = StageAssignment::withSubmissionIds([$query->assocId])
    ->withRoleIds([Role::ROLE_ID_AUTHOR])
    ->get()
    ->pluck('userId')
    ->all();
    ```

File: classes/stageAssignment/StageAssignment.php [hunk 3]
```diff
@@ -145,4 +145,15 @@ public function scopeWithContextId(Builder $query, ?int $contextId): Builder
...
         });
     }
+
+
+    /**
+     * Accessor for the userId attribute.
+     *
+     * @return int The user_id as an integer.
+     */
+    public function getUserIdAttribute(): int
```
Vitaliy-1, line 155: Do we need this method in the PR? `HasCamelCasing` trait should handle all cases of transforming between camel and snake case when Model is instantiated.

Vitaliy-1 (changes requested):

Vitaliy-1:
Thanks! I left comments, I think there is a method that was added to the StageAssignment by mistake

classes/stageAssignment/StageAssignment.php
Hafsa-Naeem: @Vitaliy-1  Yes I also wanted to avoid adding the `getUserIdAttribute` accessor. But I had to implement this accessor because the `pluck('userId')` method does not trigger the `HasCamelCasing` transformation. This trait works for direct property access and method calls, but `pluck` directly queries the database fields, which are in snake-case (`user_id`). Without the accessor, ` pluck('userId')` returns `null` values since it doesn't recognize the camel-case attribute, this is how the original code was and caused the issue of author names being visible in double-anonymous reviews. 

I initially faced this issue when trying to use `pluck('userId')` directly, which is why I added the accessor to ensure the correct transformation.

If preferred, I can remove the accessor and revert to using the snake_case user_id like `→pluck('user_id')`, but this approach is not ideal within our codebase. 

Another alternative could be using `map()` to iterate over the collection and access the `userId` attribute  but it is less efficient than `pluck()` and introduces unnecessary overhead.

controllers/grid/queries/form/QueryForm.php
Hafsa-Naeem: I'll update the code as you suggested here. But as mentioned in the  [comment](https://github.com/pkp/pkp-lib/pull/10391#discussion_r1785248281) about the accosser, I'll proceed with the final update once we decide on the approach regarding the accessor.

classes/stageAssignment/StageAssignment.php
asmecher: @Hafsa-Naeem, we're still developing consistent patterns around this so glad to have a chance to hash it out here. Sorry for the delay in reviewing!

I want to keep the use of actual column names in PHP (`snake_case` like `user_id`) limited to the storage code (Eloquent model or EntityDAO, collector, repository class) for that model. Outside of that, things should be `lowerCamelCase`.

The `HasCamelCasing` attribute covers operations related to an instance of a model, but not batch operations. So external code either has to materialize the set of models (untested; details might not be 100%)...
```php
$stageAssignments = StageAssignment::withSubmissionIds([$query->assocId])
    ->withRoleIds([Role::ROLE_ID_AUTHOR])
    ->get();
$excludeUsers = $stageAssignments->map(fn($stageAssignment) => $stageAssignment->userId);
```
...or, if that's going to be too much of a performance cost, I'd suggest adding a method to `lib/pkp/classes/stageAssignment/Repository.php`, which will keep the details of the `pluck` operation encapsulated with the storage code.

I think it would be a mistake to try to get Laravel to apply the `lowerCamelCase` conventions at the level of batch operations, and it'll move us further away from normal conventions a 3rd-party dev will expect.

Issue #10389, opened 2024-09-11
Track author Orcid verification request state in FieldOrcid component


ewhanson:
### Describe the issue

This issue concerns the ORCID verification functionality for authors within the contributor workflow. Currently, there is a button to trigger an email to request authors verify their ORCID. This button is part of a Vue component that currently tracks the state of the action locally. This means if you send a request to an author, you will see that the request has been sent, but if you reopen the contributor form, the request button state will be set back to its default. This state should be tracked to provide additional information to authors/editors.

### Describe the solution

This local state should be moved to the author settings table to track across page refreshes as something like a boolean `orcidVerificationSent`, which will need to be added to `author.json` in order to save the value to the `author_settings` table. 

The Vue component should track this value and conditionally display the `FieldOrcid` component in its "verification sent" state as long as the request has been sent once. (The ability to resend the request once it has been sent already is being handled separately, see: #10319). 

This value should be set to true when the request is sent and removed from the database once the ORCID has been verified.

This is currently only needed as part of the author/contributor workflow for sending ORCID verification requests as users are able to do so directly from within OJS without needing an email to be sent.

ewhanson:
@taslangraham, would you mind having a look at this one? Let me know if you need any further details beyond the description above. Thanks!

taslangraham:
@ewhanson Yes, I'll look into it.

taslangraham:
Ready for review @ewhanson 

### PRs
pkp-lib - https://github.com/pkp/pkp-lib/pull/10407
UI-library - https://github.com/pkp/ui-library/pull/409

**Submodule updates**
ojs - https://github.com/pkp/ojs/pull/4439
ops - https://github.com/pkp/ops/pull/771
omp - https://github.com/pkp/omp/pull/1697

ewhanson:
Hey @taslangraham, one minor comment, otherwise looks good. After you make the change, feel free to merge.

taslangraham:
All merged! Thanks for the review @ewhanson

PR #10407, opened 2024-09-12
pkp/pkp-lib#10389 Track author Orcid verification request state

taslangraham:
For https://github.com/pkp/pkp-lib/issues/10389

### DIFF
--- a/schemas/author.json
+++ b/schemas/author.json
@@ -193,6 +193,10 @@
 			"validation": [
 				"nullable"
 			]
+		},
+		"orcidVerificationRequested": {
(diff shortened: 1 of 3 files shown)


schemas/author.json
```diff
@@ -193,6 +193,10 @@
 			"validation": [
 				"nullable"
 			]
+		},
+		"orcidVerificationRequested": {
+			"type": "boolean",
+			"apiSummary": true
```
ewhanson, line 199: I don't think this needs to be included in the API summary as it's only really used in the `FieldOrcid`.

Issue #10359, opened 2024-09-03
Exposing date when editor confirmed review in API


jardakotesovec:
**Describe the issue**
One item initially I missed is that we would like to display date when editor confirms the review. Ideally it would be also the name, but at least date would suffice.

I currently use `dateDue`, `dateResponseDue`, `dateCompleted`, `dateCancelled` for these popups, but date for confirmation is not currently not available - we might not even track that in database.

<img width="235" alt="Screenshot 2024-09-03 at 10 10 48" src="https://github.com/user-attachments/assets/d78ab9a2-4c44-41f5-9b8d-ea9306117b57">

**What application are you using?**
 OJS, OMP or OPS version 3.5

jardakotesovec:
Hi @Vitaliy-1, sorry missed this one before. Would be possible still to add that?

Hafsa-Naeem:
Ready for review @Vitaliy-1

Vitaliy-1:
@Hafsa-Naeem, can you also make a PR to OJS main for tests? Now, with Github actions it's also possible to run tests on your fork

Hafsa-Naeem:
@Vitaliy-1 The tests are passing on the PR to OJS. 
OJS: [pkp/pkp-lib#10359 Update pkp-lib submodule to include the correct changes for showing editor confirm date #4462](https://github.com/pkp/ojs/pull/4462)

Vitaliy-1:
Looks good! Can we also name the column as `date_considered` following the naming for the `considered` column?

Vitaliy-1:
@Hafsa-Naeem, it still requires just column renaming. Let me know if any help here is required

Hafsa-Naeem:
@Vitaliy-1 Ready for review:
pkp-lib:[ pkp/pkp-lib#10359 Exposing date when editor confirmed review in API with dateEditorConfirmed](https://github.com/pkp/pkp-lib/pull/10450)
OJS: [[TESTS ONLY] Submodule update ##hafsanaeem/pkp/pkp-lib#10359 #4678](https://github.com/pkp/ojs/pull/4678)

Vitaliy-1:
Thanks, I left a couple comments to address in the review

PR #10450, opened 2024-09-19
pkp/pkp-lib#10359 Exposing date when editor confirmed review in API with dateEditorConfirmed

Hafsa-Naeem:
for #10359

### DIFF
--- a/schemas/reviewAssignment.json
+++ b/schemas/reviewAssignment.json
@@ -46,7 +46,6 @@
 		"dateConsidered": {
 		"type": "string",
 		"description": "Date when the editor confirmed the review",
-		"format": "date-time",
 		"validation": [
 			"date:Y-m-d H:i:s",
 			"nullable"
--- a/schemas/reviewAssignment.json
+++ b/schemas/reviewAssignment.json
@@ -43,6 +43,16 @@
 				"in:0,1,2,3,4"
 			]
 		},
+		"dateEditorConfirmed": {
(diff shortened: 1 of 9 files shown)


Vitaliy-1 (changes requested):
The comment: https://github.com/pkp/pkp-lib/issues/10359#issuecomment-2406660529

Vitaliy-1 (changes requested):
Thanks, Hafsa! I left a couple of comments. Also: 
1. Can you take a look at the `ThankReviewerForm`, I see [some logic there](https://github.com/pkp/pkp-lib/blob/a2b977dc7138e5f128e8e9144c4a9d7046a426b2/controllers/grid/users/reviewer/form/ThankReviewerForm.php#L152) to set `considered` property, does it require also setting a date? 
2. The migration should be added to the upgrade script of OJS, OMP, and, maybe, OPS - I see the `review_assignments` table there too...
3. Can you add the considered column to the installation script as well?

schemas/reviewAssignment.json
```diff
@@ -43,6 +43,15 @@
 				"in:0,1,2,3,4"
 			]
 		},
+		"dateConsidered": {
+		"type": "string",
```
Vitaliy-1, line 47: Can you check formatting here?

Vitaliy-1 (changes requested):
Thanks. Can you please make a PR to OMP and OPS?

PR #10456, opened 2024-09-19
Add support for ALTCHA, an alternative to current Google Recaptcha

Godoy0722:


classes/form/validation/FormValidatorAltcha.php
```diff
@@ -0,0 +1,126 @@
...
+
+        $challenge = (array) Altcha::createChallenge($options);
+
+        $templateMgr->assign('altchaEnabled', true);
+        $templateMgr->assign('altchaChallenge', $challenge);
+    }
+}
+
+if (!PKP_STRICT_MODE) {
+    class_alias('\PKP\form\validation\FormValidatorAltcha', '\FormValidatorAltcha');
```
asmecher, line 125: Unless it's required for some other reason, the `class_alias`es should only be provided for backwards compatibility; new code shouldn't require it except in a few exceptional cases.

js/lib/altcha/altcha.min.js
```diff
@@ -0,0 +1,8 @@
+/**
```
asmecher, line 1: Rather than committing 3rd-party code directly to our repository, is it possible to install this via an NPM dependency or, failing that, possibly even with Composer?

pages/login/LoginHandler.php
```diff
@@ -484,4 +506,37 @@ protected function sendHome($request)
...
+    }
+
+    /**
+     * Generate ALTCHA challenge to use it on form in case ALTCHA
+     * is enabled on the specific page
+     *
+     * @param string $altchaConfigKey the key to search on config.inc.php
+     * @param TemplateManager $templateMgr
+     */
+    private function generateAltchaComponent($altchaConfigKey, &$templateMgr): void
```
asmecher, line 535: Wherever possible, add typehints in new code; also remove redundant code self-documentation while you're at it. The `@param string $altchaConfigKey` adds a description, so it's valuable, but once it has a typehint, `@param TemplateManager $templateMgr` adds nothing and should be removed.

asmecher (changes requested):
Sorry for the extreme delay in reviewing this, @Godoy0722!

classes/form/validation/FormValidatorAltcha.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        $payload = (array) json_decode(base64_decode($response));
+
+        if (!Altcha::verifySolution($payload, $hmacKey)) {
+            throw new Exception('The ALTCHA validation failed.');
+        }
+    }
+
+    public static function addAltchaJavascript(TemplateManager &$templateMgr): void
```
asmecher, line 94: Reference is unnecessary here -- objects are passed by reference by default. (Here and below.)

classes/form/validation/FormValidatorAltcha.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $hmacKey = Config::getVar('captcha', 'altcha_hmackey');
+        if (empty($hmacKey)) {
+            throw new Exception('The ALTCHA is not configured correctly, the HMAC key is missing.');
+        }
+
+        $payload = (array) json_decode(base64_decode($response));
+
+        if (!Altcha::verifySolution($payload, $hmacKey)) {
+            throw new Exception('The ALTCHA validation failed.');
```
asmecher, line 90: Is this the message displayed if the ALTCHA test failed? If that's the case, it might be nicer to show a nicely-formatted page rather than an error page. I did recently add [exception handling](https://github.com/pkp/pkp-lib/issues/10027#issuecomment-2574187814), if you're curious, which opens the door to nicer error messages -- but historically we've used `frontend/pages/message.tpl` to show simple messages.

asmecher (changes requested):
Sorry for the delay, @Godoy0722! Just a couple of comments.

classes/form/validation/FormValidatorAltcha.php
Godoy0722: Hello Alec! The FormValidator catches this exception and makes an error message appear just above the login/registration form. This message is not a special one, it`s like a form validation error message. But it's nice to know about the exception handling! I'll take a look and in my next works, I'll deal with it better!

Godoy0722:
Hello again @asmecher ! Just a heads-up that I just updated your last change requests! Please let me know if you see something else for me!

asmecher:
Merged manually to resolve merge conflict.

Issue #7135, opened 2021-06-16
Need to support multiple author affiliations


kmccurley:
Apologies if this is already present in another issue or an internal development plan. I searched in issues but could not find anything related to multiple affiliations for an author.

A [recent study](https://arxiv.org/pdf/1912.05576.pdf) of 22 million articles published in 2019 showed that "almost one in three publications was (co-)authored by authors with multiple affiliations..." and "the share of authors with multiple affiliations increased from around 10% to 16% since 1996." The fact that OJS does not support multiple affiliations for authors means it is increasingly out of step with the realities of academic publishing, and my organization is reluctant to continue using OJS for this reason (among others).

I believe that an author should be able to specify multiple affiliations for a submission. This meshes nicely with the need to uniquely identify affiliations through the use of ROR identifiers. The use of a free text field alone for affiliations makes it difficult for machines to determine that "UC Berkeley" and "University of California - Berkeley" and "University of California, Berkeley" are in fact the same institution.

Having just installed the latest version of OJS, I noticed that affiliation information is stored in the underlying database as a row in `author_settings` using `setting_name` of `affiliation`, but the underlying table has a unique key of (`author_id`,`locale`,`setting_name`), which makes it impossible to store multiple affiliations unless the information is encoded in some way within the `setting_value` field. Our authors are currently listing multiple affiliations with `;` to separate them, but this is a bad practice for the future (much like journals entering bogus email addresses when the field was required).

As I mentioned before, the use of multiple affiliations is already an extremely common practice. The listing of affiliations has multiple purposes, including citation analysis to rank institutions, which strongly affects their funding. The attachment of an affiliation to a paper also strongly influences the reputation of the paper itself, and the inability to list multiple affiliations contributes to a "winner-take-all" attribution of credit, which is damaging to second-tier institutions and their authors.

Accuracy of affiliations is also important for identifying potential conflict of interest among reviewers.

Multiple affiliations are already supported by the following:
1. the [schema](https://github.com/pkp/pkp-lib/blob/main/plugins/importexport/native/pkp-native.xsd) for the `native plugin XSD` allows multiple affiliation tags per author, but when you import to OJS it [appears to discards this information](https://github.com/pkp/pkp-lib/blob/ede753267bd9ff162c981d88d0db31e86988900c/plugins/importexport/native/filter/NativeXmlPKPAuthorFilter.inc.php#L130). Obviously export fails to report multiple affiliations since the database only holds one.
[...]

PR #10460, opened 2024-09-20
pkp/pkp-lib#7135 Multiple author affiliations

GaziYucel:
Related issues:
- https://github.com/pkp/pkp-lib/issues/7135

Pull requests:
- https://github.com/pkp/pkp-lib/pull/10460
- https://github.com/pkp/ui-library/pull/417
- https://github.com/pkp/ojs/pull/4469
- https://github.com/pkp/omp/pull/1819
- https://github.com/pkp/googleScholar/pull/23
- https://github.com/pkp/jatsTemplate/pull/68
- https://github.com/asmecher/grobidMetadataExtractor/pull/1
- https://github.com/pkp/quickSubmit/pull/102
- https://github.com/pkp/crossref-ojs/pull/59

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -3,8 +3,8 @@
 /**
  * @file api/v1/submissions/PKPSubmissionController.php
  *
- * Copyright (c) 2023 Simon Fraser University
- * Copyright (c) 2023 John Willinsky
+ * Copyright (c) 2023-2024 Simon Fraser University
+ * Copyright (c) 2023-2024 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class PKPSubmissionController
@@ -665,7 +665,7 @@ class PKPSubmissionController extends PKPBaseController
 
         // Create an author record from the submitter's user account
         if ($submitAsUserGroup->roleId === Role::ROLE_ID_AUTHOR) {
-            $author = Repo::author()->newAuthorFromUser($request->getUser());
+            $author = Repo::author()->newAuthorFromUser($request->getUser(), $submission, $context);
             $author->setData('publicationId', $publication->getId());
             $author->setUserGroupId($submitAsUserGroup->id);
             $authorId = Repo::author()->add($author);
@@ -1552,6 +1552,22 @@ class PKPSubmissionController extends PKPBaseController
         }
 
         $author = Repo::author()->newDataObject($params);
+
+        $affiliations = [];
+        foreach ($params['affiliations'] as $affiliationParam) {
+            $affiliationErrors = Repo::affiliation()->validate(null, $affiliationParam, $submission, $submissionContext);
+            if (!empty($affiliationErrors)) {
+                return response()->json($affiliationErrors, Response::HTTP_BAD_REQUEST);
+            }
+            $affiliation = Repo::affiliation()->newDataObject($affiliationParam);
+            // rorObject will be an array, replace with ror object
+            if($affiliation->getRor() !== null){
+                $affiliation->setData('rorObject', Repo::ror()->newDataObject($affiliationParam['rorObject']));
+            }
+            $affiliations[] = $affiliation;
+        }
+        $author->setAffiliations($affiliations);
+
         $newId = Repo::author()->add($author);
         $author = Repo::author()->get($newId);
 
@@ -1675,6 +1691,28 @@ class PKPSubmissionController extends PKPBaseController
         if (!empty($errors)) {
             return response()->json($errors, Response::HTTP_BAD_REQUEST);
         }
+
+        $affiliations = [];
+        foreach ($params['affiliations'] as $affiliationParam) {
+            $affiliationErrors = Repo::affiliation()->validate(null, $affiliationParam, $submission, $submissionContext);
+            if (!empty($affiliationErrors)) {
+                return response()->json($affiliationErrors, Response::HTTP_BAD_REQUEST);
+            }
+            // Create a new affiliation object even if an existing affiliation is edited.
+            // This way we will have the list of all actual affiliations from the edited form.
+            // Later, in Repo::affiliation()->saveAffiliations(), we compare this list with
+            // the author affiliations that exist in the DB, to see if any was removed in the edited form.
+            $affiliation = Repo::affiliation()->newDataObject($affiliationParam);
+            // rorObject will be an array, replace with ror object
+            if($affiliation->getRor() !== null){
+                $affiliation->setData('rorObject', Repo::ror()->newDataObject($affiliationParam['rorObject']));
+            }
+            $affiliations[] = $affiliation;
+        }
+        $author->setAffiliations($affiliations);
+        // remove affiliations parameters because we have already set them properly for the author
+        // so that they are not considered once again when editing the author below
+        unset($params['affiliations']);
 
         Repo::author()->edit($author, $params);
         $author = Repo::author()->get($author->getId());
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -3,8 +3,8 @@
 /**
  * @file api/v1/submissions/PKPSubmissionController.php
  *
- * Copyright (c) 2023-2024 Simon Fraser University
- * Copyright (c) 2023-2024 John Willinsky
+ * Copyright (c) 2023-2025 Simon Fraser University
+ * Copyright (c) 2023-2025 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class PKPSubmissionController
@@ -33,6 +33,7 @@ use Illuminate\Support\Enumerable;
 use Illuminate\Support\Facades\Mail;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\LazyCollection;
+use PKP\affiliation\Affiliation;
 use PKP\components\forms\FormComponent;
 use PKP\components\forms\publication\PKPCitationsForm;
 use PKP\components\forms\publication\PKPMetadataForm;
@@ -1551,24 +1552,31 @@ class PKPSubmissionController extends PKPBaseController
             return response()->json($errors, Response::HTTP_BAD_REQUEST);
         }
 
+        $affiliationParams = $params['affiliations'];
+        unset($params['affiliations']);
         $author = Repo::author()->newDataObject($params);
+        $newId = Repo::author()->add($author);
 
-        $affiliations = [];
-        foreach ($params['affiliations'] as $affiliationParam) {
+        $affiliations = $newAffiliationErrors = [];
+        foreach ($affiliationParams as $position => $affiliationParam) {
+            $affiliationParam['authorId'] = $newId;
             $affiliationErrors = Repo::affiliation()->validate(null, $affiliationParam, $submission, $submissionContext);
+            // Map errors to the specific affiliation in the UI using the position = index
             if (!empty($affiliationErrors)) {
-                return response()->json($affiliationErrors, Response::HTTP_BAD_REQUEST);
+                $newAffiliationErrors['affiliations'][$position] = $affiliationErrors;
             }
             $affiliation = Repo::affiliation()->newDataObject($affiliationParam);
-            // rorObject will be an array, replace with ror object
-            if($affiliation->getRor() !== null){
-                $affiliation->setData('rorObject', Repo::ror()->newDataObject($affiliationParam['rorObject']));
-            }
             $affiliations[] = $affiliation;
         }
-        $author->setAffiliations($affiliations);
 
-        $newId = Repo::author()->add($author);
+        if (!empty($newAffiliationErrors)) {
+            return response()->json($newAffiliationErrors, Response::HTTP_BAD_REQUEST);
+        }
+
+        foreach ($affiliations as $affiliation) {
+            Repo::affiliation()->add($affiliation);
+        }
+
         $author = Repo::author()->get($newId);
 
         return response()->json(
@@ -1692,23 +1700,25 @@ class PKPSubmissionController extends PKPBaseController
             return response()->json($errors, Response::HTTP_BAD_REQUEST);
         }
 
-        $affiliations = [];
-        foreach ($params['affiliations'] as $affiliationParam) {
+        $affiliations = $newAffiliationErrors = [];
+        foreach ($params['affiliations'] as $position => $affiliationParam) {
             $affiliationErrors = Repo::affiliation()->validate(null, $affiliationParam, $submission, $submissionContext);
+            // Map errors to the specific affiliation in the UI using the position = index
             if (!empty($affiliationErrors)) {
-                return response()->json($affiliationErrors, Response::HTTP_BAD_REQUEST);
+                $newAffiliationErrors['affiliations'][$position] = $affiliationErrors;
             }
             // Create a new affiliation object even if an existing affiliation is edited.
             // This way we will have the list of all actual affiliations from the edited form.
             // Later, in Repo::affiliation()->saveAffiliations(), we compare this list with
             // the author affiliations that exist in the DB, to see if any was removed in the edited form.
             $affiliation = Repo::affiliation()->newDataObject($affiliationParam);
-            // rorObject will be an array, replace with ror object
-            if($affiliation->getRor() !== null){
-                $affiliation->setData('rorObject', Repo::ror()->newDataObject($affiliationParam['rorObject']));
-            }
             $affiliations[] = $affiliation;
         }
+
+        if (!empty($newAffiliationErrors)) {
+            return response()->json($newAffiliationErrors, Response::HTTP_BAD_REQUEST);
+        }
+
         $author->setAffiliations($affiliations);
         // remove affiliations parameters because we have already set them properly for the author
         // so that they are not considered once again when editing the author below
@@ -2145,7 +2155,7 @@ class PKPSubmissionController extends PKPBaseController
     protected function copyMultilingualData(Submission $submission, string $newLocale): void
     {
         $oldLocale = $submission->getData('locale');
-        $editProps = fn (Author|SubmissionFile $item, array $props): array => collect($props)
+        $editProps = fn (Author|SubmissionFile|Affiliation $item, array $props): array => collect($props)
             ->mapWithKeys(fn (string $prop): array => [$prop => ($data = $item->getData($prop)[$oldLocale] ?? null) ? [$newLocale => $data] : null])
             ->filter()
             ->toArray();
@@ -2166,11 +2176,17 @@ class PKPSubmissionController extends PKPBaseController
             'familyName',
             'preferredPublicName',
         ];
+        $affiliationProps = [
+            'name',
+        ];
         Repo::author()
             ->getCollector()
             ->filterByPublicationIds([$submission->getLatestPublication()->getId()])
             ->getMany()
-            ->each(function (Author $contributor) use ($contributorProps, $editProps, $newLocale) {
+            ->each(function (Author $contributor) use ($contributorProps, $affiliationProps, $editProps, $newLocale) {
+                foreach ($contributor->getAffiliations() as $affiliation) {
+                    Repo::affiliation()->edit($affiliation, $editProps($affiliation, $affiliationProps));
+                }
                 if (!($contributor->getData('givenName')[$newLocale] ?? null)) {
                     Repo::author()->edit($contributor, $editProps($contributor, $contributorProps));
                 }
--- a/classes/affiliation/Affiliation.php
+++ b/classes/affiliation/Affiliation.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/affiliation/Affiliation.php
  *
- * Copyright (c) 2024 Simon Fraser University
- * Copyright (c) 2024 John Willinsky
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class Affiliation
--- a/classes/affiliation/Affiliation.php
+++ b/classes/affiliation/Affiliation.php
@@ -1,0 +1,128 @@
+<?php
+
+/**
+ * @file classes/affiliation/Affiliation.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Affiliation
+ *
+ * @ingroup affiliation
+ *
+ * @see DAO
+ *
+ * @brief Basic class describing an affiliation.
+ */
+
+namespace PKP\affiliation;
+
+use APP\facades\Repo;
+use PKP\core\DataObject;
+use PKP\i18n\LocaleConversion;
+use PKP\ror\Ror;
+
+class Affiliation extends DataObject
+{
+    /**
+     * Get the default/fall back locale the values should exist for
+     * (see LocalizedData trait)
+     */
+    public function getDefaultLocale(): ?string
+    {
+        return Repo::author()->get($this->getAuthorId())->getDefaultLocale();
+    }
+
+    /**
+     * Get author id
+     */
+    public function getAuthorId(): int
+    {
+        return $this->getData('authorId');
+    }
+
+    /**
+     * Set author id
+     */
+    public function setAuthorId(int $authorId): void
+    {
+        $this->setData('authorId', $authorId);
+    }
+
+    /**
+     * Get the ROR ID
+     */
+    public function getRor(): ?string
+    {
+        return $this->getData('ror');
+    }
+
+    /**
+     * Set the ROR ID
+     */
+    public function setRor(?string $ror): void
+    {
+        $this->setData('ror', $ror);
+    }
+
+    /**
+     * Get the ROR object
+     */
+    public function getRorObject(): ?Ror
+    {
+        return $this->getData('rorObject');
+    }
+
+    /**
+     * Get affiliation name, considering also the ROR names
+     * Use getName() if only non ROR i.e.
+     * manually entered affiliation names should be considered.
+     */
+    public function getAffiliationName(?string $locale = null, ?array $allowedLocales = []): string|array|null
+    {
+        if ($rorObject = $this->getData('rorObject')) {
+            if ($locale == null) {
+                // try to map ROR locales to all submission locales
+                $names = [];
+                foreach ($allowedLocales as $allowedLocale) {
+                    $rorLocale = LocaleConversion::getIso1FromLocale($allowedLocale);
+                    $names[$allowedLocale] = $rorObject->getName($rorLocale) ?? $rorObject->getName($rorObject->getDisplayLocale());
+                }
+                return $names;
+            }
+            $rorLocale = LocaleConversion::getIso1FromLocale($locale);
+            return $rorObject->getName($rorLocale) ?? $rorObject->getName($rorObject->getDisplayLocale());
+        }
+        return $this->getData('name', $locale);
+    }
+
+    /**
+     * Get name or manually entered affiliation i.e. not considering ROR names
+     */
+    public function getName(?string $locale = null): string|array|null
+    {
+        return $this->getData('name', $locale);
+    }
+
+    /**
+     * Set name
+     */
+    public function setName(string|array|null $name, ?string $locale = null): void
+    {
+        $this->setData('name', $name, $locale);
+    }
+
+    /**
+     * Get localized affiliation name.
+     */
+    public function getLocalizedName(?string $preferredLocale = null): string|null
+    {
+        $rorObject = $this->getRorObject();
+        if ($rorObject) {
+            $preferredLocale = $preferredLocale ? LocaleConversion::getIso1FromLocale($preferredLocale) : $preferredLocale;
+            return $rorObject->getLocalizedName($preferredLocale);
+        }
+        return $this->getLocalizedData('name', $preferredLocale);
+    }
+}
--- a/classes/affiliation/Collector.php
+++ b/classes/affiliation/Collector.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/affiliation/Collector.php
  *
- * Copyright (c) 2024 Simon Fraser University
- * Copyright (c) 2024 John Willinsky
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class Collector
@@ -32,7 +32,6 @@ class Collector implements CollectorInterface
 
     public ?int $offset = null;
 
-    /** @var int|null Get affiliations with author id */
     public ?int $authorId = null;
 
     public function __construct(DAO $dao)
--- a/classes/affiliation/Collector.php
+++ b/classes/affiliation/Collector.php
@@ -1,0 +1,120 @@
+<?php
+
+/**
+ * @file classes/affiliation/Collector.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Collector
+ *
+ * @brief A helper class to configure a Query Builder to get a collection of affiliations
+ */
+
+namespace PKP\affiliation;
+
+use Illuminate\Database\Query\Builder;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\LazyCollection;
+use PKP\core\interfaces\CollectorInterface;
+use PKP\plugins\Hook;
+
+/**
+ * @template T of Affiliation
+ */
+class Collector implements CollectorInterface
+{
+    public DAO $dao;
+
+    public ?int $count = null;
+
+    public ?int $offset = null;
+
+    /** @var int|null Get affiliations with author id */
+    public ?int $authorId = null;
+
+    public function __construct(DAO $dao)
+    {
+        $this->dao = $dao;
+    }
+
+    /** @copydoc DAO::getCount() */
+    public function getCount(): int
+    {
+        return $this->dao->getCount($this);
+    }
+
+    /**
+     * @copydoc DAO::getIds()
+     *
+     * @return Collection<int,int>
+     */
+    public function getIds(): Collection
+    {
+        return $this->dao->getIds($this);
+    }
+
+    /**
+     * @copydoc DAO::getMany()
+     *
+     * @return LazyCollection<int,T>
+     */
+    public function getMany(): LazyCollection
+    {
+        return $this->dao->getMany($this);
+    }
+
+    /**
+     * Filter by single author
+     */
+    public function filterByAuthorId(?int $authorId): self
+    {
+        $this->authorId = $authorId;
+        return $this;
+    }
+
+    /**
+     * Limit the number of objects retrieved
+     */
+    public function limit(?int $count): self
+    {
+        $this->count = $count;
+        return $this;
+    }
+
+    /**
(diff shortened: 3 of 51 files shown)


GaziYucel:
Ror registry data set cache

- update bi-weekly
- api lookup feature
- migration / install
- translations
  - english is ready
  - other languages have the keys with empty message

schemas/affiliation.json
kmccurley, line 1: The schema lists ror as a required field. What happens when an author is affiliated with an institution that does not have an assigned ROR ID? While it's probably safe to assume that universities have these, it's certainly not true for industrial affiliations that are common in computer science. It took me only a few minutes looking in arxiv until I found a few such as https://arxiv.org/pdf/2410.08087 https://arxiv.org/pdf/2410.08081 https://arxiv.org/pdf/2410.08067 https://arxiv.org/pdf/2410.02445 https://arxiv.org/pdf/2410.07629

asmecher: @kmccurley, the ROR will not be a required field; it will always be possible to add plain-text affiliations in order to accommodate users working with non-ROR research institutions.

GaziYucel: Thanks @kmccurley, @asmecher for your feedback. As Alec said, ror is not a required field. I pushed a corrected version.

GaziYucel:
Hi,

Some notes. 

1. 
I have made some assumptions regarding business logic in several places. The ones I am uncertain about are the import/export classes: 
* plugins/importexport/native/filter/NativeXmlPKPAuthorFilter.php
* plugins/importexport/native/filter/PKPAuthorNativeXmlFilter.php

2. 
During an upgrade from an earlier version, the ROR cache is filled during the upgrade. After this a migration is done for all affiliations currently saved for the authors. First a match is searched for all ROR on an exact match for the names. After this all remaining is migrated as a non-ROR affiliation. 

3. 
There is an issue regarding the ROR data. Most affiliations have a name in one language, sometimes 2 -3 languages. Not all have a name in English. This is an issue regarding searching for a name in the backend. 
For now I look for an alternate name and show this if an English name is not there, with at the end the language. For example: "Technische Universiteit Delft [Dutch]". 

4.
The class in "classes/task/UpdateRorRegistryDataset.php" is responsible for updating the ROR cache. There are a lot of records without a language code. These are marked with "no_lang_code". I replace these with the locale code 'en'. 

5.
There is a question regarding the class PKP\API\v1\submissions\PKPSubmissionController.

If you look at the method getContributor, you see this at the end of this method. 

```
return response()->json(
    Repo::author()->getSchemaMap()->map($author),
    Response::HTTP_OK
);
```

I expected that the method "getSchemaMap()->map($author), also transforms "affiliations" to it's schema. But this is not happening, only author is mapped. Affiliations stays as in PHP, with "_data" and other keys still in it. 

I defined in the schema of author (schema/author.json) the following: 

```
"affiliations": {
	"type": "array",
	"description": "The scholarly institution this contributor is employed by or affiliated with.",
	"apiSummary": true,
	"readOnly": true,
	"items": {
		"type": "object",
		"$ref": "#/definitions/Affiliation"
	}
},
```
This is what I came up with, but this feels as a hack. 

```
$mappedAffiliations = [];
foreach ($author->getAffiliations() as $key => $affiliation) {
    $mappedAffiliations[] = Repo::affiliations()->getSchemaMap()->map($affiliation);
}

$author->setAffiliations(new LazyCollection($mappedAffiliations));
```

How should I do this to make affiliations also transform/map to it's schema?

GaziYucel:
Hi @bozana. This PR is ready for review. Thanks.

bozana:
Hi @GaziYucel, I am still reviewing the code and would also like to test it locally (also to understand some things better) but I wanted to already suggest something regarding the native import/export:
I think we could introduce the additional element `rorAffiliation` for the author, that could occur several times and that would contain the element `ror` and localized element `name`. 
The existing localized element `affiliation` could be used for the authors affiliations without ROR.
The examples how multiple affiliations and an affiliated institution with its identifier are represented in JATS can be found here: https://jats.nlm.nih.gov/archiving/tag-library/1.3/element/aff.html

bozana:
Hi @GaziYucel, 
and maybe regarding your last question above -- "And a question regarding the class PKP\API\v1\submissions\PKPSubmissionController" -- regarding the mapping of affiliations to the author: I think you could take a look how this is solved for issues and articles, e.g. here https://github.com/pkp/ojs/blob/main/classes/issue/maps/Schema.php#L111. I think something like this could work for this case as well.

bozana:
Hi @GaziYucel, 
And regarding the replacing `no_lang_code` with `en`: I think we cannot do it -- this way the original `en` entry can be overridden. We just cannot map `no_lang_code` to anything correctly. We would need to leave it as is.

classes/ror/Ror.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace PKP\ror;
+
+use PKP\core\DataObject;
+
+class Ror extends DataObject
+{
+    public const STATUS_ACTIVE = 1;
+    public const STATUS_INACTIVE = 0;
```
bozana, line 25: I believe we could add the getters and setters -- I think we decided to use them so that we better support type hinting...

bozana (commented):
Hi @GaziYucel, I left some comments. 
Because we decided not to copy ROR names to author_affiliation_settings table some things will change here, so I can take a look once that is done...
About a few things I would maybe also need to think a little bit more, but I left a comment -- maybe you would immediately have an idea...
And because I have not taken a look at the OJS PR yet i.e. I would need some time to see how such affiliations need to be exported into different formats, I will probably first then know what Affiliation::getLocalized... function are needed... 
So I am already submitting this comments so that you can eventually already integrate some of them when working on the new version without copying...
I am wondering if we should keep this code version in case Alec has good arguments that we should proceed with copying -- because he has not said anything yet... although I do not think it will be the case...
Thanks a lot for this great work!

classes/author/Author.php
```diff
@@ -256,4 +257,117 @@ public function getLocalizedCompetingInterests(): ?string
...
+
+    /**
+     * Get the localized affiliations as a string.
+     *
+     * @param string|null $preferredLocale
+     * @param string|null $separator
+     *
+     * @return string
+     */
+    public function getLocalizedAffiliationNamesAsString(?string $preferredLocale = null, ?string $separator = ', '): string
```
bozana, line 366: Lets use ';' as a default separator

classes/author/Author.php
```diff
@@ -256,4 +257,117 @@ public function getLocalizedCompetingInterests(): ?string
     {
         return $this->getLocalizedData('competingInterests');
     }
+
+    /**
+     * Get affiliations (position, institution, etc.).
+     *
+     * @return LazyCollection
+     */
+    public function getAffiliations(): LazyCollection
```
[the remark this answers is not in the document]
GaziYucel: I use methods which are in LazyCollection such as "all()".

classes/ror/DAO.php
```diff
@@ -0,0 +1,206 @@
...
+    /** @copydoc EntityDAO::insert() */
+    public function insert(Ror $ror): int
+    {
+        return parent::_insert($ror);
+    }
+
+    /** @copydoc EntityDAO::update() */
+    public function update(Ror $ror): void
+    {
+        if (empty($ror->getId())) {
```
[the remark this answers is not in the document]
GaziYucel: ror->getId is never empty/null. You are correct.

classes/migration/install/AffiliationsMigration.php
```diff
@@ -0,0 +1,60 @@
...
+            $table->foreign('author_id')->references('author_id')->on('authors')->cascadeOnDelete();
+        });
+
+        Schema::create('author_affiliation_settings', function (Blueprint $table) {
+            $table->comment('More data about author affiliations');
+            $table->bigInteger('author_affiliation_setting_id')->autoIncrement();
+            $table->bigInteger('author_affiliation_id');
+            $table->string('locale', 28)->default('');
+            $table->string('setting_name', 255);
+            $table->mediumText('setting_value')->nullable();
```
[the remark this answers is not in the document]
GaziYucel: Yes, the value for name can be null. I add all current locales to the table, even if they are null.

classes/migration/install/RorsMigration.php
```diff
@@ -0,0 +1,68 @@
...
+            $table->index(['is_active'], 'rors_is_active');
+        });
+
+        Schema::create('ror_settings', function (Blueprint $table) {
+            $table->comment('More data about Ror registry dataset cache');
+            $table->bigInteger('ror_setting_id')->autoIncrement();
+            $table->bigInteger('ror_id');
+            $table->string('locale', 28)->default('');
+            $table->string('setting_name', 255);
+            $table->mediumText('setting_value')->nullable();
```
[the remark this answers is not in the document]
GaziYucel: Yes, the value for name can be null. I add all current locales to the table, even if they are null.

classes/ror/maps/Schema.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $output = [];
+        foreach ($props as $prop) {
+            switch ($prop) {
+                default:
+                    $output[$prop] = $item->getData($prop);
+                    break;
+            }
+        }
+        $output = $this->schemaService->addMissingMultilingualValues($this->schema, $output, $this->context->getSupportedFormLocales());
```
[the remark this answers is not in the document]
GaziYucel: You are right, we need to add the locales as they are provided by ROR. 
Line removed.

controllers/grid/users/author/form/PKPAuthorForm.php
```diff
@@ -225,7 +227,7 @@ public function execute(...$functionParams)
         $author->setGivenName(array_map('trim', $this->getData('givenName')), null);
         $author->setFamilyName($this->getData('familyName'), null);
         $author->setPreferredPublicName($this->getData('preferredPublicName'), null);
-        $author->setAffiliation($this->getData('affiliation'), null); // localized
+        $author->setAffiliations($this->getData('affiliations'));
```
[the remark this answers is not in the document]
GaziYucel: Yes, affiliations is always LazyCollection.

classes/ror/Repository.php
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Validate properties for a ror
+     *
+     * Perform validation checks on data used to add or edit a ror.
+     *
+     * @param Ror|null $object Ror being edited. Pass `null` if creating a new submission
+     * @param array $props A key/value array with the new data to validate
+     * @param array $allowedLocales The context's supported locales
+     * @param string $primaryLocale The context's primary locale
```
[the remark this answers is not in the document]
GaziYucel: I removed the "validate" method. As you pointed out, it's not needed.

classes/affiliation/maps/Schema.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $output = [];
+        foreach ($props as $prop) {
+            switch ($prop) {
+                default:
+                    $output[$prop] = $item->getData($prop);
+                    break;
+            }
+        }
+        $output = $this->schemaService->addMissingMultilingualValues($this->schema, $output, $this->context->getSupportedFormLocales());
```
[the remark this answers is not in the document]
GaziYucel: Author has no submissionId in it's schema, only publicationId. I added get/setPublicationId. 
This is what I implemented: 
```
$author = Repo::author()->get($item->getAuthorId());
$submissionId = Repo::publication()->get($author->getPublicationId())->getData('submissionId');
$locales = Repo::submission()->get($submissionId)->getPublicationLanguages($this->context->getSupportedSubmissionMetadataLocales());
$output = $this->schemaService->addMissingMultilingualValues($this->schema, $output, $locales);
```

classes/author/Author.php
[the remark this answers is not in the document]
bozana: I am still not sure if this is the best way to go, but lets see it at the end...

classes/migration/install/AffiliationsMigration.php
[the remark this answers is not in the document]
bozana: OK.
Note that you might need to get the possible submission locales differently, i.e. `$submission->getPublicationLanguages($context->getSupportedSubmissionMetadataLocales());`

GaziYucel:
Hi @bozana
I implemented validation for affiliations in this commit: https://github.com/pkp/pkp-lib/pull/10460/commits/f92025f34b10251d3494a3f57179c0f72837d343

I had to remove Repo::affiliation()->saveAffiliations(..) from Repo::author()->dao->insert(..) and Repo::author()->dao->update(...) for this. I added Repo::affiliation()->saveAffiliations(..) to classes where the insert and update are called. Otherwise I can't check if an author exists. 

I put all in one commit so that you can review it more easily.

classes/migration/install/AffiliationsMigration.php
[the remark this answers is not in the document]
GaziYucel: Hi @bozana,
I can't figure out how to get $submission in the classes PKP\core\PKPApplication\FieldAffiliations. I do see this in PKP\components\forms\publication\ContributorForm, but it's always null. 

I do see different locales in PKPApplication::get()->getRequest()->getContext(): 
- supportedAddedSubmissionLocales
- supportedDefaultSubmissionLocale
- supportedFormLocales
- supportedLocales
- supportedSubmissionLocales
- supportedSubmissionMetadataLocales

But you pointed out that that is not correct. Thanks.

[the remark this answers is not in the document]
bozana: Hi @GaziYucel, you need submission only because of allowed/supported locales, right?
If so, maybe: the ContributorForm actually contains all needed information about locales, so maybe to provide them (locales that you need for the affiliation field) here https://github.com/pkp/ui-library/blob/main/src/components/ListPanel/contributors/ContributorsListPanel.vue#L393, similar as you provide the authorId?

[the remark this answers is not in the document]
GaziYucel: I already saw it there, but the variables are called different there, not sure if these have the values you wrote above. These are the locales in ContributorsListPanel:
- primaryLocale, e.g. 'en'
- supportedFormLocales, e.g. [ { key: 'en', label: 'English' }, { key: 'fr_CA', label: 'French (Canada)' } ]
- visibleLocales, e.g. [ 'en' ]
​
And the one I use now in PHP is $submissionContext->getSupportedSubmissionLocales()

Which one is the correct one than?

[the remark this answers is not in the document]
bozana: The locales for the ContributorForm are provided here: https://github.com/pkp/pkp-lib/blob/main/classes/components/listPanels/ContributorsListPanel.php#L120. Those are the locales of the ContributorsListPanel.
And locales provided to the ContributorsListPanel (s. different places/handler in the code) are:
```
$locales = collect($submissionContext->getSupportedSubmissionMetadataLocaleNames() + $submission->getPublicationLanguageNames())
            ->map(fn (string $name, string $locale) => ['key' => $locale, 'label' => $name])
            ->sortBy('key')
            ->values()
            ->toArray();
```
Those are the supportedFormLocales.
The primary and visibleLocales are the same for each Form.
So I think you should only work with these locales for affiliations as well.
 
`$submissionContext->getSupportedSubmissionLocales() ` is not enough because they contain only currently enabled submission locales (and not metadata locales). The submission locales are those a submission can be published in, i.e. the article full text; and metadata locales can contain/allow additional locales for metadata -- e.g. an article in German, but the metadata also in English. 
Additionally, if already some metadata exists for a submission in a locale that is in the meantime disabled in journal settings -- lets say the metadata were entered in Finish, but then the Finish is disabled -- the users/editor still needs to be able to edit those existing metadata (in Finish). That is why we consider also $submission->getPublicationLanguageNames() (that returns all existing publication languages).

I have not taken a look into the ui-library, but will take a look where you need to locales and how...
But I believe those 3 kinds of form locales should be enough and correct for affiliations...

classes/author/Author.php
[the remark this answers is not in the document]
GaziYucel: I changed them to array.

classes/affiliation/Affiliation.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function getName(): ?array
+    {
+        return $this->getData('name');
+    }
+
+    /**
+     * Set name
+     */
+    public function setName(?array $name): void
```
bozana, line 68: I think we should stay consistent with other 'settings' method that could have locale and have something like:
```
    public function setName(string|array $name, ?string $locale = null)
    {
        $this->setData('name', $name, $locale);
    }

    public function getName(?string $locale = null)
    {
        return $this->getData('name', $locale);
    }
```
When you then provide an array as $name and no locale to setName() it will/should be set correctly.

classes/affiliation/DAO.php
```diff
@@ -0,0 +1,175 @@
...
+                    unset($fromRow->_data['name'][Ror::NO_LANG_CODE]);
+                    if(empty($fromRow->_data['name'][$submissionLocale])) {
+                        $displayLocale = $rors[$fromRow->_data['ror']]->_data['displayLocale'];
+                        $fromRow->_data['name'][$submissionLocale] =
+                            $rors[$fromRow->_data['ror']]->_data['name'][$displayLocale];
+                    }
+                }
+                yield $row->author_affiliation_id => $fromRow;
+            }
+        });
```
bozana, line 127: This is not easy to read and understand, so maybe to do it similarly to author DAO and set affiliations?
To get the affiliation fromRow() and then get the Ror by ID and set the affiliations names.

bozana: Then we would also not need that function getManyRorAsCollectionId() in Ror Collector

classes/affiliation/DAO.php
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /** @copydoc EntityDAO::fromRow() */
+    public function fromRow(object $row): Affiliation
+    {
+        return parent::fromRow($row);
+    }
```
bozana, line 134: No need for this function then

classes/ror/Ror.php
```diff
@@ -0,0 +1,53 @@
...
+ * @brief Basic class describing a ror.
+ */
+
+namespace PKP\ror;
+
+use PKP\core\DataObject;
+
+class Ror extends DataObject
+{
+    public const int STATUS_ACTIVE = 1;
```
bozana, line 25: it should be `public const STATUS_ACTIVE = 1;` i.e. without the type `int`
here and below

classes/affiliation/Collector.php
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * Filter by authors
+     */
+    public function filterByAuthorIds(?array $authorIds): self
+    {
+        $this->authorIds = $authorIds;
+        return $this;
+    }
```
bozana, line 86: It seems this function is never used? 
Then maybe to leave the filtering only by one author ID?

classes/affiliation/Collector.php
```diff
@@ -0,0 +1,161 @@
...
+    public ?int $authorId = null;
+
+    /** @var int[]|null Get affiliations with author ids */
+    public ?array $authorIds = null;
+
+    /** Get affiliations with a name */
+    public ?string $name = null;
+
+    /** Get affiliations with a searchPhrase */
+    public ?string $searchPhrase = null;
```
bozana, line 45: It seems we do not consider this anywhere

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,171 @@
...
+                    $join
+                        ->on('r.ror_id', '=', 'r_s.ror_id')
+                        ->where('r_s.setting_name', '=', 'name');
+                }
+            )
+            ->join('author_settings as a_s',
+                function (JoinClause $join) {
+                    $join
+                        ->on('r_s.setting_value', '=', 'a_s.setting_value')
+                        ->on('r_s.locale', '=', 'a_s.locale')
```
bozana, line 102: does locale needs to be the same?

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,171 @@
...
+
+        foreach ($rows as $row) {
+            $ror = Repo::ror()->get($row->ror_id);
+
+            $affiliation = Repo::affiliation()->newDataObject();
+            $params = [
+                "id" => null,
+                "authorId" => $row->author_id,
+                "ror" => $ror->_data['ror'],
+                "name" => $ror->_data['name']
```
bozana, line 120: Maybe to use $ror->getData('ror') and $ror->getData('name')

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,171 @@
...
+            ];
+            $affiliation->setAllData($params);
```
bozana, line 122: Maybe to use $affiliation->setAuthorId(),  $affiliation->setRor() and $affiliation->setName()

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,171 @@
...
+
+            Repo::affiliation()->dao->updateOrInsert($affiliation);
```
bozana, line 124: In upgrade scripts we are using only SQLs and no objects and functions -- so that the upgrade can be correctly done for the given version even if we change the objects and code.
Thus, we should try to use the query builder to insert or update the affiliations.

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,171 @@
...
+                "id" => null,
+                "authorId" => $row->author_id,
+                "ror" => '',
+                "name" => [
+                    $row->locale => $row->setting_value
+                ]
+            ];
+            $affiliation->setAllData($params);
+
+            Repo::affiliation()->dao->updateOrInsert($affiliation);
```
bozana, line 168: Here as well, we should use the query builder to insert or update the affiliation

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+                    "id" => null,
+                    "authorId" => null,
+                    "ror" => $ror->_data['ror'],
+                    "name" => $ror->_data['name']
+                ];
+                break;
+            }
+        }
+
+        $affiliation->setAllData($params);
```
bozana, line 298: I think it is better to use setName, setRor and get rid of $params array.

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+            $this->delete($row);
+        }
+    }
+
+    /**
+     * Get all affiliations for a given author.
+     */
+    public function getByAuthorId(int $authorId, ?string $submissionLocale = null): array
+    {
+        return iterator_to_array(
```
bozana, line 201: instead of using the php function iterator_to_array we should use Laravel all() function:
```
$this->getCollector()
  ->filterByAuthorId($authorId)
  ->getMany($submissionLocale)
  ->all()
```

Here and elswhere...

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1558,6 +1559,15 @@ public function addContributor(Request $illuminateRequest): JsonResponse
 
         $author = Repo::author()->newDataObject($params);
         $newId = Repo::author()->add($author);
+
+        $params['id'] = $newId;
+        $author->setId($newId);
+        $author->setAffiliations($params['affiliations']);
```
bozana, line 1565: is $params['affiliations'] array of Affiliation objects?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1558,6 +1559,15 @@ public function addContributor(Request $illuminateRequest): JsonResponse
 
         $author = Repo::author()->newDataObject($params);
         $newId = Repo::author()->add($author);
+
+        $params['id'] = $newId;
```
bozana, line 1563: should this be 'authorId' instead of 'id'?

api/v1/submissions/PKPSubmissionController.php
bozana: I think you should here do:
1. Repo::affiliation()->validate($author, $params, $submission, $submissionContext);
2. if error -> return
3. Repo::affiliation()->newDataObject($params);

Then no need to have author->setAffiliations, because you get the author at the end `$author = Repo::author()->get($newId);` which will get the affiliations.

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+
+        $schemaService = app()->get('schema');
+        $allowedLocales = $submission->getPublicationLanguages($context->getSupportedSubmissionMetadataLocales());
+
+        // Check if author exists
+        if(!empty($props['id'])){
+            $validator = ValidatorFactory::make(
+                $props,
+                $schemaService->getValidationRules(PKPSchemaService::SCHEMA_AUTHOR, $allowedLocales)
+            );
```
bozana, line 110: I do not think you need this

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+    {
+        $errors = [];
+
+        $affiliations = (empty($author)) ? $props['affiliations'] : $author->getAffiliations();
+
+        $schemaService = app()->get('schema');
+        $allowedLocales = $submission->getPublicationLanguages($context->getSupportedSubmissionMetadataLocales());
+
+        // Check if author exists
+        if(!empty($props['id'])){
```
bozana, line 106: can we call it `authorId` instead of `id`?

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+            $validator = ValidatorFactory::make(
+                $props,
+                $schemaService->getValidationRules(PKPSchemaService::SCHEMA_AUTHOR, $allowedLocales)
+            );
+            $validator->after(function ($validator) use ($props) {
+                if (isset($props['id']) && !$validator->errors()->get('id')) {
```
bozana, line 112: If you check above for `isset($props['id'])` then no need here

api/v1/submissions/PKPSubmissionController.php
bozana: Maybe also to do here:
foreach ($params['affiliations'] as $affiliationParam) {
then the steps above
and having `Repo::affiliation()->validate(null, $params, $author, $submission, $submissionContext);` i.e. validating only one affiliation i.e. having the first parameter Affiliation object -- see below where the validate is implemented.

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+     * @param Author|null $author Author being edited. Pass `null` if creating a new author
+     * @param array $props A key/value array with the new data to validate
+     * @param Submission $submission The context's supported locales
+     * @param Context $context The context's primary locale
+     *
+     * @return array A key/value array with validation errors. Empty if no errors
+     *
+     * @hook Affiliation::validate [[&$errors, $object, $props, $submission, $context]]
+     */
+    public function validate(?Author $author, array $props, Submission $submission, Context $context): array
```
bozana, line 96: Maybe to have the syntax as in all validate functions: to validate only one affiliation and to have the first parameter Affiliation object that could be null ?

api/v1/submissions/PKPSubmissionController.php
bozana: Maybe to create Affiliation objects and use author->setAffiliations before adding the author.
Then in the author DAO the author is first added and then the affiliations saved, containing the author ID.

classes/affiliation/Repository.php
```diff
@@ -0,0 +1,302 @@
...
+                ->getMany($submissionLocale)
+        );
+    }
+
+    /**
+     * Save affiliations.
+     */
+    public function saveAffiliations(Author $author): void
+    {
+        $affiliations = $author->getAffiliations();
```
bozana, line 213: Here we can expect the affiliations to always be an array of Affiliation objects. (I commented above in PKPSubmissionController in add/editContributor)

classes/author/Author.php
```diff
@@ -256,4 +272,72 @@ public function getLocalizedCompetingInterests(): ?string
...
+                    "name" => $affiliation->getLocalizedName()
+                ];
+            }
+        }
+
+        return $value;
+    }
+
+    /**
+     * Get the localized affiliations as an array.
```
bozana, line 315: localized affiliations names

plugins/importexport/native/pkp-native.xsd
```diff
@@ -187,6 +187,14 @@
 			<element name="givenname" type="pkp:localizedNode" minOccurs="1" maxOccurs="unbounded" />
 			<element name="familyname" type="pkp:localizedNode" minOccurs="0" maxOccurs="unbounded" />
 			<element name="affiliation" type="pkp:localizedNode" minOccurs="0" maxOccurs="unbounded" />
+            <element name="rorAffiliation" minOccurs="0" maxOccurs="unbounded">
+                <complexType>
+                    <sequence>
+                        <element name="ror" type="string" minOccurs="0" maxOccurs="unbounded" />
+                        <element name="name" type="pkp:localizedNode" minOccurs="0" maxOccurs="unbounded" />
```
bozana, line 194: Should the minOccurs for ror and name be 1 -- if there is a rorAffiliation, then both elements should exist, no?

plugins/importexport/native/filter/NativeXmlPKPAuthorFilter.php
```diff
@@ -106,13 +106,6 @@ public function handleElement($node)
                         }
                         $author->setFamilyName($n->textContent, $locale);
                         break;
-                    case 'affiliation':
-                        $locale = $n->getAttribute('locale');
-                        if (empty($locale)) {
-                            $locale = $publication->getData('locale');
-                        }
-                        $author->setAffiliation($n->textContent, $locale);
-                        break;
```
bozana, line 115: I believe this should stay, and the new element rorAffiliation needs to be considered as well, no?
Maybe the only thing is: if we have the affiliation element, if we would like to see if the ROR exists for it, like in the upgrade script :thinking:

bozana (commented):
Hi @GaziYucel, thanks a lot for the good work and improvements!
I left a few comments.
I believe, that having the ROR object in an affiliation would make it easier to deal with names in different locales, e.g. we will not need to worry about setting ROR default name to the affiliation name with the submission locale, and also getting getLocalizedAffiliations... should be easier. I am only not sure how that would work with the ui-library... Let me know what do you think...

classes/ror/Repository.php
```diff
@@ -0,0 +1,120 @@
...
+    public function getSchemaMap(): maps\Schema
+    {
+        return app('maps')->withExtensions($this->schemaMap);
+    }
+
+    /** @copydoc DAO::insert() */
+    public function add(Ror $ror): int
+    {
+        $id = $this->dao->insert($ror);
+        Hook::call('Ror::add', [$ror]);
```
bozana, line 82: Before the hook I think we need to get the ror (by ID) again, because else the code using the hook will not bet the ror ID

controllers/grid/settings/user/form/UserDetailsForm.php
```diff
@@ -182,7 +182,7 @@ public function initData()
             $data = [
                 'givenName' => $author->getGivenName(null), // Localized
                 'familyName' => $author->getFamilyName(null), // Localized
-                'affiliation' => $author->getAffiliation(null), // Localized
```
bozana, line 185: Hmmm... This file is used in the quickSubmit plugin. I do not think we can use the new affiliationField UI component there :-( I/we need to think how to solve it best...

controllers/grid/users/author/form/PKPAuthorForm.php
```diff
@@ -225,7 +225,7 @@ public function execute(...$functionParams)
         $author->setGivenName(array_map('trim', $this->getData('givenName')), null);
         $author->setFamilyName($this->getData('familyName'), null);
         $author->setPreferredPublicName($this->getData('preferredPublicName'), null);
-        $author->setAffiliation($this->getData('affiliation'), null); // localized
```
bozana, line 228: Note for myself: For this file as well. It is only used in quickSubmit plugin and we need to see how to solve the new affiliations there for 3.5

classes/affiliation/Repository.php
GaziYucel: There are many places where Repo::author()->add() and Repo::author()->edit() are called. If I don't do the check here, I expect that there will be unforeseen errors. That's why I added the extra check here.

api/v1/submissions/PKPSubmissionController.php
GaziYucel: Here it is called 'id', it is however the author id.

controllers/grid/users/author/form/PKPAuthorForm.php
```diff
@@ -225,7 +225,7 @@ public function execute(...$functionParams)
...
+        $author->setAffiliations($this->getData('affiliations'));
```
bozana, line 228: No need to change anything in this file -- it is deprecated.

classes/migration/upgrade/v3_5_0/I7135_CreateNewRorRegistryCacheTables.php
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('rors', function (Blueprint $table) {
+            $table->comment('Ror registry dataset cache');
+            $table->bigInteger('ror_id')->autoIncrement();
+            $table->string('ror')->nullable(false);
+            $table->string('display_locale', 28)->default('');
```
blesildaramirez, line 36: Hi @GaziYucel , I encountered an SQL error regarding this column having only a length of 28. In my case, the value being set is "no_lang_code: HUCH, HUCS, HUS, HYKS" which contains 35 characters, so my migration was unsuccessful for a specific record included in the records. Can you please check? Thanks!

GaziYucel: Hi @blesildaramirez. Thank you for pointing this out. 
The schema here is actually the same as other objects in OJS. That column should be like that. As mentioned in the other issue FieldAutosuggest, there was some error in the import.

GaziYucel:
Hi @bozana,

I made a change in the database for rors/ror_settings. 

I have made more performance tests on my shared hosting. Not sure if my hoster did some downgrades in my hosting, but I found it less performant as my tests before for lookups of rors through the UI. This is due to several sql "like" lookups in rors.ror / ror_settings.setting_value. 

I added an extra column to rors, namely "search_phrase", where I concatenate the ROR iD (without https://ror.org/) and names during the import. The lookup is now done through a single like only on this column. 
With this change, I see a performance bump of 3 - 4 times (before 1,5s > after 0.2 - 0.5ms). The other side of the coin is that the table rors have grown from 20 to 30 MB.

classes/ror/Ror.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function getName(?string $locale = null): mixed
+    {
+        return $this->getData('name', $locale);
+    }
+
+    /**
+     * Get cleaned name, e.g. remove NO_LANG_CODE, set required locale
+     */
+    public function getNameCleaned(?string $requiredLocale = null): mixed
```
GaziYucel, line 89: I added this method to return the names with 'no_lang_code' cleaned. This is used in Affiliation::fromRow. I find this cleaner than mixing this in Affiliation Object.

bozana:
@asmecher, if I make ask you about that performance issue @GaziYucel mentions above -- what would be a good way to proceed/test/solve... ?
Thanks a lot!

asmecher:
@bozana and @GaziYucel, would a partial index on the `setting_value` solve the problem? I'd prefer this if it's feasible, as it doesn't require us to duplicate the data. We use this approach with `publication_settings` e.g. so that titles can be searchable.

https://github.com/pkp/pkp-lib/blob/c5ebc057b122ca01fa174e156734fdc7ae1dcb91/classes/migration/install/SubmissionsMigration.php#L90..L96

bozana:
Merged from here https://github.com/pkp/pkp-lib/pull/10880.
Thus closing

Issue #10465, opened 2024-09-23
improve GET submissions/{submissionId}/decisions endpoint


jardakotesovec:
**Describe the issue**
While triaging https://github.com/pkp/pkp-lib/issues/10360, I came to conclusion, that for listing actual recommendations from recommend only editors, we could use the /decision endpoint.  But there is couple of things that would make it nice to work with.

Useful additions would be:
- add query param to being able to filter by decision type (multiple values), to get only decisions I am interested in
- add query param to being able to filter by the editorId, if I am for example interested only in my decisions I could use that. Ideally to support also multiple editorIds
- add query param to being able to filter by reviewRoundId 
- add query param to being able to filter by stageId
- sorting by the dateDecision - for me most useful would be to sort it so the first are the newest decisions. 

**What application are you using?**
 OJS, OMP or OPS version 3.5

ewhanson:
Hey @taslangraham, could you have a look at adding the queries params under useful additions above? The filters decision type and editor ID already exist and one would need to be added for the `dateDecision` in addition to the changes in the API call. Let me know if you have any questions!

jardakotesovec:
@ewhanson Oh great, I was looking to https://docs.pkp.sfu.ca/dev/api/ojs/3.4#tag/Submissions-Decisions  and did not check  in source code if there are query params available. Would be possible to also add existing query params to the docs?

jardakotesovec:
Added also filtering by reviewRoundId and stageId to the issue description, but these are less important if that would be hassle.

taslangraham:
Will take a look

ewhanson:
Hey @jardakotesovec, the filters I mentioned are in the collector but not currently accessible through the API endpoints, though that should make adding them to the API pretty straightforward.

taslangraham:
Ready for review @ewhanson 

- pkp-lib: https://github.com/pkp/pkp-lib/pull/10469/files
- ojs: https://github.com/pkp/ojs/pull/4447/files
- omp: https://github.com/pkp/omp/pull/1707/files (submodule update)
- ops: https://github.com/pkp/ops/pull/777/files (submodule update)

ewhanson:
Thanks @taslangraham. Just one minor issue in a doc comment. Feel free to merge after updating it.

taslangraham:
All merged! Thanks for te review @ewhanson

PR #10469, opened 2024-09-26
pkp/pkp-lib#10465 Improve GET submissions/{submissionId}/decisions endpoint

taslangraham:
For pkp/pkp-lib#10465

### DIFF
--- a/classes/decision/Collector.php
+++ b/classes/decision/Collector.php
@@ -33,6 +33,17 @@ class Collector implements CollectorInterface
     public ?array $stageIds = null;
     public ?array $submissionIds = null;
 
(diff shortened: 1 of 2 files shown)


classes/decision/Collector.php
```diff
@@ -129,6 +140,21 @@ public function filterBySubmissionIds(?array $submissionIds): self
         return $this;
     }
 
+    /**
+     * Order the results
+     *
+     * Results are ordered by the date_decided column by default.
+     *
+     * @param string $column Name of column to order results by.
```
ewhanson, line 148: The doc comment type should be nullable to match the argument type below.

ewhanson (changes requested):

Issue #10468, opened 2024-09-25
[OJS 3.4] co authors receive several e-mails when a decision is notified


forgive38:
the problem here is that the mail is sent each time a user is put in “to”, but mailable->to() adds the recipient to the “to” list, so the mail is sent each time with the previous recipient list plus the current coauthor.

https://github.com/pkp/pkp-lib/blob/97635b0ca0c6630e8792ac6aa4b7971b433924b9/classes/decision/types/traits/NotifyAuthors.php#L97

Can't we just add the co-authors in “cc”, so there's no need for a specific template, and the corresponding author knows that the co-authors have been notified? 

Thank you

asmecher:
@Vitaliy-1, could you have a look at this?

mpbraendle:
See this forum post: https://forum.pkp.sfu.ca/t/editor-decision-sent-multiple-times-to-primary-contact/94668

Does https://github.com/forgive38/pkp-lib/commit/23f3575490f1cdc2a7e5c1589c569f918ec5fea4 really fix this?

Tribunal33:
@asmecher Fix is good for 3.5 and 3.4.  Before fix it sent and email from Editor to Original author which is established during the email composition section after editor selects an decision and then automated message to all the contributors with duplicate emails.  After the fix it still sends the mail from Editor to Original author but now only sends one additional email to all of the contributors excluding the original author.  Additionally, please wait for QA to close the ticket.

![Image](https://github.com/user-attachments/assets/9f574d01-5801-4f73-8de2-35280038e61a)

Vitaliy-1:
PRs

pkp-lib stable 3.4: https://github.com/pkp/pkp-lib/pull/11130
tests 3.4: https://github.com/pkp/ojs/pull/4761

With this PR, the main differences are:

- Emails are sent for each author separately
- It allows to derive email template variables from the author object

Vitaliy-1:
@Tribunal33, do you have time to test this?

PR #10479, opened 2024-09-27
pkp/pkp-lib#10468 resolve: co authors receive several e-mails when a …

forgive38:
…decision is notified
pkp/pkp-lib#10468

### DIFF
--- a/classes/decision/types/traits/NotifyAuthors.php
+++ b/classes/decision/types/traits/NotifyAuthors.php
@@ -94,8 +94,10 @@ trait NotifyAuthors
                     continue;
                 }
                 $mailable->to($author->getEmail(), $author->getFullName());
-                Mail::send($mailable);
-            }
+	    }
+            if ( sizeof($mailable->to) > 0 ) {
+		    Mail::send($mailable);
+	    }
         }
     }
 


nwoodward:
I tested this PR, and the authors are still getting several emails. I wonder if it's because there are two sections of the `sendAuthorEmails()` function that seem to be doing the same thing. The function `Mail::send()` is called in both of them. Is only once necessary?

https://github.com/pkp/pkp-lib/blob/23f3575490f1cdc2a7e5c1589c569f918ec5fea4/classes/decision/types/traits/NotifyAuthors.php#L59-L65

and 

https://github.com/pkp/pkp-lib/blob/23f3575490f1cdc2a7e5c1589c569f918ec5fea4/classes/decision/types/traits/NotifyAuthors.php#L74-L101

asmecher:
Thanks, @forgive38! Sorry for the delay. @nwoodward, I've merged this -- it does clearly fix a bug in email sending -- but I think you're raising something different; I'll follow up on [the forum thread](https://forum.pkp.sfu.ca/t/editor-decision-sent-multiple-times-to-primary-contact/94668).

Issue #10415, opened 2024-09-13
[OJS 3_4] submission Url not role dependant in discussion notification


forgive38:
If a participant of a discussion has the author role, he should receive the url for authors, but he receives the url for "editors".
same case for discussions with reviewers.

I've tried to solve the problem by adding a param to Mailable, but I'm stuck.
I did a PR with a beginning but I'm not sure it's the right way to go...

I wonder if the simplest way wouldn't be to put a redirect in dispatcher for the main url “workflow/access/$submissionId”, depending on the current user's role: if he's only an author on the submission, we'll redirect him to authorDashboard and if he's only a reviewer, we'll redirect him to reviewer dashboard.

what do you think?

Thank you

asmecher:
@Vitaliy-1, could you look at this (and the associated PR)? Thanks, @forgive38!

asmecher:
@Vitaliy-1, could you take a look at this issue (and attached PR)?

PR #10484, opened 2024-10-01
pkp/pkp-lib#10415 redirect to dashboard according to user role

forgive38:
2nd solution for pkp/pkp-lib#10415 

redirect the user depending on the role he has on the submission, precisely if he is only author he is redirect to authorDashboard else he is redirected to "editor" dashboard 

(I think I prefer this solution, more universal than the first I propose)


forgive38:
mmmmh... idealy I must add redirect for reviewer too...

PR #10494, opened 2024-10-02
I7495 assignments

jardakotesovec:


classes/submission/DashboardView.php
```diff
@@ -46,7 +46,7 @@ class DashboardView
     const TYPE_REVIEWER_ASSIGNMENTS_ARCHIVED = 'reviewer-assignments-archived';
 
     // The number of submissions in the view
-    protected int $count;
+    protected int $count = 0;
```
Vitaliy-1, line 49: Are there specific cases when `count` isn't set and needs a default value?

Vitaliy-1 (approved):
Looks good! Important not to forget address stage assignments related data for submissions list

PR #10516, opened 2024-10-12
Add Group Start Date to groups API

withanage:
Extend the usr API to support  group start date.
e.g. "dateStart": "2024-10-11 05:25:10" under  groups

```json

{
    "_href": "http://localhost:8000/index.php/publicknowledge/api/v1/users/18",
    "affiliation": {
        "en": "University of Bologna",
        "fr_CA": ""
    },
    "apiKey": null,
    "apiKeyEnabled": null,
    "authStr": null,
    "billingAddress": null,
    "biography": {
        "en": "",
        "fr_CA": ""
    },
    "country": "IT",
    "dateLastEmail": null,
    "dateLastLogin": "2024-10-11 05:25:08",
    "dateRegistered": "2024-10-11 05:25:08",
    "dateValidated": null,
    "disabled": false,
    "disabledReason": null,
    "email": "ccorino@mailinator.com",
    "familyName": {
        "en": "Corino",
        "fr_CA": ""
    },
    "fullName": "Carlo Corino",
    "givenName": {
        "en": "Carlo",
        "fr_CA": ""
    },
    "groups": [
        {
            "id": 14,
            "name": {
                "en": "Author",
                "fr_CA": "Auteur-e"
            },
            "abbrev": {
                "en": "AU",
                "fr_CA": "AU"
            },
            "roleId": 65536,
            "showTitle": true,
            "permitSelfRegistration": true,
            "permitMetadataEdit": false,
            "recommendOnly": false,
            "dateStart": "2024-10-11 05:25:10"
        },
        {
            "id": 17,
            "name": {
                "en": "Reader",
                "fr_CA": "Lecteur-trice"
            },
            "abbrev": {
                "en": "Read",
                "fr_CA": "Lect"
            },
            "roleId": 1048576,
            "showTitle": true,
            "permitSelfRegistration": true,
            "permitMetadataEdit": false,
            "recommendOnly": false,
            "dateStart": "2024-10-11 05:25:08"
        }
    ],
    "id": 18,
    "inlineHelp": true,
    "interests": [],
    "locales": [],
    "mailingAddress": null,
    "mustChangePassword": null,
    "orcid": null,
    "orcidAccessDenied": null,
    "orcidAccessExpiresOn": null,
    "orcidAccessScope": null,
    "orcidAccessToken": null,
    "orcidIsVerified": null,
    "orcidRefreshToken": null,
    "orcidReviewPutCode": null,
    "phone": null,
    "preferredPublicName": {
        "en": "",
        "fr_CA": ""
    },
    "profileImage": null,
    "signature": {
        "en": "",
        "fr_CA": ""
    },
    "stageAssignments": [],
    "url": null,
    "userName": "ccorino"
}

```

### DIFF
--- a/classes/user/maps/Schema.php
+++ b/classes/user/maps/Schema.php
@@ -21,6 +21,7 @@ use PKP\security\Role;
 use PKP\services\PKPSchemaService;
 use PKP\stageAssignment\StageAssignment;
 use PKP\user\User;
+use PKP\userGroup\relationships\UserUserGroup;
 use PKP\workflow\WorkflowStageDAO;
 use Submission;
 
@@ -156,7 +157,7 @@ class Schema extends \PKP\core\maps\Schema
                         $userGroups = Repo::userGroup()->userUserGroups($user->getId(), $this->context->getId());
                         $output[$prop] = [];
                         foreach ($userGroups as $userGroup) {
-                            $output[$prop][] = [
+                           $output[$prop][] = [
                                 'id' => (int) $userGroup->getId(),
                                 'name' => $userGroup->getName(null),
                                 'abbrev' => $userGroup->getAbbrev(null),
@@ -165,6 +166,10 @@ class Schema extends \PKP\core\maps\Schema
                                 'permitSelfRegistration' => (bool) $userGroup->getPermitSelfRegistration(),
                                 'permitMetadataEdit' => (bool) $userGroup->getPermitMetadataEdit(),
                                 'recommendOnly' => (bool) $userGroup->getRecommendOnly(),
+                                'dateStart' => UserUserGroup::withUserId($user->getId())
+                                   ->withActive()
+                                   ->withUserGroupId($userGroup->getId())
+                                   ->pluck('date_start')->first()
                             ];
                         }
                     }


bozana:
Hi @withanage, because now a user can have several entries in the user_user_groups table -- those that already ended and the active one -- just to double check what you would exactly need. Do you need only the group the user is currently active in and that start date?
The API does not reflect yet the ended groups. Thus, just to double check if you would maybe also need also the ended ones?

And, also to double check: do you need it all via API or would for example Repo::userGroup work too/be sufficient?

classes/user/maps/Schema.php
```diff
@@ -165,6 +166,9 @@ protected function mapByProperties(array $props, User $user, array $auxiliaryDat
                                 'permitSelfRegistration' => (bool) $userGroup->getPermitSelfRegistration(),
                                 'permitMetadataEdit' => (bool) $userGroup->getPermitMetadataEdit(),
                                 'recommendOnly' => (bool) $userGroup->getRecommendOnly(),
+                                'dateStart' => UserUserGroup::withUserId($user->getId())
+                                   ->withUserGroupId($userGroup->getId())
+                                   ->pluck('date_start')->first()
```
bozana, line 171: This could give you the start date of an old user_user_group, that already ended.
So you would need to also use `->withActive()`.
The API currently only considers the active user groups:
The line above: 
`$userGroups = Repo::userGroup()->userUserGroups($user->getId(), $this->context->getId());`
returns only active user groups.

withanage:
@bozana  Thanks!


> just to double check what you would exactly need. Do you need only the group the user is currently active in and that start date?
yes, at the moment according to the specificaion, we need in the API  the start  date only.

> The API does not reflect the ended groups yet. Thus, just to double check if you would maybe also need the ended ones (which would change what API returns -- there could be several items for one group)?

We will only need the active users.  I will incorporeate  your suggestion with activeUsers

classes/user/maps/Schema.php
withanage: @bozana  I have changed the PR to get only the active users

PR #10536, opened 2024-10-21
[Propsal]customized page not found error

withanage:
### Required user-cases

- When invitation  is not deleted a customized message indicating the invitation is required.
-

### DIFF
--- a/classes/core/Dispatcher.php
+++ b/classes/core/Dispatcher.php
@@ -304,10 +304,10 @@ class Dispatcher
     /**
      * Handle a 404 error (page not found).
      */
-    public static function handle404()
+    public static function handle404(string $message = "404 Not Found"): void
     {
         header('HTTP/1.0 404 Not Found');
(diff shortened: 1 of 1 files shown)


classes/core/Dispatcher.php
```diff
@@ -304,10 +304,10 @@ public function _cacheContent(string $contents): string
     /**
      * Handle a 404 error (page not found).
      */
-    public static function handle404()
+    public static function handle404(string $message = "404 Not Found"): void
     {
         header('HTTP/1.0 404 Not Found');
-        echo "<h1>404 Not Found</h1>\n";
+        echo "<h1>" . $message . "</h1>\n";
```
asmecher, line 310: `$message` should be wrapped with `htmlspecialchars`.

withanage: Thanks @asmecher , I have add the `htmlspecialchars` function and  merged.

Issue #10476, opened 2024-09-26
Ability to set casts/data types to multilingual attributes in Eloquent Model


Vitaliy-1:
**Describe the bug**
As part of #10328 multilingual attributes are introduced. For now, it's impossible to set a specific cast for the as per Laravel documentation:
```php
    protected function casts(): array
    {
        return [
            'title' => 'string',
            'description' => 'string',
        ];
    }
```
because multilingual values actually are arrays

Apart of that, consider also if ability to directly add multilingual value in the default locale is beneficial, e.g.:
```php
Announcement::create([
    "assocType" => "256",
    "assocId" => "1",
    "typeId" => null,
    "title" => "Test Announcement 201",
    "description" => "<p>Test Announcement 201</p>",
    "descriptionShort" => "<p>Test Announcement 201</p>",
    "someData" => "my new data"
]);
```
see: https://github.com/pkp/pkp-lib/pull/10382#discussion_r1764336625

touhidurabir:
Merged as part of https://github.com/pkp/pkp-lib/issues/10292

PR #10542, opened 2024-10-22
pkp/pkp-lib#10292 Fix update setting values for a Model

Vitaliy-1:
Close after issue is resolved

### DIFF
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -98,7 +98,7 @@ class SettingsBuilder extends Builder
 
     /**
      * Insert the given attributes and set the ID on the model.
-     * Overrides Builder's method to insert setting values for a models with
+     * Overrides Builder's method to insert setting values for a models with settings
      *
      * @param  string|null  $sequence
      *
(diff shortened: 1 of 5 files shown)


classes/core/SettingsBuilder.php
```diff
@@ -72,22 +72,33 @@ public function update(array $values)
...
-
-        $sql = $this->buildUpdateSql($settingValues, $us, $newQuery);
+        $schema = null;
+        if (!$this->getSchemaName()) {
+            $casts = $this->model->getCasts();
+            foreach ($this->model->getSettings() as $settingName) {
+                $schema['properties'][$settingName]['multilingual'] = in_array($settingName, $this->model->getMultilingualProps());
+
+                if (array_key_exists($settingName, $casts)) {
+                    $type = $casts[$settingName];
```
touhidurabir, line 82: This is where we may need to adapt a bit when a custom cast is going to apply like I have proposed in https://github.com/pkp/pkp-lib/issues/10476 . Though I am not sure without some testing .

touhidurabir:
@Vitaliy-1 this look good . Though I did not get a chance to test it with non schema based entity and when a custom cast is applied as mentioned at https://github.com/pkp/pkp-lib/pull/10542#discussion_r1841514212 . But I say lets merge it as it works well with schema based entity and if required, I will adapt (or make changes) for non schema based entity as I am working on pkp/pkp-lib#10292 .

PR #10557, opened 2024-10-24
User search API, return all current and future users, when search status=all

withanage:
Specificaton:

- search for users with  future roles, if the status is set to all and the user  role is in current context and the role start date is in future.
- 

Example

http://localhost:8000/index.php/publicknowledge/api/v1/users?searchPhrase=admin&status=all


ewhanson (approved):
Looks good to me. Thanks, @withanage!

withanage:
Thaks for reviewing @ewhanson !

PR #10576, opened 2024-11-01
pkp/pkp-lib#9658 user access table

ipula:
pkp/pkp-lib#9658

### DIFF
--- a/pages/invitation/InvitationHandler.php
+++ b/pages/invitation/InvitationHandler.php
@@ -21,11 +21,13 @@ use APP\core\Request;
 use APP\facades\Repo;
 use APP\handler\Handler;
 use APP\template\TemplateManager;
+use PKP\context\Context;
 use PKP\core\PKPApplication;
 use PKP\facades\Locale;
 use PKP\invitation\core\enums\InvitationAction;
 use PKP\invitation\core\Invitation;
 use PKP\invitation\stepTypes\SendInvitationStep;
+use PKP\userGroup\relationships\UserUserGroup;
 
 class InvitationHandler extends Handler
 {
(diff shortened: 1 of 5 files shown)


defstat:
@ipula is this accompanied with a ui-library PR?

ipula:
> @ipula is this accompanied with a ui-library PR?
Yes https://github.com/pkp/ui-library/pull/437

pages/invitation/InvitationHandler.php
```diff
@@ -229,6 +229,92 @@ public function invite(array $args, Request $request): void
...
+     * Edit user using user access table action
+     * @param $args
+     * @param $request
+     * @return void
+     * @throws \Exception
+     */
+    public function editUser($args, $request): void
+    {
+        $invitation = null;
+        $invitationPayload =[];
```
defstat, line 242: I am wondering if you can use the already existing `UserRoleAssignmentInvitePayload` object here, in order to avoid the many key strings. 

What do you think?

defstat (changes requested):
@ipula thanks, just one question on this, otherwise I think it is ok

pages/invitation/InvitationHandler.php
ipula: I checked but some values are different. For example I need `orcid` but in `UserRoleAssignmentInvitePayload` has `userOrcid` also same with the `country`.

defstat: @withanage, @ipula if we can't change those variable names in the front-end, we could go forward with merging it.

ipula: @defstat Issue like we are using same user invitation components. I have to change the code that already QA and merged

Issue #10571, opened 2024-10-30
Implement role check for mailables


taslangraham:
As part of the work that needs to be done for https://github.com/pkp/pkp-lib/issues/10403, we need to implement role checks for mailables and email templates:

-  **Role Designation for Email Templates**: Enable editors to assign specific user groups when creating or editing an email template. This feature should be restricted to mailables associated with the submission workflow.
    - [ ]   This will require UI work on the Edit/Add Template side modal to support selecting user groups. Maybe something like a list of checkboxes or a multi-select dropdown might be enough. @Devika008 Tagging here for your input
    
- **Role Designation for mailables** Allow mailables to be assigned to specific roles, ensuring that templates are accessible only to users with relevant roles.

- **Template-Specific Restrictions**: Users should only access templates explicitly assigned to their group within each mailable. For example, if a mailable is accessible to both editors and authors, but a specific template is restricted to editors, then only editors should see that template, and authors should not have access.

**PRs**
ojs: https://github.com/pkp/ojs/pull/4745
omp: https://github.com/pkp/omp/pull/1899
ops: https://github.com/pkp/ops/pull/916
pkp-lib: https://github.com/pkp/pkp-lib/pull/11090
ui-lib: https://github.com/pkp/ui-library/pull/559

taslangraham:
@asmecher @ewhanson,

I’ve added draft PRs for OJS, pkp-lib, and UI-Library in the [description](https://github.com/pkp/pkp-lib/issues/10571#issue-2624382729) and would appreciate any early feedback. If you have a chance, please take a look at the approach and let me know your thoughts.

Note: In discussions, @asmecher highlighted the need to allow any user group to be assignable to a template, not just those linked to $fromRoleIds on the associated mailable. As such no mailable specific checks are included; instead all checks are scoped to the template itself.

ewhanson:
Thanks, @taslangraham. It's looking good. Lots going on here!

Mostly a few minor requests, but I would like to take a look at the field components together. I think we can get away with using the existing `FieldOptions` component, which will probably remove the need for any changes in the UI library as well as reduce the number of places that state will need to be tracked. 

We can take a look together on our next call, but feel free to have a look ahead of time. Since none of user groups will be changing dynamically on the frontend, we can do all this setup in PHP and just pass the options through the form setup functions. Let me know if you have any questions.

taslangraham:
[...]

PR #10581, opened 2024-11-05
pkp/pkp-lib#10571 limit email template access by user groups

taslangraham:
For pkp/pkp-lib#10571


taslangraham:
@ewhanson Just an update -  I've made a new [commit ](https://github.com/pkp/pkp-lib/pull/10581/commits/098ea45a9d20fdabe3b0746fc768eca59e364768)since I last requested your review. 

This handles cases where a template was restored to default settings, and also cleans up template access when a template is deleted

classes/emailTemplate/DAO.php
```diff
@@ -281,6 +285,20 @@ public function installEmailTemplates(
                     $this->installAlternateEmailTemplates($contextId, $attrs['key']);
                 }
             }
+
+            if ($recordTemplateGroupAccess) {
+                // Default to true if `isUnrestricted` is not set.
+                $isUnrestricted = $attrs['isUnrestricted'] ?? '1';
```
ewhanson, line 291: I think this check is enough to cover any plugin-related issues or the email template XML not having this flag. I don't think you need the `$recordTempalteGroupAccess` parameter at all.

classes/components/forms/emailTemplate/EmailTemplateForm.php
```diff
@@ -46,6 +48,14 @@ public function __construct(string $action, array $locales)
                 'isMultilingual' => true,
                 'toolbar' => 'bold italic superscript subscript | link | blockquote bullist numlist',
                 'plugins' => 'paste,link,lists',
+            ]))->addField(new FieldEmailTemplateUnrestricted('isUnrestricted', [
```
ewhanson, line 51: I think a lot more of the setup logic for the Vue components can be done in PHP so that these can both be done with existing `FieldOptions` components. I mentioned this in the `ui-library` PR as well and we can chat about it more in our next one-on-one (probably easier to share screen and code together than writing it all out here).

classes/emailTemplate/DAO.php
```diff
@@ -235,13 +235,17 @@ public function getMainEmailTemplatesFilename()
...
+     * where the `email_template_user_group_access` table may not exist at the time of execution.
      *
      */
     public function installEmailTemplates(
         string $templatesFile,
         array $locales = [],
         ?string $emailKey = null,
-        bool $skipExisting = false
+        bool $skipExisting = false,
+        $recordTemplateGroupAccess = false
```
ewhanson, line 248: I mention this directly below, but I don't think this optional parameter will be necessary. I think the assumption of "unrestricted if no additional data provided" below is enough.

classes/emailTemplate/EmailTemplateAccessGroup.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
ewhanson, line 1: Needs file docblock.

classes/migration/upgrade/V3_6_0/I10403_EmailTemplateUserGroupAccess.php
```diff
@@ -0,0 +1,74 @@
+<?php
```
ewhanson, line 1: Needs file docblock.

classes/migration/upgrade/V3_6_0/PreflightCheckMigration.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_5_0/PreflightCheckMigration.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
```
ewhanson, line 6: Should be 2024 here and below.

classes/emailTemplate/EmailTemplateAccessGroup.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace PKP\emailTemplate;
+
+use Eloquence\Behaviours\HasCamelCasing;
+use Eloquence\Database\Model;
```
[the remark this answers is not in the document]
taslangraham: This is an error on my part. It should be extending the `Illuminate\Database\Eloquent\Model`.

schemas/emailTemplate.json
```diff
@@ -68,6 +68,37 @@
...
+			"description": "List of User Groups that can be assigned to the template",
+			"apiSummary": true,
+			"items": {
+				"type": "object",
+				"properties": {
+					"id": {
+						"type": "integer",
+						"apiSummary": true
+					},
+					"name": {
```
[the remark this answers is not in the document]
taslangraham: That's an error on my part. It should be a string.

classes/mail/Repository.php
```diff
@@ -206,6 +215,17 @@ protected function isMailableConfigurable(string $class, Context $context): bool
...
 
+
+    /**
+     * Check if the templates of a given mailable can be assigned to specific groups
+     *
+     * @param Mailable|string $mailable - Mailable class or qualified string referencing the class
+     */
+    public function isGroupsAssignableToTemplates(Mailable|string $mailable): bool
+    {
+        return !in_array(Mailable::FROM_SYSTEM, $mailable::getFromRoleIds());
```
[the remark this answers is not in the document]
taslangraham: Yes, if $mailable is string variable, with the [fully qualified name](http://php.net/manual/en/language.namespaces.rules.php), we will have a reference to the class which you can create an instance of or call static methods/properties on.

https://stackoverflow.com/questions/534159/instantiate-a-class-from-a-variable-in-php
https://www.php.net/manual/en/language.namespaces.dynamic.php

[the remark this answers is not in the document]
ewhanson: Cool, had not run into that before!

classes/emailTemplate/DAO.php
ewhanson: For posterity, we talked about it and decided to keep thing `$recordTemplateGroupAccess` parameter as described here to avoid any issues as part of the upgrade process.

Issue #1660, opened 2016-08-04
Customizable Reviewer Recommendations


stranack:
**Describe the problem you would like to solve**
Reviewers are presented with a list of recommendations they can choose from, such as accept, decline, request revisions, etc. This list is not appropriate for all journals. Some would like to change the list to reflect their review practices.

**Describe the solution you'd like**
Somewhere in the settings where review recommendations can be added, edited or removed.

**Who is asking for this feature?**
Requests on the forum, PKP's publishing services, and institutional partners.

**Additional Information**
- It should not be possible to delete reviewer recommendations that are used in exisiting reviews, but it should be possible to decommission them so they cannot be used by future reviewers.

# Specs Update - Friday, September 27th, 2024

## Workflows Affected by This Change

- [ ] Settings > Workflow > Review - Here a new tab will be added here for customisable review recommendations 

## Detailed Specs

We can add a section in Settings > Workflow > Review > Review Recommendations. This would include a default list of review recommendations that JMs and JEs can easily modify, add to, or activate/deactivate. I left out the option to delete recommendations, as the activate/deactivate feature offers a good workaround while still keeping a record of past recommendations.

**Video Recording:** https://youtu.be/cYb3U1FRH_0
**The Designs can be found here:** https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?node-id=8862-8486&t=sf2uoFGq6lBMtfCU-4
**Prototype:** https://www.figma.com/proto/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?page-id=7200%3A7069&node-id=8862-8487&node-type=frame&viewport=1704%2C-15906%2C0.23&t=mAQY2jdYCKArumzF-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8862%3A8487

### UI Development Guidance

Components to be used would be (all can be found in storybook):

- Listing: Table + DropdownActions + Badge (that need new grey variant, if you need help to extend it let me know) + Button
- Form: FieldText + FieldSelect // and don't get confused by the fact it looks different than on designs. This is how our forms will look in future.
- Form should be created on PHP side as we don't have tooling to do it client side.

In general just use the latest stuff - composition API, tailwindCSS.

**PRs (draft)**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10583
[...]

PR #10583, opened 2024-11-06
pkp/pkp-lib#1660 Customizable Reviewer Recommendations

touhidurabir:
for pkp/pkp-lib#1660


api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,218 @@
+<?php
+
+/**
+ * @file api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```
asmecher, line 4: This handler doesn't appear to check that the recommendation being edited/deleted exists in the current context! The manage of one journal can edit and delete recommendations that belong to another.

classes/submission/reviewer/recommendation/Repository.php
```diff
@@ -0,0 +1,119 @@
...
+
+class Repository
+{
+    /**
+     * Get default recommendation seed data mapped as value => localeKey
+     */
+    public function getDefaultRecommendations(): array
+    {
+        return [
+            1 => 'reviewer.article.decision.accept', // SUBMISSION_REVIEWER_RECOMMENDATION_ACCEPT
```
asmecher, line 31: I think it should be possible to eliminate the `value` column and just rely on a database ID. It would reduce the use of magic numbers, and mean that the options that ship with OJS are treated just like user-created options. To do this, `review_assignments.recommendation` would also need to be converted to a `recommendation_id` foreign key to the `reviewer_recommendations` table.

classes/userGroup/Repository.php
```diff
@@ -171,7 +171,7 @@ public function getArrayIdByRoleId(int $roleId, ?int $contextId = null): array
      *
      * @return LazyCollection<int, UserGroup>
      */
-    public function getByRoleIds(array $roleIds, int $contextId, ?bool $default = null): LazyCollection
+    public function getByRoleIds(array $roleIds, ?int $contextId, ?bool $default = null): LazyCollection
```
asmecher, line 174: Is this change intentional/necessary?

asmecher (changes requested):

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: Isn't that handled by `ContextAccessPolicy` defined in `authorize` so that only those have access to context can access that api end point ? And we are restricting the roles to `admin, manager and editor` . 

Or perhaps I am misunderstanding something ?

classes/submission/reviewer/recommendation/Repository.php
touhidurabir: @asmecher This is something I have been pondering for some time . What you have mentioned is one way to do it that I also thought about. But then we have some complex upgrade process as default values `1-6` is not available anymore when adding to table as primary `id` will be sequential . So if have 2 contexts , context-1 will have 1-6 and context-2 will have 7-12, so we have to update every review assignments for context-2 . Now for a large installation with many context and many review assignment, upgrade is even more time consuming
1. first need to create recommendations for context
2. map to value `1-6` with recommendations id
3. update all review assignments for that context . 

Also need a way to mark the default ones so that on local change, we can add new translations for the default ones (that is not difficult)

The current implementation does not require all these complex process and allow minimum changes to introduce this feature . Thats why I went with this path . But you are right about `magic numbers` and that can be the major motivation alter it other way as you have described .

classes/userGroup/Repository.php
touhidurabir: Truthfully I don't actually remember . I will test and confirm .

classes/submission/reviewer/recommendation/Repository.php
asmecher: If we were adding this as a brand new review feature, we would definitely use a foreign key rather than a `value` field. The only time we need to match a default is when we are adding a new language, and that's a boundary case (which I think is already matching on translation text, not value, right?) -- and that's a rare/boundary case already, so not justifying a different storage method.

The upgrade will be quick, even on large installs:

1. Add a `recommendation_id` column to `review_assignments`, default null, with a FK constraint
2. For each journal...
  1. Introduce default 6 options to `reviewer_recommendations`, keeping a PHP-side map from `recommendation` constant to `recommendation_id`
  2. For each default recommendation option (count=6),
    3. Update any review assignments to use the appropriate ID
3. Drop the `recommendation` column from `review_assignments`

That's a fixed number of queries per journal -- no need to e.g. execute anything e.g. per review assignment or per submission.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
asmecher: Yes, but you're accepting any recommendation ID as user-supplied data -- and that recommendation might belong to another journal that the manager or editor doesn't manage.

asmecher (changes requested):

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -233,23 +230,21 @@ public function setReviewRoundId($reviewRoundId)
...
     }
 
     /**
-     * Set reviewer recommendation.
-     *
-     * @param string $recommendation
+     * Set reviewer recommendation id
      */
-    public function setRecommendation($recommendation)
+    public function setRecommendationId(int $recommendationId)
```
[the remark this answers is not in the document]
touhidurabir: Yes, the column is nullable but it's by default . If we need to assign/set a recommendation id , then it should be a real existing reviewer_recommendation_id, no point to set null which is by default null via this method . And in the migration the relation is set to `set null` on delete of a recommendation .

classes/core/SettingsBuilder.php
```diff
@@ -263,6 +263,20 @@ public function whereIn($column, $values, $boolean = 'and', $not = false)
...
+    /**
+     * Add a "where not in" clause to the query.
+     * Overrides Illuminate\Database\Query\Builder to support settings in select queries
+     *
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
+     * @param  mixed  $values
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereNotIn($column, $values, $boolean = 'and')
```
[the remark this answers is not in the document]
touhidurabir: Not actually , see the 4th param when calling the `whereIn`  from the `whereNotId` which is set to `true` , which in general reverse of `whereIn` . This allow us to use the useful whereNotIn bit more easily like `->whereNotIn('columns_name', [])` instead `->whereIn('columns_name', [], 'and', true)` .

classes/migration/install/ReviewerSuggestionsMigration.php
```diff
@@ -106,8 +106,8 @@ public function up(): void
     public function down(): void
     {
         Schema::disableForeignKeyConstraints();
-        Schema::drop('reviewer_suggestions');
         Schema::drop('reviewer_suggestion_settings');
+        Schema::drop('reviewer_suggestions');
```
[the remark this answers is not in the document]
touhidurabir: Not really , just best practice to have the related table drop first and then the main one which I forgot to do for reviewer suggestion . I have reverted this change .

classes/submission/reviewer/recommendation/RecommendationOption.php
```diff
@@ -0,0 +1,31 @@
...
+ * @brief Enum class to define how the recommendation status filter to apply
+ */
+
+namespace PKP\submission\reviewer\recommendation;
+
+enum RecommendationOption
+{
+    case ALL;
+    case ACTIVE;
+    case DEACTIVE;
```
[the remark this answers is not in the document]
touhidurabir: > But I'm not sure this enumeration is worth the extra code, when we could just as easily use the ?bool type instead with constants defined in the ReviewerRecommendation class.

True, it's just seems a bit unreadable by what we mean by true, false and null .

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -829,37 +824,18 @@ public function getWeeksDue()
...
-            self::SUBMISSION_REVIEWER_RECOMMENDATION_SEE_COMMENTS => 'reviewer.article.decision.seeComments'
-        ];
-        return $reviewerRecommendationOptions;
-    }
-
     /**
      * Return a localized string representing the reviewer recommendation.
      */
-    public function getLocalizedRecommendation()
+    public function getLocalizedRecommendation(?Context $context = null): string
```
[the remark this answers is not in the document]
touhidurabir: > This should be able to return a null (for cases where the recommendation hasn't been set)

This return an empty string if no recommendation yet as it's return a localized data which is being used many places (in template and translation replacement) . 

> that condition should be checked below, as otherwise I think there will be an error when trying to index an array by `null`

it's being check as `$options[$this->getReviewerRecommendationId()] ?? ''` .  Also previously we had it like 
```
$options = [
    "" => "choose one",
    1 => "accept",
    2 => "decline"
];
```
so in case of `$this->getReviewerRecommendationId()` is `null`, we used to get default `choose one` . Now that we plan to remove it from recommendation list as per https://github.com/pkp/pkp-lib/pull/10583#discussion_r2017008214 , this check will still handle array index by null case .

classes/core/SettingsBuilder.php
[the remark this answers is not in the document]
asmecher: Ah, I totally missed the 4th parameter.

classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -233,23 +230,21 @@ public function setReviewRoundId($reviewRoundId)
     }
 
     /**
-     * Get reviewer recommendation.
+     * Get reviewer recommendation id
      *
      * @return string
```
asmecher, line 235: The `@return` is incorrect after this change. In this case it doesn't add anything, now that there's a typehint, so it should just be removed.

classes/submission/reviewer/recommendation/Repository.php
```diff
@@ -0,0 +1,169 @@
...
+
+            $title = $suggestion->title;
+            $title[$localeToAdd] = $localeToAddTranslation;
+            $suggestion->update(['title' => $title]);
+        }
+    }
+
+    /**
+     * Get an associative array matching reviewer recommendation id mapped to localized title.
+     * (Includes default '' => "Choose One" string.)
```
asmecher, line 129: This comment isn't accurate.

asmecher (changes requested):
Just a couple minor tweaks, thanks!

classes/submission/reviewer/form/PKPReviewerReviewStep3Form.php
```diff
@@ -166,9 +172,11 @@ public function execute(...$functionParams)
         // Persist the updated review assignment.
         Repo::reviewAssignment()->edit($reviewAssignment, [
             'dateCompleted' => Core::getCurrentDate(), // Mark the review assignment as completed.
-            'recommendation' => (int) $this->getData('recommendation'), // assign the recommendation to the review assignment, if there was one.
+            'reviewerRecommendationId' => $this->getData('reviewerRecommendationId'), // assign the recommendation to the review assignment, if there was one.
```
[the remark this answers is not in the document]
touhidurabir: This is added to `APP\submission\reviewer\form\ReviewerReviewStep3Form` which extends `PKPReviewerReviewStep3Form` as https://github.com/pkp/ojs/pull/4505/files#diff-7219eb555c9953733a864d839235582030193afbaf49b6efdcefc40460214a79R34-R43 .

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -1254,7 +1257,7 @@ protected function createMail(Mailable $mailable, string $emailBody, EmailTempla
     /**
      * Get the proper reviewer from instance
      */
-    protected function getReviewerFrom(int $selectionType, Request $request = null): ReviewerForm
+    protected function getReviewerFrom(int $selectionType, ?Request $request = null): ReviewerForm
```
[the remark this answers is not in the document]
touhidurabir: it's a typo .

PR #10585, opened 2024-11-06
add start date validation

withanage:
Hi @defstat ,

start date should be mandatory in the API. Please have a look! 

image (1)


defstat (approved):
@withanage thanks, have no comments

Issue #10620, opened 2024-11-18
New Editorial Board Member assistant role


bozana:
Add a new role with the name "Editorial Board Member", with 'Assistant' permission level and without any "Stage Assignment", that will be considered for the display on the Editorial Masthead and Editorial History page.

PRs:
- https://github.com/pkp/pkp-lib/pull/10623
- https://github.com/pkp/ojs/pull/4527
- https://github.com/pkp/omp/pull/1770
- https://github.com/pkp/ops/pull/814

bozana:
@asmecher, could you please review the PRs above? 
Thanks a lot!

bozana:
Ah, sorry @asmecher, I need to implement also the upgrade script that will install the new role for old journals... So please wait a little bit... :-)

bozana:
Hi @asmecher, now the PRs are ready for review. Could you review it?
Thanks a lot!

PR #10623, opened 2024-11-19
pkp/pkp-lib#10620 introduce new Editorial Board Member assistant role

bozana:
s. https://github.com/pkp/pkp-lib/issues/10620

### DIFF
--- a/classes/migration/upgrade/v3_5_0/I10620_EditorialBoardMemberRole.php
+++ b/classes/migration/upgrade/v3_5_0/I10620_EditorialBoardMemberRole.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_5_0/I10620_EditorialBoardMemberRole.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I10620_EditorialBoardMemberRole
+ *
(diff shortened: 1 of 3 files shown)


classes/migration/upgrade/v3_5_0/I10620_EditorialBoardMemberRole.php
```diff
@@ -0,0 +1,81 @@
...
+                'masthead' => true,
+            ]);
+            $userGroupId = (int) DB::getPdo()->lastInsertId();
+
+            DB::table('user_group_settings')->insert([
+                ['user_group_id' => $userGroupId, 'setting_name' => 'nameLocaleKey', 'setting_value' => $nameKey],
+                ['user_group_id' => $userGroupId, 'setting_name' => 'abbrevLocaleKey', 'setting_value' => $abbrevKey],
+            ]);
+
+            foreach ($installedLocales as $locale) {
```
asmecher, line 60: This is going to result in a lot of users who move to 3.5.0 immediately (before all of our translations are 100% complete) getting untranslated locale keys installed into the database. I'd rather untranslated names were just not provided in any but the required (primary) locale; that way they'll fall back properly rather than showing `##some.untranslated.key##`.

bozana: I considered this here, in the upgrade script, but it will be then different when someone installs the new 3.5 version anew. But this is maybe OK so, or? -- Those strings are editable, so if ## appears, the journal manager will eventually easier noticed that it needs to be edited.

PR #10635, opened 2024-11-26
Activity log not working

marcbria:
Old issue fixed by @mpbraendle in 2022 but never committed. 
In short, users can't see the "activity log" of 3.2 submissions. 
Patched and tested in 3.3.0-19 without any collateral effect.

Full description here:
https://forum.pkp.sfu.ca/t/activity-log-not-work/71271/9

### DIFF
--- a/classes/services/PKPSubmissionFileService.inc.php
+++ b/classes/services/PKPSubmissionFileService.inc.php
@@ -777,7 +777,7 @@ class PKPSubmissionFileService implements EntityPropertyInterface, EntityReadInt
 				}
 				$queryDao = DAORegistry::getDAO('QueryDAO'); /* @var $queryDao QueryDAO */
 				$query = $queryDao->getById($note->getAssocId());
-				return $query ? $query->getStageId() : null;
+				return $reviewRound ? $reviewRound->getStageId() : null;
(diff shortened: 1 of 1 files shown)


classes/services/PKPSubmissionFileService.inc.php
```diff
@@ -777,7 +777,7 @@ public function getWorkflowStageId($submissionFile) {
 				}
 				$queryDao = DAORegistry::getDAO('QueryDAO'); /* @var $queryDao QueryDAO */
 				$query = $queryDao->getById($note->getAssocId());
-				return $query ? $query->getStageId() : null;
+				return $reviewRound ? $reviewRound->getStageId() : null;
```
kaitlinnewson, line 780: Based on the original forum post, I think this change should be on line 771 and not here.

kaitlinnewson (changes requested):
Thanks @marcbria! Just one small change needed here.

marcbria:
Sorry a lot Kaitlin. I'm ashamed. 
Change made.

kaitlinnewson (approved):

jonasraoni (approved):

Issue #10489, opened 2024-10-01
Add a form field in the profile section that allows users to customize the initials displayed in their profile avatars.


Devika008:
**Describe the problem you would like to solve**
Using initials as avatars adds a personal touch to user profiles, giving them a unique identity without the need for personal photos. This also addresses privacy concerns related to using images. However, some users may prefer to adjust the automatically generated initials to better suit their preferences.

**Describe the solution you'd like**
Add a form field under **User Profile > Identity** that lets users customize their initials. They should be able to enter only 2 characters, no more.

taslangraham:
I've added a PR for this @ewhanson. Please take a look.
**pkp-lib** - https://github.com/pkp/pkp-lib/pull/10639

**submodule updates**
 ojs - https://github.com/pkp/ojs/pull/4533
ops - https://github.com/pkp/ops/pull/818
omp - https://github.com/pkp/omp/pull/1776


There's an avatar [component ](https://github.com/pkp/ui-library/tree/main/src/components/UserAvatar)in the UI library currently used to display initials based on the user's name. This would need to be updated, in a separate issue,  to accept custom initials

ewhanson:
Looks good, @taslangraham. Just a few questions in pkp-lib. Once this is merged, feel free to create the issue and update the avatar component as well.

taslangraham:
@ewhanson Thanks for the review

taslangraham:
@ewhanson I've made small updates to pkp-lib PR based on your initial feedback. Please take a look.

ewhanson:
Thanks, @taslangraham! Just one comment, but once you've addressed that, feel free to merge.

PR #10639, opened 2024-11-26
pkp/pkp-lib#10489 Allow users to add custom initials for avatars

taslangraham:
For pkp/pkp-lib#10489

### DIFF
--- a/classes/user/form/IdentityForm.php
+++ b/classes/user/form/IdentityForm.php
@@ -18,6 +18,7 @@ namespace PKP\user\form;
 
 use APP\core\Application;
 use APP\template\TemplateManager;
+use Illuminate\Support\Str;
 use PKP\orcid\OrcidManager;
 use PKP\user\User;
 
@@ -98,6 +99,7 @@ class IdentityForm extends BaseProfileForm
             'familyName' => $user->getFamilyName(null),
             'preferredPublicName' => $user->getPreferredPublicName(null),
             'orcid' => $user->getOrcid(),
+            'preferredAvatarInitials' => $user->getPreferredAvatarInitials(null),
         ];
     }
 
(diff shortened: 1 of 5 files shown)


schemas/user.json
```diff
@@ -217,6 +217,15 @@
 				"nullable"
 			]
 		},
+		"preferredAvatarInitials": {
+			"type": "string",
+			"description": "An optional field for users to specify the initials that they want to be displayed in their avatars.",
+			"multilingual": false,
+			"apiSummary": true,
+			"validation": [
+				"nullable"
```
ewhanson, line 226: You could also include the `max:2` validation here as well.

classes/user/form/IdentityForm.php
```diff
@@ -123,14 +124,15 @@ public function execute(...$functionArgs)
...
-        if($this->getData('removeOrcidId') === 'true') {
+        if ($this->getData('removeOrcidId') === 'true') {
             $user->setOrcid(null);
             $user->setOrcidVerified(false);
             OrcidManager::removeOrcidAccessToken($user);
         } else {
             $user->setGivenName($this->getData('givenName'), null);
             $user->setFamilyName($this->getData('familyName'), null);
             $user->setPreferredPublicName($this->getData('preferredPublicName'), null);
+            $user->setPreferredAvatarInitials($this->getData('preferredAvatarInitials'), null);
```
ewhanson, line 135: Should these be transformed to capital letters at some point? Maybe enforced at the time of typing or here? Or should it only happen in the avatar component. What do you think?

ewhanson (changes requested):

classes/user/form/IdentityForm.php
taslangraham: Good question! I think it makes sense to capitalize it at the input stage. This would keep it consistent with how it's eventually shown to user

taslangraham: I decided to handle the capitalization in both places. It's better for UX to show the change as the user types, and I prefer not to rely solely on the front-end for the final data format(in case a user populates the field with JS, etc ).

templates/user/identityForm.tpl
```diff
@@ -80,3 +92,8 @@
 
 	{fbvFormButtons hideCancel=true submitText="common.save"}
 </form>
+
+
+<script>
```
ewhanson, line 97: Did you mean to have this empty script tag here?

taslangraham: should be removed

Issue #10362, opened 2024-09-03
[OJS 3.4] variable{$editorName} not replaced in activity log for decisions


forgive38:
**Describe the bug**
When you take a decision, you see that in activity log

2024-09-03 	Administrator Testdrive 	{$editorName} sent this submission back to the copyediting stage.
2024-09-03 	Administrator Testdrive 	{$editorName} sent this submission to the production stage. 

**What application are you using?**
 OJS (testdrive 3.4 or 3.4.0.6)

aleskl:
I can also see placeholders not populating in the **Activity log**, it seems to be specific for editors:

2024-10-09 {$editorName} requested revisions for this submission.
2024-10-09 Editor {$userName} has confirmed a review for the round 1 review for submission 3286.

**What application are you using?**
OJS 3.4.0.6

kaitlinnewson:
Initial PRs for review (will port to main once approved):

- pkp-lib 3_4_0: https://github.com/pkp/pkp-lib/pull/10628
- OJS 3_4_0 (submodule update for tests only): https://github.com/pkp/ojs/pull/4530

This only covers new editorial decisions being logged though, and there will still be missing `editorName` values in the `event_log_settings` table.

kaitlinnewson:
PRs for review (will port to main once approved):

- pkp-lib 3_4_0: https://github.com/pkp/pkp-lib/pull/10628
- OJS 3_4_0 (submodule update for tests only): https://github.com/pkp/ojs/pull/4530

This only covers cases of new editorial decisions being logged though, and there still may be missing `editorName` values in the `event_log_settings` table.

kaitlinnewson:
For `PKPReviewerGridHandler`, it was `$userName` in 3.3 and 3.2, and it was only `$editorName` in 3.4 - so in that case it makes sense to stick with it and migrate any discrepancies created in 3.4 for 3.4.1 (if released) and 3.5. I'll look at those next.

kaitlinnewson:
@Vitaliy-1 I've forward-ported, added the migration, and listed the PRs above - ready for another review.

[...]

PR #10646, opened 2024-11-28
#10362 include editorName when logging an editorial decision

kaitlinnewson:


classes/migration/upgrade/v3_5_0/I10362_EventLogEditorNames.php
```diff
@@ -0,0 +1,77 @@
...
+class I10362_EventLogEditorNames extends Migration
+{
+
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        $idsToUpdate = DB::table('event_log as e')
+            ->select('e.log_id', 'e.user_id', 'e.event_type')
```
bozana, line 32: I think there is no need to select event_type

classes/migration/upgrade/v3_5_0/I10362_EventLogEditorNames.php
```diff
@@ -0,0 +1,77 @@
...
+                $join->on('es.log_id', '=', 'e.log_id')
+                    ->where('es.setting_name', '=', 'editorName');
+            })
+            ->whereNull('es.log_id')
+            ->whereIn('e.event_type', [
+                0x30000003, // PKPSubmissionEventLogEntry::SUBMISSION_LOG_EDITOR_DECISION
+                0x30000004 // PKPSubmissionEventLogEntry::SUBMISSION_LOG_EDITOR_RECOMMENDATION
+            ])
+            ->orderBy('e.log_id')
+            ->pluck('user_id', 'log_id');
```
bozana, line 43: I think this works, but maybe to use 'e.user_id' and 'e.log_id'?

PR #10654, opened 2024-12-04
pkp/pkp-lib#10616 Add ThemePlugin function to get localized theme opt…

NateWr:
…ion value

### DIFF
--- a/classes/core/DataObject.php
+++ b/classes/core/DataObject.php
@@ -20,6 +20,7 @@ namespace PKP\core;
 
 use APP\core\Application;
 use Exception;
+use PKP\core\traits\LocalizedData;
 use PKP\db\DAO;
 use PKP\db\DAORegistry;
 use PKP\facades\Locale;
@@ -33,6 +34,8 @@ use PKP\metadata\MetadataSchema;
  */
 class DataObject
 {
+    use LocalizedData;
+
     /** @var array Array of object data */
     public array $_data = [];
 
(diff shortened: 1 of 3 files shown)


asmecher (approved):

classes/core/DataObject.php
```diff
@@ -80,57 +70,11 @@ public function __construct()
...
     {
-        foreach ($this->getLocalePrecedence($preferredLocale) as $locale) {
-            $value = & $this->getData($key, $locale);
-            if (!empty($value)) {
-                $selectedLocale = $locale;
-                return $value;
-            }
-            unset($value);
+        $value = $this->getData($key);
+        if (!is_array($value)) {
```
bozana, line 74: @NateWr, is there a case where this happens: that there is a localized data and not an array (and we use getLocalizedData function to get that data)?
Just curious why we have added this now...

NateWr: Hi @bozana! No, not that I know of. But `LocalizedData::getBestLocalizedData()` has a type hint of `array`, so this will prevent such a case causing a fatal error.

bozana: I am wondering if we need this part of code at all:
```
$value = $this->getData($key);
if (!is_array($value)) {
  return $value;
}
```

bozana: I think I understand: this is only for type checking, for the function getBestLocalizedData.
Hmmm... I suppose both are OK: 
- It would be OK to get an error if the type is not an array, because then it is probably something wrong.
- But this way it is maybe also OK, because maybe it is not so important that getLocalizedData() is used correctly, for multilingual keys... 
Hmmm...

Issue #10651, opened 2024-12-03
Update UI to use user's custom initials when available


taslangraham:
Description

https://github.com/pkp/pkp-lib/issues/10489 introduced support for users to set a 2-character custom initial to represent themselves in the application. We need to update the UI components that display initials to use these custom initials when available. If no custom initial is set, the current behavior of deriving initials from the user’s name should be used as a fallback.

UI Library components that need to be updated:
- [UserAvatar ](https://github.com/pkp/ui-library/tree/main/src/components/UserAvatar)
- [InitialsAvatar ](https://github.com/pkp/ui-library/tree/main/src/components/InitialsAvatar)

**What application are you using?**
 OJS, OMP and OPS version 3.5(main)

**PR**
ui-library - https://github.com/pkp/ui-library/pull/455/files
pkp -lib - https://github.com/pkp/pkp-lib/pull/10661/files
**Submodule updates**
ojs - https://github.com/pkp/ojs/pull/4540/files
omp - https://github.com/pkp/omp/pull/1781/files
ops - https://github.com/pkp/ops/pull/823/files

taslangraham:
@blesildaramirez Do you mind reviewing the PR in the issue description, for me, please?

jardakotesovec:
@taslangraham Looking good - just suggestion whether we could do the initial calculation on server side?  So the logic whether we use the 'prefered' or calculated would happen on php. And user/participant objects would have just the 'initial' field which would be calculated. What you think?

taslangraham:
@jardakotesovec I don't have any objections to doing it that way. I'll go ahead and move the logic over to server-side

taslangraham:
@jardakotesovec I've moved the logic to calculate initials to server-side, and updated the components only to accept and render the initials. Please take a look.

This required changes to pkp-lib as well. @ewhanson Could you take a look at that PR, please? I've included it in the issue description.

ewhanson:
Thanks, @taslangraham. Just a few questions there, but the code looks good. 👍

taslangraham:
@ewhanson I've responded to your questions and made adjustments based on your review. Please take a look.

ewhanson:
Thanks, @taslangraham! They look good to me. Assuming you're not waiting on anything for the ui-library, feel free to rebase and merge.

taslangraham:
All merged! Thanks for the review @ewhanson.

PR #10661, opened 2024-12-05
pkp/pkp-lib#10651 Determine initials to be displayed for users

taslangraham:
For pkp/pkp-lib#10651

### DIFF
--- a/classes/identity/Identity.php
+++ b/classes/identity/Identity.php
@@ -22,6 +22,7 @@
 namespace PKP\identity;
 
 use APP\core\Application;
+use Illuminate\Support\Str;
 use PKP\facades\Locale;
 use PKP\orcid\traits\HasOrcid;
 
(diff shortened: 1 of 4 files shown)


classes/identity/Identity.php
```diff
@@ -376,6 +377,31 @@ public function setBiography($biography, $locale)
...
+    /***
+     * Get the initials that should be displayed when representing the user.
+     */
+    public function getDisplayInitials(): string
+    {
+        $initials = $this->getPreferredAvatarInitials(null);
+
+        if (!$initials) {
+            // Get the individual name parts. Using hyphens and spaces as delimiters
+            $words = collect(preg_split('/[\s-]+/', $this->getFullName()));
```
ewhanson, line 390: Has there been any discussion around best practice for selecting the initials when they aren't provided by the `getPreferredAvatarInitials()` above? The code here looks fine, but I'm wondering if there are any other established practices or discussions had elsewhere (I suppose that's what the preferred avatar initials are for 😅).

schemas/user.json
```diff
@@ -337,6 +337,11 @@
 					}
 				}
 			}
+		},
+		"initials": {
```
ewhanson, line 341: Will the be the preferred avatar initials if those are present? Or will this always be the automatically generated ones?

ewhanson (changes requested):

classes/identity/Identity.php
taslangraham: On further revision, I've found an earlier [issue](https://github.com/pkp/pkp-lib/issues/8899#issue-1657535800) outlining how initials should work based on the user's name; this is listed under the **_Specs Update - Tuesday, October 1st, 2024_** section. 

I'll review the code and make the necessary chnages.

schemas/user.json
taslangraham: This will be the value returned by the _getDisplayInitials_ method introduced in this PR. Therefore when preferred avatar initials are available, those will be the value. Otherwise,  initials will be determined by the user's name.

After further consideration, I believe `displayInitials `would be a more appropriate name. The current name may imply that the initials are exclusively derived from the user's name, which could lead to confusion when preferred avatar initials are returned instead.

ewhanson (approved):

PR #10666, opened 2024-12-09
[PKP-LIB][main] #4860 Add JAV support to publications

defstat:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -111,7 +112,10 @@ class PKPSubmissionController extends PKPBaseController
         'getPublicationIdentifierForm',
         'getPublicationLicenseForm',
         'getPublicationTitleAbstractForm',
-        'getChangeLanguageMetadata'
+        'getChangeLanguageMetadata',
+        'getJAVStageMetadata',
```
asmecher, line 116: There's inconsistency between these two lines on `JAV` vs. `Jav`; I don't have a strong preference but they should be consistent. In any case, I think it would be better to avoid the acronym, as most won't be familiar with it. And listing stage and numbering in the API name is overspecific. Instead, maybe `changeVersionData`?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -111,7 +112,10 @@ class PKPSubmissionController extends PKPBaseController
...
+        'changeJavStageAndNumbering',
+        'getNextAvailableJavVersionNumberingForStage',
```
asmecher, line 118: As above, maybe `getNextAvailableVersionData`?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -943,6 +957,68 @@ public function changeLocale(Request $illuminateRequest): JsonResponse
...
+        $request = $this->getRequest();
+        $publication = Repo::publication()->get((int) $illuminateRequest->route('publicationId'));
+
+        if (!$publication) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        // $submission = Repo::submission()->get((int) $illuminateRequest->route('submissionId'));
```
asmecher, line 974: Dead code should be removed

classes/publication/enums/JavStage.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * @file publication/enums/JavStage.php
```
asmecher, line 4: Because this refers to an external standard, it would be good to add a URL to where it's specified.

classes/publication/helpers/JavStageAndNumbering.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public JavStage $javStage;
+    public int $javVersionMajor;
+    public int $javVersionMinor;
+
+    public function getVersionStageDisplay(): string 
+    {
+        $versionStageValue = $this->javStage->value;
+
+        return "$versionStageValue $this->javVersionMajor.$this->javVersionMinor";
```
asmecher, line 32: This should probably be localized.

classes/publication/enums/JavStage.php
```diff
@@ -0,0 +1,24 @@
...
+ * @class JavStage
+ *
+ * @brief Enumeration for JAV versioning stages
+ */
+
+namespace PKP\publication\enums;
+
+enum JavStage: string
+{
+    case AUTHOR_ORIGINAL = 'AO';
```
asmecher, line 19: While these are fine to use in the codebase, we should be careful not to present them straight to the front end without having them map to translatable locale keys.

classes/submission/PKPSubmission.php
```diff
@@ -289,6 +291,59 @@ public function getUIDisplayString()
...
+    /**
+     * Return a collection of all existing JAV Stages 
+     * for the submission's publications
+     *
+     * @return Collection<JavStageAndNumbering>
+     */
+    public function getExistingJavStagesByPublications(): Collection
+    {
+        $publications = $this->getData('publications');
+        if ($publications->isEmpty()) {
```
asmecher, line 304: Aren't these 3 lines redundant? If the collection was empty, the result of this operation will be an empty collection anyway.

locale/en/submission.po
```diff
@@ -2454,3 +2454,51 @@ msgstr "Please disclose any competing interests this author may have with the re
...
+msgid "publication.changeVersionStage.title"
+msgstr "Version Stage"
+
+msgid "publication.changeVersionStage.buttonLabel"
+msgstr "Manage version stage"
+
+msgid "publication.versionStage.minorOrMajor"
+msgstr "Version Impact"
+
+msgid "publication.versionStage.minorOrMajor.description"
```
asmecher, line 2467: This needs a translation, or to be removed if it's redundant. But I'd suggest not including the locale file changes -- they're likely to be made redundant by someone else's UI implementation.

asmecher (changes requested):
Looks good, @defstat! I've added some specific comments, but here are two general ones:

- I think we should remove the old `version` column after migrating its data to the new columns. This will make sure we don't accidentally rely on it when it's completely supplanted by the new stuff.
- I'd suggest we don't use the JAV acronym anywhere in this. I don't think it's a formally adopted acronym by the working group, and it's likely that folks will be generally familiar with it. Better to clarify all of our conversation around versions (e.g. removing the other column) so that we *always* mean a JAV version when we talk about a version, and not complicate the code with JAV where we specifically mean that standard/working group.

defstat:
Closed in favor of #10810

Issue #9661, opened 2024-01-27
Allow Journal Managers to invite users to adopt a role - ORCiD


Devika008:
Hello,

Since https://github.com/pkp/pkp-lib/issues/3022#top was too big, upon discussion we decided to divide it into four smaller issues.

This fourth issue highlights the ORCiD prerequisite for the user flow of inviting editors, authors, readers, reviewers, etc to OJS. The information stated below reflects all the research, discussions and feedback received on this issue over the last few years along with some minor testing with a small user group at the Copenhagen Sprint.

### WIP, PRs, and TODOs

PRs:

- `pkp-lib`: pkp/pkp-lib#10679
- `ui-library`: pkp/ui-library#460
- `ojs`: pkp/ojs#4589
- `omp`: pkp/omp#1802
- `ops`: pkp/ops#839

### Summary of requirement

1. In the “Users & Roles” section of the settings, we add a new “Invite user to take a role” action that can be used to invite any user to the journal OJS software based on an email address. The email sent will add the following:
- Invitation request along with details of the role, starting date etc.
- ORCiD verification details

2. Starting with 3.5.0, ORCiD Verification would not happen through a plugin. We're rewriting ORCiD support to be integrated into OJS rather than maintaining a plugin. The option to make ORCiD mandatory should be present as part of the 3.5.0 implementation; it shouldn't be necessary to write up a plugin to add the "make ORCiDs mandatory" checkbox.

3. If the JM/System Administrator/Editor has marked ORCiD verification as mandatory, then the email invitation sent will highlight that the verification is mandatory. If it’s not enabled then the ORCiD mandatory message on the email will not be shown

4. Mandatory or not, before the user is redirected to the ORCiD website to verify they have an option to skip the step. Since we cannot customize the ORCiD verification screen to show skip or not, the email will be the vehicle for communication.

5. If the user deny's access to ORCiD, they can go back anytime in their profile to authenticate it later on in the future

### ORCID Journey in User Invitation process

a. Below is a sample of the email the user will receive informing them about ORCiD verification

![Image](https://github.com/pkp/pkp-lib/assets/112014397/f11702eb-af80-47c0-af2d-555999e75cab)

b. After the user accepts the invitation, the user needs to first create an account on OJS

![Image](https://github.com/pkp/pkp-lib/assets/112014397/2e6570f9-25f6-4d09-9e2a-22a699f19ad3)

[...]

PR #10679, opened 2024-12-10
pkp/pkp-lib#9661 Allow Journal Managers to invite users to adopt a role - ORCiD

ewhanson:


ewhanson:
Thanks @bozana! 

@defstat, would you mind having a look at this? I know you've done a lot of the backend invitation work and this touches on that a bit. Thanks!

defstat (approved):
Sorry for the delay on that. It looks ok to me. Is there a corresponding UI-LIB PR that completes this feature, or should that work on the accept invitation part of the code on ui-library `main` branch?

Issue #10486, opened 2024-10-01
Authors are able to edit submission metadata and galleys even after a submisson is rejected


alexxxmendonca:
**Describe the bug**
After a submission is rejected, autors are able to edit submission metadata and galleys. They can remove and add authors freely and remove PDF files as shown in this History:

![image](https://github.com/user-attachments/assets/62ad7b0a-e7df-43f3-a2a3-b55d113e9213)

**To Reproduce**
1. Reject any submission
2. Login as the author of that submission
3. Notice how as the author you can edit the metadata (add/delete authors) and add/delete galleys.

There are undesired consequences such as losing record of the manuscript files, authors associated to that submission and moreover, plugins such as the Author History (https://github.com/lepidus/authorsHistory) is unabled to fetch the data since it's been removed from the database, making the results shown by the plugin to be faulty. Moreover, this negatively affects statistics as well.

**What application are you using?**
Open Preprint Systems 3.3.0.14

YvesLepidus:
I made an implementation proposal to fix this issue in this [PR](https://github.com/pkp/pkp-lib/pull/10693).

asmecher:
Thanks, @YvesLepidus! I've made a few more comments [on your PR](https://github.com/pkp/pkp-lib/issues/10486).

Vitaliy-1:
Thanks again, @YvesLepidus, I left a couple of comments to finalize a PR

Vitaliy-1:
Thanks @YvesLepidus , I left a comment and it looks to me that one of the previous suggestion regarding dual assignments isn't resolved. Am I correct?

YvesLepidus:
@Vitaliy-1 I think all discussions have been resolved. Please let me know if anything has been missed.

asmecher:
@Vitaliy-1, just a heads-up that this needs your attention!

Vitaliy-1:
Sorry, taking a look!

Vitaliy-1:
tests OPS: https://github.com/pkp/ops/pull/1277

Vitaliy-1:
PRs
stable 3.3.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/12638
OPS: https://github.com/pkp/ops/pull/1277
OMP: https://github.com/pkp/omp/pull/2325

Vitaliy-1:
Looks like this is relevant for all stable and a main branch

YvesLepidus:
[...]

PR #10693, opened 2024-12-12
pkp/pkp-lib#10486 adds validation to prevent an author of a rejected submission from editing metadata in the submission.

YvesLepidus:
Related issue: [pkp/pkp-lib#10486](https://github.com/pkp/pkp-lib/issues/10486)

### DIFF
--- a/classes/services/PKPSubmissionService.inc.php
+++ b/classes/services/PKPSubmissionService.inc.php
@@ -795,6 +795,12 @@ abstract class PKPSubmissionService implements EntityPropertyInterface, EntityRe
 	public function canEditPublication($submissionId, $userId) {
 		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
 		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
+		$submission = $this->get($submissionId);
+        $isAuthorOfTheSubmission = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submissionId, ROLE_ID_AUTHOR, null, $userId)->toArray());
+        // If the submission is declined and the current user is an author of the submission
+        if ($submission->getStatus() == STATUS_DECLINED && $isAuthorOfTheSubmission) {
+            return false;
+        }
 		// Check for permission from stage assignments
 		foreach ($stageAssignments as $stageAssignment) {
 			if ($stageAssignment->getCanChangeMetadata()) {
--- a/classes/services/PKPSubmissionService.inc.php
+++ b/classes/services/PKPSubmissionService.inc.php
@@ -796,11 +796,11 @@ abstract class PKPSubmissionService implements EntityPropertyInterface, EntityRe
 		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
 		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
 		$submission = $this->get($submissionId);
-        $isAuthorOfTheSubmission = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submissionId, ROLE_ID_AUTHOR, null, $userId)->toArray());
-        // If the submission is declined and the current user is an author of the submission
-        if ($submission->getStatus() == STATUS_DECLINED && $isAuthorOfTheSubmission) {
-            return false;
-        }
+                $userIsAuthor = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submissionId, ROLE_ID_AUTHOR, null, $userId)->toArray());
+                // If the submission is declined and the current user is an author of the submission
+                if ($submission->getStatus() == STATUS_DECLINED && $userIsAuthor) {
+                        return false;
+                }
 		// Check for permission from stage assignments
 		foreach ($stageAssignments as $stageAssignment) {
 			if ($stageAssignment->getCanChangeMetadata()) {


classes/services/PKPSubmissionService.inc.php
```diff
@@ -795,6 +795,12 @@ public function delete($submission) {
 	public function canEditPublication($submissionId, $userId) {
 		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
 		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
+		$submission = $this->get($submissionId);
```
asmecher, line 798: Every case where `canEditPublication` is called, the `$submission` object is already available -- so I'd suggest changing calling code to just pass the submission object instead of `$submission->getId()`. That would save the additional (expensive) fetch of the submission object from the DB.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -795,6 +795,12 @@ public function delete($submission) {
...
+                $userIsAuthor = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submissionId, ROLE_ID_AUTHOR, null, $userId)->toArray());
+                // If the submission is declined and the current user is an author of the submission
+                if ($submission->getStatus() == STATUS_DECLINED && $userIsAuthor) {
+                        return false;
```
asmecher, line 802: I think a more complex policy will be required. If someone is an editor, for example, they probably should have permission to edit metadata on declined submissions they also authored. This restriction should probably apply to submissions only when the author is the only stage assignment.

asmecher (changes requested):
Thanks, @YvesLepidus! I made a few suggestions for change.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: Hi, @asmecher ! :smiley:   
I've [adjusted](https://github.com/pkp/pkp-lib/pull/10693/commits/4a084361365b6d8fd0875846f03d3ca102688c7e) the `canEditPublication` method and updated the calls in lib-pkp. I've also done the PR to include the method call update in [OMP](https://github.com/pkp/omp/pull/1786) and [OPS](https://github.com/pkp/ops/pull/826).

asmecher: @YvesLepidus I think you're missing a PR for OJS too, no?

classes/services/PKPSubmissionService.inc.php
YvesLepidus: > I think a more complex policy will be required. If someone is an editor, for example, they probably should have permission to edit metadata on declined submissions they also authored.

I [added a check](https://github.com/pkp/pkp-lib/pull/10693/commits/e49e800c7d3189aaf97680d724de8f67ebd3cf33) if the user doesn't have permissions to edit metadata according to the role.

YvesLepidus: > This restriction should probably apply to submissions only when the author is the only stage assignment.

Sorry, I didn't understand exactly what you meant. Is that a necessary validation in [this conditional](https://github.com/pkp/pkp-lib/blob/e49e800c7d3189aaf97680d724de8f67ebd3cf33/classes/services/PKPSubmissionService.inc.php#L801), too?

classes/services/PKPSubmissionService.inc.php
YvesLepidus: I couldn't find a call to the `canEditPublication` method in the [OJS 3.3.0 code](https://github.com/search?q=repo%3Apkp%2Fojs%20canEditPublication&type=code). :thinking: 
I only [found](https://github.com/pkp/ojs/blob/edd0f343981158db790dd6779ea4fc4f2cda7238/pages/workflow/WorkflowHandler.php#L122) it in the `main` branch, using Repository.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: > Sorry, I didn't understand exactly what you meant. Is that a necessary validation in [this conditional](https://github.com/pkp/pkp-lib/blob/e49e800c7d3189aaf97680d724de8f67ebd3cf33/classes/services/PKPSubmissionService.inc.php#L801), too?

I discussed it internally with others and understood the point. In the current PR code, there are still issues, such as an author having the role of Moderator in the OPS but not being able to edit the declined submission. I will make the necessary adjustments.

YvesLepidus: I've [adjusted](https://github.com/pkp/pkp-lib/pull/10693/commits/2819c8ed645cbfcd4e54df42e987fd059bc5fb55) the restriction to apply only in cases where the author has no role other than author/reader.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +788,34 @@ public function delete($submission) {
...
 	 * Check if a user can edit a publications metadata
 	 *
-	 * @param int $submissionId
+	 * @param Submission $submission
 	 * @param int $userId
 	 * @return boolean
 	 */
-	public function canEditPublication($submissionId, $userId) {
+	public function canEditPublication($submission, $userId) {
+                $contextId = Application::get()->getRequest()->getContext()->getId();
```
asmecher, line 796: It's risky to use `Request::getContext` here -- that'll work for web-based invocations of this function, but if it's ever called from e.g. a command-line tool or when processing a batch job at the end of a request, it may give the wrong value. Instead, use `$submission->getData('contextId')`. (I see that the existing code used to use `$request->getContext()` further down -- best to fix it now.)

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +788,34 @@ public function delete($submission) {
...
-		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
+		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submission->getId(), $userId, null)->toArray();
+                $userIsAuthor = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR, null, $userId)->toArray());
+                // If the submission is rejected and the user's only role is an author
+                if ($submission->getStatus() == STATUS_DECLINED && $userIsAuthor) {
+                        $roleDao = DAORegistry::getDAO('RoleDAO'); /* @var $roleDao RoleDAO */
+                        $roles = $roleDao->getByUserId($userId, $contextId);
+                        foreach ($roles as $role) {
+                                if ($role->getRoleId() != ROLE_ID_AUTHOR && $role->getRoleId() != ROLE_ID_READER) {
+                                        return true;
```
asmecher, line 806: I think this will erroneously bypass the checks further down in the code.

asmecher (changes requested):
Thanks, @YvesLepidus, just another couple of comments!

When you've gone through a set of review comments, don't forget to tag me on the issue to let me know; otherwise I won't know when you've finished adjusting things and have it ready for a second look.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: @asmecher i [adjusts](https://github.com/pkp/pkp-lib/pull/10693/commits/1c03478b0b4503f83a0949d18854dadcce0d2573) to get the Context identifier from Submission. And I've also changed the other parts of the code that get it from the request.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: @asmecher I've updated the code in this [commit](https://github.com/pkp/pkp-lib/pull/10693/commits/bc5e628095b2fb8b0a6c9e2b832f2724ee7f8d6c) so that it doesn't return `true` in this condition.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -176,10 +176,7 @@ public function getProperties($submission, $props, $args = null) {
 
 		// Retrieve the submission's context for properties that require it
 		if (array_intersect(['_href', 'urlAuthorWorkflow', 'urlEditorialWorkflow'], $props)) {
-			$submissionContext = $request->getContext();
```
asmecher, line 179: @YvesLepidus, these lines (and further down) were an optimization: if the request's context specifies the right ID, then we use it; otherwise fetch the context from the database. I think those were safe and reliable and shouldn't be removed.

asmecher (changes requested):
@YvesLepidus, just one more individual comment, thanks! Once that's addressed, could you forward-port this to `stable-3_4_0` and `main`?

classes/services/PKPSubmissionService.inc.php
YvesLepidus: @asmecher I reversed it in this [commit](https://github.com/pkp/pkp-lib/pull/10693/commits/2757b5137d53ad4216613d0a185b6ab95832b345).

YvesLepidus:
> @YvesLepidus, just one more individual comment, thanks! Once that's addressed, could you forward-port this to `stable-3_4_0` and `main`?

Ok @asmecher, I'll adapt it for these branchs.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -317,7 +317,7 @@ public function getPropertyReviewAssignments($submission) {
 
 			$request = \Application::get()->getRequest();
 			$currentUser = $request->getUser();
-			$context = $request->getContext();
+			$context = Services::get('context')->get($submission->getData('contextId'));
```
asmecher, line 320: Sorry to keep sending you back on this issue -- but this will potentially add a lot of overhead, as it'll mean fetching the object repeatedly if called on a  list of submissions. I would suggest using the same optimization you see elsewhere:
```php
if (!$context || $context->getId() != $submission->getData('contextId')) {
    $context = Services::get('context')->get($submission->getData('contextId'));
}
```
This will probably be faster, but still safe for batch jobs.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -396,7 +396,7 @@ public function getPropertyStages($submission, $stageIds = null) {
 		}
 
 		$currentUser = \Application::get()->getRequest()->getUser();
-		$context = \Application::get()->getRequest()->getContext();
+		$context = Services::get('context')->get($submission->getData('contextId'));
```
asmecher, line 399: Same as [here](https://github.com/pkp/pkp-lib/pull/10693/files#r1937505161)

classes/services/PKPSubmissionService.inc.php
```diff
@@ -550,11 +550,7 @@ public function getWorkflowUrlByUserRoles($submission, $userId = null) {
 			return false;
 		}
 
-		$submissionContext = $request->getContext();
-
-		if (!$submissionContext || $submissionContext->getId() != $submission->getData('contextId')) {
-			$submissionContext = Services::get('context')->get($submission->getData('contextId'));
-		}
+		$submissionContext = Services::get('context')->get($submission->getData('contextId'));
```
asmecher, line 553: This change should be removed.

YvesLepidus:
Hi @asmecher!
I've made the corrections pointed out in the last review in [this commit](https://github.com/pkp/pkp-lib/pull/10693/commits/c7402c4f10216d0b4154fdb03c31b15dc5c9b9ed). Can you check if it's ok?

asmecher:
Thanks, @YvesLepidus, this looks good to me. @Vitaliy-1, could you also review the PR?

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +794,38 @@ public function delete($submission) {
...
-		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
+		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submission->getId(), $userId, null)->toArray();
+                $userIsAuthor = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR, null, $userId)->toArray());
+                // If the submission is rejected and the user's only role is an author
+                if ($submission->getStatus() == STATUS_DECLINED && $userIsAuthor) {
+                        $userIsOnlyAuthorOrReader = true;
+                        $roleDao = DAORegistry::getDAO('RoleDAO'); /* @var $roleDao RoleDAO */
+                        $roles = $roleDao->getByUserId($userId, $contextId);
+                        foreach ($roles as $role) {
+                                if ($role->getRoleId() != ROLE_ID_AUTHOR && $role->getRoleId() != ROLE_ID_READER) {
```
Vitaliy-1, line 812: What about a reviewer role? 
Also, to double-check, if the user has, e.g., context role assistant, we allow to edit metadata of own cancelled submissions, is this the idea?

classes/services/PKPSubmissionService.inc.php
```diff
@@ -251,7 +251,7 @@ function($publication) use ($args, $submission, $submissionContext) {
 			}
 		}
 
-		$values = Services::get('schema')->addMissingMultilingualValues(SCHEMA_SUBMISSION, $values, $request->getContext()->getSupportedSubmissionLocales());
+		$values = Services::get('schema')->addMissingMultilingualValues(SCHEMA_SUBMISSION, $values, $submissionContext);
```
Vitaliy-1, line 254: `submissionContext->getSupportedSubmissionLocales()`

Vitaliy-1 (changes requested):
Thanks, @YvesLepidus!
I left a couple of comments

classes/services/PKPSubmissionService.inc.php
Vitaliy-1: (this will go to other applications as well as it's in the shared library)

classes/services/PKPSubmissionService.inc.php
YvesLepidus: Thanks, @Vitaliy-1  
I've fix it in [this commit](https://github.com/pkp/pkp-lib/pull/10693/commits/c62eaaacf76615c3c05ecdbd864155b5388d2ba2).

classes/services/PKPSubmissionService.inc.php
```diff
@@ -176,7 +176,7 @@ public function getProperties($submission, $props, $args = null) {
 
 		// Retrieve the submission's context for properties that require it
 		if (array_intersect(['_href', 'urlAuthorWorkflow', 'urlEditorialWorkflow'], $props)) {
-			$submissionContext = $request->getContext();
+                        $submissionContext = $request->getContext();
```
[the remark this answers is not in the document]
YvesLepidus: @Vitaliy-1 
I think all the inserted codes with incorrect indentation have been fixed.

- https://github.com/pkp/pkp-lib/pull/10693/commits/303e08507b2db7269bcd23741fedeec85e14a1b0
- https://github.com/pkp/pkp-lib/pull/10693/commits/e35354d9732837134cc3099f6441322eabab958c

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +794,38 @@ public function delete($submission) {
 	/**
 	 * Check if a user can edit a publications metadata
 	 *
-	 * @param int $submissionId
+	 * @param Submission $submission
 	 * @param int $userId
 	 * @return boolean
 	 */
-	public function canEditPublication($submissionId, $userId) {
+	public function canEditPublication($submission, $userId) {
```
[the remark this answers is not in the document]
YvesLepidus: @Vitaliy-1 
The motivation for the change was to save an additional search for the Submission object in the database, in all calls of the `canEditPublication` method the Submission object is present.

I’ve created two pull requests for OPS and OMP to update the method calls, but I didn’t find any occurrences in the OJS source code.

- [OPS PR](https://github.com/pkp/ops/pull/826)
- [OMP PR](https://github.com/pkp/omp/pull/1786)

classes/services/PKPSubmissionService.inc.php
```diff
@@ -397,6 +400,9 @@ public function getPropertyStages($submission, $stageIds = null) {
 
 		$currentUser = \Application::get()->getRequest()->getUser();
 		$context = \Application::get()->getRequest()->getContext();
+                if (!$context || $context->getId() != $submission->getData('contextId')) {
```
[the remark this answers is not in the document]
YvesLepidus: @Vitaliy-1 
I don't think so. I haven't found a point in the file that needs similar validation.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: > What about a reviewer role?

@Vitaliy-1 
In this initial implementation, we focused on the author role, which motivated the [issue](https://github.com/pkp/pkp-lib/issues/10486) this PR addresses. However, I can adjust it if PKP needs to restrict it to other roles.

> Also, to double-check, if the user has, e.g., context role assistant, we allow to edit metadata of own cancelled submissions, is this the idea?

Yes, in principle, only the author user would have the limitation of not editing metadata of the rejected submission.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +794,38 @@ public function delete($submission) {
...
 	 * @param int $userId
 	 * @return boolean
 	 */
-	public function canEditPublication($submissionId, $userId) {
+	public function canEditPublication($submission, $userId) {
+		$contextId = $submission->getData('contextId');
 		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO'); /* @var $stageAssignmentDao StageAssignmentDAO */
-		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submissionId, $userId, null)->toArray();
+		$stageAssignments = $stageAssignmentDao->getBySubmissionAndUserIdAndStageId($submission->getId(), $userId, null)->toArray();
+		$userIsAuthor = !empty($stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR, null, $userId)->toArray());
```
Vitaliy-1, line 805: Sorry if I'm misunderstanding. I see that it was raised before - should we also check other stage assignments? E.g., dual assignments when the user is assigned as an editor and author. That's common case for editorial articles.

classes/services/PKPSubmissionService.inc.php
```diff
@@ -788,22 +794,38 @@ public function delete($submission) {
...
+		// If the submission is rejected and the user's only role is an author
+		if ($submission->getStatus() == STATUS_DECLINED && $userIsAuthor) {
+			$userIsOnlyAuthorOrReader = true;
+			$roleDao = DAORegistry::getDAO('RoleDAO'); /* @var $roleDao RoleDAO */
+			$roles = $roleDao->getByUserId($userId, $contextId);
+			foreach ($roles as $role) {
+				if ($role->getRoleId() != ROLE_ID_AUTHOR && $role->getRoleId() != ROLE_ID_READER) {
```
Vitaliy-1, line 812: This checks only global roles within the context. Which is good as admins and managers should be able to edit the metadata anyway.

classes/services/PKPSubmissionService.inc.php
YvesLepidus: @Vitaliy-1 this was mentioned in [this review](https://github.com/pkp/pkp-lib/pull/10693#discussion_r1882578892) from Alec:

> If someone is an editor, for example, they probably should have permission to edit metadata on declined submissions they also authored. This restriction should probably apply to submissions only when the author is the only stage assignment.

Based on this, I am limiting only users who have only one stage assignment and as an author.

YvesLepidus:
@Vitaliy-1 @asmecher 

I think all discussions have been resolved. Please let me know if anything has been missed.

It is also important to integrate the related PRs.

- [OPS](https://github.com/pkp/ops/pull/826)
- [OMP](https://github.com/pkp/omp/pull/1786)

If everything is OK, I will make the adjustments for 3.4 and 3.5.

Issue #10606, opened 2024-11-14
Native XML Export/Import is not handling internal review round correctly


jardakotesovec:
**Describe the bug**
While debugging e2e tests - it turned out that export/import test is corrupting data and results in submission which is in internal review (OMP), but without review round created - which is always expected to be there and therefore it crashes at least in one place..

Test is temporarily disabled. Its not easy to pin point if its long standing bug or regression, as this test was disabled for some time - https://github.com/pkp/pkp-lib/issues/10115 .   And the crash is coming from the new code in submission listing.

**To Reproduce**
1. Run MsmithSubmission.cy.js test
2. Re-enable the Import test in NativeXmlImportExportSubmission.cy.js and run it
3. Check the data structure - its missing review round, which is crashing new submission listing

Tested only in 3.5. Once investigated, it might be good to check the older versions if they might be affected by this.

**What application are you using?**
 OJS, OMP or OPS version 3.5 

**Additional information**
Here is crash stack trace when rendering new submission listing. It fails, because it assume that there is always review round for active stage.

```
[Wed Nov 13 15:47:06 2024] 127.0.0.1:44570 [500]: GET /index.php/publicknowledge/api/v1/_submissions?status=1&offset=0&count=30&page=1&perPage=30
[Wed Nov 13 15:47:06 2024] 127.0.0.1:44570 Closing
[Wed Nov 13 15:47:06 2024] 127.0.0.1:44578 Accepted
[Wed Nov 13 15:47:07 2024] Error: Call to a member function getId() on null in /home/runner/omp/classes/submission/maps/Schema.php:183
Stack trace:
#0 /home/runner/omp/lib/pkp/classes/submission/maps/Schema.php(654): APP\submission\maps\Schema->getAvailableEditorialDecisions()
#1 /home/runner/omp/lib/pkp/classes/submission/maps/Schema.php(393): PKP\submission\maps\Schema->getPropertyStages()
#2 /home/runner/omp/classes/submission/maps/Schema.php(65): PKP\submission\maps\Schema->mapByProperties()
#3 /home/runner/omp/lib/pkp/classes/submission/maps/Schema.php(250): APP\submission\maps\Schema->mapByProperties()
#4 /home/runner/omp/lib/pkp/classes/submission/maps/Schema.php(285): PKP\submission\maps\Schema->mapToSubmissionsList()
[...]

PR #10697, opened 2024-12-12
import Native xml, restrict review stages

withanage:
### Issue
- https://github.com/pkp/pkp-lib/issues/10606

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
+++ b/plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
@@ -20,9 +20,11 @@ use APP\core\Application;
 use APP\facades\Repo;
 use APP\submission\Submission;
 use PKP\core\Core;
+use PKP\db\DAORegistry;
 use PKP\filter\Filter;
 use PKP\filter\FilterGroup;
 use PKP\observers\events\BatchMetadataChanged;
+use PKP\submission\reviewRound\ReviewRound;
 use PKP\workflow\WorkflowStageDAO;
 
 class NativeXmlSubmissionFilter extends NativeImportFilter
@@ -82,8 +84,8 @@ class NativeXmlSubmissionFilter extends NativeImportFilter
         $submission->stampModified();
         $submission->setData('status', $node->getAttribute('status'));
         $submission->setData('submissionProgress', '');
-
-        $submission->setData('stageId', WorkflowStageDAO::getIdFromPath($node->getAttribute('stage')));
+        $stageName = $node->getAttribute('stage');
+        $submission->setData('stageId', WorkflowStageDAO::getIdFromPath($stageName));
 
         // Handle any additional attributes etc.
         $submission = $this->populateObject($submission, $node);
@@ -105,6 +107,9 @@ class NativeXmlSubmissionFilter extends NativeImportFilter
 
         $deployment->addImportedRootEntity(Application::ASSOC_TYPE_SUBMISSION, $submission);
 
+        if ($stageName === "externalReview" or $stageName === "internalReview") {
+            $this->createReviewRound($submission);
+        }
         return $submission;
     }
 
@@ -173,15 +178,15 @@ class NativeXmlSubmissionFilter extends NativeImportFilter
     }
 
     /**
-     * @see Filter::process()
-     *
      * @param \DOMDocument|string $document
      *
      * @return array Array of imported documents
+     * @see Filter::process()
+     *
      */
     public function &process(&$document)
     {
-        $importedObjects = & parent::process($document);
+        $importedObjects = &parent::process($document);
 
         $deployment = $this->getDeployment();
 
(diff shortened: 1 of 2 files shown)


bozana (commented):
Hi @withanage,
I think it would be better to implement it in the appropriate filter, i.e. in NatvieXmlSubmissionFilter, in the function handleElement():
The new code is only for submissions, on submission level. You do check it in the function addReviewStage, but I believe it is cleaner and better/easier to read it all in the NativeXmlSubmissionFilter, because a developer would look at the code for submission import there and would not know/have in mind that something is also happening here.
Also, I am not 100% sure, but could it be that $this->processResult could contain multiple submissions? You are considering only one in the function addReviewStage: `$submission = $this->processResult[0];`.
Thanks a lot!

withanage:
@bozana  , I changed the implementation according to your suggestion  and pushed.

> Also, I am not 100% sure, but could it be that $this->processResult could contain multiple submissions? You are considering only one in the function addReviewStage: $submission = $this->processResult[0];.

I think this use-case does not arise, when we handle it in the  filter.

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
```diff
@@ -238,4 +243,22 @@ public function getImportFilter($elementName)
...
+     * @throws \Exception
+     */
+    public function handleSubmissionStage(string $stageName, ?Submission $submission): void
+    {
+        if ($stageName === "externalReview" or $stageName === "internalReview") {
+            $reviewRoundDao = DAORegistry::getDAO('ReviewRoundDAO');
+            $reviewRound = $reviewRoundDao->build(
+                $submission->getId(),
+                $submission->getData('stageId'),
+                ReviewRound::REVIEW_ROUND_STATUS_PENDING_REVIEWERS,
```
bozana, line 260: Here the parameter reviewRound = 1 is missing

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
```diff
@@ -238,4 +243,22 @@ public function getImportFilter($elementName)
     {
         assert(false); // Subclasses should override
     }
+
+    /**
+     * @param string $stageName
+     * @param Submission|null $submission
+     * @return void
+     * @throws \Exception
+     */
```
bozana, line 252: Maybe to remove these, but add the description for the function.

Because we actually do not handle all submission stages here, we could also do the check 
`if ($stageName === "externalReview" or $stageName === "internalReview") {`
above in the `handleElement` function and call this function `createReviewRound`. But this is just optional, as you think/would like...

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
```diff
@@ -104,6 +106,9 @@ public function handleElement($node)
         $submission = Repo::submission()->get($submission->getId());
 
         $deployment->addImportedRootEntity(Application::ASSOC_TYPE_SUBMISSION, $submission);
+        //add Empty review stage for reviewed  submissions
+        $stageName = $node->getAttribute('stage');
```
bozana, line 110: How about having this a few lines above, before the line number 88, so that there are no double `$node->getAttribute('stage')`?

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
```diff
@@ -104,6 +106,9 @@ public function handleElement($node)
         $submission = Repo::submission()->get($submission->getId());
 
         $deployment->addImportedRootEntity(Application::ASSOC_TYPE_SUBMISSION, $submission);
+        //add Empty review stage for reviewed  submissions
```
bozana, line 109: If we would rename the new function either to createReviewRound or handleSubmissionReviewRound, maybe no comment here would be needed. 
Else maybe to have: "create a review round for submissions in review" ?

bozana:
Hi @withanage, thanks a lot!
I think one change is still needed: the review round parameter is missing. Else, just cosmetic comments... :-)
Thanks a lot!

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
withanage: @bozana  thanks! added!

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
withanage: @bozana yes, that makes sense!  I changed the signature of the function.

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
withanage: @bozana  sure, it was moved.

plugins/importexport/native/filter/NativeXmlSubmissionFilter.php
withanage: got rid of this obsolete comment after above changes

withanage:
@bozana  thanks for the review, I updated a revised commit.
https://github.com/pkp/pkp-lib/pull/10697/commits/740698c55d3b5de43bacb527c37ad12e696e2143

bozana:
Hi @withanage, great! All good! Thanks a lot!
Would you maybe like to enable the test `it('Imports submissions from XML', function() {` in NativeXmlImportExportSubmission.cy.js in a PR for OJS, OMP and OPS, and also add this submodule update, so that we see if the test passes then?

PR #10711, opened 2024-12-16
pkp/pkp-lib#6998 Don't allow mixed type for  query param

taslangraham:
For pkp/pkp-lib#6998


asmecher (changes requested):
@taslangraham, after these changes, is the `Collector::UNASSIGNED` constant unused? If so, please remove it from the class.

Issue #10178, opened 2024-07-09
Add search by DOI to DOI manager


ewhanson:
There was a recent [forum post](https://forum.pkp.sfu.ca/t/unable-to-edit-mistyped-doi-in-ojs-3-4/89297/1) involving a user with a duplicate DOI that needed to be found, and it would be helpful to be able to search for DOIs, not just submissions within the search interface in the DOI management UI. As things currently stand, it is necessary to run an SQL query to search for DOIs.

taslangraham:
(WIP)Draft PRs
ojs: https://github.com/pkp/ojs/pull/4563/files
pkp-lib: https://github.com/pkp/pkp-lib/pull/10724/files

ewhanson:
Thanks @taslangraham, looks like a good approach! 👍 Just a few thoughts, but definitely looks ready to implement in OMP and OPS.

taslangraham:
All ready for review @ewhanson 

PKP - https://github.com/pkp/ojs/pull/4563/files
OPS - https://github.com/pkp/ops/pull/830/files
OJS - https://github.com/pkp/ojs/pull/4563/files
OMP - https://github.com/pkp/omp/pull/1790/files

ewhanson:
Looks good, @taslangraham! Just one comment broadly applicable to OJS/OMP/OPS, but it's otherwise ready to go.

ewhanson:
Thanks @taslangraham! Looks good to me. It looks like all the tests are failing currently, though not sure if it's related to your changes or not. After a rebase, if the tests look good, feel free to merge.

taslangraham:
All merged. Thanks for the review @ewhanson

PR #10724, opened 2024-12-16
pkp/pkp-lib#10178 Allow search by DOI in DOI manager

taslangraham:


classes/submission/Collector.php
```diff
@@ -388,8 +390,12 @@ public function getQueryBuilder(): Builder
             $q->whereIn('s.submission_id', array_map(intval(...), $this->submissionIds));
         }
 
+        // Add support to serch using DOI identifiers
+        // search phrases starting number followed by a '.'  will be interpreted as a DOI identifier. E.g: 10.1
+        $pattern = '/^\d+\./i';
```
ewhanson, line 395: Up to you, but it may be useful to extract this to something like a static method on the DOI class so if we do end up needing to update this regex check in the future, there are fewer places took. Something like `Doi::isValid(string $doiToCheck): bool`...

ewhanson (approved):

Issue #10751, opened 2024-12-23
EditorialMasthead page not loading with Uncaught TypeError (userGroup)


kaitlinnewson:
**Describe the bug**
The Editorial Masthead page is not loading due to a fatal error from the userGroup class - "`Uncaught TypeError: PKP\userGroup\Repository::getMastheadUserIdsByRoleIds()`"

**To Reproduce**
1. On a journal home page, open the Editorial Masthead page from the main navigation
2. See error

**Expected Behaviour**
When the editorial masthead link is clicked, it should open the editorial masthead page, which contains information about past editors and peer reviewers.

**What application are you using?**
 OJS, OMP or OPS version main

**Additional information**

```
[Mon Dec 23 13:50:27 2024] [::1]:57777 [500]: GET /index.php/publicknowledge/en/about/editorialMasthead - Uncaught TypeError: PKP\userGroup\Repository::getMastheadUserIdsByRoleIds(): Argument #1 ($mastheadRoles) must be of type array, Illuminate\Database\Eloquent\Collection given, called in /Users/kaitlinnewson/Code/ojs-main/lib/pkp/pages/about/AboutContextHandler.php on line 97 and defined in /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/userGroup/Repository.php:572
Stack trace:
#0 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/pages/about/AboutContextHandler.php(97): PKP\userGroup\Repository->getMastheadUserIdsByRoleIds(Object(Illuminate\Database\Eloquent\Collection), 1)
#1 [internal function]: PKP\pages\about\AboutContextHandler->editorialMasthead(Array, Object(APP\core\Request))
#2 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/core/PKPRouter.php(329): call_user_func(Array, Array, Object(APP\core\Request))
#3 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/core/PKPPageRouter.php(253): PKP\core\PKPRouter->_authorizeInitializeAndCallRequest(Array, Object(APP\core\Request), Array, false)
#4 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/core/Dispatcher.php(158): PKP\core\PKPPageRouter->route(Object(APP\core\Request))
[...]

PR #10754, opened 2024-12-27
pkp/pkp-lib#10751 EditorialMasthead page issue with eloquent based user group update

touhidurabir:
for pkp/pkp-lib#10751

### DIFF
--- a/classes/userGroup/Repository.php
+++ b/classes/userGroup/Repository.php
@@ -220,7 +220,7 @@ class Repository
     {
         return UserUserGroup::query()
             ->withUserId($userId)
-            ->withUserGroupIds($userGroupId)
+            ->withUserGroupIds([$userGroupId])
             ->withActive()
             ->exists();
     }
--- a/classes/userGroup/Repository.php
+++ b/classes/userGroup/Repository.php
@@ -13,14 +13,17 @@
 
 namespace PKP\userGroup;
 
+use stdClass;
 use DateInterval;
 use Carbon\Carbon;
 use PKP\core\Core;
+use APP\facades\Repo;
 use PKP\plugins\Hook;
 use PKP\site\SiteDAO;
 use PKP\security\Role;
 use PKP\db\DAORegistry;
 use PKP\facades\Locale;
+use APP\core\Application;
 use PKP\xml\PKPXMLParser;
 use Illuminate\Support\Collection;
 use PKP\services\PKPSchemaService;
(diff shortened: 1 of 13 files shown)


kaitlinnewson:
Hi @touhidurabir, I believe there are a couple of changes needed in `templates/frontend/pages/editorialHistory.tpl` as well, similar to what was changed in `editorialMasthead.tpl`.

touhidurabir:
@kaitlinnewson Thanks for pointing it out . I have updated it .

classes/userGroup/Repository.php
```diff
@@ -590,12 +592,38 @@ public function getMastheadUserIdsByRoleIds(array $mastheadRoles, int $contextId
...
+                                ->on('user_user_groups.user_id', '=', 'l.user_id')
+                                ->where('l.setting_name', 'family_name')
+                                ->where('l.locale', Locale::getLocale())
+                        )
+                        ->leftJoin(
+                            'user_settings AS p',
+                            fn (JoinClause $join) => $join
+                                ->on('user_user_groups.user_id', '=', 'p.user_id')
+                                ->where('p.setting_name', 'family_name')
+                                ->where('p.locale', Locale::getPrimaryLocale())
```
bozana, line 621: I think Locale::getPrimaryLocale() would return context primary locale if possible, right? The user multilingual data do not necessarily exist in the context primary locale -- the fall back locale for required users data is the site primary locale.
Also, I think we need to consider the given_name as well i.e. the logic should be the same as in the user Collector class, because it can be that there is no family name, so we should in that case consider/order by given name (that is required and thus always there at least in the site primary locale).
I hope I understand the code correctly... :-)

bozana: Hmmm... here https://github.com/pkp/pkp-lib/pull/10667/files#diff-c8cb812ae82cc024166167807c2d42818fe8c33bc61dc597835f7dbf2d7c91b2, the user collector did not change, so maybe to use the old part of the code here, that was changed/replaced in this PR, i.e. to user the user collector to get the users? :thinking:

bozana (commented):
Hi @touhidurabir, I left a comment i.e. questions regarding getting users, grouped by mastehad roles, and ordered by their family i.e. given name.

classes/userGroup/Repository.php
bozana: Do I understand it correctly, that here both family but then also given name will be considered here: https://github.com/pkp/pkp-lib/blob/main/classes/user/Collector.php#L779-L814 -- which was used earlier... ?

touhidurabir: Right, it does consider that . I have update the PR and revert back the changes to utilize the user collector features .

touhidurabir:
@bozana I have updated the PR (check the all PRs list at https://github.com/pkp/pkp-lib/issues/10751#issue-2756499673) . Please take a look again and if all tests pass, it should be good to merge .

touhidurabir:
Also added a commit [fixed context settings access issue](https://github.com/pkp/pkp-lib/pull/10754/commits/4aaf18587f774925c36891e0710b1ec6b51ebe15) which fix another regression which cause the context setting section inaccessible .

PR #10782, opened 2025-01-08
I10670

jardakotesovec:


jonasraoni (approved):
I've just skimmed through and it seems to be ok. Hopefully no problems will arise during the pre-release tests 😎

Issue #10690, opened 2024-12-11
getLocalizedData in ModelWithSettings does not fallback when a translation is missing


kaitlinnewson:
**Describe the bug**
In `ModelWithSettings.php`, [getLocalizedData()](https://github.com/pkp/pkp-lib/blob/10e1b13ad7a8a4dcc14e0a111fc92a4a9bd33dc5/classes/core/traits/ModelWithSettings.php#L109) does not fallback to the first available piece of data when a translation is missing, so a null value may be used when a fallback is available. This is different from the behaviour in the DataObject implementation.

**To Reproduce**
1. Have a journal with multiple languages enabled
2. Enable announcements in the journal
3. Create the announcement in one language only and publish it
4. In the front-end, view the announcement and toggle between languages, noting that the title appears empty in languages where it was not translated

**What application are you using?**
 OJS, OMP or OPS version main

**Additional information**
- DataObject implementation of `getLocalizedData()` for comparison: https://github.com/pkp/pkp-lib/blob/10e1b13ad7a8a4dcc14e0a111fc92a4a9bd33dc5/classes/core/DataObject.php#L71-L78
- getBestLocalizedData(): https://github.com/pkp/pkp-lib/blob/10e1b13ad7a8a4dcc14e0a111fc92a4a9bd33dc5/classes/core/traits/LocalizedData.php#L31-L49

touhidurabir:
@Vitaliy-1 can you please review the PRs at https://github.com/pkp/pkp-lib/issues/10690#issue-2733900882 .

Vitaliy-1:
@touhidurabir, sorry for waiting. I've added a comment but can be merge as is, thanks!

touhidurabir:
fixed by https://github.com/pkp/pkp-lib/pull/10785 .

PR #10785, opened 2025-01-08
pkp/pkp-lib#10690 Allow to fallback locale for localized data on eloquent model

touhidurabir:
for pkp/pkp-lib#10690


Vitaliy-1 (approved):
Thanks, looks good!

Maybe makes sense to add also an additional parameter to force the retrieval of the key in specified locale or null. That would replace calls like `DataObject::getData($key, $locale)` in the future.

touhidurabir:
> Maybe makes sense to add also an additional parameter to force the retrieval of the key in specified locale or null. That would replace calls like DataObject::getData($key, $locale) in the future.

I think thats better . Updated the PR, will merge after the tests passes .

Issue #10562, opened 2024-10-24
Update of settings by Eloquent doesn't insert new values or delete rows with new values


Vitaliy-1:
**Describe the bug**
The current implementation of the update statement in the extended Eloquent isn't suitable for the settings data model. Considering the example:
```sql
UPDATE announcement_settings SET setting_value = 
CASE WHEN setting_name='title' AND locale='en' THEN 'English Title' WHEN setting_name='title' AND locale='fr_CA' THEN 'French TItle' ELSE setting_value END
WHERE announcement_id = 16;
``` 
It doesn't allow to remove null values of already existing settings. Considering the example, if we already have a database record:
| announcement_setting_id | announcement_id | locale  | setting_name | setting_value |
| -------------------------- | ------------------ | ------- | ------------- | -------------- |
| 1 | 1 | en | title | Title in English |
| 2 | 1  | fr_FR | title | Title in French |

Then, trying to update with 
```php
$announcement->title = [
  'fr_FR' => Title in French modified
];
$announcement->save();
```
won't remove the title in English:
| announcement_setting_id | announcement_id | locale  | setting_name | setting_value |
| -------------------------- | ------------------ | ------- | ------------- | -------------- |
| 1 | 1 | en | title | Title in English |
| 2 | 1  | fr_FR | title | Title in French modified|

**update** - The EntityDAO behaviour is to remove values explicitly set to `null`, e.g.
```php
$announcement->update(
  'title' => [
    'en' => null,
    'fr_FR' => 'Title in French modified'
  ],
)
```
Also, we need to insert new settings during update. 
We have two places in the code, where the logic is implemented:

- Old DAO: https://github.com/pkp/pkp-lib/blob/b4bb4998b136322019e59a2ccc5de439c774e7bc/classes/db/DAO.php#L432
- EntityDAO: https://github.com/pkp/pkp-lib/blob/b4bb4998b136322019e59a2ccc5de439c774e7bc/classes/core/EntityDAO.php#L190

@touhidurabir's proposal: https://github.com/pkp/pkp-lib/pull/10324/files#diff-4dcc09fe73d537128c17c4bbb260ad4ba24ccc0ecb8cdc7d0b47ebc7986e2cedR79-R129

**What application are you using?**
 OJS, OMP or OPS main branch

Vitaliy-1:
PR pkp-lib: https://github.com/pkp/pkp-lib/pull/10542
tests: https://github.com/pkp/ojs/pull/4496

[...]

Issue #10668, opened 2024-12-09
Uploaded image is not saved when creating an announcement


kaitlinnewson:
**Describe the bug**
When creating an announcement and uploading an image, the image is not being saved and the following warning is output:

```
[Mon Dec  9 10:07:30 2024] PHP Warning:  Undefined property: stdClass::$temporaryFileId in /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/announcement/Announcement.php on line 120
```

**To Reproduce**
1. Enable announcements in the journal (Settings > Website > Setup > Announcements)
2. Create a new announcement and upload an image with it
3. See the error, and note that the image is not saved to the announcement.

**What application are you using?**
 OJS, OMP or OPS main (tested in OJS)

**PRs**
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/10787
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4591 [TEST ONLY]

touhidurabir:
fixed by https://github.com/pkp/pkp-lib/pull/10787

PR #10787, opened 2025-01-09
pkp/pkp-lib#10668 Announcement image handling fixed

touhidurabir:
for pkp/pkp-lib#10668

### DIFF
--- a/classes/announcement/Announcement.php
+++ b/classes/announcement/Announcement.php
@@ -119,7 +119,7 @@ class Announcement extends Model
 
         Hook::call('Announcement::add', [$this]);
 
-        $hasNewImage = $this?->image?->temporaryFileId;
+        $hasNewImage = isset($this?->image?->temporaryFileId) ? $this->image->temporaryFileId : null;
 
         // if announcement is being inserted and includes new image, upload it
         if ($newlyCreated) {
@@ -133,6 +133,16 @@ class Announcement extends Model
         if ($hasNewImage) {
             $this->deleteImage();
             $this->handleImageUpload();
+        }
+
+        // If there is no new image and image data exists in DB and it's now removed
+        // need to delete the image for this announcement model instance
+        if (!$hasNewImage && !$this?->image && $this->fresh()->image) {
+            $this->deleteImage();
+            DB::table($this->getSettingsTable())
+                ->where($this->getkeyName(), $this->getKey())
+                ->where('setting_name', 'image')
+                ->delete();
         }
 
         return $saved;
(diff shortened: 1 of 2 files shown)


classes/services/PKPSchemaService.php
```diff
@@ -376,6 +376,11 @@ public function coerce($value, $type, $schema)
                 return $newArray;
             case 'object':
                 $newObject = []; // we handle JSON objects as assoc arrays in PHP
+                
+                if (isValidJson($value)) {
```
Vitaliy-1, line 380: Hmm, in which situation we need this?

classes/announcement/Announcement.php
```diff
@@ -135,6 +135,16 @@ public function save(array $options = [])
             $this->handleImageUpload();
         }
 
+        // If there is no new image and image data exists in DB and it's now removed
+        // need to delete the image for this announcement model instance
+        if (!$hasNewImage && !$this?->image && $this->fresh()->image) {
+            $this->deleteImage();
+            DB::table($this->getSettingsTable())
```
Vitaliy-1, line 142: I would expect that settings are removed together with the main model, isn't this happening?

Vitaliy-1 (commented):
Thanks, missed that when refactoring Announcements!
My main concern is why settings aren't removed together with the main Model (announcements). Is this some sort of edge case? Or maybe this was the case before this issue was resolved https://github.com/pkp/pkp-lib/issues/10562?

classes/announcement/Announcement.php
```diff
@@ -119,7 +119,7 @@ public function save(array $options = [])
 
         Hook::call('Announcement::add', [$this]);
 
-        $hasNewImage = $this?->image?->temporaryFileId;
+        $hasNewImage = isset($this?->image?->temporaryFileId) ? $this->image->temporaryFileId : null;
```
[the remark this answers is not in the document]
touhidurabir: me too . But for simple `stdClass` object , PHP register waring also which is better to avoid .

classes/services/PKPSchemaService.php
touhidurabir: In our case. we are getting the data as JSON but here we try to access each inner prop of JSON data as array, it handle that case we for announcement files data are not converted to JSON to array .

classes/announcement/Announcement.php
touhidurabir: This handle a case when at updating an announcement image is deleted and that is the only update .

Vitaliy-1: If it's the only case, I'm ok with it)

Vitaliy-1:
Thanks, for explaining. Looks good for merging then!

classes/announcement/Announcement.php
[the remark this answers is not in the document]
touhidurabir: > What about `$hasNewImage = $this?->image?->temporaryFileId ?? null;`

That should work .

PR #10789, opened 2025-01-09
pkp/pkp-lib#8350 Allow authors to cancel submission

taslangraham:
For pkp/pkp-lib#8350

### DIFF
--- a/templates/submission/cancelled.tpl
+++ b/templates/submission/cancelled.tpl
@@ -1,0 +1,26 @@
+{**
+ * templates/submission/cancelled.tpl
+ *
+ * Copyright (c) 2014-2022 Simon Fraser University
+ * Copyright (c) 2003-2022 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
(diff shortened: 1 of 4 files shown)


templates/submission/cancelled.tpl
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2014-2022 Simon Fraser University
+ * Copyright (c) 2003-2022 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * The page shown to the user when they cancel their submission.
+ *}
+{extends file="layouts/backend.tpl"}
+
+{block name="page"}
+	<h1 class="app__pageHeading app__pageHeading--center app__pageHeading--spacious">
```
ewhanson, line 13: There's a mix of tabs and spaces here. I believe the `.tpl` files generally use spaces.

ewhanson (changes requested):

templates/submission/cancelled.tpl
taslangraham: fixed

PR #10793, opened 2025-01-10
Remove remember on Eloquent collection

kaitlinnewson:
Removing `remember()` which does not exist in Eloquent Collection - similar to fix in https://github.com/pkp/browseBySection/pull/46/files.

### DIFF
--- a/pages/catalog/PKPCatalogHandler.php
+++ b/pages/catalog/PKPCatalogHandler.php
@@ -108,8 +108,7 @@ class PKPCatalogHandler extends Handler
             'publishedSubmissions' => $submissions->toArray(),
             'authorUserGroups' => UserGroup::withRoleIds([Role::ROLE_ID_AUTHOR])
                 ->withContextIds([$context->getId()])
-                ->get()
-                ->remember(),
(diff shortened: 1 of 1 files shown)


asmecher:
@kaitlinnewson, this is worth a closer look -- `remember()` is an Eloquent remember function:

https://laravel.com/docs/11.x/collections#method-remember

In particular, I think `remember()` was made so that the collection can be iterated more than once, which I think was necessary in this case.

kaitlinnewson:
Thanks @asmecher - my understanding was that it can only be used on `LazyCollection`, and otherwise you would have to use something like `Cache::remember`. I was discussing this with @bozana as well and we thought it may not be needed here, but I can look at some alternative solutions as well (and in the BrowseBySection plugin).

asmecher:
Ah, we probably shifted from a LazyCollection to a Collection here, mitigating the need for `remember()`.

kaitlinnewson:
Yes, previously it was a `LazyCollection` but was changed during the refactor of `UserGroup` to Eloquent.

Issue #7188, opened 2021-07-14
[OPS] Display number of current assignments for each Moderator


alexxxmendonca:
**Describe the problem you would like to solve**
As Server Admin, we want to be cautions with the load of work that we give to our moderators. We try to keep assignments as even as possible.

When selecting a moderator, it's not possible to know how many manuscripts each moderator is currenly working on. We have to either a) control that separetely in a spreadsheet or some other way; b) login in as the moderator so as to see how many open submissions they currently have; or c) filter at the "Active" submission list by moderator name.

**Describe the solution you'd like**
It would save a lot of time to have a number displayed beside the name of the moderator showing the number of current assignments for each moderator:

![image](https://user-images.githubusercontent.com/4059679/125647566-b3e1fb4c-1a0c-4f94-854f-645596349636.png)

**Who is asking for this feature?**
Server Admins trying to keep assignments as even as possible.

**Additional information**
This could be applied to OJS too when selecting Editors.

asmecher:
Merged, thanks!

JhonathanLepidus:
Thank you!

PR #10800, opened 2025-01-14
[pkp/pkp-lib#7188] Adds number of user assignments on user assignment grid

JhonathanLepidus:
Related to issue [7188](https://github.com/pkp/pkp-lib/issues/7188).

### DIFF
--- a/controllers/grid/users/userSelect/UserSelectGridCellProvider.php
+++ b/controllers/grid/users/userSelect/UserSelectGridCellProvider.php
@@ -16,11 +16,16 @@
 
 namespace PKP\controllers\grid\users\userSelect;
 
+use APP\facades\Repo;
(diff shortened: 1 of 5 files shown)


asmecher (changes requested):
@JhonathanLepidus, I think this will unintentionally/unexpectedly affect the `SubscriberSelectGridHandler` class, which is a subclass of `UserSelectGridHandler`. It should only affect the participant assignment area.

JhonathanLepidus:
Hi @asmecher , I checked this possible conflict and, luckily, it does not affect the `SubscriberSelectGridHandler`. This class inherits directly from `GridHandler`. However, it uses the class `UserSelectGridCellProvider`, but since it doesn't add the `assignments` column, it's not affected by the new code.

PR #10817, opened 2025-01-17
Feature main 10816 fix redirect loop

jonasraoni:


classes/core/PKPPageRouter.php
```diff
@@ -399,81 +400,45 @@ public function redirectHome(PKPRequest $request): void
...
-                }
+            if (in_array(Role::ROLE_ID_REVIEWER, $roleIdsArray)) {
+                return $request->url(null, 'dashboard', 'reviewAssignments');
             }
 
-            return $request->url(null, 'submissions');
-        } else {
-            // The user is at the site context
-            $userGroups = UserGroup::query()
-                ->where('context_id', \PKP\core\PKPApplication::SITE_CONTEXT_ID)
```
jonasraoni, line 448: Here the query is attempting to find a user group connected to the site context (probably only an item will be retrieved, if the user is the site administrator).

classes/core/PKPPageRouter.php
```diff
@@ -399,81 +400,45 @@ public function redirectHome(PKPRequest $request): void
...
-                                ->orWhere('date_start', '<=', now());
-                        });
-                })
-                ->get();
-
-            if ($userGroups->count() == 1) {
-                $firstUserGroup = $userGroups->first();
-                $contextDao = Application::getContextDAO();
-                $context = $contextDao->getById($firstUserGroup->contextId);
-                if (!isset($context)) {
```
jonasraoni, line 466: This is supposed to be always `true`, as there's no entry in the database for the site context.
Then there will be a redirect, and here I see two problems:
- It's already known that the user is on a site context URL (see line 446 or 405)
- This code is supposed to just retrieve a URL, it shouldn't do redirections.

asmecher (approved):

classes/core/PKPPageRouter.php
```diff
@@ -399,81 +400,45 @@ public function redirectHome(PKPRequest $request): void
...
-                }
+            if (in_array(Role::ROLE_ID_AUTHOR, $roleIdsArray)) {
+                return $request->url(null, 'dashboard', 'mySubmissions');
             }
-            return $request->url(Application::SITE_CONTEXT_PATH, 'index');
         }
-    }
 
+        return $request->url(null, 'submissions');
+    }
```
jardakotesovec, line 441: @jonasraoni When this would be hit? /submissions is not used anymore.. just the /dashboard/{mySubmissions|editorial|reviewAssignments}

jonasraoni: - The user must have:
  A. More than one role on the context
  B. Just one role, but it shouldn't be the reader role
- `[features].enable_new_submission_listing` should be disabled

jardakotesovec: @jonasraoni I see, ok. Thank you

Issue #10478, opened 2024-09-26
Protect against HTTP basic auth workaround of open redirect protections


asmecher:
**Describe the bug**
The `source` parameter is occasionally used to redirect to a URL after an action. It has protections against open redirect attacks, but those can be circumvented by using HTTP basic auth strings in the redirect string.

beraoudabdelkhalek:
a CVE ID has been assigned to this issue: CVE-2024-46326

PR #10822, opened 2025-01-18
Feature main 10818 fix redirect loop

jonasraoni:


classes/core/PKPPageRouter.php
```diff
@@ -500,40 +500,44 @@ private function _getLocaleForUrl(PKPRequest $request, ?Context $context, array
      */
     private function _setLocale(PKPRequest $request, ?string $setLocale): void
     {
-        $contextPath = $this->_getRequestedUrlParts(['Core', 'getContextPath'], $request);
-        $urlLocale = $this->_getRequestedUrlParts(['Core', 'getLocalization'], $request);
+        $contextPath = $this->_getRequestedUrlParts(Core::getContextPath(...), $request);
+        $urlLocale = $this->_getRequestedUrlParts(Core::getLocalization(...), $request);
         $multiLingual = count($this->_getContextAndLocales($request, $contextPath)[1]) > 1;
-
-        if (!$multiLingual && !$urlLocale && !$setLocale || $multiLingual && !$setLocale && $urlLocale === Locale::getLocale()) {
```
jonasraoni, line 507: I simplified the common parts, but the initial intention was just to drop the requirement of knowing the operator precedence.

classes/core/PKPPageRouter.php
```diff
@@ -500,40 +500,44 @@ private function _getLocaleForUrl(PKPRequest $request, ?Context $context, array
...
+        // Quit if there's no new locale to be set and the request URL is already well-formed
+        if (!$setLocale && ($multiLingual ? $urlLocale === Locale::getLocale() : !$urlLocale)) {
             return;
         }
 
-        $sessionLocale = (function (string $l) use ($request): string {
```
jonasraoni, line 511: I didn't see a reason to create a closure here, so I extracted the code out of it and tried to simplify.

classes/core/PKPPageRouter.php
```diff
@@ -500,40 +500,44 @@ private function _getLocaleForUrl(PKPRequest $request, ?Context $context, array
...
+        }
 
+        // Do not permit basic auth strings (user:password@url) in source parameter for redirects
         if (preg_match('#^/\w#', $source = str_replace('@', '', $request->getUserVar('source') ?? ''))) {
             $request->redirectUrl($source);
         }
 
+        $newUrlLocale = $multiLingual ? "/{$newLocale}" : '';
         $indexUrl = $this->getIndexUrl($request);
-        $uri = preg_replace("#^{$indexUrl}#", '', $setLocale ? ($_SERVER['HTTP_REFERER'] ?? '') : $request->getCompleteUrl(), 1);
```
jonasraoni, line 530: On the line 525 there's a `str_replace('@', '', $url)`, was it used to do some kind of abuse (redirect user to another system and pass credentials)?

If yes, then the `HTTP_REFERER` here should receive the same treatment, as it's also unsafe, depending on what was the issue, perhaps we should check if the destiny URL belongs to us (e.g. same domain/base path).

classes/core/PKPPageRouter.php
```diff
@@ -500,40 +500,44 @@ private function _getLocaleForUrl(PKPRequest $request, ?Context $context, array
...
-        $newUrlLocale = $multiLingual ? "/{$sessionLocale}" : '';
-        $pathInfo = ($uri)
```
jonasraoni, line 532: The problem was here, when the `$uri` is `/`, then the code attempts to do the replace, nothing is replaced and we get into a loop.

jonasraoni:
Hi @jyhein! I think you were the last person to work on this part of the code, if you can do a review, great :)

classes/core/PKPPageRouter.php
bozana: This was added here: https://github.com/pkp/pkp-lib/issues/10478.
So maybe @asmecher would know best.

bozana: I think my brain is not functioning well enough to see/think what could happen with HTTP_REFERER :sweat_smile: 
It could be a fully different web page from where the request with setLocale is coming and we would then try to replace contexPath+oldLocale with contextPath+newLocale in that URL, add that string then to $indexUrl (which is checking after our base URL) and redirect to it :thinking: 
And I just cannot figure out what 'bad' could happen thereby... so that we take good care of it... :thinking:

bozana:
Hi @jonasraoni, I reviewed your changes and I think everything is fine.
Except that I am not sure what we need to do with that HTTP_REFERER...

classes/core/PKPPageRouter.php
jonasraoni: This won't be relevant after my update, as now I'm checking if the `preg_replace()` succeeded (and redirecting to `/index/{newLocale}` if not). But it might be relevant for the `stable-3_3_0`, see: https://github.com/pkp/pkp-lib/blob/0d026553661e8a5137e1fc727cf3728eafea1327/pages/user/PKPUserHandler.inc.php#L50-L52

jonasraoni: Whatever, I'll mark it as resolved, in order for this to be risky, we should be passing extra information back to the URL (e.g. credentials).

PR #10824, opened 2025-01-19
Bugfix main 10823 standardize pkprouter initialization

jonasraoni:


classes/core/PKPRouter.php
```diff
@@ -296,10 +294,10 @@ abstract public function handleAuthorizationFailure(
      * 3) initialization
      * 4) execution
      * 5) client response
-     *
-     * @param $validate whether or not to execute the validation step.
+     * @param array{0:PKPHandler,1:string} $serviceEndpoint
+     * @param bool $validate whether or not to execute the validation step.
      */
-    public function _authorizeInitializeAndCallRequest(callable $serviceEndpoint, PKPRequest $request, array $args, bool $validate = true): void
+    public function _authorizeInitializeAndCallRequest(callable|array $serviceEndpoint, PKPRequest $request, array $args, bool $validate = true): void
```
bozana, line 300: Do we need to have `array` here?
It is maybe similar to call_user_func (s. https://www.php.net/manual/en/function.call-user-func.php) which declaration only contains `collable`, no?

bozana (commented):
Hi @jonasraoni, it looks all good to me, just one question -- I am curious... :-)

classes/core/PKPRouter.php
jonasraoni: Later on we're accessing the context of the callback with an array syntax, which isn't supported by callable (it might not even have a context) :)

bozana:
Once rebased (also OJS) it can be merged.
Thanks a lot!

Issue #10792, opened 2025-01-10
Consider new ORCID guidelines for displaying not unauthenticated ORCID iDs


bozana:
The guideline for unauthenticated ORCID iDs seem to be new -- another ORCID icon should be used in that case. Currently we do not display any icon for unauthenticated ORCID iDs, so this needs to be changed.

taslangraham:
@bozana just confirming if [this](https://info.orcid.org/documentation/integration-guide/orcid-id-display-guidelines/#:~:text=%3E-,Unauthenticated%20ORCID%20iDs,acceptable%20within%20certain%20workflows%2Fsystems.) the new guideline you are referring to.

bozana:
Yes, sorry that I forgot to insert the link to it.
Thanks @taslangraham!

taslangraham:
@ewhanson @bozana  There are two public facing pages that I'm unsure about displaying the new icon for users with an unauthenticated ORCiD - the `about/editorialMasthead` page and on the article page? Do you think it we should display the authenticated icon for user's on these pages too?

Article page
![Image](https://github.com/user-attachments/assets/502d92d7-456a-45d0-b4e2-dfbfbf2cc04d)

editorialMasthead page

![Image](https://github.com/user-attachments/assets/ceba3996-5243-4717-9c05-95ea9f259adf)

bozana:
Hi @taslangraham, for the masthead: we should only display authenticated ORCID IDs there.
I am not sure about the article view page, I would say we should consider it there, but maybe @Devika008 and @ewhanson could help?

taslangraham:
Ready for review @ewhanson 

**PRs**
pkp-lib: https://github.com/pkp/pkp-lib/pull/10829
ui-library: https://github.com/pkp/ui-library/pull/490
ojs: https://github.com/pkp/ojs/pull/4614
ops: https://github.com/pkp/ops/pull/849
omp: https://github.com/pkp/omp/pull/1815

ewhanson:
Thanks, @taslangraham! Overall looks good. Just a few questions.

ewhanson:
Thanks, @taslangraham! Looks good. 

I think OJS will need a rebase/submodule update, but feel free to merge once that's done.

taslangraham:
All merged! Thanks fro the review @ewhanson

PR #10829, opened 2025-01-21
pkp/pkp-lib#10792 update ORCiD branding

taslangraham:
For pkp/pkp-lib#10792


templates/frontend/pages/editorialMasthead.tpl
```diff
@@ -24,8 +24,8 @@
 						<span class="date_start">{translate key="common.fromUntil" from=$mastheadUser['dateStart'] until=""}</span>
 						<span class="name">
 							{$mastheadUser['user']->getFullName()|escape}
-							{if $mastheadUser['user']->getData('orcid') && $mastheadUser['user']->getData('orcidAccessToken')}
-								<span class="orcid">	
+							{if $mastheadUser['user']->getData('orcid') && $mastheadUser['user']->getData('orcidAccessToken') && $mastheadUser['user']->hasVerifiedOrcid()}
```
ewhanson, line 27: I think we should remove the check for the `orcidAccessToken` as the addition of `hasVerifiedOrcid` covers this better.

templates/frontend/pages/editorialHistory.tpl
```diff
@@ -30,7 +30,7 @@
 						</span>
 						<span class="name">
 							{$mastheadUser['user']->getFullName()|escape}
-							{if $mastheadUser['user']->getData('orcid') && $mastheadUser['user']->getData('orcidAccessToken')}
+							{if $mastheadUser['user']->getData('orcid') && $mastheadUser['user']->getData('orcidAccessToken') && $mastheadUser['user']->hasVerifiedOrcid()}
```
ewhanson, line 33: I think we should remove the check for the `orcidAccessToken` as the addition of `hasVerifiedOrcid` covers this better.

classes/identity/Identity.php
```diff
@@ -280,6 +280,19 @@ public function getOrcid()
...
+     * Return the string that should be displayed when showing a user's ORCiD
+     *
+     */
+    public function getOrcidDisplayValue(): ?string
+    {
+        if (!$this->getOrcid()) {
+            return null;
+        }
+
+        return $this->hasVerifiedOrcid() ? $this->getOrcid() : $this->getOrcid() . ' (' . __('orcid.unauthenticated') . ')';
```
ewhanson, line 293: I don't really have a strong opinion on it, but did you consider having the parenthesis as part of the locale string instead of direct formatting here? I don't know if there would be any circumstances where other languages would have issues with this formatting, but just wanted to throw it out there.

ewhanson (changes requested):

classes/identity/Identity.php
taslangraham: Don't quite remember why I decided to format it in PHP, but I've found existing strings with parenthesis in the locale  files so I've moved these to the locale string.

ewhanson (approved):

Issue #1550, opened 2016-06-23
Allow a journal to define a limited set of allowed keywords and reviewer interests


mfelczak:
Add plugin support for keyword vocabularies that JMs could define to limit the options available to both authors (indexing metadata) and reviewers (reviewer interests).

This would reduce near-duplicates and synonyms. It could also be used to provide automated suggestions when assigning reviewers, i.e. based on the submission's keywords, automatically score/suggest reviewers who include some of the same keywords in their review interests.

dennmuel:
Maybe it would be possible to reuse some library work by making it possible to upload a standardised nomenclature like the [Library of Congress Subject Headings](http://id.loc.gov/authorities/subjects.html) or the equivalents in other languages (e.g. [Gemeinsame Normdatei](http://www.dnb.de/DE/Standardisierung/GND/gnd_node.html) in Germany).

asmecher:
@dennmuel, yes, if the classification can meet three conditions:
- It uses non-language-specific identifiers,
- it's translated widely, and
- it's available under a GPL-compatible license.

dennmuel:
@asmecher , thanks for your quick reply.

I fear, I don't know one, that meets these conditions. At least to my knowledge, most countries have their own standard for library subject keywords. From the top of my head, the only classification system not necesseraily used, but at least known and translated world wide and based on numeric identifiers is the [DDC](https://en.wikipedia.org/wiki/Dewey_Decimal_Classification#Design). However, I think OCLC owns the right to it.
So a "one size fits all"-solution doesn't seem possible to me, except there is something else - maybe with regards to ontologies used in the context of semantic web / linked open data like [OWL](https://www.w3.org/TR/owl2-overview/). But I don't know much about that, just heard about it some time ago.

I was rather thinking about providing an option for the site- or journaladministrator to upload a file (of any origin in a structured format) with the subject/classification and specify, which language(s) and journal(s) these keywords are for. Then the data could be loaded (as autocompletion) or selected (via a tree-view) in the keyword field for the respective language during the submission and review process. One could theoretically reuse an existing classification or write one's own if needed.

ajnyga:
Related discussion here: https://github.com/pkp/pkp-lib/issues/1828#issuecomment-296985434
And a related plugin: https://github.com/ajnyga/finto

[...]

PR #10833, opened 2025-01-22
pkp/pkp-lib#1550 Controlled vocabulary support

jyhein:
- Use external source to add new keywords etc. using plugins
- Save extra data (identifier, source) with the term to db

### DIFF
--- a/api/v1/vocabs/PKPVocabController.php
+++ b/api/v1/vocabs/PKPVocabController.php
@@ -125,7 +125,7 @@ class PKPVocabController extends PKPBaseController
                 ->withLocales([$locale])
                 ->when(
                     $term,
-                    fn ($query) => $query->withSetting($vocab, $term, ControlledVocabEntryMatch::PARTIAL)
+                    fn ($query) => $query->withSetting('name', $term, ControlledVocabEntryMatch::PARTIAL)
                 )
                 ->get();
         } else {
@@ -133,12 +133,17 @@ class PKPVocabController extends PKPBaseController
             Hook::call('API::vocabs::getMany', [$vocab, &$entries, $illuminateRequest, response(), $request]);
         }
 
-        $data = [];
-        foreach ($entries as $entry) {
-            $data[] = $entry->getLocalizedData('name', $locale);
-        }
+        $data = collect($entries)
+            ->map(fn (ControlledVocabEntry $entry): array => $entry->getEntryData($locale))
+            ->unique(fn (array $entryData): string => 
+                ($entryData[ControlledVocabEntry::CONTROLLED_VOCAB_ENTRY_IDENTIFIER] ?? '') .
+                ($entryData[ControlledVocabEntry::CONTROLLED_VOCAB_ENTRY_SOURCE] ?? '') .
+                $entryData['name']
+            )
+            ->values()
+            ->toArray();
 
-        $data = array_values(array_unique($data));
+        Hook::run('API::vocabs::external', [$vocab, $term, $locale, &$data, &$entries, $illuminateRequest, response(), $request]);
 
         return response()->json($data, Response::HTTP_OK);
     }
--- a/classes/components/forms/FieldControlledVocab.php
+++ b/classes/components/forms/FieldControlledVocab.php
@@ -15,6 +15,8 @@
 
 namespace PKP\components\forms;
 
+use PKP\controlledVocab\ControlledVocabEntry;
+
 class FieldControlledVocab extends FieldBaseAutosuggest
 {
     /** @copydoc Field::$component */
@@ -58,7 +60,7 @@ class FieldControlledVocab extends FieldBaseAutosuggest
     {
         return [
             'value' => $value,
-            'label' => $value,
+            'label' => $value['name'] ?? $value,
         ];
     }
 }
--- a/classes/controlledVocab/ControlledVocabEntry.php
+++ b/classes/controlledVocab/ControlledVocabEntry.php
@@ -27,6 +27,9 @@ use PKP\controlledVocab\ControlledVocabEntryMatch;
 class ControlledVocabEntry extends Model
 {
     use ModelWithSettings;
+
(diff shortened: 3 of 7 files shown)


classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -27,6 +27,10 @@
 class ControlledVocabEntry extends Model
 {
     use ModelWithSettings;
+
+    public const CONTROLLED_VOCAB_ENTRY_TERM = 'term';
+    public const CONTROLLED_VOCAB_ENTRY_IDENTIFIER = 'identifier';
+    public const CONTROLLED_VOCAB_ENTRY_SOURCE = 'source';
```
touhidurabir, line 33: Any particular reason we are defining these core settings columns as const here ?

classes/controlledVocab/ControlledVocabEntry.php
```diff
@@ -27,6 +27,10 @@
 class ControlledVocabEntry extends Model
 {
     use ModelWithSettings;
+
+    public const CONTROLLED_VOCAB_ENTRY_TERM = 'term';
```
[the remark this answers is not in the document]
jyhein: Before the db was different but I kept 'term' now because outside of db it sounded  more descriptive than 'name'. No other reason. I'll change it if term is not good

classes/controlledVocab/Repository.php
```diff
@@ -93,17 +99,24 @@ public function insertBySymbolic(
...
+                        fn (array|string $vocab, int $index) => 
                             ControlledVocabEntry::create([
                                 'controlledVocabId' => $controlledVocab->id,
                                 'seq' => $index + 1,
                                 'name' => [
-                                    $locale => $vocab
+                                    $locale => $vocab[$termKey] ?? $vocab
                                 ],
+                                ...is_array($vocab)
+                                    ? Arr::whereNotNull(Arr::only($vocab, $controlledVocabEntry->getSettings()))
```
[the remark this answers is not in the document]
jyhein: indeed, forgot to use it...

classes/controlledVocab/Repository.php
```diff
@@ -58,9 +59,9 @@ public function getBySymbolic(
             )
             ->when(!empty($locales), fn ($query) => $query->withLocales($locales))
             ->get()
-            ->each(function ($entry) use (&$result) {
+            ->each(function ($entry) use (&$result, $asEntryData) {
                 foreach ($entry->name as $locale => $value) {
-                    $result[$locale][] = $value;
+                    $result[$locale][] = $asEntryData ? $entry->getEntryData($locale) : $value;
```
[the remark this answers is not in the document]
jyhein: > what about that ? or perhaps that will be too much changes as I guess need to port that changes to a lot of places .
I was thinking this too

classes/controlledVocab/ControlledVocabEntry.php
jyhein: OK, where would be a better place?

touhidurabir: I am thinking that just use these as settings columns in the `getSettings` method and not define these as const . Just wondering is there any chance to specifically use the fields by any plugin or have any special use later ? Better to use const when each of that information need to be specifically use for special purpose multiple time .

jyhein: Plugins should use these when creating suggestions for the fields. Downstream services may use 'identifier' (e.g. when it's url) but probably don't have use for 'service'. Also at some point in the future there could be more data added than just 'name', 'identifier', and 'source'. Currently core doesn't use identifier and source separately but that may change in the future

classes/components/forms/FieldControlledVocab.php
```diff
@@ -58,7 +60,7 @@ public function mapSelected($value)
     {
         return [
             'value' => $value,
-            'label' => $value,
+            'label' => $value['name'] ?? $value,
```
touhidurabir, line 63: Here at the time to loading the form, we only show the metadata as `KEYWORD` but when selecting from external source, we show it as `KEYWORD[ IDENTIFIER ]` . Just wondering if this is intentional or should it preserve the same formatting as it was before saving ? Otherwise there is no way to know if a keyword selected from a external source and not sure if that is important or not . Here and also in the `PKP\components\forms\FieldAutosuggestPreset` .

jyhein: I didn't want to do that here because plugin makers can create their own label. Maybe at some point in the future ui-lib is updated to show id and source someway, e.g. using mouse hover

api/v1/vocabs/PKPVocabController.php
```diff
@@ -125,20 +125,25 @@ public function getMany(Request $illuminateRequest): JsonResponse
...
+            ->map(fn (ControlledVocabEntry $entry): array => $entry->getEntryData($locale))
+            ->unique(fn (array $entryData): string => 
+                ($entryData[ControlledVocabEntry::CONTROLLED_VOCAB_ENTRY_IDENTIFIER] ?? '') .
+                ($entryData[ControlledVocabEntry::CONTROLLED_VOCAB_ENTRY_SOURCE] ?? '') .
+                $entryData['name']
+            )
+            ->values()
+            ->toArray();
+
+        Hook::call('API::vocabs::external', [$vocab, $term, $locale, &$data, &$entries, $illuminateRequest, response(), $request]);
```
asmecher, line 146: It's recommended to use `Hook:run` instead of `Hook:call` for new hooks. See [Improve HookRegistry hook calling conventions](https://github.com/pkp/pkp-lib/discussions/8089) for details.

Using `Hook::run` instead, the hook registrant will no longer use the `$args` array in order to receive parameters. It can declare them directly in the function definition:
```php
Hook::add('someHookName', function(string $hookName, Submission $submission, Publication $publication) {
```

asmecher (changes requested):
@jyhein, just one comment -- otherwise this is ready for merge.

Issue #10660, opened 2024-12-05
Cancelling a file revision deletes the original file


luizbgomide:
**Describe the bug**
When you begin to upload a revision to an existent file, after you upload the file if you cancel the process the original file is deleted.

This is problematic if the editor notices they pick the wrong file and data is lost.

**To Reproduce**
1. Go to Submission Files
2. Click on Upload File
3. Choose "this is a revision of an existent file" and upload a file
4. Click on Cancel

**What application are you using?**
 OJS version 3.3.0.13

taslangraham:
This was [fixed](https://github.com/pkp/pkp-lib/issues/8933) on 3.4, and needs to be backported to stable3.3.
This [fix](https://github.com/pkp/pkp-lib/issues/10705) should be applied as well.

Tribunal33:
@taslangraham can we close this or mark as duplicate?  Looks similar to #10705.

I tested the steps listed in the original comments above.  When I upload a revised file and cancel it, it no longer deletes the original file.

taslangraham:
Hey @Tribunal33 This issue is similar to #10705, but they are not the same. However, it was initially fixed on stable 3.4, as indicated in my comment above, but the fix needs to be back-ported to stable3.3 where the issue still exists. Once the fix is merged into stable-3_3_0  then this issue can be closed.

taslangraham:
Ready for review. @asmecher @Vitaliy-1 Can you take a look?
PKP-LIB: https://github.com/pkp/pkp-lib/pull/10836/files
OJS: https://github.com/pkp/ojs/pull/4628/files
OMP: https://github.com/pkp/omp/pull/1818/files
OPS: https://github.com/pkp/ops/pull/853/files

asmecher:
Added a comment, thanks!

asmecher:
If this turns out to be a lot of work for 3.3.0-x, I would be OK with deciding to just support it on newer releases.

taslangraham:
@asmecher Updated and ready for another review

PKP-LIB: https://github.com/pkp/pkp-lib/pull/10836/files

Submodule updates
OJS: https://github.com/pkp/ojs/pull/4628/files
OMP: https://github.com/pkp/omp/pull/1818/files
OPS: https://github.com/pkp/ops/pull/853/files

asmecher:
These look good, thanks; with these PRs, does this mean the issue will be corrected in 3.3.0-x, 3.4.0-x, and `main`?

[...]

PR #10836, opened 2025-01-22
pkp/pkp-lib#10660 fix:cancelling revision file deletes original

taslangraham:
For pkp/pkp-lib#10660

### DIFF
--- a/classes/services/PKPFileService.inc.php
+++ b/classes/services/PKPFileService.inc.php
@@ -116,6 +116,13 @@ class PKPFileService {
 		if ($this->fs->has($path) && !$this->fs->delete($path)) {
 			throw new Exception("Unable to delete file $id at $path.");
 		}
+
+		// Delete dependent data
+		Capsule::table('submission_file_revisions')
+			->where('file_id', '=', $id)
+			->delete();
+
+		// Delete file
 		Capsule::table('files')
 			->where('file_id', '=', $file->id)
 			->delete();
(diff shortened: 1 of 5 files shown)


classes/migration/SubmissionFilesMigration.inc.php
```diff
@@ -60,7 +60,7 @@ public function up() {
 			$table->bigInteger('submission_file_id')->unsigned();
 			$table->bigInteger('file_id')->unsigned();
 			$table->foreign('submission_file_id')->references('submission_file_id')->on('submission_files');
-			$table->foreign('file_id')->references('file_id')->on('files');
+			$table->foreign('file_id')->references('file_id')->on('files')->onDelete('cascade')->onUpdate('cascade');
```
asmecher, line 63: We aren't using foreign key constraints anywhere in the 3.3.0-x codebase -- and the DBMS might not support it. We also don't introduce database changes between builds on a stable branch. You'll have to find a way to manually implement the cascading delete, I'm afraid!

asmecher (changes requested):

classes/migration/SubmissionFilesMigration.inc.php
taslangraham: Noted. I'll make the necessary updates

taslangraham: I've updated the code. Let me know if the approach taken makes sense. 
What I did was to just delete the dependent data within the `PKPFileService::delete` method before deleting the file itself

Issue #8081, opened 2022-07-06
Address invalid localization files


jonasraoni:
**Describe the bug**
Some of our localization files might have an invalid syntax, they might work, due to lax parsers, but also cause unexpected issues with other softwares:
- Weblate might wipe out unrelated content: https://github.com/pkp/ops/commit/cc6187b39b624346d60ec9b97a901d90403ac0e1
- The `gettext` package might ignore the text: https://github.com/php-gettext/Gettext/issues/282

**Solution**
- Ensure the `.po` files are valid at build time.
- For `stable-3_3_0` the lax behavior should be kept at runtime to avoid breaking plugins.
- Fix the localization files

**What application are you using?**
OJS/main

jonasraoni:
@asmecher I've created a PR for the `pkp-lib/main` here:
- https://github.com/pkp/pkp-lib/pull/8094

I've tried to separate the commits, except the 4f379aa08daa4a8703a5b1dcf4e763feceba864d, where I used the `gettext` package to setup the plurals, it formatted the entries and also removed duplicated entries inside the same file.

I've also added a generic validator for the locale files.

jonasraoni:
For the `stable-3_3_0`, I can pass my modified validator through the locale files, and fix the breaking issues.

I also think it's ok to add the [validate-locale.sh](https://github.com/pkp/pkp-lib/pull/8094/commits/27741344eef60c2b3a38d01d827b18b3050f7eeb) in a non-breaking way (e.g. emit errors without failing the build)

asmecher:
Please hold off on this until we've reached a decision about translation workflows on this discussion: https://github.com/pkp/pkp-lib/discussions/8038

jonasraoni:
Ok! So I'll just commit fixes for the broken keys (e.g. missing quotes, bad escaping, ...).

asmecher:
Yes please -- I've recently merged translations from Weblate so now is a good time.

jonasraoni:
Just to confirm, I've updated the broken files.

jonasraoni:
@asmecher I've reviewed and moved the fixes related to duplicated keys here https://github.com/pkp/pkp-lib/issues/8119

PR #10841, opened 2025-01-24
Fix localization typos

jonasraoni:


jonasraoni:
@asmecher I've asked for a review, but ignore for now... Basically I just saw a couple of issues regarding variables, including translated variables.

I remember I had created a PR with a tool to validate the localization, which wasn't merged (that time we had locales with bad syntax), I'll check what was done there and recover whatever is useful... So, I'll update the description of this issue to include ways to avoid this problem from happening again.

jonasraoni:
Ah, there's one open issue for the old PR, so I'll re-use it: https://github.com/pkp/pkp-lib/issues/8081

Issue #9822, opened 2024-03-22
Increase timeout and improve DB table indexes for jobs ProcessUsageStatsLogFile and RemoveDoubleClicks


bozana:
The name and topic of this issue changed:
It seems the ProcessUsageStatsLogFile (that reads and simply validates the log file, removes the bot entries and inserts the entries into the usage stats temporary tables) is not the problem any more, so that it does not need to be separate into smaller processes (that would process chunks of the log file at once).
The job RemoveDoubleClicks however seems to be the problem -- it takes too long for bigger log files. The job executes this function https://github.com/pkp/pkp-lib/blob/main/classes/statistics/PKPTemporaryTotalsDAO.php#L90.
To solve this we will change the index on the table usage_stats_total_temporary_records to consider user_agent and canonical_url columns, so to be something like:
`$table->index(['load_id', 'context_id', 'ip', 'user_agent', 'canonical_url'], 'ust_load_id_context_id_ip'_ua_url);`
Also, we will increase the $timeout (and $retry_after) for this job.

PRs:
stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/10844
- https://github.com/pkp/ojs/pull/4623
- https://github.com/pkp/omp/pull/1832
- https://github.com/pkp/ops/pull/862

main:
- https://github.com/pkp/pkp-lib/pull/10896
- https://github.com/pkp/ojs/pull/4646
- https://github.com/pkp/omp/pull/1833
- https://github.com/pkp/ops/pull/863

<hr>

Original issue:
Big installations, that have huge log files and that use AcronPlugin for job processing, can experience timeouts. Thus, try to separate the ProcessUsageStatsLogFile into several smaller jobs, so that each job processes only e.g. 100 lines of the log file.

bozana:
Because the current solution works also for the huge installation, I will close this issue for now...

shantanu198713:
Hi @bozana 
I am still facing `too many times or run too long. The job may have previously timed out` issue. This has been coming to light for a long time. I execute it by manually clicking the `TryAgain` button multiple times.

![Screenshot 2024-07-09 155516](https://github.com/pkp/pkp-lib/assets/108086810/a9c2f2b9-8eed-4bf2-8b0a-03b6e5d18fd6)

```
{
    "message": "APP\\jobs\\statistics\\ProcessUsageStatsLogFile has been attempted too many times or run too long. The job may have previously timed out.",
    "code": 0,
[...]

PR #10844, opened 2025-01-25
pkp/pkp-lib#9822 Increase timeout and improve DB table indexes for usage stats jobs

bozana:
s. https://github.com/pkp/pkp-lib/issues/9822

### DIFF
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -235,7 +235,7 @@ class PKPContainer extends Container
             'driver' => 'database',
             'table' => 'jobs',
             'queue' => 'default',
-            'retry_after' => 240,
+            'retry_after' => 610,
             'after_commit' => true,
         ];
         $items['queue']['failed'] = [
@@ -316,7 +316,7 @@ class PKPContainer extends Container
      */
     protected static function getDefaultMailer(): string
     {
-        $default = Config::getVar('general', 'sandbox', false) 
+        $default = Config::getVar('general', 'sandbox', false)
             ? 'log'
             : Config::getVar('email', 'default');
 
@@ -374,7 +374,6 @@ class PKPContainer extends Container
      * Override Laravel method; always false.
      * Prevents the undefined method error when the Log Manager tries to determine the driver
      *
-     * @return bool
      */
     public function runningUnitTests(): bool
     {
--- a/classes/migration/upgrade/v3_4_0/I9822_ChangeUsageStatsTemporaryTablesIndexes.php
+++ b/classes/migration/upgrade/v3_4_0/I9822_ChangeUsageStatsTemporaryTablesIndexes.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I9822_ChangeUsageStatsTemporaryTablesIndexes.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I9822_ChangeUsageStatsTemporaryTablesIndexes
+ *
+ * @brief Consider aditional columns user_agent and canonical_url for the index on temporary usage stats tables to fix/improve the removeDoubleClicks and compileUniqueClicks query.
+ */
+
+namespace PKP\migration\upgrade\v3_4_0;
+
+use Illuminate\Database\Schema\Blueprint;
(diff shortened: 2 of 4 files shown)


classes/core/PKPContainer.php
```diff
@@ -235,7 +235,7 @@ protected function loadConfiguration()
             'driver' => 'database',
             'table' => 'jobs',
             'queue' => 'default',
-            'retry_after' => 240,
+            'retry_after' => 610,
```
jonasraoni, line 238: I left a suggestion regarding these defaults. I think we shouldn't have them, because each job might have a different need, and there's no size fits all... But I don't have anything against increasing the timeout limits/retry delay/etc 😁

jonasraoni (approved):

touhidurabir (approved):

classes/core/PKPContainer.php
darius-v: the question is why it should ever time out? if there is something wrong, it should just finish with error imo. So we should let it run however long it needs. 
As I investigated with claude, using worker without installed https://www.php.net/manual/en/book.pcntl.php will ignore timoeut setting. Did not try that, but plannign soon to try in 3.5.0.3

touhidurabir: > the question is why it should ever time out? if there is something wrong, it should just finish with error imo. So we should let it run however long it needs.

setting a `$timeout` is the best practice for a queue jobs . without a timeout, a single stuck or long-running job can block worker processes indefinitely, leading to queue backlogs, delayed processing for other jobs, and potential resource exhaustion . and the `retry_after` is what define for a crashed job how long the worker will wait till it's attempt to retry it again .

darius-v: but why it can be stuck? Isnt that a bug if it is stuck? and then instead of timeout - fix the bug? Because timeout gives timeouts when it should not give and causes problems.
Also to not stop other jobs - maybe then could be somehow made posibility to run jobs in paralel - if if one is stuck or taking long, other worked could do? I think I saw laravel worker could do many processes at once but here at least in doc it tells to set 1 process. And probably 2 processes would cause problem with statistics processing job.
But if they cause problem for statistics - then maybe could be one dedicated worker for statistics only which runs however time it needs and another worker for the rest of the jobs in paralel?

Btw I see in statistics jobs when using worker, if pcntl extension is not enabled, it will not check for timeout. And thats what I will try on prod, looks like valid case. Cause I cannot see other way to run jobs without getting timeouts. Unless I would modify values of those timeouts in vendor files.

darius-v: But if the timeout is really needed, then I should say it should be configurable, not hardcoded. Looks like laravel makes it configurable, so why not leave it configurable here? As one commentor said - there is no one size fits all.

touhidurabir: I think a general misunderstanding is going here about timeout, job retry and multi worker env 

> but why it can be stuck? Isnt that a bug if it is stuck? and then instead of timeout - fix the bug? Because timeout gives timeouts when it should not give and causes problems.

`$timeout` is needed for several reason and just not related to bug of getting stuck like
 - queue jobs starvation
 - massing resource hogging if job falls in infinity loop due to unforseen reasons
 - queue can not rely on retry mechanism safely which impacts other jobs
 - Poor observability & debugging which makes it much harder to distinguish between "working slowly" and "truly stuck."
 - hitting OS level process limit
 - too long running and massively memory/cup usage hogging one can even crash the system
 
> maybe then could be somehow made posibility to run jobs in paralel - if if one is stuck or taking long, other worked could do? 

Thats a good approach and but still does not have good safety net . what really needed here is a separate queue for long running jobs but unfortunately, we do not have such configurable mechanism yet . 

> But if they cause problem for statistics - then maybe could be one dedicated worker for statistics only which runs however time it needs and another worker for the rest of the jobs in paralel?

right approach but same infrastructural lacking . 

> in statistics jobs when using worker, if pcntl extension is not enabled, it will not check for timeout. And thats what I will try on prod, looks like valid case. Cause I cannot see other way to run jobs without getting timeouts. Unless I would modify values of those timeouts in vendor files.

> But if the timeout is really needed, then I should say it should be configurable, not hardcoded. Looks like laravel makes it configurable, so why not leave it configurable here? As one commentor said - there is no one size fits all.

not a good way to solve the issue by disabling the `pcntl` but I agree that a configurable value can name it easier for installation that may hit this issue . 

I dont think this should be available for all jobs but only few selective ones with like stats processing jobs that may end up processing massive data for very large installation -- only few specific ones . 

I will register an issue with proposal for such ones

Issue #10076, opened 2024-06-18
Submissions schema filters out declined&cancelled review assignments


jardakotesovec:
**Describe the issue**
On new submission listing we would like to indicate for editors also review assignments that has been declined or cancelled. Currently Submissions Schema filters them out - https://github.com/pkp/pkp-lib/blob/main/classes/submission/maps/Schema.php#L422

The new submission listing is using the api/_submissions endpoint .

**What application are you using?**
 OJS, OMP or OPS version 3.5

Vitaliy-1:
Editors, Authors and Reviewers have access to the new dashboard. It might have sense to separate mapping of the submissions for editors from other roles, as it includes more data, for which authors and reviewers shouldn't have access to. Declined and cancelled reviews are such an example.

My proposal is to create a separate mapping for editor's list, e.g. `Schema::mapToEditorsList()`

ewhanson:
I'm wondering if rather than a separate mapping, it would be possible to do this via some sort of permissions, e.g. an editor would have the extra review assignments included because they have permission to access them. That way we won't be maintaining multiple mappings that re similar but not exactly the same. I'll do some thinking on what that could look like and if that makes sense with the existing model we have in place.

Vitaliy-1:
> it would be possible to do this via some sort of permissions,

Vitaliy-1:
@Hafsa-Naeem, I believe that the best approach here would be to create a separate submission mapping for authors (or editors). It could be together with a new endpoint or using the same endpoint (`_submissions/assigned`) with different output depending on a role and/or additional query. I'd say that a new endpoint sounds better, @jardakotesovec, what do you think? Is it inline with your idea regarding distinguishing between author's and editor's dashboard?

jardakotesovec:
@Vitaliy-1 I don't know whats better fit how to achieve it with current architecture. But just from API point of view if there is field or information that certain role should not have access to - just don't include it.  For example in graphql where you could ask for lots of different data - common practice there would be to return null if you don't have permissions to access something.

Vitaliy-1:
In the submission schema we have the method that defines mapping `mapManyToSubmissionsList()` and `mapToSubmissionsList()`. The latter has the line
```php
return $this->mapByProperties($this->getSubmissionsListProps(), $item, $anonymizeReviews);
```
[...]

PR #10847, opened 2025-01-27
pkp/pkp-lib#10076 show declined/cancelled reviews to editors

Hafsa-Naeem:
for #10076

### DIFF
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -89,6 +89,13 @@ interface iPKPApplicationInfoProvider
      * @hook PKPApplication::execute::catch ['throwable' => $t]
      */
     public static function getContextAssocType(): int;
+
+    /**
+     * Get the review workflow stages used by this application.
+     *
+     * @hook PKPApplication::execute::catch ['throwable' => $t]
+     */
+    public function getReviewStages(): array;
 }
 
 abstract class PKPApplication implements iPKPApplicationInfoProvider
@@ -454,6 +461,11 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
      * application.
      */
     abstract public function getVersionDescriptorUrl(): string;
+
+    /**
+     * Get the review workflow stages used by this application.
+     */
+    abstract public function getReviewStages(): array;
 
     /**
      * This function retrieves all enabled product versions once
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -92,8 +92,6 @@ interface iPKPApplicationInfoProvider
 
     /**
      * Get the review workflow stages used by this application.
-     *
-     * @hook PKPApplication::execute::catch ['throwable' => $t]
      */
     public function getReviewStages(): array;
 }
@@ -439,6 +437,14 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
     }
 
     /**
+     * Get the review workflow stages used by this application.
+     */
+    public function getReviewStages(): array
+    {
+        return [];
+    }
+
+    /**
      * Get the symbolic name of this application
      */
     public static function getName(): string
@@ -461,11 +467,6 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
      * application.
      */
     abstract public function getVersionDescriptorUrl(): string;
-
-    /**
-     * Get the review workflow stages used by this application.
-     */
-    abstract public function getReviewStages(): array;
 
     /**
      * This function retrieves all enabled product versions once
--- a/classes/submission/maps/Schema.php
+++ b/classes/submission/maps/Schema.php
@@ -528,14 +528,31 @@ class Schema extends \PKP\core\maps\Schema
      */
     protected function getPropertyReviewAssignments(Enumerable $reviewAssignments, bool|Collection $anonymizeReviews = false): array
     {
+        $request = Application::get()->getRequest();
+        $currentUser = $request->getUser();
+        $currentUserRoleIds = array_map(
+            function (Role $role) {
+                return $role->getId();
+            },
+            $currentUser->getRoles($this->context->getId())
+        );
+
+        // is the user a manager, sub-editor, or site admin?
+        $canSeeAllReviewAssignments = (
+            in_array(Role::ROLE_ID_MANAGER, $currentUserRoleIds) ||
+            in_array(Role::ROLE_ID_SUB_EDITOR, $currentUserRoleIds) ||
+            in_array(Role::ROLE_ID_SITE_ADMIN, $currentUserRoleIds)
+        );
         $reviews = [];
         $request = Application::get()->getRequest();
         $currentUser = $request->getUser();
 
         foreach ($reviewAssignments as $reviewAssignment) {
-            // @todo for now, only show reviews that haven't been
-            // declined or cancelled
-            if ($reviewAssignment->getDeclined() || $reviewAssignment->getCancelled()) {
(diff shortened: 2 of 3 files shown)


classes/submission/maps/Schema.php
```diff
@@ -421,16 +421,31 @@ protected function mapByProperties(array $props, Submission $submission, bool|Co
...
+            function (Role $role) {
+                return $role->getId();
+            },
+            $currentUser->getRoles($this->context->getId())
+        );
+
+        // is the user a manager, sub-editor, or site admin?
+        $canSeeAllReviewAssignments = (
+            in_array(Role::ROLE_ID_MANAGER, $currentUserRoleIds) ||
+            in_array(Role::ROLE_ID_SUB_EDITOR, $currentUserRoleIds) ||
```
Vitaliy-1, line 436: We shouldn't include here submission editors not assigned to the submission. And probably managers and admins assigned into the submission in the author role or reviewer role. Or assigned to the other stage than the review. 

We already calculated roles in the stages property: https://github.com/pkp/pkp-lib/blob/aac75ee578cc834b8d2cd99cd07071ba5098a45a/schemas/submission.json#L253

Using it you can actually check if the user has those roles, it includes global ones and those related to the assignment.  

And also you can use `array_intersect()` instead of `in_array`.

Vitaliy-1 (changes requested):
Thanks! I've added a comment regarding global roles and assignment roles.

classes/submission/maps/Schema.php
```diff
@@ -526,16 +526,30 @@ protected function canChangeMetadata(Enumerable $stageAssignments, Submission $s
...
+    protected function getPropertyReviewAssignments(Enumerable $reviewAssignments, bool|Collection $anonymizeReviews = false, Submission $submission): array
     {
-        $reviews = [];
         $request = Application::get()->getRequest();
         $currentUser = $request->getUser();
 
+        // current user's assigned roles from the submission's stages
+        $currentUserAssignedRoles = [];
+        foreach ($this->getPropertyStages($this->stageAssignments, $submission, $this->decisions ?? null, null) as $stage) {
+            if (isset($stage['currentUserAssignedRoles'])) {
```
Vitaliy-1, line 537: We are looking here for roles at the specific stage. For OJS it's `WORKFLOW_STAGE_ID_EXTERNAL_REVIEW`, for OMP it's `WORKFLOW_STAGE_ID_EXTERNAL_REVIEW` and `WORKFLOW_STAGE_ID_INTERNAL_REVIEW`

Vitaliy-1 (changes requested):
Thanks, Hafsa. I've added two minor comments

classes/submission/maps/Schema.php
```diff
@@ -526,16 +526,39 @@ protected function canChangeMetadata(Enumerable $stageAssignments, Submission $s
...
-    protected function getPropertyReviewAssignments(Enumerable $reviewAssignments, bool|Collection $anonymizeReviews = false): array
+    protected function getPropertyReviewAssignments(Enumerable $reviewAssignments, bool|Collection $anonymizeReviews = false, Submission $submission, array $stages): array
     {
-        $reviews = [];
         $request = Application::get()->getRequest();
         $currentUser = $request->getUser();
+    
+        // application specific review stages
+        $reviewStageIds = [WORKFLOW_STAGE_ID_EXTERNAL_REVIEW];
+        if (Application::get()->getName() === 'omp') {
```
Vitaliy-1, line 536: It's better to find or modify existing methods to retrieve review stages. E.g., I see the protected method in the [submission collector](https://github.com/pkp/pkp-lib/blob/9ac7230eeaf311d528f8d2e359921e0d2fba9884/classes/submission/Collector.php#L916). We could make it public and move to the PKPApplication/Application class.

classes/submission/maps/Schema.php
```diff
@@ -526,16 +526,39 @@ protected function canChangeMetadata(Enumerable $stageAssignments, Submission $s
...
+        // define the roles that can see declined or cancelled review assignments
+        $allowedRoles = [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_SITE_ADMIN];
+        $canSeeAllReviewAssignments = !empty(array_intersect($currentUserAssignedRoles, $allowedRoles));
+        $reviews = [];
         foreach ($reviewAssignments as $reviewAssignment) {
-            // @todo for now, only show reviews that haven't been
-            // declined or cancelled
-            if ($reviewAssignment->getDeclined() || $reviewAssignment->getCancelled()) {
+            // declined or cancelled are only show to editors, sub-editors and site admin
+            if (
```
Vitaliy-1, line 558: In theory, the user can have access to the internal but not external review stage (and vice versa), thus, shouldn't see declined or cancelled assignments for the latter stage. It's better to deal with review assignments from different stages separately. 
E.g., you can create something like:
```php
  if (!$this->canSeeAllReviewAssignments($reviewAssignment, $stages) && ...) {
  ...
  }
}
...
protected function canSeeAllReviewAssignments (ReviewAssignment $reviewAssignment, array $stages): bool
{
  $stageId = $reviewAssignment->getStageId();
  if (!isset($stages[$stageId]['currentUserAssignedRoles'])) {
    return false;
  }
  
  $roles = $stages[$stageId]['currentUserAssignedRoles'];
  if (array_intersect($roles, [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_SITE_ADMIN])) {
    return true;
  }
  
  return false;
}
```
(very rough)

Vitaliy-1 (changes requested):
Thanks, Hafsa! I left comments for changes to finalise the PR.

classes/submission/maps/Schema.php
```diff
@@ -526,16 +526,33 @@ protected function canChangeMetadata(Enumerable $stageAssignments, Submission $s
...
     {
-        $reviews = [];
         $request = Application::get()->getRequest();
         $currentUser = $request->getUser();
 
+        // application specific review stages
+        $reviewStageIds = Application::get()->getReviewStages();
+
+        // get roles from review stages only
+        $currentUserAssignedRoles = [];
```
Vitaliy-1, line 538: Is this variable used anywhere? It looks like we can remove it and the `foreach` block which retrieves it.

classes/core/PKPApplication.php
```diff
@@ -89,6 +89,13 @@ public static function getFileDirectories(): array;
      * @hook PKPApplication::execute::catch ['throwable' => $t]
      */
     public static function getContextAssocType(): int;
+
+    /**
+     * Get the review workflow stages used by this application.
+     *
+     * @hook PKPApplication::execute::catch ['throwable' => $t]
```
Vitaliy-1, line 96: This comment likely was added by a mistake

classes/core/PKPApplication.php
```diff
@@ -455,6 +462,11 @@ abstract public function getContextName(): string;
      */
     abstract public function getVersionDescriptorUrl(): string;
 
+    /**
+     * Get the review workflow stages used by this application.
+     */
+    abstract public function getReviewStages(): array;
```
Vitaliy-1, line 468: We can remove `abstract` and make this method returning an empty array by default as OPS doesn't have any.
Also we would need to override it on the application level (OJS and OMP).

Vitaliy-1 (changes requested):
Thanks!I left a couple comments to improve it

Issue #10859, opened 2025-01-30
Warnings logged when Invitation Added


defstat:
**Describe the bug**
There are some warnings appearing in the PHP logs like 
```
PHP Warning:  Undefined array key "userId" in ojs-main/lib/pkp/api/v1/invitations/InvitationController.php on line 260
``` 

**To Reproduce**
1. Go to 'setting->Users & Roles'
2. Click on 'Invite to a role'
3. Add an email that does not belong to an existing user. 
4. Click on 'Search User'
5. See the warning in PHP error log

**What application are you using?**
 OJS, OMP or OPS  `main` branch

---
**PR**

OJS: pkp/ojs#4674 **TESTS ONLY**
pkp-lib: #10860

defstat:
@ewhanson this is a minor change in the API, are you able to do a code review on it?

asmecher:
Merged, thanks!

PR #10860, opened 2025-01-30
[PKP-LIB][main] #10859 Fix warning related to 'userId' and 'inviteeEmail' when adding an invitation

defstat:


ewhanson (approved):
Looks good, @defstat!

asmecher (changes requested):
Suggest using `??` operator instead of `Arr::get`; the `Arr::get` is mostly for the sake of supporting Laravel's dot notation, which is overkill here.

Issue #10795, opened 2025-01-13
create a new function toBcp47()


bozana:
The following code is used lots throughout the code `str_replace(['_', '@'], '-', a_locale)`. Thus implement a function that will replace this code.

PRs:
- https://github.com/pkp/pkp-lib/pull/10887
- https://github.com/pkp/oaiJats/pull/66
- https://github.com/pkp/jatsTemplate/pull/81
- https://github.com/pkp/googleScholar/pull/26
- https://github.com/pkp/crossref-ojs/pull/60
- https://github.com/pkp/webFeed/pull/11
- https://github.com/pkp/ojs/pull/4644
- https://github.com/pkp/omp/pull/1830
- https://github.com/pkp/ops/pull/860
- https://github.com/pkp/crossref-ops/pull/46

bozana:
Hi @jonasraoni, would you maybe like to take a look at the PRs above?
Thanks a lot!

bozana:
Thanks a lot @kaitlinnewson!

PR #10887, opened 2025-02-05
pkp/pkp-lib#10795 introduce LocaleConversion::toBcp47()

bozana:
s. https://github.com/pkp/pkp-lib/issues/10795

### DIFF
--- a/classes/i18n/LocaleConversion.php
+++ b/classes/i18n/LocaleConversion.php
@@ -198,4 +198,12 @@ class LocaleConversion
         $localeMetadata = new LocaleMetadata($locale);
         return $localeMetadata->getIsoAlpha2();
     }
+
+    /**
+     * Convert Weblate locale to IETF BCP 47 language tag.
+     */
+    public static function toBcp47(string $locale): string
+    {
+        return str_replace(['_', '@'], '-', $locale);
+    }
 }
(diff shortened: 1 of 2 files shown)


classes/i18n/LocaleConversion.php
```diff
@@ -198,4 +198,12 @@ public static function getIso1FromLocale(string $locale): ?string
         $localeMetadata = new LocaleMetadata($locale);
         return $localeMetadata->getIsoAlpha2();
     }
+
+    /**
+     * Convert Weblate locale to IETF BCP 47 language tag.
+     */
+    public static function toBcp47(string $locale): string
+    {
+        return str_replace(['_', '@'], '-', $locale);
```
jonasraoni, line 207: If you believe this can create problems somewhere, then:

```suggestion
    public static function toBcp47(?string $locale): string
    {
        return str_replace(['_', '@'], '-', $locale ?? '');
```

jonasraoni (approved):

classes/i18n/LocaleConversion.php
bozana: I think it is OK to request the string. I already find out a few wrong things in our jatsTemplate tests -- an article and publication should always have 'locale'. Thus I will leave it as it is for now...

Issue #10884, opened 2025-02-05
Submission Workflow - Does not save issues form when a cover image file uploaded


Tribunal33:
**Describe the bug**
During the submission work flow.  Trying to save the form after uploading a cover image file will throw an error in console.  PHP has a log as well.
Preconditions : None, doesn't appear to be selective of the stage you are in

**To Reproduce**
1. Login as Admin or JM 
2. From the side navigation choose a view with many submissions
3. Select View on a submission from the list in a valid stage (submission, review etc..)
4. Navigate to Publication -> issues from side navigation
5. Proceed to fill out the form with valid info
a. Sections (Articles) from dropdown
b. Categories - Any
c. Cover image - Any image file (png for test)
d. Pages 4
e. URL path and Date published left blank
6. Have dev Console open to see error message
7. Select Save

Expected: Form is saved with valid messages (see success response in screenshot below).
Actual: Form thows an error in console and logs (see log info and 500 error in screenshot below).
**What application are you using?**
 OJS

**Additional information**
Error log from apache : [Tue Feb 04 11:28:40.445891 2025] [php:notice] [pid 27000:tid 1872] [client ::1:55153] TypeError: json_decode(): Argument #1 ($json) must be of type string, array given in C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\includes\\functions.php:238\nStack trace:\n#0 C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\includes\\functions.php(238): json_decode(Array)\n#1 C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\classes\\services\\PKPSchemaService.php(380): isValidJson(Array)\n#2 C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\classes\\services\\PKPSchemaService.php(323): PKP\\services\\PKPSchemaService->coerce(Array, 'object', Object(stdClass))\n#3 C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\classes\\core\\EntityDAO.php(196): PKP\\services\\PKPSchemaService->sanitize('publication', Array)\n#4 

![Image](https://github.com/user-attachments/assets/8e7b2a67-f806-46fa-82b8-c1f869f251a5)

taslangraham:
I've added a PR for this.

**PRs**
pkp-lib: https://github.com/pkp/pkp-lib/pull/10889
[...]

PR #10889, opened 2025-02-05
pkp/pkp-lib#10884 Only json_decode strings in PKPSchemaService::coerce()

taslangraham:
For pkp/pkp-lib#10884

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -185,7 +185,7 @@ function __p(string $key, int $number, array $replace = [], ?string $locale = nu
 
 /**
  * Check if run on CLI
- * 
+ *
  * @deprecated 3.5.0 use PKPContainer::getInstance()->runningInConsole()
  */
 if (!function_exists('runOnCLI')) {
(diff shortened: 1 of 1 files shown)


classes/services/PKPSchemaService.php
```diff
@@ -376,8 +378,8 @@ public function coerce($value, $type, $schema)
                 return $newArray;
             case 'object':
                 $newObject = []; // we handle JSON objects as assoc arrays in PHP
-                
-                if (isValidJson($value)) {
+
+                if (!is_object($value) && !is_array($value) && isValidJson($value)) {
```
touhidurabir, line 382: Hmm , since PHP 8+ , it got much stricter with type checking . The point is that `json_decode` now require first param to be `string` or throw fatal error while before PHP 8, it only throw a warning . How about rather than putting the check here , we put the check in the `isValidJson` method as such 
```php
if (!empty($data) && is_string($data)) { 
```
That way other also don't have be concern with these checks before using the `isValidJson` .

touhidurabir (commented):
@taslangraham I added a review , please take a look .

classes/services/PKPSchemaService.php
taslangraham: I've updated the code and will go ahead and merge this

PR #10895, opened 2025-02-06
pkp/pkp-lib#9658 changes in invitation and user access table

ipula:
see https://github.com/pkp/pkp-lib/issues/10916

### DIFF
--- a/locale/en/grid.po
+++ b/locale/en/grid.po
@@ -508,10 +508,10 @@ msgid "grid.user.email"
 msgstr "Email"
 
 msgid "grid.user.enable"
-msgstr "Enable"
+msgstr "Enable User"
 
 msgid "grid.user.disable"
-msgstr "Disable"
+msgstr "Disable User"
 
 msgid "grid.user.disableReason"
 msgstr "Reason for disabling user"
@@ -716,3 +716,17 @@ msgstr "In Browse Lists"
 
 msgid "grid.action.editContributor"
 msgstr "Edit Contributor"
+
+msgid "grid.user.remove"
+msgstr "Remove User"
+
+msgid "grid.user.logInAs"
+msgstr "Login As"
+
+msgid "grid.user.enableReasonDescription"
+msgstr "Once the user is enabled, they will regain access to the site, and you'll be able to invite them to roles as needed."
+
+msgid "grid.user.disableReasonDescription"
+msgstr "Please note that once a user is disabled, you won't be able to add them to any roles until the user is enabled again."
+
+
--- a/locale/en/invitation.po
+++ b/locale/en/invitation.po
@@ -134,16 +134,10 @@ msgid "userInvitation.searchUser.nextButtonLabel"
 msgstr "Search User"
 
 msgid "userInvitation.searchUser.stepDescription"
-msgstr "Search for the user using their email address, username or ORCID ID. Enter at least one details to get started. If the user does not exist, you can invite them to take up roles and be a part of your journal. If the user already exist in the system, you can view user information and invite to take a additional roles."
+msgstr "If the user does not exist, you can invite them to take on roles. If the user already exists in the system, you can view their information and invite them to take on additional roles."
 
-msgid "userInvitation.emailField.description"
-msgstr "e.g. aeinstein@example.com"
-
-msgid "userInvitation.usernameField.description"
-msgstr "e.g. mickeymouse"
-
-msgid "userInvitation.orcidField.description"
-msgstr "e.g. 0000-0000-0000-0000"
+msgid "userInvitation.searchField.description"
+msgstr "e.g. aeinstein@example.com or aeinstein or 0000-0002-1825-0097"
 
 msgid "userInvitation.enterDetails.stepName"
 msgstr "Enter details"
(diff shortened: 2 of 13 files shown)


locale/en/grid.po
```diff
@@ -716,3 +716,17 @@ msgstr "In Browse Lists"
...
 msgstr "Edit Contributor"
+
+msgid "grid.user.remove"
+msgstr "Remove User"
+
+msgid "grid.user.logInAs"
+msgstr "Login As"
+
+msgid "grid.user.enableReasonDescription"
+msgstr "Once the user enabled, they will regain access to OJS, and you'll be able to invite them to roles as needed."
```
blesildaramirez, line 727: Like in ui-library, let's add the missing "is" here:
```suggestion
msgstr "Once the user is enabled, they will regain access to OJS, and you'll be able to invite them to roles as needed."
```

locale/en/grid.po
```diff
@@ -716,3 +716,17 @@ msgstr "In Browse Lists"
...
+msgstr "Remove User"
+
+msgid "grid.user.logInAs"
+msgstr "Login As"
+
+msgid "grid.user.enableReasonDescription"
+msgstr "Once the user is enabled, they will regain access to OJS, and you'll be able to invite them to roles as needed."
+
+msgid "grid.user.disableReasonDescription"
+msgstr "Please note that once a user is disabled, you won't be able to add them to any roles until they are enabled again."
```
ewhanson, line 730: To me the language here is a little ambiguous. I would suggest `"Please note that once a user is disabled, you won't be able to add them to any roles until the user is enabled again."`

locale/en/invitation.po
```diff
@@ -134,16 +134,10 @@ msgid "userInvitation.searchUser.nextButtonLabel"
 msgstr "Search User"
 
 msgid "userInvitation.searchUser.stepDescription"
-msgstr "Search for the user using their email address, username or ORCID ID. Enter at least one details to get started. If the user does not exist, you can invite them to take up roles and be a part of your journal. If the user already exist in the system, you can view user information and invite to take a additional roles."
+msgstr "If the user does not exist, you can invite them to take on roles and become part of your journal. If the user already exists in the system, you can view their information and invite them to take on additional roles."
```
ewhanson, line 137: This references journals specifically, but shouldn't reference the specific context type since this is in pkp-lib.

locale/en/invitation.po
```diff
@@ -134,16 +134,10 @@ msgid "userInvitation.searchUser.nextButtonLabel"
...
-msgid "userInvitation.emailField.description"
-msgstr "e.g. aeinstein@example.com"
-
-msgid "userInvitation.usernameField.description"
-msgstr "e.g. mickeymouse"
-
-msgid "userInvitation.orcidField.description"
-msgstr "e.g. 0000-0000-0000-0000"
+msgid "userInvitation.searchField.description"
+msgstr "e.g. aeinstein@example.com or aeinstein or 0000-0000-0000-0000"
```
ewhanson, line 140: I wonder if we should use `0000-0002-1825-0097` as the ORCID example, one provided by ORCID as an official example that includes a valid checksum and is a valid ORCID (see: https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier)

locale/en/invitation.po
```diff
@@ -372,3 +366,14 @@ msgstr "Invitation not accepted. You're logged in as a different user."
...
 msgstr "Please log out and sign in with the correct account to accept this invitation."
+
+msgid "userInvitation.searchField"
+msgstr "Search for a user by email address, username, or ORCID iD. Enter only one to get started!"
+
+msgid "userInvitation.user.disableTitle"
+msgstr "The user is currently disabled."
+
+msgid "userInvitation.user.disableMessage"
+msgstr "The user was disabled. You cannot assign them a role while they are disabled. Please enable the user first to invite them to a  role."
```
ewhanson, line 377: Extra space between `a` and `role`.

locale/en/user.po
```diff
@@ -852,3 +852,18 @@ msgstr "This ORCID has not been verified. Please remove this unverified ORCID an
 
 msgid "user.removeRole.message"
 msgstr "Are you sure want remove this role permanently?"
+
+msgid "user.futureRole.notification.message"
+msgstr "Your role is scheduled to begin on "
```
ewhanson, line 857: I wonder if the scheduled begin date should be a variable in the locale key it is for others directly below. It would make it clear what goes here as well as allow for other languages where the date would not appear at the end to insert it as appropriate.

locale/en/user.po
```diff
@@ -852,3 +852,18 @@ msgstr "This ORCID has not been verified. Please remove this unverified ORCID an
...
+
+msgid "user.futureRole.notification.description"
+msgstr "Until then, you can review and update your profile. If you believe this is an error, please contact journal administrator."
```
ewhanson, line 860: Should not explicitly reference journal since this will apply to OMP and OPS as well.

ewhanson (changes requested):
Thanks @ipula! Looking really good, just a few questions and small corrections.

ewhanson (changes requested):
Thanks @ipula, just one additional reference to OJS I found, but otherwise looks good!

withanage:
@ewhanson   if you can confirm this , we can mark this as resolved for merging.

ewhanson (approved):
Looks good. Thanks @ipula. It just needs to be squashed into a single commit then it's ready for merge. cc: @withanage.

withanage:
Thanks @ewhanson for reviewing, tomorrow we will merge after squashing the commits to one

Issue #10897, opened 2025-02-06
Author affiliation migration does not explicitelly consider setting_name = affiliation


bozana:
When migrating author affiliations to new structure with multiple author affiliations and RORs also the author names are considered and the rors database search using them.

PRs:
- https://github.com/pkp/pkp-lib/pull/10898
~~- https://github.com/pkp/ojs/pull/4647 (only submodule udpate)~~

PR #10898, opened 2025-02-06
pkp/pkp-lib#10897 fix author affiliation migration to rors

bozana:
s. https://github.com/pkp/pkp-lib/issues/10897


GaziYucel:
There are two beauty changes, in the phpdoc of the migration methods. In there I wrote the used sql, which still has the "r_s" instead of the "rs" syntax for the table aliases.  

Other than that, I agree to be merged.

Issue #10858, opened 2025-01-30
Implement additional submissions list views: reviews overdue and author revisions submitted


Vitaliy-1:
Original submissions list mockups included additional views:

- Action required by me
- Reviews overdue
- Author revisions submitted

The later two is decided to add for 3.5. 

**Reviews overdue** requires additional filter to the submission collector: https://github.com/pkp/pkp-lib/blob/main/classes/submission/Collector.php#L718. One of the possibility is using `ReviewRound::REVIEW_ROUND_STATUS_REVIEWS_OVERDUE` check in the database but the status should be recalculated first. The other (and probably better) option is to construct a query that compares dates with the current.

**Author revisions submitted** already has an associated filter and should be easy to implement: https://github.com/pkp/pkp-lib/blob/main/classes/submission/Collector.php#L882

To implement views see Submission Repository: https://github.com/pkp/pkp-lib/blob/main/classes/submission/Repository.php#L847 and the Controller: https://github.com/pkp/pkp-lib/blob/main/api/v1/_submissions/PKPBackendSubmissionsController.php#L522

Vitaliy-1:
PR
pkp-lib: https://github.com/pkp/pkp-lib/pull/10919
tests: https://github.com/pkp/ojs/pull/4653

Vitaliy-1:
That appeared as a quite small PR. @jardakotesovec, can you take a look if everything looks as expected?

Vitaliy-1:
That should also solve this one: https://github.com/pkp/pkp-lib/issues/10835, OMP should apply review filters correctly...

PR #10919, opened 2025-02-10
pkp/pkp-lib#10858 reviews overdue and author revisions submitted views

Vitaliy-1:


jardakotesovec:
Can't comment directly on the code as the actual filters were added in past.

I wonder whether the currentReviewRound which is important for several of the views will work as expected here for OMP as it seems not differentiating between internal and external review rounds?

```
        // Aggregate current review round number, don't include review assignments in non-relevant rounds
        $currentReviewRound = DB::table('review_rounds', 'rr')
            ->select('rr.submission_id')
            ->selectRaw('MAX(rr.round) as current_round')
            ->groupBy('rr.submission_id');

```

classes/submission/Repository.php
```diff
@@ -961,6 +961,19 @@ protected function mapDashboardViews(Collection $types, Context $context, User $
                         'reviews',
                         ['reviewsSubmitted' => true]
                     );
+                case DashboardView::TYPE_REVIEWS_OVERDUE:
+                    $collector = Repo::submission()->getCollector()
+                        ->filterByContextIds([$context->getId()])
+                        ->filterByOverdue(true)
+                        ->filterByStatus([PKPSubmission::STATUS_QUEUED]);
+                    return new DashboardView(
```
jardakotesovec, line 969: This one is quite intense :-). There is good chance I am not interpreting things correctly..  so my concerns might be off.

I wonder whether we should filter by review stage? Otherwise even if we leave some overdue review in review stage and move it to editorial it still would show up?

Similar concern about current review round - should not we checking just that?

jardakotesovec: Oh these concerns are probably covered with this part, right?
```
                    ->whereNotIn('rr.status', [
                        ReviewRound::REVIEW_ROUND_STATUS_RESUBMIT_FOR_REVIEW,
                        ReviewRound::REVIEW_ROUND_STATUS_SENT_TO_EXTERNAL,
                        ReviewRound::REVIEW_ROUND_STATUS_ACCEPTED,
                        ReviewRound::REVIEW_ROUND_STATUS_DECLINED,
                    ])
```

Vitaliy-1: My initial change included a new filter, I can re-add that.

PR #10955, opened 2025-02-19
Remove journal and OJS references from ORCID locale strings

ewhanson:


ewhanson:
@asmecher, could you have a look at these? Thanks!

locale/en/user.po
```diff
@@ -750,13 +751,13 @@ msgstr ""
 "\tLearn more in <a href=\"https://info.orcid.org/"
 "whats-so-special-about-signing-in/\">What’s so special about signing "
 "in.</a><br>\n"
-"This journal will collect and display authenticated authors’ and coauthors’ "
-"iDs on the OJS profile and article page."
+"This site will collect and display authenticated authors’ and coauthors’ "
+"iDs on the user profile and publication landing page."
```
asmecher, line 755: We have to be a little cautious with "publication" as preprints are posted not published. Maybe "and where authorship is listed"?

asmecher (changes requested):
Just a couple suggestions!

locale/en/user.po
ewhanson: Thanks! I was looking for something like that!

Issue #9717, opened 2024-02-08
Resolve template injection risks in Smarty/Vue interactions


asmecher:
**Describe the bug**
https://github.com/pkp/pkp-lib/issues/9650 and https://github.com/pkp/pkp-lib/issues/9421 resolved the risk of template injections in Smarty/Vue interactions with an expedient but inelegant approach to escaping Smarty variables. For the `main` branch we'd like to do something cleaner. 

Meanwhile the fixes for #9650 and #9421 **have not** been merged to `main`!

See related: https://github.com/pkp/pkp-lib/discussions/9683

----
**Pull requests:**
For HTML escaping:
ui-library: https://github.com/pkp/ui-library/pull/528
pkp-lib: https://github.com/pkp/pkp-lib/pull/10944
ojs: https://github.com/pkp/ojs/pull/4669
omp: https://github.com/pkp/omp/pull/1844
ops: https://github.com/pkp/ops/pull/878

For resolving template injection:
pkp-lib: https://github.com/pkp/pkp-lib/pull/10959
ojs (_tests only_): https://github.com/pkp/ojs/pull/4680
omp (_tests only_): https://github.com/pkp/omp/pull/1849
ops (_tests only_): https://github.com/pkp/ops/pull/881

PR #10959, opened 2025-02-21
pkp/pkp-lib#9717 (main) Resolve template injection with context name

blesildaramirez:


jardakotesovec:
Hi @asmecher, just ping here, that we would like to go with same general approach as for 3.3 and 3.4. I think its still the safest option.

blesildaramirez:
Hi @jardakotesovec , @asmecher , thanks! I will merge this now.

Issue #10977, opened 2025-02-25
Merge Users Tool - Fatal error: Call to undefined method


kaitlinnewson:
**Describe the bug**
The CLI "Merge Users Tool" is currently failing with a fatal error (see log below). This may be a regression from https://github.com/pkp/pkp-lib/issues/10129 (cc @taslangraham).

**To Reproduce**
1. In the command-line, navigate to the OJS application directory (e.g. `cd my/ojs/install`)
2. Run the merge users tool, which merges the second account passed into the first (e.g. `php tools/mergeUsers.php admin fpaglieri`
3. See error

**Expected Behaviour**: The tool should merge all data from the second account into the first, removing the second user account.

**What application are you using?**
 OJS 3.5/main

**Additional information**

```
PHP Fatal error:  Uncaught BadMethodCallException: Call to undefined method Illuminate\Database\Query\Builder::as() in /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:67
Stack trace:
#0 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(4439): Illuminate\Database\Query\Builder::throwBadMethodCallException('as')
#1 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/classes/log/Repository.php(112): Illuminate\Database\Query\Builder->__call('as', Array)
#2 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(375): PKP\log\Repository->PKP\log\{closure}(Object(Illuminate\Database\Query\Builder))
#3 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(1209): Illuminate\Database\Query\Builder->createSub(Object(Illuminate\Database\Query\Builder))
#4 /Users/kaitlinnewson/Code/ojs-main/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(1259): Illuminate\Database\Query\Builder->whereIn('email_log_id', Object(Closure), 'and', true)
[...]

PR #10984, opened 2025-02-25
pkp/pkp-lib#10977 Fix issues within merge user functionality

taslangraham:
For pkp/pkp-lib#10977

### DIFF
--- a/classes/log/Repository.php
+++ b/classes/log/Repository.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file classes/log/Repository.php
  *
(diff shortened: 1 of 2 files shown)


Hafsa-Naeem (changes requested):
`pluck(...)->toArray()` loads all matching email_log_id values into PHP memory before running the update. and might not be optimal for large datasets. how about using?
```
->whereNotIn('email_log_id', function ($sub) use ($newUserId, $oldUserId) {
    $sub->select('t1.email_log_id')
        ->from(DB::raw('email_log_users as t1'))
        ->join(DB::raw('email_log_users as t2'), 't1.email_log_id', '=', 't2.email_log_id')
        ->where('t1.user_id', $newUserId)
        ->where('t2.user_id', $oldUserId);
})
```

Hafsa-Naeem (approved):

Hafsa-Naeem:
@taslangraham Looks good to me.

Issue #10974, opened 2025-02-25
[Settings] | Navigation, Editing user navigation menu is throwing a 500 error


Tribunal33:
**Describe the bug**
Currently basic dataset is throwing error in the navigation edit screen for User Navigation Menu option

**To Reproduce**
1. Login with user with access to settings
2. Select Settings
3. Select Website -> Setup -> Navigation
4. Select trigangle to open edit options
5. Select Edit

**What application are you using?**
 OJS, OMP or OPS version 3.5

**Additional information**
Please add any screenshots, logs or other information we can use to investigate this bug report.

![Image](https://github.com/user-attachments/assets/abf76f60-3723-4b4c-8318-71a524f5228d)

PHP LOGS
[Mon Feb 24 21:29:41.137322 2025] [php:warn] [pid 19724:tid 1880] [client ::1:56596] PHP Warning:  fopen(C:\\xampp\\htdocs\\ojs-main\\ojs/cache/opcache/f1/d0/f1d023de2d739c7f72fdd63066213d9814d752d6): Failed to open stream: Permission denied in C:\\xampp\\htdocs\\ojs-main\\ojs\\lib\\pkp\\lib\\vendor\\laravel\\framework\\src\\Illuminate\\Filesystem\\LockableFile.php on line 69, referer: http://localhost/ojs-main/ojs/index.php/publicknowledge/en/dashboard/editorial?currentViewId=active
[...]

PR #10985, opened 2025-02-26
[PKP-LIB][main] #10974 Add page router specific check

defstat:


defstat:
Closed in favor of #10986

PR #10986, opened 2025-02-26
[PKP-LIB][main] #10974 Add page router specific check

defstat:


classes/services/PKPNavigationMenuService.php
```diff
@@ -19,6 +19,7 @@
 use APP\core\Application;
 use APP\template\TemplateManager;
 use Illuminate\Support\Facades\Cache;
+use PageRouter;
```
asmecher, line 22: This should be `APP\core\PageRouter`. But I've fixed that in `main`.

Issue #11001, opened 2025-02-27
[General Settings] | Uploading image with TinyMCE, when uploading image through TinyMCE it results in an error


jonasraoni:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

I'm unable to upload images through TinyMCE

<img width="406" alt="Image" src="https://github.com/user-attachments/assets/d4be1a67-a28c-44bb-a822-8e5819dfadf9" />

### Steps to Reproduce

1. Access Settings > Website > Appearance > Setup > Page footer
2. Try to upload an image at the TinyMCE
3. See error

### Expected Result

The upload should work.

### Actual Result

I see an error message

### Application Version

OJS 3.5

asmecher:
@jonasraoni, the JS lint/compile/minify tool doesn't like the use of `Promise`. Running:
```bash
bash lib/pkp/tools/buildjs.sh -n
```
...eventually gives...
```
...lib/pkp/js/controllers/SiteHandler.js
	jslint:lib/pkp/js/controllers/SiteHandler.js:178:32:'Promise' was used before it was defined.
```
Could you take a look?

asmecher:
@jonasraoni, this is a blocker for 3.5.0-0; I can't compile and minify the JS until it's fixed.

jonasraoni:
Hmm, if the compiler was released before the ES6 (2015), then perhaps we can deceive it?

```js
if (typeof Promise === "undefined") {
  function Promise() {}
}
```

asmecher:
@jonasraoni, I don't mind including that kind of workaround if it works. Could you give it a try?

Tribunal33:
Besides the compiler issue @asmecher mentioned this seems to work.  I was able to upload a file and save the form with TinyMCE.  Although I couldn't find an upload images area within Settings > Journal as described in the original steps, I did find one under Settings > Website > Appearance > Setup > Page footer where you could upload via TinyMCE.  Once this compiler issue is fixed this issue can be closed.

I have updated the Title of the issue as well as the steps to reproduce.

Tribunal33:
Appologies moving this to done on my project resulted in closing the issue by accident.

Tribunal33:
To see this change might have to enable_minified = Off

jonasraoni:
@asmecher, I can confirm it's not needed to have an old Python anymore to get the ball rolling 👌
- https://github.com/pkp/pkp-lib/pull/11432
[...]

PR #11002, opened 2025-02-27
pkp/pkp-lib#11001 Updated TinyMCE image upload handler

jonasraoni:


js/controllers/SiteHandler.js
```diff
@@ -174,26 +174,27 @@
...
+							processData: false,
+							contentType: false,
+							headers: {
+								'X-Csrf-Token': pkp.currentUser.csrfToken
+							},
+							success: function(r) {
+								resolve(r.url);
+							},
+							error: function(r) {
+								reject(r.responseJSON.error);
```
jonasraoni, line 195: I saw more occurrences of `errorMessage`, but I didn't inspect the other cases because the day is over here and they didn't seem to be related with the image issue.

jonasraoni: I'll review the other `errorMessage` occurrences.

Issue #10875, opened 2025-02-03
Locale key orcidProfile.orcidCollectAuthorId.name is missing


bozana:
At OJS installation I get this message in my PHP error log:
Got error 'PHP message: Missing locale key "orcidProfile.orcidCollectAuthorId.name" for the locale "en"; PHP message: Missing locale key "orcidProfile.orcidRequestAuthorAuthorization.name" for the locale "en"; PHP message: Missing locale key "mailable.userRoleAssignmentInvitationNotify.name" for the locale "en"', referer: http://ojs-dev.bb/index.php/index/en/install

The locale key seems to be missing and is used in registry/emailTemplates.xml

**Pull requests** (Some PRs made after the 3.5 branches were created)
- pkp-lib (for user role keys): https://github.com/pkp/pkp-lib/pull/11005
- OJS (for orcid keys): https://github.com/pkp/ojs/pull/4703

- OMP 3.5: https://github.com/pkp/omp/pull/1873
- OMP main: https://github.com/pkp/omp/pull/1874
- OPS 3.5: https://github.com/pkp/ops/pull/899
- OPS main: https://github.com/pkp/ops/pull/900

kaitlinnewson:
Hi @bozana, are you able to merge the two 3.5 PRs above? I missed the PRs to OMP/OPS for this before the 3.5 branches were made. Thanks!

bozana:
Merged, can the issue then be closed?

PR #11005, opened 2025-02-28
pkp/pkp-lib#10875 missing locale keys for emails

kaitlinnewson:


defstat (approved):
Thanks @kaitlinnewson. Just a suggestion, if you agree on that, I would change the wording to "User Invited to Role Notification"

Issue #10768, opened 2025-01-03
Redesigned Dashboard & Workflow rc2 - users with multiple roles


jardakotesovec:
# Dashboard
## Dashboards (editorial)
- Being Journal Manager, any submissions where I am assigned as author or reviewer have warnings and View button is hidden on this dashboard

<img width="1069" alt="Image" src="https://github.com/user-attachments/assets/62fd0e02-791f-457e-94ec-6dcb8aa07dc2" />
<img width="1060" alt="Image" src="https://github.com/user-attachments/assets/b397c666-df6b-4c6d-8d95-913f0eedeccb" />

- Being Journal Manager, and I am assigned as editor and author to submission. This will behave as usual, due to being explicitly assigned as editor. Therefore I will be able to access workflow modal.

- Being assigned to submission via Section Editor/Assistant, only submissions that are I assigned to me via editorial roles are included. Since these are always explicitly assigned, even if I am additionally assigned as author it does not change behaviour.

## My submissions
- only submisssions where I am assigned as AUTHOR are included, no special treatment if there are additional assignments to editorial roles.

## Review Assignments
- only review assignments where I am assigned as reviewer are included, no special treatment if there are additional assignments to editorial roles

ojs tests: https://github.com/pkp/ojs/pull/4710
pkp-lib: https://github.com/pkp/pkp-lib/pull/11028
ui-library: https://github.com/pkp/ui-library/pull/547

jardakotesovec:
@Tribunal33 Merged to main and stable-3_5_0. Ready for testing :-)

Tribunal33:
This passes QA in RC2.  I have created a user with all 3 views and can see the editorial notes correctly updated for each submission specific to the role.  

![Image](https://github.com/user-attachments/assets/f9c02cdd-1d28-4d18-82d9-671463a949ea)

PR #11028, opened 2025-03-04
I10768

jardakotesovec:


jardakotesovec:
hi @Vitaliy-1 , can you give me early feedback on the approach.

I basically just extended assignedTo to be also specify which roles is user assigned to.  And than used the same logic for `notAssignedTo` to be able to exclude submissions where journal manager is assigned as author/reviewer.

The `unAssignedTo` logic is now applied only for the first View, before I apply it to the other ones, can you give me early feedback on the approach?

Separation that I am trying to achieve is covered in - https://github.com/pkp/pkp-lib/issues/10768

Vitaliy-1:
Shouldn't `assignedWithRoles()` filter by enough in this case? We then just can pass required roles there.

jardakotesovec:
@Vitaliy-1 Thats problem with "unassigned" roles like Journal Manager. For these I think I have to exclude rather than include.

jardakotesovec:
@Vitaliy-1 Ready for second review round :-)
- based on discussion, we don't need notAssigned to anymore, so thats remvoed
- and one new adjustment to the `assignedRoleIds`, to not add global roles when current user is assigned as reviewer. This was important to be able to use `assignedRoleIds` on client to be able to tell when journal manager is explicitly assigned, to be able to show warning. But I also think its quite aligned with that we don't want to provide editorial functionality to the journal manager when assigned as reviewer

classes/submission/Collector.php
```diff
@@ -722,6 +752,49 @@ public function getQueryBuilder(): Builder
         return $q;
     }
 
+    protected function buildAssignedToQuery(Builder $q, $assignedTo, $assignedWithRoles): Builder
```
Vitaliy-1, line 755: Do you still need this method?

Vitaliy-1 (commented):
Please feel free to merge this, at a glance, looks ok.

PR #11035, opened 2025-03-04
pkp/pkp-lib#10977 Fix issues within merge user functionality

taslangraham:
For pkp/pkp-lib#10977

### DIFF
--- a/classes/log/Repository.php
+++ b/classes/log/Repository.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file classes/log/Repository.php
  *
@@ -18,6 +19,8 @@ namespace PKP\log;
 use APP\core\Application;
 use APP\facades\Repo;
 use APP\submission\Submission;
+use Illuminate\Database\Query\Builder;
+use Illuminate\Database\Query\JoinClause;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Mail;
 use PKP\core\Core;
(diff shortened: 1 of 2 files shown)


Hafsa-Naeem (approved):

classes/log/Repository.php
```diff
@@ -107,13 +109,18 @@ public function changeUser(int $oldUserId, int $newUserId)
...
+                        ->from(DB::raw('(SELECT email_log_id FROM email_log_users WHERE user_id = ?) as t1'))
+                        ->join(
+                            DB::raw('(SELECT email_log_id FROM email_log_users WHERE user_id = ?) as t2'),
+                            't1.email_log_id',
+                            '=',
+                            't2.email_log_id'
+                        )
+                        ->setBindings([$newUserId, $oldUserId]);
+                })
+                ->update(['user_id' => $newUserId])
```
jonasraoni, line 123:
```suggestion
        DB::table('email_log_users AS a')
            ->leftJoin(
                'email_log_users AS b',
                fn (JoinClause $j) => $j->on('b.email_log_id', '=', 'a.email_log_id')->where('b.user_id', '=', $newUserId)
            )
            ->where('a.user_id', $oldUserId)
            ->whereNull('b.user_id')
            ->update(['a.user_id' => $newUserId]);
```

jonasraoni: Also, as there isn't any kind of "report" over deleted records, I think the `return []` can be dropped. If we ever need a report, it would be better to break this into two methods, instead of guessing that maybe in the future, we'll have one more entry in the array.

jonasraoni: p.s.: I had to look at the table structure to confirm the purpose of this join, so I think it's useful to add a comment 😊

taslangraham:
@jonasraoni I've made the suggested changes. Thanks for the review.

Issue #11024, opened 2025-03-03
Redesigned Dashboard rc2 - Editorial Activity - API copyedited files count


jardakotesovec:
**Description:**
- We would like to indicate number of copyedited files that has been uploaded in editorial activity in dashboard (submission listing).

- This would require to add that information to the /_submissions endpoint and in performant way. Meaning avoid per submission query.

**Steps to test the changes:**
This ticket cover API addition. Actual UI testable change will be handled via parent ticket.

**What application are you using?**
  OJS, OMP or OPS version 3.5

taslangraham:
Ready for review. @ewhanson @Vitaliy-1 could you take a look?
**PRs**
pkp-lib(stable3.5): https://github.com/pkp/pkp-lib/pull/11046/files

submodule updates (stable3.5)
omp: https://github.com/pkp/omp/pull/1881
ops: https://github.com/pkp/ops/pull/905
ojs: https://github.com/pkp/ojs/pull/4723

If all looks good then I'll forward port to main

taslangraham:
[PR](https://github.com/pkp/pkp-lib/pull/11046/files) for pkp lib(stable 3.5) has been merged

taslangraham:
**PR for main**
- pkp-lib: https://github.com/pkp/pkp-lib/pull/11069

Submodule update(TESTS ONLY)
- https://github.com/pkp/ojs/pull/4732

taslangraham:
Work for PKP-LIB has been ported and merged to main

PR #11046, opened 2025-03-06
pkp/pkp-lib#11024 Update API to include copyedit stage files count

taslangraham:
For pkp/pkp-lib#11024

### DIFF
--- a/classes/submission/maps/Schema.php
+++ b/classes/submission/maps/Schema.php
@@ -34,6 +34,7 @@ use PKP\submission\reviewer\suggestion\ReviewerSuggestion;
 use PKP\API\v1\reviewers\suggestions\resources\ReviewerSuggestionResource;
 use PKP\submission\reviewRound\ReviewRound;
 use PKP\submission\reviewRound\ReviewRoundDAO;
+use PKP\submissionFile\SubmissionFile;
 use PKP\user\User;
 use PKP\userGroup\relationships\UserGroupStage;
 use PKP\userGroup\relationships\UserUserGroup;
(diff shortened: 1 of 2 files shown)


classes/submission/maps/Schema.php
```diff
@@ -817,6 +837,14 @@ protected function getPropertyStages(Enumerable $stageAssignments, Enumerable $r
                 'editorAssigned' => false,
                 'currentUserAssignedRoles' => [],
             ];
+
+            if ($stageId === WORKFLOW_STAGE_ID_EDITING) {
+                $stages[$stageId]['uploadedFilesCount'] = $stageFiles->filter(fn(SubmissionFile $file) => $file->getData('fileStage') == SubmissionFile::SUBMISSION_FILE_COPYEDIT)->count();
+            } else {
+                // A `null` value is used to indicate that no count data is available.
+                // This is also done to ensure that all stage objects has the same properties
+                $stages[$stageId]['uploadedFilesCount'] = null;
```
taslangraham, line 846: An alternative approach I considered was to include the count as a top level property on the submission object.
Something like `copyeditStageFilesCount`.  Thus avoiding a property with mix value on the stage objects.

classes/submission/maps/Schema.php
```diff
@@ -783,6 +802,7 @@ protected function getPropertyParticipants(Submission $submission): array
      *  `currentUserAssignedRoles` array the roles of the current user in the submission per stage, user may be unassigned but have global manager role
      *  `currentUserCanRecommendOnly` whether the current user is an editor with the recommend only flag
      *  `currentUserRecommendation` object includes the recommendation decision of the current user
+     * `uploadedFilesCount` int || null the count of files upload to the stage. A null value indicates that the count was not included
```
ewhanson, line 805: Spacing alignment off here.

classes/submission/maps/Schema.php
ewhanson: As discussed today, I think having this be nullable here rather than at the top level is the way to go for this. 👍

ewhanson (changes requested):
Thanks @taslangraham, looking good! Just one minor spacing issue I noticed.

classes/submission/maps/Schema.php
taslangraham: updated

Vitaliy-1 (approved):
Thanks, think it's good to go as is! While we don't have this entities refactored to the Eloquent Models with eager loading, this is the best approach.

Issue #11047, opened 2025-03-07
Editorial Reminder bug : When section editor is an author


qlaboudigue:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

User assigned only as Author for a submission while being Section editor receives the Editorial Reminder e-mail.

![Image](https://github.com/user-attachments/assets/cacfe719-8c2f-48dd-ab04-9041c024a3cc)

### Steps to Reproduce

1. Log in as an Author being also Section Editor
2. Get userId
3. Create a submission
4. Locally, update submission date_last_activity with UPDATE submissions SET date_last_activity = DATE_SUB(NOW(), INTERVAL 31 DAY
5. Locally, trigger EditorialReminder :
`$job = new EditorialReminder($userId, $contextId);
 $job->handle();`

### Expected Result

Author should not receive EditorialReminder email

### Actual Result

Author receives EditorialReminder email

### Application Version

OJS version 3.4.0

asmecher:
Thanks, @qlaboudigue! @Vitaliy-1, could you review [the linked PR](https://github.com/pkp/pkp-lib/pull/11050) and see if this might also fix https://github.com/pkp/pkp-lib/issues/10573? (Or delegate that?)

asmecher:
@Vitaliy-1, just a reminder about this one needing review.

pablovp86:
We had a similar issue on one of our journals. But with an user with the Reviewer role.

The user was not assigned as the section editor for a submission, but only as a reviewer. As a result, the user received information identifying the author responsible for the manuscript,  compromising the anonymity of the process.

PR #11050, opened 2025-03-07
Editorial reminder should not trigger for submission assignee with specific roles combination

### LINKED ISSUE #10573: Editorial Reminder bug: When the section editor becomes a reviewer
**Describe the bug**
When a section editor is assigned as a reviewer for an submission, she is notified in the outstanding e-mail as if she were an editor.

 OJS version 3.4.0.5

**Additional information**
I am sharing screenshots below.

Example 1: Status "New reviews have been submitted."
image

image

Example 2: Status "Review overdue."
image

image

qlaboudigue:
Issue #11047 
Might fix #10573

### DIFF
--- a/jobs/email/EditorialReminder.php
+++ b/jobs/email/EditorialReminder.php
@@ -34,6 +34,7 @@ use PKP\submission\reviewRound\ReviewRound;
 use PKP\submission\reviewRound\ReviewRoundDAO;
 use PKP\user\User;
 use PKP\workflow\WorkflowStageDAO;
+use PKP\security\Role;
 
 class EditorialReminder extends BaseJob
 {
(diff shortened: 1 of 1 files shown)


twakeford:
Is there any update on this?

asmecher:
@qlaboudigue, thanks for the PR, and I'm sorry for the long delay in responding! I believe this won't have the desired effect. It looks like the change makes sure there is at least one `ROLE_ID_MANAGER` or `ROLE_ID_SUB_EDITOR` assignment to the submission in order for a user to be notified, which [should already be the case](https://github.com/pkp/pkp-lib/blob/0facd943c8262a6a5c2af2e0aaaf9e6be613b04e/classes/task/EditorialReminders.php#L46..L50) based on the code that invokes the reminder.

Instead, I think the desired effect is that Editors who are assigned to the submission *but who are also assigned as reviewers* should not receive the invitation. That is, the code should check whether the user has any review assignments, and if so, they should be skipped.

twakeford:
Small addition to the above. I believe it should check to see whether they have any reviewer _and/or author_ assignments to the submission, and if so, they should be skipped

twakeford:
@qlaboudigue have you been able to progress this? We have noticed this as a problem so also planned to work on it. If you already have a solution on the horizon then we will wait. If not/you'd prefer then we can schedule this into our dev timeline.

Issue #11049, opened 2025-03-07
Password resets fail with PHP 8.1


jnugent:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The method getContextList in PKPApplication is not defined as static, so this generates a fatal error in PHP 8.1 when it is called statically in PKPRrequest:

```
Got error 'PHP message: PHP Fatal error:  Uncaught Error: Non-static method Application::getContextList() cannot be called statically in lib/pkp/classes/core/PKPRequest.inc.php:293
```

### Steps to Reproduce

1. Deploy 3.3.0.20 under PHP 8.1
2. Create a user and attempt a password reset.
3. Visit the link sent to you in your email
4. log fatal error

### Expected Result

it should not generate a fatal error

### Actual Result

blank screen, fatal error in logs

### Application Version

OJS 3.3.0.20

jnugent:
PR against stable-3_3_0 https://github.com/pkp/pkp-lib/pull/11051

asmecher:
@jnugent, thanks, I've merged the `stable-3_3_0` PR. Can you confirm whether this needs forward-porting? I suspect not, but worth checking.

jonasraoni:
This method was removed in 3.4, so I'll close.

PR #11051, opened 2025-03-07
pkp/pkp-lib#11049 fix static call to non-static method

jnugent:
replace static call with non-static one.


jonasraoni (approved):
I think this is more "stable" than adding the `static` keyword :)

asmecher (approved):

Issue #11031, opened 2025-03-04
[Submissions] | Creating submissions for multiple roles have missing "Submit As" option labels


blesildaramirez:
### Description

When creating submissions for users with multiple roles (e.g: journal editor + author), the "Submit As" option labels are missing in the Submission Form. 

### Steps to Reproduce

1. Login an account with multiple roles (e.g editor + author)
2. Click `Start a New Submission` from the side menu
3. Go to "Submit As" section in the form

### Expected Result

It should display the current roles of the logged in user as the form labels.

### Actual Result
Labels for current user  roles are not displayed:
![Image](https://github.com/user-attachments/assets/98d7e059-0175-4773-a1fd-9754f3852afa)

### Application Version

OJS 3.5

-------
**Pull requests**
main:
- [x] pkp-lib: https://github.com/pkp/pkp-lib/pull/11034
- [x] ojs (test only): https://github.com/pkp/ojs/pull/4712

stable-3_5_0:
- [x] pkp-lib: https://github.com/pkp/pkp-lib/pull/11033
- [x] ojs (test only): https://github.com/pkp/ojs/pull/4711

-------
**New pull requests**
main:
- [x] pkp-lib: https://github.com/pkp/pkp-lib/pull/11053
- [x] ojs (test only): https://github.com/pkp/ojs/pull/4725
- [x] omp (test only): https://github.com/pkp/omp/pull/1884
- [x] ops (test only): https://github.com/pkp/ops/pull/907

stable-3_5_0:
- [x] pkp-lib: https://github.com/pkp/pkp-lib/pull/11052
- [x] ojs (test only): https://github.com/pkp/ojs/pull/4724
- [x] omp (test only): https://github.com/pkp/omp/pull/1883
- [x] ops (test only): https://github.com/pkp/ops/pull/906

Vitaliy-1:
Actually, I left comments. Please, go ahead and merge that once changes are applied. Thanks!

blesildaramirez:
Closing this, testing done for OJS:
- [x] Creating a new submission using an account with multiple roles
- [x] Creating a new submission from a newly registered author

blesildaramirez:
Closing this, cypress tests completed for OJS/OMP/OPS.

PR #11034, opened 2025-03-04
pkp/pkp-lib#11031 [MAIN] [Submissions] | Creating submissions for multiple roles have missing "Submit As" option labels

blesildaramirez:

PR #11052, opened 2025-03-07
pkp/pkp-lib#11031 [stable-3_5_0] Return default user group  if  userGroup is empty

blesildaramirez:


asmecher (commented):
@blesildaramirez, I think the original problem was likely in the format of the returned collection, but this change also reduces the choices a user will have in case they don't have any appropriate user groups. Is the shift from `get()` to `first()` intentional?

blesildaramirez:
Hi @asmecher . Yes that's intentional. It's originally using `first()` from this function `getFirstSubmitAsAuthorUserGroup`: https://github.com/pkp/pkp-lib/pull/11033/files#diff-c8cb812ae82cc024166167807c2d42818fe8c33bc61dc597835f7dbf2d7c91b2L454
We removed that function now because we moved the query execution at `getSubmitUserGroups()` function. Just that while moving, [I mistakenly used `get()`](https://github.com/pkp/pkp-lib/pull/11034#discussion_r1982665219) instead of the original `first()`.

touhidurabir (approved):

Issue #11063, opened 2025-03-11
[User Export] | User Export fails when no affiliation is entered


kaitlinnewson:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

In the Users XML plugin, user export fails if no affiliation is present.

### Steps to Reproduce

1. Login with User/Role admin
2. Navigate to Tools > Users XML plugin > Export Users
3. Select only the admin users and click "Export Users"

### Expected Result

The exported XML should be downloaded with no errors.

### Actual Result

The export fails with a validation error.

### Application Version

OJS/OMP/OPS 3.5

### Logs

```
[Tue Mar 11 09:07:53 2025] Exception: XML validation failed with:
Array
(
    [0] => LibXMLError Object
        (
            [level] => 2
            [code] => 1871
            [column] => 0
            [message] => Element '{http://pkp.sfu.ca}affiliation': Missing child element(s). Expected is ( {http://pkp.sfu.ca}name ).

            [file] => 
            [line] => 0
        )

)
 in /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/classes/xslt/XMLTypeDescription.php:142
Stack trace:
#0 /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/classes/filter/TypeDescription.php(143): PKP\xslt\XMLTypeDescription->checkType(Object(DOMDocument))
#1 /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/plugins/importexport/native/filter/NativeExportFilter.php(73): PKP\filter\TypeDescription->isCompatible(Object(DOMDocument))
#2 /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/classes/filter/Filter.php(465): PKP\plugins\importexport\native\filter\NativeExportFilter->supports(Array, Object(DOMDocument))
#3 /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/plugins/importexport/users/PKPUserImportExportPlugin.php(250): PKP\filter\Filter->execute(Array)
#4 /Users/kaitlinnewson/Code/ojs-3_5_0/lib/pkp/plugins/importexport/users/PKPUserImportExportPlugin.php(166): PKP\plugins\importexport\users\PKPUserImportExportPlugin->exportUsers(Array, Object(APP\journal\Journal), Object(PKP\user\User), Object(PKP\plugins\importexport\users\filter\PKPUserUserXmlFilter))
[...]

PR #11064, opened 2025-03-11
pkp/pkp-lib#11063 handle missing affiliation in user export

kaitlinnewson:


jonasraoni (commented):
I didn't have time to review, but I was just going to suggest re-using the `$user->getAffiliation(null)`.

I also have some general comments regarding the XML generation, but I've left them on a discussion :)

Issue #10835, opened 2025-01-22
Redesigned Dashboard - Views for OJS&OMP&OPS


jardakotesovec:
This ticket represents work to adjust the Views for OMP and OPS context. 

Best to orient where the dashboard views are being used is to search for `getDashboardViews` in code base.

Explanation for the views is covered in [figma](https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?node-id=7755-12313&t=WyFYNXT6Fj6vK1Qw-4).

## OJS
- Update labels according to figma mockup - https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?node-id=9303-11153&t=UByqbCDCSnXlfqpd-4
- View 'Needs Editor', which is already implemented is missed on the mock up, but we want to keep it

## OMP
- based on @Devika008 research - she concluded that internal & external reviews would be under same Views - not separated
- Therefore views are the same, we just need to ensure it works correctly both for internal and external review

![Image](https://github.com/user-attachments/assets/c6ccb9ff-a632-43f3-9eee-5dab85742e8c)

## OPS
- Views for OPS are reduced, as some of them would never have item
-
![Image](https://github.com/user-attachments/assets/41c8ddef-1bfe-4412-bb07-7a696e111d00)

taslangraham:
Ready for review @Vitaliy-1 @jardakotesovec 
**PRs**
pkp-lib (stable 3.5): https://github.com/pkp/pkp-lib/pull/11067
OMP (stable 3.5): https://github.com/pkp/omp/pull/1891
OPS (stable 3.5): https://github.com/pkp/ops/pull/910

**Submodule update only**
OJS (stable 3.5):  https://github.com/pkp/ojs/pull/4731

Once everything looks good, I'll port to `main`

Vitaliy-1:
Thanks! @jardakotesovec, do you have time to look also?

jardakotesovec:
@Vitaliy-1 @taslangraham Looks great to me! I did merge it to stable 3.5 and cherry picked to main.

PR #11067, opened 2025-03-11
pkp/pkp-lib#10835 Update dashboard views (stable 3.5)

taslangraham:
For pkp/pkp-lib#10835

### DIFF
--- a/locale/en/submission.po
+++ b/locale/en/submission.po
@@ -2416,11 +2416,14 @@ msgstr "Active submissions"
 msgid "submission.dashboard.view.needsEditor"
 msgstr "Needs editor"
 
-msgid "submission.dashboard.view.initialReview"
-msgstr "All in desk/initial review"
+msgid "submission.dashboard.view.submissionStageAll"
+msgstr "All in submission stage"
 
 msgid "submission.dashboard.view.reviewExternal"
 msgstr "All in peer review"
+
+msgid "submission.dashboard.view.reviewAll"
+msgstr "All in review stage"
 
 msgid "submission.dashboard.view.needsReviews"
 msgstr "Needs reviews"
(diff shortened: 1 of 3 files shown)


locale/en/submission.po
```diff
@@ -2416,12 +2416,15 @@ msgstr "Active submissions"
 msgid "submission.dashboard.view.needsEditor"
 msgstr "Needs editor"
 
-msgid "submission.dashboard.view.initialReview"
-msgstr "All in desk/initial review"
+msgid "submission.dashboard.view.submissionStageAll"
+msgstr "All in submission stage"
```
Vitaliy-1, line 2420: Not sure if we should change the wording here and other places

Vitaliy-1 (commented):
This PR changes the localization strings associated with submissions list naming. @Devika008, can you take a look?

locale/en/submission.po
taslangraham: The [ticket](https://github.com/pkp/pkp-lib/issues/10835#issue-2804871826) does mention that the labels should be updated to match this [design](https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?node-id=9303-11153&t=UByqbCDCSnXlfqpd-4), but @jardakotesovec or @Devika008 can confirm if this is still the case.

Vitaliy-1 (approved):

locale/en/submission.po
jardakotesovec: Yeah, this is correct change. 'Desk review' was supposed to replace 'submission' stage in early designs..  But we back out from that.

jardakotesovec (approved):

Issue #10929, opened 2025-02-13
[Create Submission] | Upload Files - Production editor adding a file and updating article text receives an access denied 401 error


Tribunal33:
**Describe the bug**
Please tell us what happens, what you expected to happen, and why you think it is a bug in the software.

Preconditions: Have a user with the role Production Editor and nothing else

**To Reproduce**
1. Login as the Production Editor User
2. Select Create New Submission to start the workflow
3. On Details stage fill in the form as usual (title, keywords, Abstract etc..)
4. On upload Files stage Upload a file
5. Have dev console open
6. Select Article Text for the file type
a. Alternatively you can select Other and see a different error but same auth problem (see screenshot 2)

Expected: Article text type is selected without error.  Or other file type is selected without error / auth popup
Act;ual:  401 Error in console (screen shot 1) and does not select the file type.  Other brings up a auth server window to sign in

**Additional information**
Please add any screenshots, logs or other information we can use to investigate this bug report.

![Image](https://github.com/user-attachments/assets/2bdeaba1-fd7a-4614-a066-2b5d80c24136)

Other option unexpected result.

![Image](https://github.com/user-attachments/assets/de3fb1a3-f87e-425f-9bbc-5eaabf0c3a9f)

taslangraham:
This issue arises because Production Editor does not have access to submission stage, but when selecting a file type, stage related policy (PolicyA & PolicyB) checks are made thus resulting in a authorization failure.
A short term solution suggested by @Vitaliy-1 is to have Production Editor automatically submit as author, instead of production editor, when creating new submission, thus, granting access to the submission stage.

taslangraham:
**PRs**
pkp-lib(stable 3.5): 
- https://github.com/pkp/pkp-lib/pull/11075/files
- ojs(tests only, stable 3.5): https://github.com/pkp/ojs/pull/4738

@Vitaliy-1 Could you give this a review, please?

Vitaliy-1:
Thans, @taslangraham! I left a comment regarding implementation.

asmecher:
@taslangraham, it looks like this one is nearly ready; is it possible to fix it as a high priority for 3.5.0?

Vitaliy-1:
Thanks @taslangraham, merged! Leaving the issue open as we need to port it to main

taslangraham:
Thanks @Vitaliy-1! I'll port to main shortly

Tribunal33:
[...]

PR #11075, opened 2025-03-12
pkp/pkp-lib#10929 Temporary fix for Production Editor's file upload issues

taslangraham:
For pkp/pkp-lib#10929


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -664,8 +664,34 @@ public function add(Request $illuminateRequest): JsonResponse
                     : $submitAsUserGroup->permitMetadataEdit
             );
 
+        $isSubmittingAsProductionEditor = $submitAsUserGroup->getLocalizedData('name') === __('default.groups.name.productionEditor');
```
taslangraham, line 667: Not sure if this is the best way to check for a specific role. Open to suggestions

Vitaliy-1: I think the best way would be to check the access to the submission stage through the user group stage table

Vitaliy-1: Forgot to mention that OPS might require different treatment - in 3.5 stage ID has only value `5` there, submission stage was removed.

taslangraham:
@Vitaliy-1 Thanks for the review. I've made the updates based on your suggestions.

Also, given that OPS does not have Production Editor group, and production stage (stage ID `5`), I've added app level implementation for the `SubmissionController::add` method in OPS. PR can be found [here](https://github.com/pkp/ops/pull/930).

Vitaliy-1 (changes requested):
Sorry for the late reply!

The part of code that checks user roles and automatically assigned author role if necessary is here: https://github.com/pkp/pkp-lib/blob/main/pages/submission/PKPSubmissionHandler.php#L615

Thus, it makes senes to move the check there. 
Moreover, I don't know what was the initial idea behind that part. In my opinion, it makes sense to check if user has access to the submission stage in general, and if not, assign an author role. That applies, for production editors, which are managers, copyeditors, layout editors, reviewers... and custom user groups. 

Also, we might want to exclude user groups without access to submission stage from from the list here: https://github.com/pkp/pkp-lib/blob/main/classes/components/forms/submission/StartSubmission.php#L159

taslangraham:
Thanks @Vitaliy-1, I've made updates based on your suggestions:
1. Move checks and group assignment to https://github.com/pkp/pkp-lib/blob/main/pages/submission/PKPSubmissionHandler.php#L615
    * Assign user to Author role once user doesn't have access to submission stage
3. Exclude user groups without access to submission stage from list of "Submit as" options
4. Keep original behaviour for https://github.com/pkp/pkp-lib/blob/main/pages/submission/PKPSubmissionHandler.php#L615 in OPS ([PR](https://github.com/pkp/ops/pull/930))


Take a look and let me know if it makes sense

pages/submission/PKPSubmissionHandler.php
```diff
@@ -614,22 +614,29 @@ protected function getReviewerSuggestionsListPanel(
...
-            ->withUserIds([$user->getId()])
-            ->withRoleIds([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_AUTHOR])
-            ->get();
+        $request = Application::get()->getRequest();
+        $isAdmin = $request->getUser()->hasRole([Role::ROLE_ID_SITE_ADMIN], \PKP\core\PKPApplication::SITE_CONTEXT_ID);
+        $query = UserGroup::query()->withContextIds([$context->getId()])->withUserIds([$user->getId()]);
 
-        // Users without a submitting role can submit as an
-        // author role that allows self registration
+        $userGroups = $isAdmin
```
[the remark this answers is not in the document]
taslangraham: Yes, what this does is that once a user is not an admin it queries for user groups that has access to the submission stage. If the user has no such groups, then the author role gets assigned.

taslangraham:
Thanks for the review @Vitaliy-1! I've made the suggested updated and resolved all merge conflicts.

pages/submission/PKPSubmissionHandler.php
```diff
@@ -614,25 +614,33 @@ protected function getReviewerSuggestionsListPanel(
...
-            })
-            ->withRoleIds([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_AUTHOR])
-            ->get();
+            });
 
-        // Users without a submitting role can submit as an
-        // author role that allows self registration
+        $userGroups = $isAdmin
+            ? $query->withRoleIds([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN])->get()
+            : $query->withStageIds([WORKFLOW_STAGE_ID_SUBMISSION])->get(); // For non-admin users, query for the groups tht give them access to the submission stage
```
Vitaliy-1, line 627: I think, we also need to check start and end dates

Vitaliy-1 (changes requested):
Thanks, just one small comment

Issue #10363, opened 2024-09-04
Minimum Reviews Required for a Submission


Devika008:
# Specs Update as of Thursday 27th February, 2025

## Workflows affected by this change
- [x] In **Dashboard > Editorial Activity**, the **"Add Reviewer"** button should be displayed **only when no reviewers are assigned** to the submission. However, if at least **one reviewer is assigned**, but the **Minimum confirmed reviews required have not been met**, the button will be **removed**, and the submission will still appear under the **"Needs Reviewers"** view.
- [x] In **Dashboard > Editorial Activity**, a **new status** should be added when the **Minimum confirmed reviews required have been confirmed by the editor**, indicating:  **"Minimum required number of reviews have been confirmed. A decision is needed."**  
- [x] - [ ] In **Dashboard > View > Submission Workflow page**, a **new status** should be displayed ** when the minimum number of reviews are confirmed by the editor", indicating:  **"Minimum number of reviews have been confirmed. A decision is needed."**  
- [x] In **Settings > Workflow > Review Setup**, a **new setting** should be added, allowing the **Admin, Journal Manager, and Journal Editor** to define the **Minimum number of confirmed reviews required for a submission**.
- [x] In **Dashboard > View > Submission Workflow page**, a **new status** should be displayed **at all times**, indicating:  **"Minimum number of confirmed reviews required: {number}"**  
- [x] If the **minimum number of confirmed reviews has not been met** and the editor attempts to take an **editorial decision**—**"Create a New Round," "Request Revisions," or "Accept Submission"**—an **alert** should be displayed stating:  **"The minimum number of confirmed reviews has not been met. Do you still want to proceed with this editorial decision?"**  
If the editor selects **"Yes, Continue,"** the **user journey** for these editorial decisions will follow the **existing process from OJS 3.5** without changes.

### Minimum Reviews vs. Minimum Reviewers Required

Question: Which option makes the most sense?

- Option 1: Minimum Number of Reviewers Only → 27 respondents
- Option 2: Minimum Required Reviews Only → 59 respondents ✅ (Most preferred)
- Option 3: Hybrid Approach (Combination of the two) → 47 respondents
- No response → 14 respondents

📌 Key Insight: The majority of respondents (40%) preferred Option 2, as it provides flexibility in setting required reviews per submission.

#### Key Takeaways:

✅ Editors prefer tracking completed reviews rather than just the number of assigned reviewers.  
[...]

PR #11080, opened 2025-03-12
pkp/pkp-lib#10363 Context setting to control the minimum number of reviews per submission

Vitaliy-1:


blesildaramirez (approved):
Reviewed and tested the "Needs reviews" view. 
Just wanted to note that we probably need to update this swagger docs as well for OJS: https://github.com/pkp/ojs/blob/main/docs/dev/swagger-source.json#L7569

taslangraham (approved):

Issue #11025, opened 2025-03-03
Redesigned Dashboard rc2 -  API published in issue


jardakotesovec:
**Description:**
- We would like to indicate **issue** that would be submission published in. It would be indicated in editorial activity on dashboard (submission listing).

- This would require to add that information to the /_submissions endpoint and its important to do it in performant way. Meaning avoid per submission query.

**Steps to test the changes:**
This ticket cover API addition. Actual UI testable change will be handled via parent ticket.

**What application are you using?**
  OJS, OMP or OPS version 3.5

taslangraham:
We can determine a submission's issue through its associated publication. However, when fetching publications for a submission, we have two considerations: the latest publication and the current publication. 

- The **current publication** is the version the submission was last published with.
- The **latest publication** is the most recent version created, regardless of whether it has been published

Asking for clarity here - for this feature, are we interested in the issue attached to the latest published publication or the issue attached to the latest created publication?
Tagging you as well @Vitaliy-1 incase you have any input

jardakotesovec:
Summary from the call:

I will list options, with the first one to be preferred, but if that would cause performance problems (many queries), than we can fallback to other ones

1) Attach issue directly to the publication object, instead of submission. That would move responsibility to client to pick the relevant publication, which I think would be good. And than the issue information directly on publication object could be useful for other use cases as well.
2) If we would need to stick attaching issue to submission object - than for current purposes, it should be called `issueToBePublished` and it would latest publication which has status "Scheduled"

taslangraham:
Ready for review. @Vitaliy-1 @jardakotesovec 

pkp-lib (stable-3.5): https://github.com/pkp/pkp-lib/pull/11102
ojs(stable-3.5, tests only): https://github.com/pkp/ojs/pull/4746

I initially attempted to take the first approached mentioned by Jarda in his [comment](https://github.com/pkp/pkp-lib/issues/11025#issuecomment-2729888258), but was unsuccessful in my attempts. Given the time constraint I decided to go with option 2

asmecher:
All merged for 3_5_0rc2 -- thanks, @taslangraham and @ewhanson! This still needs forward-porting to `main`, so leaving open.

taslangraham:
PRs for main

ojs: https://github.com/pkp/ojs/pull/4779
omp: https://github.com/pkp/omp/pull/1914
ops: https://github.com/pkp/ops/pull/927
[...]

PR #11102, opened 2025-03-14
pkp/pkp-lib#11025 Include issue info on submission API data

taslangraham:
For pkp/pkp-lib#11025 
tests - https://github.com/pkp/ojs/pull/4746

### DIFF
--- a/classes/submission/maps/Schema.php
+++ b/classes/submission/maps/Schema.php
@@ -22,7 +22,6 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Enumerable;
 use Illuminate\Support\LazyCollection;
 use PKP\API\v1\reviewers\suggestions\resources\ReviewerSuggestionResource;
-use PKP\context\Context;
 use PKP\db\DAORegistry;
 use PKP\decision\DecisionType;
 use PKP\plugins\Hook;
@@ -152,6 +151,7 @@ class Schema extends \PKP\core\maps\Schema
         $this->decisions = $decisions ?? Repo::decision()->getCollector()->filterBySubmissionIds([$item->getId()])->getMany()->remember();
         $this->reviewerSuggestions = $reviewerSuggestions ?? ReviewerSuggestion::withSubmissionIds($item->getId())->get();
         $this->submissionStageFiles = $stageFiles ?? $this->getStageFilesBySubmissions(collect([$item]), [SubmissionFile::SUBMISSION_FILE_COPYEDIT]);
+        $this->addAppSpecificData(collect([$item]));
 
         return $this->mapByProperties($this->getProps(), $item, $anonymizeReviews);
     }
@@ -184,6 +184,8 @@ class Schema extends \PKP\core\maps\Schema
         $this->stageAssignments = $stageAssignments ?? $this->getStageAssignmentsBySubmissions(collect([$item]));
         $this->reviewerSuggestions = $reviewerSuggestions ?? ReviewerSuggestion::withSubmissionIds($item->getId())->get();
         $this->submissionStageFiles = $stageFiles ?? $this->getStageFilesBySubmissions(collect([$item]), [SubmissionFile::SUBMISSION_FILE_COPYEDIT]);
+        $this->addAppSpecificData(collect([$item]));
+
         return $this->mapByProperties($this->getSummaryProps(), $item, $anonymizeReviews);
     }
 
@@ -214,6 +216,7 @@ class Schema extends \PKP\core\maps\Schema
         $this->stageAssignments = $this->getStageAssignmentsBySubmissions($collection);
         $this->decisions = Repo::decision()->getCollector()->filterBySubmissionIds($submissionIds)->getMany()->remember();
         $this->submissionStageFiles = $this->getStageFilesBySubmissions($collection, [SubmissionFile::SUBMISSION_FILE_COPYEDIT]);
+        $this->addAppSpecificData($collection);
 
         $associatedReviewAssignments = $this->reviewAssignments->groupBy(fn (ReviewAssignment $reviewAssignment, int $key) =>
             $reviewAssignment->getData('submissionId'));
@@ -268,7 +271,7 @@ class Schema extends \PKP\core\maps\Schema
         $this->reviewAssignments = Repo::reviewAssignment()->getCollector()->filterBySubmissionIds($collection->keys()->toArray())->getMany()->remember();
         $this->stageAssignments = $this->getStageAssignmentsBySubmissions($collection);
         $this->submissionStageFiles = $this->getStageFilesBySubmissions($collection, [SubmissionFile::SUBMISSION_FILE_COPYEDIT]);
-
+        $this->addAppSpecificData($collection);
         $associatedReviewAssignments = $this->reviewAssignments->groupBy(
             fn (ReviewAssignment $reviewAssignment, int $key) =>
             $reviewAssignment->getData('submissionId')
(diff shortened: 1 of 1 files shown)


classes/submission/maps/Schema.php
```diff
@@ -541,6 +567,9 @@ protected function mapByProperties(array $props, Submission $submission, bool|Co
                 case 'reviewerSuggestions':
                     $output[$prop] = $this->getPropertyReviewerSuggestions($this->reviewerSuggestions);
                     break;
+                case 'issueToBePublished':
```
Vitaliy-1, line 570: We already have all publications associated with a submission, can we derive this information from there?
Above we have a call `submission->getData('publications')`,
publication has a property `issueId`, `status` and `version`. That should give all needed data to us, shouldn't it?

classes/publication/Collector.php
```diff
@@ -115,6 +140,10 @@ public function getQueryBuilder(): Builder
         $qb = DB::table('publications as p')
             ->select(['p.*']);
 
+        if ($this->isScheduled) {
+            $qb->where('p.status', PKPSubmission::STATUS_SCHEDULED);
```
Vitaliy-1, line 144: Do we need this filter? For this particular task we can look at the `status` property of the `publication`

And to double check, do we need only publication with the scheduled status but not published?

Vitaliy-1 (changes requested):
I don't think that we need changes to the publication collector here. Also, we can avoid making making additional DB calls but retrieving data from the publications attached to the each submission.


As for logic, I would show all issues, in which publications associated with the submission are scheduled or published. We have clear one to many relationship between submission and publication, and this looks rationally for me as it shows the full picture to the user.  But don't mind to show the data only regarding the latest publication

taslangraham:
> As for logic, I would show all issues, in which publications associated with the submission are scheduled or published. We have clear one to many relationship between submission and publication, and this looks rationally for me as it shows the full picture to the user. But don't mind to show the data only regarding the latest publication.

Okay, understood. So I guess instead of an `issueToBePublished` property on the submission schema, we'd just go with `issues` which has list of all relevant ones?

classes/publication/Collector.php
taslangraham: >And to double check, do we need only publication with the scheduled status but not published?

That was the request [here](https://github.com/pkp/pkp-lib/issues/11025#issuecomment-2729888258), on Option#2 which is the implementation I went with.

Vitaliy-1:
The part with deriving issues from submissions collection to retrieve label is ok.

We can add a property `issueToBePublished` but derive it from publications objects associated with the submission

Vitaliy-1:
Something like:

```php
$publications = $submission->getData('publications');

case 'issueToBePublished':
  $output[$prop]  = $this->getPropertyIssueToBePublished($publications);
   break;

protected function getPropertyIssueToBePublished(array $publications) {
   $issueId = [];

   foreach (publications as $publication) {
     if (!in_array(publication->getData('status'), [PKPSubmission::STATUS_SCHEDULED, PKPSubmission::STATUS_PUBLISHED]) {
       continue
     }

    $issueId[] = $publication->getData('issueId)
 
   }
}
```

Vitaliy-1 (approved):
Thanks, looks good!

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
...
+    protected function getIssueBySubmissions(Enumerable $submissions): Enumerable
+    {
+        $submissionIds = [];
+        $submissions->each(function (Submission $submission) use (&$submissionIds) {
+            $submissionIds[] = $submission->getId();
+        });
+
+        $publications = Repo::publication()->getCollector()->filterBySubmissionIds($submissionIds)->getMany();
+
+        $issueIds = $publications->map(fn(Publication $publication) => $publication->getData('issueId'))
```
jonasraoni: I think you can remove most of the code above, the publications are available at the `Submission`  class, through `getData('publications')`

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
...
+            ->toArray();
+
+        $issues = Repo::issue()->getCollector()
+            ->filterByContextIds([$this->context->getId()])
+            ->filterByIssueIds($issueIds)
+            ->getMany();
+
+        return $submissions->mapWithKeys(function ($submission) use (&$issues, $publications) {
+            $submissionIssueIds = $publications
+                ->filter(fn(Publication $publication) => $publication->getData('submissionId') === $submission->getId())
```
jonasraoni: As there will be many loops here, it would make sense to group the items by submission ID before.
But given that this `$publications` variable might disappear (ref: comment where I said it's possible to drop most of the code), maybe all this piece will be refactored.

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
...
+                ->map(fn(Publication $publication) => $publication->getData('issueId'))
+                ->toArray();
+
+            return [$submission->getId() => $issues->filter(fn($issue) => in_array($issue->getId(), $submissionIssueIds))];
```
jonasraoni: I think this can be optimized too, but I'll review better in a next iteration.

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
             ->filterByFileStages($stageIds)
             ->getMany();
     }
+
+    /***
+     * Get detail about submission issues. Results are keyed by submission ID.
```
jonasraoni: I think it's nice to add more typing information on the `@return`

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
...
+                ->filter(fn(Publication $publication) => $publication->getData('submissionId') === $submission->getId())
+                ->map(fn(Publication $publication) => $publication->getData('issueId'))
+                ->toArray();
+
+            return [$submission->getId() => $issues->filter(fn($issue) => in_array($issue->getId(), $submissionIssueIds))];
+        });
+    }
+
+    /***
+     * Get details about the issue a submission will be published in.
```
jonasraoni: Add more typing to the `@return`, see https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

classes/submission/maps/Schema.php
```diff
@@ -1191,4 +1221,60 @@ protected function getStageFilesBySubmissions(Enumerable $submissions, array $st
...
+     */
+    protected function getPropertyIssueToBePublished(Enumerable $publications): ?array
+    {
+        /** @var Publication $latestScheduledPublication */
+        $latestScheduledPublication = $publications
+            ->filter(fn($publication) => $publication->getData('status') === Submission::STATUS_SCHEDULED)
```
jonasraoni: Not sure if I understood well the task, but perhaps the `STATUS_PUBLISHED` should be included here.

jonasraoni (changes requested):

taslangraham:
The initial functionality implemented on this PR was moved to https://github.com/pkp/ojs/pull/4746 as the changes needed to be more app specific to OJS

ewhanson (approved):

PR #11106, opened 2025-03-16
Update emails.po

fgnievinski:
enclose submission title in double quotes for better readability.

it's already present in the following email templates: 
- emails.orcidRequestUpdateScope.body
- emails.orcidRequestAuthorAuthorization.body

(It might be possible to extend this change automatically to translations, although the quotation mark is language specific. Also notice some translations already include quotes, even if they are absent in English.)


asmecher:
Thanks, @fgnievinski -- I've merged into `main` and cherry-picked back to `stable-3_5_0`.

Issue #10766, opened 2025-01-03
Redesigned Dashboard & Workflow - clean up previous implementations


jardakotesovec:
PRs:
- https://github.com/pkp/ojs/pull/4756
- https://github.com/pkp/omp/pull/1910
- https://github.com/pkp/ops/pull/922
- https://github.com/pkp/pkp-lib/pull/11111

asmecher:
@jardakotesovec and @Vitaliy-1, I merged the attached PRs and will close this for 3.5.0. It's fairly likely that there are still a few pieces of dead code, but it's not critical that we get everything for 3.5.0.

PR #11111, opened 2025-03-17
pkp/pkp-lib#10766 Remove support code for old submission lists

asmecher:


classes/core/PKPApplication.php
```diff
@@ -170,6 +170,7 @@ class_alias('\PKP\payment\QueuedPayment', '\QueuedPayment'); // QueuedPayment in
         }
 
         // Ensure that nobody registers for hooks that are no longer supported
+        Hook::addUnsupportedHooks('API::_submissions::params', 'Template::Workflow::Publication', 'Template::Workflow', 'Workflow::Recommendations'); // pkp/pkp-lib#10766 Removed with new submission lists for 3.5.0
```
jardakotesovec, line 173: @Vitaliy-1 Just wondering, should not we add hooks for the new _submissions endpoints?

Issue #11087, opened 2025-03-13
Consider role masthead options in the old UserForm


bozana:
Provide possibility to define if the user will be displayed on the masthead for a role in the old UserForm, that currently can be accessed by admin.

bozana:
Hi @asmecher, would you maybe be able to take a look at the PR above?
Thanks a lot!

asmecher:
Added [a review](https://github.com/pkp/pkp-lib/pull/11128#pullrequestreview-2699652914) to the `stable-3_5_0` PR. Looks like the link to the `main` PR got lost in the head comment?

bozana:
Hi @asmecher, thanks, I will consider the comments. I then thought to cherry-pick it all to main, when finished, that's why the main PR is missing...

bozana:
Hi @asmecher, oh good comments -- I realized also other parts of UserForm should be improved.
Could you please take another look?
Thanks a lot!

asmecher:
@bozana, I made another few individual comments, thanks! This might have to wait post-3.5.0rc2.

asmecher:
@Tribunal33, is this ready to be closed?

bozana:
Hi @Tribunal33,
The checkboxes "Appear on Masthead" should be all selected per default (for the roles the user does not have) and according the existing/saved settings for the roles that user has. This checkboxes do not relate to the User&Roles > Roles setting -- they define if the user agrees to be displayed on the masthead for that specific role. Even if a role should not be displayed on the masthead we need to know if the user agrees i.e. should be displayed on the masthead for that role, in case journal manager decides later to display that role on the masthead.
When the form is saved, only the settings for the roles the user currently has are considered -- if the user should be displayed for that role on the masthead. Thus, in your screenshot, it seems that the user does not have the translator role, that is why the changes are not saved?

The issue number 2 you have mentioned is strange and has something to do with the invitations -- not with this old admin user form. Could you please double check this invitation issue again -- I will do the same locally -- and if it does not behave correctly, we should create a new issue and fix it.

Thanks a lot!

Tribunal33:
Hi @bozana you're correct.  The admin > hosted Journal > Setting Wizard > User > Edit works exactly how you describe.  I think we can mark that as working as intended.  

For number 2.  You're also correct.  I think this has something to do with Initivation workflow and is unrelated to this case.  I will create a bug for it as I can reproduce it.  We can close this issue, the editorial masthead works without issue.

PR #11128, opened 2025-03-19
pkp/pkp-lib#11087 consider masthead options in the old UserForm

bozana:
s. https://github.com/pkp/pkp-lib/issues/11087


controllers/grid/settings/user/form/UserForm.php
```diff
@@ -62,12 +64,24 @@ public function initData()
...
                 $userGroupIds[] = $userGroup->id;
-                $masthead[$userGroup->id] = Repo::userGroup()->userOnMasthead($this->userId, $userGroup->id);
             }
+
+            // Get user group IDs for user groups this user should not be displayed on the masthead for
+            // The others will be selected per default
+            $notOnMastheadUserGroupIds = UserUserGroup::query()
+                ->withUserId($this->userId)
+                ->withActive()
+                ->where('user_user_groups.masthead', UserUserGroupMastheadStatus::STATUS_OFF)
```
asmecher, line 74: It would be better to use a scoping function in `UserUserGroup` rather than exposing a `lower_camel_case` table name here and adding an explicit `where` clause. SQL building should be kept in the container.

controllers/grid/settings/user/form/UserForm.php
```diff
@@ -159,6 +177,24 @@ public function saveUserGroupAssignments(Request $request): void
...
+            $reviewerUserGroupIds = Repo::userGroup()->getArrayIdByRoleId(Role::ROLE_ID_REVIEWER, $contextId);
+            collect($this->getData('userGroupIds'))
+                ->filter(
+                    function ($userGroupId) use ($reviewerUserGroupIds) {
+                        return !in_array($userGroupId, $reviewerUserGroupIds);
+                    }
+                )
+                ->each(
+                    fn ($userGroupId) =>
+                    UserUserGroup::query()
```
asmecher, line 192: I think this is missing a check to ensure that the user-specified `userGroupIds` are in the right context.

asmecher (changes requested):
A couple changes to request!

controllers/grid/settings/user/form/UserForm.php
```diff
@@ -62,12 +64,24 @@ public function initData()
...
+
+            // Get user group IDs for user groups this user should not be displayed on the masthead for
+            // The others will be selected per default
+            $notOnMastheadUserGroupIds = UserUserGroup::query()
+                ->withUserId($this->userId)
+                ->withActive()
+                ->withMastheadOff()
+                ->get()
+                ->map(
+                    fn (UserUserGroup $userUserGroup) => $userUserGroup->user_group_id
```
asmecher, line 77: Does the `HasCamelCasing` trait not allow the use of `userGroupId` here? `lower_camel_case` columns should be abstracted by the container level.

classes/userGroup/relationships/UserUserGroup.php
```diff
@@ -100,6 +100,11 @@ public function scopeWithMasthead(Builder $query): Builder
         return $query->where('user_user_groups.masthead', 1);
     }
 
+    public function scopeWithMastheadOff(Builder $query): Builder
+    {
+        return $query->where('user_user_groups.masthead', 0);
```
asmecher, line 105: Maybe too late to change this for 3.5.0, but I think when we introduce enumerations, we should use the same values in the DB that we do in PHP-land. This introduces string constants (`null`, `on`, `off`, `all`) in PHP, but different numeric constants in the DB (`1`, `0`, presumably other values). It would be better just to use one set of constants, even if they're numeric.

controllers/grid/settings/user/form/UserForm.php
```diff
@@ -92,17 +106,21 @@ public function display($request = null, $template = null)
...
         $templateMgr = TemplateManager::getManager($request);
 
-        $allUserGroups = [];
+        $allUserGroups = $defaultMastheadUserGroups = [];
 
         $userGroups = UserGroup::withContextIds([$contextId])->get();
 
         foreach ($userGroups as $userGroup) {
             $allUserGroups[(int) $userGroup->id] = $userGroup->getLocalizedData('name');
+            if ($userGroup->role_id != Role::ROLE_ID_REVIEWER) {
```
asmecher, line 115: `lower_camel_case` outside of the storage level

controllers/grid/settings/user/form/UserForm.php
bozana: Oh, yes, my fault... Corrected...

classes/userGroup/relationships/UserUserGroup.php
bozana: Hmmm... :thinking: 
These constants are originally meant to be for the user Repository, for filtering there: If one would like to filter by mastheadStatusOff i.e. to get users that are not displayed on the masthead, it would consider both masthead = 0 in the DB table user_groups, as well as masthead = 0 or null in the DB table user_user_groups. Also mastheadStatusAll would get all users no matter what masthead status they have. 
But then these constants are also used for user-user-groups functions and filtering there, to only get the information if the user has agreed to be displayed, if the user has disagreed, or if it is undefined (no matter if the user group is defined to be displayed on the masthead). 
So, maybe we would need two enums, for user and userUserGroup? -- It seems this way it would be cleaner... :thinking:

bozana:
Ah, sorry @asmecher, I pushed the wrong branch... :see_no_evil:

asmecher:
Thanks, @bozana, this looks good. I've reviewed but not tested; please go ahead, then assign the issue over to Beau for a spin through QA.

Issue #11118, opened 2025-03-18
Clean up specific invitation code mixed into general toolset


asmecher:
The [generic `InvitationHandler` class](https://github.com/pkp/pkp-lib/blob/main/pages/invitation/InvitationHandler.php) accidentally got some invitation-specific code mixed into it with https://github.com/pkp/pkp-lib/pull/10576. The code specific to the user invitation should be moved to its respective handler, leaving `InvitationHandler` a generic tool for any kind of invitation.

defstat:
@asmecher @withanage, @ipula has prepared the following PRs for this issue:

- `main`: #11272

- `stable_3_5_0`: #11266

These PRs complete the separation of generic and invitation-specific code relevant to this issue.
The respective ui-lib changes were merged yesterday separately. 

@withanage @ipula @Tribunal33
There's a separate case in the "Edit User" flow — when a userId is provided that doesn't exist, it currently throws an HTTP 500 error. I believe that could be handled in a separate issue.

@asmecher
I don’t have the necessary permissions to fully merge the above PRs. Would you be able to assist with that?

jardakotesovec:
Merged, sorry on jumping gun on the ui-library part..  I did not realise that is pkp-lib part while going through various PRs.. even though its quite obvious..

Tests are passing for me locally with these changes.. will watch out for that on the GA as well..

ipula:
> [@asmecher](https://github.com/asmecher) [@withanage](https://github.com/withanage), [@ipula](https://github.com/ipula) has prepared the following PRs for this issue:
> 
>     * `main`: [refactor invitation page handler #11272](https://github.com/pkp/pkp-lib/pull/11272)
> 
>     * `stable_3_5_0`: [refactor invitation page handler #11266](https://github.com/pkp/pkp-lib/pull/11266)
> 
> 
> These PRs complete the separation of generic and invitation-specific code relevant to this issue. The respective ui-lib changes were merged yesterday separately.
> 
> [@withanage](https://github.com/withanage) [@ipula](https://github.com/ipula) [@Tribunal33](https://github.com/Tribunal33) There's a separate case in the "Edit User" flow — when a userId is provided that doesn't exist, it currently throws an HTTP 500 error. I believe that could be handled in a separate issue.
> 
> [@asmecher](https://github.com/asmecher) I don’t have the necessary permissions to fully merge the above PRs. Would you be able to assist with that?

[...]

PR #11137, opened 2025-03-19
clean up invtation handler

ipula:
https://github.com/pkp/pkp-lib/issues/11118

### DIFF
--- a/classes/invitation/invitations/userRoleAssignment/resources/BaseUserRoleAssignmentInviteResource.php
+++ b/classes/invitation/invitations/userRoleAssignment/resources/BaseUserRoleAssignmentInviteResource.php
@@ -17,8 +17,11 @@ namespace PKP\invitation\invitations\userRoleAssignment\resources;
 use APP\facades\Repo;
 use Illuminate\Http\Request;
 use Illuminate\Http\Resources\Json\JsonResource;
+use PKP\context\Context;
+use PKP\facades\Locale;
 use PKP\invitation\invitations\userRoleAssignment\payload\UserRoleAssignmentInvitePayload;
 use PKP\user\User;
+use PKP\userGroup\relationships\UserUserGroup;
 
 class BaseUserRoleAssignmentInviteResource extends JsonResource
 {
@@ -90,5 +93,25 @@ class BaseUserRoleAssignmentInviteResource extends JsonResource
         $newUser->setEmail($payload->sendEmailAddress);
 
         return $newUser;
+    }
+
+    protected function transformCurrentUserGroups(int $id , Context $context): array
+    {
+        $userGroups = [];
+        $userUserGroups = UserUserGroup::query()
+            ->withUserId($id)
+            ->withContextId($context->getId())
+            ->get()
+            ->toArray();
+        foreach ($userUserGroups as $key => $userUserGroup) {
+            $userGroup = Repo::userGroup()
+                ->get($userUserGroup['userGroupId'])
+                ->toArray();
+            $userGroups[$key] = $userUserGroup;
+            $userGroups[$key]['masthead'] = $userUserGroup['masthead'] === 1;
+            $userGroups[$key]['name'] = $userGroup['name'][Locale::getLocale()];
+            $userGroups[$key]['id'] = $userGroup['userGroupId'];
+        }
+        return $userGroups;
     }
 }
--- a/classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
+++ b/classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
@@ -15,6 +15,8 @@
 namespace PKP\invitation\invitations\userRoleAssignment\resources;
 
 use Illuminate\Http\Request;
+use PKP\context\Context;
+use PKP\invitation\invitations\userRoleAssignment\payload\UserRoleAssignmentInvitePayload;
 use PKP\user\User;
 
 class UserRoleAssignmentInviteResource extends BaseUserRoleAssignmentInviteResource
@@ -64,5 +66,56 @@ class UserRoleAssignmentInviteResource extends BaseUserRoleAssignmentInviteResou
             'existingUser' => $this->transformUser($this->getExistingUser()),
             'newUser' => $this->transformUser($newUser),
         ]);
+    }
+
+    /**
+     * Transform invitation payload
+     * @param int|null $userId
+     * @param array $payload
+     * @param Context $context
+     * @return UserRoleAssignmentInvitePayload
+     */
+    public function transformInvitationPayload(?int $userId, array $payload, Context $context): object
+    {
+        $invitationPayload = UserRoleAssignmentInvitePayload::fromArray($payload)->toArray();
+        $invitationPayload['userId'] = !$userId ? null : $userId;
+        $invitationPayload['inviteeEmail'] = !$userId ?$invitationPayload['sendEmailAddress']:$payload['email'];
+        $invitationPayload['userGroupsToAdd'] = !$invitationPayload['userGroupsToAdd'] ? [] :$invitationPayload['userGroupsToAdd'];
+        $invitationPayload['currentUserGroups'] = !$userId ? [] : $this->transformCurrentUserGroups($userId,$context);
+
+        return (object)$invitationPayload;
+    }
+
+    /**
+     * Transform user data for invitation view
+     * @param User $user
+     * @param $context
+     * @return array
+     */
+    public function transformInvitationUserData(User $user, $context): array
+    {
+        $userData = [];
+        $userData['country'] = $user->getCountryLocalized();
+        $userData['biography'] = $user->getBiography(null);
+        $userData['phone'] = $user->getPhone();
+        $userData['mailingAddress'] = $user->getMailingAddress();
+        $userData['signature'] = $user->getSignature(null);
+        $userData['locales'] = $this->transformUserWorkingLanguages($context, $user->getLocales());
+        $userData['reviewInterests'] = $user->getInterestString();
+        $userData['homePageUrl'] = $user->getUrl();
+        $userData['disabled'] = $user->getData('disabled');
+        return $userData;
+    }
+
+    /**
+     * get user working languages
+     * @param Context $context
+     * @param array $userLocales
+     * @return string
+     */
+    private function transformUserWorkingLanguages(Context $context, array $userLocales): string
+    {
+        $locales = $context->getSupportedLocaleNames();
+        return join(__('common.commaListSeparator'), array_map(fn($key) => $locales[$key], $userLocales));
     }
 }
--- a/pages/invitation/InvitationHandler.php
+++ b/pages/invitation/InvitationHandler.php
@@ -20,16 +20,9 @@ use APP\core\Application;
 use APP\core\Request;
 use APP\facades\Repo;
 use APP\handler\Handler;
-use APP\template\TemplateManager;
-use PKP\context\Context;
-use PKP\core\PKPApplication;
-use PKP\facades\Locale;
-use PKP\i18n\LocaleMetadata;
 use PKP\invitation\core\enums\InvitationAction;
 use PKP\invitation\core\Invitation;
-use PKP\invitation\stepTypes\SendInvitationStep;
-use PKP\user\User;
-use PKP\userGroup\relationships\UserUserGroup;
+use PKP\invitation\invitations\userRoleAssignment\handlers\UserRoleAssignmentInviteUIController;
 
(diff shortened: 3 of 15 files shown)


classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteRedirectController.php
```diff
@@ -13,22 +13,26 @@
...
 use PKP\invitation\core\InvitationActionRedirectController;
+use PKP\invitation\invitations\userRoleAssignment\resources\UserRoleAssignmentInviteResource;
 use PKP\invitation\invitations\userRoleAssignment\UserRoleAssignmentInvite;
 use PKP\invitation\stepTypes\AcceptInvitationStep;
+use PKP\invitation\stepTypes\SendInvitationStep;
 
 /**
  * @extends InvitationActionRedirectController<UserRoleAssignmentInvite>
  */
 class UserRoleAssignmentInviteRedirectController extends InvitationActionRedirectController
```
defstat, line 33: Here I find it a little confusing. The RedirectController of an invitation is designed to be used for the accept and decline handle of the invitation. 

Here it is used for other purposes, that seem to mostly meet the needs of the UI rather, than the needs of the business logic. 

Could we explore the possibility of assigning the "other" operations to another handler, that can reside within the invitation/userRoleAssignment/handlers folder, but will be used for the needs of the UI instantiation etc.

This will remove the need to, for example, cover the case of a null Invitation at the level of the RedirectController. Also, and most importantly, will separate the concerns of the handlers, and reduce the complexity of the core part of the code. 

Could we also utilize the `preRedirectActions` for the setup of the UI? 

What do you think?

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
```diff
@@ -65,4 +67,40 @@ public function toArray(Request $request)
             'newUser' => $this->transformUser($newUser),
         ]);
     }
+
+    public function transformInvitationPayload($userId, array $payload, Context $context)
```
defstat, line 71: Revisiting this, can't we use the $invitationPayload object instead of an array? If no, what prevents us from doing that?

pages/management/ManagementHandler.php
```diff
@@ -579,6 +583,25 @@ public function manageEmails(array $args, Request $request): void
...
+     * @param $request
+     * @return void
+     * @throws \Exception
+     */
+    public function editUser(array $args, $request): void
+    {
+        $this->setupTemplate($request);
+        $invitation = null;
+        if(!empty($args)) {
+            $invitationController = new UserRoleAssignmentInviteRedirectController($invitation);
```
defstat, line 598: See comment above regarding this.

pages/management/ManagementHandler.php
```diff
@@ -579,6 +583,25 @@ public function manageEmails(array $args, Request $request): void
...
+            $invitationController->inviteHandle($request,$args[0],'editUser');
```
defstat, line 599: Here, perhaps, define in a separate $variable what the `$args[0]` is, for clarity?

Here, we should also consider standardizing the 'editUser' in a Enumeration instead of the "hardcoded" string?

classes/invitation/invitations/userRoleAssignment/resources/BaseUserRoleAssignmentInviteResource.php
```diff
@@ -91,4 +94,36 @@ protected function createNewUserFromPayload(UserRoleAssignmentInvitePayload $pay
...
+        return $userGroups;
+    }
+
+    /**
+     * get user working languages
+     * @param Context $context
+     * @param $userLocales
+     * @return string
+     */
+    protected function transformWorkingLanguages(Context $context,$userLocales): string
```
defstat, line 124: Here can we define all the parameters types? 

Also could you leave an empty space after the `,` here and elsewhere?

defstat (changes requested):
@ipula Thanks. I have added some review comments for this.

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteRedirectController.php
ipula: I think better to move this in to separate class like `UserRoleAssignmentInviteUIController` under the `invitation/userRoleAssignment/handlers ` for UI instantiation

pages/invitation/InvitationHandler.php
```diff
@@ -123,250 +124,12 @@ public static function getActionUrl(InvitationAction $action, Invitation $invita
      */
     public function invite(array $args, Request $request): void
```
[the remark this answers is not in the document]
defstat: @ipula this is still unresolved

pages/invitation/InvitationHandler.php
```diff
@@ -20,16 +20,9 @@
...
-use PKP\facades\Locale;
-use PKP\i18n\LocaleMetadata;
 use PKP\invitation\core\enums\InvitationAction;
 use PKP\invitation\core\Invitation;
-use PKP\invitation\stepTypes\SendInvitationStep;
-use PKP\user\User;
-use PKP\userGroup\relationships\UserUserGroup;
+use PKP\invitation\invitations\userRoleAssignment\handlers\UserRoleAssignmentInviteUIController;
 
 class InvitationHandler extends Handler
```
defstat, line 27: at the first line of the `public function accept(array $args, Request $request): void` is the `$this->setupTemplate($request);` line needed?

defstat (changes requested):
@ipula thanks! There are still unresolved and new comments on that change.

pages/invitation/InvitationHandler.php
defstat: Is it is needed, why isn't it needed in at the `decline` function also?

ipula: this is needed initialized the template. without that UI will not render

ipula: image

pages/invitation/InvitationHandler.php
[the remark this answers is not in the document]
defstat: @ipula this is still unresolved

pages/invitation/InvitationHandler.php
```diff
@@ -117,256 +110,18 @@ public static function getActionUrl(InvitationAction $action, Invitation $invita
...
-     * get user working languages
-     * @param Context $context
-     * @param $userLocales
-     * @return string
-     */
-    private function getWorkingLanguages(Context $context,$userLocales): string
-    {
-        $locales = $context->getSupportedLocaleNames();
-        return join(__('common.commaListSeparator'), array_map(fn($key) => $locales[$key], $userLocales));
+        $invitationController = new UserRoleAssignmentInviteUIController($invitation);
```
defstat, line 124: @ipula This is still unresolved. we have an Invitation specific code here. Can you tell me the need of this the `userInvite` function here?

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
```diff
@@ -65,4 +68,51 @@ public function toArray(Request $request)
...
+        // removing security related data
+        unset($invitationPayload["username"]);
+        unset($invitationPayload["password"]);
+        unset($invitationPayload["orcidAccessToken"]);
+        unset($invitationPayload["orcidRefreshToken"]);
+        unset($invitationPayload["passwordHashed"]);
+
+        return [
+            'invitationPayload' => $invitationPayload,
+            'user' => $user
```
defstat, line 115: Here why do we need to return the User Object?

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
```diff
@@ -65,4 +68,51 @@ public function toArray(Request $request)
...
+     */
+    public function transformInvitationPayload(?int $userId, array $payload, Context $context): array
+    {
+        $invitationPayload = UserRoleAssignmentInvitePayload::fromArray($payload)->toArray();
+        $user = null;
+        if($userId){
+            $user = Repo::user()->get($userId,true);
+            $invitationPayload = UserRoleAssignmentInvitePayload::fromArray($user->getAllData())->toArray();
+        }
+        $invitationPayload['userId'] = $user ? $user->getId() : $userId;
```
defstat, line 87: Can we use the Payload object instead of the array here? If not, why?

pages/invitation/InvitationHandler.php
```diff
@@ -117,256 +110,18 @@ public static function getActionUrl(InvitationAction $action, Invitation $invita
     }
```
defstat, line 110: in `public function accept(array $args, Request $request): void` the first call is `$this->setupTemplate($request);`. Don't we need to add the same in `public function decline(array $args, Request $request): void` function?

defstat (changes requested):
@ipula thanks. some comments from the previous review remained unresolved. Just adding them again here with 1 addition.

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
ipula: because we used this with edit user as well. I used this user query here because I have to use that in two places if I not used here

pages/invitation/InvitationHandler.php
ipula: we do not need that because , there is no view for decline the invitations. Once user click the decline in email , user will be redirect to back

pages/invitation/InvitationHandler.php
ipula: this will use the page route URL function `userInvite`

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
defstat: OK but I am afraid that this does not separate the concerns of the code. Do you think we could prevent that?

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php
ipula: I convert this to array to add additional data indexes for the payload that I need in UI

pages/invitation/InvitationHandler.php
```diff
@@ -117,256 +110,29 @@ public static function getActionUrl(InvitationAction $action, Invitation $invita
...
 
     /**
-     * Create an invitation to accept new role
-     *
+     * Create an invitation for a user to accept new roles
      * @throws \Exception
      */
-    public function invite(array $args, Request $request): void
+    public function userInvite(array $args, Request $request): void
```
defstat, line 116: This function constitutes a security breach here. This handler does not have an `authorize` function. 

`accept` and `decline` functions depend on a provided key and id. The key is secret and is send only to the indented user. That is why this handler does not have any `authorization` code.

But having this  `userInvite` function, anyone can access any invitation I think.

If we can't solely depend on the API for the creation of the UI, I think we should create another handler here, with the appropriate `authorization` checks and policies.

pages/invitation/InvitationHandler.php
```diff
@@ -117,256 +110,29 @@ public static function getActionUrl(InvitationAction $action, Invitation $invita
...
-                        ['name' => $invitationPayload['givenName'][Locale::getLocale()]]
-                    ),
-            ]);
-            $templateMgr->assign([
-                'pageComponent' => 'Page',
-                'breadcrumbs' => $breadcrumbs,
-                'pageWidth' => TemplateManager::PAGE_WIDTH_FULL,
-            ]);
-            $templateMgr->display('/invitation/userInvitation.tpl');
+        $invitationType = $args[0]; // invitation type
```
defstat, line 123: Here, the way it is developed, I think there is no need for both args, right? either the type or the id is needed.

defstat (changes requested):
@ipula Thanks for your work on this! I’ve left a few comments that are important for ensuring everything works as expected.

I also proposed some changes here: https://github.com/ipula/pkp-lib/pull/4 — could you please review them and test if they work for the specific invitation case?

You'll also need to make sure the UI is updated to call the new handler introduced in that PR.

Once we confirm this works, we can move on to test the other invitation types and address any potential security concerns.

Thanks again — let me know if anything’s unclear or if you'd like to discuss further. Ping me when ready.

ipula:
close this PR. this related to `stable-3_5_0`

Issue #11011, opened 2025-02-28
[Website Settings] | Setup - Language, Unable to set navigation item title with multiple languages


bg364:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

No ability to update the navigation item title when you don't complete all languages during it's creation. After saving titles in other languages are not saved.
Using OJS 3.4.0.8 (latest at the moment).

### Steps to Reproduce

You need to have a journal with enabled more than one language in it.

![Image](https://github.com/user-attachments/assets/e54bf913-3889-43df-9a7d-95d2d98187b6)

1. Log in as administrator
2. Go to Journals website settings -> Setup tab -> Navigation (/management/settings/website#setup/navigationMenus)
3. Create new Navigation Menu Item - but fill the title only in one language

![Image](https://github.com/user-attachments/assets/0806a3e5-2b15-48a8-9340-1a7e25f6b802)
4. Select any type (it doesn't matter) and save it.
5. Edit it and try to fill other languages:

![Image](https://github.com/user-attachments/assets/4b490e1b-5ab5-48c9-94d0-f220d5e32661)
6. Save it

### Expected Result

Titles should be saved in all languages

### Actual Result

Only filled in languages are updated

### Environment Details

PHP Version 8.2.27, Debian, MySQL
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36

### Application Version

OJS 3.4.0.8

### Additional Information

It seems like a bug in `lib/pkp/controllers/grid/navigationMenus/form/PKPNavigationMenuItemsForm.php`
In line 190 there is a loop `foreach ($localizedTitles as $locale => $title) {` that goes over a list of languages that are currently saved in said navigation item. It should go over of list of languages send by user instead.

---
**PRs**

Branch **stable-3_4_0**
pkp-lib: #11139
OJS: pkp/ojs#4766

~~pkp-lib: #11103~~
~~OJS: pkp/ojs#4747 - **TESTS ONLY**~~

Branch **stable-3_5_0**
pkp-lib: #11140
OJS: pkp/ojs#4769

~~pkp-lib: #11104~~
~~OJS: pkp/ojs#4748 - **TESTS ONLY**~~

Branch **main**
pkp-lib: #11141
OJS: pkp/ojs#4768

~~pkp-lib: #11105~~
[...]

PR #11139, opened 2025-03-19
[PKP-LIB][stable-3_4_0] #11011 Fix title assignment functionality

defstat:


controllers/grid/navigationMenus/form/PKPNavigationMenuItemsForm.php
```diff
@@ -180,19 +180,44 @@ public function execute(...$functionParams)
...
+            // This is not always the case for the default NMIs that, by default, get their titles
+            // from the titleLocaleKey setting of the NMI.
+
+            // Get the title that have been explicitely set in the settings tatble
+            $localizedTitlesFromDB = $navigationMenuItem->getTitle(null) ?? [];
+
+            // Get the set of all displayed titles, no matter the source. 
+            Services::get('navigationMenu')
+                ->setAllNMILocalizedTitles($navigationMenuItem);
+            $originalDisplayTitles = $navigationMenuItem->getTitle(null);
```
jonasraoni, line 194: As this is basically the same as the variable `$localizedTitlesFromDB`, I think it can be removed.

jonasraoni: I just remembered this code is cursed, ignore the comment 🫠

controllers/grid/navigationMenus/form/PKPNavigationMenuItemsForm.php
```diff
@@ -180,19 +180,44 @@ public function execute(...$functionParams)
...
+                // Check if there is a display title for the NMI. If not for this locale, explicitely define one
+                if (!isset($originalDisplayTitles[$locale])) {
+                    $newTitles[$locale] = $trimmedTitle;
+                    continue;
+                }
+
+                // Update title if explicitly defined or changed from default
+                if (isset($localizedTitlesFromDB[$locale]) || $originalDisplayTitles[$locale] !== $trimmedTitle) {
+                    $newTitles[$locale] = $trimmedTitle;
+                }
```
jonasraoni, line 220: I think this can be left as:

```suggestion
                // Update title if explicitly defined or changed from default
                if (isset($localizedTitlesFromDB[$locale]) || ($originalDisplayTitles[$locale] ?? null) !== $trimmedTitle) {
                    $newTitles[$locale] = $trimmedTitle;
                }
```

defstat: @jonasraoni are those two equivalent or you think that the suggested one incorporates the correct approach for this case?

defstat: What do you say, to leave it in two different clauses for clarity (to an already "difficult" code segment to read)?

PR #11140, opened 2025-03-19
[PKP-LIB][stable-3_5_0] #11011 Fix title assignment functionality

defstat:


jonasraoni (changes requested):
The review was left on the other branch.

Issue #11123, opened 2025-03-18
Submission Workflow- Internal Review, Author cannot upload revisions in internal review stage


mbakioglu:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Test: Take a new submission through the internal and external review stages
Status: Failed

Description:
The submission progresses through the internal review stage as expected. However, after the Press Editor requests a revision, the author is unable to submit their revisions. When attempting to do so, the author encounters the following error:
"Error: You are not allowed to add and edit these files."

### Steps to Reproduce

Steps to Reproduce
1.	Log in as a Press Editor.
2.	Request a revision while the submission is in the internal review stage.
3.	Log in as the Author.
4.	View the revision request notification and the “Submit Revision” option.
 
![Image](https://github.com/user-attachments/assets/e6fbd8ad-6dd3-47aa-9556-ba93ecf35cb9)

5.	Click "Submit Revision."
6.	Observe the error message: "You are not allowed to add and edit these files."
7.	Alternatively, there is no available option to submit a revision within the submission interface.
 ![Image](https://github.com/user-attachments/assets/43bc0f22-3945-402c-bfe5-695c1a0a229d)

### Expected Result

The author should be able to submit their revisions either via the button on the dashboard or through the submission screen.

### Actual Result

The author is unable to submit their revisions either via the button on the dashboard or through the submission screen.

### Application Version

OMP 3.5

Tribunal33:
I believe it could be part of this added feature #10688. This is why I gave it the interal milestone.

Tribunal33:
We will check in RC2.  I'm seeing some potential issues.  I'm not sure what the requirements are.  I will get these confirmed before we attempting more fixes for this one.

jardakotesovec:
One issue that I identified is that in 3.4 editorial decisions were refactored and separate decisions were introduced for internal review round.
When author attempts to upload file - we have logic that will let him to do it only if there is decision for given review round - it has to be accepted/revisions requested/new round created.

So basically the fix is just to check for correct decisions for internal review stage.

pkp-lib (3.4) - https://github.com/pkp/pkp-lib/pull/11183
@Vitaliy-1 Can you check it out.

@Tribunal33 Is there any other scenario that you are concerned about and we should check?

Tribunal33:
This passes now.  Tested with latest on main.  I am able to upload revisions in the internal review round once revisions have been requested.  I can use the editorial activity button and view the internal review stage as an author.

PR #11183, opened 2025-03-27
pkp/pkp-lib#11123 Fix: consider correct editorial decisions when auth…

jardakotesovec:
…or is uploading revision in internal review round


Vitaliy-1 (approved):
Looks good!

Issue #10692, opened 2024-12-12
Structured Citations


GaziYucel:
Update 01.10.2025

Still to finish:
- [ ] test native import export
- [ ] adapt article_details.tpl for each theme
- [ ] adapt jatsTemplate
- [ ] adapt crossrefReferenceLinking 

<hr>

### Overview
The structured citations feature in OJS is designed to streamline the process of managing references within submissions. It allows authors and editors to input, process, and manage citations efficiently, ensuring accurate metadata retrieval and integration.

### Acceptance criteria and detailed design specifications

**`User Path 1: Enable Structured Citations Setting at the Journal Level`**

- A toggle is available under Settings > Workflow > Submission > Metadata to enable or disable Structured Citations by default. 
- However, this setting is not mandatory—editors can override it for individual submissions by going to the Publication > References tab and adjusting the setting there 

**`User Path 2: Structured Citations during Article Submission Process`**

- When a new submission is started by the author or editor but has not yet been submitted, the References field will appear as a simple text input with no structured citations.
- Once the submission is officially submitted to the journal, the Structured Citations feature becomes available in the Publication tab and can be accessed from all workflow stages.

**`User Path 3: Author or Editor Adds Citations During Workflow`**

**Submission > References Tab**
- A dedicated References section allows authors (or editors) to input references as free-text, with each reference entered on a new line.

- A brief instruction appears above the field:

> “Enter each reference on a new line. You can paste one or multiple references at a time. Click ‘Add’ to process them into structured citations.”

**"Add" Button Behavior**
- Clicking Add triggers metadata lookup via external services (e.g., Crossref) and attempts to match the reference with available DOI and citation metadata.
- While processing, a status message appears:

> “Processing references… 0 of 4 added so far. This may take a few moments. You can continue working while the references are being retrieved.”

- A caution is also displayed:

> “Some references may not return metadata. You can edit them manually once they’re added to the list.”

**`User Path 4: Display of Processed References in Structured Format`**

- Once processed, structured references appear in a tabular list below the text box.
- A success message is displayed:

> “All references successfully added. You can review, edit, or remove them at any time.”

**Table Actions**

Each reference row includes:
- Formatted citation (retrieved or entered manually)
- DOI (if matched)
- Action buttons: Edit | Delete
- Clicking on "Edit" will open a modal where the user can fill more information about the citations or modify the existing one to make it structured

If users paste new references into the textbox and click Add again:

- The system detects and adds only unique entries.
[...]

PR #11213, opened 2025-04-01
Structured Citations DB

GaziYucel:
https://github.com/pkp/pkp-lib/issues/10692


bozana (commented):
Hi @GaziYucel, 
I went through the code and left a few comments. I have not tested it yet, but I will do soon, and will then let you know...
Thanks a lot!

classes/citation/Citation.php
```diff
@@ -20,99 +16,70 @@
...
+class Citation extends DataObject
+{
     /**
-     * Get the rawCitation
-     *
-     * @return string
+     * Get the rawCitation.
      */
-    public function getRawCitation()
+    public function getRawCitation(): string|null
```
bozana, line 26: I am wondering when it is the case where the rawCitation would not exist? :thinking:

classes/citation/Citation.php
```diff
@@ -20,99 +16,70 @@
...
         $this->setData('seq', $seq);
     }
 
-    //
-    // Private methods
-    //
+    /**
+     * Replace URLs through HTML links, if the citation does not already contain HTML links.
+     */
+    public function getRawCitationWithLinks(): string|null
```
bozana, line 59: The same here, if rawCitation always exist, this function will always return string

classes/citation/Collector.php
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        switch ($this->orderBy) {
+            case self::ORDERBY_SEQUENCE:
+                $qb->orderBy('c.seq', 'asc');
+                break;
+            default:
+                $qb->orderBy('c.citation_id', 'asc');
+                break;
+        }
```
bozana, line 128: I think we can say that the citations are always ordered by sequence, no?

classes/citation/Repository.php
```diff
@@ -0,0 +1,226 @@
...
+    {
+        $id = $this->dao->insert($citation);
+        Hook::call('Citation::add', [$citation]);
+        return $id;
+    }
+
+    /** @copydoc DAO::update() */
+    public function edit(Citation $citation, array $params): void
+    {
+        $newRow = clone $citation;
```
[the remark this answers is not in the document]
bozana: Would you like to change the name of this variable?

schemas/citation.json
```diff
@@ -0,0 +1,257 @@
...
+		"lastModified": {
+			"type": "string",
+			"description": "The last time a modification was made to this record.",
+			"apiSummary": true,
+			"validation": [
+				"date_format:Y-m-d H:i:s"
+			]
+		}
+	}
+}
```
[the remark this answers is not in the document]
bozana: I think we do not need lastModified now, or?

classes/citation/CitationDAO.php
```diff
@@ -1,253 +0,0 @@
...
-                    $citation->setSequence($seq + 1);
-
-                    $this->insertObject($citation);
-
-                    $importedCitations[] = $citation;
-                }
-            }
-        }
-
-        Hook::call('CitationDAO::afterImportCitations', [$publicationId, $existingCitations, $importedCitations]);
```
bozana, line 121: Can you please add this hook removal to the list here https://github.com/pkp/pkp-lib/blob/main/classes/core/PKPApplication.php#L173

schemas/publication.json
```diff
@@ -59,12 +59,14 @@
 			"description": "Optional metadata that contains an array of references for works cited in this submission. References have been split and parsed from the raw text.",
 			"readOnly": true,
 			"items": {
-				"type": "string"
+				"type": "object",
+				"$ref": "#/definitions/Citation"
 			}
 		},
-		"citationsRaw": {
+		"rawCitations": {
```
bozana, line 66: I believe it was called "citationsRaw" in order to be direct behind "citations", because the properties are alphabetically ordered. 
So maybe to leave it as it was i.e. "citationsRaw"

PR #11214, opened 2025-04-01
ReCAPTCHA on User Lost Password Page

Godoy0722:
*Description*: Add ReCAPTCHA on the `userLostPassword.tpl` page. Solves the issue #6984.

### DIFF
--- a/classes/template/PKPTemplateManager.inc.php
+++ b/classes/template/PKPTemplateManager.inc.php
@@ -205,12 +205,15 @@ class PKPTemplateManager extends Smarty {
 			}
 
 			// Register recaptcha on relevant pages
-			if (Config::getVar('captcha', 'recaptcha') && Config::getVar('captcha', 'captcha_on_register')) {
+			if (Config::getVar('captcha', 'recaptcha') && (
+					Config::getVar('captcha', 'captcha_on_register') ||
+					Config::getVar('captcha', 'captcha_on_password_reset'
+				))) {
 				$this->addJavaScript(
 					'recaptcha',
 					'https://www.recaptcha.net/recaptcha/api.js?hl=' . substr(AppLocale::getLocale(),0,2),
 					[
-						'contexts' => ['frontend-user-register', 'frontend-user-registerUser'],
+						'contexts' => ['frontend-user-register', 'frontend-user-registerUser', 'frontend-login-lostPassword'],
 					]
 				);
 			}
--- a/pages/login/LoginHandler.inc.php
+++ b/pages/login/LoginHandler.inc.php
@@ -159,6 +159,17 @@ class LoginHandler extends Handler {
 
 		$this->setupTemplate($request);
 		$templateMgr = TemplateManager::getManager($request);
+
+		// Add recaptcha if enabled
+		$captchaEnabled = Config::getVar('captcha', 'captcha_on_password_reset') && Config::getVar('captcha', 'recaptcha');
+		if ($captchaEnabled) {
+			$publicKey = Config::getVar('captcha', 'recaptcha_public_key');
+			$reCaptchaHtml = '<div class="g-recaptcha" data-sitekey="' . $publicKey . '"></div><label for="g-recaptcha-response" style="display:none;">Recaptcha response</label>';
+			$templateMgr->assign([
+				'reCaptchaHtml' => $reCaptchaHtml,
+				'captchaEnabled' => true,
+			]);
+		}
 		$templateMgr->display('frontend/pages/userLostPassword.tpl');
 	}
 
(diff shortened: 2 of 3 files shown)


classes/template/PKPTemplateManager.inc.php
```diff
@@ -205,12 +205,15 @@ function initialize($request) {
...
+			if (Config::getVar('captcha', 'recaptcha') && (
+					Config::getVar('captcha', 'captcha_on_register') ||
+					Config::getVar('captcha', 'captcha_on_password_reset'
+				))) {
 				$this->addJavaScript(
 					'recaptcha',
 					'https://www.recaptcha.net/recaptcha/api.js?hl=' . substr(AppLocale::getLocale(),0,2),
 					[
-						'contexts' => ['frontend-user-register', 'frontend-user-registerUser'],
+						'contexts' => ['frontend-user-register', 'frontend-user-registerUser', 'frontend-login-lostPassword'],
```
jonasraoni, line 216: This should be a conditional include, see this for reference: https://github.com/pkp/pkp-lib/pull/6541/files#diff-8100c829b3c8e2b0f5bbf2d00cbc29f5b67dfa0b7183132bfa8694c7d6e476a7R202-R218

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
 		$this->setupTemplate($request);
 		$templateMgr = TemplateManager::getManager($request);
 
+		// Check if reCaptcha is enabled and validate it if so
+		$captchaEnabled = Config::getVar('captcha', 'captcha_on_password_reset') && Config::getVar('captcha', 'recaptcha');
+		if ($captchaEnabled) {
+			import('lib.pkp.classes.form.Form');
+			$form = new Form('');
```
jonasraoni, line 187:
```suggestion
			$form = new Form();
```

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
...
+				$templateMgr->display('frontend/pages/userLostPassword.tpl');
+				return;
+			}
+		}
+
 		$email = $request->getUserVar('email');
 		$userDao = DAORegistry::getDAO('UserDAO'); /** @var UserDAO $userDao */
 		$user = $userDao->getUserByEmail($email);
 
+		$rateLimitingEnabled = true;
```
jonasraoni, line 214: I think this should be added to a configuration, and perhaps disabled by default to avoid breaking changes.
Then a user can opt-in for both the recaptcha and the rate limiter.

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
...
+		if ($rateLimitingEnabled && $user !== null) {
+			$sessionManager = SessionManager::getManager();
+			$session = $sessionManager->getUserSession();
+
+			// Store reset requests with timestamps in the session
+			$resetRequests = $session->getSessionVar('passwordResetRequests') ?: [];
```
jonasraoni, line 220: Hmm, if we store the retries on the session, then I think it's not very useful, as clearing the session cookie will be enough to skip the check.

I think this piece of code should be removed or moved to another issue, there's a relevant discussion here: https://github.com/pkp/pkp-lib/discussions/11084

templates/frontend/pages/userLostPassword.tpl
```diff
@@ -39,6 +39,14 @@
...
 				</label>
 			</div>
+
+			{if $captchaEnabled}
+			<div class="captcha">
+				<div class="pkp_form_locale_field">
+					{$reCaptchaHtml}
+				</div>
+			</div>
+			{/if}
```
jonasraoni, line 49: Just to follow the pattern of other places:

```suggestion
			{if $reCaptchaHtml}
			<div class="captcha">
				<div class="pkp_form_locale_field">
					{$reCaptchaHtml}
				</div>
			</div>
			{/if}
```

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
...
+			$captchaValidator = new FormValidatorReCaptcha(
+				$form,
+				$request->getRemoteAddr(),
+				'common.captcha.error.invalid-input-response',
+				$request->getServerHost()
+			);
+
+			if (!$captchaValidator->isValid()) {
+				$templateMgr->assign([
+					'captchaEnabled' => true,
```
jonasraoni, line 200:
```suggestion
```

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
...
+
+			// Remove entries older than 1 hour (3600 seconds)
+			foreach ($resetRequests as $resetEmail => $timestamp) {
+				if ($currentTime - $timestamp > 3600) {
+					unset($resetRequests[$resetEmail]);
+				}
+			}
+
+			// Check if this email has a recent request (within the last 10 minutes)
+			if (isset($resetRequests[$email]) && ($currentTime - $resetRequests[$email] < 600)) {
```
jonasraoni, line 231: If the user was blocked, I think it's good to inform that a new request can be made in "X minutes"

pages/login/LoginHandler.inc.php
```diff
@@ -169,12 +180,71 @@ function requestResetPassword($args, $request) {
...
+			$sessionManager = SessionManager::getManager();
+			$session = $sessionManager->getUserSession();
+
+			// Store reset requests with timestamps in the session
+			$resetRequests = $session->getSessionVar('passwordResetRequests') ?: [];
+			$currentTime = time();
+
+			// Remove entries older than 1 hour (3600 seconds)
+			foreach ($resetRequests as $resetEmail => $timestamp) {
+				if ($currentTime - $timestamp > 3600) {
```
jonasraoni, line 225: If the user is allowed to try again after 10 minutes, then I think we can clear the entries also after 10 minutes.

classes/template/PKPTemplateManager.inc.php
Godoy0722: Gotcha. I'll send a new commit with these updates. Thanks @jonasraoni !

pages/login/LoginHandler.inc.php
Godoy0722: Gotcha. I'll remove this piece of code for now.

Godoy0722:
Hello @jonasraoni, just a heads-up that I implemented your change requests, as far as I also removed the rate limiter as you previously mentioned. Please let me know if I have something else to do in this issue and I'll update it. :smile:

Godoy0722:
Hello @asmecher . Sorry for the late response as well! I just committed the requested changes to improve the tab usage consistency. Please let me know if you need anything else from me! And thanks for letting me know that the other versions are ok to merge! :smile:

asmecher:
Thanks, @Godoy0722, that looks good! This will also need to be added to the themes we maintain that override the changed templates, or else they'll stop working if the option is enabled.

Issue #11191, opened 2025-03-27
[Redesigned Dashboard] | Review assignment Views - Reviews Overdue, preivous round overdue reviews are being displayed


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Part of the edge case testing for previous review rounds.  When a round has overdue reviews it would appear that the submission is showing up in the list but the value does not match.  Two issues.

They should match
I don't think it should show up in the view list unless the current round Editorial activity has overdue reviews.

### Steps to Reproduce

1. Preconditions : have a submission in review with reviewers that have overdue reviews.
2. Login with User/Role ' with access to move a submission to another review round
3. Navigate to the submission and select view
4. Select Create New review round
5. Select Reviews Overdue

### Expected Result

Should only show submission with current round overdue reviews in this view

### Actual Result

This view will detect overdue reviews from previous rounds even though the editorial activity is showing no overdue review activity.

![Image](https://github.com/user-attachments/assets/78308d82-4917-4a39-8a6d-9441d0cc78b3)

### Application Version

OJS 3.5 RC2

jardakotesovec:
@Tribunal33 That was great scenario to test.

@Vitaliy-1 Seems as easy fix, can you review: https://github.com/pkp/pkp-lib/pull/11203

jardakotesovec:
@Tribunal33 Merged to 3.5 and main. Should be fine now

Tribunal33:
Still seeing the submissions with overdue reviews from previous rounds being filtered into this Reviews overdue view.

![Image](https://github.com/user-attachments/assets/49dad04a-963e-43e0-9694-edc73f716a85)

jardakotesovec:
@Tribunal33 Oki, so my testing scenario was with review being overdue. While yours is with request for review being overdue. It turned out there is one more bug to address for that scenario.

@Vitaliy-1 I think right fix for this is just to improve condition grouping - https://github.com/pkp/pkp-lib/pull/11215 . Can you please review.

Before the change we had:
```
SELECT "s".*
FROM "submissions" AS "s"
LEFT JOIN "publications" AS "po" ON "s"."current_publication_id" = "po"."publication_id"
WHERE "s"."context_id" IN (?)
  AND "s"."status" IN (?)
  AND "s"."stage_id" IN (?,
                         ?)
  AND "s"."submission_id" IN
    (SELECT "ra"."submission_id"
     FROM "review_assignments" AS "ra"
[...]

PR #11215, opened 2025-04-02
pkp/pkp-lib#11191 Fix sql query condition grouping to work same for r…

jardakotesovec:
…esponse overdue and review overdue


Vitaliy-1 (approved):
Thanks.  I'd use arrow function to shorten the expression. Otherwise looks good

Vitaliy-1 (commented):
Looks good! haven't I reviewed that already? :) It definitely requires wrapping that `or` condition with the additional `where` statement.

Issue #10779, opened 2025-01-07
Redesigned Workflow rcX - Help modal


jardakotesovec:
This has been discussed internally. And for 3.5 we are moving away from built-in help. Instead we will just provide link to "Learning OJS" resources which are being worked on.

@Devika008 Could you please add mock up, where to place the help icon?

# Specs Update - Tuesday, February 25th, 2025

## Workflows Affected by This Change

- [x] The **Help icon** should be displayed in the **software bar** at the top of the OJS interface for easy access at all times.
- [x] Consider only if time allows or if it's an easy fix: **Make the software bar sticky** so it remains visible when scrolling.
- [x] Any reference to the built-in help should be removed. Including the built-in help UI.

## Updates since Thursday, March 27th, 2025
- [x] Create the top navigation as a component to display the top header actions (Help, Tasks, User Dropdown) on the site and side modals.
- [x] Remove **docs/manual** submodule and the tools for checking help `.md` files from the [OJS](https://github.com/pkp/ojs-user-guide)/[OMP](https://github.com/pkp/omp-user-guide)/[OPS](https://github.com/pkp/ops-user-guide) user-guide repo. These will be archived once this issue is complete.

### Below are not related to the Help button, but was necessary to fix for related top navigation changes:
- [x] Prevent the user from clicking the Tasks button (bell icon) again when the Tasks side-modal is already opened.
- [x] Remove unused css styling for old top navbar implementation and use tailwind classes instead
- [x] Set correct Logout As redirect on the User Dropdown when a submission workflow page is opened
- [x] Add logout as button on Participants section
- [x] When user is logged in as another user, make the red circle on the User Dropdown look bigger

HELP URL:
- [x] Change the Help link (#) to the actual user-guide link. ~This was not available, will need to follow up.~ Below are the links to be used for the Help buttons. These are confirmed by Emma:
- [x] ojs: https://docs.pkp.sfu.ca/learning-ojs/
- [x] omp: https://docs.pkp.sfu.ca/learning-omp/en/
- [x] ops: https://docs.pkp.sfu.ca/learning-ops/en/

## Mockup

![Image](https://github.com/user-attachments/assets/c03916a0-b910-49ef-aa81-617ecda6ef77)

--- 
Pull Requests
**stable-3_5_0**:
ui-library: https://github.com/pkp/ui-library/pull/579
pkp-lib: https://github.com/pkp/pkp-lib/pull/11221
[...]

PR #11221, opened 2025-04-02
pkp/pkp-lib#10779 [stable-3_5_0] Redesigned Workflow rcX - Help modal

blesildaramirez:


classes/template/PKPTemplateManager.php
```diff
@@ -1422,10 +1416,26 @@ public function display($template = null, $cache_id = null, $compile_id = null,
...
+                        'username' => $loggedInAsUser->getData('userName'),
+                        'initials' => $loggedInAsUser->getDisplayInitials(),
+                    ];
+                }
+
                 $currentUser = [
                     'csrfToken' => $this->_request->getSession()->token(),
                     'id' => (int) $user->getId(),
                     'roles' => array_values(array_unique($userRoles)),
+                    'unreadTasksCount' => $this->getTemplateVars('unreadNotificationCount'),
```
jardakotesovec, line 1433: Similar concern here.

cypress/support/commands.js
```diff
@@ -873,7 +873,7 @@ Cypress.Commands.add('uploadSubmissionFiles', (files, options) => {
 Cypress.Commands.add('changeLanguage', (language, contextPath) => {
 	contextPath = contextPath || 'publicknowledge';
 
-	cy.get('.app__userNav > button').click();
-	cy.get('.app__userNav a:contains("français")').click();
+	cy.get('[data-cy="app-user-nav"] > button').click();
+	cy.get('[data-cy="app-user-nav"] a:contains("français")').click();
 	cy.wait(2000);
```
jardakotesovec, line 878: right.. I need to delete this file :-D..  keep forgetting..  only commands_new_workflow.js should be used I think  But thats not to worry about in this PR

templates/layouts/backend.tpl
```diff
@@ -74,89 +74,8 @@
 				</div>
 			{/if}
 			{if $currentUser}
```
jardakotesovec, line 76: Always happy to see smarty reduction :-)

jardakotesovec (changes requested):
Looking great, only minor concerns.  And wondering where is the code that was responsible for rendering that help UI?

classes/notification/Notification.php
```diff
@@ -199,4 +199,19 @@ public function scopeWithRead(Builder $query, bool $read)
             fn ($q) => $q->whereNull('date_read')
         );
     }
+
+    public static function getUnreadNotificationsCount(?int $userId): int
```
jonasraoni, line 203: In general it's better to avoid static methods that depend on global state, they are harder to test, but this whole notification stuff is going to be refactored one day, so nothing to do here, just a note.

classes/core/PKPApplication.php
```diff
@@ -469,6 +469,11 @@ abstract public function getContextName(): string;
      */
     abstract public function getVersionDescriptorUrl(): string;
 
+    /**
+     * Get the help URL for this application.
+     */
+    abstract public function getHelpUrl(): string;
```
jonasraoni, line 475: This will probably be a fixed string

```suggestion
    abstract public static function getHelpUrl(): string;
```

jonasraoni (approved):
Just requested minor updates.

jardakotesovec (approved):

PR #11225, opened 2025-04-03
pkp/pkp-lib#10263 Relax editing metadata on published/posted materials

Hafsa-Naeem:
for [issue##10263](https://github.com/pkp/pkp-lib/issues/10263)

### DIFF
--- a/classes/submission/Repository.php
+++ b/classes/submission/Repository.php
@@ -514,6 +514,25 @@ abstract class Repository
      */
     public function canEditPublication(int $submissionId, int $userId): bool
     {
+        // block authors can never edit a published publication even if an editor granted them canChangeMetadata
+        $submission = $this->get($submissionId);
+        if ($submission) {
+            $currentPub = $submission->getCurrentPublication();
+            if (
+                $currentPub
+                && $currentPub->getData('status') === Submission::STATUS_PUBLISHED
+            ) {
+                // fetch this user’s stage assignments
+                $assignments = StageAssignment::withSubmissionIds([$submissionId])->withUserId($userId)->get();
+
+                // if all of their assignments are to an author group then block them
+                $hasNonAuthor = $assignments->contains(fn($sa) => $sa->userGroup && $sa->userGroup->roleId !== Role::ROLE_ID_AUTHOR);
+                if (!$hasNonAuthor) {
+                    return false;
+                }
+            }
+        }
+        
         // Replaces StageAssignmentDAO::getBySubmissionAndUserIdAndStageId
         $stageAssignments = StageAssignment::withSubmissionIds([$submissionId])
             ->withUserId($userId)
--- a/classes/submission/Repository.php
+++ b/classes/submission/Repository.php
@@ -515,36 +515,31 @@ abstract class Repository
     public function canEditPublication(int $submissionId, int $userId): bool
     {
         // block authors can never edit a published publication even if an editor granted them canChangeMetadata
-        $submission = $this->get($submissionId);
-        if ($submission) {
-            $currentPub = $submission->getCurrentPublication();
-            if (
-                $currentPub
-                && $currentPub->getData('status') === Submission::STATUS_PUBLISHED
-            ) {
-                // fetch this user’s stage assignments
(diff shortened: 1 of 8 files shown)


Vitaliy-1 (changes requested):
With these changes, authors with explicit approval from editors can change metadata of the published publication, like title, abstract. 

It would be great to fix the following or open a separate issue: there is an inconsistency with the UI in changing contributors - with the permission from editor, author can, e.g., add a contributor through the API but doesn't have that ability through  the UI. Also I found a bug where author can set a contributor with  any user group ID through the API.

Vitaliy-1 (commented):
Thanks! Two changes that I'd like to be made here:
1) Don't allow authors to access if there is not only published publication, but also scheduled for publication `STATUS_SCHEDULED`
2) Can we also make a correspondent change to the UI if there is publication that is published - for other publications of this submission for the author

classes/submission/Repository.php
```diff
@@ -514,18 +514,34 @@ public function canCurrentUserDelete(Submission $submission): bool
...
 
-        // Check for permission from stage assignments
-        if ($stageAssignments->contains(fn ($stageAssignment) => $stageAssignment->canChangeMetadata)) {
+        $submission = $this->get($submissionId);
+        if ($submission) {
+            foreach ($submission->getData('publications') as $pub) {
+                $status = $pub->getData('status');
+                // if any publication is published or scheduled
+                if (in_array($status, [Submission::STATUS_PUBLISHED, Submission::STATUS_SCHEDULED])) {
+                    $hasNonAuthor = $assignments->contains(fn($sa) =>
```
Vitaliy-1, line 528: Such checks are better to be done outside of the loop, to avoid executing several times per each publication.

Vitaliy-1: Also, we are including `ROLE_ID_ASSISTANT` here but in several places, e.g., preprint grid handler, catalog grid handler, we don't. Is it done deliberately?

asmecher (approved):

Vitaliy-1 (approved):

classes/submission/Repository.php
```diff
@@ -514,18 +514,32 @@ public function canCurrentUserDelete(Submission $submission): bool
...
             ->withUserId($userId)
             ->get();
 
-        // Check for permission from stage assignments
-        if ($stageAssignments->contains(fn ($stageAssignment) => $stageAssignment->canChangeMetadata)) {
+        $submission = $this->get($submissionId);
+        // any published or scheduled then probe
+        $hasLockedPublication = $submission
+            && $submission->getData('publications')
+                ->contains(fn($p) =>
```
jonasraoni, line 526: It's nice to add types here to help the code editor :)

classes/submission/Repository.php
```diff
@@ -514,18 +514,32 @@ public function canCurrentUserDelete(Submission $submission): bool
...
+        $assignments = StageAssignment::withSubmissionIds([$submissionId])
             ->withUserId($userId)
             ->get();
 
-        // Check for permission from stage assignments
-        if ($stageAssignments->contains(fn ($stageAssignment) => $stageAssignment->canChangeMetadata)) {
+        $submission = $this->get($submissionId);
+        // any published or scheduled then probe
+        $hasLockedPublication = $submission
+            && $submission->getData('publications')
```
jonasraoni, line 525: You can use `$submission?->getData()` to avoid the extra check.

classes/submission/Repository.php
```diff
@@ -514,18 +514,32 @@ public function canCurrentUserDelete(Submission $submission): bool
...
+                ->contains(fn($p) =>
+                    in_array(
+                        $p->getData('status'),
+                        [Submission::STATUS_PUBLISHED, Submission::STATUS_SCHEDULED]
+                    )
+                );
+
+        if ($hasLockedPublication && !$assignments->contains(fn($sa) => $sa->userGroup && $sa->userGroup->roleId !== Role::ROLE_ID_AUTHOR)) {
```
jonasraoni, line 533: Better to avoid using `!==` with integer fields, I've explained on the other PR

Issue #11194, opened 2025-03-28
Upgrade from OJS 3.4.0-8 to OJS 3.5.0rc2


shantanu198713:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

PHP-8.3.14
MYSQL-9.1

I attempted to upgrade from **OJS 3.4.0-8** to **OJS 3.5.0rc2**, but the process failed due to the following error. The upgrade could not be completed successfully, and troubleshooting is required to identify and resolve the issue.

### Steps to Reproduce

I attempted to upgrade from OJS 3.4.0-8 to OJS 3.5.0rc2.

### Expected Result

Continue to upgrade.

### Actual Result

Upgradation failed.

### Application Version

OJS-3.5.0rc2

### Logs

```
PHP Fatal error:  Uncaught TypeError: PKP\invitation\core\Invitation::initialize(): Argument #2 ($contextId) must be of type ?int, string given, called in H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\migration\upgrade\v3_5_0\I9197_MigrateAccessKeys.php on line 92 and defined in H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\invitation\core\Invitation.php:112
Stack trace:
#0 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\migration\upgrade\v3_5_0\I9197_MigrateAccessKeys.php(92): PKP\invitation\core\Invitation->initialize(45, 'MailListContext', NULL, NULL)
#1 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\install\Installer.php(444): PKP\migration\upgrade\v3_5_0\I9197_MigrateAccessKeys->up()
#2 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\install\Installer.php(297): PKP\install\Installer->executeAction(Array)
#3 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\install\Installer.php(206): PKP\install\Installer->executeInstaller()
#4 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\classes\install\form\UpgradeForm.php(49): PKP\install\Installer->execute()
#5 H:\wamp64\www\ojs-3.5.0rc2\lib\pkp\pages\install\InstallHandler.php(148): PKP\install\form\UpgradeForm->execute()
[...]

PR #11233, opened 2025-04-04
pkp/pkp-lib#11194 Remove c_v_e_s_pkey unique index from cv entry settings table

touhidurabir:
This PR is a subset of pkp/pkp-lib#11194 which aims to resolve the following error at upgrade , see this comment https://github.com/pkp/pkp-lib/issues/11194#issuecomment-2777486855

Basically after the merge of pkp/pkp-lib#10292, in the `controlled_vocab_entry_settings` table, we have updated the `setting_name` from different symbolic to just `name` which may clash with `c_v_e_s_pkey` unique index which is a combination of `controlled_vocab_entry_id,locale,setting_name` . Also it's not required anymore .

### DIFF
--- a/classes/migration/install/ControlledVocabMigration.php
+++ b/classes/migration/install/ControlledVocabMigration.php
@@ -58,7 +58,6 @@ class ControlledVocabMigration extends \PKP\migration\Migration
             $table->string('locale', 28)->default('');
             $table->string('setting_name', 255);
             $table->mediumText('setting_value')->nullable();
-            $table->unique(['controlled_vocab_entry_id', 'locale', 'setting_name'], 'c_v_e_s_pkey');
         });
 
         // Reviewer Interests Associative Table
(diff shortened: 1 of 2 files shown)


touhidurabir:
@asmecher can you please take a look at this PR ? I got the idea when looking into the issue https://github.com/pkp/pkp-lib/issues/11194#issuecomment-2779587475 and I don't think this PR is the solution of that issue . 

But again, do we need this unique index `c_v_e_s_pkey` in `controlled_vocab_entry_settings` table anymore ? As it was a combination of `controlled_vocab_entry_id,locale,setting_name` where previously the `setting_name` was different symbolic and we update it to just `name` as part of pkp/pkp-lib#10292 . Still with current implementation, the combination of `controlled_vocab_entry_id,locale` should be always unique but was there any chance that in older implementation these 2 can be same which can lead to a duplication for unique index . However I am not able to reproduce it.

asmecher:
@touhidurabir, I think it's worth exploring the problem data to understand it better before we consider dropping the key.

touhidurabir:
closing this in favour of https://github.com/pkp/pkp-lib/pull/11234

Issue #11238, opened 2025-04-08
Prepare DB for structured citations to come


bozana:
DB should be prepared so that the work on structured citation can come later, any time, and would not require a DB upgrade, so that structured citations can be merged without waiting for next major release where DB upgrades are allowed.
Structured citations require following DB changes:
Removing the column `setting_type` in the table `citation_settings`, because citations are moved to the EntityDAO. This column will not be removed right now, but made nullable, so that old code and classes CitationDAO and Citation can still work. After structured citations are merged, this column can be removed in the next major release.
Removing the setting `citationsRaw` from the `publication_settings` table. This will be done now, for 3.5.0. The raw citations for publication object will be constructed by concatenating the each/every single raw citation.
Note: Multilingual citations would require more DB changes, e.g. moving the entries from the column `raw_citation` from the table `citations` into the `citation_settings` table, but for now multiple citations will not be considered, thus these changes will be done once the work on multilingual citations starts/is done.

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/11239
- https://github.com/pkp/ojs/pull/4819
- https://github.com/pkp/omp/pull/1947
- https://github.com/pkp/ops/pull/948
- https://github.com/pkp/googleScholar/pull/28
- https://github.com/pkp/jatsTemplate/pull/85

main:
- https://github.com/pkp/pkp-lib/pull/11278
- https://github.com/pkp/ojs/pull/4835
- https://github.com/pkp/omp/pull/1955
- https://github.com/pkp/ops/pull/957
- https://github.com/pkp/googleScholar/pull/29
- https://github.com/pkp/jatsTemplate/pull/86

https://github.com/pkp/crossrefReferenceLinking/pull/31#
https://github.com/pkp/medra/pull/44

bozana:
Hi @asmecher, these are the minimal DB changes needed for 3.5.0, so that structured citations can come later at any point.
Could you please review the PRs?
Thanks a lot!

bozana:
Hmmm... @asmecher, do you think the hook renaming should also take place now, before 3.5.0:
`CitationDAO::afterImportCitations` -> `Citation::importCitations::after` ?

asmecher:
> Hmmm... @asmecher, do you think the hook renaming should also take place now, before 3.5.0:

[...]

Issue #10629, opened 2024-11-21
Saving empty references throws an error


GaziYucel:
**Describe the bug**
If the References are emptied and saved, an error is shown. Saving a second time does not show the error anymore. The change is however saved to the database. 

**To Reproduce**
1. Activate metadata References: Settings > Workflow > Submission > Metadata > References
2. Go to Submissions and open any submission
3. Open References tab: Publication > References
4. Add any text in the textbox
5. Save
6. Empty the references textbox
7. Save => error: the empty references are saved to the database (citations, publication_settings table rows are deleted)
8. Save again, no more error

**What application are you using?**
 OJS version 3.4.0-5, 3.4.0.7

**Additional information**
The error message shown, is thrown here: https://github.com/pkp/pkp-lib/blob/a1caad182db02667ecebe229fb888c8d23423743/classes/filter/Filter.php#L444

This is the error logged in apache2
```
[php:notice] [pid 19534] [client 127.0.0.1:40484]
[...]

PR #11239, opened 2025-04-08
pkp/pkp-lib#11238 prepare DB for structured citations

bozana:
s. https://github.com/pkp/pkp-lib/issues/11238

and 

https://github.com/pkp/pkp-lib/issues/10629

### DIFF
--- a/classes/publication/DAO.php
+++ b/classes/publication/DAO.php
@@ -25,6 +25,7 @@ use PKP\citation\CitationDAO;
 use PKP\controlledVocab\ControlledVocab;
 use PKP\core\EntityDAO;
 use PKP\core\traits\EntityWithParent;
+use PKP\db\DAORegistry;
 use PKP\services\PKPSchemaService;
 
 /**
@@ -174,6 +175,12 @@ class DAO extends EntityDAO
             ->where('s.submission_id', '=', $publication->getData('submissionId'))
             ->value('locale');
         $publication->setData('locale', $locale);
+
+        $citationDao = DAORegistry::getDAO('CitationDAO'); /** @var CitationDAO $citationDao */
+        $citations = $citationDao->getByPublicationId($publication->getId())->toArray();
+        $citationsRaw = $citationDao->getRawCitationsByPublicationId($publication->getId())->implode(PHP_EOL);
+        $publication->setData('citations', $citations);
+        $publication->setData('citationsRaw', $citationsRaw);
 
         $this->setAuthors($publication);
         $this->setCategories($publication);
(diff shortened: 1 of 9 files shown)


classes/publication/DAO.php
```diff
@@ -175,6 +177,12 @@ public function fromRow(object $row): Publication
             ->value('locale');
         $publication->setData('locale', $locale);
 
+        $citationDao = DAORegistry::getDAO('CitationDAO'); /** @var CitationDAO $citationDao */
```
asmecher, line 180: Hmm, this will potentially be expensive for operations that iterate through a list of publications -- e.g. the submission lists. We're going through the whole citation object lifecycle for every citation attached to each submission, just to get the `citationsRaws`, which in most cases likelihood won't be used.

It wouldn't completely eliminate the overhead, but it might be better to add a function to CitationDAO that just gets the set of citations glued together as a string rather than going through the whole object instantiation part.

Better still, for the few operations that still use `citations` and `citationsRaw`, they could be adapted to use the DAO directly rather than relying on the data being in the DataObject.

asmecher (changes requested):

classes/publication/DAO.php
bozana: Hi @asmecher, hmmm... yes...
But the citationsRaw is everywhere expected to be part of the publication object -- not only in smarty (this is minor), but also in the Form, API and Vue.js... So I would rather leave this as it is for now.
I can try to have a function in CitationDAO that would generate that citationsRaw string.
I can then see with Vitaly and Jarda, if we should change something -- for citations not to be part of the publication object -- when we work on structured citations.
OK?

bozana: Hi @asmecher, hmmm.... another problem/question: if we concatenate it with SQL, the resulting string will be truncated to the `group_concat_max_len`, which is 1024 per default, which seems not to be that much for citations, no?

bozana: Hi @asmecher, one more thing: The `citations` and `citationsRaw` are not part of submission list properties, so these do not influence the performance of it.
However it is maybe good to get raw citations with SQL and then implode them to a string -- instead of going through every object.
And, actually we should always get citation from the publication object -- then they are queried only once here and used everywhere else. Thus, I will now also change those other places that use $citationDao->getByPublicationId() over and over again.

asmecher: @bozana, I think we don't need to concatenate in SQL -- we can still do it in PHP. But it'll be a lot more efficient to just iterate through a result set:
```php
DB::table('citations')->where('publication_id', $publicationId)->implode(...);
```
...rather than going through the whole `Citation` / `CitationDAO` object lifecycle.

asmecher (approved):

Issue #10404, opened 2024-09-12
Allow arbitrary nesting depth in categories


asmecher:
last updated: Sept 2024

Journal Managers can currently create Categories and Subcategories in Settings > Journal > Categories:
https://github.com/user-attachments/assets/42e25d1e-e681-42a5-a3f8-b23930a6caa3

However, this artificially limits the nesting depth to 2 levels. It should be easy to support arbitrary nesting depth. The database already supports this, but the user interface does not.

asmecher:
The UI library aspect of this is filed in https://github.com/pkp/pkp-lib/issues/10449.

bozana:
@mreiko, do you maybe mean @ewhanson ?

mreiko:
Yes! Thanks Bozana.

taslangraham:
**Draft PRs**

pkp-lib: https://github.com/pkp/pkp-lib/pull/11243/files
ui-library: https://github.com/pkp/ui-library/pull/550
ojs: https://github.com/pkp/ojs/pull/4822

taslangraham:
Ready for review @ewhanson 

**PRs**
**pkp-lib**: https://github.com/pkp/pkp-lib/pull/11243/files
- Allow arbitrary levels of categories to be created
- Introduce API endpoints to manage categories
    - The endpoint to fetch categories returns the top-level categories, each with their sub-categories. These sub-categories are recursively populated with their own sub-categories, returning the complete category tree
- Introduce new Form component to create/edit categories
- Added migration to move category images to public directory for each Context so that they're discoverable when using the `FieldUploadImage` field component on the newly introduced form.
- Update category breadcrumb generation to handle arbitrary nesting depth in categories
- Remove old classes related to category management


**ui-library**: https://github.com/pkp/ui-library/pull/550
- Introduce new `CategoryManager.vue` component that uses reka-ui's [tree component](https://reka-ui.com/docs/components/tree) to display categories with arbitrary nesting depth

**ojs**: https://github.com/pkp/ojs/pull/4822
- Added API documentation for new endpoints to swagger schema

**omp**: https://github.com/pkp/omp/pull/1949
ops: https://github.com/pkp/ops/pull/949

**_Note_**: I've temporarily disabled tests related to categories, and will update those on a separate issue.

ewhanson:
[...]

PR #11243, opened 2025-04-08
pkp/pkp-lib#10404 Allow nested categories

taslangraham:
For pkp/pkp-lib#10404


schemas/category.json
```diff
@@ -54,20 +57,79 @@
...
+					"name": {
+						"type": "string"
+					},
+					"editorDisplayInitials": {
+						"type": "string"
+					}
+				}
+			}
+		},
+		"subEditors": {
```
ewhanson, line 129: subEditors should probably a type. I'm assuming an array of integers.

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+use PKP\core\Core;
+use PKP\core\PKPBaseController;
+use PKP\core\PKPRequest;
+use PKP\file\TemporaryFile;
+use PKP\file\TemporaryFileManager;
+use PKP\security\authorization\CanAccessSettingsPolicy;
+use PKP\security\authorization\ContextAccessPolicy;
+use PKP\security\Role;
+
+class PkpCategoryCategoryController extends PKPBaseController
```
jonasraoni, line 38: I'm personally against prefixing classes with PKP, it was probably used in the past as a kind of namespace when namespaces didn't exist.
But let me tag @asmecher to confirm the pattern.

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
+<?php
+
+/**
+ * @file api/v1/categories/PkpCategoryCategoryController.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2003-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getRouteGroupMiddleware(): array
+    {
+        return [
+            'has.user',
+            'has.context',
```
jonasraoni, line 66: Don't we have a non-stringified version of this? I think we should use less hardcoded strings.

```suggestion
            'has.user',
            'has.context',
```

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        $context = $this->getRequest()->getContext();
+        $sortedCategories = $illuminateRequest->all();
+        $categoryIds = array_map(function ($category) {return $category['id'];}, $sortedCategories);
+
+        $categoriesFound = Repo::category()->getCollector()
+            ->filterByContextIds([$context->getId()])
+            ->filterByIds($categoryIds)
+            ->getMany();
+
+        $categoriesFoundIds = $categoriesFound->map(fn ($category) => $category->getId())->all();
```
jonasraoni, line 106: Not sure if your code editor is understanding that `$category` is a `Category`, if not, I'm all for typing at least the arguments.

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+/**
+ * @file classes/components/CategoryManager.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
+<?php
+
+/**
+ * @file classes/components/form/context/CategoryForm.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+
+        if (array_diff($categoryIds, $categoriesFoundIds)) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        // Only allow top-level categories to be ordered
+        foreach ($categoriesFound as $category) {
+            if ($category->getData('parentId') !== null) {
```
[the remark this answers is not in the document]
jonasraoni: If this is something that shouldn't happen in a normal usage, then you can also move this check to the `$categoriesFound` query (there's a `filterByParentIds()`, which accepts `[null]`)

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        $category->setDescription($illuminateRequest->input('description'), null);
+        $category->setParentId($parentId);
+        $category->setPath($illuminateRequest->input('path'));
+        $category->setSortOption($illuminateRequest->input('sortOption'));
+
+        // Update or insert the category object
+        if ($categoryId == null) {
+            $category->setSequence(REALLY_BIG_NUMBER);
+            $categoryId = Repo::category()->add($category);
+            Repo::category()->dao->resequenceCategories($context->getId());
```
jonasraoni, line 203: Instead of using the `REALLY_BIG_NUMBER`, then resequencing (which will run another query/loop), you can just set the sequence with the right value (i.e. `max() + 1`).

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        $subEditors = $illuminateRequest->input('subEditors');
+
+        // Subeditors are assigned to user group that has access to WORKFLOW_STAGE_ID_SUBMISSION stage, but OPS does not have that stage.
+        // So $subEditors will be null in OPS.
+        if ($subEditors !== null) {
+            Repo::category()->updateEditors($categoryId, $subEditors, Category::$ASSIGNABLE_ROLES, $context->getId());
+        }
+
+        $submittedImageData = $illuminateRequest->input('image') ?: [];
+        $temporaryFileId = $submittedImageData ? $submittedImageData['temporaryFileId'] : null;
```
jonasraoni, line 219:
```suggestion
        $temporaryFileId = $submittedImageData['temporaryFileId'] ?? null;
```

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        // Delete the old image if a new one was submitted or if the existing one was removed
+        if ($temporaryFileId || !$submittedImageData) {
+            $oldImageData = $category->getImage();
+            if ($oldImageData) {
+                $publicFileManager = new PublicFileManager();
+                $publicFileManager->removeContextFile($category->getContextId(), $oldImageData['uploadName']);
+                $publicFileManager->removeContextFile($category->getContextId(), $oldImageData['thumbnailName']);
+                $category->setImage(null);
+            }
+        }
```
jonasraoni, line 230: Just to decrease the indentation

```suggestion
        $oldImageData = $category->getImage();
        if (($temporaryFileId || !$submittedImageData) && $oldImageData) {
            $publicFileManager = new PublicFileManager();
            $publicFileManager->removeContextFile($category->getContextId(), $oldImageData['uploadName']);
            $publicFileManager->removeContextFile($category->getContextId(), $oldImageData['thumbnailName']);
            $category->setImage(null);
        }
```

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+     *
+     * @return array - assoc array with thumbnail details
+     * Example:
+     * ```
+     * [
+     *  'thumbnailName' => (string),
+     *  'thumbnailWidth' => (int),
+     *  'thumbnailHeight' => (int),
+     * ]
+     * ```
```
jonasraoni, line 362: I think we can use the array shape syntax (https://phpstan.org/writing-php-code/phpdoc-types#array-shapes)

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+     */
+    private function generateThumbnail(TemporaryFile $temporaryFile, Context $context, $categoryId): array
+    {
+        $temporaryFileManager = new TemporaryFileManager();
+        $imageExtension = $temporaryFileManager->getImageExtension($temporaryFile->getFileType());
+        $sizeArray = getimagesize($temporaryFile->getFilePath());
```
jonasraoni, line 368: What happens if I upload a bad image?

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        $temporaryFilePath = $temporaryFile->getFilePath();
+
+        // Generate the surrogate images. Used later to create thumbnail
+        $image = match ($imageExtension) {
+            '.jpg' => imagecreatefromjpeg($temporaryFilePath),
+            '.png' => imagecreatefrompng($temporaryFilePath),
+            '.gif' => imagecreatefromgif($temporaryFilePath),
+        };
```
jonasraoni, line 376: I'm not sure which image formats we're supporting, but if the user uploads something unexpected, an exception should be raised here. Do we have a validation (e.g. "only jpg/png/gif are allowed")?

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        $coverThumbnailsMaxWidth = $context->getData('coverThumbnailsMaxWidth');
+        $coverThumbnailsMaxHeight = $context->getData('coverThumbnailsMaxHeight');
+        $thumbnailFilename = $categoryId . '-category-thumbnail' . $imageExtension;
+        $xRatio = min(1, ($coverThumbnailsMaxWidth ? $coverThumbnailsMaxWidth : 100) / $sizeArray[0]);
+        $yRatio = min(1, ($coverThumbnailsMaxHeight ? $coverThumbnailsMaxHeight : 100) / $sizeArray[1]);
+        $ratio = min($xRatio, $yRatio);
+        $thumbnailWidth = round($ratio * $sizeArray[0]);
+        $thumbnailHeight = round($ratio * $sizeArray[1]);
+        $thumbnail = imagecreatetruecolor($thumbnailWidth, $thumbnailHeight);
+        imagecopyresampled($thumbnail, $image, 0, 0, 0, 0, $thumbnailWidth, $thumbnailHeight, $sizeArray[0], $sizeArray[1]);
```
jonasraoni, line 387: Perhaps this should be moved to another place, to be re-used. Or perhaps we could include a library to deal with the cropping/resizing (something for another task).

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+        // store the thumbnail
+        $fullPath = $publicFileManager->getContextFilesPath($context->getId()) . '/' . $thumbnailFilename;
+        match ($imageExtension) {
+            '.jpg' => imagejpeg($thumbnail, $fullPath),
+            '.png' => imagepng($thumbnail, $fullPath),
+            '.gif' => imagegif($thumbnail, $fullPath),
+        };
+
+        imagedestroy($thumbnail);
+        imagedestroy($image);
```
jonasraoni, line 400: Such things are nice to be covered with a try/finally

classes/category/Collector.php
```diff
@@ -129,6 +139,10 @@ public function getQueryBuilder(): Builder
             ->leftJoin('categories AS pc', 'c.parent_id', '=', 'pc.category_id')
             ->select(['c.*']);
 
+        $qb->when($this->categoryIds !== null, function ($query) {
+            $query->whereIn('c.category_id', $this->categoryIds);
+        });
```
jonasraoni, line 144:
```suggestion
        $qb->when($this->categoryIds !== null, fn (Builder $query) => $query->whereIn('c.category_id', $this->categoryIds));
```

classes/category/Repository.php
```diff
@@ -63,32 +71,27 @@ public function exists(int $id, ?int $contextId = null): bool
...
-            /** @var ?Category $parent */
-            $parent = $categories->first(
-                fn (Category $c) => $c->getId() === $category->getParentId()
-            );
-            return $this->getBreadcrumb($category, $parent);
+            $currentCategory = $category;
+            $result = $currentCategory->getLocalizedTitle();
+
+            // Traverse up category tree until we reach a top-level category
+            while ($currentCategory->getParentId() && $parent = $categories->get($currentCategory->getParentId())) {
```
jonasraoni, line 84: Such loops can be dangerous, it's better to protect against circular references.

p.s.: I'm not a big fan of building trees with `parentId`, there are better ways :)

classes/category/Repository.php
```diff
@@ -143,6 +147,58 @@ public function validate(?Category $object, array $props, array $allowedLocales,
...
+        }
+
+        if (isset($props['image']) && $props['image']['temporaryFileId']) {
+            $validator->after(function (Validator $validator) use ($props, $context) {
+                $temporaryFileId = $props['image']['temporaryFileId'];
+                $temporaryFileManager = new TemporaryFileManager();
+                $user = Application::get()->getRequest()->getUser();
+                $temporaryFile = $temporaryFileManager->getFile((int)$temporaryFileId, $user->getId());
+                $imageExtension = $temporaryFile ? $temporaryFileManager->getImageExtension($temporaryFile->getFileType()) : [];
+                $isValidExtension = in_array($imageExtension, ['.jpg', '.png', '.gif']);
```
jonasraoni, line 190: If we're using this list of extensions somewhere else, I think it should be turned into a constant (e.g. `SUPPORTED_IMAGE_TYPES`).

classes/category/maps/Schema.php
```diff
@@ -73,15 +77,40 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
         foreach ($props as $prop) {
             switch ($prop) {
+                case 'subCategories':
+                    $subCategories = Repo::category()->getCollector()
+                        ->filterByParentIds([$category->getId()])
+                        ->filterByContextIds([$context->getId()])
+                        ->getMany();
+
+                    if ($subCategories->isNotEmpty()) {
+                        $output['subCategories'] = $this->mapMany($subCategories)->values();
```
jonasraoni, line 90: This is the sad part of such structure, hopefully users won't have huge nested datasets.
Anyway, I think here it's also needed to protect against circular references.

classes/category/maps/Schema.php
```diff
@@ -73,15 +77,40 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
+                        ->filterByRoleIds(Category::$ASSIGNABLE_ROLES)
+                        ->assignedToCategoryIds([$category->getId()])
+                        ->getMany()
+                        ->map(function (User $user) {
+                            return [
+                                'id' => $user->getId(),
+                                'name' => $user->getFullName(),
+                                'editorDisplayInitials' => $user->getDisplayInitials(),
+                            ];
+                        })->values();
```
jonasraoni, line 106:
```suggestion
                        ->map(fn (User $user) => [
                              'id' => $user->getId(),
                              'name' => $user->getFullName(),
                              'editorDisplayInitials' => $user->getDisplayInitials(),
                            ])->values();
```

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,96 @@
...
+use PKP\facades\Repo;
+
+class CategoryManager
+{
+    public const COMPONENT_CATEGORY = 'CategoryManager';
+    public $id = self::COMPONENT_CATEGORY;
+
+    public function __construct(
+        private Context $context,
+    ) {
```
jonasraoni, line 33:
```suggestion
    public function __construct(private Context $context) {
```

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,96 @@
...
+
+    }
```
jonasraoni, line 35:
```suggestion
    public function __construct(private Context $context)
    {
    }
```

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function getConfig(): array
```
jonasraoni, line 37: The methods are missing the phpdoc, if you have patience, it's nice to document the return with the array shape from PHPStan.

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,96 @@
...
+    private function getComponentTableColumns(): array
+    {
+        return [
+            [
+                'name' => 'category Name',
+                'label' => __('grid.category.categoryName'),
+            ],
+            [
+                'label' => __('manager.category.assignedTo'),
+                'name' => 'Assigned To',
```
jonasraoni, line 79:
```suggestion
                'name' => 'Assigned To',
                'label' => __('manager.category.assignedTo'),
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+            ->get()
+            ->map(function (UserGroup $userGroup) use ($request) {
+                return [
+                    'userGroup' => $userGroup,
+                    'users' => Repo::user()
+                        ->getCollector()
+                        ->filterByUserGroupIds([$userGroup->id])
+                        ->filterByContextIds([$request->getContext()->getId()])
+                        ->getMany()
+                        ->mapWithKeys(fn ($user, $key) => [$user->getId() => $user->getFullName()])
```
jonasraoni, line 65:
```suggestion
                        ->mapWithKeys(fn (User $user) => [$user->getId() => $user->getFullName()])
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+            'label' => __('manager.category.form.assignEditors'),
+            'description' => __('manager.categories.form.assignEditors.description'),
+            'id' => 'subEditors',
+        ]);
+        $this->addField(new FieldText('title', [
+            'label' => __('grid.category.name'),
+            'isMultilingual' => true,
+            'isRequired' => true,
+            'groupId' => 'categoryDetails',
+            'value' => $category ? $category->getData('title') : ''
```
jonasraoni, line 92:
```suggestion
            'value' => $category?->getData('title') ?? ''
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+                        Application::get()->getRequest(),
+                        PKPApplication::ROUTE_PAGE,
+                        null,
+                        'catalog',
+                        'category',
+                        ['path']
+                    )]),
+                    'isRequired' => true,
+                    'groupId' => 'categoryDetails',
+                    'value' => $category ? $category?->getData('path') : ''
```
jonasraoni, line 107:
```suggestion
                    'value' => $category?->getData('path') ?? ''
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+                ])
+            )
+            ->addField(
+                new FieldPreparedContent('description', [
+                    'label' => __('grid.category.description'),
+                    'isMultilingual' => true,
+                    'toolbar' => 'bold italic superscript subscript | link | blockquote bullist numlist',
+                    'plugins' => ['link'],
+                    'groupId' => 'categoryDetails',
+                    'value' => $category ? $category->getData('description') : ''
```
jonasraoni, line 117:
```suggestion
                    'value' => $category?->getData('description') : ''
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+            ->addField(
+                new FieldUploadImage('image', [
+                    'label' => __('category.coverImage'),
+                    'baseUrl' => $baseUrl,
+                    'options' => [
+                        'url' => $temporaryFileApiUrl,
+                        'acceptedFiles' => 'image/jpeg,image/png,image/gif,image/jpg',
+                    ],
+                    'groupId' => 'categoryDetails',
+                    'value' => $category ? $category->getData('image') : null
```
jonasraoni, line 137:
```suggestion
                    'value' => $category?->getData('image') ?? null
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+        // Conditionally add this field. This is done because assignable user groups would have access to WORKFLOW_STAGE_ID_SUBMISSION,
+        // but submission stage does not exist in OPS, so we don't display this field in OPS.
+        if (!empty($assignableUserGroups)) {
+            $assignedSubeditors = $category ? Repo::user()
+                ->getCollector()
+                ->filterByContextIds([Application::get()->getRequest()->getContext()->getId()])
+                ->filterByRoleIds($this->assignableRoles)
+                ->assignedToCategoryIds([$category->getId()])
+                ->getIds()
+                ->toArray() : [];
```
jonasraoni, line 150:
```suggestion
            $assignedSubeditors = $category
                ? Repo::user()
                    ->getCollector()
                    ->filterByContextIds([Application::get()->getRequest()->getContext()->getId()])
                    ->filterByRoleIds($this->assignableRoles)
                    ->assignedToCategoryIds([$category->getId()])
                    ->getIds()
                    ->toArray()
                : [];
```

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,191 @@
...
+                $subEditorsDao = DAORegistry::getDAO('SubEditorsDAO');
+                /** @var SubEditorsDAO $subEditorsDao */
+
+                //  A list of user group IDs for each assigned editor, keyed by user ID.
+                $subeditorUserGroups = $subEditorsDao->getAssignedUserGroupIds(
+                    Application::get()->getRequest()->getContext()->getId(),
+                    Application::ASSOC_TYPE_CATEGORY,
+                    $category->getId(),
+                    $assignedSubeditors
+                )->toArray();
```
jonasraoni, line 163:
```suggestion
                )->all();
```

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    /**
+     *
+     * @param Enumerable $updates List of Category image data to update. Keyed by category ID.
+     */
+    private function updateCategoriesImageNameFields(int $contextId, Enumerable $updates): void
+    {
+        $updates->chunk(self::CHUNK_SIZE)->each(function ($chunk) use ($contextId) {
```
jonasraoni, line 67: The `$updates` will be already chunked, so I think it's not needed to chunk again.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,101 @@
...
+                            [$imageDecoded['name'], $imageDecoded['uploadName']] = [$imageDecoded['uploadName'], $imageDecoded['name']];
+                            $imageRecordsToUpdate[$category->category_id] = json_encode($imageDecoded);
+                        }
+                    }
+
+                    $this->updateCategoriesImageNameFields($contextId, collect($imageRecordsToUpdate));
+                });
+
+            // Move images
+            $this->moveContextCategoryImagesToPublicFolder($contextId, $fileNamesToMove);
```
jonasraoni, line 57: I think this should be moved inside the chunked part, to avoid having a large `$fileNamesToMove`.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     *
+     * @param Enumerable $updates List of Category image data to update. Keyed by category ID.
+     */
+    private function updateCategoriesImageNameFields(int $contextId, Enumerable $updates): void
+    {
+        $updates->chunk(self::CHUNK_SIZE)->each(function ($chunk) use ($contextId) {
+            $caseStatement = 'UPDATE categories SET image = CASE category_id ';
```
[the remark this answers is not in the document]
jonasraoni: I see the intention is just to swap values from two JSON properties, so another approach would be a simple string replacement:
`SELECT REPLACE(REPLACE(REPLACE('{"a": "aaa", "b": "bbb"}', '"a"', '"x"'), '"b"', '"a"'), '"x"', '"b"')`

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,101 @@
...
+        });
+    }
+
+
+    /**
+     * Moves the Category images from file system to public dir of given context.
+     */
+    private function moveContextCategoryImagesToPublicFolder(int $contextId, $fileNames): void
+    {
+        $contextFileManager = new ContextFileManager($contextId);
```
jonasraoni, line 85: We're trying to not use classes from the code base on the migrations, otherwise small code updates might end up break expectations at the migrations.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,101 @@
...
+        $basePath = $contextFileManager->getBasePath() . 'categories/';
+        $publicFileManager = new PublicFileManager();
+
+        foreach ($fileNames as $fileName) {
+            $success = $publicFileManager->copyContextFile($contextId, $basePath . $fileName, $fileName);
+            if ($success) {
+                $contextFileManager->deleteByPath($basePath . $fileName);
+            }
```
jonasraoni, line 93: I think it's better to use a move operation.

cypress/tests/integration/Categories.cy.js
```diff
@@ -6,7 +6,7 @@
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  */
 
- describe('Tests categories in the submission wizard', function() {
+ describe.skip('Tests categories in the submission wizard', function() {
```
jonasraoni, line 9: Was the skip forgotten or is the test broken?

jonasraoni (changes requested):
I just got curious to see if we changed the tree implementation and left some comments.

api/v1/categories/PkpCategoryCategoryController.php
taslangraham: There are no non-stringified version for `'has.user'` & `'has.context'`, but I agree it would be good to eventually move away from repeatedly using these hardcoded strings

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+            self::roleAuthorizer([
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SITE_ADMIN,
+            ]),
+        ])->group(function () {
+            Route::get('', $this->getMany(...));
+            Route::get('categoryFormComponent', $this->getCategoryFormComponent(...));
+            Route::post('', $this->add(...));
+            Route::put('saveOrder', $this->saveOrder(...));
+            Route::put('{categoryId}', $this->edit(...))
```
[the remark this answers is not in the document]
taslangraham: Right now, the way category management is set up, it's not really needed — but I’m fine adding the endpoint for completeness.

api/v1/categories/PkpCategoryCategoryController.php
```diff
@@ -0,0 +1,409 @@
...
+                ->whereNumber('categoryId');
+            Route::delete('{categoryId}', $this->delete(...))
+                ->whereNumber('categoryId');
+        });
+    }
+
+    /**
+     * Saves the ordering of categories.
+     */
+    public function saveOrder(Request $illuminateRequest, Context $context): JsonResponse
```
taslangraham, line 95: We've decided to drop the Order functionality from this feature. This method will be removed

taslangraham, line 95: We've decided to drop the Order functionality from this feature. This method will be removed

api/v1/categories/PkpCategoryCategoryController.php
taslangraham: Given that we are removing the ability to order categories, resequencing can be removed as if no longer serves any purpose

api/v1/categories/PkpCategoryCategoryController.php
asmecher: Agreed, the `PKP` class prefix is a deprecated pattern from before there was namespacing. We shouldn't use it for new code (and should gradually remove it as we disturb old code with breaking changes).

classes/category/Category.php
```diff
@@ -14,8 +14,12 @@
 
 namespace PKP\category;
 
+use PKP\security\Role;
+
 class Category extends \PKP\core\DataObject
 {
+    public static array $ASSIGNABLE_ROLES = [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_ASSISTANT];
```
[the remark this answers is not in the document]
taslangraham: no, no particular reason. I'll change to constant

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
taslangraham: The FileManager does not provide such a method

cypress/tests/integration/Categories.cy.js
taslangraham: All test related to categories are skipped, and will be updated on a separate issue

classes/category/Collector.php
```diff
@@ -159,8 +173,6 @@ public function getQueryBuilder(): Builder
             }
         });
 
-        $qb->orderBy(DB::raw('(COALESCE((pc.seq * 8192) + pc.category_id, 0) * 8192) + CASE WHEN pc.category_id IS NULL THEN 8192 * ((c.seq * 8192) + c.category_id) ELSE c.seq END'));
```
taslangraham, line 162: The category ordering feature was removed, so I also dropped the `seq` column. As a result, I’ve removed this orderBy clause as well.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
taslangraham: Understood — but for file management, is there a another pattern in the codebase that can be used within a migration?

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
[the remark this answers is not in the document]
taslangraham: > This all looks pretty straight forward, but have you confirmed these raw SQL statements work with postgres as well?

Yes, works as expected in postgres

api/v1/categories/PkpCategoryCategoryController.php
taslangraham: Thanks, I've dropped the PKP from the class name

api/v1/categories/PkpCategoryCategoryController.php
taslangraham: [Validation](https://github.com/pkp/pkp-lib/blob/a253727e6a9665d3cd749458cf6bc2364a180b77/classes/category/Repository.php#L184-L201) takes place before this method gets called

api/v1/categories/PkpCategoryCategoryController.php
taslangraham: Yea, a validation is done earlier before this function call

classes/category/Collector.php
```diff
@@ -159,7 +175,16 @@ public function getQueryBuilder(): Builder
...
+        $locale = Locale::getLocale();
+        $qb->leftJoin(
+            'category_settings as category_settings',
+            fn (JoinClause $join) => $join->on('category_settings.category_id', '=', 'c.category_id')
+                ->where('category_settings.setting_name', '=', 'title')
+                ->where('category_settings.setting_value', '!=', '')
+                ->where('category_settings.locale', '=', $locale)
+        );
+
+        $qb->orderByRaw('COALESCE(category_settings.setting_value) ASC');
```
taslangraham, line 187: Order categories by title

locale/en/manager.po
```diff
@@ -1439,6 +1439,53 @@ msgstr "Assign {$name} as {$role}"
...
+"</div>"
+"<div class=\"text-secondary\">"
+"<p class=\"mt-6\">This action cannot be undone. Deleting the category will:</p>"
+"<ul role=\"list\" class=\"list-disc ml-8\">"
+"<li>Permanently remove all nested sub-categories</li>"
+"<li>Unassign this category from any submissions currently using it</li>"
+"</ul>"
+"<p class=\"mt-4\">This will not delete the submissions themselves — they will simply be left without a category.</p>"
+"</div>"
+"<p class=\"mt-4 text-primary text-lg-bold\">To confirm, please type the name of the category \"{$title}\" below to proceed</p>"
```
taslangraham, line 1467: I decided to add the styling and formatting here to avoid having to split these across multiple locale entries then styling them individually on the frontend.

taslangraham:
@ewhanson @jonasraoni thanks for the review. I've made updates based on the feedback given.

@jonasraoni In regard to your comments about circular references -- after discussing with @ewhanson we've decided that it is not a major concern at this point, thus, we'll keep the current implementation as is. However, taking your concerns into consideration, I've added checks to prevent a category from being updated to have circular references for parent-child relationships

classes/components/CategoryManager.php
```diff
@@ -0,0 +1,116 @@
...
+use APP\file\PublicFileManager;
+use Illuminate\Support\Enumerable;
+use PKP\category\Category;
+use PKP\components\forms\context\CategoryForm;
+use PKP\context\Context;
+use PKP\facades\Repo;
+
+class CategoryManager
+{
+    public const COMPONENT_CATEGORY = 'CategoryManager';
```
jardakotesovec, line 29: For new UI we don't create PHP component classes anymore. There were too many limitations for more complex UI, so we are basically moving extensibility capabilities to client side - which is covered in https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-plugins--docs

Only thing where we still rely on the PHP side is creating the forms. We have not had enough time to bring that capability to JS. All other interactions and data exchanges ideally would happen just via API.

So you can remove this class, and just directly pass the CategoryForm config in the handler

jardakotesovec (changes requested):
Feel free to reach out, if any of the feedback is not clear, thanks!

jonasraoni:
@taslangraham thanks, I'll take a look. Considering we're not using transactions, N users updating the same dataset can lead to curious situations, so I think it's important (unless locks/transactions were added in this PR).

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
jonasraoni: I just remember about seeing one migration with file operations at `lib/pkp/classes/migration/upgrade/v3_4_0/I8073_RemoveNotesWithoutQueriesAndRelatedObjects.php`

It's still using the `Config` class, but it sounds less likely to be broken than the `ContextFileManager`. I can confirm that one of these file classes have a fixed path, which will need to be updated.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
jonasraoni: I think it's ok to create it! But such update will conflict with another comment that I left on the other migration (I10404_UpdateCategoryImageNameFields.php)

api/v1/categories/CategoryCategoryController.php
```diff
@@ -0,0 +1,364 @@
...
+        $category->setPath($illuminateRequest->input('path'));
+        $category->setSortOption($illuminateRequest->input('sortOption'));
+
+        // Prevent categories from referencing themselves as their parent
+        if ($parentId && $categoryId && $$categoryId === $parentId) {
+            return response()->json(__('api.categories.400.invalidParentId'), Response::HTTP_BAD_REQUEST);
+        }
+
+        // Prevent category from being updated to have a circular parent reference
+        if ($parent && $category->getParentId() === $parent->getId() && $parent->getParentId() === $category->getId()) {
```
jonasraoni, line 162: Given that you're setting `$category->setParentId($parentId);` some lines above., then `$category->getParentId() === $parent->getId()` will be always `true` and could be removed.

jonasraoni: And as I said before, while we're doing our integrity checks here, another user might be changing the database and breaking the checks which you validated.

This `parentId` structure is bad, but I don't think it should be changed in this PR.

Besides fixing the code (visit the parents until you reach the top, and ensure a node isn't visited twice), I think it's needed to lock the involved items for update (the current item + parents).

ewhanson: Thanks @jonasraoni. I think for this PR, adding the circular reference checks you suggest is a good idea, but as you mention the rest we should wait until this PR has been merged. If you have an idea for performing the circular references, could you share your suggested approach here?

For the other things you've mentioned, could you create a separate issue as a sub-issue of #11307 outlining some of the concerns you've mentioned here as well as the solutions you have in mind. Definitely seems like a good idea but it would be helpful to have this code solidified and merged first. Thanks again for taking such a thorough look!

jonasraoni: Something like this should be enough:

```php
public function hasCircularReference(Category $category, ?Context $context): bool
{
    for ($visited = []; $category?->parentId !== null;) {
        // An item cannot be visited twice
        if (isset($visited[$category->id])) {
            return true;
        }

        $visited[$category->id] = true;
        $category = Repo::category()->get($category->parentId, $context?->getId());
    }

    return false;
}
```

jonasraoni: I've created the sub-issue here: #11360

jonasraoni: For the sake of safety (even an intentionally broken database should be safe), a similar protection should be added to the code, on the places I've highlighted before.

To avoid duplicating code, a generic method that retrieves the parents, while doing the validation (you can update the method I've shared above to store the parents and retrieve them if there's no circular reference) sounds very okayish.

taslangraham:
@jonasraoni @ewhanson I've made updates based on last review . Please take a look
 
**specific commits for main changes**
        - Remove usage of FileManager in migrations https://github.com/pkp/pkp-lib/pull/11243/commits/9383896beec87c61a59baa6a2f93e6fbda12487b  
        - Implement circular reference checks and validations https://github.com/pkp/pkp-lib/pull/11243/commits/ec5a18a20622eacd343a8837931bb98152124199

pages/management/ManagementHandler.php
```diff
@@ -586,19 +604,17 @@ public function manageEmails(array $args, Request $request): void
...
-     * @return void
+     *
      * @throws \Exception
      */
     public function editUser(array $args, $request): void
     {
         $this->setupTemplate($request);
         $userId = $args[0];
-        if(!empty($userId)) {
+        if (!empty($userId)) {
```
jonasraoni, line 614: In general I think it's better to invert the condition (the `handle404()` will quit the application), then we'll have a smaller indentation and be able to drop the `else`.

pages/management/ManagementHandler.php
```diff
@@ -165,11 +167,27 @@ public function context($args, $request)
...
+
+        $categoriesApiUrl = Application::get()->getRequest()->getDispatcher()->url(
+            Application::get()->getRequest(),
+            Application::ROUTE_API,
+            $context->getPath(),
+            'categories'
+        );
+        $categoryForm = new CategoryForm($categoriesApiUrl, $locales, $baseUrl, $temporaryFileApiUrl);
+
+
```
jonasraoni, line 182:
```suggestion
```

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,159 @@
...
+            return;
+        }
+
+        $caseStatement = 'UPDATE categories SET image = CASE category_id ';
+
+        foreach ($updates as $categoryId => $json) {
+            $caseStatement .= "WHEN {$categoryId} THEN ? ";
+        }
+
+        $caseStatement .= 'END WHERE category_id IN (' . implode(',', $updates->keys()->all()) . ') AND context_id = ?';
```
jonasraoni, line 87: It's possible to drop these `CASEs` with a string replace, I think I've mentioned it before, then you'll solve this issue with a single query, which will cover all contexts.

```sql
UPDATE categories SET image = REPLACE(REPLACE(REPLACE(image, '"name":', '"swapper":'), '"uploadName":', '"name":'), '"swapper":', '"uploadName":')
```

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,159 @@
...
+                'dir' => [
+                    'public' => self::DIRECTORY_MODE_MASK & ~$umask,
+                    'private' => self::DIRECTORY_MODE_MASK & ~$umask,
+                ]
+            ]),
+            LOCK_EX,
+            LocalFilesystemAdapter::DISALLOW_LINKS
+        );
+
+        $fileSystem = new Filesystem($adapter);
```
jonasraoni, line 120: I think this can be moved to a property and initialized in the constructor.

jonasraoni (changes requested):
I just reviewed again, I requested minor updates and a bigger one on the migration.

classes/category/Collector.php
```diff
@@ -99,6 +102,15 @@ public function filterByPaths(?array $paths): self
         return $this;
     }
 
+    /**
+     * Filter categories by IDs
+     */
+    public function filterByIds(?array $categoryIds): self
```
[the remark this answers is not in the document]
taslangraham: Interested to know what are the advantages/difference in using `static` here?

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,169 @@
...
+     * @param string $action URL to submit the form to.
+     * @param array $locales Supported locales.
+     * @param string $baseUrl Site's base URL. Used for image previews.
+     * @param string $temporaryFileApiUrl URL to upload files to.
+     */
+    public function __construct(string $action, array $locales, $baseUrl, $temporaryFileApiUrl)
+    {
+        $this->action = $action;
+        $this->method = 'POST';
+        $this->locales = $locales;
```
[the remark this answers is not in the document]
taslangraham: Conflict with signature of parent class' constructor -- error:`Type must not be defined (as in base class '\PKP\components\forms\FormComponent')`. Updating the parent constructor meant I also needed to update quite a lot of other classes due to type errors. Maybe a type of cleanup that can be done on a separate issue

classes/components/forms/context/CategoryForm.php
```diff
@@ -0,0 +1,169 @@
...
+use PKP\components\forms\FieldUploadImage;
+use PKP\components\forms\FormComponent;
+use PKP\core\PKPApplication;
+use PKP\user\User;
+use PKP\userGroup\UserGroup;
+
+class CategoryForm extends FormComponent
+{
+    public const FORM_CATEGORY = 'editCategory';
+    public $id = self::FORM_CATEGORY;
```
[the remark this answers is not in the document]
taslangraham: Suggestion not added for reasons similar to this other comment https://github.com/pkp/pkp-lib/pull/11243#discussion_r2080811436

taslangraham:
All updated @jonasraoni. Thanks again for the review!

classes/category/Category.php
```diff
@@ -48,22 +53,6 @@ public function setParentId(?int $parentId)
...
-    }
-
-    /**
-     * Set sequence of category.
-     */
-    public function setSequence(float $sequence)
-    {
-        return $this->setData('sequence', $sequence);
-    }
-
```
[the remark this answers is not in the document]
taslangraham: I've confirmed with Erik that these are fine to remove given that the actual database columns they are interacting with have been remove as well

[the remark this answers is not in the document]
jonasraoni: Yes, but such changes to the "public API" of classes can break customizations/plugins of users, so in theory the best would be to leave a dummy deprecated method on 3.6, then wipe out the methods on 3.7.

As we don't have a rock solid architecture yet, I think it's ok to break contracts, there's much more ahead.

Tagging @asmecher just to get some general directions.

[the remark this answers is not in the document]
asmecher: Since the feature is going to be removed, I'd rather an integration that expected it died with a fatal error rather than having some functionality stop working for an unclear reason. This will make it really clear for 3rd party developers when they port their plugins forward across the change.

classes/category/Category.php
```diff
@@ -14,8 +14,13 @@
 
 namespace PKP\category;
 
+use PKP\security\Role;
+
 class Category extends \PKP\core\DataObject
 {
+    public const ASSIGNABLE_ROLES = [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_ASSISTANT];
+    public const SUPPORTED_IMAGE_TYPES = ['.jpg', '.png', '.gif'];
+    public static string $PATH_REGEX = '/^[a-zA-Z0-9\/._-]+$/';
```
jonasraoni, line 23:
```suggestion
    public const $PATH_REGEX = '/^[a-zA-Z0-9\/._-]+$/';
```

classes/category/Category.php
```diff
@@ -14,8 +14,13 @@
 
 namespace PKP\category;
 
+use PKP\security\Role;
+
 class Category extends \PKP\core\DataObject
 {
+    public const ASSIGNABLE_ROLES = [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_ASSISTANT];
```
jonasraoni, line 21: I didn't review the visibility modifiers of anything in this PR, but in general I prefer to use private > protected > public, to decrease the surface of the public API and decrease breaking changes in the future. And if turning things into `private` makes it tough to customize/re-use stuff, then we probably have structural/organization issues (e.g. classes doing too much things).

Does everything that's marked as public should be public? If not, then it's better to hide the implementation.

classes/category/Repository.php
```diff
@@ -143,6 +156,58 @@ public function validate(?Category $object, array $props, array $allowedLocales,
...
+                if ($context->getId() !== $props['contextId']) {
+                    $validator->errors()->add('contextId', __('api.categories.400.invalidContext'));
+                }
+            });
+        }
+
+        if (isset($props['path'])) {
+            $validator->after(function (Validator $validator) use ($props, $context, $object) {
+                // Check if path matches the allowed pattern
+                if (!preg_match(\Category::$PATH_REGEX, $props['path'] ?: '')) {
```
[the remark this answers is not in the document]
jonasraoni: Also, if this `PATH_REGEX` is only used here, you probably can move it to this class.

classes/migration/upgrade/v3_6_0/I10404_UpdateCategoryImageNameFields.php
```diff
@@ -0,0 +1,143 @@
...
+                    ->whereIn('category_id', $ids)
+                    ->update([
+                        'image' => DB::raw(
+                            'REPLACE(
+                                REPLACE(REPLACE(image, \'"name":\', \'"temporaryNameField":\'), \'"uploadName":\', \'"name":\'),
+                                \'"temporaryNameField":\', \'"uploadName":\'
+                            )'
+                        ),
+                    ]);
+            });
```
jonasraoni, line 80: The inner query can do the whole work, but doing "pauses" in such queries is a healthy approach, to avoid hitting timeouts/hardware limits.

But instead of retrieving all rows, we can just get the last ID of the chunk and iterate from there... Something like this pseudo-code:

```
$lastId = 0;
do {
  $nextId = `SELECT id FROM table WHERE id > $lastId LIMIT $chunkSize ORDER BY id DESC LIMIT 1`;
  $affectedRows = `UPDATE table WHERE id BETWEEN $lastId AND $nextId`;
  $lastId = $nextId;
} while ($affectedRows);
```

locale/en/submission.po
```diff
@@ -2189,6 +2189,14 @@ msgstr "Before you begin"
 msgid "submission.wizard.categories.description"
 msgstr "Select only the categories that are appropriate for your submission."
 
+msgid "submission.wizard.categories.descriptionWithCircularReferenceWarning"
+msgstr ""
+"Select only the categories that are appropriate for your submission.<br>"
+"<sub class=\"text-negative\">Some categories are not displayed due to circular references. Please contact your Admin to resolve this.</sub>"
```
ewhanson, line 2195: For this and the one below, I would add some additional context, something like "due to circular references in the category hierarchy." I would also say "Site Administrator" rather than "Admin" as this info will be listed on the site with contact info.

ewhanson (approved):
Just had one final comment on the circular references warning, but otherwise looks good!

Issue #11219, opened 2025-04-02
[Submission Workflow] | Internal Review Round - Status, Revisions requested status does not change even when author has uploaded a revision


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Part of an ongoing investigation with understanding how the internal and external review rounds work.  This was found during the internal review round.  An author will upload a file but the status doesn't update for author or editor views.

Preconditions: Have a submission in the internal review round stage.  Can be seen with submission number 9 from default dataset. 

### Steps to Reproduce

1. Login with an Editor (Rvaca) 
2. Select the Request Revisions button
3. Complete the Request Revisions flow
a. Editoral activity will be "Revisions requested from author"
b. Round 1 status will be "Revisions have been requested"
4. Select the three dots on the author that made the submission
5. Select Login As on that Author (Fernando Perini)
6. Editoral activity will be "Revisions requested" + Submit Revisions button
7. Select either "Upload", "Upload Revisions" or "Submit revisions"
8. Complete the upload process
a. Select "This is not a revision of an existing file"
b. Select Bibliogrphy
c. Choose a file (example is PDF) and select continue
9. On the second stage select Continue
10. On 3rd stage Select Complete

### Expected Result

For the author view the status message should change when a revision files is uploaded, The editorial activity message should change after a revision file is uploaded.  For the editor view the editorial activity message should change and the Round 1 status should change.

### Actual Result

The message and editorial activity does not change even when a file has been uploaded into the revisions uploaded table area.

![Image](https://github.com/user-attachments/assets/141b54b7-94e3-49bf-8926-769c72e33336)

### Application Version

OMP 3.3, 3.4, 3.5

jardakotesovec:
PRs (more to come)
pkp-lib 3.3: https://github.com/pkp/pkp-lib/pull/11247

Vitaliy-1:
Thanks, @jardakotesovec. This can be merged!

jardakotesovec:
Great, merged.
@Tribunal33 Ready for you, its updated in 3.3, 3.4 and 3.5

Tribunal33:
@jardakotesovec I test on 3.3, 3.4, 3.5 OMP VMs and uploading does indeed change the status now.  However there is one problem which I am not sure how to resolve.  It is part of my ongoing investigation into the differences between all of these versions.  

[...]

PR #11248, opened 2025-04-09
pkp/pkp-lib#11219 Fix: Detect correctly that the revision was uploade…

jardakotesovec:
…d in internal review round


Vitaliy-1 (approved):
Looks good!

Issue #11212, opened 2025-03-31
[Submissions] Published article - Download statistics display is ambiguous


almereyda:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Download statistics sometimes show empty bars, which often leads people to assume no downloads exist for a certain period.

### Steps to Reproduce

1. Go to an article page.
2. See the Downloads visualisation.

### Expected Result

The downloads are presented in an unambiguous form.

Common praxis is to zero the y-axis at zero, or to find a suitable lower value than the lowest value in the distribution.

### Actual Result

The y-axis is zeroed on the lowest value instead of zero or another option. It is easy for readers to not see this.

### Application Version

3.3.0-20

#### Downloads visualisation in OJS

> ![Downloads visualisation in OJS](https://github.com/user-attachments/assets/c4865d2e-47e8-428d-a43a-68490148ca9a)

Taken from [3D Printing of Continuous-Fibers Cementitious Composites: Anisotropic 3D Mortar | Open Conference Proceedings](https://www.tib-op.org/ojs/index.php/ocp/article/view/193).

#### Downloads visualisation with Vega Lite

> ![Downloads visualisation with Vega Lite](https://github.com/user-attachments/assets/c7bd2b31-ab7c-469b-bfa7-1e0c2fe8473b)

<details><summary>Vega Lite Manifest used for generation of comparative display</summary>
<p>

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {
    "url": "https://corsproxy.io/?url=https://path/to/pkp_usage_downloads.csv"
  },
  "mark": "bar",
  "transform": [
    {
      "filter": {
        "and": [
          "datum.Month >= 202404",
          "test(/^3D Printing of/, datum.Submission)"
        ]
      }
    }
  ],
  "encoding": {
    "x": {"field": "Month", "type": "ordinal"},
    "y": {"field": "Count", "type": "quantitative"},
    "color": {"field": "Submission"}
  }
}

```

</p>
</details> 

PRs:
stable-3_3_0:
- https://github.com/pkp/usageStats/pull/59
- https://github.com/pkp/ui-library/pull/597
- https://github.com/pkp/ojs/pull/4836 (only submodules updates)
- https://github.com/pkp/omp/pull/1956 (only submodules updates)
- https://github.com/pkp/ops/pull/958 (only submodules updates)

stable-3_4_0:
[...]

PR #11256, opened 2025-04-10
pkp/pkp-lib#11212 chart.js y-axis beginAtZero

bozana:
s. https://github.com/pkp/pkp-lib/issues/11212

### DIFF
--- a/js/usage-stats-chart.js
+++ b/js/usage-stats-chart.js
@@ -89,6 +89,9 @@
 					color: 'rgba(0,0,0,0.05)',
 					drawTicks: false,
 				},
+				ticks: {
+					beginAtZero: true
+				},
 			},
 		},
 	};


jardakotesovec (approved):

js/usage-stats-chart.js
```diff
@@ -89,6 +89,9 @@
 					color: 'rgba(0,0,0,0.05)',
 					drawTicks: false,
 				},
+				ticks: {
+					beginAtZero: true
```
blesildaramirez, line 93: Probably something that we need to add in the https://github.com/pkp/ui-library/blob/main/src/components/Chart/LineChart.vue as well? Not sure.

blesildaramirez (approved):
Looks good, just that maybe we need to add the "beginAtZero" param in the LineChart for ui-library.

js/usage-stats-chart.js
blesildaramirez: Yes please, you can merge it after applying the fix. Thanks @bozana !

bozana:
I will merge and add the submodule updates in another PR that I need to merge soon...

PR #11266, opened 2025-04-11
refactor invitation page handler

ipula:
https://github.com/pkp/pkp-lib/issues/11118


defstat (approved):
Thanks @ipula - I think the code concerns are properly separated now.

PR #11270, opened 2025-04-14
structured citations

GaziYucel:
Issue https://github.com/pkp/pkp-lib/issues/10692


api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        if (!Repo::citation()->exists((int)$illuminateRequest->route('citationId'))) {
+            return response()->json([
+                'error' => __('api.citations.404.citationNotFound')
+            ], Response::HTTP_OK);
+        }
+
+        $citation = Repo::citation()->get((int)$illuminateRequest->route('citationId'));
```
bozana: Maybe to have this line first, and then check 
if (!$citation) {...
This way we save a DB query, I think.

api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+            }
+        }
+
+        Hook::call('API::citations::params', [$collector, $illuminateRequest]);
+
+        $citations = $collector->getMany();
+
+        return response()->json([
+            'itemsMax' => $collector->getCount(),
+            'items' => Repo::citation()->getSchemaMap()->summarizeMany($citations->values())->values(),
```
bozana: I think only $citations could be provided to summarizeMany()?

api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+        if (!$citation) {
+            return response()->json([
+                'error' => __('api.citations.404.citationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_CITATION, $illuminateRequest->input());
+        $params['id'] = $citation->getId();
+
+        $errors = Repo::citation()->validate(null, $params);
```
bozana: Because the citation exist, the citation should be given as first argument for the validate() function

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+     *
+     * @return array A key/value array with validation errors. Empty if no errors
+     *
+     * @hook Citation::validate [[&$errors, $citation, $props]]
+     */
+    public function validate(?Citation $citation, array $props): array
+    {
+        $errors = [];
+
+        Hook::call('Citation::validate', [&$errors, $citation, $props]);
```
bozana, line 126: I think something like this needs to be added before this line:
```
        $validator = ValidatorFactory::make(
            $props,
            $this->schemaService->getValidationRules($this->dao->schema, [])
        );

        // Check required fields 
        ValidatorFactory::required(
            $validator,
            $citation,
            $this->schemaService->getRequiredProps($this->dao->schema),
            $this->schemaService->getMultilingualProps($this->dao->schema),
            [],
            ''
        );

        $errors = [];

        if ($validator->fails()) {
            $errors = $this->schemaService->formatValidationErrors($validator->errors());
        }
```

api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+        $citation = Repo::citation()->get((int)$illuminateRequest->route('citationId'));
+
+        if (!$citation) {
+            return response()->json([
+                'error' => __('api.citations.404.citationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_CITATION, $illuminateRequest->input());
+        $params['id'] = $citation->getId();
```
bozana, line 181: I am wondering if it would work without this line, I believe it should.
I think something like this would be needed not to allow the editing of the readOnly properties, but readOnly properties are considered on the DB update in https://github.com/pkp/pkp-lib/blob/main/classes/core/EntityDAO.php#L194.
I am wondering why was this needed -- I see it is also in other API endpoints... :thinking: 
I can test it... but to note it for now...

locale/en/submission.po
```diff
@@ -469,7 +469,174 @@ msgid "submission.citations"
...
+
+msgid "submission.citations.structured.processed.description"
+msgstr "All references have been processed and added below. You can review, edit or remove them at any time."
+
+msgid "submission.citations.structured.deleteDialog.title"
+msgstr "Delete this structured citation (title)"
+
+msgid "submission.citations.structured.deleteDialog.confirm"
+msgstr "This will remove this citation (message)."
+
```
jardakotesovec, line 547: that (title) and (message) I don't think thats intended?

api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+    {
+        $citation = Repo::citation()->get((int)$illuminateRequest->route('citationId'));
+
+        if (!$citation) {
+            return response()->json([
+                'error' => __('api.citations.404.citationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_CITATION, $illuminateRequest->input());
```
bozana, line 174: After this line, there should be something like:
 ```
$readOnlyErrors = $this->getWriteDisabledErrors(PKPSchemaService::SCHEMA_SUBMISSION, $params);
        if (!empty($readOnlyErrors)) {
            return response()->json($readOnlyErrors, Response::HTTP_BAD_REQUEST);
        }
```
S. as example https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/PKPSubmissionController.php#L653, and the function getWriteDisabledErrors() implemented there.

File: classes/citation/job/MetadataLookupHandler.php [hunk 8]
```diff
@@ -0,0 +1,68 @@
+<?php
+
+/**
+ * @file classes/citation/job/externalServices/MetadataLookup.php
```
kaitlinnewson:
```suggestion
 * @file classes/citation/job/MetadataLookup.php
```

classes/citation/job/MetadataLookupHandler.php
```diff
@@ -0,0 +1,68 @@
...
+    public function __construct(int $citationId)
+    {
+        $this->citationId = $citationId;
+        $this->execute();
+    }
+
+    public function execute(): void
+    {
+        $citation = Repo::citation()->get($this->citationId);
+        if($citation->getData('isProcessed')){
```
kaitlinnewson:
```suggestion
        if ($citation->getData('isProcessed')) {
```

classes/citation/job/MetadataLookupHandler.php
```diff
@@ -0,0 +1,68 @@
...
+
+        if (
+            $extractPidsStatus &&
+            $crossrefStatus &&
+            $openAlexStatus &&
+            $orcidStatus
+        ) {
+            $citation->setData('isProcessed', true);
+            Repo::citation()->edit($citation, []);
+        }
```
kaitlinnewson:
```suggestion
        } else {
```

classes/citation/job/MetadataLookupHandler.php
```diff
@@ -0,0 +1,68 @@
...
+        else{
```
kaitlinnewson:
```suggestion

```

classes/citation/job/externalServices/ExternalServicesHelper.php
```diff
@@ -0,0 +1,89 @@
...
+     * Makes HTTP request to the API and returns the response as an array.
+     *
+     * @param string $method The HTTP method (e.g., 'POST', 'GET').
+     * @param string $url The API endpoint URL.
+     * @param array $options Additional options for the HTTP request.
+     * @return array The response data as an associative array.
+     */
+    public static function apiRequest(string $method, string $url, array $options): array
+    {
+        if ($method !== 'POST' && $method !== 'GET') return [];
```
kaitlinnewson:
```suggestion
        if ($method !== 'POST' && $method !== 'GET') {
            return [];
        }
```

classes/citation/job/externalServices/ExternalServicesHelper.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $httpClient = Application::get()->getHttpClient();
+//        $httpClient = new \GuzzleHttp\Client(
```
kaitlinnewson: Possibly some commented out code to be removed here?

classes/citation/job/externalServices/ExternalServicesHelper.php
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            $response = $httpClient->request($method, $url, $options);
+
+            if (!str_contains('200,201,202', (string)$response->getStatusCode())) {
+                return [];
+            }
+
+            $result = json_decode($response->getBody(), true);
+
+            if (empty($result) || json_last_error() !== JSON_ERROR_NONE) return [];
```
kaitlinnewson:
```suggestion
        if (empty($result) || json_last_error() !== JSON_ERROR_NONE) {
            return [];
        }
```

classes/citation/job/externalServices/openAlex/Inbound.php
```diff
@@ -0,0 +1,131 @@
...
+            'GET',
+            $this->url . '/works/doi:' . $this->citation->getData('doi'),
+            []
+        );
+
+        if (empty($response)) {
+            return false;
+        }
+
+        // error_log(json_encode($response, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE));
```
kaitlinnewson: To be removed?

classes/citation/job/pid/Arxiv.php
```diff
@@ -0,0 +1,49 @@
...
+        $string = parent::extractFromString($string);
+
+        $class = get_called_class();
+
+        // check if prefix found in extracted string
+        $prefixes = $class::prefixInCorrect;
+        $prefixes[] = $class::prefix;
+
+        foreach($prefixes as $prefix){
+            if(str_contains($string, $prefix)) return $string;
```
kaitlinnewson:
```suggestion
        if (str_contains($string, $prefix)) {
            return $string;
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+    public const string defaultTrimCharacters = ' ./';
+
+    /**
+     * Add prefix
+     * @param string|null $string e.g. 10.123/tib123
+     * @return string e.g. https://doi.org10.123/tib123
+     */
+    public static function addPrefix(?string $string): string
+    {
+        if (empty($string)) return '';
```
kaitlinnewson:
```suggestion
        if (empty($string)) {
            return '';
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+
+        /* @var BasePid $class */
+        $class = get_called_class();
+
+        // no prefix defined, return original string
+        if (empty($class::prefix)) return $string;
```
kaitlinnewson:
```suggestion
        if (empty($class::prefix)) {
            return $string;
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Remove prefix
+     * @param string|null $string e.g. https://doi.org10.123/tib123
+     * @return string e.g. 10.123/tib123
+     */
+    public static function removePrefix(?string $string): string
+    {
+        if (empty($string)) return '';
```
kaitlinnewson:
```suggestion
        if (empty($string)) {
            return '';
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+
+        /* @var BasePid $class */
+        $class = get_called_class();
+
+        // no prefix defined, return original string
+        if (empty($class::prefix)) return $string;
```
kaitlinnewson:
```suggestion
        if (empty($class::prefix)) {
            return $string;
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+     */
+    public static function normalize(?string $string): string
+    {
+        if (empty($string)) return '';
+
+        /* @var BasePid $class */
+        $class = get_called_class();
+
+        // no prefix defined, return original string
+        if (empty($class::prefix)) return $string;
```
kaitlinnewson:
```suggestion
        if (empty($class::prefix)) {
            return $string;
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Normalize PID by removing any incorrect prefixes.
+     * @param string|null $string e.g. doi:10.123/tib123
+     * @return string e.g. https://doi.org/10.123/tib123
+     */
+    public static function normalize(?string $string): string
+    {
+        if (empty($string)) return '';
```
kaitlinnewson:
```suggestion
        if (empty($string)) {
            return '';
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Extract from string with regex
+     * @param string|null $string
+     * @return string e.g. 10.123/tib123
+     */
+    public static function extractFromString(?string $string): string
+    {
+        if (empty($string)) return '';
```
kaitlinnewson:
```suggestion
        if (empty($string)) {
            return '';
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+
+        /* @var BasePid $class */
+        $class = get_called_class();
+
+        // no regex defined, return empty
+        if (empty($class::regex)) return '';
```
kaitlinnewson:
```suggestion
        if (empty($class::regex)) {
            return '';
        }
```

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $matches = [];
+
+        preg_match($class::regex, $string, $matches);
+
+        if (empty($matches[0])) return '';
```
kaitlinnewson:
```suggestion
        if (empty($matches[0])) {
            return '';
        }
```

classes/citation/job/pid/Handle.php
```diff
@@ -0,0 +1,49 @@
...
+        $string = parent::extractFromString($string);
+
+        $class = get_called_class();
+
+        // check if prefix found in extracted string
+        $prefixes = $class::prefixInCorrect;
+        $prefixes[] = $class::prefix;
+
+        foreach($prefixes as $prefix){
+            if(str_contains($string, $prefix)) return $string;
```
kaitlinnewson:
```suggestion
        if (str_contains($string, $prefix)) {
            return $string;
        }
```

api/v1/citations/PKPCitationController.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    public function delete(Request $illuminateRequest): JsonResponse
+    {
+        if (!Repo::citation()->exists((int)$illuminateRequest->route('citationId'))) {
+            return response()->json([
+                'error' => __('api.citations.404.citationNotFound')
+            ], Response::HTTP_OK);
+        }
+
+        $citation = Repo::citation()->get((int)$illuminateRequest->route('citationId'));
```
bozana, line 222: Put this line in front

classes/publication/DAO.php
```diff
@@ -175,16 +161,10 @@ public function fromRow(object $row): Publication
...
-        $publication->setData('citationsRaw', new class($publication->getId()) implements \Stringable {
-            public function __construct(public int $publicationId) {}
-            function __toString() {
-                $citationDao = DAORegistry::getDAO('CitationDAO'); /** @var CitationDAO $citationDao */
-                return $citationDao->getRawCitationsByPublicationId($this->publicationId)->implode(PHP_EOL);
-            }
-        });
+        $publication->setData('citations',
+            Repo::citation()->getByPublicationId($publication->getId()));
+        $publication->setData('citationsRaw',
```
bozana: This we should leave as Alec changed it receantly (s. the removed code on right)

classes/publication/DAO.php
```diff
@@ -108,10 +93,11 @@ public function getIds(Collector $query): Collection
      */
     public function getMany(Collector $query): LazyCollection
     {
-        return LazyCollection::make(function () use ($query) {
-            $rows = $query
-                ->getQueryBuilder()
-                ->get();
+        $rows = $query
```
bozana: This we should leave as Alec improved it recently, s. the reorganized code on the right

classes/citation/job/MetadataLookupHandler.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+namespace PKP\citation\job;
+
+use APP\facades\Repo;
+use PKP\citation\job\externalServices\crossref\Inbound as CrossrefInbound;
+use PKP\citation\job\externalServices\openAlex\Inbound as OpenAlexInbound;
+use PKP\citation\job\externalServices\orcid\Inbound as OrcidInbound;
+use PKP\citation\job\pid\ExtractPidsHelper;
+use PKP\jobs\citation\MetadataLookup;
```
kaitlinnewson:
```suggestion

```

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2227,4 +2224,81 @@ protected function copyMultilingualData(Submission $submission, string $newLocal
                 }
             });
     }
+
+    /**
+     * Update citationsMetadataLookup of a publication.
+     */
+    protected function editCitationsMetadataLookup(Request $illuminateRequest): JsonResponse
```
bozana, line 2232: Is this API endpoint needed? Is it not enough to edit the publication with that setting?
I need to see the new UI for this... but just a note for now...

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2227,4 +2224,81 @@ protected function copyMultilingualData(Submission $submission, string $newLocal
...
+        if (!$publication) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $publicationId = $publication->getId();
+
+        $existingCitations = [];
+        foreach (Repo::citation()->getByPublicationId($publicationId) as $citation) {
```
bozana: Can't we use $publication->getData('citations') instead of getting them again?

classes/publication/Repository.php
```diff
@@ -360,10 +361,10 @@ public function version(Publication $publication): int
             }
         }
 
-        if (!empty((string) $newPublication->getData('citationsRaw'))) {
-            $citationDao = DAORegistry::getDAO('CitationDAO'); /** @var \PKP\citation\CitationDAO $citationDao */
-            $citationDao->importCitations($newPublication->getId(), $newPublication->getData('citationsRaw'));
-        }
+        Repo::citation()->importCitations(
+            $newPublication->getId(),
+            $newpublication->getData('citationsRaw')
```
kaitlinnewson:
```suggestion
            $newPublication->getData('citationsRaw')
```

classes/citation/job/externalServices/ExternalServicesHelper.php
```diff
@@ -0,0 +1,89 @@
...
+
+            return $result;
+
+        } catch (GuzzleException|Exception $e) {
```
kaitlinnewson:
```suggestion
        } catch (GuzzleException | Exception $e) {
```

classes/citation/job/pid/Arxiv.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @brief Arxiv class
+ */
+
+namespace PKP\citation\job\pid;
+
+class Arxiv extends BasePid
+{
+    /** @copydoc AbstractPid::regex */
+    public const string regex = '%\b(([\w-]+://?|www[.])[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|/)))%s';
```
kaitlinnewson: Since 3.5.0-x uses a minimum PHP version of 8.2, typed constants should not be used as they're only allowed since 8.3.

classes/citation/Citation.php
```diff
@@ -20,99 +16,102 @@
...
     {
         return $this->getData('rawCitation');
     }
 
     /**
-     * Set the rawCitation
+     * Set the rawCitation.
      */
-    public function setRawCitation(?string $rawCitation)
+    public function setRawCitation(?string $rawCitation): void
```
bozana: I think now rawCitation cannot be null any more here -- earlier `?string` was needed because one could create a Citation using the constructor that expects rawCitation, and if it is not given then the null would be given to setRawCitation.

classes/citation/Citation.php
```diff
@@ -20,99 +16,102 @@
...
     {
-        $rawCitation = $this->_cleanCitationString($rawCitation ?? '');
+        $rawCitation = $this->cleanCitationString($rawCitation ?? '');
```
bozana: Then also here and below for function description: there is always rawCitation parameter for cleanCitationString

bozana: if preg_replace from cleanCitationString returns null, then there will be no rawCitation -- to double check other places in the code getting the rawCitation
but maybe better to change the returning value from cleanCitationString()

classes/citation/Citation.php
```diff
@@ -20,99 +16,102 @@
...
+
     /**
      * Take a citation string and clean/normalize it
      */
-    public function _cleanCitationString(string $citationString) : string
+    public function cleanCitationString(?string $citationString = null): string|null
     {
-        // 1) Strip slashes and whitespace
         $citationString = trim(stripslashes($citationString));
+        return preg_replace('/[\s]+/u', ' ', $citationString);
```
bozana: Maybe to return citationString as it is if preg_replace returns null

classes/citation/Citation.php
```diff
@@ -20,99 +16,102 @@
...
+    }
 
-        // 2) Normalize whitespace
-        $citationString = preg_replace('/[\s]+/u', ' ', $citationString);
+    /**
+     * Determine if citation is structured.
+     */
+    public function calculateIsStructured(): bool
```
bozana: maybe to call it isStructured()

classes/citation/Citation.php
```diff
@@ -20,99 +16,102 @@
...
+    {
+        foreach ($this->requirementIsStructured as $set) {
+            $isStructured = false;
+
+            foreach ($set as $item) {
+                if (!empty($this->getData($item))) {
+                    $isStructured = true;
+                    break;
```
bozana, line 105: I think you can r`eturn true` directly here

classes/citation/DAO.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Get a collection of citations matching the configured query
+     *
+     * @return LazyCollection<int,T>
+     */
+    public function getMany(Collector $query): LazyCollection
+    {
+        $rows = $query
```
bozana: Put this into 
`return LazyCollection::make(function () use ($rows) {`
As Alec change it everywhere recently.
S. other getMany()

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * Get all citations for a given publication.
+     *
+     * @return array<Citation>
+     */
+    public function getByPublicationId(int $publicationId): array
+    {
+        return $this->getCollector()
+            ->filterByPublicationId($publicationId)
+            ->orderBy(Collector::ORDERBY_SEQUENCE)
```
bozana: Because it is default sequencing no need to call it extra. It seems there is also no need for that constant at the moment -- it is all always ordered by sequence

classes/core/PKPApplication.php
```diff
@@ -173,6 +173,7 @@ class_alias('\PKP\payment\QueuedPayment', '\QueuedPayment'); // QueuedPayment in
         }
 
         // Ensure that nobody registers for hooks that are no longer supported
+        Hook::addUnsupportedHooks('CitationDAO::afterImportCitations'); // pkp/pkp-lib#10692 Removed with introduction of structured citations
```
bozana: This hook does not exist any more, we removed it during the DB preparation for structured citations, s. https://github.com/pkp/pkp-lib/issues/11238

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+
+        // Tokenize raw citations
+        $citationTokenizer = new CitationListTokenizerFilter();
+        $citationStrings = $rawCitationList ? $citationTokenizer->execute($rawCitationList) : [];
+
+        $existingRawCitations = collect($this->getByPublicationId($publicationId))
+            ->map(fn(Citation $citation) => $citation->getData('rawCitation'))
+            ->all();
+
+        if ($existingRawCitations !== $citationStrings) {
```
bozana, line 223: Are the raw citations here the same? -- Because in $citation->setRawCitation we use cleanCitationString(). Is this fully the same what we get from the filter?

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+    {
+        $existingCitations = $this->getByPublicationId($publicationId);
+
+        Hook::call('Citation::importCitations::before', [$publicationId, $existingCitations, $rawCitationList]);
+
+        // Tokenize raw citations
+        $citationTokenizer = new CitationListTokenizerFilter();
+        $citationStrings = $rawCitationList ? $citationTokenizer->execute($rawCitationList) : [];
+
+        $existingRawCitations = collect($this->getByPublicationId($publicationId))
```
bozana: There is a function getRawCitationsByPublicationId in this class

jobs/citation/MetadataLookup.php
```diff
@@ -0,0 +1,42 @@
...
+    public function __construct(int $citationId)
+    {
+        parent::__construct();
+
+        $this->citationId = $citationId;
+    }
+
+    public function handle(): void
+    {
+        if (!$this->citationId) {
```
bozana: How could this happen?

classes/citation/job/MetadataLookupHandler.php
bozana: It could also be that the citation is deleted in the meantime

classes/citation/job/MetadataLookupHandler.php
```diff
@@ -0,0 +1,68 @@
...
+            $extractPidsStatus &&
+            $crossrefStatus &&
+            $openAlexStatus &&
+            $orcidStatus
+        ) {
+            $citation->setData('isProcessed', true);
+            Repo::citation()->edit($citation, []);
+        }
+        else{
+            Repo::citation()->addJobForCitation($this->citationId, 300);
```
bozana: When is this the case?
And why the delay?

jobs/citation/MetadataLookup.php
```diff
@@ -0,0 +1,42 @@
...
+            throw new JobException(JobException::INVALID_PAYLOAD);
+        }
+
+        new MetadataLookupHandler($this->citationId);
```
bozana: Why do we need an extra class? Couldn't the code be in this job class?

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+     * Add a new job for metadata lookup for a given publication.
+     */
+    public function addJobForCitation(int $citationId, int $jobDelay = 60): void
+    {
+        $citation = $this->get($citationId);
+
+        if ($citation->getData('isProcessed') || !$citation->getData('rawCitation')) {
+            return;
+        } else {
+            dispatch(new MetadataLookup($citation->getId()))->delay(now()->addSeconds($jobDelay));
```
bozana: Why do we need the delay?

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+
+        foreach ($citations as $citation) {
+            $this->addJobForCitation($citation->getId());
+        }
+    }
+
+    /**
+     * Add a new job for metadata lookup for a given publication.
+     */
+    public function addJobForCitation(int $citationId, int $jobDelay = 60): void
```
bozana: Maybe we do not need this function, the job can be directly dispatched. It is only called when importing a citation, that is not processed in that moment.

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+    /** @copydoc DAO::existsRawCitation() */
+    public function existsRawCitation(int $publicationId, string $rawCitation): bool
+    {
+        return $this->dao->existsRawCitation($publicationId, $rawCitation);
+    }
+
+    /**
+     * Add a new job for metadata lookup for a given publication.
+     */
+    public function addJobForPublication(int $publicationId): void
```
bozana: Is this function used?

classes/citation/Repository.php
```diff
@@ -0,0 +1,288 @@
...
+        $existingSeq = collect($this->getByPublicationId($publicationId))
+            ->map(fn(Citation $citation) => $citation->getData('seq'))
+            ->all();
+        $lastSeq = end($existingSeq) ? end($existingSeq) + 1 : 1;
+
+        $rejectedCitations = [];
+        if (is_array($citationStrings) && !empty($citationStrings)) {
+            foreach ($citationStrings as $rawCitationString) {
+                if (!empty(trim($rawCitationString))) {
+                    if (!$this->existsRawCitation($publicationId, $rawCitationString)) {
```
bozana, line 265: here again, if the string is the same because we call also cleanCitationString() in setRawCitation()

api/v1/submissions/PKPSubmissionController.php
GaziYucel: I created this because I find it dangerous to have the whole publication updated from the UI for a single field. But other than that, it would be of course possible to have the update done by updating the publication. 
I could maybe make it more generic to have a single field updated.

jobs/citation/MetadataLookup.php
GaziYucel: It should actually be this: 
```
$citation = Repo::citation()->get($this-citationId); 
if(!$citation) {...}
```
for the case that maybe the citation was deleted in the meantime.

classes/citation/Repository.php
GaziYucel: Yes, they are the same. There was however an error, which I just corrected. I was comparing Collection and Array, which is not correct. Also added a for each to avoid another DB round trip.

classes/citation/Repository.php
GaziYucel: Removed this method.

classes/citation/Repository.php
GaziYucel: Without the delay, the script waits for the job. With the delay, it adds and continues.

classes/citation/job/externalServices/ExternalServicesHelper.php
GaziYucel: Hi Kaitlin, this keeps going back to "GuzzleException|Exception" in my PhpStorm. Do you have an extra setting which creates the extra spaces around the pipe? In my knowledge, there should be no spaces around a pipe in this context.

GaziYucel:
Issues

- [ ] references added through submission wizard aren't being added for metadata lookup
- [ ] default setting metadata lookup for new submissions under Settings > Workflow > Submission > Metadata > References not implemented yet
- [ ] add action for every citation to add job for metadata lookup (currently under consideration)
- [ ] ...

Notes

- [ ] enabling metadata lookup adds jobs for metadata lookup for all current saved citations
- [ ] triggering metadata lookup for all existing citations is possible from the UI
- [ ] ...

schemas/citation.json
```diff
@@ -0,0 +1,260 @@
+{
+	"title": "Citation",
+	"description": "Citation is a reference in a publication to another publication.",
+	"required": [
+		"publicationId",
+		"seq"
+	],
+	"properties": {
```
bozana: The properties should be sorted alphabetically... so that one can find them better...

classes/citation/Repository.php
bozana: Hmmm... From the code it does not look like they are fully the same: 
cleanCitationString additionally strips the slashes and removes multiple empty spaces from a citation string.
Maybe to move the code from cleanCitationString into the CitationListTokenizerFilter::process() -- so everything is done in the same, one place?

classes/citation/Repository.php
```diff
@@ -0,0 +1,310 @@
...
+            $existingRawCitations[] = $citation->getRawCitation();
+        }
+
+        if ($existingRawCitations !== $citationStrings) {
+            $importedCitations = [];
+            $this->deleteByPublicationId($publicationId);
+            if (is_array($citationStrings) && !empty($citationStrings)) {
+                $publication = Repo::publication()->get($publicationId);
+                foreach ($citationStrings as $seq => $rawCitationString) {
+                    if (!empty(trim($rawCitationString))) {
```
bozana: This trim should have already be done in the filter, no?

classes/citation/Repository.php
```diff
@@ -0,0 +1,310 @@
...
+     * Import citations from a raw citation list of the particular publication to existing citations.
+     */
+    public function importAdditionalCitations(int $publicationId, ?string $rawCitationList): string
+    {
+        $citationTokenizer = new CitationListTokenizerFilter();
+        $citationStrings = $rawCitationList ? $citationTokenizer->execute($rawCitationList) : [];
+
+        $existingSeq = collect($this->getByPublicationId($publicationId))
+            ->map(fn(Citation $citation) => $citation->getData('seq'))
+            ->all();
```
bozana: Maybe it would be good to have a function getLastSeq($publicationId) ? -- that would get the max(seq) -- so that we not need this map function and go through all the collection.

classes/citation/Repository.php
```diff
@@ -0,0 +1,310 @@
...
+        $lastSeq = end($existingSeq) ? end($existingSeq) + 1 : 1;
+
+        $rejectedCitations = [];
+        if (is_array($citationStrings) && !empty($citationStrings)) {
+            $publication = Repo::publication()->get($publicationId);
+            foreach ($citationStrings as $rawCitationString) {
+                if (!empty(trim($rawCitationString))) {
```
bozana: again, isn't it done in the filter?

classes/citation/job/externalServices/ExternalServicesHelper.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Makes HTTP request to the API and returns the response as an array.
+     *
+     * @param string $url The API endpoint URL.
+     * @return array|int|null The response as an associative array, request status code or null.
+     */
+    public static function apiRequest(string $url, array $options = []): array|int|null
+    {
+        $httpClient = Application::get()->getHttpClient($options);
```
bozana: getHttpClient should not get options parameter

classes/citation/job/pid/ExtractPidsHelper.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Exception;
+use PKP\citation\Citation;
+
+class ExtractPidsHelper
+{
+    public function execute(Citation $citation): ?Citation
+    {
+        try {
+            $rowRaw = $citation->cleanCitationString($citation->getRawCitation());
```
bozana: This cleanCitationString is not needed, because rawCitation was cleaned when it was set, no?

classes/citation/job/pid/ExtractPidsHelper.php
```diff
@@ -0,0 +1,62 @@
...
+
+            // extract doi
+            $doi = Doi::extractFromString($rowRaw);
+            if (!empty($doi)) {
+                $citation->setData('doi', $doi);
+            }
+
+            // remove doi from raw
+            $rowRaw = str_replace(Doi::addPrefix($doi), '', Doi::normalize($rowRaw));
```
bozana: Why do we need this?

classes/citation/job/pid/BasePid.php
```diff
@@ -0,0 +1,166 @@
...
+        $prefixAlt = str_replace('https://', '', $class::prefix);
+
+        // make secure
+        $string = str_replace('http://', 'https://', $string);
+
+        // process longer first, e.g. dx.doi.org before doi.org
+        usort($prefixInCorrect, function ($a, $b) {
+            return strlen($b) - strlen($a);
+        });
+        $string = str_replace($prefixInCorrect, $prefixAlt, $string);
```
bozana: Should this maybe be case insensitive? One could maybe also use ORCID, ORCID ID, DOI..., no?

classes/citation/job/pid/ExtractPidsHelper.php
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            // urn
+            $urn = Urn::extractFromString($rowRaw);
+            if (!empty($urn)) {
+                $citation->setData('urn', Urn::extractFromString($rowRaw));
+            }
+
+            return $citation;
+        } catch (Exception) {
```
bozana: Why is this try and catch ? -- Does any of the function in try trows an exception?

jobs/citation/ExtractPidsJob.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        if($citation->getIsProcessed()) {
+            return;
+        }
+
+        $extractPids = new ExtractPidsHelper();
+        $citationChanged = $extractPids->execute($citation);
+
+        if (empty($citationChanged)) {
```
bozana: Maybe to check if is null -- to be more precise?

jobs/citation/ExtractPidsJob.php
```diff
@@ -0,0 +1,58 @@
...
+            throw new JobException(__('common.error'));
```
bozana: In this case the job should fail, correct?

jobs/citation/OrcidJob.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $authors = $citation->getData('authors');
+        if (empty($authors)) {
+            return;
+        }
+
+        $publication = Repo::publication()->get($citation->getData('publicationId'));
+        $submission = Repo::submission()->get($publication->getData('submissionId'));
+        $context = Application::getContextDAO()->getById($submission->getData('contextId'));
+        $contactEmail = $context->getContactEmail();
```
bozana: maybe to provide contactEmail to each job, when the job is dispatched -- so that we do not need to get all the objects here...

jobs/citation/OrcidJob.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $service = new Inbound($contactEmail);
+        $authorsChanged = [];
+        foreach ($authors as $author) {
+            if (empty($author['orcid'])) {
+                $authorsChanged[] = $author;
+                continue;
+            }
+
+            $authorChanged = $service->getAuthor($author);
```
bozana, line 68: getAuthor gets an array, but $author is Author, no?

classes/citation/Repository.php
```diff
@@ -0,0 +1,310 @@
...
+     * Add a new job chain for a citation.
+     */
+    public function reprocessCitation(int $citationId): void
+    {
+        $jobs = [
+            new ExtractPidsJob($citationId),
+            new CrossrefJob($citationId),
+            new OpenAlexJob($citationId),
+            new OrcidJob($citationId),
+            new IsProcessedJob($citationId)
```
bozana: Maybe no need for this job -- it only sets isProcessed for the citation. This can be done at the end of OrcidJob, I think.

classes/citation/job/pid/ExtractPidsHelper.php
GaziYucel: Doi::normalize does some cleaning/restructuring so that a doi results in https://doi.org/10.1234/123 (e.g original doi:10.123/123). This is than removed from the rawCitation, so that the url, arxiv, handle regex don't catch these.

classes/citation/job/pid/ExtractPidsHelper.php
GaziYucel: No, no exceptions. I did this to get a similar class as the other classes. But there is no situation that it will throw an exception. Thus makes no sense to check for it. I will remove this. Also the check for null in ExtractPidsJob will be removed.

jobs/citation/ExtractPidsJob.php
GaziYucel: No need, will be removed. See comment at ExtractPidsHelper.

jobs/citation/ExtractPidsJob.php
GaziYucel: Yes, but removed. Won't happen.

jobs/citation/OrcidJob.php
GaziYucel: This is not the OJS Author object! This $author is an array as defined under schemas/citation.json > property authors.

```
"authors": {
    "type": "array",
    "description": "A list of the authors for this work.",
    "multilingual": false,
    "apiSummary": true,
    "validation": [
        "nullable"
    ],
    "items": {
        "type": "object",
        "properties": {
            "givenName": {
                "type": "string"
            },
            "familyName": {
                "type": "string"
            },
            "orcid": {
                "type": "string"
            },
            "openAlex": {
                "type": "string"
            }
        }
    }
},
```

classes/citation/Repository.php
GaziYucel: I would argue that this is cleaner this way. But if you prefer to have it at the end of Orcid than that's also possible.

classes/citation/job/pid/ExtractPidsHelper.php
GaziYucel: Shall move this to the job? Than this class can be removed.

classes/citation/job/externalServices/ExternalServicesHelper.php
kaitlinnewson: Hi Gazi, have you set up PHP CodeSniffer in PHPStorm (in Settings > PHP > Quality Tools > PHP_CodeSniffer) and set it to use PSR12 as the coding standard? I have that setup in mine which will flag this and other formatting issues.

schemas/submission.json
```diff
@@ -47,6 +47,14 @@
 			"apiSummary": true,
 			"writeDisabledInApi": true
 		},
+		"contextCitationsMetadataLookup": {
```
bozana, line 50: Do we need this for submission?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2227,4 +2225,89 @@ protected function copyMultilingualData(Submission $submission, string $newLocal
...
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $citationsMetadataLookup = (bool)$illuminateRequest->input('citationsMetadataLookup');
+
+        /** @var Citation $citation */
+        foreach ($publication->getData('citations') as &$citation) {
+            if ($citationsMetadataLookup && !$publication->getData('citationsMetadataLookup')) {
```
bozana: This should go before the foreach loop. And here we do not check the context setting which we should, no?

schemas/submission.json
GaziYucel: I need this in the UI for the initial decision for the checkbox. This was the only sensible place I could think of to provide this value. Or am I missing another place where I can access the context values? This value is saved in the context object and can be found in pkp-lib/schemas/context.json, the second checkbox in the screenshot below. (@jardakotesovec)

api/v1/submissions/PKPSubmissionController.php
GaziYucel: I don't think we need the context value here. The context value is only used if there is no value in publication yet, so is_null. In any other situation, publication::citationsMetadataLookup is leading. No?

GaziYucel: I moved the if logic before the foreach loop. That's indeed better for reading.

schemas/submission.json
GaziYucel: references

api/v1/submissions/PKPSubmissionController.php
bozana: How is this journal setting meant to be? Does it define if metadata lookup should be possible, or only suggests the default value for each publication (but actually the editor can decide different for each publication)? 
If it is the first case, we do need to always consider that setting -- further metadata lookup should not be possible if JM disables it, correct?
Lets discuss it in chat...

classes/citation/Repository.php
bozana: using this as an example: 
`Barone, Tom\' & Eisner, Elliot W. (2012). Arts based research. London: Sage.`

I will fix it...

classes/citation/Repository.php
bozana: I have removed it in favor of less processing

schemas/submission.json
bozana: I think similarly to `contextMinReviewsPerSubmission` we can provide this setting to the Vue page. I will try to test it.

api/v1/submissions/PKPSubmissionController.php
bozana: The journal setting is just to define the default state for new submissions, not to enable/disable the structured citations fully.

schemas/submission.json
bozana: Worked thanks to Jarda's help :-)

Issue #11244, opened 2025-04-09
[Review Assignments] | Respond to request - Title, missing a space after the colon


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Not sure if you can do a global search and check for other areas. If so please list them.  Noticed a few areas where the title is missing a space.  This one is when logged in as a reviewer and responding to a request to review.  

I think it's coming from here
`lib\pkp\pages\reviewer\PKPReviewerHandler.php`
and it's the pageTitle var here : 
`'pageTitle' => __('semicolon', ['label' => __('submission.review')]) . $reviewSubmission->getCurrentPublication()->getLocalizedTitle()`

### Steps to Reproduce

1. Login with User/Role Reviewer with a request for a review
2. Select "Respond to request"

### Expected Result

Title should have a space between "Review: {submission name}"

### Actual Result

Space is missing
![Image](https://github.com/user-attachments/assets/4d5a446c-5306-4a81-a60d-88434bec0879)

### Application Version

OJS, OMP, OPS 3.5.0rc2

#### pkp-lib 

- pkp-lib main: https://github.com/pkp/pkp-lib/pull/11280
- pkp-lib 3.5: https://github.com/pkp/pkp-lib/pull/11279
- pkp-lib 3.4: https://github.com/pkp/pkp-lib/pull/11281
- pkp-lib 3.3: https://github.com/pkp/pkp-lib/pull/11282

#### submodule updates

- OJS 3.3: https://github.com/pkp/ojs/pull/4838
- OJS 3.4: https://github.com/pkp/ojs/pull/4837
- OPS 3.3: https://github.com/pkp/ops/pull/959
- OPS 3.4: https://github.com/pkp/ops/pull/960
- OMP 3.3: https://github.com/pkp/omp/pull/1958
- OMP 3.4: https://github.com/pkp/omp/pull/1957

kaitlinnewson:
Hi @bozana, can you review the small PRs listed above? I just updated submodules for 3.3/3.4 but can do the others if needed. There are also unrelated test failures for 3.4 which I've filed [in a separate issue](https://github.com/pkp/pkp-lib/issues/11283). Thanks!

bozana:
Hi @kaitlinnewson, the submodule updates need to be re-done for stable-3_4_0

kaitlinnewson:
Hi @bozana, the 3.4 submodule updates have now been updated.

bozana:
[...]

PR #11279, opened 2025-04-15
pkp/pkp-lib#11244 add whitespace to submission review page title

kaitlinnewson:


bozana:
I see some other places also uses this locale key, so I hope those places stay OK with this change -- I could not see a problem quickly...

bozana:
Also, maybe we can mark the key with `#, fuzzy`, for other languages to update as well, if needed...

PR #11282, opened 2025-04-15
pkp/pkp-lib#11244 add whitespace to submission review page title

kaitlinnewson:


pages/reviewer/PKPReviewerHandler.inc.php
```diff
@@ -42,7 +42,7 @@ function submission($args, $request) {
 		if ($step > $reviewStep) $step = $reviewStep; // Reviewer can't go past incomplete steps
 		if ($step < 1 || $step > 4) throw new Exception('Invalid step!');
 		$templateMgr->assign([
-			'pageTitle' => __('semicolon', ['label' => __('submission.review')]) . $reviewerSubmission->getLocalizedTitle(),
+			'pageTitle' => __('semicolon', ['label' => __('submission.review')]) . ' ' . $reviewerSubmission->getLocalizedTitle(),
```
bozana, line 45: I think this should be solved differently, always having a locale key where this concatenation happens and never in the code, because the other languages could have it fully differently...
But maybe to double check with @asmecher :-)

kaitlinnewson: Perhaps it would make sense to apply the proper fix to 3.5/main since those have active translation, then apply this fix for 3.3/3.4 (since they are already done this way)?

kaitlinnewson: Hi @bozana, I've updated the 3.5/main PRs, ready for another look.

Issue #11252, opened 2025-04-09
[Submission List] | Reviewer list - Dashboard, Duplicate reviews are populated.


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Noticed that previous review round review request are being populated in the reviewer submission view.  I'm unsure if it would matter at what stage the review was in so I will only do the overdue request from round 1 and a new review request in round 2.  I felt like this would be a common use case as an editor wanting to start a new round with fresh reviewers.

Precondition: Start with a submission in the review stage with at least one reviewer added.  You can create a new submission and work it through but it isn't necessary and won't reflect from where the steps will start from.

### Steps to Reproduce

1. Login with User/Role with editor permissions
2. Navigate to a submission in the reivew round and has at least one reviewer added
3. Select the three dots menu and edit the review request of the reviewer
4. Edit the response due date to before the current date so that it will be overdue
5. Select OK to save
6. Select Create New Review Round
7. Select Add Reviewer after new round is created
8. Select the same reviewer from the previous round with overdue status
9. Select the three dots and Login As this reviewer
10. Observe the submission dashboard of this reviewer

### Expected Result

There are two solutions that I would be comfortable with.
1. Only show current review round review requests in the submission list.  Make sure to keep the access to previous review round decisions
2. Update the editorial message with something like "This review request has moved to another round" and remove the Respond to Request button

This is the current way to see previous review round details as outlined in #9453
![Image](https://github.com/user-attachments/assets/0acc9a42-640b-426e-b7d4-dffce673ef7a)

### Actual Result

There are two reviews to respond to.  Only the most recent review round is needed to be displayed (option 1).

![Image](https://github.com/user-attachments/assets/a939d89b-8e8e-4eef-9670-244ee7b830cc)

### Application Version

OJS, OMP, OPS 3.5.0 RC2

Vitaliy-1:
PRs stable 3.5.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/11285
tests: https://github.com/pkp/ojs/pull/4841

Vitaliy-1:
Ready for review! @jardakotesovec, do you have time to take a look?

Vitaliy-1:
Not closing the issue as it still requires testing

Tribunal33:
@Vitaliy-1 a few questions.
1.  Should a declined submission show this when
2. When reviewing a previous review round in which it was declined I'm getting this error
[...]

PR #11285, opened 2025-04-16
pkp/pkp-lib#11252 Show only reviews in the last round in action-required view

Vitaliy-1:
…red view

### DIFF
--- a/api/v1/_submissions/PKPBackendSubmissionsController.php
+++ b/api/v1/_submissions/PKPBackendSubmissionsController.php
@@ -374,7 +374,8 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
         foreach ($illuminateRequest->query() as $param => $val) {
             switch ($param) {
                 case 'actionRequired':
-                    $collector->filterByActionRequiredByReviewer(true);
+                    $collector->filterByActionRequiredByReviewer(true)
+                        ->filterByLastReviewRound(true);
                     break;
                 case 'active':
                     $collector->filterByActive(true);
--- a/classes/submission/Repository.php
+++ b/classes/submission/Repository.php
@@ -1095,7 +1095,7 @@ abstract class Repository
                         ->filterByStatus([PKPSubmission::STATUS_QUEUED]);
                     return new DashboardView(
                         $key,
-                        in_array(Role::ROLE_ID_AUTHOR, $selectedRoleIds) ? __('submission.list.revisionsSubmitted')  : __('submission.dashboard.view.revisionsSubmitted'),
+                        in_array(Role::ROLE_ID_AUTHOR, $selectedRoleIds) ? __('submission.list.revisionsSubmitted') : __('submission.dashboard.view.revisionsSubmitted'),
                         [Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_ASSISTANT, Role::ROLE_ID_AUTHOR],
                         $canAccessUnassignedSubmission
                             ? $collector
@@ -1123,7 +1123,8 @@ abstract class Repository
                         Repo::reviewAssignment()->getCollector()
                             ->filterByReviewerIds([$user->getId()])
                             ->filterByContextIds([$context->getId()])
-                            ->filterByActionRequiredByReviewer(true),
+                            ->filterByActionRequiredByReviewer(true)
+                            ->filterByLastReviewRound(true),
                         'reviewerAssignments',
                         ['actionRequired' => true]
                     );


classes/submission/reviewAssignment/Collector.php
```diff
@@ -435,8 +445,24 @@ public function getQueryBuilder(): Builder
...
+                ->when(
+                    $this->isLastReviewRoundByReviewer,
+                    fn (Builder $q) => $q
+                        // Aggregate the latest review round number for the given assignment and reviewer
+                        ->leftJoinSub(
+                            DB::table('review_assignments as ramax')
+                                ->leftJoin('review_rounds as rrmax', 'ramax.review_round_id', '=', 'rrmax.review_round_id')
+                                ->select(['ramax.submission_id', 'ramax.reviewer_id'])
+                                ->selectRaw('MAX(rrmax.round) as latest_round')
+                                ->groupBy(['ramax.submission_id', 'ramax.reviewer_id']),
```
jardakotesovec, line 459: I think this temporary sub table is calculated for all review assignments in the system, right?

While we need this only for given reviewerIds. So additionally adding WHERE condition for reviewerIds here I think could be good optimisation

classes/submission/reviewAssignment/Collector.php
```diff
@@ -435,8 +445,24 @@ public function getQueryBuilder(): Builder
...
-            $q->whereIn('ra.reviewer_id', $this->reviewerIds)
+            fn (Builder $q) => $q
+                ->whereIn('ra.reviewer_id', $this->reviewerIds)
+                ->when(
+                    $this->isLastReviewRoundByReviewer,
+                    fn (Builder $q) => $q
+                        // Aggregate the latest review round number for the given assignment and reviewer
+                        ->leftJoinSub(
+                            DB::table('review_assignments as ramax')
+                                ->leftJoin('review_rounds as rrmax', 'ramax.review_round_id', '=', 'rrmax.review_round_id')
```
jardakotesovec, line 456: Interestingly review assignment seems to have both roundId and round columns, so probably no need to join it with review rounds?

classes/submission/reviewAssignment/Collector.php
```diff
@@ -435,8 +445,24 @@ public function getQueryBuilder(): Builder
 
         $q->when(
             $this->reviewerIds !== null,
-            fn (Builder $q) =>
-            $q->whereIn('ra.reviewer_id', $this->reviewerIds)
+            fn (Builder $q) => $q
```
jardakotesovec, line 448: Only open question in my mind is what to do about scenarios when theoretically (no idea if that could happen practically) same reviewer could by assigned both in internal/external review stage in OMP.  And in such scenario we might want to separate and add couple stageId conditions. But will ask first Devika what would be her expectation how that should work and whether it should be considered at all.

jardakotesovec (approved):
I think its looking good, suggested only some optimisations.

And will follow-up with Devika whether the internal/external stages should be considered and what would be expected behaviour.

classes/submission/reviewAssignment/Collector.php
Vitaliy-1: Yes, you are right

classes/submission/reviewAssignment/Collector.php
Vitaliy-1: According to the current logic, just applying the stage ID filter for OMP should be enough, will check. 
For action required by me, in my mind is additionally check submission's relevant stage.

jardakotesovec: @Vitaliy-1 But not sure when/if to apply such filter.. because we are showing review assignments regardless of what stage the submission are in..  And I don't think we have specific views for OMP..

jardakotesovec: Devika is suggesting just to keep it together. So if reviewer teoretically would have review assignment both in internal and external review stage - we would just show the latest thing.

So I can imagine either just find max round_id, I think we can assume these will always be higher with new rounds. Or somehow combine stage and round.

classes/submission/reviewAssignment/Collector.php
```diff
@@ -435,8 +445,25 @@ public function getQueryBuilder(): Builder
...
+                                ->select(['ramax.submission_id', 'ramax.reviewer_id'])
+                                ->selectRaw('MAX(ramax.round) as latest_round')
+                                ->selectRaw('MAX(ramax.stage_id) as latest_stage')
+                                ->whereIn('ramax.reviewer_id', $this->reviewerIds)
+                                ->groupBy(['ramax.submission_id', 'ramax.reviewer_id']),
+                            'agrmax',
+                            fn (JoinClause $join) => $join->on('ra.submission_id', '=', 'agrmax.submission_id')
+                        )
+                        ->whereColumn('ra.round', 'agrmax.latest_round') // assignments from the last review round per reviewer only
+                        ->whereColumn('ra.stage_id', 'agrmax.latest_stage') // assignments for the current review stage only (for OMP)
```
jardakotesovec, line 465: Sorry to be pain.. but I think this still suffers with the edge case, if you have 2 rounds in internal and 1 round in external. You would end up with latest_round:2 and latest_stage: 3, but the latest review assignment would be having round 1 and stage 3.

Vitaliy-1: Sorry, it shows the latest in the internal stage or external in this scenario?

jardakotesovec: My point is that the round is calculated independently from the stage.

So if there are 3 rounds in internal stage and 2 rounds in the external stage - it will be trying to find review assignment which is in 3rd round and external stage, which does not exist. Or do I miss something

Issue #11171, opened 2025-03-25
[Installer] | Setup - Page, files dir path not set on first installation when default path is not edited


ronste:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When the installer is run for the first time and the default value of files dir is not edited, the files dir internally is not set correctly despite being shown on the installer page as "/var/www/html/files" (seems to be the default value on the installer page).

The value "/var/www/html/files" is written as such into the config file, however, internally it still is "files" (probably loaded from config.inc.php):

![Image](https://github.com/user-attachments/assets/39f7ce5d-715e-4924-b424-3a3abf723d33)

Subsequently this reults in the following errors:

```bash
PHP Warning:  mkdir(): Permission denied in /var/www/html/ojs/lib/pkp/classes/file/FileManager.php on line 424
PHP Fatal error:  Uncaught TypeError: Cannot assign null to property PKP\\scheduledTask\\ScheduledTask::$executionLogFile of type string in /var/www/html/ojs/lib/pkp/classes/scheduledTask/ScheduledTask.php:68
```

When ignoring the errors and calling `http://localhost:7020` again OJS shows a blanck page and additional PHP errors:

```bash
PHP Fatal error:  Uncaught Error: Call to a member function getVersionString() on null in /var/www/html/ojs/lib/pkp/classes/core/PKPApplication.php:244
PHP Fatal error:  Uncaught Error: Call to a member function getContext() on null in /var/www/html/ojs/lib/pkp/classes/core/PKPApplication.php:487
```

When setting config.inc.php back to `installed = off` and run it again it works fine (as the correct value of files dir is loaded from config.inc.php).

I never edited the input field for the files folder since it was correct from the beginning. It seems there is a syncing issue between config file and default value on the install screen.

### Steps to Reproduce

1. Start a new installation of OJS on a system where the files dir is created as "/var/www/html/files"
2. Finish the installation screen without editing the files dir value (i.e. using the default value)
3. Look at PHP error log

### Expected Result

OJS should start up an run.

### Actual Result

White screen and PHP errors.

### Environment Details

Debian GNU/Linux 12

### Application Version

OJS 3.5rc2

### Logs

Full error logs

```bash
[...]

PR #11295, opened 2025-04-18
pkp/pkp-lib#11171 Config data reset for writing at installation

touhidurabir:
for pkp/pkp-lib#11171

### DIFF
--- a/classes/install/Installer.php
+++ b/classes/install/Installer.php
@@ -546,6 +546,8 @@ class Installer
             $this->wroteConfig = false;
         }
 
+        Config::resetData();
+
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


classes/install/Installer.php
```diff
@@ -949,8 +951,11 @@ public function checkPhpVersion()
      */
     public function updateRorRegistryDataset(): bool
     {
+        PKPApplication::upgrade();
```
touhidurabir, line 954: Had to add this as below class calls internally `PKPApplication::getHttpClient` which try to retrieve the version information but we register the version details at the very end of installation phase and without this get `null` which raise exception .

Issue #11303, opened 2025-04-21
[Queue Jobs] | Mail translation does not correspond to the defined locale using queue worker


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

See details at 
1. https://forum.pkp.sfu.ca/t/translation-of-announcement-notify-mail-does-not-correspond-to-the-main-language-of-the-journal-when-using-the-worker/93720
2. https://forum.pkp.sfu.ca/t/using-the-supervisor-in-combination-with-the-default-job-runner-ojs-3-4/95557

### Steps to Reproduce

Run the jobs via job runner and worker mode .

### Expected Result

proper translation handle .

### Actual Result

missing translation . 

### Application Version

OJS, OMP and OPS 3.4.0 and above

**PRs**
#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11304
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4848 
[omp](https://github.com/touhidurabir/omp) --> https://github.com/pkp/omp/pull/1991 [TEST ONLY] 
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/979 [TEST ONLY] 

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11414
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4890 
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1992 [TEST ONLY] 
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/980 [TEST ONLY] 

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11415 
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4891
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1993 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/981 [TEST ONLY]

Vitaliy-1:
We might want to overwrite the data when it's set. 
[...]

PR #11304, opened 2025-04-22
pkp/pkp-lib#11303 job worker mode locale set issue

touhidurabir:
for pkp/pkp-lib#11303 .


Vitaliy-1 (approved):
Thanks, tested and it works as expected!

asmecher (approved):

PR #11312, opened 2025-04-23
pkp/pkp-lib#11238 fix hook run so array is passed

kaitlinnewson:
For Crossref reference linking plugin.

PR to main: https://github.com/pkp/pkp-lib/pull/11313

### DIFF
--- a/classes/citation/CitationDAO.php
+++ b/classes/citation/CitationDAO.php
@@ -85,7 +85,7 @@ class CitationDAO extends \PKP\db\DAO
      * @param int $publicationId
      * @param string $rawCitationList
      *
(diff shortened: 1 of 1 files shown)


bozana:
Hi @kaitlinnewson, because we renamed the hook, and because I think this hook is not used that much (maybe only in the crossrefReferenceLinking plugin), I think we do not need to adapt the `Hook::run` to be backward compatible, I think we could treat it as all the other new hooks using `run`, i.e. to then change the parameters in the plugin's function, like this:
`public function citationsChanged(string $hookName, int $publicationId, array $existingCitations, array $importedCitations): bool`
I think this should then work?
Thanks a lot!

PR #11313, opened 2025-04-23
pkp/pkp-lib#11238 fix hook run so array is passed

kaitlinnewson:


classes/citation/CitationDAO.php
```diff
@@ -119,7 +119,7 @@ public function importCitations($publicationId, $rawCitationList)
             }
         }
 
-        Hook::run('Citation::importCitations::after', [$publicationId, $existingCitations, $importedCitations]);
+        Hook::run('Citation::importCitations::after', [[$publicationId, $existingCitations, $importedCitations]]);
```
asmecher, line 122: @kaitlinnewson, is this for compatibility with plugins that were included in previous versions?

kaitlinnewson: In the CrossrefReferenceLinking plugin I've been getting the following error without this change for 3.5:

```
[Thu Apr 24 13:37:33 2025] Plugin APP\plugins\generic\crossrefReferenceLinking\CrossrefReferenceLinkingPlugin failed to handle the hook Citation::importCitations::after
TypeError: APP\plugins\generic\crossrefReferenceLinking\CrossrefReferenceLinkingPlugin::citationsChanged(): Argument #2 ($params) must be of type array, int given in /Users/kaitlinnewson/Code/crossrefReferenceLinking/CrossrefReferenceLinkingPlugin.php:328
```

I assumed it was because of the unpacking happening with the arguments in `Hook::run` (it was previously using `Hook::call`), but perhaps there something I'm missing about hooks that we need to update on the plugin side?

Here's the Hook::add in the plugin: https://github.com/pkp/crossrefReferenceLinking/blob/d61d6d8a7f26310e91dc3b7bdbb052d06b8fa8a2/CrossrefReferenceLinkingPlugin.php#L77

...and the function: https://github.com/pkp/crossrefReferenceLinking/blob/d61d6d8a7f26310e91dc3b7bdbb052d06b8fa8a2/CrossrefReferenceLinkingPlugin.php#L327

Issue #11319, opened 2025-04-24
[Settings] | Edit Profile - Roles, removing a role doesn't reset the self-imposed user roles


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

There were a lot of trace elements to this problem, but I believe I have found the root cause.  It has to do with the new invite to a role system and the self-imposed roles that are enabled by the user in the edit profile section.  When you select the Author role from the Settings > Edit Profile > Roles section the role is applied correctly to the user & roles table.  However, if you are to remove this role using the new invite to a role system then this Author role is not removed from the profile section and can have some unintended consequences.  

![Image](https://github.com/user-attachments/assets/6dd105df-3d44-45c6-ae9b-4da1b74a2b31)

### Steps to Reproduce

1. Login with JE role (dbarnes if using basic dataset)
2. Navigate to edit profile section
3. From the Roles tabe select the Author role checkbox
4. Go to the dashboard and observe that you indeed have a new Author Dashboard view
5. Go to Settings > Users & roles and edit the JE user to bring up the new users & roles area
6. Remove the Author Role from the roles associated with this user
7. Observe that the dashboard (upon page refresh) will now have the Author Dashboard view removed even though Author role is in profile section
8. ![Image](https://github.com/user-attachments/assets/cebb400e-f92a-4bde-97b5-b984149e6379)
9. Select Edit Profile from the user Account option in top right navigation
10. Select Roles and observe that the Author role is still selected.
11. The following shows the problem with this not being deselected correclty.
12. From the dashboard menu select Create New Submission
13. Under the first step "Submit As" section notice that Author is available <-- select this
14. ![Image](https://github.com/user-attachments/assets/7f0ed81a-6308-4354-9888-9f1d953362b6)
15. Continue through the workflow as normal until Contributors Step 3 and Observe that Daniel Barnes is automatically added as an Author
16. ![Image](https://github.com/user-attachments/assets/d4521d90-4911-41bd-9752-af55e722368f)
17. Continue through the workflow until the final stage and Submit the form to complete the submission.
18. The Submission complete page will show up with the option "Review this submission" <-- Select this link
19. ![Image](https://github.com/user-attachments/assets/25217dce-c928-42c7-a1ee-71517f457040)7. Observe the authorization denied message
20. ![Image](https://github.com/user-attachments/assets/b8770a45-1e1c-4263-9308-937b44519cb2)

### Expected Result

[...]

PR #11324, opened 2025-04-28
fix edit profile role removeing reset user permission

ipula:
https://github.com/pkp/pkp-lib/issues/11319


defstat (approved):
Thanks @ipula! I am OK with those changes.

Issue #11339, opened 2025-04-29
UI Language Toggle - Language change redirects to index page


bozana:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Changing the UI language in the language block on any page redirects user to the index page.
This was find out on 3.5rc2 test installations, and it happens while $_SERVER['HTTPS_REFERER'], that we rely on, does not exist or is different.

### Steps to Reproduce

1. Navigate to any page, e.g. an article landing page
2. Click on a different UI language in the sidebar block, to change the UI language
3. See that you are redirected to the site index/landing page

### Expected Result

When changing the UI language, the user should stay on the same page, seeing the content in the newly selected language. The new language should also appear in the URL.

### Actual Result

Instead of staying on the same page when changing the UI language, the user is redirected to the site index page.

### Application Version

OJS, OMP, OPS 3.5 and main

### Additional Information

s. PKPPageRouter::_setLocale() and how we rely on $_SERVER['HTTPS_REFERER']

### PRs ###
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/11340
- https://github.com/pkp/ui-library/pull/608
- https://github.com/pkp/ojs/pull/4855
- https://github.com/pkp/omp/pull/1979
- https://github.com/pkp/ops/pull/970

main:
- https://github.com/pkp/pkp-lib/pull/11370
- https://github.com/pkp/ui-library/pull/615
- https://github.com/pkp/ojs/pull/4869
- https://github.com/pkp/omp/pull/1982
- https://github.com/pkp/ops/pull/973

bozana:
Hi @asmecher, could you please review the PRs above?
When we added the locales in URLs we changed the behavior of the language toggle plugin a little bit -- we removed the source parameter and relied on HTTPS_REFERER. This was wrong, because HTTPS_REFERER is not always there. So now I introduced the source parameter again, this time as $smarty.server.SERVER_NAME and $smarty.server.REQUEST_URI (instead of only $smarty.server.REQUEST_URI), so that the further replacement (of the old to the new locale in the URL) can correctly happen.
Also, the language toggle in the right top navigation was wrong -- it also relied on HTTPS_REFERER. Now I added a source parameter (= the current URL) there as well. I will ask Jarda to take a look at that.
Thanks a lot!

asmecher:
[...]

PR #11340, opened 2025-04-29
pkp/pkp-lib#11339 do not rely on http_referrer when using language to…

bozana:
s. https://github.com/pkp/pkp-lib/issues/11339

### DIFF
--- a/classes/core/PKPPageRouter.php
+++ b/classes/core/PKPPageRouter.php
@@ -522,20 +522,56 @@ class PKPPageRouter extends PKPRouter
             $request->setCookieVar('currentLocale', $newLocale);
         }
 
+        $protocol = $request->getProtocol();
         // Do not permit basic auth strings (user:password@url) in source parameter for redirects
-        if (preg_match('#^/\w#', $source = str_replace('@', '', $request->getUserVar('source') ?? ''))) {
-            $request->redirectUrl($source);
+        $source = str_replace('@', '', $request->getUserVar('source') ?? '');
+
+        if (isset($setLocale)) {
+            $targetUrl = '';
+            // The source parameter is coming either from the languageToggle block plugin
+            // and contains $smarty.server.SERVER_NAME|cat:$smarty.server.REQUEST_URI
+            // i.e. the whole URL except the protocol,
+            // or from TopNavActions and contain the whole URL with protocol
+            if (!empty($source) && $this->isAllowedHost($protocol, $source)) {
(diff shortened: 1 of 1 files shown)


classes/core/PKPPageRouter.php
```diff
@@ -522,14 +521,33 @@ private function _setLocale(PKPRequest $request, ?string $setLocale): void
...
+            // or from TopNavActions and contain the whole URL with protocol
+            $source = $request->getUserVar('source');
+            if (isset($source)) {
+                $targetUrl = str_replace('@', '', $source);
+                $protocol = $request->getProtocol();
+                if (!str_starts_with($source, $protocol)) {
+                    $targetUrl = $request->getProtocol() . '://' . $targetUrl;
+                }
+            } elseif (isset($_SERVER['HTTP_REFERER'])) {
+                $targetUrl = $_SERVER['HTTP_REFERER'];
```
asmecher, line 539: I think the `HTTP_REFERER` can be whatever the requester specifies, so this could potentially be a dangerous URL (e.g. something that goes to another domain). I think the only way we can prevent this from being done is to match the `allowed_hosts`.

classes/core/PKPPageRouter.php
```diff
@@ -522,14 +521,33 @@ private function _setLocale(PKPRequest $request, ?string $setLocale): void
...
+            $targetUrl = '';
+            // The source parameter is coming either from the languageToggle block plugin
+            // and contains $smarty.server.SERVER_NAME|cat:$smarty.server.REQUEST_URI
+            // i.e. the whole URL except the protocol,
+            // or from TopNavActions and contain the whole URL with protocol
+            $source = $request->getUserVar('source');
+            if (isset($source)) {
+                $targetUrl = str_replace('@', '', $source);
+                $protocol = $request->getProtocol();
+                if (!str_starts_with($source, $protocol)) {
```
asmecher, line 535: This could, I think, be used to redirect to an arbitrary domain if the leading subdomain looks like a protocol, e.g. `http.myevilsite.com`.

asmecher (changes requested):
A couple things to check, thanks!

classes/core/PKPPageRouter.php
bozana: It will not be directly redirected to this source or targetUrl -- later in the code our index URL is assumed to be part of it, but I will maybe however add a check here for allowed hosts as well...

asmecher (approved):

Issue #11318, opened 2025-04-24
[Submission Workflow] | Internal Review stage - Editorial revision upload flow, upload button uploads file to the files for review table


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

This is part of an ongoing investigation into #11219 as an attempt to see if I can figure out the root cause of the issue.  There is a similar issue here #7144.  My current hypothesis is that this button is simply adding the file to the wrong table or the author should not be able to view the Files for Review list when uploading a revision.  If I click on the upload file in the revision table I expect the file to be uploaded to this table.  There are several problems when an author is requested to do revisions with regards to this issue.  This occurs on OMP 3.4 as well as 3.5 the buttons and table names are slightly different.

Preconditions: Should have a submission in review that has completed the submission round with files that are expected to be reviewed.

### Steps to Reproduce

As an Editor:
1. Login with Editor role that can login as author
2. Navigate to the submission in the internal review round
3. Observe the files in the Files for Review table.  In the next step we will revise one of them
4. Click on Upload from the Revisions Uploaded table
5. From the drop down select one of the files from the list to revise
6. Upload a revised file and change the name so it is easy to spot
7. Complete the file upload flow
8. Observe the Review Files table

As an Author:
1. Follow the same steps as editor except you cannot observe the Review Files table
2. Upload a revision from an existing file from the drop down list
3. Complete the upload and observe that nothing is displayed for the author.

![Image](https://github.com/user-attachments/assets/8de4c1fd-6d7f-497c-943f-f572e483cf6f)

### Expected Result

As an Editor I believe if you upload a file to the revisions table it should be placed in the revisions table.  Or These files should not be part of the list of files to revise.  If you wanted to replace a file then you should do that in the upload files in the Files for Review table.

As an author.  I believe you should see the file that was revised in the Revisions table.  Or these files should be removed from the list of potential files to be revised.  

### Actual Result

The file that is revised will be replaced with the newly revised file in the Review files table.  Only if you select "This is not a revision of an existing file" will the newly added file go into the Revisions table.  

For the author even though they upload a revision of an existing file it is hidden from them because authors can not see the Review files table.

### Application Version

OMP 3.5.0rc2

jardakotesovec:
Even though this bug report is focused on internal review. Lets first establish how it should work for external review round in OJS. Behaviour in internal is probably expected to be the same.

I did test 3.3, 3.4 and 3.5. 
[...]

PR #11341, opened 2025-04-29
pkp/pkp-lib#11318 Bug fix, Don't mix revisions for different file types

jardakotesovec:


controllers/wizard/fileUpload/form/PKPSubmissionFilesUploadBaseForm.php
```diff
@@ -217,6 +217,7 @@ public function getSubmissionFiles()
                         ->getCollector()
                         ->filterByReviewRoundIds([(int) $reviewRound->getId()])
                         ->filterBySubmissionIds([$submissionId])
+                        ->filterByFileStages([(int) $this->getData('fileStage')])
                         ->getMany()
```
jardakotesovec, line 221: This make sure that we list only from the same "file grid" for the "If you are uploading a revision of an existing file, please indicate which file."

controllers/wizard/fileUpload/form/PKPSubmissionFilesUploadBaseForm.php
```diff
@@ -262,7 +263,7 @@ public function getRevisionSubmissionFilesSelection($user, $uploadedFile = null)
             }
             if (
                 ($submissionFile->getFileStage() == SubmissionFile::SUBMISSION_FILE_REVIEW_ATTACHMENT || $submissionFile->getFileStage() == SubmissionFile::SUBMISSION_FILE_REVIEW_FILE) &&
-                $stageAssignmentDao->getBySubmissionAndRoleIds($submissionFile->getData('submissionId'), [Role::ROLE_ID_AUTHOR], $this->getStageId(), $user->getId())
+                $stageAssignmentDao->getBySubmissionAndRoleIds($submissionFile->getData('submissionId'), [Role::ROLE_ID_AUTHOR], $this->getStageId(), $user->getId())->next()
             ) {
```
jardakotesovec, line 267: This is fix, to make this condition working as intended. But not sure if thats necessary anymore. As the logic seemed that `getSubmissionFiles` would return ALL files for given review round. And than the exclusions made sense to not expose some of these files to the Author.   But if getSubmissionFiles is now returning just the same fileType, I don't think thats necessary.  But keeping the changes minimum due to my limited understanding of this area :-).

Issue #11356, opened 2025-05-04
[API Routing] | Add API controller default middleware to apply on plugin override routes


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Plugin can add/override routes in 2 different approach, overriding the target controller or injecting directly in the router itself. When injecting directly in the router in following manner , the API controller's default middleware are not applied to that route . 
```php
Hook::add('APIHandler::endpoints::users', function(string $hookName, PKPBaseController &$apiController, APIHandler $apiHandler): bool {

    $apiHandler->addRoute(
        'GET/POST/PUT/PATCH/DELETE', // HTTP Request METHOD
        'some/route/path/to/add',   // The route uri
        function (IlluminateRequest $request): JsonResponse { // The closure/callback of route action handler when the route url got hit
            return response()->json([
                'message' => 'A new route added successfully',
            ], Response::HTTP_OK);
        },
        'name.of.the.route', // Name of the route
        [Role::ROLE_ID_..., Role::ROLE_ID_..., ...] // The route accessable role from `Role::ROLE_ID_*`
    );
    
    return false;
});
```

This can/may cause security issue and inconsistent behaviour when adding/overriding api routes form plugins. 

### Steps to Reproduce

Try to add a new route from plugin via directly injecting that into the router and see that none of controller's default middleware defined in `getRouteGroupMiddleware` are not applied to that routes. Try with the above code snippets in the issue description . 

### Expected Result

All of the API controller default middleware that defined in controller method `getRouteGroupMiddleware` must be applied to any routes added from plugin . 

### Actual Result

Missing controller default middleware . 

### Application Version

OJS/OMP/OPS 3.5.0 and main 

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11357
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4865 [TEST ONLY] .

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11358
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4866 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/1986 [TEST ONLY]
[...]

PR #11357, opened 2025-05-04
pkp/pkp-lib#11356 Apply default API controller middleware to routes from plugin

touhidurabir:
for pkp/pkp-lib#11356 .

### DIFF
--- a/classes/handler/APIHandler.php
+++ b/classes/handler/APIHandler.php
@@ -17,8 +17,8 @@ namespace PKP\handler;
 use Illuminate\Http\Response;
 use Illuminate\Routing\Pipeline;
 use PKP\core\PKPBaseController;
-use PKP\core\PKPContainer;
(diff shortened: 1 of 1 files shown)


classes/handler/APIHandler.php
```diff
@@ -17,8 +17,8 @@
 use Illuminate\Http\Response;
 use Illuminate\Routing\Pipeline;
 use PKP\core\PKPBaseController;
-use PKP\core\PKPContainer;
 use PKP\core\PKPRoutingProvider;
+use PKP\middleware\HasRoles;
```
ewhanson, line 21: It doesn't look like this is used anywhere in this file.

ewhanson (changes requested):
Thanks @touhidurabir! Just one question about `PKP\middleware\HasRoles` but otherwise looks good to me.

PR #11358, opened 2025-05-04
pkp/pkp-lib#11356 Apply default API controller middleware to routes from plugin

touhidurabir:
for pkp/pkp-lib#11356 .

### DIFF
--- a/classes/handler/APIHandler.php
+++ b/classes/handler/APIHandler.php
@@ -17,8 +17,8 @@ namespace PKP\handler;
 use Illuminate\Http\Response;
 use Illuminate\Routing\Pipeline;
 use PKP\core\PKPBaseController;
-use PKP\core\PKPContainer;
 use PKP\core\PKPRoutingProvider;
+use PKP\middleware\HasRoles;
 use PKP\plugins\Hook;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Throwable;
@@ -190,14 +190,31 @@ class APIHandler extends PKPHandler
         $router = app('router'); /** @var \Illuminate\Routing\Router $router */
 
         foreach ($this->routesFromHook as $routeParams) {
-            $router
-                ->addRoute(
-                    $routeParams['method'],
-                    $this->getEndpointPattern() . '/' . $routeParams['uri'],
-                    $routeParams['callback']
-                )
+            $route = $router->addRoute(
+                $routeParams['method'],
+                $this->getEndpointPattern() . '/' . $routeParams['uri'],
+                $routeParams['callback']
+            );
+
+            $route
                 ->name($routeParams['name'])
                 ->middleware($this->apiController->roleAuthorizer($routeParams['roles']));
+
+            foreach ($this->apiController->getRouteGroupMiddleware() as $middleware) {
+                
+                // As roles are already supplied for routes directly injecting in the router
+                // we do not want to add any other roles middleware form controller if defined
+                if (strstr($middleware, 'has.roles') !== false || strstr($middleware, HasRoles::class) !== false) {
+                    continue;
+                }
+
+                // We don't want to set any middleware to the route which is already set
+                if (in_array($middleware, $route->middleware())) {
+                    continue;
+                }
+
+                $route->middleware($middleware);
+            }
         }
     }
 }


ewhanson (approved):

taslangraham:
@touhidurabir Adding a note here based on our conversation from yesterday.

I've done some testing and this seems to behave in an inconsistent or rather, unclear manner. I've done my best to document the scenarios I've tested and their results below.

For testing I used the [PkpUserController](https://github.com/pkp/pkp-lib/blob/main/api/v1/users/PKPUserController.php)

I then injected an endpoint via a plugin:
```php
        Hook::add('APIHandler::endpoints::users', function(string $hookName, PKPBaseController &$apiController, \PKP\handler\APIHandler $apiHandler): bool {
            $apiHandler->addRoute(
                'GET',
                'tests',
                function ( $request): JsonResponse {
                    return response()->json([
                        'message' => 'A new route added successfully on fly',
                    ], 200);
                },
                'test.onfly',
                [
                    Role::ROLE_ID_SUB_EDITOR,
                ]
            );
            return true;
        });
```
Here's a table breakdown of the scenarios tested an the results. 


| Test scenario                                                                                | Roles In Plugin Route | Roles in Controller's Default getRouteGroupMiddleware | Logged in as                  | Expectations                                                                                             | Result                                                                                                                          |
|----------------------------------------------------------------------------------------------|-----------------------|-------------------------------------------------------|-------------------------------|----------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
| Role(ROLE_ID_SUB_EDITOR) applied in default getRouteGroupMiddleware, but not at plugin level | ROLE_ID_MANAGER       | ROLE_ID_SITE_ADMIN,ROLE_ID_MANAGER,ROLE_ID_SUB_EDITOR | sberardo (ROLE_ID_SUB_EDITOR) | ROLE_ID_SUB_EDITOR applied from default getRouteGroupMiddleware and user will get access to the endpoint | Roles in default getRouteGroupMiddleware are not applied - {"error":"You are not authorized to access the requested resource."} |
| Role(ROLE_ID_SUB_EDITOR) applied at both plugin level and default getRouteGroupMiddleware    | ROLE_ID_SUB_EDITOR    | ROLE_ID_SITE_ADMIN,ROLE_ID_MANAGER,ROLE_ID_SUB_EDITOR | sberardo (ROLE_ID_SUB_EDITOR) | ROLE_ID_SUB_EDITOR should be given access to endpoint                                                    | User accessed the endpoint successfully                                                                                         |
| Role(ROLE_ID_SUB_EDITOR) applied at plugin level, but not in default getRouteGroupMiddleware | ROLE_ID_SUB_EDITOR    | ROLE_ID_SITE_ADMIN,ROLE_ID_MANAGER                    | sberardo (ROLE_ID_SUB_EDITOR) | ROLE_ID_SUB_EDITOR at plugin level should grant access to endpoint?                                      | {"error":"You are not authorized to access the requested resource."}                                                            |

I’ve repeated these tests using different roles and the outcome were the same.


Based on these tests, it looks like the roles in default getRouteGroupMiddleware are still not been applied. Interestingly enough, the user only had access to the endpoint when the role was applied at both locations.

I don't know what the best approach is, but I think there are a few questions we need to consider here:
1. Should the roles at default getRouteGroupMiddleware override those at plugin level?
2. Should roles from both locations be merged? If yes, how does this affect existing endpoints that are been overridden?
3. Should the roles at plugin level override those at default getRouteGroupMiddleware? If yes, how does this affect existing endpoints that are been overridden?

touhidurabir:
@taslangraham thanks a lot for the detailing out the `role` behaviours . Thinking and testing this , I have following points to 

1. through the direct inject approach into the `router` directly , it's not possible to override an existing route as all routes are defined sequentially and if we have a exact same route and route method(e.g. `GET`, `POST`, `PUT` etc) injected in the `router` directly , will be after the controller defined routes so routing service will always execute the controller defined one . 
2. as it's a new route through injecting into the `router`, I think it's up to the plugin developer to decide what roles to be applied to it . If we try to force the controller defined roles of try to merge the roles with the provided roles, i think the plugin lost the freedom for custom routes . 
3. for all the routes, the API controller defined `authorize` will get to run and policies will be applied so that is great . 
4. About overriding an existing route, plugin has to extend the API controller and add it's own implementation and there is nothing we can do about it if the plugin override the default roles and the `authorize` policies . I have an example plugin https://github.com/touhidurabir/apiExample which demonstrate how to extend and override an API controller via plugin . 

I think plugin developer need to be careful when injecting or overriding a route with proper roles and authorize policies . 

about the issue of conflict of plugin provided roles and controller global roles, an update has been pushed . 

@taslangraham @ewhanson what do you think ?

ewhanson:
Thanks @touhidurabir, this all makes sense to me.

If a plugin is going to override a controller to override an existing endpoint, they are already inherently taking on risk by modifying an existing route rather than adding an existing route or modifying the data returned in another way so it makes sense that they will need to consider the roles and authorization policies. 

I did have one other question about your example plugin: When you show the example of [assigning the overridden controller](https://github.com/touhidurabir/apiExample/blob/9506eec759c8ea31a9d8e6ce35da3268c0a5f71f/ApiExamplePlugin.php#L91), it's just assigned to a local variable. Is this enough for it to get picked up, or will there be any other changes/additions necessary. For some reason, it's not clear to me from the example.

taslangraham:
Thanks for the explanation, @touhidurabir. I agree—it makes sense to give the plugin developer full control over the roles for an injected endpoint, as well as the roles and authorization for any endpoints they choose to override.

I've reviewed the updated code and approved!

taslangraham (approved):

touhidurabir:
> I did have one other question about your example plugin: When you show the example of [assigning the overridden controller](https://github.com/touhidurabir/apiExample/blob/9506eec759c8ea31a9d8e6ce35da3268c0a5f71f/ApiExamplePlugin.php#L91), it's just assigned to a local variable. Is this enough for it to get picked up, or will there be any other changes/additions necessary. For some reason, it's not clear to me from the example.

@ewhanson via the hook we are taking the reference of the api controller which is getting [passed from here](https://github.com/pkp/pkp-lib/blob/main/classes/handler/APIHandler.php#L62) and then replacing that with an extended one . However in some cases one entity may be associated with multiple api controller so in those cases we may need to check if proper instance of as we are going to check for [Plagiarism Plugin](https://github.com/pkp/plagiarism/pull/74/files#diff-ec595da12fac1c043bc6a5328d206fd14ad7786f0332ac5057aa8852c78b2255R163-R165) .

Issue #11374, opened 2025-05-06
Reviews - Reviewer is able to identify author if not assigned to the current review round


Vitaliy-1:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

If a reviewer is assigned to the submission in the double blind mode, he can access to the author identifying information if the new review round is started. See https://github.com/pkp/pkp-lib/blob/a3c62d196615af6b0ad3461eb1ff95f02658c2f7/api/v1/submissions/PKPSubmissionController.php#L1048

### Steps to Reproduce

1. Make a submission, move to the review stage and assign a reviewer in a double blind mode
2. Login as a reviewer and make a review
3. Login as an editor and create new review round
4. Make an API call as a reviewer to the `...publications` endpoint 
5. See info about the author

### Expected Result

If a reviewer is assigned in an double anonymous review, the author info should never be exposed

### Actual Result

The author name is exposed

### Application Version

OJS and OMP 3.5

Vitaliy-1:
PR stable 3.5.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/11375/files
OMP tests: https://github.com/pkp/omp/pull/1985

main pkp-lib: https://github.com/pkp/pkp-lib/pull/11376

Vitaliy-1:
@jardakotesovec, do you have time to review this?

PR #11375, opened 2025-05-06
pkp/pkp-lib#11374 Fix reviewer latest review related issues

Vitaliy-1:


classes/submission/reviewAssignment/Collector.php
```diff
@@ -284,12 +284,12 @@ public function getQueryBuilder(): Builder
         $q = DB::table($this->dao->table . ' as ra')
             // Aggregating data regarding latest review round and stage. For OMP the latest round isn't equal to the round with the highest number per submission
             ->leftJoinSub(
-                DB::table('review_assignments as rr')
+                DB::table('review_rounds as rr')
```
Vitaliy-1, line 287: Check latest round based on the review rounds table. If new review round has been started which doesn't have any review assignments, the review assignment-based approach won't show the correct result.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1050,8 +1050,7 @@ public function getPublications(Request $illuminateRequest): JsonResponse
 
         $currentUserReviewAssignment = Repo::reviewAssignment()->getCollector()
             ->filterBySubmissionIds([$submission->getId()])
-            ->filterByReviewerIds([$request->getUser()->getId()])
-            ->filterByLastReviewRound(true)
+            ->filterByReviewerIds([$request->getUser()->getId()], true)
```
Vitaliy-1, line 1053: Fix the issue with broken anonymity. Check the latest review round per reviewer per submission instead of just per submission

jardakotesovec (approved):

classes/submission/reviewAssignment/Collector.php
jardakotesovec: Only concern here would be (if its used) that there is no where condition inside this most inner look up. So I think it calculates this for all submissions in the system.  But hard to improve it until its used in some context to see whether filtering by submissionId for example woudl make sense

PR #11382, opened 2025-05-09
Beau/11381 adding projects defaults to the bug report

Tribunal33:


kaitlinnewson:
Hi @Tribunal33, based on [the GitHub docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms), I don't think this will work:

> The person opening the issue must have write permissions for the specified projects. If you don't expect people using this template to have write access, consider enabling your project's auto-add workflow. For more information, see [Adding items automatically](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/adding-items-automatically).

Tribunal33:
Hmm or maybe I can make the project public so everyone can access it?  From what I read you can't add "type" yet since it is still in beta and I couldn't find it in their schema.  Maybe it might work though and they didn't announce it -- I could give it a try first.  I originally thought about using labels for bugs but it would of relied on people adding bug labels or bug types manually.  I'll try default bug type and if that works I will auto add the project with workflows.

kaitlinnewson:
"Type" is listed [here in the docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax) as an option, so worth giving that a test to see if it works.

Tribunal33:
Added type instead of projects (due to private/write access restrictions).  Adding 'type' throws an error on my repo, but probably due to not having type in the schema.   I didn't opt in to the new beta features for sub-issues and types on my repo.  If it fails I will quickly roll it back :).

Tribunal33 (commented):
changed.

Issue #9202, opened 2023-08-01
Migrate missing email templates


Vitaliy-1:
**Describe the bug**
3.4.0-0 and 3.4.0-1 releases have missing migration for new email templates. 
The following templates should be added during migration if not exist: `REVIEW_COMPLETE`, `REVIEW_EDIT`.
The following template should be removed during migration if exists: `NOTIFICATION`. 
Alongside remove the workaround with a generic email message added here: https://github.com/pkp/pkp-lib/issues/9109#issuecomment-1652395657

**What application are you using?**
 OJS, OMP 3.4.0-0, 3.4.0-1

**Additional information**
For more details see: https://github.com/pkp/pkp-lib/issues/9109

nils-stefan-weiher:
I found one more missing e-mail template in a migration for OMP 3.4.0.
See the issue above.

Tribunal33:
@Vitaliy-1 can we close this one? The sub-issues 9109, 9217 have been closed.

Vitaliy-1:
This requires double check. @Hafsa-Naeem, can you check if all templates are migrated as expected, from version 3.3 in particular?

Vitaliy-1:
PRs (from @Hafsa-Naeem)

stable 3.5.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/11384
tests: https://github.com/pkp/ojs/pull/4874

Vitaliy-1:
Thanks, @Hafsa-Naeem!

Can you also check `NOTIFICATION` template? It was removed but I still seeing it on instances updated to 3.4.0-0 and then to 3.5.0:
```sql
SELECT * FROM email_templates_default_data WHERE email_key = 'NOTIFICATION'
```

Hafsa-Naeem:
Ready for review @Vitaliy-1 

pkp-lib: stable 3.5.0: [pkp/pkp-lib#9202 Include 3.4 missing email templates in 3.5 upgrade #11384](https://github.com/pkp/pkp-lib/pull/11384)

stable 3_4_0 test: [pkp/pkp-lib#9202 Remove obsolete NOTIFICATION rows in 3.4 upgrade #4852](https://github.com/pkp/ojs/pull/4852)

stable 3.5.0 tests: https://github.com/pkp/ojs/pull/4874

Vitaliy-1:
Thanks, @Hafsa-Naeem! Merged to 3.5.
[...]

PR #11384, opened 2025-05-12
pkp/pkp-lib#9202 Include 3.4 missing email templates in 3.5 upgrade

Hafsa-Naeem:
for [issue#9202](https://github.com/pkp/pkp-lib/issues/9202)


Vitaliy-1 (changes requested):
Thanks, that works! But also we need to remove `NOTIFICATION` as per issue description

Vitaliy-1 (approved):

Issue #9295, opened 2023-09-11
Add support for early and continuous access to version of record (continuous publication)


asmecher:
**Describe the bug**

As an Editor, I would like to be able to publish the work of authors as early as possible rather than waiting for a complete issue to be ready for simultaneous publication.

- I may want to schedule an article for an issue, but still have it available in advance of the issue’s formal publication. (I may want to use issue scheduling as organizational method for when I have a lot of articles awaiting posting.)
- I may or may not want to designate an issue at all, for example my journal may organize content through thematic areas (categories) and have no use at all for issues.
- I may decide to publish e.g. quarterly issues based on some of the content that has been posted early; this may mean going through already-posted content and publishing a formal issue based on things that have already been made available online. (For articles, this will mean that previously issueless content will have an issue.)
- I would like major functions e.g. DOIs to function in a predictable way regardless of the mix of the above.

To be determined:
- Implications for integrations – see sprint log (Hanover 2023)
- Upgrade path
- Publication dates may be a problem.

Of interest but out of scope for this issue:
- Open reviews
- Making non-final versions of articles available early
- Reconciling preprint functionality in OJS with OPS

radjr2:
We are interested in this and decided that OJS could support this as long as continuous page numbering was not used. eg page numbers restart at 1 for each issue. That way, we can assign articles to a defined issue with real page numbers and point the active issue to the current calendar issue while the published future issues are still available. What would make this really possible is if the future calendar issues could be hidden in some way or displayed some where else. We like this model as the page numbers are set as soon as they are added to the issue. That all said, a more defined process that allows continuous numbering would be great.

pmangahis:
Flagging this as an issue that would need to be solved - https://github.com/pkp/pkp-lib/issues/8998 in order to support the continuous publication model.

asmecher:
(See also NISO JAV revision for terminology around versions: https://www.niso.org/press-releases/nisos-draft-revision-journal-article-version-jav-recommended-practice-now-open)

radjr2:
Thanks Alec.


What PKP's plans to support multiple revisions and preprints?


Cheers!


Richard A. DeVito, Jr

Publisher

Weston Medical Publishing, LLC

470 Boston Post Road, Suite 301

Weston, MA 02493

www.wmpllc.org<http://www.wmpllc.org> - ***@***.***

781-899-2702 Ext. 107

339-222-4364 Cell


________________________________

touhidurabir:
[...]

PR #11385, opened 2025-05-12
pkp/pkp-lib#9295 publishing without issue assignment in editorial workflow

touhidurabir:
for pkp/pkp-lib#9295 .

### DIFF
--- a/classes/components/forms/dashboard/PKPSubmissionFilters.php
+++ b/classes/components/forms/dashboard/PKPSubmissionFilters.php
@@ -61,7 +61,7 @@ class PKPSubmissionFilters extends FormComponent
         );
     }
 
-    protected function addSectionFields(): self
+    protected function addSectionFields(): static
     {
         if (count($this->sections) === 1) {
             return $this;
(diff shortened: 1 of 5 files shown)


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1273,6 +1273,11 @@ public function editPublication(Request $illuminateRequest): JsonResponse
             return response()->json($errors, Response::HTTP_BAD_REQUEST);
         }
 
+        // If no issue is selected, it will be considered as continuous publication
```
asmecher, line 1276: If possible, any references to `issueId` should be kept in the OJS codebase. (There are a few places where we aren't 100% tidy with this.)

asmecher (changes requested):

api/v1/submissions/PKPSubmissionController.php
touhidurabir: Agreed, I was having a bit trouble where to move this part without duplicating a lot of code but now it's fine as moved to repository with different check and implementation .

touhidurabir:
> None of these changes look relevant to the tagged issue; I see the logic behind the `self` to `static` change, but what is the need for the changes related to the locale parameters?

right , most of the change of adding `?` before a nullable param is to suppress the annoying deprecation message like `DEPRECATED  PKP\core\DataObject::getData(): Implicitly marking parameter $locale as nullable is deprecated, the explicit nullable type must be used instead in lib/pkp/classes/core/DataObject.php` which is flooding the error_log file using PHP 8.4 .

PR #11388, opened 2025-05-13
Vocabulary for autosuggest fields

jardakotesovec:


classes/components/forms/FieldBaseAutosuggest.php
```diff
@@ -32,6 +32,31 @@ abstract class FieldBaseAutosuggest extends Field
...
+     *       'source' => string  // (Optional) Source of the vocabulary (e.g., 'Frascati')
+     *       'items' => [  // (Optional) Child items for hierarchical vocabularies
+     *         // Each child follows the same structure (identifier, name, source, items)
+     *       ]
+     *     ],
+     *     // Additional top-level vocabulary items...
+     *   ]
+     * ]
+     */
+    public $vocabularies = [];
```
taslangraham, line 56:
```suggestion
    public array $vocabularies = [];
```

The other properties above could be typed as well

locale/en/common.po
```diff
@@ -2105,6 +2105,13 @@ msgstr "Expand all"
 msgid "list.collapseAll"
 msgstr "Collapse all"
 
+msgid "list.expand"
+msgstr "Expand all"
+
+msgid "list.collapse"
+msgstr "Collapse all"
+
+
```
taslangraham, line 2114:
```suggestion
```

taslangraham (changes requested):
Just the two minor suggestions, but it looks good otherwise @jardakotesovec

locale/en/common.po
jardakotesovec: Actually intended to introduce new translation to have just "Collapse" and "Expand", so also fixed that.

taslangraham (approved):

asmecher (approved):

classes/category/Repository.php
```diff
@@ -104,6 +104,45 @@ public function getBreadcrumbs(LazyCollection $categories): LazyCollection
...
+
+        // Collect root items (those with no parentId)
+        $hierarchy = [];
+        $categories->each(function ($item) use (&$map, &$hierarchy) {
+            if ($item->getData('parentId') === null) {
+                $hierarchy[] = &$map[$item->getId()];
+            }
+        });
+
+        return $hierarchy;
```
taslangraham, line 142: With the new arbitrary levels of nesting in categories, we've added guards against circular references whenever we are building out the category hierarchy. To keep consistent, I think it would be beneficial to do same here.

An example can be found in the `getBreadcrumbs` [method](https://github.com/pkp/pkp-lib/blob/3aa127313b5366cd01deddb48a096164d8d12484/classes/category/Repository.php#L78)  of this same file or in the `mapByProperties` [method](https://github.com/pkp/pkp-lib/blob/3aa127313b5366cd01deddb48a096164d8d12484/classes/category/maps/Schema.php#L85-L105) of the category schema class. There's also a `hasCircularReference` [method](https://github.com/pkp/pkp-lib/blob/3aa127313b5366cd01deddb48a096164d8d12484/classes/category/Repository.php#L322C21-L322C41) in the category repository class which  makes this check easier, but at the expense of extra DB calls

If a category has a circular reference then you'd just skip over it when building out the results

taslangraham:
Just the one suggestion @jardakotesovec. Otherwise, looks good

classes/category/Repository.php
jardakotesovec: @taslangraham OKi, I did merge it. So Alec can easily incorporate it to the m10 install. But opened to improve this.

Wondering how the circular dependency can happen? What was the motivation to deal with it? Sounds like corrupted data to me.

Issue #11305, opened 2025-04-22
Add a Hook for PKPReviewerReviewStep3Form::saveForLater


JakobRing:
## Issue Description
I'm currently developing the [RQC plugin](https://github.com/prechelt/ojs-rqcplugin) for OJS 3.3 (and after that 3.5) and I want to hook into saveForLater() from Step3Form the same way as one can hook into the execute() method of the same form. For this operation, there simply is no hook available.

I see no downsides in this, so it would be an easy addition.

## What application are you using?
OJS 3.3 and 3.4

## Suggestion
I would add the hook here
https://github.com/pkp/pkp-lib/blob/f1ce2e9bf1a7d2984cfb6dbdb124942eb34b5e26/classes/submission/reviewer/form/PKPReviewerReviewStep3Form.inc.php#L214
as there is no parent method in ReviewerReviewForm or Form.

Is there a better way than adding
```
$returner = null;
if (HookRegistry::call(strtolower_codesafe(get_class($this)) . '::saveForLater', array($this, &$returner))) {
	return $returner;
}
``` 
there?

Edit:

```
HookRegistry::call(strtolower_codesafe(get_class($this)) . '::saveForLater', array($this));
``` 
is better I guess.

JakobRing:
These merge request are for stable 3.3 and 3.4 now. Should I also make merge request for 3.5 and main? Or just main?

asmecher:
Thanks, @JakobRing, I've merged the changes for `stable-3_3_0` and `stable-3_4_0`, which will be released in OJS 3.3.0-22 and 3.4.0-10, respectively. There's just one comment needing attention before I can merge this for `stable-3_5_0` and `main`, though.

asmecher:
@JakobRing, all merged, thanks!

PR #11404, opened 2025-05-18
Add a Hook for PKPReviewerReviewStep3Form::saveForLater() for stable_3_5_0

JakobRing:
See #11305

### DIFF
--- a/classes/submission/reviewer/form/PKPReviewerReviewStep3Form.php
+++ b/classes/submission/reviewer/form/PKPReviewerReviewStep3Form.php
@@ -22,6 +22,7 @@ use APP\notification\NotificationManager;
 use APP\submission\Submission;
 use APP\template\TemplateManager;
 use Illuminate\Support\Facades\Mail;
+use PKP\plugins\Hook;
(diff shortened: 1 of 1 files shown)


classes/submission/reviewer/form/PKPReviewerReviewStep3Form.php
```diff
@@ -277,6 +278,7 @@ public function saveForLater()
             'recommendation' => (int) $this->getData('recommendation'), // save the recommendation to the review assignment
         ]);
 
+        Hook::call(strtolower_codesafe(get_class($this)) . '::saveForLater', array($this));
```
asmecher, line 281: As of PHP8.2, the PHP `strtolower` function is no longer locale-dependent, so we've removed the `strtolower_codesafe` function from our codebase in `stable-3_5_0` and `main`. Use `strtolower` instead. Also, the `[ ... ]` syntax is preferred to `array( ... )`.
```suggestion
        Hook::call(strtolower(get_class($this)) . '::saveForLater', [$this]);
```

JakobRing: I did replace the commit with the suggested changes from you.

PR #11427, opened 2025-05-27
data citations field authors

GaziYucel:


classes/components/forms/FieldAuthors.php
```diff
@@ -0,0 +1,56 @@
...
+    public $component = 'field-authors';
+
+    /** @copydoc Field::$component */
+    public $default = [];
+
+    /**
+     * Primary language
+     * Filled in components/ListPanel/ContributorsListPanel.vue
+     */
+    public string $primaryLocale = '';
```
jardakotesovec: Just to clean up things bit more. I don't think thats necessary. Thats something thats being passed to the field on the vue side. But its not something you would want to configure when creating the field configuration in php. Same I think is true for supportedFormLocales below

classes/components/forms/FieldAuthors.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @copydoc Field::getConfig()
+     */
+    public function getConfig(): array
+    {
+        $config = parent::getConfig();
+        $config['value'] = $this->value ?? $this->default ?? null;
+        $config['label'] = $this->label;
+        $config['description'] = $this->description;
```
jardakotesovec: name, label, value are already handled in parent, right?
So I think we need to handle only the addButtonLabel

PR #11441, opened 2025-06-02
pkp/pkp-lib#10133 Port Genre and GenreDAO to Use Eloquent Model

Hafsa-Naeem:
for [#10133](https://github.com/pkp/pkp-lib/issues/10133)

### DIFF
--- a/api/v1/_submissions/PKPBackendSubmissionsController.php
+++ b/api/v1/_submissions/PKPBackendSubmissionsController.php
@@ -196,8 +196,7 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
 
         $userGroups = UserGroup::withContextIds($contextId)->cursor();
 
-
-        $genres = Repo::genre()->getByContextId($context->getId());
+        $genres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
 
         return response()->json([
             'itemsMax' => $collector->getCount(),
@@ -231,7 +230,7 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
         $contextId = $context->getId();
         $userGroups = UserGroup::withContextIds($contextId)->cursor();
 
-        $genres = Repo::genre()->getByContextId($context->getId());
+        $genres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
 
         return response()->json([
             'itemsMax' => $collector->getCount(),
@@ -309,7 +308,7 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
         $contextId = $context->getId();
         $userGroups = UserGroup::withContextIds($contextId)->cursor();
 
-        $genres = Repo::genre()->getByContextId($context->getId());
+        $genres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
 
 
         return response()->json([
--- a/classes/submission/genre/Genre.php
+++ b/classes/submission/genre/Genre.php
@@ -33,7 +33,7 @@ class Genre extends Model
 
     protected $table = 'genres';
     protected $primaryKey = 'genre_id';
-    public    $timestamps = false;
+    public $timestamps = false;
     protected $guarded = ['genreId','id'];
 
     public function getSettingsTable(): string
@@ -95,22 +95,17 @@ class Genre extends Model
     
     protected static function booted(): void
     {
-        parent::booted();
-
-        // always order by seq
         static::addGlobalScope('orderBySeq', function (EloquentBuilder $builder) {
             $builder->orderBy('seq');
         });
 
-        // ensure seq is set on insert
-        static::creating(function ($genre) {
+        static::creating(function (self $genre) {
             if ($genre->getAttribute('seq') === null) {
-                $maxSeq = (int) self::where(
-                    'context_id',
-                    $genre->getAttribute('context_id')
-                )->max('seq');
+                $maxSeq = static::withoutGlobalScope('orderBySeq')
+                    ->where('context_id', $genre->getAttribute('context_id'))
+                    ->max('seq');
 
-                $genre->setAttribute('seq', $maxSeq + 1);
+                $genre->setAttribute('seq', ((int) $maxSeq) + 1);
             }
         });
     }
@@ -191,7 +186,8 @@ class Genre extends Model
      */
     public static function findById(int $id, ?int $contextId = null): ?self
     {
-        $query = self::where('genre_id', $id);
+        $query = static::withoutGlobalScope('orderBySeq')
+            ->where('genre_id', $id);
         if ($contextId !== null) {
             $query->withContext($contextId);
         }
@@ -203,8 +199,8 @@ class Genre extends Model
      */
     public static function findByKey(string $key, ?int $contextId = null): ?self
     {
-        $lowerKey = strtolower($key);
-        $query = self::where('entry_key', $lowerKey);
+        $query = static::withoutGlobalScope('orderBySeq')
+            ->where('entry_key', strtolower($key));
         if ($contextId !== null) {
             $query->withContext($contextId);
         }
@@ -227,7 +223,7 @@ class Genre extends Model
     protected function getDefaultKeys(): array
     {
         $xmlDao = new XMLDAO();
-        $data   = $xmlDao->parseStruct('registry/genres.xml', ['genre']);
+        $data = $xmlDao->parseStruct('registry/genres.xml', ['genre']);
         if (empty($data['genre'])) {
             return [];
         }
--- a/classes/submission/genre/Repository.php
+++ b/classes/submission/genre/Repository.php
@@ -25,10 +25,11 @@ class Repository
     public function getByDependenceAndContextId(bool $dependentFilesOnly, int $contextId): Collection
     {
         return Genre::query()
-                    ->withEnabled()
-                    ->withContext($contextId)
-                    ->withDependent($dependentFilesOnly)
-                    ->get();
+            ->withEnabled()
+            ->withContext($contextId)
+            ->withDependent($dependentFilesOnly)
+            ->get()
+            ->keyBy('genre_id');
     }
 
     /**
@@ -37,10 +38,11 @@ class Repository
     public function getBySupplementaryAndContextId(bool $supplementaryFilesOnly, int $contextId): Collection
     {
         return Genre::query()
-                    ->withEnabled()
-                    ->withContext($contextId)
-                    ->withSupplementary($supplementaryFilesOnly)
-                    ->get();
+            ->withEnabled()
+            ->withContext($contextId)
+            ->withSupplementary($supplementaryFilesOnly)
+            ->get()
+            ->keyBy('genre_id');
     }
 
     /**
@@ -49,11 +51,12 @@ class Repository
     public function getPrimaryByContextId(int $contextId): Collection
     {
         return Genre::query()
-                    ->withEnabled()
-                    ->withContext($contextId)
-                    ->where('dependent', 0)
-                    ->where('supplementary', 0)
-                    ->get();
+            ->withEnabled()
+            ->withContext($contextId)
+            ->where('dependent', 0)
+            ->where('supplementary', 0)
+            ->get()
+            ->keyBy('genre_id');
     }
 
     /**
@@ -61,7 +64,7 @@ class Repository
      */
     public function getByContextId(int $contextId): Collection
     {
-        return Genre::withContext($contextId)->get();
+        return Genre::withContext($contextId)->get()->keyBy('genre_id');
     }
 
     /**
@@ -103,10 +106,11 @@ class Repository
     public function getRequiredToSubmit(int $contextId): Collection
     {
         return Genre::query()
-                    ->withEnabled()
-                    ->withContext($contextId)
-                    ->withRequired(true)
-                    ->get();
+            ->withEnabled()
+            ->withContext($contextId)
+            ->withRequired(true)
+            ->get()
+            ->keyBy('genre_id');
     }
 
     /**
@@ -137,17 +141,17 @@ class Repository
         }
 
         $rows = [];
-        $seq = 0;
+        $seq = 1;
         foreach ($data['genre'] as $entry) {
             $attrs = $entry['attributes'];
 
             $rows[] = [
-                'entry_key' => $attrs['key'],
+                'entry_key' => strtolower($attrs['key']),
                 'seq' => $seq,
                 'context_id' => $contextId,
                 'category' => (int) $attrs['category'],
-                'dependent' => $attrs['dependent'],
-                'supplementary' => $attrs['supplementary'],
+                'dependent' => (int) $attrs['dependent'],
+                'supplementary' => (int) $attrs['supplementary'],
                 'required' => $attrs['required'] ?? false,
                 'enabled' => 1,
             ];
@@ -155,17 +159,13 @@ class Repository
             $seq++;
         }
 
-        Genre::upsert(
-            $rows,
-            ['context_id', 'entry_key'],
-            ['seq', 'category', 'dependent', 'supplementary', 'required', 'enabled']
-        );
+        Genre::insert($rows);
 
         foreach ($data['genre'] as $entry) {
             $attrs = $entry['attributes'];
 
             $genre = Genre::findByKey($attrs['key'], $contextId);
-            if (! $genre) {
+            if (!$genre) {
                 continue;
             }
 
@@ -184,7 +184,7 @@ class Repository
     public function deleteSettingsByLocale(string $locale): void
     {
         DB::table('genre_settings')
-          ->where('locale', $locale)
-          ->delete();
+            ->where('locale', $locale)
+            ->delete();
     }
 }
--- a/controllers/wizard/fileUpload/FileUploadWizardHandler.php
+++ b/controllers/wizard/fileUpload/FileUploadWizardHandler.php
@@ -457,15 +457,15 @@ class FileUploadWizardHandler extends Handler
         ]);
 
         $context = $request->getContext();
+        $submission = $this->getSubmission();
         $submissionFile = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION_FILE);
         $form = new SubmissionFilesMetadataForm($submissionFile, $this->getStageId(), $this->getReviewRound());
         $form->initData();
 
-        $fileGenres = Repo::genre()->getByContextId($context->getId());
+        $fileGenres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
 
         $fileData = Repo::submissionFile()
-            ->getSchemaMap()
-            ->map($submissionFile, $fileGenres->all());
+            ->getSchemaMap($submission, $fileGenres->all())->map($submissionFile);
 
         $json = new JSONMessage(true, $form->fetch($request));
         $json->setGlobalEvent('submissionFile:added', $fileData);
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -485,7 +485,7 @@ class PKPSubmissionController extends PKPBaseController
 
         $userGroups = UserGroup::withContextIds($context->getId())->cursor();
 
-        $genres = Repo::genre()->getByContextId($context->getId());
+        $genres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
 
         return response()->json([
             'itemsMax' => $collector->getCount(),
@@ -597,7 +597,7 @@ class PKPSubmissionController extends PKPBaseController
 
         $anonymizeReviews = $this->anonymizeReviews($submission, $reviewAssignments);
 
-        $genres = Repo::genre()->getByContextId($submission->getData('contextId'));
+        $genres = Repo::genre()->getByContextId($submission->getData('contextId'))->keyBy('genre_id');
 
 
         return response()->json(Repo::submission()->getSchemaMap()->map(
--- a/api/v1/submissions/tasks/EditorialTaskController.php
+++ b/api/v1/submissions/tasks/EditorialTaskController.php
@@ -57,7 +57,6 @@ use PKP\security\authorization\SubmissionAccessPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
 use PKP\stageAssignment\StageAssignment;
-use PKP\submission\GenreDAO;
 use PKP\submission\reviewAssignment\ReviewAssignment;
 use PKP\submissionFile\SubmissionFile;
 use PKP\user\User;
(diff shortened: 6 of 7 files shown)


classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        return $query->where('supplementary', (int) $supplementary);
+    }
+
+    public function scopeRequired(EloquentBuilder $query, bool $required = true): EloquentBuilder
+    {
+        return $query->where('required', (int) $required);
+    }
+
+    public function scopeInContext(EloquentBuilder $query, int $contextId): EloquentBuilder
```
Vitaliy-1: `scopeWithContexts` is more common in our patterns

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        return Attribute::make(
+            get: fn($value) => (bool) $value,
+            set: fn($value) => (int) $value
+        );
+    }
+
+    // Scopes
+
+    public function scopeEnabled(EloquentBuilder $query): EloquentBuilder
```
Vitaliy-1: It looks that our usual pattern is using `scope` key prefix together with `with`, i.e. `scopeWithEnabled`. For boolean values I see several cases with prefix `scopeIs...`. Anyway, it's better to use such prefix to easily identify the `scope`-related method and I'd use the former option.

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        return $query->where('enabled', 1);
+    }
+
+    public function scopeDependent(EloquentBuilder $query, bool $dependent = true): EloquentBuilder
+    {
+        return $query->where('dependent', (int) $dependent);
```
Vitaliy-1: Doesn't eloquent automatically resolve boolean values? Do we need to case to int in this and similar cases?

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,154 @@
...
+            get: fn($value) => strtoupper($value),
+            set: fn($value) => strtolower($value)
+        );
+    }
+
+    protected function enabled(): Attribute
+    {
+        return Attribute::make(
+            get: fn($value) => (bool) $value,
+            set: fn($value) => (int) $value
```
Vitaliy-1: The same as above

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+     */
+    public function get(int $genreId, ?int $contextId = null): ?Genre
+    {
+        return Genre::findById($genreId, $contextId);
+    }
+
+    /**
+     * Retrieve a Genre by its entry_key (and optionally by context) using the model’s own finder.
+     */
+    public function getByKey(string $key, ?int $contextId = null): ?Genre
```
Vitaliy-1: Do we need this method and the method above in this helper? They are just calling the similar method of the Model

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+    {
+        return Genre::findByKey($key, $contextId);
+    }
+
+    /**
+     * Does a Genre with this key already exist? Exclude one ID if needed.
+     */
+    public function keyExists(string $key, int $contextId, ?int $excludeGenreId = null): bool
```
Vitaliy-1: What about moving this and similar logic to Model's scope? We just could do something like
```php
Genre::findByKey($key, $contextId)->withoutIds($excludeGenreIds)->exists();
```

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+        if ($excludeGenreId !== null) {
+            $query->where('genre_id', '<>', $excludeGenreId);
+        }
+        return $query->exists();
+    }
+
+    /**
+     * Get all enabled genres in a given context, ordered by seq.
+     */
+    public function getEnabledByContextId(int $contextId): Collection
```
Vitaliy-1: I'd say that it's questionable which call is better in terms of readability and conciseness between:
```php
Repo::genre()->getEnabledByContextId(int $contextId); 
```
```php
Genre::withEnabled()->withContextIds([$ids])->get();
```
As you can see, they are quite the same. I'd prefer to keep such calls through the Model directly.

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+    {
+        return Genre::query()
+                    ->enabled()
+                    ->inContext($contextId)
+                    ->orderBy('seq')
```
Vitaliy-1: I see that this ordering is used in almost all methods, does it make sense to add an anonymous global scope? https://laravel.com/docs/12.x/eloquent#anonymous-global-scopes

classes/submission/genre/Repository.php
Vitaliy-1: I think that's relevant for most methods in this class

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Permanently delete all genres (and their locale settings) for a given context_id.
+     * (Mirrors the old DAO’s deleteByContextId().)
+     */
+    public function deleteByContextId(int $contextId): void
+    {
+        // First, remove all locale‐specific rows from genre_settings:
+        $all = Genre::inContext($contextId)->get();
+        foreach ($all as $genre) {
+            DB::table('genre_settings')
```
Vitaliy-1: can you double check if settings aren't remove automatically?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+                    ->enabled()
+                    ->inContext($contextId)
+                    ->required(true)
+                    ->get();
+    }
+
+    /**
+     * Soft‐delete a genre: set enabled = 0.
+     */
+    public function deleteById(int $genreId): bool
```
Vitaliy-1: Is this used anywhere?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+        foreach ($data['genre'] as $entry) {
+            $attrs = $entry['attributes'];
+
+            // Try to find an existing Genre by (entry_key, context_id):
+            $existing = Genre::findByKey($attrs['key'], $contextId);
+
+            if (! $existing) {
+                // Create a brand‐new Genre model (ModelWithSettings will insert into genre_settings automatically)
+                $genre = new Genre([
+                    'entry_key'     => $attrs['key'],
```
Vitaliy-1: I don't think we use such formatting with multiple spaces

Vitaliy-1: Can we avoid using snake_case here?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+                    'category'      => (int) $attrs['category'],
+                    'dependent'     => (bool) $attrs['dependent'],
+                    'supplementary' => (bool) $attrs['supplementary'],
+                    'required'      => (bool) ($attrs['required'] ?? false),
+                    'enabled'       => 1,
+                ]);
+                $genre->save();
+            } else {
+                // Already exists: re‐enable and update any changed flags (keep the same ID):
+                $existing->enabled       = 1;
```
Vitaliy-1: It's better to use `fill()` method. e.g.,
```php
$attributes = [
   'seq' => $seq,
   'category' => $category,
   ...
];
```
```php
new Genre(array_merge(['entryKey' => $entryKey, 'contextId' => $contextId], $attibutes))
```
```
$existing->fill($attributes);
```

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+                $existing->save();
+                $genre = $existing;
+            }
+            // collect all localized names first, then fill/save once
+            $allNames = [];
+            foreach ($locales as $locale) {
+                $allNames[$locale] = __($attrs['localeKey'], [], $locale);
+            }
+            $genre->fill([ 'name' => $allNames ]);
+            $genre->save();
```
Vitaliy-1, line 177: Above there is at least one additional `save` call, is it required?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,256 @@
...
+        if (empty($data['genre'])) {
+            return;
+        }
+
+        $seq = 0;
+        foreach ($data['genre'] as $entry) {
+            $attrs = $entry['attributes'];
+
+            // Try to find an existing Genre by (entry_key, context_id):
+            $existing = Genre::findByKey($attrs['key'], $contextId);
```
Vitaliy-1: So the logic is here is identical to `upsert`? In this case Laravel supports this in one operation and we can simply form and array of values and perform one db operation: https://laravel.com/docs/12.x/eloquent#upserts, avoiding several update/insert operations in each iteration

controllers/grid/settings/genre/form/GenreForm.php
```diff
@@ -85,21 +84,19 @@ public function initData($args = [])
...
-                'dependent' => $genre->getDependent(),
-                'supplementary' => $genre->getSupplementary(),
-                'required' => $genre->getRequired(),
-                'key' => $genre->getKey(),
+                'name' => [], // this would be replaced by the localized name retrieval logic.
+                'category' => $genre->category,
+                'dependent' => $genre->dependent,
+                'supplementary' => $genre->supplementary,
+                'required' => $genre->required,
+                'key' => $genre->entry_key,
```
Vitaliy-1: `$genre->entryKey`

controllers/grid/settings/genre/form/GenreForm.php
```diff
@@ -146,33 +143,38 @@ public function readInputData()
...
-        $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
         $request = Application::get()->getRequest();
         $context = $request->getContext();
 
         // Update or insert genre
         if (!$this->getGenreId()) {
-            $genre = $genreDao->newDataObject();
-            $genre->setContextId($context->getId());
+            $genre = new Genre();
+            $genre->context_id = $context->getId();
```
Vitaliy-1: `$genre->contextId`

controllers/grid/settings/genre/form/GenreForm.php
```diff
@@ -146,33 +143,38 @@ public function readInputData()
...
-        $genre->setCategory($this->getData('category'));
-        $genre->setDependent($this->getData('dependent'));
-        $genre->setSupplementary($this->getData('supplementary'));
-        $genre->setRequired((bool) $this->getData('required'));
+        $localizedNames = $this->getData('name');
+        $genre->fill([
+            'name' => $localizedNames
+        ]);
+
+        $genre->category = $this->getData('category');
```
Vitaliy-1: Can all the data be filled through the correspondent method, all-together?

controllers/grid/settings/genre/form/GenreForm.php
```diff
@@ -146,33 +143,38 @@ public function readInputData()
...
+        $genre->dependent = $this->getData('dependent');
+        $genre->supplementary = $this->getData('supplementary');
+        $genre->required = (bool)$this->getData('required');
 
         if (!$genre->isDefault()) {
-            $genre->setKey($this->getData('key'));
+            $genre->entry_key = $this->getData('key');
```
Vitaliy-1: `$genre->entryKey`

pages/decision/DecisionHandler.php
```diff
@@ -255,15 +253,14 @@ protected function getFileGenres(Context $context): array
...
+
         /** @var Genre $genre */
-        while ($genre = $genreResults->next()) {
+        foreach ($genreResults as $genre) {
             $fileGenres[] = [
-                'id' => $genre->getId(),
-                'name' => $genre->getLocalizedName(),
-                'isPrimary' => !$genre->getSupplementary() && !$genre->getDependent(),
+                'id' => $genre->getKey(),
+                'name' => $genre->name, // TODO: Handle localization once settings handling is complete
```
Vitaliy-1: Shouldn't this be changed according to the comment?

pages/submission/PKPSubmissionHandler.php
```diff
@@ -209,16 +208,14 @@ protected function showWizard(array $args, Request $request, Submission $submiss
             ->withContextIds([$context->getId()])
             ->get();
 
-        /** @var GenreDAO $genreDao */
-        $genreDao = DAORegistry::getDAO('GenreDAO');
-        $genres = $genreDao->getEnabledByContextId($context->getId())->toArray();
+        $genres = Repo::genre()->getEnabledByContextId($context->getId());
```
Vitaliy-1: Better to do conversion to array in one place 
```
$genres = Repo::genre()->getEnabledByContextId($context->getId())->all();
```

plugins/importexport/native/filter/NativeXmlSubmissionFileFilter.php
```diff
@@ -88,11 +88,10 @@ public function handleElement($node)
         // Build a cached list of genres by context ID by name
         if ($genreName) {
             if (!isset($genresByContextId[$context->getId()])) {
-                $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
-                $genres = $genreDao->getByContextId($context->getId());
-                while ($genre = $genres->next()) {
-                    foreach ($genre->getName(null) as $locale => $name) {
-                        $genresByContextId[$context->getId()][$name] = $genre;
+                $genres = Genre::where('context_id', $context->getId())->get();
```
Vitaliy-1: Better to use here scope

schemas/genre.json
Vitaliy-1, line 1: Can you recall why json file is needed? Don't we have other ways to specify this data types in the Model?

schemas/submissionFile.json
```diff
@@ -127,7 +127,7 @@
 			"apiSummary": true
 		},
 		"genreName": {
-			"type": "string",
+			"type": "object",
```
Vitaliy-1: Does the `string` data type lead to errors somewhere? I see that it differs from the usual pattern

api/v1/jats/PKPJatsController.php
```diff
@@ -192,11 +190,12 @@ public function delete(Request $illuminateRequest): JsonResponse
         $publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION);
 
         $context = Application::get()->getRequest()->getContext();
-        $genreDao = DAORegistry::getDAO('GenreDAO');
-        $genres = $genreDao->getEnabledByContextId($context->getId());
+
+        $genres = Repo::genre()->getEnabledByContextId($context->getId());
```
Vitaliy-1: It's better to convert to array with `all()` here in a single call. There several other places where it can be improved

classes/migration/install/GenresMigration.php
```diff
@@ -57,7 +57,7 @@ public function up(): void
             $table->string('locale', 28)->default('');
             $table->string('setting_name', 255);
             $table->mediumText('setting_value')->nullable();
-            $table->string('setting_type', 6)->comment('(bool|int|float|string|object)');
+            $table->string('setting_type', 6)->default('string')->comment('(bool|int|float|string|object)');
```
Vitaliy-1, line 60: If changing the default type in the installation script, it also should be reflected in the upgrade migration

classes/services/PKPSchemaService.php
```diff
@@ -686,3 +688,29 @@ public function addMissingMultilingualValues($schemaName, $values, $localeKeys)
         return $values;
     }
 }
+
+if (!PKP_STRICT_MODE) {
+    class_alias('\PKP\services\PKPSchemaService', '\PKPSchemaService');
```
Vitaliy-1: Not sure if we need this. Is there a plugin that uses old calls?

classes/services/PKPSchemaService.php
```diff
@@ -53,6 +53,8 @@ class PKPSchemaService
     public const SCHEMA_USER_GROUP = 'userGroup';
     public const SCHEMA_EVENT_LOG = 'eventLog';
     public const SCHEMA_EMAIL_LOG = 'emailLog';
+    public const SCHEMA_GENRE = 'genre';
```
Vitaliy-1, line 61: I don't think we need to use schema. Take a look at the `ModelWithSettings` trait, it should have everything required. I believe @touhidurabir has some examples

api/v1/_submissions/PKPBackendSubmissionsController.php
```diff
@@ -233,16 +230,14 @@ public function assigned(Request $illuminateRequest): JsonResponse
         $contextId = $context->getId();
         $userGroups = UserGroup::withContextIds($contextId)->cursor();
 
-        /** @var \PKP\submission\GenreDAO $genreDao */
-        $genreDao = DAORegistry::getDAO('GenreDAO');
-        $genres = $genreDao->getByContextId($context->getId())->toAssociativeArray();
+        $genres = Repo::genre()->getByContextId($context->getId())->keyBy('genre_id');
```
Vitaliy-1, line 233: It looks that `keyBy` is already applied at `getByContextId`

api/v1/_submissions/PKPBackendSubmissionsController.php
```diff
@@ -313,16 +308,15 @@ public function reviews(Request $illuminateRequest): JsonResponse
...
```
Vitaliy-1, line 311: The same. Can you revise all such cases

classes/publication/Repository.php
```diff
@@ -414,20 +414,10 @@ public function version(Publication $publication, ?VersionStage $versionStage =
             }
         }
 
-        // Clone data citations if any
-        $dataCitations = $publication->getData('dataCitations');
```
Vitaliy-1, line 418: Was this part intentionally removed?

classes/services/PKPSchemaService.php
Vitaliy-1: Do we need schema name for Genre?

classes/submission/genre/Genre.php
```diff
@@ -0,0 +1,237 @@
...
+            get: fn($value) => (bool) $value,
+            set: fn($value) => !empty($value) ? 1 : 0,
+        );
+    }
+
+    protected function dependent(): Attribute
+    {
+        return Attribute::make(
+            get: fn($value) => (bool) $value,
+            set: fn($value) => !empty($value) ? 1 : 0,
```
Vitaliy-1, line 125: Doesn't Eloquent handle such cases automatically when casts are set?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,190 @@
...
+     * Get genres by dependent flag in a context, ordered by seq.
+     */
+    public function getByDependenceAndContextId(bool $dependentFilesOnly, int $contextId): Collection
+    {
+        return Genre::query()
+            ->withEnabled()
+            ->withContext($contextId)
+            ->withDependent($dependentFilesOnly)
+            ->get()
+            ->keyBy('genre_id');
```
Vitaliy-1, line 32: Here and in other cases, could we use camelCase attribute rather than snake_case column name as we already have a collection of models?

classes/submission/genre/Repository.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Permanently delete all genres (and their locale settings) for a given context_id.
+     * (Mirrors the old DAO’s deleteByContextId().)
+     */
+    public function deleteByContextId(int $contextId): void
+    {
+        $genres = Genre::withContext($contextId)->get();
+        foreach ($genres as $genre) {
+            $genre->delete();
```
Vitaliy-1, line 124: What about deleting in one query without model instantiation?

controllers/grid/settings/genre/GenreGridHandler.php
```diff
@@ -42,10 +41,10 @@ public function __construct()
     {
         parent::__construct();
         $this->addRoleAssignment([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], [
-            'fetchGrid', 'fetchRow',
-            'addGenre', 'editGenre', 'updateGenre',
-            'deleteGenre', 'restoreGenres', 'saveSequence'
-        ]);
+                'fetchGrid', 'fetchRow',
```
Vitaliy-1, line 44: Is the indentation here and this class in general correct?

controllers/grid/settings/genre/GenreGridHandler.php
```diff
@@ -120,8 +119,7 @@ protected function loadData($request, $filter)
     {
         // Elements to be displayed in the grid
         $context = $request->getContext();
-        $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
-        return $genreDao->getEnabledByContextId($context->getId(), self::getRangeInfo($request, $this->getId()));
```
Vitaliy-1, line 124: The before logic includes get the `self::getRangeInfo` which I don't see in the new code

controllers/wizard/fileUpload/FileUploadWizardHandler.php
```diff
@@ -499,13 +497,12 @@ public function finishFileSubmission($args, $request)
...
-        /** @var GenreDAO $genreDao */
-        $genreDao = DAORegistry::getDAO('GenreDAO');
-        $fileGenres = $genreDao->getByContextId($request->getContext()->getId())->toAssociativeArray();
+        $contextId = $request->getContext()->getId();
+        $fileGenres = Repo::genre()->getByContextId($contextId);
 
         $fileData = Repo::submissionFile()
-            ->getSchemaMap($submission, $fileGenres)
-            ->map($submissionFile);
+            ->getSchemaMap()
```
Vitaliy-1, line 504: The method accepts 2 arguments

pages/authorDashboard/PKPAuthorDashboardHandler.php
```diff
@@ -168,8 +168,9 @@ public function setupTemplate($request)
             $submissionContext = app()->get('context')->get($submission->getData('contextId'));
         }
 
-        $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
-        $contextGenres = $genreDao->getEnabledByContextId($submission->getData('contextId'))->toArray();
+        $contextUserGroups = Repo::userGroup()->getByRoleIds([Role::ROLE_ID_AUTHOR], $submission->getData('contextId'));
```
Vitaliy-1, line 171: Is this used anywhere?

schemas/genre.json
Vitaliy-1: Do we need genre schema JSON?

templates/controllers/wizard/fileUpload/form/submissionFileMetadataForm.tpl
```diff
@@ -33,7 +33,7 @@
 		{/fbvFormSection}
 
 		{* Supplementary file metadata *}
-		{if $genre && $genre->getCategory() == PKP\submission\Genre::GENRE_CATEGORY_SUPPLEMENTARY}
+		{if $genre && $genre->category == $smarty.const.GENRE_CATEGORY_SUPPLEMENTARY}
```
Vitaliy-1, line 36: Probably fully qualified class name is better but not sure

PR #11446, opened 2025-06-03
add public file url visible to the grid

ajnyga:


ajnyga:
This is for Oslo Sprint testing...

Issue #11417, opened 2025-05-23
Identified regressions with #9658


ewhanson:
# Identified issues

## Part 1 (breaking or "critical")
 
- [x] When removing a role from a user via the invitation list, the role is not included in the email. 👇

<details>
<img width="1151" alt="Image" src="https://github.com/user-attachments/assets/72cc5d56-2906-45cb-8109-2e1513a2918e" />
</details>

- [x] Invitation to a new role for an existing user: grammar/style issues. 👇

<details>
<img width="1253" alt="Image" src="https://github.com/user-attachments/assets/0e43a08e-6b80-4675-b662-c3767f376545" />
</details>

- [x] When an existing user being added to a role wants to verify their ORCID and the successfully log in to ORCID, the redirect does not correctly advance the page to the next step. The ORCID is correctly attached to the user, but the user will need to click "Skip ORCID verification" to get to the next step. ORCID "verification" badge also missing. 👇

<details>
https://github.com/user-attachments/assets/5eade6df-4a7e-42fc-87b5-0d4a7258dd54
</details>

- [x] When an ORCID is added via invitation process for an existing user when adding a role, it does not show up in the user edit screen. Seems like it's not being saved properly as part of invitation process. 👇

<details>
<img width="1167" alt="Image" src="https://github.com/user-attachments/assets/f59f1022-f078-48fc-b5f0-3c1b551a5f32" />
</details>

- [x]  Wrong timestamp format for existing user with a role adding a new role email: 👇

<details>
<img width="1269" alt="Image" src="https://github.com/user-attachments/assets/771b2997-c7d0-4c79-ad81-3e0f4bf9fc5c" /> 
 
NB: see `UserRoleAssignmentInvitationNotify::getUserUserGroupSection()`, date should be formatted, taken as-is from db.
</details>

- [x] "Search for user by ORCID" doesn't work without the URL to find an existing user even though description say search just by ORCID. Should probably work with both: 👇

<details>
https://github.com/user-attachments/assets/0cfb8eff-52f5-4fe5-a7ed-2df44e3c08b6
</details>

- [x] Roles beyond first existing role not represented in email list of existing roles: 👇

<details>
<img width="1101" alt="Image" src="https://github.com/user-attachments/assets/e3d162b2-a3df-4d59-b2f7-66d42b15ecd6" />

<img width="1248" alt="Image" src="https://github.com/user-attachments/assets/38653c15-20bd-4b42-9a8d-86f6b68e62d9" />
[...]

PR #11459, opened 2025-06-06
pkp/pkp-lib#11417 Fix pre-release invitation regressions

ewhanson:


api/v1/users/PKPUserController.php
```diff
@@ -321,7 +321,7 @@ public function endRole(Request $request): JsonResponse
         Repo::userGroup()->endAssignments($context->getId(), $userId, $userGroupId);
 
         // Send email notification
-        $mailable = new UserRoleEndNotify($context, $userGroup);
+        $mailable = new UserRoleEndNotify($context, Repo::userGroup()->get($userGroupId));
```
taslangraham, line 324: Why do we need to fetch for the user group here instead of using the data `$userGroup` variable?

classes/invitation/invitations/userRoleAssignment/helpers/UserGroupHelper.php
```diff
@@ -40,8 +40,8 @@ public static function fromArray(array $data): self
         return new self(
             $data['userGroupId'],
             $data['masthead'],
-            $data['dateStart'],
-            $data['dateEnd'] ?? null
+            self::formatDate($data['dateStart']),
+            $data['dateEnd'] ? self::formatDate($data['endDate']) : null
```
taslangraham:
```suggestion
            $data['dateEnd'] ? self::formatDate($data['dateEnd']) : null
```

classes/invitation/invitations/userRoleAssignment/helpers/UserGroupHelper.php
```diff
@@ -50,8 +50,18 @@ public static function fromUserUserGroup(UserUserGroup $userUserGroup): self
         return new self(
             $userUserGroup->userGroupId,
             $userUserGroup->masthead,
-            $userUserGroup->dateStart,
-            $userUserGroup->dateEnd
+            self::formatDate($userUserGroup->dateStart),
+            self::formatDate($userUserGroup->dateEnd)
```
taslangraham: Maybe it's worth adding a safety check on `dateEnd` here as well?

taslangraham (changes requested):
Just a few questions here and there, and a suggestion, but looks good overall @ewhanson.

api/v1/users/PKPUserController.php
ewhanson: Since the original `$userGroup` variable is from the lazy collection, it doesn't have the settings table info filled in, so it needs to be manually fetched to populate those fields. This will be the case for any models fetch in a lazy collection.

classes/invitation/invitations/userRoleAssignment/helpers/UserGroupHelper.php
ewhanson: Yup, good idea!

Issue #11461, opened 2025-06-06
Consider user_user_groups masthead in the user API


bozana:
Users API (api/v1/users) does not consider masthead column from the user_user_groups table i.e. if the user has agreed to be displayed on the masthead for a specific role.
This leads to the problem that all existing user roles are displayed as "Do not appear on masthead" when an invitation is started for an existing user: Invite to a role > search for an existing user by email for example > see the User Group/Role table displayed.
This issue is reported here https://github.com/pkp/pkp-lib/issues/11424, under number 2.

Steps to reproduce:

1. Create a new user using invite to role feature.
2. Add a few roles and select combination of Appear/Not appear on masthead (make sure to remember the selections)
3. Send the invite
4. Accept the invite email
5. Now try to invite this same user with invite to role feature
6. After it finds the user Observe the table with all the previous roles.

Expected: All roles should match the current Edit User Role Table
Actual: All of the roles have "Does not appear on the masthead"

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/11462
- https://github.com/pkp/ojs/pull/4913 (only submodule update)
- https://github.com/pkp/omp/pull/2018 (only submodule update)
- https://github.com/pkp/ops/pull/1004 (only submodule update)

bozana:
Hi @taslangraham, could you maybe review the PR above?
Thanks a lot!

taslangraham:
@bozana sure thing.

taslangraham:
Approved @bozana

bozana:
Hi @Tribunal33 
This is all merged and could be doubled checked by you.
Thanks a lot!

Tribunal33:
Thanks @bozana it lines up now.  Even existing users have matching mastehead option.  Not sure why these tables have the roles sorted differently between edit user and Invite existing user (see screenshot).

![Image](https://github.com/user-attachments/assets/805a8c79-f4e6-46c6-b9bd-6b99e5a393dd)

PR #11462, opened 2025-06-06
pkp/pkp-lib#11461 consider user_user_groups masthead in the users API

bozana:
s. https://github.com/pkp/pkp-lib/issues/11461


taslangraham (approved):
Reviewed and tested. Works as expected.

PR #11467, opened 2025-06-09
Add default country for contributors

thuanhuynhh:
After setting up, when adding a new contributor:
The Country field will automatically default to the country set in the Journal settings
Users can still change the country if needed
Benefits:
- Save time when adding contributors
- Ensure data consistency
- Reduce errors when users forget to select a country
- Compatible with both new and old forms of the system

Feature requests: https://forum.pkp.sfu.ca/t/setting-a-default-contributor-country-in-ojs-3-4-submission-wizard/95984/3

### DIFF
--- a/classes/components/forms/publication/ContributorForm.php
+++ b/classes/components/forms/publication/ContributorForm.php
@@ -90,6 +90,7 @@ class ContributorForm extends FormComponent
                 'label' => __('common.country'),
                 'options' => $countries,
                 'isRequired' => true,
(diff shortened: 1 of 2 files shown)


asmecher:
@thuanhuynhh, thanks for the well written proposal, but we're hesitant to set UI defaults like this because it risks poisoning the data with incorrect assumptions when users don't correct it. If this is important to your journal, it might be better to maintain it as a modification, or perhaps write a plugin to modify the default behaviour.

PR #11473, opened 2025-06-09
add a new property to allow links to open on new tabs

Godoy0722:
**Description**: This PR adds a new property to HTMLPurifier that allows users to open links in new tabs. We discovered this issue when clients reported that saved settings (like Journal Summary with links configured to open in new tabs) were being stripped of their target attributes, causing all links to open in the same tab instead of respecting the original configuration.


Godoy0722:
Hi @asmecher ! Could you take a look at this when you get a chance? We have a client asking about this minor tweak, and I'll update their OJS once it's merged. Thanks!

asmecher:
Looks good, thanks, @Godoy0722! Could you forward-port this to `stable-3_5_0` and `main` too?

Godoy0722:
For sure! I'll make it and let you know.

PR #11475, opened 2025-06-09
Fix a tweak about tabla de contenido

Godoy0722:
**Description**: this minor tweak was figured out by a client and they asked for this update.

### DIFF
--- a/locale/es/common.po
+++ b/locale/es/common.po
@@ -1311,7 +1311,7 @@ msgid "help.help"
 msgstr "Ayuda"
 
 msgid "help.toc"
-msgstr "Tabla de contenidos"
+msgstr "Tabla de contenido"
 
 msgid "help.previous"
 msgstr "Anterior"


Godoy0722:
Hello @asmecher ! This PR addresses a minor issue with a Spanish translation that was reported. Thanks for reviewing!

marcbria:
Hello, @asmecher and @Godoy0722.

We finally had the DIG-ES meeting today and discussed this request.
I am speaking on their behalf to say that we are sorry, but we have to reject this PR.

The reasons?

- Both expressions are correct in Spanish and are the most commonly used forms in Spain and Mexico.
- The translation prior to the PR, ‘Tablas de contenidos’ (plural), has been in the ES translation glossary for years.
- If we accept this PR, the translation will be inconsistent between tools (OMP), screenshots, and documentation.
- If we accept the change, some OJS will include the plural form and others the singular.
- This client requires this change, while others will require the opposite, as both translations are correct.

Therefore, for all these reasons, we believe it is best NOT to change this string and to keep it as it was until now.

The same applies to PR https://github.com/pkp/pkp-lib/pull/49178

asmecher:
Thanks, @marcbria!

Thanks for the PR, @Godoy0722; if it's crucial to the client, maybe it's best kept as a local change (or using the Custom Locale plugin).

Issue #11437, opened 2025-05-29
Reviewers are always displayed on masthead


bozana:
Currently:
- journal manager can choose if a user, that is being invited to a reviewer role, will appear on the masthead or not. 
- reviewers will always be displayed on the masthead if they completed a review in the previous year.
This is confusing for journal managers and reviewers, so this should be consistent.

There are two possible solutions:
1) To treat reviewers in the same way as other roles, so that they can agree or disagree to be displayed. 
Pros: This solution does not introduce special cases that needs to be considered as such in several parts of the code. It is also cleaner and better to understand.
Note: Currently the reviewers can be added also via review stage (Add Reviewer: Create New Reviewer and Enroll Existing User), and in the admin UserForm. So the masthead setting should be added to these parts of the code as well.
2) To always provide only the option "Appear on masthead" for reviewer roles in the invitations.
Pros: This way it is clear to journal managers, editors and reviewers, that they will always be displayed on masthead. This way we ignore what is in the DB table user_user_groups for reviewer roles, so assigning reviewer roles via review stage (Add Reviewer: Create New Reviewer and Enroll Existing User) is not a problem.
Cons: This solution introduces special cases in additional part of the code.

According to the Journal Integrity Initiative, it is very important for active reviewers to be listed, for the reviewer approvals to be transparent to the outsiders -- reviewer invisibility would make it hard to detect some malicious behaviors/approvals by outsiders.
Thus, the right solution would be number 2.

PRs:
~~Solution 1:~~
~~stable-3_5_0:~~
~~- https://github.com/pkp/pkp-lib/pull/11471~~

Solution2:
stable-3_5_0:
- https://github.com/pkp/ui-library/pull/643
- https://github.com/pkp/pkp-lib/pull/11492
- https://github.com/pkp/ojs/pull/4933 (only submodules updates)
- https://github.com/pkp/omp/pull/2029 (only submodules updates)
- https://github.com/pkp/ops/pull/1012 (only submodules updates)

main:
- https://github.com/pkp/ui-library/pull/648
- https://github.com/pkp/pkp-lib/pull/11528

bozana:
Hi @taslangraham, could you please review the PR above?
Thanks a lot!

taslangraham:
Hey @bozana, I'm off today, but will a look tomorrow when I'm back online.

bozana:
[...]

PR #11492, opened 2025-06-11
pkp/pkp-lib#11437 Reviewers are always displayed on masthead

bozana:
s. https://github.com/pkp/pkp-lib/issues/11437

### DIFF
--- a/classes/invitation/stepTypes/SendInvitationStep.php
+++ b/classes/invitation/stepTypes/SendInvitationStep.php
@@ -175,15 +175,6 @@ class SendInvitationStep extends InvitationStepTypes
      */
     private function getAllUserGroups(Context $context): array
     {
-        $allUserGroups = [];
-        $userGroups = UserGroup::withContextIds([$context->getId()])->get();
-        foreach ($userGroups as $userGroup) {
-            $allUserGroups[] = [
-                'value' => (int) $userGroup->id,
-                'label' => $userGroup->getLocalizedData('name'),
-                'disabled' => false
-            ];
-        }
-        return $allUserGroups;
+        return UserGroup::withContextIds([$context->getId()])->get()->all();
     }
 }
--- a/classes/invitation/stepTypes/SendInvitationStep.php
+++ b/classes/invitation/stepTypes/SendInvitationStep.php
@@ -15,7 +15,6 @@
 namespace PKP\invitation\stepTypes;
 
 use APP\core\Application;
-use APP\facades\Repo;
 use Exception;
 use PKP\components\forms\invitation\UserDetailsForm;
 use PKP\context\Context;
@@ -25,7 +24,6 @@ use PKP\invitation\sections\Form;
 use PKP\invitation\sections\Sections;
 use PKP\invitation\steps\Step;
 use PKP\mail\mailables\UserRoleAssignmentInvitationNotify;
-use PKP\security\Role;
 use PKP\user\User;
 use PKP\userGroup\UserGroup;
 use stdClass;
@@ -111,7 +109,6 @@ class SendInvitationStep extends InvitationStepTypes
             [
                 'validateFields' => [],
                 'userGroups' => $this->getAllUserGroups($context),
-                'reviewerUserGroupIds' => Repo::userGroup()->getArrayIdByRoleId(Role::ROLE_ID_REVIEWER, $context->getId())
             ]
         );
         $step = new Step(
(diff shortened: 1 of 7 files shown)


classes/invitation/stepTypes/SendInvitationStep.php
```diff
@@ -107,7 +110,8 @@ private function invitationDetailsForm(Context $context): stdClass
             ),
             [
                 'validateFields' => [],
-                'userGroups' => $this->getAllUserGroups($context)
+                'userGroups' => $this->getAllUserGroups($context),
+                'reviewerUserGroupIds' => Repo::userGroup()->getArrayIdByRoleId(Role::ROLE_ID_REVIEWER, $context->getId())
```
taslangraham: Suggestion: Since you're already fetching all user groups with `getAllUserGroups($context)`, you could filter that result for `reviewerUserGroupIds` instead of making another trip to the DB.

controllers/grid/settings/user/form/UserForm.php
```diff
@@ -111,15 +111,14 @@ public function display($request = null, $template = null)
...
-                $defaultMastheadUserGroups[(int) $userGroup->id] = $userGroup->getLocalizedData('name');
-            }
+            $defaultMastheadUserGroups[(int) $userGroup->id] = $userGroup->getLocalizedData('name');
         }
 
         $templateMgr->assign([
             'allUserGroups' => $allUserGroups,
             'assignedUserGroups' => array_map(intval(...), $this->getData('userGroupIds')),
             'defaultMastheadUserGroups' => $defaultMastheadUserGroups,
+            'reviewerUserGroupIds' => Repo::userGroup()->getArrayIdByRoleId(Role::ROLE_ID_REVIEWER, $contextId)
```
taslangraham: Similar comment as above about filtering for `reviewerUserGroupIds` from the existing list in `$userGroups`

templates/controllers/grid/users/reviewer/form/createReviewerForm.tpl
```diff
@@ -67,6 +67,10 @@
 		{fbvElement type="text" multilingual="true" name="affiliation" id="affiliation" value=$affiliation size=$fbvStyles.size.LARGE}
 	{/fbvFormSection}
 
+	{fbvFormSection title="manager.setup.masthead" list=true}
+		{fbvElement type="checkbox" name="masthead" id="masthead" checked=true label="invitation.masthead.show" translate="true" disabled="true"}
```
taslangraham: Since this field isn’t used during form submission and serves only a visual purpose, I think it makes sense to remove the `name` attribute so that this is not sent with the form data.

templates/controllers/grid/users/reviewer/form/enrollExistingReviewerForm.tpl
```diff
@@ -50,6 +50,10 @@
 		}
 	{/fbvFormSection}
 
+	{fbvFormSection title="manager.setup.masthead" list=true}
+		{fbvElement type="checkbox" name="masthead" id="masthead" checked=true label="invitation.masthead.show" translate="true" disabled="true"}
```
taslangraham: Same comment about dropping the name attribute

classes/invitation/stepTypes/SendInvitationStep.php
```diff
@@ -171,15 +175,6 @@ private function invitationInvitedEmail(Context $context): stdClass
...
-        $userGroups = UserGroup::withContextIds([$context->getId()])->get();
-        foreach ($userGroups as $userGroup) {
-            $allUserGroups[] = [
-                'value' => (int) $userGroup->id,
-                'label' => $userGroup->getLocalizedData('name'),
-                'disabled' => false
-            ];
-        }
-        return $allUserGroups;
+        return UserGroup::withContextIds([$context->getId()])->get()->all();
```
taslangraham: Checking if the `all()` is necessary here?

taslangraham (changes requested):
Just a few questions and suggestions, but overall looks good @bozana

classes/invitation/stepTypes/SendInvitationStep.php
taslangraham: Thoughts on using `Repo::userGroup()->getSchemaMap()->mapMany()` here? I think this would remove the need to do `Object.values(userGroups)` when looping through the userGroups in ui-library

bozana: For some reason, when provided here as an array, it is not an array in the UserInvitationDetailsStep.vue

PR #11498, opened 2025-06-12
add a new property to allow links to open on new tabs

Godoy0722:
**Description**: This PR adds a new property to HTMLPurifier that allows users to open links in new tabs. We discovered this issue when clients reported that saved settings (like Journal Summary with links configured to open in new tabs) were being stripped of their target attributes, causing all links to open in the same tab instead of respecting the original configuration.


Godoy0722:
Hello @asmecher . That's the PR to add the target _blank also on main branch.

PR #11507, opened 2025-06-13
Credit Roles

jyhein:
- Adds Credit Roles and degrees of contribution to contributor form and landing page
- Adds 2 db tables
- Downloads translations

### DIFF
--- a/classes/author/Author.php
+++ b/classes/author/Author.php
@@ -286,4 +286,20 @@ class Author extends Identity
             $this->getLocalizedAffiliationNames($preferredLocale)
         );
     }
+
+    /**
+     * Get contrubutor credit roles and degrees
+     */
+    public function getCreditRoles(): array
+    {
+        return $this->getData('creditRoles') ?? [];
+    }
+
+    /**
+     * Set contrubutor credit roles and degrees.
+     */
+    public function setCreditRoles(?array $creditRoles): void
+    {
+        $this->setData('creditRoles', $creditRoles);
+    }
 }
--- a/classes/author/creditContributorRole/CreditContributorRole.php
+++ b/classes/author/creditContributorRole/CreditContributorRole.php
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * @file lib/pkp/classes/author/creditContributorRole/CreditContributorRole.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class CreditContributorRole
+ *
+ * @brief CreditContributorRole model class
+ */
+
+namespace PKP\author\creditContributorRole;
+
+use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Query\JoinClause;
+
+class CreditContributorRole extends Model
+{
+    use HasCamelCasing;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$table
+     */
+    protected $table = 'credit_contributor_roles';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$primaryKey
+     */
+    protected $primaryKey = 'credit_contributor_role_id';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\GuardsAttributes::$guarded
+     */
+    protected $guarded = [
+        'creditRoleId',
+        'contributorRoleId',
+    ];
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasTimestamps::$timestamps
+     */
+    public $timestamps = false;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasAttributes::casts
+     */
+    protected function casts(): array
+    {
+        return [
+            'contributor_id' => 'integer',
+            'credit_role_id' => 'integer',
+            'credit_degree' => 'string',
+            'contributor_role_id' => 'integer',
+        ];
+    }
+
+    /**
+     * Accessor and Mutator for primary key => id
+     */
+    protected function id(): Attribute
+    {
+        return Attribute::make(
+            get: fn ($value, $attributes) => $attributes[$this->primaryKey] ?? null,
+            set: fn ($value) => [$this->primaryKey => $value],
+        )->shouldCache();
+    }
+
+    /**
+     * Scope a query to only include credit and contributor roles with a specific contributor id.
+     */
+    public function scopeWithContributorId(Builder $query, int $contributorId): Builder
+    {
+        return $query->where('contributor_id', $contributorId);
+    }
+
+    /**
+     * Scope a query to only include credit roles with a specific contributor id.
+     */
+    public function scopeWithCreditRoles(Builder $query): Builder
+    {
+        return $query
+            ->whereNotNull('credit_contributor_roles.credit_role_id')
+            ->leftJoin('credit_roles as cr', fn (JoinClause $join) => $join
+                ->on('credit_contributor_roles.credit_role_id', '=', 'cr.credit_role_id'));
+    }
+}
--- a/classes/author/creditRole/CreditRole.php
+++ b/classes/author/creditRole/CreditRole.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * @file lib/pkp/classes/author/creditRole/CreditRole.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class CreditRole
+ *
+ * @brief CreditRole model class
+ */
+
+namespace PKP\author\creditRole;
+
+use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Builder;
+
+class CreditRole extends Model
+{
+    use HasCamelCasing;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$table
+     */
+    protected $table = 'credit_roles';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$primaryKey
+     */
+    protected $primaryKey = 'credit_role_id';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasTimestamps::$timestamps
+     */
+    public $timestamps = false;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasAttributes::casts
+     */
+    protected function casts(): array
+    {
+        return [
+            'credit_role_id' => 'integer',
+            'credit_role_identifier' => 'string',
+        ];
+    }
+
+    /**
+     * Accessor and Mutator for primary key => id
+     */
+    protected function id(): Attribute
+    {
+        return Attribute::make(
+            get: fn ($value, $attributes) => $attributes[$this->primaryKey] ?? null,
+            set: fn ($value) => [$this->primaryKey => $value],
+        )->shouldCache();
+    }
+
+    /**
+     * Scope a query to include specific credit roles.
+     */
+    public function scopeWithCreditRoleIdentifiers(Builder $query, array $creditRoles): Builder
+    {
+        return $query->whereIn('credit_role_identifier', $creditRoles);
+    }
+}
--- a/classes/author/creditRole/Repository.php
+++ b/classes/author/creditRole/Repository.php
@@ -1,0 +1,120 @@
+<?php
+
+/**
+ * @file classes/author/creditRole/Repository.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Repository
+ *
+ * @brief A repository to manage actions related to credit roles
+ */
+
+namespace PKP\author\creditRole;
+
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Cache;
+use PKP\author\creditRole\CreditRoleDegree;
+use APP\core\Application;
+use PKP\core\Core;
+use PKP\facades\Locale;
+
+class Repository {
+    /** @var string Max lifetime for the cache */
+    protected const MAX_CACHE_LIFETIME = '1 year';
+    protected array $localeMapping = [
+        'cs' => 'cz',
+        'da' => 'dk',
+        'zh_Hans' => 'cn',
+        'el' => 'gr',
+        'nb_NO' => 'no_bk',
+        'nn' => 'no_nn',
+        'zh_Hant' => 'tc',
+    ];
+    // One of the credit role file locales
+    protected string $defaultLocale = 'en';
+    protected array $creditRoleTerms = [];
+
+    /**
+     * Get credit role terms and degrees of contribution
+     */
+    public function getTerms(?string $locale = null): array
+    {
+        $locale ??= Locale::getLocale();
+        $key = __METHOD__ . static::MAX_CACHE_LIFETIME;
+
+        if (!$this->creditRoleTerms) {
+            $this->creditRoleTerms = Cache::get($key) ?? [];
+        }
+
+        if (!array_key_exists($locale, $this->creditRoleTerms)) {
+            $this->getLocalizedTerms($locale, $key);
+        }
+
+        return $this->creditRoleTerms[$locale] ?? $this->creditRoleTerms[$this->defaultLocale] ?? [];
+    }
+
+    /**
+     * Type of roles in an associative array URI => Term
+     * Degrees in an associative array db-value => translation
+     */
+    protected function getLocalizedTerms(string $locale, string $key): void
+    {
+        $expiration = \DateInterval::createFromDateString(static::MAX_CACHE_LIFETIME);
+
+        $setRoles = fn (array $roles, string $localeKey): array => [
+            'roles' => $roles,
+            'degrees' => [
+                CreditRoleDegree::NULL->getLabel() => '',
+                CreditRoleDegree::LEAD->getLabel() => __('submission.submit.creditRoles.degrees.lead', [], $localeKey),
+                CreditRoleDegree::EQUAL->getLabel() => __('submission.submit.creditRoles.degrees.equal', [], $localeKey),
+                CreditRoleDegree::SUPPORTING->getLabel() => __('submission.submit.creditRoles.degrees.supporting', [], $localeKey),
+            ],
+        ];
+        $getJson = fn (string $file): array => is_array($json = json_decode(file_get_contents($file) ?: "", true)) ? $json : [];
+        $getRoles = fn (array $json): array => Arr::map($json['translations'] ?? [], fn (array $items) => $items['name']);
+
+        // Init with ui locales and default locale
+        if (!$this->creditRoleTerms) {
+            $this->creditRoleTerms = Cache::remember($key, $expiration, fn () =>
+                collect(Application::get()->getRequest()->getContext()->getSupportedLocales())
+                    ->push($this->defaultLocale)
+                    ->unique()
+                    ->mapWithKeys(fn (string $l): array => [$l => ($filename = $this->getFileName($l)) ? $setRoles($getRoles($getJson($filename)), $l) : null])
+                    ->toArray()
+            );
+        }
+
+        // Add new locale if not in array
+        if (!array_key_exists($locale, $this->creditRoleTerms)) {
+            $filename = $this->getFileName($locale);
+            $this->creditRoleTerms[$locale] = $filename ? $setRoles($getRoles($getJson($filename)), $locale) : null;
+            Cache::put($key, $this->creditRoleTerms, $expiration);
+        }
+    }
+
+    /**
+     * Get file name of the localized terms
+     */
+    protected function getFileName(string $locale): ?string
+    {
+        static $filenames;
+        $prefix = Core::getBaseDir() . '/' . PKP_LIB_PATH . "/lib/creditRoles/translations/";
+        $suffix = '.json';
+        $filenames ??= collect(glob("{$prefix}*{$suffix}"));
+        // Determine the language we're looking for from $locale
+        $language = \Locale::getPrimaryLanguage($locale);
+        // Get a list of available options from the filesystem and return file name
+        return $filenames
+            ->first(fn (string $filename) =>
+                // 1. Look for an exact match and return it.
+                ($filename === "{$prefix}{$locale}{$suffix}") ||
+                // 2. Look in the preference list for a preferred fallback.
+                (isset($this->localeMapping[$locale]) && $filename === "{$prefix}{$this->localeMapping[$locale]}{$suffix}") ||
+                // 3. Find the first match by language.
+                (strpos($filename, "{$prefix}{$language}{$suffix}") === 0 || strpos($filename, "{$prefix}{$language}_") === 0)
+            );
+    }
+}
--- a/classes/author/maps/Schema.php
+++ b/classes/author/maps/Schema.php
@@ -103,6 +103,9 @@ class Schema extends \PKP\core\maps\Schema
                 case 'hasVerifiedOrcid':
                     $output[$prop] = $item->hasVerifiedOrcid();
                     break;
+                case 'creditRoles':
+                    $output[$prop] = $item->getCreditRoles();
+                    break;
                 case 'orcidDisplayValue':
                     $output[$prop] = $item->getOrcidDisplayValue();
                     break;
--- a/classes/facades/Repo.php
+++ b/classes/facades/Repo.php
@@ -27,6 +27,8 @@ namespace PKP\facades;
 use PKP\affiliation\Repository as AffiliationRepository;
 use PKP\announcement\Repository as AnnouncementRepository;
 use PKP\author\Repository as AuthorRepository;
+use PKP\author\creditContributorRole\Repository as CreditContributorRoleRepository;
+use PKP\author\creditRole\Repository as CreditRoleRepository;
 use PKP\category\Repository as CategoryRepository;
 use PKP\controlledVocab\Repository as ControlledVocabRepository;
 use PKP\decision\Repository as DecisionRepository;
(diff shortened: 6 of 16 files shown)


classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+        $this->setData('creditRoles', $creditRoles);
+    }
+
+    /**
+     * Get credit role terms and degrees of contribution
+     */
+    public static function getCreditRoleTerms(?string $locale = null): array
+    {
+        $locale ??= Locale::getLocale();
+        static $key = __METHOD__ . static::MAX_CACHE_LIFETIME;
```
bozana: This static in compiled-time constants is not allowed in PHP 8.2

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+        // One of the credit role file locales
+        static $defaultLocale = 'en';
+        static $creditRoleTerms = Cache::get($key)[0] ?? [];
```
bozana: This static in compiled-time constants is not allowed in PHP 8.2

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+
+    /**
+     * Get credit role terms and degrees of contribution
+     */
+    public static function getCreditRoleTerms(?string $locale = null): array
+    {
+        $locale ??= Locale::getLocale();
+        static $key = __METHOD__ . static::MAX_CACHE_LIFETIME;
+        // One of the credit role file locales
+        static $defaultLocale = 'en';
```
bozana: Maybe without `static`?

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+            'cz' => ['cs'],
+            'fr' => ['fr', 'fr_CA'],
+            'cn' => ['zh_Hans'],
+            'gr' => ['el'],
+            'no_bk' => ['nb_NO'],
+            'no_nn' => ['nn'],
+            'tc' => ['zh_Hant'],
+        ];
+        static $creditRoleLocales = Cache::get($key)[1] ?? [];
+        static $expiration = \DateInterval::createFromDateString(static::MAX_CACHE_LIFETIME);
```
bozana: The both `static` in compiled-time constants are not allowed in PHP 8.2

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+
+        static $setRoles = fn (array $roles, string $localeKey): array => [
```
bozana: `static` again

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+            'roles' => $roles,
+            'degrees' => [
+                CreditRoleDegree::NO_DEGREE->value => '',
+                CreditRoleDegree::LEAD->value => __('submission.submit.creditRoles.degrees.lead', [], $localeKey),
+                CreditRoleDegree::EQUAL->value => __('submission.submit.creditRoles.degrees.equal', [], $localeKey),
+                CreditRoleDegree::SUPPORTING->value => __('submission.submit.creditRoles.degrees.supporting', [], $localeKey),
+            ],
+        ];
+        static $getJson = fn (string $file): array => is_array($json = json_decode(file_get_contents($file) ?: "", true)) ? $json : [];
+        static $getRoles = fn (array $json): array => Arr::map($json['translations'] ?? [], fn (array $items) => $items['name']);
```
bozana: These `static` also

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+        return $creditRoleTerms[$locale] ?? $creditRoleTerms[$defaultLocale] ?? [];
+    }
+
+    /**
+     * Type of roles in an associative array URI => Term
+     * Degrees in an associative array db-key => translation
+     */
+    protected static function getLocalizedCreditRoleTerms(string $locale, array $creditRoleTerms, string $defaultLocale, string $key): array
+    {
+        static $localeMapping = [
```
bozana: Maybe without `static`?

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+                CreditRoleDegree::EQUAL->value => __('submission.submit.creditRoles.degrees.equal', [], $localeKey),
+                CreditRoleDegree::SUPPORTING->value => __('submission.submit.creditRoles.degrees.supporting', [], $localeKey),
+            ],
+        ];
+        static $getJson = fn (string $file): array => is_array($json = json_decode(file_get_contents($file) ?: "", true)) ? $json : [];
+        static $getRoles = fn (array $json): array => Arr::map($json['translations'] ?? [], fn (array $items) => $items['name']);
+
+        if (!$creditRoleTerms) {
+            [$creditRoleTerms, $creditRoleLocales] = Cache::remember($key, $expiration, function () use ($localeMapping, $defaultLocale, $setRoles, $getJson, $getRoles): array {
+                $folder = dirname(__FILE__, 3) . "/lib/creditRoles/";
```
bozana: Maybe to use `Core::getBaseDir() . '/' . PKP_LIB_PATH`

classes/author/creditContributorRoles/CreditContributorRoles.php
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+namespace PKP\author\creditContributorRoles;
+
+use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Builder;
+
+class CreditContributorRoles extends Model
```
bozana: Maybe to call the Class and the folder using singular? -- I think it is so everywhere -- to have CreditContributorRole (class name) and creditContributorRole (folder name)
Then also all the calls CreditContributorRoles:: should be changed to CreditContributorRole::

classes/author/creditRoles/CreditRoles.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+namespace PKP\author\creditRoles;
+
+use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Builder;
+
+class CreditRoles extends Model
```
bozana: Maybe also here, to use singular: CreditRole (class) and creditRole (folder)

classes/author/creditContributorRoles/Repository.php
```diff
@@ -0,0 +1,103 @@
...
+                ['contributor_id' => $contributorId, 'credit_role_id' => $creditRoles[$role]],
+                ['credit_degree' => $rd['degree'] ?? null]
+            );
+        }
+    }
+
+    /**
+     * Delete all contributor's credit roles and contributor roles.
+     */
+    public function deleteByContributorId(int $contributorId): void
```
bozana: I think this function is not used?

classes/author/creditContributorRoles/Repository.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        CreditContributorRoles::query()->withContributorId($contributorId)->delete();
+    }
+
+    /**
+     * Delete all contributor's credit roles.
+     */
+    public function deleteCreditRolesByContributorId(int $contributorId): void
```
bozana: this function is also not used?

classes/author/creditContributorRoles/Repository.php
```diff
@@ -0,0 +1,103 @@
...
+        CreditContributorRoles::query()
+            ->withContributorId($contributorId)
+            ->whereNotNull('credit_role_id')
+            ->delete();
+    }
+
+    /**
+     * Get all contributor's credit roles and contributor roles.
+     */
+    public function getByContributorId(int $contributorId): array
```
bozana: this as well?

classes/author/creditRoles/CreditRoleDegree.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+/**
+ * @file classes/author/creditRoles/CreditRoleDegree.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
```
bozana: 2025

classes/author/creditRoles/Repository.php
```diff
@@ -0,0 +1,17 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Repository
+ *
+ * @brief A repository to manage actions related to credit roles
+ */
+
+namespace PKP\author\creditRoles;
+
+class Repository {}
```
bozana: Then no need for this class? I.e. maybe to move those two functions from Author here?

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Credit roles listing table
+        Schema::create('credit_roles', function (Blueprint $table) {
+            $table->comment('The list of the CRediT Roles');
+            $table->bigInteger('credit_roles_id')->autoIncrement();
+            $table->bigInteger('credit_role_id');
```
bozana: I think we need only one, credit_role_id, and it can be with autoIncrement()

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+            $table->string('credit_role_identifier', 255);
+            $table->unique(['credit_role_id'], 'credit_role_id_unique');
```
bozana: I think credit_role_identifier should be unique, no?

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+        });
+
+        // Load en json, and fill table with values
+        $creditRoles = json_decode(file_get_contents(dirname(__FILE__, 5) . '/lib/creditRoles/en.json') ?: "", true) ?: [];
```
bozana: Maybe to use Core::getBaseDir() . '/' . PKP_LIB_PATH

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+        $creditRolesData = Arr::map(array_keys($creditRoles['translations'] ?? []), fn (string $role, int $i): array => ['credit_role_id' => $i, 'credit_role_identifier' => $role]);
```
bozana: Maybe to leave credit_role_id to autoincrement? The URIs in en.json seem to be alphabetically ordered. If for example a new role would be added to the standard, we could not keep the same indexes any more...
What was the motivation to have it like this?

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+            $table->bigInteger('credit_contributor_role_id')->autoIncrement();
+            $table->bigInteger('contributor_id');
+            $table->bigInteger('credit_role_id')->nullable();
+            $table->string('credit_degree', 255)->nullable();
+            $table->bigInteger('contributor_role_id')->nullable();
+            $table->foreign('contributor_id', 'contributor_id_author_id_foreign')->references('author_id')->on('authors')->onDelete('cascade');
+            $table->foreign('credit_role_id', 'credit_role_id_foreign')->references('credit_role_id')->on('credit_roles')->onDelete('cascade');
+            $table->unique(['contributor_id', 'credit_role_id'], 'contributor_id_credit_role_id_unique');
+            $table->unique(['contributor_id', 'contributor_role_id'], 'contributor_id_contributor_role_id_unique');
+            $table->enum('degree', [
```
bozana: This should be credit_degree, no? And it should also contain NO_DEGREE, correct?

classes/migration/install/SubmissionsMigration.php
```diff
@@ -127,6 +128,47 @@ public function up(): void
             $table->unique(['author_id', 'locale', 'setting_name'], 'author_settings_unique');
         });
 
+        // Credit roles listing table
```
bozana, line 132: See comments for I857_CreditRoles.php -- they would apply also here.

classes/author/maps/Schema.php
```diff
@@ -103,6 +103,9 @@ protected function mapByProperties(array $props, Author $item): array
                 case 'hasVerifiedOrcid':
                     $output[$prop] = $item->hasVerifiedOrcid();
                     break;
+                case 'creditRoles':
+                    $output[$prop] = $item->getCreditRoles($prop);
```
bozana: No need for parameter $prop for this function, no?

classes/dev/ComposerScript.php
```diff
@@ -34,6 +34,41 @@ public static function isoFileCheck(): void
         }
     }
 
+    /**
+     * A post-install-cmd custom composer script that
+     * creates folder creditRoles and downloads Credit Role translations to json files.
+     */
+    public static function creditRolesFilesDownload(): void
```
bozana: I think we can have lib/creditTranslations as a submodule in the pkp-lib, pointing to https://github.com/contributorshipcollaboration/credit-translation, instead of this composer sript. This way we do not need to maintain this hard coded list of locales.

classes/author/Author.php
```diff
@@ -286,4 +293,103 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+     */
+    public function setCreditRoles(?array $creditRoles): void
+    {
+        $this->setData('creditRoles', $creditRoles);
+    }
+
+    /**
+     * Get credit role terms and degrees of contribution
+     */
+    public static function getCreditRoleTerms(?string $locale = null): array
```
bozana: Maybe to move these two functions, getCreditRoleTerms and getLocalizedCreditRoleTerms, to creditRoles/Repository.php ? -- They do not have anything to do with Author, right?

classes/author/creditContributorRoles/Repository.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Get all contributor's credit roles.
+     */
+    public function getCreditRolesByContributorId(int $contributorId): array
+    {
+        return CreditContributorRoles::query()
+            ->withContributorId($contributorId)
+            ->whereNotNull('credit_contributor_roles.credit_role_id')
+            ->leftJoin('credit_roles as cr', fn (JoinClause $join) => $join
+                ->on('credit_contributor_roles.credit_role_id', '=', 'cr.credit_role_id'))
```
bozana: Maybe to move this whereNotNull and leftJoin to CreditContributorRoles.php, so that we call it here:
->withCreditRoles()

bozana (changes requested):
Hi @jyhein, I went through the code and left some comments. Thanks a lot!

jardakotesovec:
@jyhein @bozana Any idea why I am getting error when running migration? Maybe its postgres specific issue?

```
10:ojs-main jarda$ php lib/pkp/tools/migration.php \\PKP\\migration\\upgrade\\v3_6_0\\I857_CreditRoles up
PHP Deprecated:  PKP\core\DataObject::getData(): Implicitly marking parameter $locale as nullable is deprecated, the explicit nullable type must be used instead in /Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/classes/core/DataObject.php on line 83
PHP Deprecated:  PKP\plugins\ThemePlugin::getLocalizedOption(): Implicitly marking parameter $preferredLocale as nullable is deprecated, the explicit nullable type must be used instead in /Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/classes/plugins/ThemePlugin.php on line 499
PHP Deprecated:  PKP\plugins\ThemePlugin::getLocalizedOption(): Implicitly marking parameter $selectedLocale as nullable is deprecated, the explicit nullable type must be used instead in /Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/classes/plugins/ThemePlugin.php on line 499
PHP Deprecated:  PKP\submission\DAO::getByPubId(): Implicitly marking parameter $contextId as nullable is deprecated, the explicit nullable type must be used instead in /Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/classes/submission/DAO.php on line 165
PHP Warning:  file_get_contents(/Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/lib/creditRoles/en.json): Failed to open stream: No such file or directory in /Users/jarda/git/pkp/pkp-main/ojs-main/lib/pkp/classes/migration/upgrade/v3_6_0/I857_CreditRoles.php on line 41
ERROR: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "NOT"
LINE 4:             CHECK (`credit_role_id` IS NOT NULL AND `contrib...
                                               ^ (Connection: pgsql, SQL: 
            ALTER TABLE credit_contributor_roles
            ADD CONSTRAINT check_xor_credit_contributor_role
            CHECK (`credit_role_id` IS NOT NULL AND `contributor_role_id` IS NULL OR `contributor_role_id` IS NOT NULL AND `credit_role_id` IS NULL)
        )
```

bozana:
Hi @jardakotesovec, 
Could you please change that line for you locally -- to be able to test it -- like this:
`CHECK ((credit_role_id IS NOT NULL AND contributor_role_id IS NULL) OR (contributor_role_id IS NOT NULL AND credit_role_id IS NULL))`

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,80 @@
...
+                'EQUAL',
+                'SUPPORTING',
+            ]);
+        });
+
+        // Add contstraint to only allow either credit or contributor role per row
+        DB::statement('
+            ALTER TABLE credit_contributor_roles
+            ADD CONSTRAINT check_xor_credit_contributor_role
+            CHECK (`credit_role_id` IS NOT NULL AND `contributor_role_id` IS NULL OR `contributor_role_id` IS NOT NULL AND `credit_role_id` IS NULL)
```
bozana: it seems this is not the right syntax for PostgreSQL. This maybe works:
`CHECK ((credit_role_id IS NOT NULL AND contributor_role_id IS NULL) OR (contributor_role_id IS NOT NULL AND credit_role_id IS NULL))` for PostgreSQL

classes/author/creditContributorRoles/Repository.php
jyhein: It's there ready when needed.

classes/author/creditContributorRoles/Repository.php
jyhein: It's there ready when needed.

classes/author/creditContributorRoles/Repository.php
jyhein: It's there ready when needed.

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
jyhein: It's there to allow the Uris to be kept in alphabetical order when additions happen

bozana: But the number will eventually change when additions happen. And we point to it in the other table.
I do not think we need to keep the alphabetical order, or? If additions happen, we can just add them to the table, no?

jyhein: ok

classes/author/creditContributorRoles/Repository.php
bozana: I think you can delete them -- they seem not to be needed -- the addCreditRoles function above seems to manage it all, as the only one needed. I do not know the scenario where these would be needed, and if it happens, then we can add them...

bozana: Ah, or do you think you will need them for the coming contributor roles? -- I have forgotten...
Then they can stay... :-)

EDIT: but these are credit contributor roles, so maybe you do not need them for contributor roles... So maybe to delete them :sweat_smile:

classes/author/creditContributorRole/CreditContributorRole.php
```diff
@@ -0,0 +1,93 @@
...
+        return $query->where('contributor_id', $contributorId);
+    }
+
+    /**
+     * Scope a query to only include credit roles with a specific contributor id.
+     */
+    public function scopeWithCreditRoles(Builder $query, int $contributorId): Builder
+    {
+        return $query
+            ->withContributorId($contributorId)
```
bozana: I think you can leave the contributorId parameter here out and call this line below in Repository::getCreditRolesByContributorId().

classes/author/creditRole/CreditRole.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$table
+     */
+    protected $table = 'credit_roles';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$primaryKey
+     */
+    protected $primaryKey = 'credit_roles_id';
```
bozana: maybe to call it `credit_role_id` -- similarly as in every other table...

classes/facades/Repo.php
```diff
@@ -67,6 +69,16 @@ public static function author(): AuthorRepository
         return app(AuthorRepository::class);
     }
 
+    public static function CreditContributorRole(): CreditContributorRoleRepository
```
bozana: should start with small letter, i.e. creditContributorRole?

classes/migration/install/SubmissionsMigration.php
```diff
@@ -127,6 +129,45 @@ public function up(): void
...
+        Schema::create('credit_roles', function (Blueprint $table) {
+            $table->comment('The list of the CRediT Roles');
+            $table->bigInteger('credit_role_id')->autoIncrement();
+            $table->string('credit_role_identifier', 255);
+            $table->unique(['credit_role_identifier'], 'credit_role_identifier_unique');
+        });
+
+        // Load en json, and fill table with values
+        $creditRoles = json_decode(file_get_contents(Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/creditRoles/translations/en.json') ?: "", true) ?: [];
```
bozana: shell we maybe throw exception if the file does not exist? :thinking:

classes/author/creditRole/Repository.php
```diff
@@ -0,0 +1,114 @@
...
+    /** @var string Max lifetime for the cache */
+    protected const MAX_CACHE_LIFETIME = '1 year';
+    protected array $localeMapping = [
+        'cz' => ['cs'],
+        'fr' => ['fr', 'fr_CA'],
+        'cn' => ['zh_Hans'],
+        'gr' => ['el'],
+        'no_bk' => ['nb_NO'],
+        'no_nn' => ['nn'],
+        'tc' => ['zh_Hant'],
```
bozana: Could we maybe implement this mapping in the same way as it is in other places in the code, e.g. here: https://github.com/pkp/citationStyleLanguage/blob/main/CitationStyleLanguagePlugin.php#L1014.
Here it would be something like:
```
public function getCreditRoleLocale(string $locale, string $defaultLocale = 'en'): string
    {
        $prefix = Core::getBaseDir() . '/' . PKP_LIB_PATH . "/lib/creditRoles/translations";
        $suffix = '.json';
        $preferences = [
            'cs' => 'cz',
            'da' => 'dk',
            'zh_Hans' => 'cn',
            'el' => 'gr',
            'nb_NO' => 'no_bk',
            'nn' => 'no_nn',
            'zh_Hant' => 'tc',
        ];
        // Determine the language and region we're looking for from $locale
        $language = \Locale::getPrimaryLanguage($locale);
        $region = \Locale::getRegion($locale) ?? null;
        $localeAndRegion = $language . ($region ? "-{$region}" : '');
        // Get a list of available options from the filesystem.
        $availableLocaleFiles = glob("{$prefix}*{$suffix}");
        // 1. Look for an exact match and return it.
        if (in_array("{$prefix}{$locale}{$suffix}", $availableLocaleFiles)) {
            return $locale;
        }
        // 2. Look in the preference list for a preferred fallback.
        if ($preference = $preferences[$localeAndRegion] ?? false) {
            return $preference;
        }
        // 3. Find the first match by language.
        foreach ($availableLocaleFiles as $filename) {
            if (strpos($filename, "{$prefix}{$language}{$suffix}") === 0 || strpos($filename, "{$prefix}{$language}_") === 0) {
                return substr($filename, strlen($prefix), -strlen($suffix));
            }
        }
        // 4. Use the supplied default.
        return $defaultLocale;
    }

```
That means, we would have a list of preferred locales that will map our Weblate locales to their locales, but also only those fully different ones. fr_CA we would not need to have in the list, for example, because `strpos($filename, "{$prefix}{$language}{$suffix") === 0` would find their fr.json.

classes/migration/upgrade/v3_6_0/I857_CreditRoles.php
```diff
@@ -0,0 +1,79 @@
...
+            $table->foreign('credit_role_id', 'credit_role_id_foreign')->references('credit_role_id')->on('credit_roles')->onDelete('cascade');
+            $table->unique(['contributor_id', 'credit_role_id'], 'contributor_id_credit_role_id_unique');
+            $table->unique(['contributor_id', 'contributor_role_id'], 'contributor_id_contributor_role_id_unique');
+        });
+
+        // Add contstraint to only allow either credit or contributor role per row
+        DB::statement('
+            ALTER TABLE credit_contributor_roles
+            ADD CONSTRAINT check_xor_credit_contributor_role
+            CHECK ((`credit_role_id` IS NOT NULL AND `contributor_role_id` IS NULL) OR (`contributor_role_id` IS NOT NULL AND `credit_role_id` IS NULL))
```
bozana: I think PostgreSQL does not like these inverted commas/quotation marks, it should be without it for PostgreSQL

classes/migration/install/SubmissionsMigration.php
asmecher: Rather, I would suggest falling back on English. Lots of translations won't be present, at least not at first.

bozana: @asmecher, what do you mean? 
We do read en.json to insert the URLs into DB... Or do you mean something else?

asmecher: Sorry, I misunderstood. I thought the exception would be thrown when a non-English translation was expected, but this one is fixed as English.

classes/author/creditRole/Repository.php
```diff
@@ -0,0 +1,121 @@
...
+        $language = \Locale::getPrimaryLanguage($locale);
+        $region = \Locale::getRegion($locale) ?? null;
+        $localeAndRegion = $language . ($region ? "_{$region}" : '');
+        // Get a list of available options from the filesystem and return file name
+        return collect(glob("{$prefix}*{$suffix}"))
+            ->first(fn (string $filename) =>
+                // 1. Look for an exact match and return it.
+                ($filename === "{$prefix}{$locale}{$suffix}") ||
+                // 2. Look in the preference list for a preferred fallback.
+                (isset($this->localeMapping[$localeAndRegion]) && $filename === "{$prefix}{$this->localeMapping[$localeAndRegion]}{$suffix}") ||
```
bozana: this can be $this->localeMapping[$locale] -- because we map our locales (the whole locale as it is) to those of credit

bozana:
Merged with https://github.com/pkp/pkp-lib/pull/11639.

Issue #11502, opened 2025-06-12
[Admin] | New submission - Author Contributor, admin created user creates unknown affiliation and is flagged during final step of submission process


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

There are two scenarios to create a user. 
1. New Invite User to a Role **works correctly**.  If affiliation field is left blank during this invitation process, then this seems to be a correct empty string.  Maybe it isn't set to NULL by default?
2. If you use Admin method it fails.  Use Admin > Hosted site > Setting Wizard > Users > Add user.  For some reason this empty affiliation field is not being parsed correctly.

![Image](https://github.com/user-attachments/assets/cab50be3-27da-4e41-a96d-3d7043722360)

If an author user is created using method 2 (Admin) then when that author is working through the submission creation process they will be automatically added as a contributor.  If you continue to the final step of the submission creation process there will be an error with this contributor.  

![Image](https://github.com/user-attachments/assets/d003c337-8cdb-49b9-acee-f5175f138152)

Upon editing the author contributor you will see that the affiliation field is missing a valid English translation.  It isn't obvious where to enter the language.  You will need to select the 0/1 languages completed.

![Image](https://github.com/user-attachments/assets/bf98ff76-ef10-47a9-ad24-b77401acb00c)

### Steps to Reproduce

1. Prerequisites: Must have a user created with author role assigned using the Admin scenario above without entering in affliation
2.  Login with this Author user
3. Select "Start a new Submission" from side menu
4. Fill out the new submission form
5. Continue until Step 3 - Contributors and it should automatically add the current author as the contributor
6. Select Continue and proceed until the final step
7. Observe the warning errors

### Expected Result

If the affliation field is empty for the Invite User to a Role scenario and works without issue, then I would expect when creating a user with Admin to follow the same pattern.  I believe the problem lies in the creation of the user with Admin and having that extra section with affliation being set to a non-string value.

Here is the Affiliation Field

![Image](https://github.com/user-attachments/assets/b2d6c134-c2bd-4bd1-8f3f-cba07ff36d77)

### Actual Result

You get an error message during the last stage of submit.

![Image](https://github.com/user-attachments/assets/d15a1d45-b8d5-4337-b0c8-ee2220a416ad)

And when you look into the contributor's Affiliation it will say this

![Image](https://github.com/user-attachments/assets/f8802b2b-ac38-4368-8ddd-3009453e65fc)

### Application Version

[...]

PR #11509, opened 2025-06-13
pkp/pkp-lib#11502 don't migrate affiliations with empty names

kaitlinnewson:


classes/affiliation/Repository.php
```diff
@@ -258,6 +258,9 @@ public function migrateUserAffiliation(User $user, Submission $submission, Conte
             return null;
         }
         foreach ($userAffiliations as $locale => $name) {
+            if (empty($name)) {
+                return null;
+            }
```
bozana: I think we should check this above, in line 257:
`if (count(array_filter($user->getAffiliation(null))) == 0) return null;`
Because `$userAffiliations` there we be something like:
```
Array
(
    [en] => 
    [fr_CA] => 
)
```
which will not be empty and thus `null` will not be returned above.
Here, in this foreach, we need to go through all locales, maybe one name in a locale will match with ror entry.

kaitlinnewson: Updated - if I'm understanding the possible cases correctly, I think `$userAffiliations` could be null (e.g. no entries at all for the user when invited), so we need to check that as well before.

Issue #11532, opened 2025-06-17
Install form time zone is not considered during initial installation


prbt2016:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Hello,

I was in the process of manual installation of OJS, OMP, OPS 3.5.0.0 downloaded from official website on AlmaLinux release 9.6 (Sage Margay) with PHP 8.2.28, MYSQL 5.7.44, Apache 2.4.63.

Additional details : Enter admin details, timezone selection : America/New_York and Enter db details.

After a successful installation when I log in to the installation the URL redirects to (in case of OJS) : 

http://example.com/ojs3500/index.php/index/en

instead of 

http://example.com/ojs3500/index.php/index/en/admin/contexts

(Same case with OMP, AND OPS)

I see this content : 

OJS:

![Image](https://github.com/user-attachments/assets/37b5743a-97e2-4c87-a2bc-124cd16b062a)

OMP:

![Image](https://github.com/user-attachments/assets/2fc07c54-6cd5-44c8-8677-13c3a8d21dd6)

OPS:

![Image](https://github.com/user-attachments/assets/fc15dd55-3832-49f3-9229-11eb39afa0dc)

Hence, I am not able to see the Create Journal, Create Press or Create Server option.

What could be the issue?.

 

### Steps to Reproduce

-  Install OJS/OMP/OPS
-  Log in using the login form
- Post log in, URL redirects to http://example.com/ojs3500/index.php/index/en

### Expected Result

URL should redirect to http://example.com/ojs3500/index.php/index/en/admin/contexts post login

### Actual Result

URL redirects to http://example.com/ojs3500/index.php/index/en post login

### Environment Details

AlmaLinux release 9.6
PHP 8.2.28 
MYSQL 5.7.44 
Apache 2.4.63

### Application Version

OJS, OMP, OPS 3.5.0.0

## Pull Requests

- pkp-lib 3.5: https://github.com/pkp/pkp-lib/pull/11543
- pkp-lib main: https://github.com/pkp/pkp-lib/pull/11561

### Submodule Updates/Tests

- OJS 3.5: https://github.com/pkp/ojs/pull/4944
- OMP 3.5: https://github.com/pkp/omp/pull/2042
- OPS 3.5: https://github.com/pkp/ops/pull/1025
- OJS main (tests only): https://github.com/pkp/ojs/pull/4956

asmecher:
[...]

PR #11543, opened 2025-06-19
pkp/pkp-lib#11532 set database time zone during installation

kaitlinnewson:


kaitlinnewson:
Hi @jonasraoni, please ignore the review request for now - I see tests failing so I've got more investigating to do

kaitlinnewson:
Hi @jonasraoni, can you give this an initial review? Tests are now mostly passing in OJS except for one unrelated test failure, and I'm going to set up OMP/OPS tests soon.

classes/install/PKPInstall.php
```diff
@@ -278,6 +291,40 @@ public function createData()
...
+     */
+    protected function initializeDatabaseTimeZone(string $timeZone): bool
+    {
+        try {
+            // Validate and get the timezone
+            $timeZoneName = (new DateTimeZone($timeZone))->getName();
+
+            // Get the current offset for this timezone
+            $dateTime = new DateTime('now', new DateTimeZone($timeZoneName));
+            $offset = $dateTime->format('P');
```
bozana: Maybe to use `date_default_timezone_set` (the next line below) first, before getting the offset, and then to use `$offset = (new DateTime())->format('P');` (to get offset for the current time zone).
Because we say: `$timeZone ?: ini_get('date.timezone') ?: 'UTC'` in `date_default_timezone_set` -- The timeZone will probably be correctly/successfully constructed, because it comes from the install form where it is also validated, but if for some reason the `timeZone` is not successfully created, then other time zone will be set in PHP. So maybe it is better to calculate offset based on that, on the actually set time zone?

classes/install/PKPInstall.php
```diff
@@ -102,7 +109,13 @@ public function preInstall()
         // initial details from the `config.inc.php` rather than what is set through the install form.
         app()->register(new \Illuminate\Database\DatabaseServiceProvider(app()));
 
-        return parent::preInstall();
+        $result = parent::preInstall();
+
+        if ($result && $this->getParam('timeZone')) {
+            $this->initializeDatabaseTimeZone($this->getParam('timeZone'));
```
bozana, line 114: Do we need to consider the return value here -- if the time zone could not be set correctly in the DB, when the function returns false? Shell the installation proceed however? :thinking:

bozana:
Hi @kaitlinnewson, just two questions :-)

kaitlinnewson:
Thanks @bozana - I've updated the PR based on your feedback as well as the submodule PRs.

Issue #11544, opened 2025-06-19
Author ORCID request email fails with new author


ewhanson:
**Description:**
- Author ORCID verification requests do not work when creating a new author as the ORCID verification request email expects an author ID and author to already exist.
- Users should be able to request author ORCID verification when creating a new author without saving the form first. 

**Steps to test the changes:**
1. Go to an unpublished submission.
2. Open the "add a contributor" form for a publication.
3. Click the "Request verification" ORCID button, and confirm.
4. Submit the form.
5. Check email to confirm ORCID email verification receipt.

**Solution:** Include a value as part of the form data sent to the create contributor API endpoint to trigger the ORCID request email after the user has been created for the first time.

**What application are you using?**
  OJS `main` and `stable-3_5_0`

**Additional information**
N/A

ewhanson:
@taslangraham, could you have a look at these PRs? Thanks!

taslangraham:
@ewhanson  Approved with minor suggestions on the pkp-lib and ui-library PRs. I think the failing tests might be due to changes made for this [issue](https://github.com/pkp/pkp-lib/issues/11538) which changed how category options are being displayed.

ewhanson:
Thanks @taslangraham, I see you've been reviewing that one as well, so I'll wait and see what happens with that (before merging into `stable-3_5_0`).

taslangraham:
@ewhanson I don't think the changes made for https://github.com/pkp/pkp-lib/issues/11538 by @jardakotesovec were ported to `stable-3_5_0`, so it should be fine for your to port your changes to `stable-3_5_0`. Currently, only category tests on `main` needs to be updated

taslangraham:
Category tests has been fixed by @jardakotesovec!

Tribunal33:
This passes QA. I have checked a few combinations of forms and user flows. I would say that as an Author you are not able to edit the metadata for Contributors so I would recommend removing that as a testing step.

As an author you are able to create a new submission and then request ORCiD verification from yourself as the primary author or any additional co-author. I have recorded these test scenarios in the QA Zone for plugin testing of ORCiD.

Additional note that the email will only be sent after the contributor has been created. Not when you initially select that button and there is a dialogue box that explains this detail.

![Image](https://github.com/user-attachments/assets/ceee466e-adbd-44b8-b196-d041df509738)

PR #11545, opened 2025-06-19
pkp/pkp-lib#11544 Author ORCID request email fails with new author

ewhanson:


classes/components/forms/FieldOrcid.php
```diff
@@ -43,6 +43,12 @@ public function getConfig()
         $config['authorId'] = $this->authorId;
         $config['isVerified'] = $this->isVerified;
 
+        // This field must not be inert to let the contributor API endpoint know the ORCID verification request email
+        // should be sent once the new author has been created.
+        if ($this->authorId === 0) {
```
taslangraham:
```suggestion
        if (!$this->authorId) {
```

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1616,6 +1618,19 @@ public function addContributor(Request $illuminateRequest): JsonResponse
...
 
         $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_AUTHOR, $illuminateRequest->input());
+
+        // Allows author ORCID request email to be triggered from frontend before author ID exists
+        $shouldSendOrcidEmail = false;
+        switch ($params['orcid']) {
+            case null:
+                unset($params['orcid']);
+                break;
+            case 'shouldUpdate':
```
taslangraham: Up to you if you think it it's worth changing, but I think something like `'shouldRequestVerification'` would be more descriptive here

taslangraham (approved):
Overall, this looks good. I've left two minor suggestions which you can include if you agree with

ewhanson:
Thanks @taslangraham!

Issue #11517, opened 2025-06-13
[Suggest Reviewers] | Various areas - small adjustments to 4787 feature, to align with design consistency


Tribunal33:
**Description:**
There were a few areas that can be fixed to align more with the original intended designs without overhauling the full redesign layout.  This are some minor tweaks that were missed during the original development cycle.

3 areas for improvement.
1.  New Submission Final Step should not have Reasons for Suggesting Reviewer

![Image](https://github.com/user-attachments/assets/49a0eed8-e2f3-47a8-934f-268fe513d5ae)
As per Design it should only have {Name}, {Email} and {Affiliation}

![Image](https://github.com/user-attachments/assets/96c6c6dc-a18a-421b-98cd-2b4c4b14a27d)

2. Remove the Extra Author Table at the top of the Add Reviewer form

![Image](https://github.com/user-attachments/assets/1c8e7dcf-eed7-4eaa-b199-ccc57e4bdf5d)

3. Add the More Information button to the three dots menu during the Review Stage

![Image](https://github.com/user-attachments/assets/5bcf8a19-2ec3-455f-a102-05fde233ce8b)
 

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
1. Check the new submission Final Step for part 1
2. Check the Add reviewer Form when submission is in Review Stage for part 2
3. Check the Three dots menu for a suggested reviewer for More Information option for part 3

**Solution:** What the proposed solutions are and additional affected areas

###Requirements
- [ ] Remove the Reasons for Suggesting in the New Submission Review final step and replace with affiliation as per design
- [ ] Remove the Extra Author table at the top of the +Add reviewer page
- [ ] Add the More Information button to the three dots during the Review Stage

**What application are you using?**
 OJS, OMP 3.5

**Additional information**
OPS does not have review stage so not sure if it is relevant.

Tribunal33:
After reviewing with @Devika008 I think the only relevant one now is "New Submission Final Step should not have Reasons for Suggesting Reviewer"

I will remove the other 2 as they are expected.

taslangraham:
@kaitlinnewson @blesildaramirez could you give these a review, please?

**PRs**

pkp-lib(main): https://github.com/pkp/pkp-lib/pull/11554
ojs (main)[TESTS ONLY]: https://github.com/pkp/ojs/pull/4950

**stable 3.5 PRs**
pkp-lib: https://github.com/pkp/pkp-lib/pull/11555
[...]

PR #11554, opened 2025-06-20
pkp/pkp-lib#11517 Remove suggestion reason for suggested reviewer in submission wizard(main)

taslangraham:
For pkp/pkp-lib#11517

### DIFF
--- a/locale/en/submission.po
+++ b/locale/en/submission.po
@@ -45,9 +45,6 @@ msgstr "A ROR name is required."
 msgid "ror.displayLocaleRequired"
 msgstr "ROR display locale is required."
 
-msgid "submission.author.list"
-msgstr "Submission Author List"
-
 msgid "submission.authors.label"
 msgstr "Authors"
 
--- a/templates/submission/review-reviewer-suggestions.tpl
+++ b/templates/submission/review-reviewer-suggestions.tpl
@@ -28,7 +28,7 @@
             submissionWizard__reviewPanel__body--{$step.id|escape}
         "
     >
-        <notification 
+        <notification
             v-if="!submission.reviewerSuggestions?.length ?? 0"
             type="warning"
             class="submissionWizard__reviewEmptyWarning"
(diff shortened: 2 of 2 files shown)


locale/en/submission.po
```diff
@@ -45,9 +45,6 @@ msgstr "A ROR name is required."
 msgid "ror.displayLocaleRequired"
 msgstr "ROR display locale is required."
 
-msgid "submission.author.list"
```
taslangraham, line 48: Removed because it it duplicated on line 57

kaitlinnewson (approved):

templates/submission/review-reviewer-suggestions.tpl
```diff
@@ -54,14 +54,14 @@
...
                     <h5> {{ reviewerSuggestion.email }} </h5>
                 </div>
-                
+
                 <div
-                    class="submissionWizard__reviewPanel__item__value semantic-defaults" 
-                    v-strip-unsafe-html="localize(reviewerSuggestion.suggestionReason)"
+                    class="submissionWizard__reviewPanel__item__value"
+                    v-strip-unsafe-html="localize(reviewerSuggestion.affiliation)"
                 ></div>
```
blesildaramirez: I don't think we need to use `v-strip-unsafe-html` for displaying the user's affiliation. We normally use this when we expect the data to have html content. I checked other places where we display the user information (e.g. Users & Roles page) and we don't use html sanitization there.

I suggest to do the same here:
```suggestion
                    class="submissionWizard__reviewPanel__item__value">
                    <h5> {{ localize(reviewerSuggestion.affiliation) }} </h5>
                </div>
```

blesildaramirez (changes requested):

templates/submission/review-reviewer-suggestions.tpl
taslangraham: updated

Issue #11559, opened 2025-06-23
Fatal error Uncaught TypeError in NotificationsGridCellProvider.php when handling notifications for a deleted query.


thuanhuynhh:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The error occurs when a notification is associated with discussion of the submission that no longer exists in the database. (I have declined Submission after user add a discussion

### Steps to Reproduce

1. Login with User/Role Author add new submission and discussion
2. Back to admin decline submission and delete submission 
3. Click on Notification

### Expected Result

OJS should handle the orphaned notification gracefully.  It should either skip rendering the invalid notification or display a placeholder title (e.g., "-") without causing the application to crash.

### Actual Result

Warning: Attempt to read property "assocType" on null in /path/to/ojs/lib/pkp/controllers/grid/notifications/NotificationsGridCellProvider.php on line 158

Warning: Undefined variable $submissionId in /path/to/ojs/lib/pkp/controllers/grid/notifications/NotificationsGridCellProvider.php on line 182

Fatal error: Uncaught TypeError: PKP\submission\Repository::get(): Argument #1 ($id) must be of type int, null given, called in /path/to/ojs/lib/pkp/controllers/grid/notifications/NotificationsGridCellProvider.php on line 182...

### Application Version

3.5.0

## Pull Requests

- pkp-lib 3.5: https://github.com/pkp/pkp-lib/pull/11564
- pkp-lib main: https://github.com/Vitaliy-1/pkp-lib/pull/5 (filed against existing PR branch to avoid conflicts with work in https://github.com/pkp/pkp-lib/issues/10406)

### Submodule Updates

- OJS 3.5: https://github.com/pkp/ojs/pull/4957
- OMP 3.5: https://github.com/pkp/omp/pull/2044
- OPS 3.5: https://github.com/pkp/ops/pull/1029

thuanhuynhh:
quick fix add below line 156 `$query = Query::find($notification->assocId);` file lib\pkp\controllers\grid\notifications\NotificationsGridCellProvider.php
```
if (!$query) {
   return '—';
}
```

thuanhuynhh:
quick fix add below line 156 $query = Query::find($notification->assocId); file lib\pkp\controllers\grid\notifications\NotificationsGridCellProvider.php

```
if (!$query) {
   return '—';
}
```

and lib\pkp\classes\notification\managerDelegate\QueryNotificationManager.php line 41

```
[...]

PR #11564, opened 2025-06-25
pkp/pkp-lib#11559 clean up related data when deleting a Query

kaitlinnewson:


classes/submission/DAO.php
```diff
@@ -277,9 +277,12 @@ public function deleteById(int $id): int
...
 
-        // Delete the queries associated with a submission
-        Query::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $id)
-            ->delete();
+        // Delete the queries associated with a submission. Load models to ensure cleanup happens.
+        $queries = Query::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $id)
+            ->get();
+        foreach ($queries as $query) {
+            $query->delete();
+        }
```
taslangraham: I understand the overall approach and it makes sense. However, one concern I have is the potential for a high number of database calls, especially if a submission has a lot of Queries. Since each Query deletion also triggers deletion of associated Notes and Notifications, the total number of DB calls can add up quickly.

One possible optimization would be to place Query delete logic in the Query Repository class and handle deletions in bulk. For example:

```php
function deleteBySubmissionId(int $submissionId)
{
    $queries = Query::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $submissionId)->get();
    $ids = $queries->pluck('query_id')->toArray();

    if (!empty($ids)) {
        Query::whereIn('query_id', $ids)->delete();
        Note::whereIn('assoc_id', $ids)->delete();
        Notification::whereIn('assoc_id', $ids)->delete();
    }
}


// Or a more generic method
function deleteQueryByAssocType(int $assocType, int $assocId)
{
    $queries = Query::withAssoc($assocType, $assocId)->get();
    $ids = $queries->pluck('query_id')->toArray();

    if (!empty($ids)) {
        Query::whereIn('query_id', $ids)->delete();
        Note::whereIn('assoc_id', $ids)->delete();
        Notification::whereIn('assoc_id', $ids)->delete();
    }
}
```

I haven't tried it, but this could probably be added as a static method on the Query model itself

taslangraham (approved):

controllers/grid/queries/QueriesGridHandler.php
```diff
@@ -177,17 +177,21 @@ public function initialize($request, $args = null)
...
+            case WORKFLOW_STAGE_ID_PRODUCTION:
+                $this->setTitle('submission.queries.production');
                 break;
             case WORKFLOW_STAGE_ID_INTERNAL_REVIEW:
             case WORKFLOW_STAGE_ID_EXTERNAL_REVIEW:
                 $this->setTitle('submission.queries.review');
                 break;
-            default: assert(false);
+            default:
+                assert(false);
```
asmecher, line 194: Use an exception; this case really shouldn't happen and most systems have exceptions disabled. Probably lowest risk to do this just on the `main` branch, though.

Issue #11333, opened 2025-04-28
Highlights | Homepage - Long Descriptions ignores paragraph breaks


pmangahis:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When adding a highlight with a long description, the field for description appears to be treated as a Title, with minimal styling options. 

<img width="1086" alt="Image" src="https://github.com/user-attachments/assets/dc4b5cc3-123d-4d9b-98a6-3f9d69e40ef8" />

Highlight with a long description is displayed as a single, continuous body of text. 

<img width="1193" alt="Image" src="https://github.com/user-attachments/assets/ddd2f1ee-8cb1-4e08-b7ff-95e2218bbab0" />

### Steps to Reproduce

1. Login with as a Editor or Journal Manager
2. Navigate to Settings
3. Click on Website
4. Click Setup
5. Click Highlights
6. Add a Highlight 
7. Copy and Paste Long description

### Expected Result

The description field should allow for additional options, similar to the Announcement > Introduction text box (shown below). Allowing users to add text with breaks and make additional formatting changes
<img width="636" alt="Image" src="https://github.com/user-attachments/assets/ddbcdcb1-9e00-46aa-968c-84e5fafb128b" />

Text on the home page to be displayed with paragraph breaks.

### Actual Result

Highlight with a long description is displayed as a single, continuous body of text. 

<img width="1193" alt="Image" src="https://github.com/user-attachments/assets/ddd2f1ee-8cb1-4e08-b7ff-95e2218bbab0" />

### Application Version

OJS 3.5 RC2 Release

taslangraham:
@kaitlinnewson @blesildaramirez Do you mind taking a look at these PRs, please?

pkp-lib(3.5): https://github.com/pkp/pkp-lib/pull/11552
ojs(3.5): https://github.com/pkp/ojs/pull/4948
ops(3.5): https://github.com/pkp/ops/pull/1023
omp(3.5): https://github.com/pkp/omp/pull/2040

pkp-lib(main): https://github.com/pkp/pkp-lib/pull/11553
ojs(main)[TEST ONLY]: https://github.com/pkp/ojs/pull/4949

**stable 3.4**
pkp-lib: https://github.com/pkp/pkp-lib/pull/11569
ops: https://github.com/pkp/ops/pull/1031
ojs: https://github.com/pkp/ojs/pull/4960
omp: https://github.com/pkp/omp/pull/2045

[...]

PR #11569, opened 2025-06-26
pkp/pkp-lib#11333 Use FieldRichTextarea for Highlights form description field

taslangraham:
For pkp/pkp-lib#11333

### DIFF
--- a/classes/components/forms/highlight/HighlightForm.php
+++ b/classes/components/forms/highlight/HighlightForm.php
@@ -50,6 +50,7 @@ class HighlightForm extends FormComponent
             ->addField(new FieldRichTextarea('description', [
                 'label' => __('common.description'),
                 'isMultilingual' => true,
+                'size' => 'large',
             ]))
             ->addField(new FieldText('url', [
                 'label' => __('common.url'),


kaitlinnewson (approved):

classes/components/forms/highlight/HighlightForm.php
```diff
@@ -42,10 +44,10 @@ public function __construct(string $action, string $baseUrl, string $temporaryFi
...
-                'label' => __('common.title'),
-                'isMultilingual' => true,
-            ]))
-            ->addField(new FieldRichText('description', [
+            'label' => __('common.title'),
+            'isMultilingual' => true,
+        ]))
+            ->addField(new FieldRichTextarea('description', [
                 'label' => __('common.description'),
                 'isMultilingual' => true,
```
blesildaramirez, line 52: Probably doesn't matter but we can also add the 'size' => 'large', here as we do for the other versions.
```suggestion
                'isMultilingual' => true,
                'size' => 'large',
```

That would only be required if we enable the feature flag on v3.4, but I'm not sure.

blesildaramirez (approved):

Issue #8920, opened 2023-04-14
Improve search framework


jonasraoni:

diegoabadan:
Hello!

To our knowledge, an institution with the recently released OJS 3.4.0-7 and which hosts several journals was experiencing immense slowness and contacted us. We had to apply this fix:
https://forum.pkp.sfu.ca/t/ojs-3-4-0-4-performance-issue-ojs-3-3-0-14-to-ojs-3-4-0-4/83655/5?u=abadan

asmecher:
@ewhanson -- just a heads-up that we're interested in OpenSearch as an option.

asmecher:
I've implemented a first cut at replacing the built-in search index (and hook-based option for other implementations) using Laravel Scout as a middle layer. You can specify and configure driver in `config.inc.php`:
```
;;;;;;;;;;;;;;;;;;;
; Search Settings ;
;;;;;;;;;;;;;;;;;;;

[search]

; The search driver to use. Options: database (default), opensearch
driver = database

; For indexing engines supporting it (e.g. opensearch), provide a name for the submission index.
search_index_name = "submissions"

; When using the OpenSearch driver, the following must be specified...
;opensearch_hosts = "[\"https://localhost:9200\"]"
;opensearch_username = "admin"
;opensearch_password = "password"
;opensearch_ssl_verification = On
```

The `database` driver (implemented in `lib/pkp/classes/search/engines/DatabaseEngine.php`) uses fulltext indexing in MySQL, MariaDB, or PostgreSQL. This seems to be widely enough supported that we can make use of it now.

The `opensearch` driver (implemented in `lib/pkp/classes/search/engines/OpenSearchEngine.php`), once you configure it in `config.inc.php`, doesn't need any additional action on the OpenSearch side -- just run `tools/rebuildSearchIndex.php` to create and build the index.

(We can't use the built-in Scout engines because they presume we use Eloquent for submission content, and they don't match our use case very well. But the Scout engines are nice and simple so maintaining our own isn't nearly the overhead we had with our previous implementations.)

I would like to review and improve some things, but consider this a good enough "first cut" to get merged into `main` so later improvements can happen there iteratively.

@jonasraoni, would you mind taking a first review of the PRs linked in the issue description for general approach? Then I can port to OMP and OPS, and work through the set of issues referred to in the head of the list to make sure we have good coverage of the full set of requests/features.

[...]

PR #11578, opened 2025-06-27
pkp/pkp-lib#8920 Implement Laravel Scout and Opensearch

asmecher:
See https://github.com/pkp/pkp-lib/discussions/7968#discussioncomment-13645386 for details.


composer.json
```diff
@@ -27,14 +27,17 @@
...
 		"moxiecode/plupload": "2.*",
 		"mpdf/mpdf": "^8.2",
+		"opensearch-project/opensearch-php": "^2.4",
 		"phpmailer/phpmailer": "6.*",
 		"ralouphie/getallheaders": "*",
 		"smarty/smarty": "4.*",
 		"sokil/php-isocodes": "^4.1",
 		"sokil/php-isocodes-db-i18n": "^4.0",
 		"symfony/mailer": "^7.0",
+		"teamtnt/tntsearch": "^5.0",
```
jonasraoni: I think this was forgotten.

classes/core/PKPContainer.php
```diff
@@ -83,6 +83,7 @@ protected function registerBaseBindings(): void
         $this->instance('app', $this);
         $this->instance(Container::class, $this);
         $this->instance('path', $this->basePath);
+        $this->instance('path.config', $this->basePath . DIRECTORY_SEPARATOR . 'config'); // Necessary for Scout to let CLI happen
```
jonasraoni:
```suggestion
        $this->instance('path.config', "{$this->basePath}/config"); // Necessary for Scout to let CLI happen
```

classes/core/PKPContainer.php
```diff
@@ -185,8 +186,9 @@ public function registerConfiguredProviders(): void
         $this->register(new InvitationServiceProvider($this));
         $this->register(new ScheduleServiceProvider($this));
         $this->register(new ConsoleCommandServiceProvider($this));
-        $this->register(new \PKP\core\ValidationServiceProvider($this));
+        $this->register(new ValidationServiceProvider($this));
         $this->register(new \Illuminate\Foundation\Providers\FormRequestServiceProvider($this));
+        $this->register(new \Laravel\Scout\ScoutServiceProvider($this));
```
jonasraoni, line 192: What's our pattern (I'm using A)?
A. `use namespace`
B. Full class names when there's just one usage

classes/core/PKPContainer.php
```diff
@@ -469,8 +471,19 @@ protected function loadConfiguration(): void
...
         ];
 
+        $items['scout'] = [
+            'driver' => Config::getVar('search', 'driver', 'database'),
+        ];
+
         // Create instance and bind to use globally
         $this->instance('config', new Repository($items));
+
+        app()->extend(\Laravel\Scout\EngineManager::class, function ($service, $app) {
```
jonasraoni:
```suggestion
        app()->extend(\Laravel\Scout\EngineManager::class, fn (Type $service) =>
```

classes/install/Installer.php
```diff
@@ -971,4 +972,24 @@ public function downloadIPGeoDB(): bool
         return true;
     }
 
+    /**
+     * Rebuild the search index.
+     */
+    public function rebuildSearchIndex(): bool
```
jonasraoni, line 978: Comments from here might be re-used: https://github.com/pkp/ojs/pull/4963/files#diff-3636d0a25f479555f372835202415ca8bfb80e855c4676a75eafd3a09a7174bb

classes/migration/install/SubmissionSearchMigration.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function up(): void
+    {
+        Schema::create('submissions_fulltext', function (Blueprint $table) {
+            $table->comment('Fulltext search index for submission content');
+
+            $table->bigInteger('submissions_fulltext_id')->autoIncrement();
+
+            $table->bigInteger('submission_id');
+            $table->foreign('submission_id')->references('submission_id')->on('submissions')->onDelete('cascade');
```
jonasraoni, line 33: Given that a submission is related to a publication, this column can be removed

classes/observers/listeners/UpdateSubmissionInSearchIndex.php
```diff
@@ -43,11 +42,11 @@ public function subscribe(Dispatcher $events): void
 
     public function handleUnpublished(PublicationUnpublished $event)
     {
-        dispatch(new UpdateSubmissionSearchJob($event->submission->getId()));
+        app(\Laravel\Scout\EngineManager::class)->engine()->delete(collect([$event->submission]));
```
jonasraoni, line 45: As the search can be forwarded to external engines, I think it should be still working with a job.

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+    public const MINIMUM_DATA_LENGTH = 4096;
+
+    protected function getIndexName(): string
+    {
+        return Config::getVar('search', 'search_index_name', 'submissions');
+    }
+
+    protected function getTableName(): string
+    {
+        return $this->getIndexName() . '_fulltext';
```
jonasraoni: For the default search engine I think it's better to leave the table name fixed (e.g. if the user changes the index name it will lead to the creation of a new one and some trash will be left behind)

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public function update($models)
+    {
+        $models->each(function ($submission) {
+            Repo::publication()->getCollector()->filterBySubmissionIds([$submission->getId()])->getMany()->each(function ($publication) use ($submission) {
```
jonasraoni: The publications should be available at `Submission::getData('publications')`

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+
+                // Index all galleys
+                $submissionFiles = Repo::submissionFile()
+                    ->getCollector()
+                    ->filterByAssoc(Application::ASSOC_TYPE_REPRESENTATION, [$publication->getId()])
+                    ->filterByFileStages([SubmissionFile::SUBMISSION_FILE_PROOF])
+                    ->getMany();
+
+                foreach ($submissionFiles as $submissionFile) {
+                    $galley = Application::getRepresentationDAO()->getById($submissionFile->getData('assocId'));
```
jonasraoni: The galleys should be available at `Publication::getData('galleys')`, from there perhaps it's possible to retrieve the files in a simpler way (e.g. `filterByFileIds()`). There's a `Galley::getFile()`, but it will lead to many DB requests.

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+                        ],
+                        ['submission_id', 'publication_id', 'locale'],
+                        ['title', 'abstract', 'body']
+                    );
+                }
+            });
+        });
+    }
+
+    public function delete($models)
```
jonasraoni, line 39: Missing typing (perhaps through phpdoc since this might be an inherited method) + documentation on the methods.

If the `@copydoc` isn't needed for the "documentation tool", then I'd say we can drop the copydoc from everywhere.

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+    {
+        $models->each(function ($submission) {
+            DB::table($this->getTableName())->where('submission_id', $submission->getId())->delete();
```
jonasraoni: A `map()` with `whereIn()` can be used here.

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+        });
+    }
+
+    protected function buildQuery(SearchBuilder $builder): DatabaseBuilder
+    {
+        // Handle "where" conditions
+        $contextId = null;
+        $publishedFrom = $publishedTo = null;
```
jonasraoni:
```suggestion
        $contextId = $publishedFrom = $publishedTo = null;
```

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+        foreach ($builder->wheres as $field => $value) {
+            switch ($field) {
```
jonasraoni:
```suggestion
            $$field = match ($field) {
```

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+                    $$field = $value ? new \Carbon\Carbon($value) : null;
+                    break;
+                default: throw new \Exception("Unsupported field {$field}!");
+            }
+        };
+
+        // Handle "whereIn" conditions
+        $sectionIds = $categoryIds = null;
+        foreach ($builder->whereIns as $field => $list) {
+            switch ($field) {
```
jonasraoni:
```suggestion
            $$field = match ($field) {
```

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+                default: throw new \Exception("Unsupported field {$field}!");
+            }
+        };
+
+        // Handle options
+        foreach ($builder->options as $option => &$value) {
+            switch ($option) {
+                default: throw new \Exception("Unsupported options {$option}!");
+            }
+        };
```
jonasraoni, line 70: Remove block or `if ($builder->options) { throw ... }`?

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+
+        return DB::table($this->getTableName() . ' AS ftsearch')
+            ->join('submissions AS s', 'ftsearch.submission_id', 's.submission_id')
+            ->when($contextId, fn ($q) => $q->where('context_id', $contextId))
```
jonasraoni:
```suggestion
            ->when($contextId, fn (Builder $q) => $q->where('context_id', $contextId))
```

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+            ->when($publishedFrom || $publishedTo || is_array($sectionIds), fn ($q) => $q->whereExists(
+                fn ($q) => $q->select(DB::raw(1))
```
jonasraoni:
```suggestion
                fn ($q) => $q->selectRaw(1)
```

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+                    ->from('publications AS p')
+                    ->whereColumn('p.submission_id', 's.submission_id')
+                    ->where('p.published', 1)
+                    ->when($publishedFrom, fn ($q) => $q->where('p.date_published', '>=', $publishedFrom))
+                    ->when($publishedTo, fn ($q) => $q->where('p.date_published', '<', $publishedTo))
```
jonasraoni: If the comparison isn't considering the time, then there's the `whereDate()`.

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+    public function paginate(SearchBuilder $builder, $perPage, $page): LengthAwarePaginator
+    {
+        return $this->buildQuery($builder)
+            ->select(['s.submission_id AS submissionId'])
+            ->paginate($perPage, ['submissionId'], 'submissions', $page);
+    }
+
+    public function mapIds($results)
+    {
+        throw new \BadFunctionCallException('Unimplemented function.');
```
jonasraoni: Not important, but the `BadMethodCallException` suits better

classes/search/engines/DatabaseEngine.php
```diff
@@ -0,0 +1,221 @@
...
+    {
+    }
+
+    public function deleteIndex($name)
+    {
+    }
+
+    public function flush($model)
+    {
+        DB::table($this->getTableName())->delete();
```
jonasraoni: Perhaps the `truncate()` performs better

```suggestion
        DB::table($this->getTableName())->truncate();
```

classes/search/engines/OpenSearchEngine.php
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Get an OpenSearch client.
+     */
+    protected function getClient(): Client
+    {
+        $hosts = Config::getVar('search', 'opensearch_hosts', null);
+        if ($hosts === null) {
+            throw new \Exception('The opensearch_hosts configuration is missing. Review your config.inc.php file for details.');
+        }
```
jonasraoni: This construction looks more tidy in my opinion:

```suggestion
        $hosts = Config::getVar('search', 'opensearch_hosts', null)
            ?? throw new \Exception('The opensearch_hosts configuration is missing. Review your config.inc.php file for details.');
```

classes/search/engines/OpenSearchEngine.php
```diff
@@ -0,0 +1,260 @@
...
+
+        $username = Config::getVar('search', 'opensearch_username', null);
+        $password = Config::getVar('search', 'opensearch_password', null);
+        if ($username === null || $password === null) {
+            throw new \Exception('The opensearch username and/or password is missing. Review your config.inc.php file for details.');
+        }
+
+        return (new \OpenSearch\ClientBuilder())
+            ->setHosts(json_decode($hosts, flags: JSON_OBJECT_AS_ARRAY))
```
jonasraoni:
```suggestion
            ->setHosts(json_decode($hosts, flags: JSON_OBJECT_AS_ARRAY | JSON_THROW_ON_ERROR))
```

classes/search/engines/OpenSearchEngine.php
```diff
@@ -0,0 +1,260 @@
...
+            ->setBasicAuthentication($username, $password)
+            ->setSSLVerification((bool) Config::getVar('search', 'opensearch_ssl_verification', true))
+            ->build();
+    }
+
+    public function update($models)
+    {
+        $client = $this->getClient();
+        $models->each(function ($submission) use ($client) {
+            $publication = $submission->getCurrentPublication();
```
jonasraoni, line 64: Perhaps there should be a try/catch here and in similar places, to allow the code to proceed in case there's a bad submission.

classes/search/engines/OpenSearchEngine.php
```diff
@@ -0,0 +1,260 @@
...
+                try {
+                    $parser->open();
+                    do {
+                        for ($buffer = ''; ($chunk = $parser->read()) !== false && strlen($buffer .= $chunk) < static::MINIMUM_DATA_LENGTH;);
+                        if (strlen($buffer)) {
+                            $bodies[$galley->getLocale()] = ($bodies[$galley->getLocale()] ?? '') . $buffer;
+                        }
+                    } while ($chunk !== false);
+                } catch (\Throwable $e) {
+                    error_log($e);
```
jonasraoni: I think the submission ID won't be included in the error, so it makes sense to include more information here and in similar occurrences

classes/search/engines/OpenSearchEngine.php
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    public function delete($models)
+    {
+        $client = $this->getClient();
+        $models->each(function ($submission) use ($client) {
+            $client->delete([
+                'index' => $this->getIndexName(),
+                'id' => $submission->getId(),
+            ]);
```
jonasraoni, line 128: I didn't watch the documentation, but there's a `deleteByQuery()`, perhaps it can be used to do a batch process

classes/search/parsers/SearchFileParser.php
```diff
@@ -1,26 +1,18 @@
...
- */
-
-/**
- * @file classes/search/SearchFileParser.php
+ * @file classes/search/parsers/SearchFileParser.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7: Perhaps a script to find the earliest date of each file in Git and change the copyrights to `Copyright (c) created-year`?

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,63 @@
...
+    public function newCollection(array $models = [])
+    {
+        $contextCache = [];
+        $sectionCache = [];
+
+        return LazyCollection::make(function () use ($models, &$contextCache, &$sectionCache) {
+            foreach ($models as $data) {
+                $submissionId = is_scalar($data) ? (int) $data : (int) $data->submissionId;
+
+                $submission = Repo::submission()->get($submissionId);
```
jonasraoni, line 44: Do a loop to pre-collect the submission IDs, then fire a single request to load all submissions

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,63 @@
...
+                if (!$submission || $submission->getData('status') !== PKPSubmission::STATUS_PUBLISHED) {
```
jonasraoni:
```suggestion
                if ($submission?->getData('status') !== PKPSubmission::STATUS_PUBLISHED) {
```

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,63 @@
...
+                    continue;
+                }
+                $currentPublication = $submission->getCurrentPublication();
+
+                $contextId = $submission->getData('contextId');
+                $context = $contextCache[$contextId] ?? ($contextCache[$contextId] = Application::getContextDAO()->getById($contextId));
```
jonasraoni: The same idea I proposed previously can be done here too (pre-collect the contextIds)
```suggestion
                $context = $contextCache[$contextId] ??= Application::getContextDAO()->getById($contextId);
```

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,63 @@
...
+
+                $sectionId = $currentPublication->getData('sectionId');
+                $section = $sectionId ? $sectionCache[$sectionId] ?? ($sectionCache[$sectionId] = Repo::section()->get($sectionId)) : null;
```
jonasraoni: Pre-collect the section IDs

```suggestion
                $section = $sectionId ? ($sectionCache[$sectionId] ??= Repo::section()->get($sectionId)) : null;
```

classes/submission/Collector.php
```diff
@@ -47,10 +46,11 @@ abstract class Collector implements CollectorInterface, ViewsCount
     public const ORDERBY_LAST_MODIFIED = 'lastModified';
     public const ORDERBY_SEQUENCE = 'sequence';
     public const ORDERBY_TITLE = 'title';
-    public const ORDERBY_SEARCH_RANKING = 'ranking';
     public const ORDER_DIR_ASC = 'ASC';
     public const ORDER_DIR_DESC = 'DESC';
 
+    public const SEARCH_STOPWORDS_FILE = 'lib/pkp/registry/stopwords.txt';
```
jonasraoni: I think this is outdated with the fulltext search, then the file can also be removed from Git

jonasraoni (approved):

classes/core/PKPContainer.php
asmecher: I have a slight preference for B, but don't especially care if we mix approaches. If we do enforce one, it should be using a code formatter so we don't have to worry about it anymore.

classes/search/SubmissionSearchResult.php
asmecher: I'll get rid of the exact test (in case of database type issues), but won't use `?->` along with that, because it would result in a constant value `0` testing as `true` in comparison with `null`.

classes/search/engines/DatabaseEngine.php
asmecher: The Laravel scout parent class [doesn't have types either](https://github.com/laravel/scout/blob/10.x/src/Engines/Engine.php) so we can't add them here.

classes/search/engines/DatabaseEngine.php
asmecher: I'm expecting to add additional cases as we go forward.

classes/search/parsers/SearchFileParser.php
asmecher: Sure, but out of scope for this PR.

classes/migration/install/SubmissionSearchMigration.php
asmecher: I think I'd like to keep it; the cost is very little and it makes life easier since most of the result presentation shows submissions, not publications. We may have to adjust this later, but for the moment, indexing publications and listing submissions as results makes the most sense.

classes/search/SubmissionSearchResult.php
asmecher: This is a generator, and I chose to optimize memory use (not instantiating all objects at once) rather than execution time. For `n` result submissions there will be `n-1` additional queries this way, I think, but there are quite a few per-submission queries and I think this won't be the bottleneck.

classes/search/engines/DatabaseEngine.php
asmecher: I'm not too bothered about optimizing this; I think it'll work well enough as coded.

classes/search/engines/OpenSearchEngine.php
asmecher: I'm working with @defstat to close the loopholes where this data can get in e.g. via Import/Export; I'd rather we set better expectations/protections in the database rather than spreading try/catch clauses around. There would need to be quite a lot of them to catch this circumstance, and it's only caused by broken imports.

classes/search/SubmissionSearchResult.php
jonasraoni: There will be many sub-queries, with or without this optimization, I'm just trying to not create more of them.
Given the result is limited and it's unlikely that someone will display 1000 items on the search, I think it's better to optimize for performance (our current bottleneck). To protect against a possible high page size, chunking would do it, but well, it can be done later...

classes/search/engines/OpenSearchEngine.php
jonasraoni: My main concern was the `$client->create(...)`, if a external system is used for the search (Algolia, Solr, etc), it might fail and then the whole batch will be damned.

If you've covered the batch with a retryable job, then it should be ok 👌

classes/search/engines/DatabaseEngine.php
jonasraoni: Yeah, I expected something like that. But if we know what we're going to receive, then I support using the `@param`, it's also feasible to [type complex arrays](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes), and I'm trying to do it.

Issue #10962, opened 2025-02-21
[Email] | Email metadata - Subject Text is not rendering variable {$contextAcronym} for notify email messages when templates are used


Tribunal33:
**Describe the bug**
Not sure if this affects all email templates but I've been noticing a few of them are not rendering the subject variable {$contextAcronym}

preconditions: Default datastet works and any user that would be part of the participants list of a submission.

**To Reproduce**
1. Login with any user
2. Navigate to dashboard submission list
3. Select view on a submission
4. In the participants area select the three dots
5. Select Notify from the options
6. Select any predefined message from the drop down
a. Body text will be filled in do no need to modify
7. Select Notify button

Expected : Email is sent with subject fully rendered with the journal name 
Actual : message reads "Submission 19 is ready to be copyedited for {$contextAcronym}"

**What application are you using?**
 OJS 3.5

**Additional information**
Please add any screenshots, logs or other information we can use to investigate this bug report.

![Image](https://github.com/user-attachments/assets/4f3bfc56-d8a1-454f-810d-40bae17e51a5)

#### stable-3_5_0
- pkp-lib: [ Move migration to pkp-lib for reuse](https://github.com/pkp/pkp-lib/pull/11778) 
- OJS: [Map `contextAcronym` to `journalAcronym` in email template installer](https://github.com/pkp/ojs/pull/4898) 
- OMP: [pkp/omp#2163](https://github.com/pkp/omp/pull/2163)
- OPS: [pkp/ops#1133](https://github.com/pkp/ops/pull/1133)

asmecher:
The template in the screenshot is `LAYOUT_REQUEST` (in the `emails.layoutRequest.subject` locale key).

forgive38:
Is variable {$context*} replace by {$journal*} in OJS context ?

https://github.com/pkp/ojs/blob/main/classes/migration/upgrade/v3_4_0/I7264_UpdateEmailTemplates.php

asmecher:
@Vitaliy-1 and @Hafsa-Naeem, is this one possible to fix as a high priority for the 3.5.0 release? Or should it be deferred?

Vitaliy-1:
@Hafsa-Naeem, I left a comment. It looks that instances already upgraded to 3.4 won't be affected by the fix

Vitaliy-1:
[...]

PR #11585, opened 2025-06-30
pkp/pkp-lib#10962 Adjust email template variable mappings

Hafsa-Naeem:
for [#10962](https://github.com/pkp/pkp-lib/issues/10962)

### DIFF
--- a/classes/core/ValidationServiceProvider.php
+++ b/classes/core/ValidationServiceProvider.php
@@ -168,6 +168,15 @@ class ValidationServiceProvider extends \Illuminate\Validation\ValidationService
                     if (is_null($this->resolver)) {
                         return new class ($this->translator, $data, $rules, $messages, $attributes) extends \Illuminate\Validation\Validator
                         {
+                            /*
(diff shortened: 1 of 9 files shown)


Vitaliy-1 (changes requested):
Thanks! If it's used more than in 1 app, rather than modifying the localized string, we need to add the variable in the `APP\emailTemplate\DAO::variablesToRename()`.

classes/emailTemplate/DAO.php
```diff
@@ -420,7 +420,9 @@ protected function renameApplicationVariables(string $localizedData): string
      */
     protected function variablesToRename(): array
     {
-        return [];
+        return [
+            'contextAcronym' => 'journalAcronym',
```
touhidurabir: As this is Journal specific (e.g. OJS specific), should it go into the `APP\emailTemplate\DAO::variablesToRename` of OJS app ?

touhidurabir (changes requested):

Issue #10406, opened 2024-09-12
Extend Discussions to add configurable workflow tasks


Devika008:
## Summary

**Describe the problem you would like to solve**
Editorial staff currently face challenges in managing tasks and checklists required to move a submission to the next stage. They also need an efficient way to track and oversee the work assigned to team members and monitor overall progress.

**Describe the solution you'd like**
Incorporating task management into the workflow will help journal managers and editors efficiently track and oversee tasks, monitor team progress, and ensure everything is on track.

**Who is asking for this feature?**
Some community members. This feature was also mentioned as one of the possible solutions for tackling the following issue: https://github.com/pkp/pkp-lib/issues/4976#top

### Option 1
Here, tasks and discussion components on the workflow page are merged into one. When adding something new, you can choose to make it a discussion, a task, or both.

Video: https://www.youtube.com/watch?v=lRblqgea5Ek
Prototype: https://www.figma.com/proto/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?page-id=7200%3A7069&node-id=7430-8478&viewport=824%2C36%2C0.05&t=e2niKeiSTSedNFjR-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=7430%3A8478 

**CONSIDERATIONS FOR THIS OPTION**

- Participants: Only those assigned to the submission can be added to tasks/discussions.
- While you can't change the name of a task/discussion once it's created, you can still add participants even after it's in progress.
- Only participants added to tasks/discussions will appear in the "responsible for" section.
- Once a task/discussion is closed or completed, you can't make changes unless it’s reopened or the completed checkbox is unchecked.
- Suggestion from the participant at the Minneapolis Sprint:
- Editors can determine tasks based on the journal’s policies through the workflow configuration settings. However, tasks like plagiarism check and ethical approval can be pre-determined and set within the configuration.
- A toggle switch in the configuration settings will enable or disable the creation of new tasks.
- Task Visibility: Tasks assigned to a user are only visible to that user. Managing editors, journal editors, and higher roles have access to the entire task list.
- Deadline Management: Deadlines can be set as workflow defaults. Individual task deadlines can be modified as needed.

**OPEN QUESTIONS**

- If a task is closed, does that mean the associated discussion will automatically close too? To what extent should the task and discussion be interdependent, given their close connection in this option.
[...]

PR #11587, opened 2025-06-30
pkp/pkp-lib#10406 Migrations and API for tasks and discussions

Vitaliy-1:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -122,6 +127,11 @@ class PKPSubmissionController extends PKPBaseController
         'getChangeLanguageMetadata',
         'changeVersion',
         'getNextAvailableVersion',
+        'addTask',
+        'editTask',
+        'deleteTask',
```
touhidurabir, line 133: these 2 actions (`editTask` and `deleteTask`) does not seems to have any implementation yet . are these left out intentionally ?

api/v1/submissions/formRequests/AddTask.php
```diff
@@ -0,0 +1,107 @@
...
+                'date',
+                'after:today',
+            ],
+            EditorialTask::ATTRIBUTE_PARTICIPANTS => [
+                'required',
+                'array',
+                function (string $attribute, array $value, Closure $fail) {
+                    $reponsibles = array_filter(Arr::pluck($this->input('participants'), 'isResponsible'));
+                    if (count($reponsibles) > 1) {
+                        return $fail('There should be the only one user responsible for the task');
```
touhidurabir: should we pass translated error message here ?

api/v1/submissions/formRequests/AddTask.php
```diff
@@ -0,0 +1,107 @@
...
+                    }
+
+                    // If the creator is manager or admin, allow participation
+                    if ($currentUser && $currentUser->hasRole([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $submission->getData('contextId'))) {
+                        return true;
+                    }
+
+                    // TODO allow assigning reviewers if it doesn't break anonymity
+
+                    return $fail();
```
touhidurabir: This one does not seems to have any fail message yet .

api/v1/submissions/resources/EditTaskParticipantResource.php
```diff
@@ -0,0 +1,31 @@
...
+ * @brief Transforms the API response of the editorial task participant into the desired format
+ *
+ */
+
+namespace PKP\API\v1\submissions\resources;
+
+use Illuminate\Http\Request;
+use Illuminate\Http\Resources\Json\JsonResource;
+
+class EditTaskParticipantResource extends JsonResource
```
touhidurabir: perhaps class name as only `TaskParticipantResource` as we only have `TaskResource` to have same pattern in naming ?

api/v1/submissions/resources/EditTaskParticipantResource.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function toArray(Request $request): array
+    {
+        return [
+            'id' => $this->id,
+            'userId' => $this->userId,
+            'isResponsible' => (bool) $this->isResponsible,
```
touhidurabir: As we are already casting this one as bool in the related model class , do we need to do another type cast here ?

api/v1/submissions/formRequests/AddTask.php
```diff
@@ -0,0 +1,107 @@
...
+use PKP\stageAssignment\StageAssignment;
+
+class AddTask extends FormRequest
+{
+    public function rules()
+    {
+        $userDao = Repo::user()->dao;
+        $currentUser = Application::get()->getRequest()?->getUser();
+        $input = $this->all();
+        $submission = Repo::submission()->get((int) $input['assocId']);
```
touhidurabir: if we only need to get the `assocId` , can we not get it like `$this->get('assocId')` instead of getting all inputs . 

Also , I did mentioned that in the form request class, we can not use `getAuthorizedContextObject` but probably we can as at this point all the authorization level policies are done, so we can do perhaps
```php
$routeController = PKPBaseController::getRouteController($this->request);
$submission = $routeController->getAuthorizedContextObject(ASSOC_TYPE_SUBMISSION);
```
I haven't tested the above code but it may just work (theoretically) .

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1935,6 +1985,78 @@ public function addDecision(Request $illuminateRequest): JsonResponse
...
     }
 
+    /**
+     * Creates a task or discussion associated with the submission
+     */
+    public function addTask(AddTask $illuminateRequest): JsonResponse
+    {
+        $validated = $illuminateRequest->validated();
+
+        $editorialTask = new EditorialTask($validated);
```
touhidurabir, line 2005: out of curiosity, why not just `EditorialTask::create` ?

classes/editorialTask/EditorialTask.php
```diff
@@ -0,0 +1,206 @@
...
+    protected $table = 'edit_tasks';
+    protected $primaryKey = 'edit_task_id';
+
+    protected $guarded = ['edit_tasks_id', 'id'];
+
+    protected $fillable = [
+        'assocType', 'assocId', 'stageId', 'seq',
+        'createdAt', 'updatedAt', 'closed', 'dateDue',
+        'createdBy', 'type', 'status'
+    ];
```
touhidurabir: Just wondering, any specific reason we need to define both the `guarded` and `fillable` ? The usual convention is to define just one .

classes/editorialTask/EditorialTask.php
```diff
@@ -0,0 +1,206 @@
...
+    use ModelWithSettings;
+
+    // Type of the editorial task
+    public const TYPE_DISCUSSION = 1;
+    public const TYPE_TASK = 2;
+
+    // Current status of the editorial task
+    public const STATUS_NEW = 1;
+    public const STATUS_REPLIED = 2;
+    public const STATUS_CLOSED = 3;
```
touhidurabir: How about refactor these to 2 enum (EditorialTaskType and EditorialTaskStatus) unless they have very minimal usage . Just have better organization with some added ability . Not anything major, so up to you .

classes/editorialTask/EditorialTask.php
```diff
@@ -0,0 +1,206 @@
...
+    public function save(array $options = []): bool
+    {
+        $success = parent::save($options);
+        if (!$success) {
+            return false;
+        }
+
+        $this->participants()->saveMany($this->participants);
+        return $success;
+    }
```
touhidurabir: If I understand correctly, the reason to override these method to allow create task and related participants form a single dataset . do we loose any major ability if we do like `$task->participants()->createMany()` ? I think the less we override core eloquent methods, better for us .

classes/editorialTask/Participant.php
```diff
@@ -12,37 +12,43 @@
...
         return [
-            'queryId' => 'int',
-            'userId' => 'int'
+            'editTaskId' => 'int',
+            'userId' => 'int',
+            'isResponsible' => 'bool',
         ];
     }
-    public function toQuery(): BelongsTo
+    public function toTask(): BelongsTo
```
touhidurabir: why not just `task` ? I think the it was previously `toQuery` as laravel itself have a `query` method and probably had a conflict with that but just `task` should work .

classes/editorialTask/Participant.php
```diff
@@ -12,37 +12,43 @@
...
     public $timestamps = false;
 
+    protected $guarded = [
+        'editTaskParticipantId', 'id'
+    ];
+
     protected $fillable = [
-        'queryId', 'userId'
+        'editTaskId', 'userId', 'isResponsible'
     ];
```
touhidurabir: same as `EditorialTask`, any specific reason we need both the `guarded` and `fillable` ?

classes/editorialTask/Repository.php
```diff
@@ -83,31 +83,34 @@ public function resequence($assocType, $assocId): void
      */
     public function addQuery(int $submissionId, int $stageId, string $title, string $content, User $fromUser, array $participantUserIds, int $contextId, bool $sendEmail = true): int
     {
-        $maxSeq = Query::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $submissionId)
+        $maxSeq = EditorialTask::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $submissionId)
             ->max('seq') ?? 0;
 
-        $query = Query::create([
+        $query = EditorialTask::create([
```
touhidurabir: In many places , we are still refereeing as `$query` and probably better to have named as `$task` as long as it's backward compatible . And maybe for both variable and method .

classes/editorialTask/Repository.php
```diff
@@ -83,31 +83,34 @@ public function resequence($assocType, $assocId): void
...
         $addParticipants = [];
         foreach (array_unique($participantUserIds) as $participantUserId) {
             $addParticipants[] = ([
-                'query_id' => $query->id,
+                'edit_task_id' => $query->id,
                 'user_id' => $participantUserId
             ]);
         }
-        QueryParticipant::insert($addParticipants);
+        Participant::insert($addParticipants);
```
touhidurabir: perhaps we can do it like `query->participants()->createMany(array_unique($participantUserIds))` ?

classes/facades/Repo.php
```diff
@@ -30,6 +30,7 @@
 use PKP\category\Repository as CategoryRepository;
 use PKP\controlledVocab\Repository as ControlledVocabRepository;
 use PKP\decision\Repository as DecisionRepository;
+use PKP\editorialTask\Repository as QueryRepository;
```
touhidurabir: why not just say it as `EditorialTaskRepository` ? and have a new repo method as `task` or `editorialTask`, alter the `query` as 
```php
/**
 * @deprecated 3.6, use Repo::task()
 */
public static function query(): EditorialTaskRepository
{
    return static::task();
}
```
more align with current changes .

controllers/grid/queries/form/QueryForm.php
```diff
@@ -543,21 +546,21 @@ public function execute(...$functionArgs)
...
         $addParticipants = [];
         foreach ($newParticipantIds as $userId) {
             $addParticipants[] = ([
-                'query_id' => $query->id,
+                'edit_task_id' => $query->id,
                 'user_id' => $userId
             ]);
         }
-        QueryParticipant::insert($addParticipants);
+        Participant::insert($addParticipants);
```
touhidurabir: same as above, can we use `createMany` ?

api/v1/submissions/formRequests/AddTask.php
```diff
@@ -0,0 +1,107 @@
...
+        ];
+    }
+
+    /**
+     * Prepare the data for validation.
+     */
+    protected function prepareForValidation(): void
+    {
+        $this->merge([
+            'createdBy' => Application::get()->getRequest()?->getUser()?->getId(),
```
touhidurabir, line 49: wondering if we can start using `Auth::user` instead of `Application::get()->getRequest()` to get the authenticated user ?

classes/editorialTask/EditorialTask.php
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function getSettingsTable(): string
+    {
+        return 'edit_task_settings';
+    }
+
```
touhidurabir, line 81: usually we define a `getSettings` method to have specific setting data but it's missing here , so are we going to have any number of settings ? just asking as I am not sure .

and also `getMultilingualProps` if there are any multilingual ones .

classes/migration/upgrade/v3_6_0/I10406_EditorialTasks.php
```diff
@@ -0,0 +1,139 @@
...
+
+            // Identify and set all users created a discussion
+            DB::table('notes as n')
+                ->select(['n.assoc_id', 'n.user_id', 'n.date_created'])
+                ->joinSub(
+                    // Get the first (latest) note for each query
+                    DB::table('notes as nt')
+                        ->select('nt.assoc_id')
+                        ->selectRaw('MIN(nt.date_created) as min_date')
+                        ->where('nt.assoc_type', '=', 0x010000a) // ASSOC_TYPE_QUERY
```
touhidurabir, line 53: any particular reason not to use like `Application::ASSOC_TYPE_QUERY` ?

touhidurabir (commented):
@Vitaliy-1 sorry about the delay, I have added few comments. None of these are major except 1/2 that may need a bit consideration, please take a look .

api/v1/submissions/PKPSubmissionController.php
Vitaliy-1: Yes, just added

api/v1/submissions/PKPSubmissionController.php
Vitaliy-1: Here it doesn't matter. For individual operations I usually prefer Model's methods over Eloquent Builder's

api/v1/submissions/formRequests/AddTask.php
Vitaliy-1: Interesting. I see that there is `registerUserSession()` and it's implemented for usual login: https://github.com/pkp/pkp-lib/blob/main/classes/security/Validation.php#L53. But it doesn't look to be implemented for API. Probably, there should be similar code when we check the token.

touhidurabir: you are right . lets go with usual approach for now and have a issue for 3.6 to update user token set process via API . we should add a fallback code for api key/token based authentication

classes/migration/upgrade/v3_6_0/I10406_EditorialTasks.php
Vitaliy-1: Yes, I recall that was our agreement to hard code values in migration scripts

PR #11588, opened 2025-07-01
Update user.register.form.usernameAlphaNumeric in pt_BR

diegoabadan:
to add the requirement to use lowercase letters, which was omitted in the previous translation for OJS 3.3


asmecher:
Thanks, @diegoabadan!

Issue #11586, opened 2025-06-30
[DOI Deposit] | XMLs generated are not well-formed or does not validate:


mbakioglu:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

In OJS 3.5, the XML files generated for DOI registration are not well-formed and fail validation when submitted to the Crossref system. As a result, the XML files are unable to pass through the parser, which causes the DOI registration process to fail.

### Steps to Reproduce

1- Go to the OJS 3.5 test instance:
https://alec.testing.publicknowledgeproject.org/ojs-3_5_0-docs/index.php/publicknowledge/en/

2- Select a published submission and use the Crossref Export Plugin to export the XML file.

3- Navigate to the Crossref XML parser tool:
 https://www.crossref.org/XSDParse

4- Upload the exported XML file.

5- Observe the following validation error returned by the parser:

<record_diagnostic status="Failure" msg_id="29">
   <doi />
   <msg>Deposited XML is not well-formed or does not validate: Error on line 89: Content is not allowed in trailing section.</msg>
</record_diagnostic>

Note: The line number (e.g., line 89) points to the final line of the XML file.

![Image](https://github.com/user-attachments/assets/afe9bf50-f0c8-4756-8e7c-ec01cfc1710e)

### Expected Result

The exported XML should be well-formed and valid according to Crossref schema standards, allowing it to successfully pass through the Crossref parser and complete the DOI registration process without errors.

### Actual Result

The XML file fails validation when deposited.

### Application Version

OJS 3.5

Tribunal33:
After deleting this trialing character `[]` I was able to successfully upload.  Not sure why we are now adding this new character to the XML export.

![Image](https://github.com/user-attachments/assets/614bd435-0c82-48a8-b6b5-cacfd765854c)

kaitlinnewson:
Hi @Tribunal33, are you able to see if any errors are logged when the Crossref XML is generated in this test install? I'm not able to reproduce this locally but can see it in the test install export. And is it using the latest 3.5.0 code?

Tribunal33:
@kaitlinnewson I got Alec to pull the latest code for that VM to check that as well.  I don't have access to those logs, but I can repro on my 3.5.0 VM and I'm not getting any errors.  Seems like it is appending an end of file character to the xml doc perhaps default charset has changed?

kaitlinnewson:
[...]

PR #11596, opened 2025-07-03
pkp/pkp-lib#11586 fix empty json responses appended to exported xml

kaitlinnewson:


api/v1/dois/PKPDoiController.php
```diff
@@ -688,20 +689,21 @@ function (DoiException $item) {
 
     /**
      * Download exported DOI XML from temporary file ID
+     * @throws BindingResolutionException
      */
-    public function getExportedFile(Request $illuminateRequest): JsonResponse
+    public function getExportedFile(Request $illuminateRequest): Response
```
taslangraham, line 694: Just curious to know the reason for using `Response` here instead of `JsonResponse`?

asmecher: `$response()->noContent` returns a `Response` object, not a `JSONResponse`. Because the `$tempFileManager->downloadById` function sends the file as output to the browser, we need this function to not create any further output.

Issue #11583, opened 2025-06-30
[PHPUnit] | GitHub Action tests are missing locale keys and passing for wrong reasons


Tribunal33:
**Description:**
- Include specific context details that have been discovered that surround this developer task.
- Link related tickets

I'm not well versed in PHP so this might not be an issue but I've noticed that the GitHub Actions runs are showing missing locale keys for Validator tests.  I think the reason is that it passes is because it is asserting false but I don't think it is failing for the right reasons.  Could we take a look at these validator AssertFalse tests so they aren't throwing so many missing keys.

GitHub Actions
![Image](https://github.com/user-attachments/assets/6fead9c3-9153-46f0-a5e7-07610be489e3)

Sample code where they are being tested in the PHPunit Tests

![Image](https://github.com/user-attachments/assets/d8443012-2c37-465c-96a9-d246c8964221)

**Steps to test the changes:**
1. Run the phpunit tests and it should pass validator tests with correct validation rule. 

**Solution:** What the proposed solutions are and additional affected areas
Either we can remove these validator checks or we can add a rule so that it passes without missing locale keys.   I gave a reason for what is happening.  I'm not sure if this is expected but when going through these logs it seems like the test isn't doing what it should be doing.

**What application are you using?**
 OJS 3.5 and main but GitHub Actions all PHPunit test runs.

**Additional information**
Please add any screenshots, videos or other information we can use to understand this dev task report.

After parsing the key with the rule it returns with a key that isn't available so it concatenates to an unknown gibberish key like "validation.values.value.anything else".  This will return false because of a missing key and still assert true because the test is expected to fail.  I am not sure if the intent of this isValid method is correctly being asserted by this test

#### Stable-3.5.0
~~[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4966~~
~~[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1043~~
~~[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2056~~
~~[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11597~~
~~[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4978 [TEST ONLY]~~
~~[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2075 [TEST ONLY]~~
[...]

PR #11597, opened 2025-07-03
pkp/pkp-lib#11583 Missing locale key logging handle at validation

touhidurabir:
for pkp/pkp-lib#11583 .


classes/core/ValidationServiceProvider.php
```diff
@@ -176,12 +185,14 @@ protected function resolve(array $data, array $rules, array $messages, array $at
                              */
                             protected function getMessage($attribute, $rule)
                             {
+                                $this->rehydrateValidationErrorMessages();
+
```
touhidurabir, line 189: This is required to rehydrate the validation message with our custom rules related validation messages .

Issue #11594, opened 2025-07-02
[Site-wide Language] | Reader profile language, UI Language change not possible if allowed_hosts is not enabled


bozana:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

s. https://forum.pkp.sfu.ca/t/ojs-3-5-error-when-change-the-language/96296

PKPPageRouter::isAllowedHost() does not consider that allowed_hosts in the config.inc.php could be disabled or empty string.

### Steps to Reproduce

1. Comment out allowed_hosts in the config.inc.php or set it to ""
2. Log in as a user in a journal with multiple UI languages
3. Try to change the UI language in the backend

### Expected Result

The UI language changes

### Actual Result

PHP Fatal error:  Uncaught TypeError: array_map(): Argument #2 ($array) must be of type array, null given in ojs-stable-3_5_0/lib/pkp/classes/core/PKPPageRouter.php:568

### Application Version

OJS, OMP, OPS version 3.5.0

<hr>

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/11599
- https://github.com/pkp/ojs/pull/4967 (only submodule update)
- https://github.com/pkp/omp/pull/2050 (only submodule update)
- https://github.com/pkp/ops/pull/1037 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/11623

bozana:
@touhidurabir, could you please review the PR above? Thanks a lot!

bozana:
Hi @Tribunal33,
The fix is merged, so the issue is all yours :-)
Thanks a lot!

Tribunal33:
This passes QA on main and stable 3.5.0.  I tested on both OJS and OMP for both branches with and without the fix.  Once to reproduce the problem and once with the fix which was confirmed following the removal of allowed_hosts config setting.  The language changes do work as expected without this config file setting removed or empty `""`.

PR #11599, opened 2025-07-04
pkp/pkp-lib#11594 consider disabled and empty allowed_hosts config, w…

bozana:
…hen checking redirect URL

s. https://github.com/pkp/pkp-lib/issues/11594

### DIFF
--- a/classes/core/PKPPageRouter.php
+++ b/classes/core/PKPPageRouter.php
@@ -560,11 +560,13 @@ class PKPPageRouter extends PKPRouter
      */
     private function isAllowedHost(string $protocol, string $redirectUrl): bool
     {
+        $allowedHosts = Config::getVar('general', 'allowed_hosts');
+        if ($allowedHosts == '') { // disabled (null) or empty string ''
+            return true;
+        }
         if (str_starts_with($redirectUrl, $protocol)) {
             $redirectUrl = preg_replace('/^' . preg_quote($protocol . '://', '/') . '/', '', $redirectUrl);
-            ;
         }
-        $allowedHosts = Config::getVar('general', 'allowed_hosts');
         $allowedHosts = array_map(strtolower(...), json_decode($allowedHosts));
         foreach ($allowedHosts as $allowedHost) {
             if (str_starts_with(strtolower($redirectUrl), $allowedHost)) {


classes/core/PKPPageRouter.php
```diff
@@ -560,11 +560,13 @@ private function _setLocale(PKPRequest $request, ?string $setLocale): void
...
+        $allowedHosts = Config::getVar('general', 'allowed_hosts');
+        if ($allowedHosts == '') { // disabled (null) or empty string ''
+            return true;
+        }
         if (str_starts_with($redirectUrl, $protocol)) {
             $redirectUrl = preg_replace('/^' . preg_quote($protocol . '://', '/') . '/', '', $redirectUrl);
-            ;
         }
-        $allowedHosts = Config::getVar('general', 'allowed_hosts');
         $allowedHosts = array_map(strtolower(...), json_decode($allowedHosts));
```
touhidurabir, line 570: This will still cause issue if someone set the `allowed_hosts` to some string or invalid data as `json_decode` will return `null` . If I understand correctly , `allowed_hosts` should be valid structure which is can be json decoded . Wondering rather than allowing any kind of format which will cause issue, should we force check above if `allowed_hosts` is and also properly json decodable, only then run the array mapping ?  and in that case , probably do error log or perhaps exception(if we want to be very strict) ?

touhidurabir (commented):
@bozana just one question .

classes/core/PKPPageRouter.php
bozana: I think the AllowedHostsPolicy (https://github.com/pkp/pkp-lib/blob/stable-3_5_0/classes/security/authorization/AllowedHostsPolicy.php) will decline the access to the web site even earlier if the allowed_hosts is some invalid structure that cannot be json decoded. 
However, maybe I can add the check here and log an error if it cannot be json decoded, as you suggested...

bozana: Hi @touhidurabir, I have just tested and thought about this once again: I think there is no much sense in handling it here -- if there is something wrong in the allowed_hosts, that policy will handle it i.e. the error will pop up earlier. And it is probably not going to happen, that someone just changes this setting manually before changing the UI language.
What do you think? 
Maybe I am thinking wrong and/or missing something...

touhidurabir: @bozana you are right , we can leave it to that policy .

touhidurabir (approved):

Issue #10034, opened 2024-06-10
Role assignment checkboxes behave unpredictably in some PHP environments


asmecher:
**Describe the bug**
From [the support forum](https://forum.pkp.sfu.ca/t/participant-permissions-not-automatically-enabled-for-journal-editors-ojs-3-3-0-16/87872):
> Users who have the Journal editor role and are assigned as participants on articles do not automatically have “Allow this person to make changes to the publication, such as the title, abstract, metadata and other publication details…” checked under Edit Assignment.

The options to restrict user activities on assignment shouldn't even be presented for journal editor/manager roles.

**To Reproduce**
This seems to be specific to certain PHP environments in cases where we do strict tests in PHP (`===` / `!==`). Different manifestation of the same problem e.g. [here](https://forum.pkp.sfu.ca/t/ojs-3-4-0-1-authors-unable-to-initiate-discussion-in-workflow/79738/15).

I was able to induce the same behaviour in my PHP environment using (prior to the affected database fetches):
```
\Illuminate\Database\Capsule\Manager::schema()->getConnection()->getPdo()->setAttribute(PDO::ATTR_EMULATE_PREPARES,true);
\Illuminate\Database\Capsule\Manager::schema()->getConnection()->getPdo()->setAttribute(PDO::ATTR_STRINGIFY_FETCHES,true);
```
This causes some numeric data to come back from the database in string form.

**What application are you using?**
 OJS, OMP or OPS version 3.3.0-x

**Additional information**
It appears not to affect `EntityDAO`-based implementations, as the schema contains type information that results in an explicit cast.

asmecher:
This doesn't appear to affect `stable-3_4_0` or `main`, as the `UserGroup` implementation has been moved to SchemaDAO.

PR #11605, opened 2025-07-07
Cast userId column to ensure auth logic works properly

kennydude:
We've had an issue where "Login As" doesn't show up correctly, and when tracing this through I found it to be caused by permission checks not quite working at https://github.com/pkp/pkp-lib/blob/f9e374720827d15b8e14a0ca93cf81a161011e3e/classes/security/Validation.php#L519

Following this through, it's seemingly caused by the Laravel/Eloquent model loading in the column for `userId` as a string and not an integer, so this small PR ensures the comparison is between an integer and an integer,

### DIFF
--- a/classes/userGroup/relationships/UserUserGroup.php
+++ b/classes/userGroup/relationships/UserUserGroup.php
@@ -33,6 +33,7 @@ class UserUserGroup extends \Illuminate\Database\Eloquent\Model
     protected $casts = [
         'dateStart' => 'datetime',
         'dateEnd' => 'datetime',
+        'userId' => 'int',
     ];
 
     public function user(): Attribute


asmecher:
@kennydude, thanks for tracking this down through a doubtlessly huge call stack... This is probably related to [#10034](https://github.com/pkp/pkp-lib/issues/10034), which saw similar behaviour with some PHP environments where some data that was expected to come back from the database as an `int` came back `string` instead. I haven't been able to pin down why some MySQL/PHP environments do this and others don't.

The fix looks good, but can you clarify what version of OJS you're using, and maybe a little about your server stack (PHP, MySQL version, etc?)

kennydude:
We're running OJS 3.5.0 (the latest released version) and the server is running `10.6.22-MariaDB-cll-lve-log - MariaDB Server` according to phpMyAdmin which is on a cPanel-based webhost

asmecher:
Thanks, @kennydude, I've merged and ported to `stable-3_5_0` for the 3.5.0-1 build!

Issue #11603, opened 2025-07-07
[Invitation] | User role assignment email template missing at upgrade to 3.5.0


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Original forum post https://forum.pkp.sfu.ca/t/unable-to-invite-user-in-ojs-3-5-0/96162/10 . 

after the upgrade from 3.4.0-x to 3.5.0 , when trying to invite an user to a certain role, the email template is missing at the final stage of invitation . 

### Steps to Reproduce

1. Checkout the latest codebase of 3.5.0
2. import `stable-3.4.0` database from the dataset
3. run the upgrade process as `php tools/upgrade.php upgrade`
4. After successful upgrade, login and try to invite an user to a certain role (e.g. Author, Journal Editor, etc)
5. See the error popup `The email template you requested was not found.` at final step along with empty email body and subject. 

### Expected Result

Should have the pre filled email subject/body and no error popup . 

### Actual Result

error popup `The email template you requested was not found.` along with empty email body and subject. 

### Application Version

OJS (probably also OMP/OPS) 3.5.0

### Additional Information

checking into the database table `email_templates_default_data`, there is no row of column `email_key` with value `USER_ROLE_ASSIGNMENT_INVITATION` . for the upgrade process, there should be migration to have the pre filled . 

However, this does work for a fresh installation of 3.5.0 . 

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11607
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4975 
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2057
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1044 

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11786
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5081
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2113
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1090

touhidurabir:
[...]

PR #11607, opened 2025-07-08
pkp/pkp-lib#11603 Missing email temaple for role assignment invitation

touhidurabir:
for pkp/pkp-lib#11603 .

### DIFF
--- a/classes/migration/upgrade/v3_5_0/I11603_AddMissingUserRoleAssignmentInvitationEmail.php
+++ b/classes/migration/upgrade/v3_5_0/I11603_AddMissingUserRoleAssignmentInvitationEmail.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_5_0/I11603_AddMissingUserRoleAssignmentInvitationEmail.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I11603_AddMissingUserRoleAssignmentInvitationEmail
+ *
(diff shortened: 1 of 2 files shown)


classes/migration/upgrade/v3_5_0/I11603_AddUserRoleAssignmentInvitationEmail.php
```diff
@@ -0,0 +1,45 @@
...
+
+class I11603_AddUserRoleAssignmentInvitationEmail extends Migration
+{
+
+    /**
+     * @inheritDoc
+     */
+    public function up(): void
+    {
+        Repo::emailTemplate()->dao->installEmailTemplates(
```
Vitaliy-1: Can we just re-use `InstallEmailTemplates` to rely solely on the query builder, avoiding using external methods? https://github.com/pkp/ojs/blob/fa1efaeb0cd58492a33bb80f3db0703929d9ce2b/dbscripts/xml/upgrade.xml#L136

touhidurabir: @Vitaliy-1 my motivation for this approach came form few other migrations such as `COA75_AddUserRoleEndEmail`, `I9771_OrcidMigration` but your suggestion is better . Updated it . 

I have also added a block in `upgrade.xml` as 
```xml
<upgrade minversion="3.5.0.0" maxversion="3.5.0.1">
        <migration class="PKP\migration\upgrade\v3_5_0\I11603_AddMissingUserRoleAssignmentInvitationEmail"/>
</upgrade>
```
to handle the upgrade case `3.5.0-0 --> 3.5.0-1` but doubting if this is necessary as only few people probably upgraded to `3.5.0-0` . But again those who upgraded, need to run some sql to address this issue .

Issue #11327, opened 2025-04-28
Add front end UI for commenting in themes


ewhanson:
[To be updated]

ewhanson:
@Devika008, the frontoffice UI for public comments is ready for the mockup stage. Would you be able to prepare them? Please let me know if you have any questions or need any additional info about it. Thanks!

taslangraham:
@Devika008 @ewhanson Just a quick note: the API returns comments as paginated data, so the UI mockup should also include a “View more” button (or similar) to let users load additional comments.

taslangraham:
Ready for review @ewhanson. PRs are listed at https://github.com/pkp/pkp-lib/issues/11327#issue-3025465513.

@jardakotesovec Could you take a look at these PRs as well, especially the [ui-library PR](https://github.com/pkp/ui-library/pull/659)? I’ve added a few more icons and components to the frontend UI, along with minor updates to some of the components you had already added. I’m also using the `useDate` and `useLocalize` composables from the backend UI, but I can create frontend-specific versions if you think that would be better.

taslangraham:
@jardakotesovec I've merged the PRs for public comments.

Thanks to both of you @jardakotesovec and @ewhanson for the reviews

jardakotesovec:
@taslangraham Hi, finally have my PRs ready. Can you please review them?

Intended changes:
- delete as much as possible from the vue.js components that were brought from backend. For example having just one dialog component now, to keep things simple and scalable across themes
- moved the styling to the default theme, while the components don't have any styling (headless)
- I made some choices about the styling based on what we have in default theme and I checked with Devika some of my choices to make sure its good enough for now.
- I moved the comments to the left column as its in designs, following the mock up in this case. Let me know if there was some decision to change that. Since we don't expect much content there I think its better if its taking up less space and its not push down by the content in the right column
- I make the version indication consistent with how we indicate versions currently on the right side. This might be subject to change..  but it would be changed on both places.  More details here - https://github.com/pkp/pkp-lib/issues/11790
- I moved the logic to the store, which makes it easier to connect to that data source from multiple places (scrolling component as well) and possibly extend things from plugin in future. Also its easier to find the logic given the amount of the components for the headless version

ojs: https://github.com/pkp/ojs/pull/5088
pkp-lib: https://github.com/pkp/pkp-lib/pull/11812
ui-library: https://github.com/pkp/ui-library/pull/702/

taslangraham:
[...]

PR #11620, opened 2025-07-09
WIP: pkp/pkp-lib#11327 Implement public facing UI for user comments

taslangraham:
For pkp/pkp-lib#11327

### DIFF
--- a/classes/template/PKPTemplateManager.php
+++ b/classes/template/PKPTemplateManager.php
@@ -802,8 +802,9 @@ class PKPTemplateManager extends Smarty
         );
     }
 
-    public function requiresVueRuntime() {
-        if(!$this->isVueRuntimeIncluded) {
+    public function requiresVueRuntime()
+    {
+        if (!$this->isVueRuntimeIncluded) {
             $this->isVueRuntimeIncluded = true;
             $baseUrl = $this->_request->getBaseUrl();
 
@@ -815,7 +816,17 @@ class PKPTemplateManager extends Smarty
                     'contexts' => ['frontend']
                 ]
             );
-    
+
+            $request = Application::get()->getRequest();
+            // Include css styles used by UI-Library Vue.js components
+            $this->addStyleSheet(
+                'build',
+                $request->getBaseUrl() . '/styles/build.css',
+                [
+                    'priority' => self::STYLE_SEQUENCE_LAST,
+                    'contexts' => 'frontend',
+                ]
+            );
         }
     }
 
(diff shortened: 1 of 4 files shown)


classes/template/PKPTemplateManager.php
```diff
@@ -815,7 +816,17 @@ public function requiresVueRuntime() {
...
+            $request = Application::get()->getRequest();
+            // Include css styles used by UI-Library Vue.js components
+            $this->addStyleSheet(
+                'build',
+                $request->getBaseUrl() . '/styles/build.css',
+                [
+                    'priority' => self::STYLE_SEQUENCE_LAST,
+                    'contexts' => 'frontend',
+                ]
+            );
```
taslangraham: This is a workaround, and might need to be changed. I added it so that the styles for the ui-library components are available when using said components in the front-office.

@jardakotesovec do you think this an appropriate approach or should it be done differently?

classes/template/PKPTemplateManager.php
```diff
@@ -1466,7 +1478,7 @@ public function display($template = null, $cache_id = null, $compile_id = null,
             $output,
             [
                 'priority' => self::STYLE_SEQUENCE_NORMAL,
-                'contexts' => ['backend'],
+                'contexts' => $this->isVueRuntimeIncluded ? ['backend','frontend'] : ['backend'],
```
taslangraham: Done so that the `pkpAppData` is available in vue components on the front-office

classes/components/UserCommentComponent.php
```diff
@@ -0,0 +1,133 @@
...
+    public function getConfig()
+    {
+        return [
+            'publicationIds' => $this->publishedPublicationIds->sortDesc()->values(),
+            'latestPublicationId' => $this->submission->getCurrentPublication()->getId(),
+            'itemsPerPage' => Repo::userComment()->getPerPage(),
+            'loginUrl' => $this->getLoginUrl(),
+            'allCommentsCount' => $this->allCommentsCount,
+            'commentsCountPerPublication' => $this->commentsCountPerPublication,
+        ];
```
jardakotesovec, line 122: Still wondering, whether we could just have the good enough API to get most of these things from the API. 

Specifically  publicationIds, latestPublicationId, allCommentsCount and commentsCountPerPublication are very data oriented and something thats good fit for API.  The other ones are more matter of configuration.. so having that option to configure these server side is good convenience, but it would not prevent someone to implement this and just rely on the API.

@ewhanson @asmecher To iterate my question from last review. I understand we might want to take some shortcuts because of the timelines or current lack of the front office APIs. But would really appreciate your comment on this, whether you care about this or not.

asmecher: I don't have super well formed opinions about this, but on further reflection, I think you're right -- we should be careful about loading down `getConfig()` with things that should be in the API. It's a half-step away from our old server-side rendering patterns and we should be cleanly moving towards building out the API with the new VueJS stuff.

PR #11625, opened 2025-07-09
10553

bozana:


classes/publication/Repository.php
```diff
@@ -574,10 +575,29 @@ public function publish(Publication $publication)
...
 
+        $context = $submission->getData('contextId') === Application::get()->getRequest()->getContext()?->getId()
+        ? Application::get()->getRequest()->getContext()
+        : app()->get('context')->get($submission->getData('contextId'));
+
         // Mark DOIs stale (if applicable).
-        if ($newPublication->getData('status') === Publication::STATUS_PUBLISHED) {
-            $staleDoiIds = Repo::doi()->getDoisForSubmission($newPublication->getData('submissionId'));
-            Repo::doi()->markStale($staleDoiIds);
+        if ($newPublication->getData('status') === Submission::STATUS_PUBLISHED) {
```
ewhanson: I'm not sure why it would have been `Publication::STATUS_PUBLISHED` before instead of `Submission::STATUS_PUBLISHED`, but is there a difference here? Are we in general using the `Submission` ones instead? Not a problem, just curious.

api/v1/dois/PKPDoiController.php
```diff
@@ -299,14 +303,25 @@ public function edit(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_NOT_FOUND);
         }
 
+        // All minor versions of the same submission, version stage and version major have one, the same DOI.
+        // So get all the minor versions that have the same DOI.
+        // Eventuell we could just edit all the same DOIs, but this way we eventually keep
```
ewhanson:
```suggestion
        // Eventually we could just edit all the same DOIs, but this way we eventually keep
```

classes/publication/Repository.php
bozana: Actually you are right -- this should be Publication::STATUS_PUBLISHED -- now they are the same, but maybe they will differ in the future, and I believe we should rather watch for publication statuses here...

Issue #11631, opened 2025-07-10
Add ability for plugin to attach authorization policies to API controller


touhidurabir:
**Description:**
As part of https://github.com/pkp/pkp-lib/issues/9434 , we have introduced the mechanism for plugin to add/override API routes . Initially there was 2 different approach and API controller override mechanism allow the override the controller's `authorize` method to add/override new policy authorizer. However this mechanism has been removed because of a certain drawback , see details at https://github.com/pkp/pkp-lib/issues/11609 . 

We need some kind of enhancement in the available approach to pass one or more authorization policies to attach from plugins . 

**Solution:** 
The proposed solution allow a callable param to the `APIHandler::addRoute`  which will follows the same method signature as the core `PKPBaseController::authorize` which will return an array to attachable policy authorizer from the plugin in the following manner 
```php
Hook::add('APIHandler::endpoints::submissions', function(string $hookName, PKPBaseController &$apiController, APIHandler $apiHandler): bool {  
  $apiHandler->addRoute(
  	'POST',
  	'{submissionId}/plagiarism/status',
  	function (\Illuminate\Http\Request $request) use ($apiController): JsonResponse {
  
  		$submission = $apiController->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
  
  		// return response after required logic
  	},
  	'submission.plagiarism.status',
  	[
  		Role::ROLE_ID_SITE_ADMIN,
  		Role::ROLE_ID_MANAGER,
  		Role::ROLE_ID_SUB_EDITOR,
  		Role::ROLE_ID_ASSISTANT,
  	],
  	new class implements HasAuthorizationPolicy
	{
		public function getPolicies(PKPRequest $request, array &$args, array $roleAssignments): array
		{
			return [
				new SubmissionAccessPolicy($request, $args, $roleAssignments),
				new SubmissionCompletePolicy($request, $args),
			];
		}
	}
  );
  
  return Hook::CONTINUE;
});
```

**What application are you using?**
  OJS, OMP or OPS version 3.5.0-2 and main

#### stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11632
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4984 [TEST ONLY] 
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2073 [TEST ONLY]
[...]

PR #11632, opened 2025-07-10
pkp/pkp-lib#11631 Ability to inject policies in api route from plugin

touhidurabir:
for pkp/pkp-lib#11631 .

### DIFF
--- a/classes/handler/APIHandler.php
+++ b/classes/handler/APIHandler.php
@@ -20,6 +20,7 @@ use PKP\core\PKPBaseController;
 use PKP\core\PKPRoutingProvider;
 use PKP\middleware\HasRoles;
 use PKP\plugins\Hook;
+use PKP\plugins\interfaces\HasAuthorizationPolicy;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Throwable;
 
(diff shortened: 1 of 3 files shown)


classes/middleware/PolicyAuthorizer.php
```diff
@@ -52,6 +53,14 @@ public function handle(Request $request, Closure $next)
         $args = [$request];
         $roleAssignments = $this->getRoleAssignmentMap($router->getRoutes());
 
+        // if route has extra policy authorizer from plugin, add those to current policy stack
+        if ($policyAuthorizer = $currentRoute->getAction('policyAuthorizer')) { /** @var callable $policyAuthorizer */
+            $policies = $policyAuthorizer($pkpRequest, $args, $roleAssignments->toArray()); /** @var array $policies */
+            foreach ($policies as $policie) { /** @var AuthorizationPolicy|PolicySet $policie */
+                $routeController->addPolicy($policie);
+            }
```
touhidurabir: I am thinking , perhaps we should wrap it in a `try/catch` block as it's not possible in PHP to force a callable/closure to implement any interface . we can catch the error and log it and in that process, plugins will not break the app . 

also probably the `instanceof` checking against each of the `policy` and if does not match , a error log so what plugin does not break the app .

touhidurabir: Or it can be better(and more strict) if we take an instance of certain contract like `new class implements ....` instead of callback ? probably bit few lines of code for plugin developer but more structured with certain contract .

ewhanson: I think your stricter suggestion is a good approach. It may be a few more lines for plugin developers, but better to have a safer way to do this that's less likely to break things and easier to debug.

ewhanson (changes requested):
Hey @touhidurabir, this overall looks good to me, and I agree with your suggestion on adding a new contract to improve the access to this via plugins. Let me know once you have that implemented and I can have another look.

taslangraham:
@touhidurabir My availability is a little patchy today, but I'll review this either tonight or first thing tomorrow

classes/plugins/interfaces/HasAuthorizationPolicy.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @file classes/plugins/interfaces/HasAuthorizationPolicy.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class HasAuthorizationPolicy
+ *
+ * @brief Provides an interface for plugins which allows the to injection of policies in the API request
```
taslangraham:
```suggestion
 * @brief Provides an interface for plugins which allows the injection of policies in the API request
```

classes/handler/APIHandler.php
```diff
@@ -215,6 +225,14 @@ protected function registerRoute(): void
 
                 $route->middleware($middleware);
             }
+
+            // If route has extra policy authorizer from plugin, add those to route's action stack with custom key
+            // which will be later added in the route controller's authorization stack via `PolicyAuthorizer` global middleware
+            if ($routeParams['policyAuthorizer'] && $routeParams['policyAuthorizer'] instanceof HasAuthorizationPolicy) {
+                $route->setAction(array_merge($route->getAction(), [
+                    'policyAuthorizer' => $routeParams['policyAuthorizer']
```
taslangraham: Would it make sense to prefix this with `plugin` to make it clearer that it comes from a plugin? That way, it’d be named `pluginPolicyAuthorizer`, which might make its origin more obvious.

taslangraham (approved):
@touhidurabir This looks good! I've approved it, just left two minor suggestions that you can address without needing another review round.

Issue #11589, opened 2025-07-02
[Implementation] Consider JAV versions for indexing


bozana:
Now that we have introduced JAV and versions (s. https://github.com/pkp/pkp-lib/issues/10669), and that we are introducing different DOIs for different versions (s. https://github.com/pkp/pkp-lib/issues/10553), we need to consider versions for indexing i.e. in export tools.
Currently only the requirement exist to export/deposit/index version of records (VoRs), so for now we will consider only published VoRs for export and indexing.
We will use the existing ImportExportPlugin and grid tools, and (similarly to articles display) a publication/version tab in the export plugins. For that ExportPublishedPublicationsListGridHandler, similar to ExportPublishedSubmissionListGridHandler, needs to be implemented.

It seems this is only needed for OJS -- OMP and OPS do not have any export plugins from this kind.

Explanation about what the words/letters, that are used below, mean:
A = one DOI for all versions
B = one DOI for a version
settings = e.g. status

Note: One and the same DOI is assigned to all minor versions with the same version stage and version major. Thus, all these minor versions are considered as one objects for indexing services.

Listen and act accordingly to the activities in the workflow:

publication()::version() -- The settings are copied to the new publication object.
A: The export plugins save their settings for submission object (not publication), so nothing happens here
B: If it is a major version the settings will be set to null. Nothing to do for minor versions.

publication()::publish()
A: Mark the registered submission stale if the published publication is current publication (to update the registered metadata with the indexing service)
B: If it is a minor version of an already registered version (it will have the copied status registered from the previous version), and it is the last minor version, mark this publication stale (to update the metadata and to point to the new current/latest minor publication). If it is a major version do nothing -- the version has no status (because it was removed by versioning), so it will be registered at next registration attempt. The older versions do not need to be updated => only new versions will point/relate to the older ones, and not vice versa.

publication()::unpublish()
Note: this should actually not happen!
A: If it is a current publication, that was registered or marked registered, mark submission stale (to eventually update the metadata)
B: If it was the last minor version, and if it was registered or marked registered, mark the next previous published minor version stale (to update the metadata to point (again) to that next previous minor version). 
(Note:
A: if all publications are unpublished 
of
B: if a major publication is unpublished:
We cannot remove already indexed items from indexers, so this case should actually be retraction notice on the submission landing page).

Edge cases: 
If A is enabled after B:
[...]

Issue #10553, opened 2024-10-24
[Implementation] Different DOIs for different article versions


bozana:
The DOI backend allows assignment of different DOIs to different article versions. This option currently only exists in OPS. For the project OSS ORE an assignment of different DOIs to different article versions is needed also in OJS.

The decisions made:
All JAV will get a DOI, so that preprints can get a DOI.
The minor versions in the same version stage and same version major will have one and the same DOI. This requires changes in the current code.

Only current publication is displayed in submission listing on the DOI management page (as it was till now). The link 'View all' opens a modal where the last minor version for all major versions will be listed. This is slightly different as till now (in OPS, where all versions are listed and all versions get a DOI). 
We cannot display all versions i.e. also all minor versions, because each version can be edited, which could lead in wrongly editing minor versions (so that it could happen that they have different DOIs) i.e. we would need to consider this logic also in the edit function in UI (so that when editing a minor version all sibling minor version would be edited in the same way). This is a bigger task (for me) and I think does not need to be implemented right now.

Tasks to do:
- [x] Add setting for a journal to define if each article version will get a new DOI
- [x] Consider it on the DOI management page
- [x] Consider it for display on the article landing page. If a minor version does not have a DOI, we check if there is a sibling minor version, to display the DOI from there, because all minor versions (of a submission and with the same version stage and major) have the same DOI. This is considered also for chapters, but not for galleys, publication formats and files, because it is not possible to find out the sibling minor version of them. Currently we do not display DOIs for galley and files, and DOIs for these object types are not much used.
- ~~[ ] Consider it in native import export plugin~~ -- The DOIs will be exported for publications (and other objects) that have a DOI in the DB. Differently to the DOI display on the landing page, if a minor version do not have a DOI, it will not be searched for a DOI of a sibling minor version for export/import. 

PR:

s. comment regarding development notes: https://github.com/pkp/pkp-lib/issues/10553#issuecomment-3045723046

- https://github.com/pkp/pkp-lib/pull/11625
- https://github.com/pkp/ui-library/pull/690
- https://github.com/pkp/ojs/pull/4982
- https://github.com/pkp/omp/pull/2091
- https://github.com/pkp/ops/pull/1076

Fix marking DOIs stale:
- https://github.com/pkp/pkp-lib/pull/11835
- https://github.com/pkp/ojs/pull/5099
- https://github.com/pkp/omp/pull/2124
- https://github.com/pkp/ops/pull/1099

Another fix marking DOIs stale:
[...]

PR #11635, opened 2025-07-11
pkp/pkp-lib#11589 Consider JAV versions for indexing

bozana:
s. https://github.com/pkp/pkp-lib/issues/11589

The new introduces function getMinorVersionsSettingValues() is used in the DOAJ plugin.

### DIFF
--- a/classes/publication/DAO.php
+++ b/classes/publication/DAO.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/publication/DAO.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2000-2021 John Willinsky
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class DAO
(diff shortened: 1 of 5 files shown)


locale/en/grid.po
```diff
@@ -698,6 +698,12 @@ msgstr "{$userGroupName} role assigned to {$stageName} stage."
 msgid "grid.userGroup.unassignedStage"
 msgstr "{$userGroupName} role unassigned from {$stageName} stage."
 
+msgid "grid.publication.itemVoR"
+msgstr "VoR"
```
kaitlinnewson: Would abbreviating this present any issues for translations and/or users understanding what it means? I wonder if it should be written out, even though it's longer to display in the grid.

kaitlinnewson (approved):
Just one small comment on this one!

bozana:
Hi @kaitlinnewson, I added a new commit to this PR -- the possibility to filter publications by version stage and major -- used in PubObjectsExportPlugin::handlePublicationUnpublishing() (also needed for this issue https://github.com/pkp/pkp-lib/issues/10553, used in this PR: https://github.com/pkp/pkp-lib/pull/11625/commits/0cca1e19d3900a7b2d15257d0c35de17d7d5ad9a#diff-013b0fb476685dc4247dfa2b46fa4d770dfa07662e7d30c8d264774de1002744)

classes/publication/DAO.php
```diff
@@ -505,4 +508,20 @@ protected function setDoiObject(Publication $publication)
...
     }
+
+    /**
+     * Get setting values for the given setting name of all submission's minor versions in the given stage and a given version major
+     */
+    public function getMinorVersionsSettingValues(int $submissionId, string $versionStage, int $versionMajor, string $settingName): Collection
+    {
+        return DB::table('publication_settings AS ps')
+            ->join('publications AS p', 'p.publication_id', '=', 'ps.publication_id')
+            ->join('submissions AS s', 'p.submission_id', '=', 's.submission_id')
```
kaitlinnewson: Is the join on the `submissions` table needed?

bozana: Yes, you are right -- I thought in order to be sure the submission is there, but because now we have foreign keys, it is not needed...

kaitlinnewson (approved):

Issue #11643, opened 2025-07-16
Remove redundant calls to DB when loading a submissions list


Vitaliy-1:
**Description:**
In situations when submission doesn't have review assignments, stage assignments or decisions, Submission Schema tries to get them from the database additional time per each submission. See: https://github.com/pkp/pkp-lib/blob/2ae905c0070b213a42c4d854cb900005e88785ce/classes/submission/maps/Schema.php#L332-L336

This decreases performance. 

**Steps to test the changes:**
1. E.g., have user with an editor role assigned to the large amount of submissions
2. Open browser dev tools. 
3. Go to the dashboard an click on the assigned to me view
4. See the loading time for the request to the `_submissions` endpoint

**Solution:** What the proposed solutions are and additional affected areas
Remove redundant calls

**What application are you using?**
  OJS, OMP and OPS version 3.5 and main

Vitaliy-1:
PRs
main: https://github.com/pkp/pkp-lib/pull/11644
tests: https://github.com/pkp/ojs/pull/4994

stable 3.5.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/11738
tests: https://github.com/pkp/ojs/pull/5065

Vitaliy-1:
@jonasraoni, can you take a look at this PR. Also, do you think it worth to port it to 3.5, meaning if there is a substantial benefit? 

According to my test, it doesn't resolve much the problem with performance, although significantly reduces the number of db calls, those which are unnecessary.

jonasraoni:
Done! I don't think this is going to bring visible performance benefits, but given the update is small and that 3.5 is a LTS release, I think it's worth to do it.

asmecher:
@Vitaliy-1, this looks good to me. Is it ready to merge? If so, I'd suggest merging both for `stable-3_5_0` *and* `main` -- I'll need to do some quality testing for https://github.com/pkp/pkp-lib/issues/11682, which is also going to both, so now is good timing.

Vitaliy-1:
@Tribunal33, this addresses only performance, not sure if additional testing is required here.

kaitlinnewson:
Hi @Vitaliy-1, it looks like this may have broken tests in OMP `stable-3_5_0`, e.g.: https://github.com/pkp/omp/actions/runs/17247187578/job/48999773900?pr=2108

Vitaliy-1:
PRs
stable 3.5.0
[...]

PR #11644, opened 2025-07-16
pkp/pkp-lib#11643 Remove redunant database calls

Vitaliy-1:


classes/submission/maps/Schema.php
```diff
@@ -559,19 +562,21 @@ protected function mapByProperties(array $props, Submission $submission, bool|Co
...
         $canChangeMetadata = false;
 
         // Check if stage assignment is associated with the current user and edit metadata flag
-        foreach ($stageAssignments as $stageAssignment) {
-            if ($stageAssignment->userId === $currentUser->getId()) {
-                $isAssigned = true;
-                if ($stageAssignment->canChangeMetadata) {
-                    $canChangeMetadata = true;
-                    break;
+        if ($stageAssignments) {
```
jonasraoni: Unrelated to this change, but the comment on the line 588 and the following `if` sound suspicious: `// If user is ASSIGNED` then `if (!$isAssigned)`.

If it should be `if ($isAssigned)` then this line/update can also be simplified to `if (!$stageAssignments) { return false; }`

classes/submission/maps/Schema.php
```diff
@@ -858,72 +864,76 @@ protected function getPropertyStages(Enumerable $stageAssignments, Enumerable $r
         $isCurrentUserDecidingEditor = false;
 
         // Determine stage assignment related data
-        foreach ($stageAssignments as $stageAssignment) {
+        if ($stageAssignments) {
+            foreach ($stageAssignments as $stageAssignment) {
```
jonasraoni: To decrease the number of changes and the indentation I'd replace these lines by (can be applied to a similar update some lines above):

```suggestion
        foreach ($stageAssignments ?? [] as $stageAssignment) {
```

jonasraoni (approved):

classes/submission/maps/Schema.php
Vitaliy-1: Likely just the comment is wrong, the code logic looks correct to me

Issue #11576, opened 2025-06-27
Add backoffice UI for comment moderation


ewhanson:
## Requirements

There are three cases where comments will need editorial staff intervention via the backoffice.

1. When a comment is first posted, it must be approved to "ensure the quality of the comments" to prevent "spam, abusive language, or irrelevant content."
2. Users can report a comment as inappropriate, so editorial staff need to be able to review these reports in the context of the comments. 
3. Any hidden comments can be restored/unhidden.

For design/mockups, the following pages will be required:

1. A settings section where user comment-related things can live.
2. A list of incoming, new comments requiring approval (including comment text, user info?, publication/submission info/link), action to approve or hide.
3. A list of reported comments, the same info as above as well as the reporting user and and the text of the report. 
4. A way to view hidden comments, relevant details, and a way to unhide them.

## Acceptance Criteria

TBD

## PRs

UI-Library: https://github.com/pkp/ui-library/pull/664
pkp-lib: https://github.com/pkp/pkp-lib/pull/11647
ojs: https://github.com/pkp/ojs/pull/5006
omp: https://github.com/pkp/omp/pull/2066
ops: https://github.com/pkp/ops/pull/1053

ewhanson:
@Devika008, the backoffice UI for public comments is ready for the mockup stage. Would you be able to prepare them? Please let me know if you have any questions or need any additional info about it. Thanks!

taslangraham:
@Devika008 @ewhanson Below are screenshots of the initial UI design I've implemented for this to get things underway. Once we have final designs from @Devika008 then these will be updated accordingly.
 

#### **Table showing paginated list of comments**
<img width="895" height="383" alt="Image" src="https://github.com/user-attachments/assets/d48b37a3-d03f-4891-b6cf-b5040fb45862" />

_The "status" column on the table shows all statuses that the comment currently falls under. For example, a comment can be approved(publicly visible) and reported at the same time.  If said comment was to be hidden(unapproved) the the status in the table would be "needs approval, reported"._

The dropdown in the top right-hand allows the user to select which set of comments they'd like to view, i.e approved, need approval, reported

<img width="277" height="108" alt="Image" src="https://github.com/user-attachments/assets/c98c0ae4-2430-45bc-bf3c-6aa1b3caf913" />


## **Side modal is used to display comment details**
[...]

PR #11647, opened 2025-07-16
pkp/pkp-lib#11576 Add backoffice UI for comment moderation

taslangraham:
For pkp/pkp-lib#11576


locale/en/manager.po
```diff
@@ -3731,4 +3731,70 @@ msgid "plugins.importexport.common.validation"
...
+msgstr "Reason"
+
+msgid "manager.userComment.dateReported"
+msgstr "Date Reported"
+
+msgid "manager.userComment.approveThisComment"
+msgstr "Approve This Comment"
+
+msgid "manager.userComment.approveThisComment.description"
+msgstr "Approving a comment will make it visible to all users on the article page."
```
ewhanson: This shouldn't be "article" specific if it's in pkp-lib. It will also be visible to all visitors, not just authenticated users. I think it's probably okay to just say "on the page" or "on the site."

Issue #11637, opened 2025-07-14
[Installation] Replace RoR dataset import process with batch job


touhidurabir:
**Description:**
In the 3.5, we have introduced RoR functionality which added the process of importing all of RoR dataset from API at the install phase and also upgrade (2 times every month) of dataset via schedule task . As this RoR bit is quite big, downloading, extracting and importing process at the install time cause the application install time a bit . Also sometimes occasional failure to download the dataset with long running process to download cause 502 or timeout at install time . 

**Solution:** What the proposed solutions are and additional affected areas
If possible, break the importing process into smaller batch jobs . 

**What application are you using?**
  OJS, OMP or OPS version 3.5.0-1

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11652
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4998
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2064
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1051

touhidurabir:
@asmecher @bozana can you please review the PRs at https://github.com/pkp/pkp-lib/issues/11637#issue-3227790097 .

touhidurabir:
Hi @jonasraoni can you take a look at the PRs at https://github.com/pkp/pkp-lib/issues/11637#issue-3227790097 as seconds reviewer .

bozana:
Hi @touhidurabir, I took a quick another look and it looks good. Maybe we can wait for Jonas' review, then, if wished, I can test it once again...

bozana:
Maybe @jonasraoni have missed the message above? :-)

bozana:
Maybe to also note/ask something here:
Now, with these changes, we first introduce the possibility for a scheduled task and a job to write to the same log file. What if the scheduled task is run in the CLI mode (with a CLI user) and the job executed within a request (with the web-user)? The scheduled task will create the file (and I believe also some cache files) that then the job will not be able to access?
Thanks!

touhidurabir:
Before proceed with more work on this one, we need a resolution on pkp/pkp-lib#9345 first .

PR #11652, opened 2025-07-17
pkp/pkp-lib#11637 Refactor RoR dataset import to jobs

touhidurabir:
for pkp/pkp-lib#11637 .


classes/install/PKPInstall.php
```diff
@@ -192,10 +194,17 @@ public function createDirectories()
     public function createConfig()
     {
         $request = Application::get()->getRequest();
-        return $this->updateConfig(
+        $appKey = PKPAppKey::generate();
+
+        // Need to update the new app key immediately in the facade config after generation
+        $appConfig = FacadesConfig::get('app');
+        $appConfig['key'] = $appKey;
+        FacadesConfig::set('app', $appConfig);
```
touhidurabir, line 201: need to set the new app key upon set or change for any service provider to have the latest one .

classes/install/PKPInstall.php
```diff
@@ -221,6 +230,13 @@ public function createConfig()
                 ]
             ]
         );
+
+        // Need to re-register the encryption service provider after the new app key set as
+        // this will be used for signed url and closure serialization, see details at 
+        // \Illuminate\Encryption\EncryptionServiceProvider::EncryptionServiceProvider()
+        app()->register(new \PKP\core\PKPEncryptionServiceProvider(app()));
```
touhidurabir, line 236: still need to re-register as this already registered with no or older app key which will be used to sign the job closure or url signing later . But as we have updated the key , need to make sure it's have the latest one other will be unable to verify any singed url/closure in the middle of installation process .

classes/scheduledTask/ScheduledTask.php
```diff
@@ -157,6 +152,33 @@ public function execute(): bool
...
     }
+
+    /**
+     * Write an entry into the execution log.
+     *
+     * @param string $logFile Path to the log file.
+     * @param string $message A translated message.
+     * @param ?string $type One of the ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE... constants
+     */
+    public static function writeToExecutionLogFile(string $logFile, string $message, ?string $type = null): void
```
touhidurabir: as we do not have any good logging system and this task dispatch several jobs, we need a way to write log to task's log file . We can pass the task instance but that introduce a possible chances to handle closure in future . This static interface and new method `getExecutionLogFile` allow a mechanism to write logs to task's log file even after task complete and control passed to jobs for better log centralization for specific task .

jobs/BaseJob.php
```diff
@@ -65,22 +64,10 @@ abstract class BaseJob implements ShouldQueue
      */
     public function __construct()
     {
-        $this->connection = $this->defaultConnection();
+        $this->connection = Config::getVar('queues', 'default_connection', 'database');
```
touhidurabir, line 67: previously we have set the queue driver to `sync` when app in maintenance mode to run jobs but this is flawed . No jobs should run at all in maintenance mode .

jobs/orcid/SendUpdateScopeMail.php
```diff
@@ -38,7 +38,9 @@ public function __construct(
         /** @var int $itemId Will be a publication ID or a review assignment ID depending on deposit type */
         private int $itemId,
         private OrcidDepositType $depositType,
-    ) {
+    )
+    {
+        parent::__construct();
```
touhidurabir, line 43: here and above was missing the call to parent __construct, nothing critical but this is how we have all jobs set up .

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
     /** @copydoc ScheduledTask::executeActions() */
     public function executeActions(): bool
     {
-        $this->fileManager = new PrivateFileManager();
-        $pathZipDir = $this->fileManager->getBasePath() . DIRECTORY_SEPARATOR . $this->prefix;
-        $pathZipFile = $pathZipDir . '.zip';
+        $downloadUrl = $this->getDownloadUrl();
+        if (empty($downloadUrl)) {
+            UpdateRorRegistryDataset::log(
```
bozana: here and below we use `log` function, and further below also `addExecutionLogEntry`. Could we everywhere in this class only use `addExecutionLogEntry`?

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+     */
+    public int $timeout = 600;
+
+    protected PrivateFileManager $fileManager;
+
+    public function __construct(
+        protected string $downloadUrl, 
+        protected string $csvNameContains, 
+        protected string $prefix,
+        protected ?string $scheduledTaskLogFilesPath = null
```
bozana: maybe to call it in singular `scheduledTaskLogFilePath` ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
+                $fileManager->deleteByPath($path);
             }
         }
     }
 
-    /**
-     * Create temporary table
-     */
-    private function createTemporaryTable(): void
+    public static function log(string $message, ?string $logFilePath = null, ?string $type = null): void
```
bozana: Is there a case when `logFilePath` is/should be null?

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+                ->then(function (Batch $batch) use ($pathZipDir, $pathZipFile, $chunkDir, $fileSize, $chunkSize, $prefix, $csvNameContains, $scheduledTaskLogFilesPath) {
+                    // after successfully completing all the batch jobs, 
+                    // first we need to verify if all chunks downloaded
+                    $chunks = glob($chunkDir . '/chunk_*');
+
+                    if (count($chunks) !== (int) ceil($fileSize / $chunkSize)) {
+                        // not all chunks downloaded
+                        // we will not attempt any more chunk downloads and cancel batch process
+                        // and fall back to sync download
+                        $batch->cancel();
```
bozana: Is this needed, because if `then` is called, all jobs were successfully completed ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
+                ),
+                (new ImportRorData(
+                    $this->prefix,
+                    $this->csvNameContains,
+                    $this->dataMapping,
+                    $this->dataMappingIndex,
+                    $this->noLocale,
+                    $this->temporaryTable,
+                    $this->getExecutionLogFile()
+                ))->delay(now()->addSeconds(60)),
```
bozana: Why do we need a delay here?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
-                return false;
-            }
-
-            $this->processCsv($pathCsv);
+            $fileManager = new PrivateFileManager();
+            $pathZipDir = $fileManager->getBasePath() . DIRECTORY_SEPARATOR . $this->prefix;
+            $pathZipFile = $pathZipDir . '.zip';
+            $logfile = $this->getExecutionLogFile();
+
+            // Following jobs will be dispatch in chain and process as follow :-
```
bozana: will be **dispatched** 
process **is** as **follows**

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
+            //  - Download RoR dataset in chunk, build final zip file and extract it
```
bozana: in **chunks**

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
-            $table->smallInteger('is_active')->nullable(false)->default(1);
-            $table->mediumText('search_phrase')->nullable();
-            $table->string('locale', 28)->default('');
-            $table->string('setting_name', 255);
-            $table->mediumText('setting_value')->nullable();
-            $table->unique(['ror', 'locale', 'setting_name'], $this->temporaryTable . '_unique');
-        });
-    }
+        // debug in error log
+        error_log($message);
```
bozana: does this need to be removed?

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+                    $startByte,
+                    $endByte,
+                    $chunkFile,
+                    $pathZipDir,
+                    $this->csvNameContains,
+                    $this->scheduledTaskLogFilesPath
+                );
+            }
+
+            // we need to get thses in local variables to pass into the batch closures(e.g. then, catch, etc)
```
bozana: these

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+                        return;
+                    }
+
+                    UpdateRorRegistryDataset::log(
+                        'All chunks downloaded successfully, proceeding with merging and extraction',
+                        $scheduledTaskLogFilesPath,
+                        ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_COMPLETED
+                    );
+
+                    // chunk downloaded successfully
```
bozana: chunks

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+                ->finally(function (Batch $batch) use ($downloadUrl, $pathZipFile, $pathZipDir, $chunkDir, $csvNameContains, $scheduledTaskLogFilesPath) {
+                    UpdateRorRegistryDataset::cleanup([$pathZipFile, $chunkDir]);
+                    UpdateRorRegistryDataset::log(
+                        'RoR dataset downloading in chunk batch jobs have finished.', 
+                        $scheduledTaskLogFilesPath,
+                        ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_NOTICE
+                    );
+
+                    // Regardless of chunk process success/fail, the fallback sync process will kick off with delay,
+                    // however will run check if the extract and target CSV file is present and if so,
```
bozana: however **it** will run check

jobs/ror/DownloadRoRDatasetInChunks.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        try {
+
+            $pathCsv = $this->fileManager->fileExists($this->pathZipDir, 'dir')
+                ? UpdateRorRegistryDataset::getPathCsv($this->pathZipDir, $this->csvNameContains)
+                : '';
+            
+            if (!empty($pathCsv) && $this->fileManager->fileExists($pathCsv)) {
+                $this->batch()->cancel();
+                return; // No need to download the CSV file, job can skip and cancel the batch
```
bozana, line 70: job can be skipped

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+            // Clean up any existing files
+            UpdateRorRegistryDataset::cleanup([$pathZipFile, $pathZipDir, $chunkDir]);
+
+            $fileSize = $this->getFileSize($this->downloadUrl);
+
+            if ($fileSize === 0) {
+                throw new Exception('Failed to determine file size');
+            }
+
+            $this->performChunkDownload(
```
bozana, line 85: Do we maybe have to check if the server allows range download?

jobs/ror/DownloadRoRDatasetInSync.php
```diff
@@ -0,0 +1,142 @@
...
+                    ? UpdateRorRegistryDataset::getPathCsv($this->pathZipDir, $this->csvNameContains)
+                    : '';
+                
+                if (!empty($pathCsv) && $this->fileManager->fileExists($pathCsv)) {
+                    UpdateRorRegistryDataset::log(
+                        'Cancelling sync download as the CSV alerady exists',
+                        $this->scheduledTaskLogFilesPath, 
+                        ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+                    );
+                    return true; // No need to download the CSV file, job can skip
```
bozana: job can be skipped

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    protected function countCsvRows(string $pathCsv): int
+    {
+        $rowCount = 0;
+        $handle = fopen($pathCsv, 'r');
+        while (!feof($handle)) {
+            $buffer = fread($handle, 8192);
+            $rowCount += substr_count($buffer, "\n");
```
bozana: would it maybe be better to use `fgets` instead of `fread` here, or maybe even SplFileObject to get the total number of lines?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
-            $this->dropTemporaryTable();
-            $this->cleanup([$pathZipFile, $pathZipDir]);
         }
     }
 
-    /**
-     * Get download url
-     */
-    private function getDownloadUrl(): string
+    protected function getDownloadUrl(): string
```
bozana, line 189: Maybe to add function description?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
-
-            return false;
-        }
-    }
-
-    /**
-     * Get path to csv file
-     */
-    private function getPathCsv(string $pathZipDir): string
+    public static function getPathCsv(string $pathZipDir, string $csvNameContains): string
```
bozana, line 225: Function description?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -116,330 +112,131 @@ public function getName(): string
...
-            'searchPhrase' => trim($searchPhrase),
-            'name' => $namesOut,
-        ];
-    }
-
-    /**
-     * Cleanup temporary files and directories.
-     */
-    private function cleanup(array $paths = []): void
+    public static function cleanup(array $paths = []): void
```
bozana, line 241: Function description

classes/task/UpdateRorRegistryDataset.php
bozana: Function description

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+            UpdateRorRegistryDataset::log(
+                "RoR dataset downloading in chunk and extract main job have failed: {$e->getMessage()}",
+                $this->scheduledTaskLogFilesPath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+            throw $e;
+        }
+    }
+
+    protected function performChunkDownload(
```
bozana, line 108: Function description?

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,226 @@
...
+            UpdateRorRegistryDataset::cleanup([$pathZipDir, $pathZipFile, $chunkDir]);
+            UpdateRorRegistryDataset::log(
+                "RoR dataset downloading in chunk batch jobs have failed: {$e->getMessage()}",
+                $this->scheduledTaskLogFilesPath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+        }
+    }
+
+    protected function getFileSize(string $downloadUrl): int
```
bozana: Function description

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,168 @@
...
+                    $endRow,
+                    $this->dataMapping,
+                    $this->dataMappingIndex,
+                    $this->noLocale,
+                    $this->temporaryTable,
+                    $this->scheduledTaskLogFilesPath
+                );
+            }
+
+            // we need to get thses in local variables to pass into the batch closures(e.g. then, catch, etc)
```
bozana: these

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,168 @@
...
+                "RoR dataset importing failed: {$e->getMessage()}",
+                $this->scheduledTaskLogFilesPath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+
+            throw $e;
+        }
+    }
+
+    protected function countCsvRows(string $pathCsv): int
```
bozana, line 160: Function description

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,281 @@
...
+        $this->dropTemporaryTable();
+
+        UpdateRorRegistryDataset::log(
+            "Importing batch chunk job has failed : {$exception->getMessage()}",
+            $this->scheduledTaskLogFilesPath,
+            ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+        );
+    }
+
+    protected function processRow(array $row): array
```
bozana, line 139: Function description

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,281 @@
...
+        return [
+            'ror' => $ror,
+            'displayLocale' => $displayLocale,
+            'isActive' => $isActive,
+            'searchPhrase' => trim($searchPhrase),
+            'name' => $namesOut,
+        ];
+    }
+
+    protected function processBatch(array $rows): void
```
bozana, line 177: Function description

jobs/ror/DownloadRoRDatasetInChunks.php
```diff
@@ -0,0 +1,108 @@
...
+
+            UpdateRorRegistryDataset::log(
+                "Downloading chunk {$this->startByte}-{$this->endByte} to {$this->chunkFile}",
+                $this->scheduledTaskLogFilesPath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_NOTICE
+            );
+
+            $response = $client->request('GET', $this->downloadUrl, [
+                'headers' => ['Range' => "bytes={$this->startByte}-{$this->endByte}"],
+                'sink' => fopen($this->chunkFile, 'wb'),
```
bozana: if fopen for some reason did not work the failure would be: "sink must not be a boolean" (here and below)
which is maybe OK...

classes/task/UpdateRorRegistryDataset.php
touhidurabir: this is as per laravel's note that we should not use `$this` within the job closure when possible as 
```
Since chain callbacks are serialized and executed at a later time by the Laravel queue, you should not use the `$this` variable within chain callbacks.
```
at https://laravel.com/docs/11.x/queues#chain-failures . In fact in my experience , as laravel use https://github.com/laravel/serializable-closure, we should never pass `$this` to any batch/chain callbacks as the serialization of `$this` is buggy when the callbacks/closure are serialized and store in DB .

classes/task/UpdateRorRegistryDataset.php
touhidurabir: not needed anymore .

classes/scheduledTask/ScheduledTask.php
```diff
@@ -157,6 +152,33 @@ public function execute(): bool
...
+        }
+
+        $fp = fopen($logFile, 'ab');
+        if (flock($fp, LOCK_EX)) {
+            fwrite($fp, $log . PHP_EOL);
+            flock($fp, LOCK_UN);
+        } else {
+            throw new Exception("Couldn't lock the file.");
+        }
+        fclose($fp);
```
jonasraoni, line 180:
```suggestion
        file_put_contents($this->executionLogFile, $log . PHP_EOL, FILE_APPEND | LOCK_EX);
```

classes/scheduledTask/ScheduledTask.php
```diff
@@ -157,6 +152,33 @@ public function execute(): bool
...
+     */
+    public static function writeToExecutionLogFile(string $message, string $logFile, ?string $type = null): void
+    {
+        $date = '[' . Core::getCurrentDate() . '] ';
+
+        if ($type) {
+            $log = $date . '[' . __($type) . '] ' . $message;
+        } else {
+            $log = $date . $message;
+        }
```
jonasraoni, line 171:
```suggestion
        $type = $type ? '[' . __($type) . '] ' : '';
        $log = "{$date}{$type}{$message}";
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -64,21 +62,22 @@
...
+use PKP\jobs\ror\DownloadAndExtractRorDataset;
+use PKP\jobs\ror\ImportRorData;
 use PKP\scheduledTask\ScheduledTask;
 use PKP\scheduledTask\ScheduledTaskHelper;
-use ZipArchive;
 
 class UpdateRorRegistryDataset extends ScheduledTask
 {
-    private PrivateFileManager $fileManager;
+    public const CACHE_KEY_CHUNK_BATCH = 'ror:chunk:batch:processing';
```
jonasraoni, line 80: I see this is used to control if the chunk jobs are being executed and etc, but I think it shouldn't be needed 🤔

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -64,21 +62,22 @@
...
 
     /** @var string API Url of the data dump versions. */
     private string $urlVersions = 'https://zenodo.org/api/communities/ror-data/records?q=&sort=newest';
```
jonasraoni, line 83:
```suggestion
    private const URL_VERSIONS = 'https://zenodo.org/api/communities/ror-data/records?sort=newest';
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
     /** @var string Name of temporary table */
     private string $temporaryTable = 'rors_temporary';
```
jonasraoni, line 107:
```suggestion
    private const TEMPORARY_TABLE = 'rors_temporary';
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
-            }
-
-            $pathCsv = $this->getPathCsv($pathZipDir);
-            if (empty($pathCsv || !$this->fileManager->fileExists($pathCsv))) {
-                return false;
-            }
-
-            $this->processCsv($pathCsv);
+            $fileManager = new PrivateFileManager();
+            $pathZipDir = $fileManager->getBasePath() . DIRECTORY_SEPARATOR . $this->prefix;
```
jonasraoni, line 134: We're using `/` for the sake of simplicity, there are other cases in the PR that can be modified as well.

```suggestion
            $pathZipDir = $fileManager->getBasePath() . "/{$this->prefix}";
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
             return false;
-        } finally {
-            $this->dropTemporaryTable();
-            $this->cleanup([$pathZipFile, $pathZipDir]);
         }
     }
 
     /**
-     * Get download url
+     * generate and get the ROR dataset download url
```
jonasraoni, line 187:
```suggestion
     * Generate and get the ROR dataset download url
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
      */
-    private function getDownloadUrl(): string
+    protected function getDownloadUrl(): string
```
jonasraoni, line 189: If there's no plan to extend the class, I think it can be kept as private.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
     {
         try {
             $client = Application::get()->getHttpClient();
+            $response = $client->request('GET', $this->urlVersions, [
+                'connect_timeout' => 10
```
jonasraoni, line 194: This type of thing shouldn't be hard coded, but at the same time, I don't think a proper implementation should be covered here (i.e. there are other important settings which should be configurable), so let it be.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
+            ]);
 
-            $response = $client->request('GET', $this->urlVersions);
             $responseArray = json_decode($response->getBody(), true);
 
-            if ($response->getStatusCode() !== 200 || json_last_error() !== JSON_ERROR_NONE || empty($responseArray)) {
+            if ($response->getStatusCode() !== 200
+                || json_last_error() !== JSON_ERROR_NONE
```
jonasraoni, line 200: I think this is a less error prone approach (e.g. the logic will be broken if another indirect method gets inserted before the `json_last_error()` check)

```suggestion
$responseArray = json_decode($response->getBody(), true, JSON_THROW_ON_ERROR);

            if ($response->getStatusCode() !== 200
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
+                );
                 return '';
             }
 
-            if (!empty($responseArray['hits']['hits'][0]['files'][0]['links']['self'])) {
-                return $responseArray['hits']['hits'][0]['files'][0]['links']['self'];
-            }
-
-            return '';
+            return $responseArray['hits']['hits'][0]['files'][0]['links']['self'] ?? '';
```
jonasraoni, line 211: I don't think this is ok, in my opinion it should fail the process, otherwise we'll be hiding problems under the carpet.

```suggestion
            return $responseArray['hits']['hits'][0]['files'][0]['links']['self'] ?? throw new Exception('Unexpected response format');
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
         } catch (GuzzleException|Exception $e) {
-            $this->addExecutionLogEntry(
+            static::writeToExecutionLogFile(
                 $e->getMessage(),
+                $this->getExecutionLogFile(),
                 ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
             );
-
             return '';
```
jonasraoni, line 218: Same here, let the exception go out or return null, as a user of this method I wouldn't expect to check if it's an empty string.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
-            if (!$fileInfo->isDot()) {
-                if (str_contains($fileInfo->getFilename(), $this->csvNameContains)) {
-                    return $fileInfo->getPathname();
-                }
+            if (!$fileInfo->isDot() && str_contains($fileInfo->getFilename(), $csvNameContains)) {
+                return $fileInfo->getPathname();
             }
         }
 
         return '';
```
jonasraoni, line 235: If the return is required, an exception should be raised or a null should be returned.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
+    public static function getPathCsv(string $pathZipDir, string $csvNameContains): string
     {
         $iterator = new DirectoryIterator($pathZipDir);
+
         foreach ($iterator as $fileInfo) {
-            if (!$fileInfo->isDot()) {
-                if (str_contains($fileInfo->getFilename(), $this->csvNameContains)) {
-                    return $fileInfo->getPathname();
-                }
+            if (!$fileInfo->isDot() && str_contains($fileInfo->getFilename(), $csvNameContains)) {
```
jonasraoni, line 230: If we use the `FilesystemIterator`, then it's not needed to call the `isDot()`

```suggestion
$iterator = new FilesystemIterator($pathZipDir);

foreach ($iterator as $fileInfo) {
            if (str_contains($fileInfo->getFilename(), $csvNameContains)) {
```

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
+        $fileManager = new PrivateFileManager();
+        
         foreach ($paths as $path) {
-            if ($this->fileManager->fileExists($path, 'dir')) {
+            if ($fileManager->fileExists($path, 'dir')) {
                 foreach (new DirectoryIterator($path) as $fileInfo) {
                     if (!$fileInfo->isDot()) {
-                        $this->fileManager->deleteByPath($fileInfo->getPathname());
+                        $fileManager->deleteByPath($fileInfo->getPathname());
                     }
```
jonasraoni, line 250:
```suggestion
                foreach (new DirectoryIterator($path) as $fileInfo) {
                    $fileManager->deleteByPath($fileInfo->getPathname());
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
+<?php
+
+/**
+ * @file jobs/ror/DownloadAndExtractRorDataset.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/DownloadRoRDatasetInChunks.php
```diff
@@ -0,0 +1,116 @@
+<?php
+
+/**
+ * @file jobs/ror/DownloadRoRDatasetInChunks.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/DownloadRoRDatasetInSync.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+/**
+ * @file jobs/ror/DownloadRoRDatasetInSync.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+/**
+ * @file jobs/ror/ImportRorData.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/MergeAndExtractRorDatasetChunks.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+/**
+ * @file jobs/ror/MergeAndExtractRorDatasetChunks.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
+<?php
+
+/**
+ * @file jobs/ror/ProcessRorCsv.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+use PKP\task\UpdateRorRegistryDataset;
+use GuzzleHttp\Exception\GuzzleException;
+use PKP\scheduledTask\ScheduledTaskHelper;
+use PKP\jobs\ror\DownloadRoRDatasetInChunks;
+use Illuminate\Support\Facades\Cache;
+
+class DownloadAndExtractRorDataset extends BaseJob
+{
+    /**
+     * The maximum number of SECONDS a job should get processed before consider failed
```
jonasraoni, line 34:
```suggestion
     * The maximum number of SECONDS a job should be processed before being considered as failed
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+     */
+    public int $timeout = 300;
```
jonasraoni, line 36: Given we don't know the network/disk/cpu speed of the users, I'm against hard coded timeouts in jobs, I think everything should be expected to finish, sooner or later, unless there are infinite loops (which require a different type of protection).

But timeouts placed at the request (connect + read timeout) are great 👌

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+
+    protected PrivateFileManager $fileManager;
```
jonasraoni, line 38: If this isn't going to be extended, I think it can be private.

```suggestion
    private PrivateFileManager $fileManager;
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+
+    public function __construct(
+        protected string $downloadUrl, 
+        protected string $csvNameContains, 
+        protected string $prefix,
+        protected string $scheduledTaskLogFilePath
```
jonasraoni, line 44: I think these could be private as well.

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+                throw new Exception('Failed to create chunk directory');
+            }
+
+            $chunkSize = 1024 * 1024 * 3 ; // 3 MB
+            $numChunks = ceil($fileSize / $chunkSize);
+            $jobs = [];
+
+            for ($i = 0; $i < $numChunks; $i++) {
+                $startByte = $i * $chunkSize;
+                $endByte = min(($i + 1) * $chunkSize - 1, $fileSize - 1);
```
jonasraoni, line 128: Is the generated download URL immutable? If not, there must be a hash validation.

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+            // Dispatch batch
+            Bus::batch($jobs)
+                ->then(function (Batch $batch) use ($pathZipDir, $pathZipFile, $chunkDir, $fileSize, $chunkSize, $prefix, $csvNameContains, $scheduledTaskLogFilePath) {
+                    // after successfully completing all the batch jobs, 
+                    // first we need to verify if all chunks downloaded
+                    $chunks = glob($chunkDir . '/chunk_*');
+
+                    if (count($chunks) !== (int) ceil($fileSize / $chunkSize)) {
+                        UpdateRorRegistryDataset::writeToExecutionLogFile(
+                            'Not all chunks downloaded, cancelling batch chunk downloading',
```
jonasraoni, line 156:
```suggestion
                            "Not all chunks were downloaded, expected {$expectedChunks} found {$foundChunks}, the process will be retried",
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+                    UpdateRorRegistryDataset::writeToExecutionLogFile(
+                        "RoR dataset chunk download batch at progress of {$batch->progress()}% have failed: {$e->getMessage()}",
+                        $scheduledTaskLogFilePath,
+                        ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+                    );
+                })
+                ->finally(function (Batch $batch) use ($prefix, $downloadUrl, $pathZipFile, $pathZipDir, $chunkDir, $csvNameContains, $scheduledTaskLogFilePath) {
+                    UpdateRorRegistryDataset::cleanup([$pathZipFile, $chunkDir]);
+
+                    Cache::put(UpdateRorRegistryDataset::CACHE_KEY_CHUNK_BATCH, true, 600);
```
jonasraoni, line 196: It's not clear for me the 600 timeout.

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+                "RoR dataset downloading in chunk batch jobs have failed: {$e->getMessage()}",
+                $this->scheduledTaskLogFilePath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+            throw $e;
+        }
+    }
+
+    /**
+     * Get the target file size and verify range download support.
```
jonasraoni, line 219:
```suggestion
     * Get the target file size and verify range download support.
     * 
     * @return array{int, bool} Returns an array, the first item contains the file size, the second whether range download is supported
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+    {
+        try {
+            $client = Application::get()->getHttpClient();
+
+            // Step 1: Get file size without Range header
+            $sizeResponse = $client->request('HEAD', $downloadUrl, [
+                'connect_timeout' => 10,
+            ]);
+            $contentLength = $sizeResponse->getHeaderLine('Content-Length');
+            $fileSize = $contentLength ? (int) $contentLength : 0;
```
jonasraoni, line 231:
```suggestion
            $fileSize = (int) abs($sizeResponse->getHeaderLine('Content-Length'));
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+        try {
+            // Clean up any existing files
+            UpdateRorRegistryDataset::cleanup([$pathZipFile, $pathZipDir, $chunkDir]);
+            Cache::forget(UpdateRorRegistryDataset::CACHE_KEY_CHUNK_BATCH);
+
+            [$fileSize, $supportsRange] = $this->getFileSizeAndVerifyRangeDownload($this->downloadUrl);
+
+            // if can not determine file size or range download support, 
+            // we will not kick off the chunk download process and exit this job process
+            if ($fileSize === 0 || !$supportsRange) {
```
jonasraoni, line 76:
```suggestion
            if (!$fileSize || !$supportsRange) {
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+            $client = Application::get()->getHttpClient();
+
+            // Step 1: Get file size without Range header
+            $sizeResponse = $client->request('HEAD', $downloadUrl, [
+                'connect_timeout' => 10,
+            ]);
+            $contentLength = $sizeResponse->getHeaderLine('Content-Length');
+            $fileSize = $contentLength ? (int) $contentLength : 0;
+
+            if ($fileSize === 0) {
```
jonasraoni, line 233:
```suggestion
            if (!$fileSize) {
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+            return [$fileSize, $supportsRange];
+
+        } catch (GuzzleException|Exception $e) {
+            UpdateRorRegistryDataset::writeToExecutionLogFile(
+                "Failed to get file size: {$e->getMessage()}",
+                $this->scheduledTaskLogFilePath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+
+            return [0, false];
```
jonasraoni, line 268: In case of error I think the task should be retried.

jobs/ror/DownloadRoRDatasetInChunks.php
```diff
@@ -0,0 +1,116 @@
...
+            $fileHandle = fopen($this->chunkFile, 'wb');
+            if ($fileHandle === false) {
+                throw new Exception("Failed to open chunk file {$this->chunkFile} for writing");
+            }
+
+            $response = $client->request('GET', $this->downloadUrl, [
+                'headers' => ['Range' => "bytes={$this->startByte}-{$this->endByte}"],
+                'sink' => $fileHandle,
+                'connect_timeout' => 10,
+                'timeout' => $this->timeout - 1, // set the download time 1 second less than the job timeout
```
jonasraoni, line 91: I think we need to use the `read_timeout` instead of a generic timeout, as long as the request is progressing we can wait.

classes/scheduledTask/ScheduledTask.php
```diff
@@ -78,6 +79,14 @@ public function getProcessId(): ?string
...
     }
 
+    /**
+     * Get the execution log file path.
+     */
+    public function getExecutionLogFile(): string
+    {
+        return $this->executionLogFile;
+    }
+
```
jonasraoni, line 89: Nothing against keeping it, but as the class is deprecated, I don't think it's needed to extend its public API, AFAICS the calls that use this method could be replaced by calls to `addExecutionLogEntry()`.

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Define middleware for preventing overlaps
+     */
+    public function middleware()
+    {
+        return [
+            (new WithoutOverlapping($this->prefix . ':download'))->expireAfter(300), // 5 mins lock
+        ];
```
jonasraoni, line 58: I think these type of checks shouldn't be needed with a good "wiring".

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+                    ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+                );
+                return [0, false];
+            }
+
+            // Step 2: Verify range support with a small GET range request (GET instead of HEAD for better server compatibility)
+            $rangeResponse = $client->request('GET', $downloadUrl, [
+                'connect_timeout' => 10,
+                'timeout' => 30,
+                'headers' => ['Range' => 'bytes=0-1023'],
```
jonasraoni, line 246: I think we can use the smallest possible range (1 byte), but not important.

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+                    $this->scheduledTaskLogFilePath,
+                    ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+                );
+                return [0, false];
+            }
+
+            // Step 2: Verify range support with a small GET range request (GET instead of HEAD for better server compatibility)
+            $rangeResponse = $client->request('GET', $downloadUrl, [
+                'connect_timeout' => 10,
+                'timeout' => 30,
```
jonasraoni, line 245: Instead of a "total" timeout, I think we should use a read timeout (as long as the transfer is having progress, we can wait).

jobs/ror/DownloadRoRDatasetInSync.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Indicate if the job should be marked as failed on timeout.
+     */
+    public bool $failOnTimeout = false;
+
+    /**
+     * The number of SECONDS to wait before retrying the job.
+     */
+    public int $backoff = 60;
```
jonasraoni, line 44: The backoff value for a failed download, including the chunked ones, can make use of the `retry-after` header, if it exists. (e.g. `max(retry-after, 60)`)

Depending on the type of the failure (e.g. unresponsive), the backoff could be MUCH higher, like a couple of days.

jobs/ror/DownloadRoRDatasetInSync.php
```diff
@@ -0,0 +1,159 @@
...
+                || !$this->fileManager->fileExists($this->pathZipFile)
+            ) {
+                throw new Exception('Failed to download ROR dataset');
+            }
+
+            $zip = new ZipArchive();
+            if ($zip->open($this->pathZipFile) !== true) {
+                throw new Exception('Failed to open ZIP file');
+            }
+            $zip->extractTo($this->pathZipDir);
```
jonasraoni, line 134: Here we can extract only the needed CSV file.

Also, it's needed how the extraction is done. If the `ZipArchive` creates the file right away at the right path and start writing to it, then it's better to extract it to a temporary folder instead, then move once it's finished. I mean... If the extraction fails at the middle, the job code will consider it was successfully (AFAICS there's no checksum check, just a lookup for the file existence).

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,178 @@
...
+            $pathZipDir = $this->fileManager->getBasePath() . DIRECTORY_SEPARATOR . $this->prefix;
+
+            if (!$this->fileManager->fileExists($pathZipDir, 'dir')) {
+                throw new Exception("Ror dataset extracted directory at {$pathZipDir} not found");
+            }
+
+            $pathCsv = UpdateRorRegistryDataset::getPathCsv($pathZipDir, $this->csvNameContains);
+            if (empty($pathCsv) || !$this->fileManager->fileExists($pathCsv)) {
+                throw new Exception('CSV file not found');
+            }
```
jonasraoni, line 84: Instead of passing all these variables, I think this job could be created later on, and just receive the needed path. I mean, it's possible to include extra jobs on the current chain.

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,178 @@
...
+
+            // Count total rows
+            $totalRows = $this->countCsvRows($pathCsv);
```
jonasraoni, line 87: Counting all rows will require parsing and counting all rows, I think we could do it in a better way.

Given it's possible to append extra jobs to a chain, perhaps we can attempt to process 5000 rows, then prepend an extra job to resume if there are still bytes ahead.

On the new job, to avoid re-parsing/skipping the already processed lines, you can receive the file position.

jobs/ror/ImportRorData.php
```diff
@@ -0,0 +1,178 @@
...
+        $rowCount = 0;
+
+        $handle = fopen($pathCsv, 'r');
+        if ($handle === false) {
+            throw new Exception("Failed to open CSV file: {$pathCsv}");
+        }
+
+        // Skip header row
+        fgets($handle);
+        while (fgets($handle) !== false) {
```
jonasraoni, line 171: This implementation **might work**, but it's invalid, a CSV can have line-breaks in a row.

jobs/ror/MergeAndExtractRorDatasetChunks.php
```diff
@@ -0,0 +1,118 @@
...
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_NOTICE
+            );
+
+            $zipHandle = fopen($this->pathZipFile, 'wb');
+            if ($zipHandle === false) {
+                throw new Exception('Failed to open ZIP file for writing');
+            }
+
+            $chunks = glob($this->chunkDir . DIRECTORY_SEPARATOR . 'chunk_*');
+            natsort($chunks); // Ensure chunks are in order
```
jonasraoni, line 64: Alternatively you can attempt to open the chunk_0, chunk_1, ..., until a file isn't found, I think it should be faster, even though it's not relevant.

jobs/ror/MergeAndExtractRorDatasetChunks.php
```diff
@@ -0,0 +1,118 @@
...
+            $zip = new ZipArchive();
+            if ($zip->open($this->pathZipFile) !== true) {
+                throw new Exception('Failed to open ZIP file');
+            }
+            $zip->extractTo($this->pathZipDir);
+            $zip->close();
+
+            if (!$this->fileManager->fileExists($this->pathZipDir, 'dir')) {
+                throw new Exception('Extraction failed');
+            }
```
jonasraoni, line 91: I saw this piece of code in the other job, so I think it can be moved to the next job (`ProcessRorCsv`) or an intermediary one (if it's heavy enough).

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+            );
+
+            $this->createTemporaryTable();
+
+            $batchRows = [];
+            $currentRow = 0;
+            $isHeader = true;
+
+            if (($handle = fopen($this->pathCsv, 'r')) !== false) {
+                while (($rowCsv = fgetcsv($handle, null, ',', '"', '\\')) !== false) {
```
jonasraoni, line 76: These are the default values, so it's ok to use:

```suggestion
                while (($rowCsv = fgetcsv($handle)) !== false) {
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_NOTICE
+            );
+
+            $this->createTemporaryTable();
+
+            $batchRows = [];
+            $currentRow = 0;
+            $isHeader = true;
+
+            if (($handle = fopen($this->pathCsv, 'r')) !== false) {
```
jonasraoni, line 75: Alternatively, you can use more the `SplFileObject`, to avoid dealing with file handles and passing them around.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                while (($rowCsv = fgetcsv($handle, null, ',', '"', '\\')) !== false) {
+                    if ($isHeader) {
+                        foreach ($this->dataMapping as $keyDB => $keyCsv) {
+                            $this->dataMappingIndex[$keyDB] = array_search($keyCsv, $rowCsv, true);
+                        }
+                        $isHeader = false;
+                        continue;
+                    }
```
jonasraoni, line 83: Instead of doing this check on every row, you can move it out of the loop, and add an initial fetch, like `if (($header = fgetcsv($handle)) !=== false) { ... } while (...)`

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+
+                    $currentRow++;
+                    if ($currentRow < $this->startRow) {
+                        continue;
+                    }
+                    if ($currentRow > $this->endRow) {
+                        break;
+                    }
```
jonasraoni, line 91: As I said on the previous comment, you can do a file seek to skip to avoid skipping rows.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_COMPLETED
+            );
+        } catch (Throwable $e) {
+            UpdateRorRegistryDataset::writeToExecutionLogFile(
+                "Importing batch chunk {$this->startRow}-{$this->endRow} failed: {$e->getMessage()}",
+                $this->scheduledTaskLogFilePath,
+                ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+            );
+
+            $this->dropTemporaryTable();
```
jonasraoni, line 116: This type of code that should be executed always can be moved to a `finally` section.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+        $displayLocale = (!empty($row[$this->dataMappingIndex['displayLocale']]))
+            ? $row[$this->dataMappingIndex['displayLocale']]
+            : $this->noLocale;
+        $isActive = (strtolower($row[$this->dataMappingIndex['isActive']]) === 'active') ? 1 : 0;
+        $searchPhrase = $ror;
+
+        $namesIn = $row[$this->dataMappingIndex['names']];
+        $namesOut = [];
+        if (!empty($namesIn)) {
+            $names = array_map('trim', explode(';', $namesIn));
```
jonasraoni, line 151: Is there a escape value for `;`, like `\;`? If yes, this should be considered.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+            for ($i = 0; $i < count($names); $i++) {
+                $name = array_map('trim', explode(':', $names[$i]));
```
jonasraoni, line 153: I didn't check the file, but I guess this works like a `name: Jonas`, if yes, then I think it's good to specify a maximum number of items to be `exploded`, to cover errrrr `name: Jonas:blabla`

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * Process the CSV row to generate the ROR data array.
+     */
+    protected function processRow(array $row): array
+    {
+        $ror = $row[$this->dataMappingIndex['ror']];
+        $displayLocale = (!empty($row[$this->dataMappingIndex['displayLocale']]))
+            ? $row[$this->dataMappingIndex['displayLocale']]
+            : $this->noLocale;
```
jonasraoni, line 144: Easier to read

```suggestion
        $displayLocale = empty($row[$this->dataMappingIndex['displayLocale']])
            ? $this->noLocale
            : $row[$this->dataMappingIndex['displayLocale']];
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+        $isActive = (strtolower($row[$this->dataMappingIndex['isActive']]) === 'active') ? 1 : 0;
```
jonasraoni, line 145:
```suggestion
        $isActive = (int) (strtolower($row[$this->dataMappingIndex['isActive']]) === 'active');
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+            ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+        );
+    }
+
+    /**
+     * Process the CSV row to generate the ROR data array.
+     */
+    protected function processRow(array $row): array
+    {
+        $ror = $row[$this->dataMappingIndex['ror']];
```
jonasraoni, line 141: As it's an external data source, I think it's better to expect the worst (rows might not be formatted properly) and include a coalesce (or throw a warning) when accessing the `$row`.

```suggestion
        $ror = $row[$this->dataMappingIndex['ror']] ?? '';
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+            'displayLocale' => $displayLocale,
+            'isActive' => $isActive,
+            'searchPhrase' => trim($searchPhrase),
+            'name' => $namesOut,
+        ];
+    }
+
+    /**
+     * Process the CSV row to store in DB
+     */
```
jonasraoni, line 176: I think it's good to type these array params.

```suggestion
     *
     * @param array<array{name: string, isActive: int, ...}> $rows
     */
```

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                        'setting_value' => $name
+                    ];
+                }
+            }
+            DB::table($this->temporaryTable)->insert($values);
+
+            $values = DB::table($this->temporaryTable . ' as tmp')
+                ->select('tmp.ror', 'tmp.display_locale', 'tmp.is_active', 'tmp.search_phrase')
+                ->distinct()
+                ->leftJoin('rors as r', 'tmp.ror', '=', 'r.ror')
```
jonasraoni, line 200: Looks like it's missing a `whereNull('r.ror_id')` (if the idea is to not insert RORs that already exist). Otherwise, if the intention is to update existing RORs, the join can be removed.

I didn't inspect the dataset, so the only reason for the usage of this temporary table seems to be the `distinct()`. The distinct can be done at the PHP side, then we can proceed directly with the `upsert()` using the `$rows`.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                        'display_locale' => $row['displayLocale'],
+                        'is_active' => $row['isActive'],
+                        'search_phrase' => $row['searchPhrase'],
+                        'locale' => $locale,
+                        'setting_name' => 'name',
+                        'setting_value' => $name
+                    ];
+                }
+            }
+            DB::table($this->temporaryTable)->insert($values);
```
jonasraoni, line 195: The temporary table has a unique key, which I guess will fail if a duplicated row is found here, so some protection is needed.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+            DB::table('rors')->upsert(
+                $values,
+                ['ror'],
+                ['display_locale', 'is_active', 'search_phrase']
+            );
+
+            $orphanedSettings = DB::table('ror_settings AS rs')
+                ->select('rs.ror_setting_id')
+                ->join('rors as r', 'rs.ror_id', '=', 'r.ror_id')
+                ->join($this->temporaryTable . ' as tmp1', 'tmp1.ror', '=', 'r.ror')
```
jonasraoni, line 215: It seems to me that this join isn't needed, I think the next join should be enough, just need to replace the first filter by `tmp2.ror = r.ror`.

I'd say the `distinct()` isn't needed too, unless there are duplicated settings on the temporary table (which shouldn't exist due to the unique key).

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                ->leftJoin($this->temporaryTable . ' AS tmp2', function ($join) {
+                    $join->on('tmp2.ror', '=', 'tmp1.ror')
+                        ->on('tmp2.locale', '=', 'rs.locale')
+                        ->on('tmp2.setting_name', '=', 'rs.setting_name');
+                })
+                ->whereNull('tmp2.locale')
+                ->distinct()
+                ->pluck('rs.ror_setting_id');
+            DB::table('ror_settings')->whereIn('ror_setting_id', $orphanedSettings)->delete();
```
jonasraoni, line 224: Perhaps it's possible to join this with the previous and run a single query.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+
+            $values = DB::table($this->temporaryTable . ' as tmp')
+                ->select('r.ror_id', 'tmp.locale', 'tmp.setting_name', 'tmp.setting_value')
+                ->join('rors as r', 'tmp.ror', '=', 'r.ror')
+                ->leftJoin('ror_settings as rs', function ($join) {
+                    $join->on('r.ror_id', '=', 'rs.ror_id')
+                        ->on('rs.locale', '=', 'tmp.locale')
+                        ->on('rs.setting_name', '=', 'tmp.setting_name');
+                })
```
jonasraoni, line 233: This is pointless too, the same type of join that I've commented previously.

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+                ->distinct()
```
jonasraoni, line 234: This seems to be not needed, the columns which are retrieved are covered by a unique key (just missing the `tmp.ror` which is referenced indirectly)

jobs/ror/ProcessRorCsv.php
```diff
@@ -0,0 +1,287 @@
...
+        });
+    }
+
+    /**
+     * Drop temporary table
+     */
+    protected function dropTemporaryTable(): void
+    {
+        Schema::dropIfExists($this->temporaryTable);
+    }
```
jonasraoni, line 286: I think we don't need a temporary table for these operations.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -107,339 +106,153 @@ class UpdateRorRegistryDataset extends ScheduledTask
...
+                    $this->prefix,
+                    $logfile
+                ),
+                new DownloadRoRDatasetInSync(
+                    $this->csvNameContains,
+                    $downloadUrl, 
+                    $pathZipFile, 
+                    $pathZipDir, 
+                    $logfile
+                ),
```
jonasraoni, line 155: I think we don't need to have "two paths" here, the server does support the Range header and I think it's unlikely they will block it or move to a rustic HTTP server that doesn't support it.

The snippet below should cover both situations (if the header isn't supported, the server will probably just ignore it, and feed the whole content). I'll leave the good points that I see on this approach:
- Only a single job is needed, it can be retried while there's progress.
- Partitioning by time instead of data size sounds good (in case the server is fast, it might download everything in a single request)
- As we're writing to a single file, it will avoid the extra task of joining the chunks into a single file at the end

```php
$client = new Client();
$headers = [];
// Check if the file already exists and get its size for resuming
if (file_exists($filePath)) {
    $headers['Range'] = 'bytes=' . filesize($filePath) . '-'; //From the current file position until the end, but I think it's better to pass 
}

try {
    $response = $client->request('GET', $url, [
        'headers' => $headers,
        'sink' => fopen($filePath, 'a'), // Append...
        'timeout' => 30, // Optional, the job timeout will have the same effect...
    ]);

    // Check if the download was successful
    if ($response->getStatusCode() === count($headers) ? 206 : 200) {
        ...
    }
} catch (RequestException $e) {
  // Delay more time in case a 429 status is received, check the retry-after header value, in case the server is unresponsive, can delay for some days, etc.
}
```

jobs/ror/DownloadAndExtractRorDataset.php
```diff
@@ -0,0 +1,271 @@
...
+            }
+
+            $chunkSize = 1024 * 1024 * 3 ; // 3 MB
+            $numChunks = ceil($fileSize / $chunkSize);
+            $jobs = [];
+
+            for ($i = 0; $i < $numChunks; $i++) {
+                $startByte = $i * $chunkSize;
+                $endByte = min(($i + 1) * $chunkSize - 1, $fileSize - 1);
+                $chunkFile = $chunkDir . "/chunk_{$i}";
```
jonasraoni, line 129: Instead of writing the chunks to different files, then join them at the end, it sounds better to create a single file, reserve the total size (`ftruncate($file, $fileSize)`) and write the chunks over the right places.

jonasraoni (changes requested):
In general I think there are too many workarounds to make this work (regarding the `CACHE_KEY_CHUNK_BATCH` and the `WithoutOverlapping` middleware), in my opinion the Laravel wiring should be enough to accomplish the task without having to deal with overlaps, so I think the whole process can be simplified.

I also don't think it's needed to provide an alternative for the chunked download (I left a short snippet that probably cover both cases) nor create temporary tables for this task (I know this has been there before, but as it's a refactoring I think it was worth mentioning).

Given there's a 3rd-party in the game, a middleware which seems to be useful is the `ThrottlesExceptions`, which can help us to postpone the job chain in case we've got certain exceptions (e.g. a `429 Too Many Requests` could wait up to the header `Retry-After`, a network outage could be postponed for perhaps 1 day, etc).

Regarding the integration we're doing... Requesting a [ROR](https://ror.org/) data dump directly from Zenodo, which is another institution, sounds risky and unreliable for a software which doesn't receive automatic updates like ours.
Even though that's what they recommend (https://ror.readme.io/docs/data-dump), notice there were **recent breaking changes**:

image

In general I think we should have a modular/package based software instead of a huge codebase, it would simplify dealing with such problems, but it is what it is, so what we can do to help users is to provide perhaps a configuration to override the default behavior URL and "JSON path" (`['hits']['hits'][0]['files'][0]['links']['self']`) in case things get changed again 🤦

I also left some comments regarding performance optimizations and not important 
nitpicks.

bozana:
Hi @jonasraoni, I would like to see with you the SQLs for inserting the RORs, so I will contact you on MM to understand better what exactly you mean...

jobs/ror/ProcessRorCsv.php
bozana: We do not look for not inserting RORs that already exist -- also the properties of the existing needs to be considered and updated -- an upsert is used after this query, using $values.

There are DB tables rors and ror_settings. We need to update all existing RORs, remove the one that do not exist any more, and add the new RORs -- the entries in both DB tables -- according to the CSV. That's why we are using the temporary table, from where we can then do all of that. If you could help me understand how else we could solve it, it would be great! So maybe I can then help with changing it, so that Touhidur does not need to go too deep into this topic... :-)

jonasraoni:
Some extra notes:

- Looks like what I've commented previously just happened, and the integration was broken. So, we need to think about a better way to handle this, we can't have unstable pieces in an application that doesn't self-update itself, some ideas:
  - Get in touch with the data provider and ensure they can provide a stable API/integration (e.g. fixed URL that redirects to the right resource)
  - Man-in-the-middle: We already provide some fixed infra for the application (plugins.xml, version.xml, ...), so we could also provide a route that redirects the user to the right URL (and contains perhaps also instructions, like the filename pattern).
  - Provide a self-updating plugin (a bit against our rules, but pretty useful for some cases)
- Looks like this job is executed as part of the installation, and if it breaks, the installation breaks as well, it cannot happen, unless this was a strict requirement to use the software.

asmecher:
@jonasraoni, see https://github.com/pkp/pkp-lib/issues/12154 and https://github.com/pkp/pkp-lib/issues/12144 on that.

jonasraoni:
@asmecher Thanks for the references 👌
Is it possible to get the RoR folks to provide a more stable integration (e.g. versioned API), harvesting data directly from another domain (no matter how stable it is), without a proxy/redirect, already sounds kind of unsafe for me. If not, then I think we should go with another route (I've left some suggestions), as it's not uncommon to [have breaking changes](https://ror.readme.io/docs/data-dump#download-ror-data-dumps-programmatically-with-the-zenodo-api).

Issue #11389, opened 2025-05-13
API accepts invalid values when updating author user group


Hafsa-Naeem:
### Description

When updating an author's user group via the API, the endpoint currently accepts arbitrary or invalid values for the `userGroupId` field. This could result in inconsistent data or unintended role assignments.

### Steps to Reproduce

1. Send a PUT request to update a submission’s author with a wrong userGroupId.
2. Observe that the API accepts the request without validation errors.

### Expected Result

The API should validate `userGroupId` against available author user groups in the current context.

### Actual Result

The API accepts any value for `userGroupId` when updating an author's user group. No validation error is returned, and the invalid value is saved.

### What application are you using?

OJS 3.5

Tribunal33:
@asmecher I'm considering upgrading this one to bug lvl 3.  This is a security issue?

asmecher:
@Tribunal33, no, I don't see a security issue here, just data consistency. The user group ID attached to the author is never used for access permissions, so it can't be used to gain privileges. It should be fixed for 3.5.0, though.

Hafsa-Naeem:
@Vitaliy-1  Ready for review
OJS (stable-3_5_0) : [pkp/pkp-lib#11389 Submodule update ##Hafsa-Naeem/i11389-stable_3_5_0_fix #4885](https://github.com/pkp/ojs/pull/4885)

pkp-lib: [pkp/pkp-lib#11389 Validate contributor user group role is author before allowing add or edit in API #11412](https://github.com/pkp/pkp-lib/pull/11412)

Hafsa-Naeem:
@Vitaliy-1  Ready for review
OJS (stable-3_5_0) : [pkp/pkp-lib#11389 Submodule update ##Hafsa-Naeem/i11389-stable_3_5_0_fix #4885](https://github.com/pkp/ojs/pull/4885)

pkp-lib: [pkp/pkp-lib#11389 Validate contributor user group role is author before allowing add or edit in API #11412](https://github.com/pkp/pkp-lib/pull/11412)

Vitaliy-1:
Thanks, merged! This also requires porting to main. I've made a comment regarding new validation pattern, it could be adapter to this case for the main branch.

Hafsa-Naeem:
@Vitaliy-1 fixed tests for 3.5
[...]

PR #11659, opened 2025-07-18
pkp/pkp-lib#11389 Add validation to ensure userGroupId belongs to an author group when adding/editing contributor

Hafsa-Naeem:
for [#11389](https://github.com/pkp/pkp-lib/issues/11389)

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -1678,15 +1678,23 @@ class PKPSubmissionController extends PKPBaseController
         }
 
         // validate that if userGroupId is passed, it really is an author group and that email is present.
-        $validator = Validator::make($illuminateRequest->all(), [
+        $data = $illuminateRequest->only(['userGroupId']);
+        $validator = Validator::make($data, [
             'userGroupId' => [
+                'sometimes',
                 'nullable',
                 'integer',
-                Rule::exists('user_groups', 'user_group_id')
-                    ->where('role_id', Role::ROLE_ID_AUTHOR)
-                    ->where('context_id', $submission->getData('contextId')),
+                function ($attribute, $value, $fail) use ($submission) {
+                    $exists = UserGroup::query()
+                        ->withRoleIds([Role::ROLE_ID_AUTHOR])
+                        ->withContextIds([$submission->getData('contextId')])
+                        ->whereKey($value)
+                        ->exists();
+                    if (!$exists) {
+                        $fail(__('api.submission.400.invalidId', ['id' => $value]));
+                    }
+                },
             ],
-            'email' => ['required', 'email'],
         ]);
         if ($validator->fails()) {
             return response()->json($validator->errors()->toArray(), Response::HTTP_BAD_REQUEST);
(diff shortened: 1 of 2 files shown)


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1617,6 +1617,21 @@ public function addContributor(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_FORBIDDEN);
         }
 
+        // validate that if userGroupId is passed, it really is an author group and that email is present.
+        $validator = Validator::make($illuminateRequest->all(), [
```
touhidurabir: rather than passing all data through the validation, it's better to pass only what we want to validate . perhaps `$illuminateRequest->only(['userGroupId', ...])` will work .

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1617,6 +1617,21 @@ public function addContributor(Request $illuminateRequest): JsonResponse
...
+            'userGroupId' => [
+                'nullable',
```
touhidurabir: If we are adding `nullable` and also `integer` does that mean it can be present sometimes or not , right ? in that case better to add `sometimes` also as first rule .

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1617,6 +1617,21 @@ public function addContributor(Request $illuminateRequest): JsonResponse
...
+                'integer',
+                Rule::exists('user_groups', 'user_group_id')
+                    ->where('role_id', Role::ROLE_ID_AUTHOR)
+                    ->where('context_id', $submission->getData('contextId')),
```
touhidurabir: we are trying not to use database column aside from inside of model, repo and migration , basically not in core logics if possible as this conflict with our camelcase patten when the DB column name use snake case pattern . How about we try something like 
```php
function ($attribute, $value, $fail) use ($submission) {
    $exists = UserGroup::query()
        ->withRoleIds(.......)
        ->withContextIds(........)
        ->withUserGroupIds(.........)
        ->exists();

    if (!$exists) {
        $fail(__('api.submission.400.invalidId', ['id' => $value]));
    }
}
```
instead of `Rule::exists` .

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1617,6 +1617,21 @@ public function addContributor(Request $illuminateRequest): JsonResponse
...
+            ],
+            'email' => ['required', 'email'],
```
touhidurabir: do we need to validate the email here as I see we do validate it below when we run `Repo::author()->validate` as in `author.json` , it's validation rule set as `email_or_localhost` .

touhidurabir (changes requested):
@Hafsa-Naeem I have added some reviews . 

wondering as we have same logic in the `addContributor/editContributor` and associated with author, it is wise to move it to `PKP\author\Repository::validate` ? we can add it as part of `$validator->after` and no duplicate logic to handle as `userGroupId` is defined in `author.json` so it's part of author schema . @Vitaliy-1 your thought on this ?

touhidurabir:
> @Hafsa-Naeem I have added some reviews .
> 
> wondering as we have same logic in the `addContributor/editContributor` and associated with author, it is wise to move it to `PKP\author\Repository::validate` ? we can add it as part of `$validator->after` and no duplicate logic to handle as `userGroupId` is defined in `author.json` so it's part of author schema . @Vitaliy-1 your thought on this ?

Hi @Hafsa-Naeem , this looks good but can you give it a shot of what I described above if it make sense ? If we can have this in a single place rather than 2 duplicate code blocks , that better .

classes/author/Repository.php
```diff
@@ -135,6 +137,28 @@ public function validate($author, $props, Submission $submission, Context $conte
                 }
             }
         });
+        $validator->after(function ($validator) use ($props, $submission) {
+            if (
+                array_key_exists('userGroupId', $props) &&
+                !is_null($props['userGroupId']) &&
```
touhidurabir: I think just `isset` will work as that check for both the key exist and null .

classes/author/Repository.php
```diff
@@ -135,6 +137,28 @@ public function validate($author, $props, Submission $submission, Context $conte
                 }
             }
         });
+        $validator->after(function ($validator) use ($props, $submission) {
```
touhidurabir: we already have one `$validator->after` and I think we can use that one .

touhidurabir (approved):
@Hafsa-Naeem approved . I have left 2 minor comments and probably better to address those before merge .

Issue #11296, opened 2025-04-19
Publication – Add Versions Menu and Move "Create New Version" to the Side Menu of the Workflow Page


blesildaramirez:
### Description:
Under the "Publication" section of the submission workflow page, all versions created for a submission should be displayed as a collapsible menu in the left sidebar. Each version item should have child links for sections like "Title & Abstract", "Contributors", etc.
- [ ] Display the publication versions as collapsible menu
- [ ] The latest publication version should be selected by default when the workflow page loads.
- [ ] A "Create New Version" option should appear at the bottom of the Publication menu. The design should follow the provided mockup: 
https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-3.4-3.5?node-id=11903-19130&t=nJoIe0yWUFDmer9c-4 / https://github.com/pkp/pkp-lib/issues/4860
- [ ] Remove the "Create a New Version" button from the header of any Publication pages in the submission workflow.
- [ ] Remove the Versions dropdown from the header as well

blesildaramirez:
Hi @Tribunal33 , all the PRs for the frontend work have just been merged. You can now test this on the `main` branch, assigning this to you. Thank you!

Tribunal33:
This passes QA.  Confirmed that the tasks is complete.  

Not sure this applies in all cases: "The latest publication version should be selected by default when the workflow page loads." I believe this is only true for submissions in the published stage.  In the submission stage for example the submission stage is selected by default.

PR #11664, opened 2025-07-18
pkp/pkp-lib#11296 Sort the order of the version stages to show in dropdown field

blesildaramirez:


pages/dashboard/PKPDashboardHandler.php
```diff
@@ -158,6 +158,7 @@ public function index($args, $request)
 
         $versionStageOptions = [];
         $allVersionStages = VersionStage::cases();
+        usort($allVersionStages, fn($a, $b) => $a->order() <=> $b->order());
```
blesildaramirez, line 161: This change improves the ordering of the dropdown options for Publication Stage when creating a new publication version, to follow the expected order of publication stages:
OJS: AO, SM, AM, PR, VoR
OMP: AO, SM, AM. PR, VoR
OPS: AO, SM

blesildaramirez:
@defstat Thank you for the review.
> That said, the code you are using is PHP native, faster and also clear.

I'm after it being faster in execution, since this code will always be executed in the workflow and dashboard pages. Merging the current code to the main branch now..

PR #11668, opened 2025-07-21
Fix notification bug: Cast notification type to integer

kennydude:
Part of an issue with notifications is the text could not be displayed for notifications. This seems to be another issue with casting data to/from database which this was able to correct this issue for us.


asmecher:
Thanks, @kennydude -- merged and back-ported to `stable-3_5_0` for inclusion in 3.5.0-2 and later.

Issue #11673, opened 2025-07-24
[Reviewer Suggestion] | Unable to fetch appropriate submission in multi context system


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

In a multi context system or single context system without a context ID 1 , reviewer suggestion sometimes unable to fetch proper associated submission . The source of problem at https://github.com/pkp/pkp-lib/blob/5f37ccd8222df6d16e660aa3f7c2e62d77ac34ae/classes/submission/reviewer/suggestion/ReviewerSuggestion.php#L197 where where mistakenly passes true as second param which cause it to convert to 1 by PHP's type cast system and not finding the submission .

### Steps to Reproduce

Add reviewer suggestion in multi context system for the second context and try to assign that suggestion as reviewer for that submission . 

### Expected Result

Should be able to assign the reviewer suggestion without any problem . 

### Actual Result

Unable to load the submission after the suggestion to add as reviewer . Also at the assignment phase , there is endless spinner after assignment . 

### Application Version

OJS/OMP 3.5/main 

### Logs

error trace as 
```
[...]

PR #11674, opened 2025-07-24
pkp/pkp-lib#11673 Fix reviewer sugeestion resolve submission in multi context env

touhidurabir:
for pkp/pkp-lib#11673 .

### DIFF
--- a/classes/migration/upgrade/v3_5_0/I11673_AddMissingApprovalToReviewerSuggestion.php
+++ b/classes/migration/upgrade/v3_5_0/I11673_AddMissingApprovalToReviewerSuggestion.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_5_0/I11673_AddMissingApprovalToReviewerSuggestion.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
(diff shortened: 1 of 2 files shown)


classes/migration/upgrade/v3_5_0/I11673_AddMissingApprovalToReviewerSuggestion.php
```diff
@@ -0,0 +1,68 @@
...
+                        '=',
+                        'review_assignments.submission_id'
+                    );
+            })
+            ->whereNull('reviewer_suggestions.approved_at')
+            ->whereNull('reviewer_suggestions.reviewer_id')
+            ->select('reviewer_suggestions.*', 'users.user_id', 'review_assignments.review_id')
+            ->get();
+
+        $primaryKey = (new ReviewerSuggestion())->getKeyName();
```
asmecher: It would be better not to instantiate the model in the migration scripts; just hard-code `reviewer_suggestion_id` below. This will work for now, but down the road it might cause e.g. model lifecycle events to happen before a migration has run to adjust the schema. As much as possible, migration code should not interact with application code.

asmecher (changes requested):
@touhidurabir, just one small change to suggest, then please go ahead.

Issue #11686, opened 2025-07-28
[New Feature] | Publication - Contributors, Adding 'Email all authors' feature


salmanm2003:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

(This feature was discussed and approved for implementation in OJS 3.5)

Currently, editors cannot email all authors of a submission outside of decision emails. This functionality was available in OJS 2 and is a valuable feature that should be reintroduced.

The proposed feature would add a button that opens an email modal with all contributors pre-filled in the "To" field. Editors would be able to edit the recipients and optionally add CC and BCC fields.

The subject line would be blank and editable.

The email body would be populated from a predefined template (which could be left empty if a blank message is preferred by editors).

<img width="1440" height="862" alt="Image" src="https://github.com/user-attachments/assets/eee633e7-35de-4301-bccf-df6db15d5a53" />

### Steps to Reproduce

- 

### Expected Result

 - 

### Actual Result

 - 

### Application Version

OJS 3.5

salmanm2003:
Hi,
Here are the PRs for this feature. They are based on the main branch and can be applied to OJS 3.5 as well later.

Tribunal33:
@salmanm2003, I have a few questions which would help define the requirements. who, for example, has access to this email all button?  What are the permission levels?

Who is considered an author?  For instance, what about a section editor who has the author role and is in the participants list?

What about a contributor who is not an author but a translator?

Another more broader question is should we be encouraging un-tracked emails as a form of communication?

salmanm2003:
HI @Tribunal33,
Thanks for the questions — please see the clarifications below.

**Who, for example, has access to this email all button? What are the permission levels?**
Editors and Section Editors

**Who is considered an author? For instance, what about a section editor who has the author role and is in the participants list?**
If the Section Editor is assigned as a Author, they will see it, if they are not not assigned as a Section Editor then they will not see it

**What about a contributor who is not an author but a translator?**
The email is sent to everyone listed in the Contributors table, including translators. Given this, it may be better to label the button as “Email All” rather than “Email All Authors.”

**Another more broader question is should we be encouraging un-tracked emails as a form of communication?**
The emails will be tracked and recorded in the Activity Log.

PR #11687, opened 2025-07-28
pkp-lib#11686 Add email all authors feature to publication contributors-main

salmanm2003:
This is for [11686](https://github.com/pkp/pkp-lib/issues/11686)
`main` branch
OJS
pkp-lib repo

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -28,6 +28,7 @@ use APP\publication\Publication;
 use APP\section\Section;
 use APP\submission\Collector;
 use APP\submission\Submission;
+use PKP\log\SubmissionEmailLogEventType;
 use Illuminate\Http\JsonResponse;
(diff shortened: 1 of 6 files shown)


fradeve:
@jardakotesovec is there anything that is blocking this PR?

jardakotesovec:
@fradeve We dropped ball on this one. @Devika008 will get back to you with some details.  We are fine with the feature, but from technical point of view make sense to use our Composer component (which you can see for example when making editorial decision). And that might also require some work on the templates. But @Devika008 will catch up with you on that, to figure out whether the templates makes sense for this use case.

Devika008:
Hello @jardakotesovec and @fradeve,

Here is the design for the email screen :D Apologies for sharing this so late.

image

Figma Link -https://www.figma.com/design/Wf7sDlUg2372jaKKTJ0Mgz/OJS-Design?node-id=17047-10470&t=7c5iB9oRgw7tMBfn-4

Issue #11692, opened 2025-07-31
[New Submission] | Adding Keyword - Final Step summary, when reloading the page or selecting save and continue the keyword field is summarized incorrectly


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When going through the workflow of a new submission if the author adds some keywords and either refreshes the page or uses the save and continue option to come back later the final step summary shows [object Object] for each keyword instead of the keywords.

<img width="617" height="339" alt="Image" src="https://github.com/user-attachments/assets/ac1143a6-f9c5-452d-ba09-6399de0e85a2" />

### Steps to Reproduce

1. Login with author or JM
2. Start a new submission
3. Complete all the steps until the review Summary page.
4. Observe that the keywords are accurate
5. Either refresh the page or save for later.
6. If you refreshed the page then click continue until you get back to the summary review page
a. Observe the Details section with **keywords**
7. If you choose Save for Later button.  Select Complete the submission from the editorial activity submission list button
8. Upon returning you should already be on the summary and review step.
a. Observe the Details section with **keywords**

### Expected Result

- [ ] Should be able to refresh the page and display the keywords without displaying the object data type.
- [ ] Should be able to return to the incomplete submission and display the keywords without displaying the object data type. 

### Actual Result

Shows the object datatype instead of the keywords.  
This does not prevent the user from completing the submission or editing section.  After editing any section the keywords did show up so this seems like a UI only issue.

### Application Version

3.5 and main OJS

taslangraham:
@blesildaramirez @kaitlinnewson Could you give the small PRs listed in https://github.com/pkp/pkp-lib/issues/11692#issue-3281773539 a review, please? I'll port to main once the PRs for stable3.5 are approved

kaitlinnewson:
All approved @taslangraham 👍

Tribunal33:
This passes QA on main and 3.5 OJS.  Keywords are displayed both after page reload and returning to the incomplete submission.  Both requirements have passed.

<img width="604" height="281" alt="Image" src="https://github.com/user-attachments/assets/4ce71af9-4c6f-4426-baef-7ec8f87ea6ab" />

PR #11694, opened 2025-08-01
pkp/pkp-lib#11692 Fix keywords display

taslangraham:
For pkp/pkp-lib#11692

### DIFF
--- a/templates/submission/review-publication-field.tpl
+++ b/templates/submission/review-publication-field.tpl
@@ -44,19 +44,18 @@
     >
         {if $type === 'array'}
             <template v-if="publication.{$localizedProp|escape} && publication.{$localizedProp|escape}.length">
-                <template v-if="'{$dataField}'">
+                {if $dataField}
                     {{
                         publication.{$localizedProp|escape}
                             .map(item => item['{$dataField}'])
                             .join(t('common.commaListSeparator'))
                     }}
-                </template>
-                <template v-else>
-                    {{
-                        publication.{$localizedProp|escape}
-                            .join(t('common.commaListSeparator'))
-                    }}
-                </template>
+               {else}
+                   {{
+                       publication.{$localizedProp|escape}
+                           .join(t('common.commaListSeparator'))
+                   }}
+               {/if}
             </template>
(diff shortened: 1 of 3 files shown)


kaitlinnewson:
Hi @taslangraham, it looks like this change causes the keywords not to display on the initial submission review (only commas), or in the case where there is only 1 keyword.

taslangraham:
> Hi @taslangraham, it looks like this change causes the keywords not to display on the initial submission review (only commas), or in the case where there is only 1 keyword.

Thanks for catching that, @kaitlinnewson. I’ve fixed it. The issue was caused by publication metadata (keywords, subjects, etc.) in the API response not being properly hydrated.

I also noticed that this [issue](https://github.com/pkp/pkp-lib/issues/11692) was affecting the `Supporting Agencies`, `Disciplines`, and `Subjects` metadata. I’ve updated the PR to address those as well.

taslangraham:
There are a few failing tests that I need to look into

templates/submission/review-publication-field.tpl
```diff
@@ -43,12 +44,19 @@
...
+                            .map(item => item['{$dataField}'])
+                            .join(t('common.commaListSeparator'))
+                    }}
+                </template>
+                <template v-else>
+                    {{
+                        publication.{$localizedProp|escape}
+                            .join(t('common.commaListSeparator'))
+                    }}
+                </template>
```
blesildaramirez: It would be better to use the Smarty `{if}` condition here since `$dataField` is a Smarty variable.
Something like this:
```suggestion
                {if $dataField}
                    {{
                        publication.{$localizedProp|escape}
                            .map(item => item['{$dataField}'])
                            .join(t('common.commaListSeparator'))
                    }}
                {else}
                    {{
                        publication.{$localizedProp|escape}
                            .join(t('common.commaListSeparator'))
                    }}
                {/if}
```

blesildaramirez (approved):
Approving the PR. Just a small note - the current `v-if` works, but it might be better to use a Smarty `{if}` since the variable is already a Smarty one.

Issue #11677, opened 2025-07-24
[Disabled user] | New submission workflow - final step, there is a fatal error when trying to complete a new submission


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

I believe this is caused when a new submission is expecting to add editors to the new submission but one of them has been disabled.  

To set this up you will have to have preconditions met.
- [ ] Section editor assigned to the submission (Settings Journal > Section > Edit > Editorial assignment
- [ ] Disabled editor (section editor Stephanie Berardo used from default dataset)

### Steps to Reproduce

1. Login with an author role
2. Navigate to create new submission
3. Complete the submission with typical data it isn't important for the bug
4. Once at the final step in the new submission workflow select `Submit` button
5. Select `Submit` on the popup dialogue window.

### Expected Result

The submission should be successful and not error displayed. 
Requirements to pass the bug.  Should be tested on OJS, OMP, and OPS by disabling existing user.
- [x] Will not recieve an email
- [x] Will be removed from the Journal > Sections > Editorial Assignments list
- [x] Will be removed from particpants list
- [x] Will not be completely removed from the submission and can be brought back once enabled again.  
- [x] From Original steps -> Author can complete a submission without an error. 

### Actual Result

Error (see log for details) and page does not continue.  

<img width="782" height="278" alt="Image" src="https://github.com/user-attachments/assets/b2237a53-0f13-4ccd-b3bf-07a139bb7c3a" />

### Application Version

OJS, OMP, OPS 3.5.0-1 but older versions too

### Logs

[...]

PR #11706, opened 2025-08-06
pkp/pkp-lib#11677 Don't auto assign disabled editors to new submissions

taslangraham:
For pkp/pkp-lib#11677


classes/context/SubEditorsDAO.php
```diff
@@ -197,16 +197,23 @@ public function assignEditors(Submission $submission, Context $context): Collect
...
         $assignments = collect($assignments)->mapWithKeys(fn ($assignment, $key) => [$assignment->userId . '-' . $assignment->userGroupId => $assignment]);
-
         $userGroups = UserGroup::query()
             ->withContextIds([$submission->getData('contextId')])
             ->get();
 
         $userGroupIds = $userGroups->keys();
 
-        $assignments = $assignments->filter(function ($assignment) use ($userGroupIds) {
+        $activeUsers = Repo::user()
```
asmecher: @taslangraham, rather than building out a whole new set of queries to filter by active users, I'd suggest building this capability into the `getBySubmissionGroupIds` function directly in the form of a new `$enabledOnly` parameter. (Eventually the is DAO will be refactored to newer patterns, and this can be more cleanly done with a `Collector`, but let's not do that yet.) This would be less code, and would avoid additional queries.

asmecher (approved):

Issue #11699, opened 2025-08-05
[REST API] | Tasks and Discussions - get the list of tasks


Vitaliy-1:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Fatal error when trying to get the list of task/discussions as an author from the API, when one of the discussions doesn't have a due date. 
As an example, endpoint `GET`:`http://localhost:8000/publicknowledge/api/v1/submissions/21/stage/1/tasks`

### Steps to Reproduce

Add a discussion through the API without a due date. Hit get the list of the discussions endpoint

### Expected Result

Should return a list of discussions, e.g.:
```
{
    "items": [
        {
            "id": 4,
            "type": 1,
            "assocId": 21,
            "stageId": 1,
            "status": 1,
            "createdBy": 18,
            "dateDue": null,
            "participants": [
                {
                    "id": 8,
                    "userId": 3,
                    "isResponsible": false
                },
                {
                    "id": 9,
                    "userId": 6,
                    "isResponsible": false
                },
                {
                    "id": 10,
                    "userId": 18,
                    "isResponsible": false
                }
            ]
        }
    ],
    "itemMax": 1
}
```

### Actual Result

Return a fatal error

### Application Version

OJS main

### Logs
```
Error: Call to a member function format() on null in /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/api/v1/submissions/resources/TaskResource.php:32
Stack trace:
#0 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Collections/HigherOrderCollectionProxy.php(65): PKP\API\v1\submissions\resources\TaskResource->toArray(Object(Illuminate\Http\Request))
#1 [internal function]: Illuminate\Support\HigherOrderCollectionProxy->Illuminate\Support\{closure}(Object(PKP\API\v1\submissions\resources\TaskResource), 4)
#2 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Collections/Arr.php(609): array_map(Object(Closure), Array, Array)
#3 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(799): Illuminate\Support\Arr::map(Array, Object(Closure))
[...]

PR #11720, opened 2025-08-21
pkp/pkp-lib#11699 Enrich task resource with additional data

Vitaliy-1:


api/v1/submissions/resources/EnrichesWithData.php
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @var array $data [
+     *   'submission' => Submission,
+     *   'stageAssignments' => Collection,
+     *   'users' => LazyCollection,
+     *   'userGroups' => Collection,
+     *   'reviewAssignments' => Collection,
+     * ]
+     */
+    protected array $data = [];
```
touhidurabir: here we may start expecting different set of data with different keys and right now no way to know or validate the passed extra data format . Wondering having this trait a abstract method like `requiredKyes` or something that will return an array and any resource class will be force to implement that with custom keys as it needed . 

Then we can have a `validate` method to validate the passed data as 
```php
protected function validateData(array $data): void
{
    $requiredKeys = $this->requiredKyes(); // abstract method in each resource class to have
    foreach ($requiredKeys as $key) {
        if (!array_key_exists($key, $data)) {
            throw new \InvalidArgumentException("Missing required data key: $key");
        }
    }
}
```
and in the constructor as 
```php
public function __construct($resource, $key = null, ?array $data = [])
{
    $this->validateData($data);
    $this->data = $data ?? [];
    parent::__construct($resource);
}
```
it's just that right now we are too blind to know what that resource class needed as extra it it's used multiple places .

api/v1/submissions/resources/EnrichesWithData.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function __construct($resource, $key = null, ?array $data = [])
+    {
+        if (!empty($data)) {
+            $this->data = $data;
+        }
```
touhidurabir: as we kind of and should only except array , any need to mark it nullable ? and we can do it as `$this->data = $data` as it's also marked as array in the properly declaration .

api/v1/submissions/resources/EnrichesWithData.php
```diff
@@ -0,0 +1,87 @@
...
+             *
+             */
+            public function toArray(Request $request): array
+            {
+                return $this->collection->map->toArray(
+                    $request,
+                    $this->data
+                )->all();
+            }
+        };
```
touhidurabir: just some experiment , it is possible to pass the data to each collection class as (not tested but theoretically may work )
```php
$this->collection = $resource->map(function ($item) use ($collects, $data) {
    return in_array(EnrichesWithData::class, class_uses_recursive($collects))
        ? new $collects($item, null, $data)
        : new $collects($item);
});
```
by adding this int the constructor . we are forced to have the `toArray` as `public function toArray(Request $request, ?array $data = []): array`  that is always defined `$data` there when we are using the trait there also . And in that case we wont have to override the `toArray` here also . But need to consider the performance also here .

api/v1/submissions/resources/EditorialTaskParticipantResource.php
```diff
@@ -17,15 +17,97 @@
...
+            $this->data = $data;
+        }
+
+        [
+            'submission' => $submission,
+            'stageAssignments' => $stageAssignments,
+            'users' => $users,
+            'userGroups' => $userGroups,
+            'reviewAssignments' => $reviewAssignments,
+        ] = $this->data;
```
touhidurabir: this kind of require to know what when have passed and then get that back , just wondering what do you thing of following approach in the trait to have 
```php
public function getData(string ...$keys): array
{
    $result = [];
    foreach ($keys as $key) {
        if (!array_key_exists($key, $this->data)) {
            throw new \InvalidArgumentException(
                "Missing required data key: $key"
            );
        }
        $result[] = $this->data[$key];
    }
    return $result;
}
```
and we can use it like from resource class
```php
[$submission, $users] = $this->getData('submission', 'users')
```
nothing special or important so up to you .

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2112,6 +2163,127 @@ public function deleteTask(Request $illuminateRequest): JsonResponse
...
+        }
+
+        if (!$editTask->dateClosed) {
+            return response()->json([
+                'error' => __('api.409.resourceActionConflict'),
+            ], Response::HTTP_CONFLICT);
+        }
+
+        $editTask->fill(['dateClosed' => null])->save();
+        $editTask->refresh();
```
touhidurabir, line 2215: just out of curiosity, why not just `->update`  ?

classes/editorialTask/EditorialTask.php
```diff
@@ -59,7 +60,9 @@ class EditorialTask extends Model
         'dateDue' => 'datetime:Y-m-d',
         'createdBy' => 'int',
         'type' => 'int',
-        'status' => 'int',
+        'dateStarted' => 'datetime',
+        'dateClosed' => 'datetime',
+        'title' => 'string',
```
touhidurabir, line 65: is this going to be a multilingual input ?

touhidurabir (commented):
@Vitaliy-1 , I have added few review comments but mostly as questions and suggestions, please take a look. 
BTW, this approach to override the resource class with a trait looks quite good .

classes/editorialTask/EditorialTask.php
Vitaliy-1: No

api/v1/submissions/resources/EnrichesWithData.php
Vitaliy-1: I'll double check, but since we can pass the data in 2 places, we need either to override or merge

Issue #10045, opened 2024-06-12
Email variable for the review reminder is not being shared with the UI


jonasraoni:
**Describe the bug**
The email variable `reviewAssignmentUrl` for the template REVIEW_REMIND (review reminder) is not being passed to the UI.

I also think the constants/classes should be renamed to "review reminder".

**To Reproduce**
1. Enter on the review stage.
2. Click on the button to send a reminder to the reviewer.
3. See the variable is not stylized and also doesn't appear in the list of variables.

**What application are you using?**
OJS 3.4

**Additional information**
![image](https://github.com/pkp/pkp-lib/assets/361921/c98fc3dd-5b06-4e25-9262-260743eebeee)

jonasraoni:
Also, it was unexpected for me that the email editor is working differently from the one at the Workflow > Emails, where the variables are displayed in plain text and the icon to insert variables is different:
![image](https://github.com/pkp/pkp-lib/assets/361921/ee3ab134-e505-4df3-80a7-c7cb6a344074)

Assuming it's a work in progress.

Vitaliy-1:
@jonasraoni, is at an old template that should be replaced during upgrade?

Vitaliy-1:
Apart from it, reviewAssignmentUrl isn't exposed deliberately: https://github.com/pkp/pkp-lib/blob/3536404e545ef14ffed6fbe390916d74fc1c875c/controllers/grid/users/reviewer/form/ReviewReminderForm.php#L89

Vitaliy-1:
> Also, it was unexpected for me that the email editor is working differently from the one at the Workflow > Emails, where the variables are displayed in plain text and the icon to insert variables is different

I recall variables are replaced before being sent

Vitaliy-1:
The new wording for the template is different: https://github.com/pkp/ojs/blob/main/locale/en/emails.po#L251
I assume, the old template had custom modifications and that's why it was preserved, right?

jonasraoni:
@Vitaliy-1 this is a mix of UI and backend issue.

The UI of the email editor is different, the email editor that appears when I click on the button `Submission > Workflow > Review > Send Reminder` has:
- `+` button, which allows me to select variables, but only the `URL` is available (it refers to the `{$passwordResetUrl}`.
- The variable is not displayed as text, but as a kind of icon, which looks nice for me, see:
[...]

PR #11728, opened 2025-08-23
pkp/pkp-lib#10045 Fix email preview to show placeholder text instead of unresolved reviewAssignmentUrl link

Hafsa-Naeem:
for #10045


controllers/grid/users/reviewer/form/ReviewReminderForm.php
```diff
@@ -90,6 +90,11 @@ public function initData()
         // Don't expose the reviewer's one-click access URL to editors
         $data[ReviewAssignmentEmailVariable::REVIEW_ASSIGNMENT_URL] = '{$' . ReviewAssignmentEmailVariable::REVIEW_ASSIGNMENT_URL . '}';
         $body = Mail::compileParams($template->getLocalizedData('body'), $data);
+        $body = preg_replace(
```
Vitaliy-1, line 93: I think we should change the wording in the template to localize it. Then the variable can be replaced by a placeholder

PR #11735, opened 2025-08-24
Add BeforeSend hook to Mailer

kennydude:
Adds a hook to allow email changes before sending an email - i.e changing emails to follow policies

For example to force a from address no matter what (i.e we don't want the return path or any address pointing to the contact email address), or to add content to every email being sent out

### DIFF
--- a/classes/mail/Mailer.php
+++ b/classes/mail/Mailer.php
@@ -159,7 +159,7 @@ class Mailer extends IlluminateMailer
     {
         $sentMessage = null;
         try {
-            Hook::call('Email::BeforeSend', [$message]);
+            Hook::call('Email::send::before', [$message]);
             $sentMessage = $this->transport->send($message, Envelope::create($message));
         } catch (TransportException $e) {
             error_log($e->getMessage());
--- a/classes/mail/Mailer.php
+++ b/classes/mail/Mailer.php
@@ -159,7 +159,7 @@ class Mailer extends IlluminateMailer
     {
         $sentMessage = null;
         try {
-            Hook::call('Email::send::before', [$message]);
+            Hook::call('Email::send::before', ['message' => $message, 'mailer' => $this]);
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @kennydude! That's a good idea -- but we have a slightly different naming convention for these. See e.g. [this hook](https://github.com/pkp/pkp-lib/blob/7d2d3738477c449f8023e6651cc5673fe3536ee9/classes/affiliation/Repository.php#L176). Can you adjust the naming in your PR to suit, then I can get it merged?

kennydude:
Changed the name to `Email::send::before` which should be more in-line with the convention @asmecher

asmecher:
Sorry, one more thing -- please add `$this` to the parameter list, so that any hook registrant wanting to configure the mailer before sending can do that too. Thanks!

kennydude:
Sure - would that be best as the second argument?

asmecher:
I don't have a strong preference for ordering. We made some changes [a few years ago](https://github.com/pkp/pkp-lib/discussions/8089) to make hook registration a little less arbitrary; now that this is in place, you can typehint your callback function:

```php
Hook::add('Email::send::before', function(Message $message, Mailer $mailer]) {
    ...
});
```
This reduces the chances that you'll get your parameters mistyped or out of order.

Further, if you like, you can also use named parameters:
```php
Hook::call('Email::send::before', ['message' => $message, 'mailer' => $this]);
```
...and then you can put your callback parameters in any order you like, as long as they're named (and typed) right.

We don't have strict requirements for these, but I think the named parameter formulation is the most helpful.

kennydude:
Ah that is quite nice to see. I've updated accordingly

asmecher:
Thanks, I've merged this to `main` and cherry-picked into `stable-3_5_0` for release in OJS/OMP/OPS 3.5.0-2!

Issue #11730, opened 2025-08-24
[API] Fatal error if any request to `submissions` endpoints includes invalid submission ID


Vitaliy-1:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Any `submissions` GET API endpoint throws a fatal error instead of 404 not found response error when invalid submission ID is specified in the request. 

### Steps to Reproduce

1. Login as an admin and set API key in the profile menu
2. Open any tool for making API request, e.g., Postman
4. Hit the following endpoint with unexisting submission ID: `http://localhost:8000/publicknowledge/api/v1/submissions/200`
5. See the 500 internal server error

### Expected Result

Response with 404 not found error

### Actual Result

500 internal server error

### Application Version

OJS main branch

### Logs

```
 Symfony\Component\HttpKernel\Exception\NotFoundHttpException in /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/classes/security/authorization/internal/SubmissionRequiredPolicy.php:43
Stack trace:
#0 [internal function]: PKP\security\authorization\internal\SubmissionRequiredPolicy->PKP\security\authorization\internal\{closure}()
#1 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/classes/security/authorization/AuthorizationDecisionManager.php(158): call_user_func(Object(Closure))
#2 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/classes/core/PKPBaseController.php(365): PKP\security\authorization\AuthorizationDecisionManager->decide()
#3 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/api/v1/submissions/PKPSubmissionController.php(487): PKP\core\PKPBaseController->authorize(Object(APP\core\Request), Array, Array)
#4 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/classes/middleware/PolicyAuthorizer.php(78): PKP\API\v1\submissions\PKPSubmissionController->authorize(Object(APP\core\Request), Array, Array)
#5 /Users/vitalii/Documents/development/ojs-main/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(209): PKP\middleware\PolicyAuthorizer->handle(Object(Illuminate\Http\Request), Object(Closure))
[...]

PR #11737, opened 2025-08-25
pkp/pkp-lib#11730 Handle exception at policy authorization

touhidurabir:
for pkp/pkp-lib#11730 .

### DIFF
--- a/classes/middleware/PolicyAuthorizer.php
+++ b/classes/middleware/PolicyAuthorizer.php
@@ -30,6 +30,8 @@ use PKP\plugins\interfaces\HasAuthorizationPolicy;
 use PKP\security\authorization\PolicySet;
 use PKP\security\authorization\AuthorizationPolicy;
 use ReflectionFunction;
+use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
+use Throwable;
 
 class PolicyAuthorizer
 {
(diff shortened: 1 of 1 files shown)


classes/middleware/PolicyAuthorizer.php
```diff
@@ -74,20 +76,34 @@ public function handle(Request $request, Closure $next)
...
+            $hasAuthorized = $routeController->authorize(
+                $pkpRequest,
+                $args,
+                $roleAssignments->toArray()
+            );
+        } catch (NotFoundHttpException $e) {
+            $exceptionStatusCode = $e->getStatusCode();
+        } catch (Throwable $e) {
+            $exceptionStatusCode = $e->getCode();
+        }
```
taslangraham: Can this be done in a single catch block? Maybe something like the following:

```php
} catch (Throwable $e) {
    $exceptionStatusCode = $e instanceof NotFoundHttpException 
        ? $e->getStatusCode() 
        : $e->getCode();
}
```

taslangraham (changes requested):
Hey @touhidurabir, this looks good in general. Just the one question.

classes/middleware/PolicyAuthorizer.php
touhidurabir: This works for now . I have a feeling that in we may need to generate different response based on different exception , but we can tackle that when such need arise in future .

asmecher (approved):

Issue #11540, opened 2025-06-18
Metadata | Add support for a plain language summary


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Some journals will want the ability to add a plain-language summary ("lay summary"), distinct from the more formal Abstract.

This is supported in some 3rd party standards ; see e.g. [JATS4R's `@abstract-type` attribute](https://jats4r.niso.org/abstracts/).

### Requirements
- [x] User will be able to enable Plain Language Summary via Workflow Settings > MetaData > Plain Language Summary with checkbox (seen below)

<img width="734" height="147" alt="Image" src="https://github.com/user-attachments/assets/499bc23b-a0e0-4aaf-98e6-b83d9fae8c4b" />

- [x] Once enabled 3 new options are available by radio button: Do not request a plain language summary from the author during submission, Ask the author to provide a plain language summary during submission and Require the author to provide a plain language summary bfore accepting their submission
- [x] Once enabled and either option to display the plain language summary during a new submission is choosen a new open text field will be provided to the user during the submission process.   
- [ ] If the Require the author to provide a plain language summary before accepting their submission is selected then a required `*` for the field and a warning message will be displayed if it has not been filled in.  The submission will not complete until this field is filled in.

<img width="268" height="92" alt="Image" src="https://github.com/user-attachments/assets/01770fc9-8231-4bed-8acc-821335def9d4" />

- [x] When enabled the plain language summary field can be seen in a submission under Publication > Title & Abstract > Plain Language Summary.
- [x] If the field option is required then the field wiill have an `*` indicating so and the option to save an empty Plain language summary field will not work
- [x] When enabled and option to not display during submission process will hide field on step 2 of submission workflow but still display in `Publication > Title & Abstract` and on the Front-office submission view page.

<img width="381" height="114" alt="Image" src="https://github.com/user-attachments/assets/7546677c-baa3-41cc-b166-e18eeb7442f2" />

### Steps to Reproduce

(N/A)

### Expected Result

(N/A)

### Actual Result

(N/A)

### Application Version

OJS/OMP/OPS 3.6.0

### PRs 
~~[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11570~~
~~[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/4961~~
[...]

PR #11741, opened 2025-08-26
pkp/pkp-lib#11540 Missing validations for PLS

touhidurabir:
for pkp/pkp-lib#11540 .


classes/submission/traits/HasWordCountValidation.php
```diff
@@ -0,0 +1,61 @@
...
+     * Validate the word count of a publication localized data
+     *
+     * @return array
+     */
+    protected function validateWordCount(
+        Context $context,
+        Submission $submission,
+        int $wordLimit,
+        array $localizedData = [],
+        string $errorMessageKey = 'publication.wordCountLong'
```
asmecher: The `publication.wordCountLong` key is specific to the abstract -- I'd suggest renaming it to `publication.abstract.wordCountLong` and removing it as a default parameter  value here. (You'll need to rename they key in other locales besides English too.)

Issue #11557, opened 2025-06-22
Centralise the publication's object metadata structure


touhidurabir:
**Description:**
With the https://github.com/pkp/pkp-lib/issues/1550 , it's now possible to have plugins that can offer keyword from certain vocab list . This will add few more extra information in the `controlled_vocab_entry_settings` table such as `source`, `identifier` . However this also add 2 different vocab structure in the `publication` object . 

basically in the publication object , the metadata say `keywords` are presented as
```
[
    "en" => [
      "Sociology",
      "Chemical sciences",
      "vocab-test-101",
    ],
  ]
```
The above structure what is expected when trying to retrieve the `keywords` from `publication` object as `$publication->getData('keywords')`

But then to have more meta information , we are expecting data structure like
```
    [
      "name" => "Sociology",
      "identifier" => "5.3",
      "source" => "Frascati",
    ],
    [
      "name" => "Chemical sciences",
      "identifier" => "1.4",
      "source" => "Frascati",
    ],
    [
      "name" => "vocab-test-101",
    ],
  ]
```
through the publication object, it's not possible to get that structure but need to use method as `Repo::controlledVocav()->getBySymbolic(...)` and se the last param `$asEntryData` to `true` .

However this cause major issue with component rendering to edit publication/submission/metadata as this structure with extra information is missing in the `publication` object , it cause the extra meta information missing in vue component which lead to silently loosing existing extra meta information . More details of this case at https://github.com/pkp/pkp-lib/issues/11481 . 

As part of fix for https://github.com/pkp/pkp-lib/issues/11481 , we have applied the rehydrate approach only in those scenario when we need to make sure these extra meta information do persist before editing/updating , more details at https://github.com/pkp/pkp-lib/issues/11481#issuecomment-2973542794 . This approach is ok but still there is now 2 different data structure of metadata which leads to major technical debt in 3.5/main branch . 

**Solution:** 
The best solution should be to centralise to one certain metadata structure for `publication` object and to the structure of 
```
    [
      "name" => "Sociology",
      "identifier" => "5.3",
      "source" => "Frascati",
    ],
    [
      "name" => "Chemical sciences",
      "identifier" => "1.4",
      "source" => "Frascati",
    ],
    [
      "name" => "vocab-test-101",
    ],
  ]
```
without making a breaking change to `$publication->getData('keywords')` which return back as 
```
[
    "en" => [
      "Sociology",
      "Chemical sciences",
      "vocab-test-101",
    ],
  ]
[...]

PR #11764, opened 2025-08-31
pkp/pkp-lib#11557 Use single format for keywords in publication object

ajnyga:


api/v1/reviews/PKPReviewController.php
```diff
@@ -181,7 +181,7 @@ public function getHistory(Request $illuminateRequest): JsonResponse
         }
 
         $publicationAbstract = $publication->getData('abstract');
-        $publicationKeywords = $publication->getData('keywords');
+        $publicationKeywords = array_map(fn($items) => array_column($items, 'name'), (array) $publication->getData('keywords'));
```
touhidurabir: this is a bit hard to understand what's it's doing , how about something like 
```php
collect($publication->getData('keywords'))
    ->map(
        fn(array $items): array => collect($items)
            ->pluck("name")
            ->all()
    )
    ->all();
```
but nothing important, so up to you .

classes/components/forms/publication/PKPMetadataForm.php
```diff
@@ -167,8 +167,7 @@ protected function getVocabEntryData(string $symbolic): array
             $symbolic,
             Application::ASSOC_TYPE_PUBLICATION,
             $this->publication->getId(),
-            [],
-            Repo::controlledVocab()::AS_ENTRY_DATA
+            []
```
touhidurabir: as the default is set to empty array , no need to pass an empty array as argument

classes/controlledVocab/Repository.php
```diff
@@ -51,9 +49,8 @@ public function getBySymbolic(
         int $assocType,
         ?int $assocId,
         ?array $locales = [],
-        bool $asEntryData = !Repository::AS_ENTRY_DATA
-    ): array
-    {
+        bool $asStringOnly = false
```
touhidurabir: usual convention for a boolean param is to be `true` by default as setting something as `true` as method argument is kind of confusing . Also I think it's better to stick with const event if thats true/false as thats easy to understand what it means.

classes/publication/DAO.php
```diff
@@ -360,6 +360,8 @@ protected function deleteAuthors(int $publicationId)
      */
     protected function setControlledVocab(Publication $publication)
     {
+        # f11557
+        # These now return an object with all the data
```
touhidurabir: if add comment , better to add the issue number as `pkp/pkp-lib#11557` or just the issue link . Also better to add a bit more description comments like return an array of localize vocabs with name, identifier or source .

api/v1/reviews/PKPReviewController.php
ajnyga: ok

classes/components/forms/publication/PKPMetadataForm.php
ajnyga: check

classes/controlledVocab/Repository.php
ajnyga: ok, but then most calls to getBySymbolic would need to set it explicitly to false since most of the time we are returning something else than a string?

ajnyga: nevermind the last comment, gotcha.

returned the const from the old code, that is probably what you meant

classes/publication/DAO.php
ajnyga: will just remove this, it was just for myself...

ajnyga:
I think the failing unit tests have something todo with using the publication object to fetch the keywords and subjects in dc11?

ajnyga:
I noticed now that I missed one direct call to the getBySymbolic in classes/search/ArticleSearch.php

However, I changed the new logic for fetching data from the publication object to other places and if the approach is ok I can add it to ArticleSearch as well.

The unit test like I mention above is probably due to the way the test is written. I am not that familiar with the unittests part of the application though. The dc11 itself seems to work ok when you look for example the oai endpoint. I will study the unittests more carefully and will try to suggest a new implementation for that.

PR #11765, opened 2025-09-01
Contributor Roles and Type

jyhein:
Contributor Roles and Type

- Add contributor Roles to Author
- Add contributor Type to Author
- Remove user group functions from Author
- Remove translator from user groups
- Add possibility to add more roles than author and translator

- Selected contributor type changes what contributor form fields are required
- More than one contributor role can be selected in the contributor form

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -30,6 +30,7 @@ use APP\submission\Submission;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Mail;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\Validator;
@@ -39,6 +40,8 @@ use PKP\affiliation\Affiliation;
 use PKP\API\v1\submissions\formRequests\AddTask;
 use PKP\API\v1\submissions\formRequests\EditTask;
 use PKP\API\v1\submissions\resources\TaskResource;
+use PKP\author\contributorRole\ContributorRoleIdentifier;
+use PKP\author\contributorRole\ContributorType;
 use PKP\components\forms\FormComponent;
 use PKP\components\forms\publication\PKPCitationsForm;
 use PKP\components\forms\publication\PKPMetadataForm;
@@ -745,7 +748,8 @@ class PKPSubmissionController extends PKPBaseController
         if ($submitAsUserGroup->roleId === Role::ROLE_ID_AUTHOR) {
             $author = Repo::author()->newAuthorFromUser($request->getUser(), $submission, $context);
             $author->setData('publicationId', $publication->getId());
-            $author->setUserGroupId($submitAsUserGroup->id);
+            $author->setData('contributorType', ContributorType::PERSON->getName());
+            $author->setContributorRoles([ContributorRoleIdentifier::AUTHOR->getName()], $context->getId());
             $authorId = Repo::author()->add($author);
             Repo::publication()->edit($publication, ['primaryContactId' => $authorId]);
         }
@@ -1677,10 +1681,12 @@ class PKPSubmissionController extends PKPBaseController
             ], Response::HTTP_FORBIDDEN);
         }
         $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_AUTHOR, $illuminateRequest->input());
+        // Remove extra data that is irrelevant to the selected contributor type
+        $this->removeIrrelevantContributorTypeData($params);
 
         // Allows author ORCID request email to be triggered from frontend before author ID exists
         $shouldSendOrcidEmail = false;
-        switch ($params['orcid']) {
+        switch ($params['orcid'] ?? null) {
             case null:
                 unset($params['orcid']);
                 break;
@@ -1697,6 +1703,10 @@ class PKPSubmissionController extends PKPBaseController
             $submissionContext = app()->get('context')->get($submission->getData('contextId'));
         }
 
+        if (isset($params['contributorRoles'])) {
+            $params['contributorRoles'] = $this->contributorRoleTranslationsToParams($params['contributorRoles'], $submissionContext->getId());
+        }
+
         $errors = Repo::author()->validate(null, $params, $submission, $submissionContext);
 
         if (!empty($errors)) {
@@ -1704,6 +1714,7 @@ class PKPSubmissionController extends PKPBaseController
         }
         $affiliationParams = $params['affiliations'] ?? [];
         unset($params['affiliations']);
+
         $author = Repo::author()->newDataObject($params);
         $newId = Repo::author()->add($author);
 
@@ -1834,6 +1845,8 @@ class PKPSubmissionController extends PKPBaseController
         }
 
         $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_AUTHOR, $illuminateRequest->input());
+        // Remove extra data that is irrelevant to the selected contributor type
+        $this->removeIrrelevantContributorTypeData($params);
         $params['id'] = $author->getId();
 
         $submissionContext = $request->getContext();
@@ -1853,6 +1866,10 @@ class PKPSubmissionController extends PKPBaseController
             return response()->json([
                 'error' => __('api.submissions.403.userCantEdit'),
             ], Response::HTTP_FORBIDDEN);
+        }
+
+        if (isset($params['contributorRoles'])) {
+            $params['contributorRoles'] = $this->contributorRoleTranslationsToParams($params['contributorRoles'], $submissionContext->getId());
         }
 
         $errors = Repo::author()->validate($author, $params, $submission, $submissionContext);
@@ -2225,10 +2242,7 @@ class PKPSubmissionController extends PKPBaseController
 
         $submissionLocale = $submission->getData('locale');
         $locales = $this->getPublicationFormLocales($context, $submission);
-        $authorUserGroups = UserGroup::withRoleIds([Role::ROLE_ID_AUTHOR])
-            ->withContextIds([$submission->getData('contextId')])
-            ->get();
-        $publicationLicenseForm = new PKPPublicationLicenseForm($publicationApiUrl, $locales, $publication, $context, $authorUserGroups);
+        $publicationLicenseForm = new PKPPublicationLicenseForm($publicationApiUrl, $locales, $publication, $context);
 
         return response()->json($this->getLocalizedForm($publicationLicenseForm, $submissionLocale, $locales), Response::HTTP_OK);
     }
@@ -2454,6 +2468,7 @@ class PKPSubmissionController extends PKPBaseController
             'givenName',
             'familyName',
             'preferredPublicName',
+            'organizationName',
         ];
         $affiliationProps = [
             'name',
@@ -2468,8 +2483,38 @@ class PKPSubmissionController extends PKPBaseController
                         Repo::affiliation()->edit($affiliation, $editProps($affiliation, $affiliationProps));
                     }
                 }
-                if (!($contributor->getData('givenName')[$newLocale] ?? null)) {
+                if (!(
+                        $contributor->getContributorType() === ContributorType::PERSON->getName() && ($contributor->getData('givenName')[$newLocale] ?? null) ||
+                        $contributor->getContributorType() === ContributorType::ORGANIZATION->getName() && ($contributor->getData('organizationName')[$newLocale] ?? null)
+                    )) {
                     Repo::author()->edit($contributor, $editProps($contributor, $contributorProps));
+                }
+            });
+    }
+
+    /**
+     * Remove irrelevant data based on the selected contributor type
+     */
+    protected function removeIrrelevantContributorTypeData(array &$params): void
+    {
+        $contributorType = $params['contributorType'];
+        $typeFields = [
+            ContributorType::PERSON->getName() => Author::CONTRIBUTOR_TYPE_PERSON_FORM_FIELDS,
+            ContributorType::ORGANIZATION->getName() => Author::CONTRIBUTOR_TYPE_ORGANIZATION_FORM_FIELDS,
+            ContributorType::ANONYMOUS->getName() => Author::CONTRIBUTOR_TYPE_ANONYMOUS_FORM_FIELDS,
+        ];
+        $selectedTypeField = Arr::pull($typeFields, $contributorType);
+        $dataToRemove = collect($typeFields)
+            ->flatten()
+            ->diff($selectedTypeField)
+            ->unique();
+        $mprops = $dataToRemove
+            ->intersect(app()->get('schema')->getMultilingualProps(PKPSchemaService::SCHEMA_AUTHOR));
+
+        $dataToRemove
+            ->each(function (string $key) use (&$params, $mprops) {
+                if (isset($params[$key])) {
+                    $params[$key] = $mprops->contains($key) ? array_map(fn ($_) => null, $params[$key]) : null;
                 }
             });
     }
@@ -2497,5 +2542,14 @@ class PKPSubmissionController extends PKPBaseController
             FILTER_VALIDATE_BOOLEAN,
             FILTER_NULL_ON_FAILURE
         );
+    }
+
+    private function contributorRoleTranslationsToParams(?array $roles, int $contextId): array
+    {
+        // If values are roles
+        if (is_array($roles) && array_intersect($roles, ContributorRoleIdentifier::getRoles())) {
+            return Repo::contributorRole()->getByContextIdAndIdentifiers($contextId, $roles);
+        }
+        return $roles;
     }
 }
--- a/classes/author/Author.php
+++ b/classes/author/Author.php
@@ -18,13 +18,26 @@
 
 namespace PKP\author;
 
+use APP\facades\Repo;
+use PKP\author\contributorRole\ContributorType;
 use PKP\affiliation\Affiliation;
 use PKP\facades\Locale;
 use PKP\identity\Identity;
-use PKP\userGroup\UserGroup;
 
 class Author extends Identity
 {
+    public const CONTRIBUTOR_TYPE_PERSON_FORM_FIELDS = [
+        'givenName', 'familyName', 'preferredPublicName', 'email', 'country',
+        'rorId', 'url', 'biography', 'affiliations',
+    ];
+
+    public const CONTRIBUTOR_TYPE_ORGANIZATION_FORM_FIELDS = [
+        'organizationName', 'email', 'country', 'rorId',
+        'url', 'biography', 'affiliations',
+    ];
+
+    public const CONTRIBUTOR_TYPE_ANONYMOUS_FORM_FIELDS = ['email', 'country'];
+
     /**
      * Get the default/fall back locale the values should exist for
      * (see LocalizedData trait)
@@ -58,6 +71,39 @@ class Author extends Identity
         return $this->getFamilyName($this->getData('submissionLocale'));
     }
 
+    /**
+     * Get the contributor's localized name based on contributor type.
+     *
+     * @param bool $preferred If the preferred public name should be used, if exist
+     * @param bool $familyFirst False / default: Givenname Familyname
+     * 	If true: Familyname, Givenname
+     * @param string $preferredLocale The locale the full name is requested for. If null, the user locale will be used.
+     */
+    public function getFullName(bool $preferred = true, bool $familyFirst = false, ?string $preferredLocale = null): string
+    {
+        return match ($this->getData('contributorType')) {
+            ContributorType::PERSON->getName() => parent::getFullName($preferred, $familyFirst, $preferredLocale),
+            ContributorType::ORGANIZATION->getName() => $this->getLocalizedOrganizationName($preferredLocale),
+            ContributorType::ANONYMOUS->getName() => __('submission.submit.contributorType.anonymous'),
+        };
+    }
+
+    /**
+     * Get the contributor's names based on contributor type.
+     *
+     * @param bool $preferred If the preferred public name should be used, if exist
+     * @param bool $familyFirst False / default: Givenname Familyname
+     * 	If true: Familyname, Givenname
+     */
+    public function getFullNames(bool $preferred = true, bool $familyFirst = false): array
+    {
+        return match ($this->getData('contributorType')) {
+            ContributorType::PERSON->getName() => parent::getFullNames($preferred, $familyFirst),
+            ContributorType::ORGANIZATION->getName() => $this->getOrganizationName(),
+            ContributorType::ANONYMOUS->getName() => [Locale::getLocale() => __('submission.submit.contributorType.anonymous')],
+        };
+    }
+
     //
     // Get/set methods
     //
@@ -83,26 +129,6 @@ class Author extends Identity
     }
 
     /**
-     * Set the user group id
-     *
-     * @param int $userGroupId
-     */
-    public function setUserGroupId($userGroupId)
-    {
-        $this->setData('userGroupId', $userGroupId);
-    }
-
-    /**
-     * Get the user group id
-     *
-     * @return int
-     */
-    public function getUserGroupId()
-    {
-        return $this->getData('userGroupId');
-    }
-
-    /**
      * Set whether or not to include in browse lists.
      *
      * @param bool $include
@@ -120,29 +146,6 @@ class Author extends Identity
     public function getIncludeInBrowse()
     {
         return $this->getData('includeInBrowse');
-    }
-
-    /**
-     * Get the "show title" flag (whether or not the title of the role
-     * should be included in the list of submission contributor names).
-     * This is fetched from the user group for performance reasons.
-     *
-     * @return bool
-     */
-    public function getShowTitle()
-    {
-        return $this->getData('showTitle');
-    }
-
-    /**
-     * Set the "show title" flag. This attribute belongs to the user group,
-     * NOT the author; fetched for performance reasons only.
-     *
-     * @param bool $showTitle
-     */
-    public function _setShowTitle($showTitle)
-    {
-        $this->setData('showTitle', $showTitle);
     }
 
     /**
@@ -183,33 +186,6 @@ class Author extends Identity
     public function setSequence($sequence)
     {
         $this->setData('seq', $sequence);
-    }
-
-    /**
-     * Get the user group for this contributor.
-     *
-     * @return \PKP\userGroup\UserGroup
-     */
-    public function getUserGroup()
-    {
-        //FIXME: should this be queried when fetching Author from DB? - see #5231.
-        static $userGroup; // Frequently we'll fetch the same one repeatedly
-        if (!$userGroup || $this->getData('userGroupId') != $userGroup->id) {
-            $userGroup = UserGroup::find($this->getData('userGroupId'));
-        }
-        return $userGroup;
-    }
-
-    /**
-     * Get a localized version of the User Group
-     *
-     * @return string
-     */
-    public function getLocalizedUserGroupName()
-    {
-        $userGroup = $this->getUserGroup();
-        return $userGroup ? $userGroup->getLocalizedData('name') : null;
-
     }
 
     /**
@@ -287,8 +263,34 @@ class Author extends Identity
         );
     }
 
+        /**
+     * Get organization name.
+     */
+    public function getOrganizationName(string $locale): string|array|null
+    {
+        return $this->getData('organizationName', $locale);
+    }
+
     /**
-     * Get contrubutor credit roles and degrees
+     * Get the localized given name
+     *
+     * @return string
+     */
+    public function getLocalizedOrganizationName(string $preferredLocale = null)
+    {
+        return $this->getLocalizedData('organizationName', $preferredLocale);
+    }
+
+    /**
+     * Set organization name.
+     */
+    public function setOrganizationName(string $organizationName, string $locale): void
+    {
+        $this->setData('organizationName', $organizationName, $locale);
+    }
+
+    /**
+     * Get contributor credit roles and degrees.
      */
     public function getCreditRoles(): array
     {
@@ -296,10 +298,30 @@ class Author extends Identity
     }
 
     /**
-     * Set contrubutor credit roles and degrees.
+     * Set contributor credit roles and degrees.
      */
     public function setCreditRoles(?array $creditRoles): void
     {
         $this->setData('creditRoles', $creditRoles);
+    }
+
+    /**
+     * Get contributor roles.
+     */
+    public function getContributorRoles(bool $withTranslations = true): array
+    {
+        $roles = $this->getData('contributorRoles') ?? [];
+        return $withTranslations ? $roles : array_keys($roles);
+    }
+
+    /**
+     * Set contributor roles.
+     */
+    public function setContributorRoles(?array $contributorRoles, int $contextId = null): void
+    {
+        $this->setData('contributorRoles', !$contextId
+            ? $contributorRoles
+            : Repo::contributorRole()->getByContextIdAndIdentifiers($contextId, $contributorRoles)
+        );
     }
 }
--- a/classes/author/DAO.php
+++ b/classes/author/DAO.php
@@ -53,7 +53,7 @@ class DAO extends EntityDAO
         'includeInBrowse' => 'include_in_browse',
         'publicationId' => 'publication_id',
         'seq' => 'seq',
-        'userGroupId' => 'user_group_id',
+        'contributorType' => 'contributor_type',
     ];
 
     /**
@@ -162,6 +162,8 @@ class DAO extends EntityDAO
 
         $author->setCreditRoles(Repo::creditContributorRole()->getCreditRolesByContributorId($author->getId()));
 
+        $author->setContributorRoles(Repo::creditContributorRole()->getContributorRolesByContributorId($author->getId()));
+
         return $author;
     }
 
@@ -179,7 +181,9 @@ class DAO extends EntityDAO
             Repo::affiliation()->add($affiliation);
         }
 
-        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(), $newAuthorId);
+        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(false), $newAuthorId);
+
+        Repo::creditContributorRole()->addContributorRoles($author->getContributorRoles(false), $newAuthorId);
 
         return $newAuthorId;
     }
@@ -191,7 +195,9 @@ class DAO extends EntityDAO
     {
         Repo::affiliation()->saveAffiliations($author);
 
-        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(), $author->getId());
+        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(false), $author->getId());
+
+        Repo::creditContributorRole()->addContributorRoles($author->getContributorRoles(false), $author->getId());
 
         parent::_update($author);
     }
--- a/classes/author/Repository.php
+++ b/classes/author/Repository.php
@@ -19,6 +19,8 @@ use APP\author\DAO;
 use APP\core\Request;
 use APP\facades\Repo;
 use APP\submission\Submission;
+use PKP\author\contributorRole\ContributorRoleIdentifier;
+use PKP\author\contributorRole\ContributorType;
 use PKP\context\Context;
 use PKP\identity\Identity;
 use PKP\plugins\Hook;
@@ -136,24 +138,18 @@ class Repository
             if (isset($props['publicationId']) && !$validator->errors()->get('publicationId')) {
                 $publication = Repo::publication()->get($props['publicationId']);
                 if (!$publication) {
-                    $validator->errors()->add('publicationId', __('author.publicationNotFound'));
+                    $validator->errors()->adsd('publicationId', __('author.publicationNotFound'));
                 } elseif ($publication->getData('status') === PKPSubmission::STATUS_PUBLISHED) {
                     $validator->errors()->add('publicationId', __('author.editPublishedDisabled'));
                 }
             }
-            // userGroupId must be an Author group within the current context
-            if (isset($props['userGroupId']) && !$validator->errors()->get('userGroupId')) {
-                $userGroupId = (int) $props['userGroupId'];
-                $exists = UserGroup::query()
-                    ->withRoleIds([Role::ROLE_ID_AUTHOR])
-                    ->withContextIds([$submission->getData('contextId')])
-                    ->whereKey($userGroupId)
-                    ->exists();
-
-                if (!$exists) {
+            // At least one contributor role
+            if (isset($props['contributorRoles']) && !$validator->errors()->get('contributorRoles')) {
+                $contributorRoles = $props['contributorRoles'];
+                if (!$contributorRoles) {
                     $validator->errors()->add(
-                        'userGroupId',
-                        __('api.submission.400.invalidId', ['id' => $userGroupId])
+                        'contributorRoles',
+                        __('api.submission.400.emptyContributorRoles')
                     );
                 }
             }
@@ -264,6 +260,8 @@ class Repository
         $author->setUrl($user->getUrl());
         $author->setIncludeInBrowse(1);
         $author->setOrcid($user->getOrcid());
+        $author->setData('contributorType', ContributorType::PERSON->getName());
+        $author->setContributorRoles([ContributorRoleIdentifier::AUTHOR->getName()], $context->getId());
 
         Hook::call('Author::newAuthorFromUser', [$author, $user]);
 
@@ -284,7 +282,7 @@ class Repository
             ->getMany();
 
         foreach ($authors as $author) {
-            if (empty($author->getGivenName($newLocale))) {
+            if (ContributorType::PERSON->getName() && empty($author->getGivenName($newLocale))) {
                 if (empty($author->getFamilyName($newLocale)) && empty($author->getPreferredPublicName($newLocale))) {
                     // if no name exists for the new locale
                     // copy all names with the old locale to the new locale
@@ -296,6 +294,8 @@ class Repository
                     // copy only the given name with the old locale to the new locale, because the given name is required
                     $author->setGivenName($author->getGivenName($oldLocale), $newLocale);
                 }
+            } else if (ContributorType::ORGANIZATION->getName() && !$author->getOrganizationName($newLocale)) {
+                $author->setOrganizationName($author->ggetOrganizationName($oldLocale), $newLocale);
             }
 
             $newAffiliations = [];
--- a/classes/author/contributorRole/ContributorRole.php
+++ b/classes/author/contributorRole/ContributorRole.php
@@ -1,0 +1,158 @@
+<?php
+
+/**
+ * @file lib/pkp/classes/author/contributorRole/ContributorRole.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class ContributorRole
+ *
+ * @brief ContributorRole model class
+ */
+
+namespace PKP\author\contributorRole;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use PKP\author\creditContributorRole\CreditContributorRole;
+use PKP\core\traits\ModelWithSettings;
+
+class ContributorRole extends Model
+{
+    use ModelWithSettings;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$table
+     */
+    protected $table = 'contributor_roles';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Model::$primaryKey
+     */
+    protected $primaryKey = 'contributor_role_id';
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\GuardsAttributes::$guarded
+     */
+    protected $guarded = [
+        'contextId',
+        'contributorRoleIdentifier',
+    ];
+
+    /**
+     * @copydoc \PKP\core\traits\ModelWithSettings::getSettingsTable
+     */
+    public function getSettingsTable(): string
+    {
+        return 'contributor_role_settings';
+    }
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasTimestamps::$timestamps
+     */
+    public $timestamps = false;
+
+    /**
+     * @copydoc \Illuminate\Database\Eloquent\Concerns\HasAttributes::casts
+     */
+    protected function casts(): array
+    {
+        return [
+            'contributor_role_id' => 'integer',
+            'context_id' => 'integer',
+            'contributor_role_identifier' => 'string',
+        ];
+    }
+
+    /**
+     * @copydoc \PKP\core\traits\ModelWithSettings::getSchemaName
+     */
+    public static function getSchemaName(): ?string
+    {
+        return null;
+    }
+
+    /**
+     * @copydoc \PKP\core\traits\ModelWithSettings::getMultilingualProps
+     */
+    public function getMultilingualProps(): array
+    {
+        return array_merge($this->multilingualProps, ['name']);
+    }
+
+    /**
+     * @copydoc \PKP\core\traits\ModelWithSettings::getSettings
+     */
+    public function getSettings(): array
+    {
+        return array_merge($this->settings, ['name']);
+    }
+
+    /**
+     * Accessor and Mutator for primary key => id
+     */
+    protected function id(): Attribute
+    {
+        return Attribute::make(
+            get: fn ($value, $attributes) => $attributes[$this->primaryKey] ?? null,
+            set: fn ($value) => [$this->primaryKey => $value],
+        )->shouldCache();
+    }
+
+    /**
+     * The contributor roles that belong to the contributor.
+     */
+    public function contributorRole(): BelongsTo
+    {
+        return $this->belongsTo(CreditContributorRole::class, 'contributor_role_id', 'contributor_role_id');
+    }
+
+    public function getData(): array
+    {
+        return [];
+    }
+
+    /**
+     * Scope a query to include specific context id.
+     */
+    public function scopeWithContextId(Builder $query, int $contextId): Builder
+    {
+        return $query->where('context_id', $contextId);
+    }
+
+    /**
+     * Scope a query to include specific contributor role id.
+     */
+    public function scopeWithRoleId(Builder $query, int $roleId): Builder
+    {
+        return $query->where('contributor_role_id', $roleId);
+    }
+
+    /**
+     * Scope a query to include specific contributor role identifiers.
+     */
+    public function scopeWithIdentifiers(Builder $query, array $identifiers): Builder
+    {
+        return $query->whereIn('contributor_role_identifier', $identifiers);
+    }
+
+    /**
+     * Scope a query to only include entries for a specific contributor role id and locales
+     */
+    public function scopeWithSettings(Builder $query): Builder
+    {
+        return $query->whereExists(
+            fn ($query) => $query
+                ->select($this->primaryKey)
+                ->from($this->getSettingsTable())
+                ->whereColumn(
+                    "{$this->getSettingsTable()}.{$this->primaryKey}",
+                    "{$this->getTable()}.{$this->primaryKey}"
+                )
+        );
+    }
+}
--- a/classes/author/contributorRole/Repository.php
+++ b/classes/author/contributorRole/Repository.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * @file classes/author/contributorRole/Repository.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Repository
+ *
+ * @brief A repository to manage actions related to contributor roles
+ */
+
+namespace PKP\author\contributorRole;
+
+use APP\core\Application;
+
+class Repository
+{
+    // Add or edit. Use role id to edit. Use identifier and context id to add or edit.
+    public function add(array $translations, ?string $identifier = null, ?int $contextId = null, ?int $roleId = null): int
+    {
+        return ContributorRole::updateOrCreate(
+            [
+                ...$roleId
+                    ? ['contributor_role_id' => $roleId] /* Edit */
+                    : ['context_id' => $contextId, 'contributor_role_identifier' => $identifier] /* Add/Edit */
+            ],
+            [
+                'name' => $translations,
+                ...($roleId && $identifier) /* Edit */
+                    ? ['contributor_role_identifier' => $identifier]
+                    : []
+            ]
+        )
+        ->id;
+    }
+
+    public function delete(int $contextId = null, string $identifier = null, int $roleId = null): void
+    {
+        if ($roleId) {
+            ContributorRole::query()
+                ->withRoleId($roleId)
+                ->delete();
+            return;
+        }
+        ContributorRole::query()
+            ->withContextId($contextId)
+            ->withIdentifiers([$identifier])
+            ->delete();
+    }
+
+    /**
+     * Get localized, or all, context's contributor roles with settings in assoc: identifier => translation(s).
+     */
+    public function getByContextId(int $contextId, string $locale = null): array
+    {
+        return ContributorRole::query()
+            ->withContextId($contextId)
+            ->withSettings()
+            ->get()
+            ->mapWithKeys(fn (ContributorRole $role) => [
+                $role->contributor_role_identifier => $locale ? $role->getLocalizedData('name', $locale) : $role->name
+            ])
+            ->toArray();
+    }
+
+    /**
+     * Get localized, or all, context's contributor role with settings in assoc: identifier => translation(s).
+     */
+    public function getByContextIdAndIdentifiers(int $contextId, array $identifiers, string $locale = null): array
+    {
+        return ContributorRole::query()
+            ->withContextId($contextId)
+            ->withIdentifiers($identifiers)
+            ->withSettings()
+            ->get()
+            ->mapWithKeys(fn (ContributorRole $role) => [
+                $role->contributor_role_identifier => $locale ? $role->getLocalizedData('name', $locale) : $role->name
+            ])
+            ->toArray();
+    }
+
+    /**
+     * Get context's contributor role with settings by identifier in assoc: identifier => translation(s).
+     */
+    public function getByRoleId(int $roleId, string $locale = null): array
+    {
+        return ContributorRole::query()
+            ->withRoleId($roleId)
+            ->withSettings()
+            ->get()
+            ->mapWithKeys(fn (ContributorRole $role) => [
+                $role->contributor_role_identifier => $locale ? $role->getLocalizedData('name', $locale) : $role->name
+            ])
+            ->toArray();
+    }
+}
--- a/classes/author/creditContributorRole/CreditContributorRole.php
+++ b/classes/author/creditContributorRole/CreditContributorRole.php
@@ -15,10 +15,12 @@
 namespace PKP\author\creditContributorRole;
 
 use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Casts\Attribute;
-use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Query\JoinClause;
+use PKP\author\contributorRole\ContributorRole;
 
 class CreditContributorRole extends Model
 {
@@ -77,6 +79,14 @@ class CreditContributorRole extends Model
     public function scopeWithContributorId(Builder $query, int $contributorId): Builder
     {
         return $query->where('contributor_id', $contributorId);
+    }
+
+    /**
+     * Scope a query to only include contributors with a specific contributor role id.
+     */
+    public function scopeWithContributorRoleId(Builder $query, int $roleId): Builder
+    {
+        return $query->where('contributor_role_id', $roleId);
     }
 
     /**
--- a/classes/author/creditContributorRole/Repository.php
+++ b/classes/author/creditContributorRole/Repository.php
@@ -14,44 +14,86 @@
 
 namespace PKP\author\creditContributorRole;
 
+use APP\core\Application;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Arr;
+use PKP\author\contributorRole\ContributorRole;
 use PKP\author\creditRole\CreditRole;
 
 class Repository
 {
-    public function addCreditRoles(array $newCreditRoles, int $contributorId): void
+    public function addContributorRoles(array $newRoleIdentifiers, int $contributorId, int $contextId = null): void
     {
-        $creditRolesWithId = Arr::mapWithKeys(
-            CreditRole::withCreditRoleIdentifiers(Arr::pluck($newCreditRoles, 'role'))->get()->toArray(),
-            fn (array $cr): array => [$cr['creditRoleIdentifier'] => $cr['creditRoleId']]
-        );
+        $contextId ??= Application::get()->getRequest()->getContext()->getId();
+        $roleIds = ContributorRole::withContextId($contextId)
+            ->withIdentifiers($newRoleIdentifiers)
+            ->select(['contributor_role_id'])
+            ->get()
+            ->pluck('contributor_role_id');
 
-        $currentCreditRoles = CreditContributorRole::query()
+        // Disallow removal of all roles
+        if ($roleIds->isEmpty()) return;
+
+        // Delete old roles
+        CreditContributorRole::query()
+            ->withContributorId($contributorId)
+            ->whereNotNull('contributor_role_id')
+            ->whereNotIn('contributor_role_id', $roleIds)
+            ->delete();
+
+        $roleIds->each(fn (int $roleId) =>
+            CreditContributorRole::updateOrCreate(
+                ['contributor_id' => $contributorId, 'contributor_role_id' => $roleId],
+                []
+            ));
+    }
+
+    public function addCreditRoles(array $newRoles, int $contributorId): void
+    {
+        $identifiersWithIds = CreditRole::withCreditRoleIdentifiers(Arr::pluck($newRoles, 'role'))
+            ->get()
+            ->mapWithKeys(fn (CreditRole $role) => [$role->creditRoleIdentifier => $role->creditRoleId]);
+
+        // Delete old roles
+        CreditContributorRole::query()
             ->withContributorId($contributorId)
             ->whereNotNull('credit_role_id')
-            ->select(['credit_contributor_role_id', 'credit_role_id'])
-            ->get();
+            ->whereNotIn('credit_role_id', $identifiersWithIds)
+            ->delete();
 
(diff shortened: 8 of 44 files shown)


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1834,6 +1845,8 @@ public function editContributor(Request $illuminateRequest): JsonResponse
         }
 
         $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_AUTHOR, $illuminateRequest->input());
+        // Remove extra data that is irrelevant to the selected contributor type
+        $this->removeIrrelevantContributorTypeData($params);
```
bozana: I am wondering if we should rather put this into the Repository validate() function...
Here and above...
Let me think... :thinking:

classes/author/Author.php
```diff
@@ -58,48 +71,61 @@ public function getLocalizedFamilyName()
...
+     * @param string $preferredLocale The locale the full name is requested for. If null, the user locale will be used.
      */
-    public function getSubmissionId()
+    public function getFullName(bool $preferred = true, bool $familyFirst = false, ?string $preferredLocale = null): string
     {
-        return $this->getData('submissionId');
+        return match ($this->getData('contributorType')) {
+            ContributorType::PERSON->getName() => parent::getFullName($preferred, $familyFirst, $preferredLocale),
+            ContributorType::ORGANIZATION->getName() => $this->getLocalizedOrganizationName($preferredLocale),
+            ContributorType::ANONYMOUS->getName() => __('submission.submit.contributorType.anonymous'),
```
bozana: Provide $preferredLocale as parameter for __():
__('submission.submit.contributorType.anonymous', locale: $preferredLocale)
or
__('submission.submit.contributorType.anonymous', [], $preferredLocale)

classes/author/Author.php
```diff
@@ -287,19 +263,65 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
         );
     }
 
+        /**
+     * Get organization name.
+     */
+    public function getOrganizationName(string $locale): string|array|null
```
bozana: also null can be given, to get an array of all localized names, no? It also seems to be the default parameter, because above, in the function getFullNames() we call it without parameters.

classes/author/Author.php
```diff
@@ -287,19 +263,65 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+        return $this->getData('organizationName', $locale);
+    }
+
     /**
-     * Get contrubutor credit roles and degrees
+     * Get the localized given name
+     *
+     * @return string
+     */
+    public function getLocalizedOrganizationName(string $preferredLocale = null)
```
bozana: to add return type like this:
public function getLocalizedOrganizationName(string $preferredLocale = null): string
?

classes/author/DAO.php
```diff
@@ -179,7 +181,9 @@ public function insert(Author $author): int
             Repo::affiliation()->add($affiliation);
         }
 
-        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(), $newAuthorId);
+        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(false), $newAuthorId);
```
bozana: author->getCreditRoles() does not expect any parameter

classes/author/DAO.php
```diff
@@ -191,7 +195,9 @@ public function update(Author $author)
     {
         Repo::affiliation()->saveAffiliations($author);
 
-        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(), $author->getId());
+        Repo::creditContributorRole()->addCreditRoles($author->getCreditRoles(false), $author->getId());
```
bozana: author->getCreditRoles() does not expect any parameter

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2498,4 +2543,13 @@ private function validateVersionIsMinor(Request $illuminateRequest): ?bool
             FILTER_NULL_ON_FAILURE
         );
     }
+
+    private function contributorRoleTranslationsToParams(?array $roles, int $contextId): array
```
bozana: It seems $roles can be null as well? 
Then the return value can be null i.e. `?array` too?
I see we check above `if (isset($params['contributorRoles'])...` so maybe it can be only array?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2468,12 +2483,42 @@ protected function copyMultilingualData(Submission $submission, string $newLocal
...
+            ->flatten()
+            ->diff($selectedTypeField)
+            ->unique();
+        $mprops = $dataToRemove
+            ->intersect(app()->get('schema')->getMultilingualProps(PKPSchemaService::SCHEMA_AUTHOR));
+
+        $dataToRemove
+            ->each(function (string $key) use (&$params, $mprops) {
+                if (isset($params[$key])) {
+                    $params[$key] = $mprops->contains($key) ? array_map(fn ($_) => null, $params[$key]) : null;
```
bozana: Is `$_` locale?
is it not enough to just say `unset($params[$key])` ?

classes/author/Repository.php
```diff
@@ -136,24 +138,18 @@ public function validate($author, $props, Submission $submission, Context $conte
             if (isset($props['publicationId']) && !$validator->errors()->get('publicationId')) {
                 $publication = Repo::publication()->get($props['publicationId']);
                 if (!$publication) {
-                    $validator->errors()->add('publicationId', __('author.publicationNotFound'));
+                    $validator->errors()->adsd('publicationId', __('author.publicationNotFound'));
```
bozana: is this a typo?

classes/author/contributorRole/ContributorRole.php
```diff
@@ -0,0 +1,158 @@
...
+            set: fn ($value) => [$this->primaryKey => $value],
+        )->shouldCache();
+    }
+
+    /**
+     * The contributor roles that belong to the contributor.
+     */
+    public function contributorRole(): BelongsTo
+    {
+        return $this->belongsTo(CreditContributorRole::class, 'contributor_role_id', 'contributor_role_id');
```
bozana, line 107: Should this relationship also be present in the CreditRole ?

classes/author/contributorRole/ContributorRole.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public function getData(): array
```
bozana: Why do we need this function?

classes/author/contributorRole/ContributorRole.php
```diff
@@ -0,0 +1,158 @@
...
+     */
+    public function scopeWithIdentifiers(Builder $query, array $identifiers): Builder
+    {
+        return $query->whereIn('contributor_role_identifier', $identifiers);
+    }
+
+    /**
+     * Scope a query to only include entries for a specific contributor role id and locales
+     */
+    public function scopeWithSettings(Builder $query): Builder
```
bozana: Do we need this function?

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Get localized, or all, context's contributor roles with settings in assoc: identifier => translation(s).
+     */
+    public function getByContextId(int $contextId, string $locale = null): array
+    {
+        return ContributorRole::query()
+            ->withContextId($contextId)
+            ->withSettings()
```
bozana: Do we need withSettings()

classes/services/PKPContextService.php
```diff
@@ -585,6 +587,13 @@ public function add($context, $request)
         Repo::emailTemplate()->dao->installAlternateEmailTemplates($context->getId());
         Repo::emailTemplate()->dao->setTemplateDefaultUnrestirctedSetting($context->getId());
 
+        $contributorRoleTranslations = Arr::map(
+            ['default.groups.name.author', 'default.groups.name.translator'],
+            fn (string $str): array => Arr::mapWithKeys($context->getData('supportedLocales'), fn (string $l) => [$l => __($str, locale: $l)])
+        );
+        Repo::contributorRole()->add($contributorRoleTranslations[0], ContributorRoleIdentifier::AUTHOR->getName(), $context->getId());
+        Repo::contributorRole()->add($contributorRoleTranslations[1], ContributorRoleIdentifier::TRANSLATOR->getName(), $context->getId());
```
bozana: Shouldn't rather a new ContributorRole object be created and that object provided to the Repository add function? -- This is very unusual to provide this kind of parameters to the add...

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+namespace PKP\author\contributorRole;
+
+use APP\core\Application;
+
+class Repository
+{
+    // Add or edit. Use role id to edit. Use identifier and context id to add or edit.
+    public function add(array $translations, ?string $identifier = null, ?int $contextId = null, ?int $roleId = null): int
```
bozana: We usually provide an object that should be added to this function -- it is unusual to provide all properties here, but to rather construct an object somewhere else and then to call this function with the object.

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,99 @@
...
+                'name' => $translations,
+                ...($roleId && $identifier) /* Edit */
+                    ? ['contributor_role_identifier' => $identifier]
+                    : []
+            ]
+        )
+        ->id;
+    }
+
+    public function delete(int $contextId = null, string $identifier = null, int $roleId = null): void
```
bozana: Here as well, an object should be provided

The usual add, edit (that can get additional parameters as second argument), delete function should be provided here

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Constructor.
+     */
+    public function __construct(int $roleId = null)
+    {
+        $this->setRoleId($roleId);
+        parent::__construct('controllers/grid/settings/contributorRoles/form/contributorRoleForm.tpl');
+
+        // Validation checks for this form
+        $this->addCheck(new \PKP\form\validation\FormValidatorCustom($this, 'name', 'required', 'manager.setup.contributorRoles.nameRequired', fn (array $name) => count($name) === count(array_filter($name))));
```
bozana: I think here FormValidatorLocale can be used:
`$this->addCheck(new \PKP\form\validation\FormValidatorLocale($this, 'name', 'required', 'manager.setup.contributorRoles.nameRequired'));`

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
```diff
@@ -0,0 +1,135 @@
...
+     * @return bool
+     */
+    public function execute(...$functionArgs)
+    {
+        // Edit or add contributor role
+        $roleId = Repo::contributorRole()->add(
+            $this->getData('name'),
+            $this->getData('identifier'),
+            Application::get()->getRequest()->getContext()->getId(),
+            $this->getData('rowId')
```
bozana: Create a new ContributorRole object and provide it to the Repo add or edit function

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
```diff
@@ -0,0 +1,135 @@
...
+            ];
+        } else {
+            $this->_data = [
+                'name' => [],
+            ];
+        }
+
+        // grid related data
+        $this->_data['gridId'] = $args['gridId'];
+        $this->_data['rowId'] = $args['rowId'] ?? null;
```
bozana: initData function should get the object, if it exists and map the objects properties to the form fields (_data). 
So maybe here an eloquent object needs to be get: ContribuorRole::find($this->getRoleId()). It is much cleaner, I think.
Also no need to localize any data, I think -- all existing languages for the name should be provided and the form maps them automatically.
The identifier names are provided to the form template in the fetch function, because they do not change, they are always the same. S. also comment below.

Maybe you can take a look at the form GenreForm as example: it contains `submissionFileCategories` in the similar way as this form contains role identifiers.

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * @copydoc Form::fetch()
+     *
+     * @param null|mixed $template
+     */
+    public function fetch($request, $template = null, $display = false)
+    {
+        $templateMgr = TemplateManager::getManager($request);
+        $templateMgr->assign('contributorRoles', collect(ContributorRoleIdentifier::getRoles())
+            ->diff(array_keys(Repo::contributorRole()->getByContextId($request->getContext()->getId())))
```
bozana: Why is it here needed to get the diff? I think we just need to provide all possible ContributorRoleIdentifiers.
Then, in the form template: if the role has role ID we only display already existing identifier, else we display all for the user to choose one.

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
```diff
@@ -0,0 +1,135 @@
...
+        $roleId = Repo::contributorRole()->add(
+            $this->getData('name'),
+            $this->getData('identifier'),
+            Application::get()->getRequest()->getContext()->getId(),
+            $this->getData('rowId')
+        );
+
+        if (!$this->getRoleId()) {
+            $this->setRoleId($roleId);
+        }
```
bozana: I think there is no need for this here at the end

controllers/grid/settings/contributorRoles/form/ContributorRoleForm.php
bozana: also identifier is required, no?

schemas/author.json
```diff
@@ -40,6 +39,28 @@
 				"regex:/^[A-Z]{2}$/"
 			]
 		},
+		"contributorRoles": {
+			"type": "object",
```
bozana: isn't this an array?

schemas/author.json
```diff
@@ -40,6 +39,28 @@
...
+			"description": "The names of this contributor's roles in the publication, such as 'Author' or 'Translator', and the translations",
+			"apiSummary": true,
+			"readOnly": true,
+			"items": {
+				"type": "object",
+				"multilingual": "true"
```
bozana: shell here also the properties be defined/listed?

schemas/author.json
```diff
@@ -40,6 +39,28 @@
 				"regex:/^[A-Z]{2}$/"
 			]
 		},
+		"contributorRoles": {
+			"type": "object",
+			"description": "The names of this contributor's roles in the publication, such as 'Author' or 'Translator', and the translations",
```
bozana, line 47: Wouldn't it be better to have the array of objects here: That way we know also the role identifier and can match it when exporting?

pages/authorDashboard/PKPAuthorDashboardHandler.php
```diff
@@ -270,7 +269,7 @@ public function setupTemplate($request)
         })->values();
 
         // Get full details of the working publication and the current publication
-        $mapper = Repo::publication()->getSchemaMap($submission, $contextUserGroups, $contextGenres);
+        $mapper = Repo::publication()->getSchemaMap($submission, $contextGenres);
```
bozana, line 265: should then the arguments in the function getSchemaMap() in lib/pkp/classes/publication/Repository be changed as well?

classes/author/Author.php
```diff
@@ -287,19 +263,65 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
+    }
+
+    /**
+     * Set contributor roles.
+     */
+    public function setContributorRoles(?array $contributorRoles, int $contextId = null): void
+    {
+        $this->setData('contributorRoles', !$contextId
+            ? $contributorRoles
+            : Repo::contributorRole()->getByContextIdAndIdentifiers($contextId, $contributorRoles)
```
bozana: Somehow I think it would be cleaner to have array of objects/ContributorRoles in this author's property.

classes/author/Repository.php
```diff
@@ -284,7 +282,7 @@ public function changePublicationLocale($publicationId, $oldLocale, $newLocale)
             ->getMany();
 
         foreach ($authors as $author) {
-            if (empty($author->getGivenName($newLocale))) {
+            if (ContributorType::PERSON->getName() && empty($author->getGivenName($newLocale))) {
```
bozana: Shouldn't we check if author->getData('contributorType') == ContributorType::PERSON->getName() instead?

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,246 @@
...
+    public function getMany(Request $illuminateRequest): JsonResponse
+    {
+        $roles = ContributorRole::query()->withContextId($this->getRequest()->getContext()->getId());
+        return response()->json(Repo::contributorRole()->getSchemaMap()->summarizeMany($roles->get())->values(), Response::HTTP_OK);
+    }
+
+    /**
+     * Get form data
+     */
+    public function getContributorRoleFormData(Request $illuminateRequest): JsonResponse
```
jardakotesovec: I would suggest not making it form specific. Just expose it as any other data. I think its still valuable even if its not data from the database. 

Endpoint could be contributorRoles/contributorRoleIdentifier
just with response
["EDITOR", "CHAIR"]

Tagging @ewhanson If he has some preference on this.

jyhein: Just to note that the identifiers are not all but what are currently not in use

jardakotesovec: @jyhein Not sure if I understand. Can you elaborate more? Thanks.

jyhein: The identifiers are enum and some of them are saved to the db with translations. That function returns the identifiers that are not in the db for the form. Identifier per context is unique in db

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,246 @@
...
+     * Create or update a role.
+     *
+     * Used internally to handle both new role creation and editing existing ones.
+     */
+    private function saveRole(Request $illuminateRequest, ?ContributorRole $role): JsonResponse
+    {
+        $context = $this->getRequest()->getContext();
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_CONTRIBUTOR_ROLE, $illuminateRequest->input());
+        $params['id'] = $role?->id;
+        $params['contextId'] ??= $context->getId();
```
bozana: Context ID parameter should not be provided via API, it should only be taken from the request, I think

classes/author/contributorRole/ContributorRole.php
```diff
@@ -0,0 +1,159 @@
...
+            [
+                'name' => $name,
+                ...($roleId && $identifier) /* Edit */
+                    ? ['contributor_role_identifier' => $identifier]
+                    : []
+            ]
+        );
+    }
+
+    public function delete(): bool
```
bozana, line 121: is this needed when it only calls the parent function?

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,73 @@
...
+     * announcements to their schema
+     */
+    public function getSchemaMap(): maps\Schema
+    {
+        return app('maps')->withExtensions($this->schemaMap);
+    }
+
+    public function validate(?ContributorRole $role, array $props, Context $context): array
+    {
+        $locales = $context->getSupportedLocales();
```
bozana: it should be getSupportedFormLocales()

schemas/contributorRole.json
```diff
@@ -0,0 +1,38 @@
+{
+	"title": "ContributorRole",
+	"description": "Contributor role for contributors",
+	"required": [
+		"id",
+		"identifier",
+		"name"
```
bozana, line 7: contextId is missing here

schemas/contributorRole.json
```diff
@@ -0,0 +1,38 @@
+{
+	"title": "ContributorRole",
+	"description": "Contributor role for contributors",
+	"required": [
+		"id",
```
bozana: we usually do not have the id here, I think

api/v1/contributorRoles/ContributorRoleController.php
bozana: The heck for readOnly, as e.g. here https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/PKPSubmissionController.php#L613, is also needed

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,73 @@
...
+        $identifiers = ContributorRoleIdentifier::getRoles();
+        $schema = ContributorRole::getSchemaName();
+        // Identifier and context id required when adding
+        // Name must have all the ui locales, and each filled
+        $rules = array_merge_recursive($this->schemaService->getValidationRules($schema, $locales), [
```
bozana: I think there is no need to write an extra validation function/rules -- you can use already existing functions here, they check the schema, allowed, required, and multilingual fields. 
Also the name is not needed in all UI locales, but only in context primary locale -- as for all other similar objects. Then, if there is no name in an UI locale, we fall back to the primary locale using getLocalizedData('name').
It also does not make sense to save the name in all UI locales, they can change with the time.
So validation would be something like this:
```
        $validator = ValidatorFactory::make(
            $props,
            $schemaService->getValidationRules($this->dao->schema, $allowedLocales)
        );

        ValidatorFactory::required(
            $validator,
            $role,
            $this->schemaService->getRequiredProps($this->dao->schema),
            $this->schemaService->getMultilingualProps($this->dao->schema),
            $allowedLocales,
            $primaryLocale
        );

        ValidatorFactory::allowedLocales($validator, $schemaService->getMultilingualProps(PKPSchemaService::SCHEMA_CONTRIBUTOR_ROLE), $allowedLocales);

```

classes/author/contributorRole/maps/Schema.php
```diff
@@ -0,0 +1,105 @@
...
+                    break;
+                case 'identifier':
+                    $output[$prop] = $item->contributor_role_identifier;
+                    break;
+                case 'name':
+                    $output[$prop] = $item->name;
+                    break;
+            }
+        }
+
```
bozana, line 103: You would then also need this:
```
$output = $this->schemaService->addMissingMultilingualValues($this->schema, $output, $this->context->getSupportedFormLocales());
ksort($output);

```

classes/author/creditContributorRole/Repository.php
```diff
@@ -14,46 +14,78 @@
...
-            fn (array $cr): array => [$cr['creditRoleIdentifier'] => $cr['creditRoleId']]
-        );
+        $contextId ??= Application::get()->getRequest()->getContext()->getId();
+        $roleIds = ContributorRole::withContextId($contextId)
+            ->withIdentifiers($newRoleIdentifiers)
+            ->select(['contributor_role_id'])
+            ->get()
+            ->pluck('contributor_role_id');
+
+        // Disallow removal of all roles
```
bozana: I think the comment is wrong -- we look if any contributor role exists for the given identifier, and in this case it does not exist.

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,142 @@
...
+            ->unique();
+        $person = ContributorType::PERSON->getName();
+
+        // Create Author and Translator contributor roles
+        $contextIds
+            ->each(fn (int $contextId) =>
+                $translations
+                    ->filter(fn (array $t) => $t['contextId'] === $contextId)
+                    ->each(fn (array $t) =>
+                        ContributorRole::add(['name' => $t['name'], 'identifier' => $roles[$t['key']], 'contextId' => $contextId])
```
bozana: We should use direct queries, and not the existing functions, in case something changes in the code, the old migration script would not be needed to be adapted too

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,142 @@
...
+
+        // Add contributor roles and types to existing contributors
+        DB::table('authors as a')
+            ->select(['a.author_id', 'a.user_group_id', 's.context_id'])
+            ->join('publications as p', 'a.publication_id', '=', 'p.publication_id')
+            ->join('submissions as s', 'p.submission_id', '=', 's.submission_id')
+            ->get()
+            ->each(function ($row) use ($translations, $roles) {
+                $role = $roles->get($translations->get($row->user_group_id)['key']);
+                Repo::creditContributorRole()->addContributorRoles([$role], $row->author_id, $row->context_id);
```
bozana: Here as well, the direct query

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,246 @@
...
+        $role = ContributorRole::find((int) $illuminateRequest->route('roleId'));
+
+        if (!$role) {
+            return response()->json([
+                'error' => __('api.contributorRole.404.roleNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $contextId = $this->getRequest()->getContext()->getId();
+
```
bozana, line 141: In think we also need to check  ` if ($contextId !== $role->contextId) {` , no?

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,246 @@
...
+    public function getContributorRoleFormData(Request $illuminateRequest): JsonResponse
+    {
+        $context = $this->getRequest()->getContext();
+        // Available options only
+        $roleOptions = collect(ContributorRoleIdentifier::getRoles())
+            ->diff(Repo::contributorRole()
+                ->getSchemaMap()
+                ->summarizeMany(ContributorRole::query()->withContextId($context->getId())->get())
+                ->pluck('identifier')
+            )
```
bozana: It is unusual to use schema map and summarizeMany in such a case.
Why not use normal collection and functions?
E.g.
```
ContributorRole::query()->withContextId($context->getId())->get()->map(function ($role) {
            return $role->contributorRoleIdentifier;
        });
```

classes/author/contributorRole/ContributorRole.php
bozana: should here be checked if it is the last role and already in use? -- this is above all a note for me to think about it later...

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -702,7 +706,8 @@ public function add(Request $illuminateRequest): JsonResponse
         if ($submitAsUserGroup->roleId === Role::ROLE_ID_AUTHOR) {
             $author = Repo::author()->newAuthorFromUser($request->getUser(), $submission, $context);
             $author->setData('publicationId', $publication->getId());
-            $author->setUserGroupId($submitAsUserGroup->id);
+            $author->setData('contributorType', ContributorType::PERSON->getName());
+            $author->setContributorRoles([ContributorRoleIdentifier::AUTHOR->getName()], $context->getId());
```
bozana: Why do we not work with IDs, rather then with identifiers?
I think we should work with IDs, and only pass those, instead of always passing the context ID as well.
I think the form can also get IDs instead of identifiers.
And author object would contain only the array of ContributorRoles, instead of identifier => ContributorRole.
Except there is a really important reason I do not see to use identifiers I would like this to change....

classes/author/Author.php
```diff
@@ -287,19 +264,72 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
     public function setCreditRoles(?array $creditRoles): void
     {
         $this->setData('creditRoles', $creditRoles);
     }
+
+    /**
+     * Get contributor roles as assoc array [ identifier => translation(s) ].
+     * To get ContributorRole[] use getData
+     */
+    public function getContributorRoles(?string $locale = null): array
```
bozana: I think we need getLocalizedContributorRoles(?string $preferredLocale = null) that would return the localized names only i.e. also fall back if the preferredLocale is not found:
return collect($this->getData('contributorRoles'))
            ->map(fn (ContributorRole $role) => $role->getLocalizedData('name', $preferredLocale))
            ->toArray();

That getLocalizedContributorRoles function would be for example used in PKPPublicationLicenseForm and article_details.tpl.
Everywhere else, we could use getData('contributorRoles')
Also, we do not need to fill the empty names translations, I think. Always for display on the frontend: if there is no translation it should fall back to the default locale which is journal primary locale.

api/v1/contributorRoles/ContributorRoleController.php
bozana: EDIT: Forget about my previous comment, Jarda explained it to me :-) Sorry!

classes/services/PKPContextService.php
```diff
@@ -585,6 +588,13 @@ public function add($context, $request)
         Repo::emailTemplate()->dao->installAlternateEmailTemplates($context->getId());
         Repo::emailTemplate()->dao->setTemplateDefaultUnrestirctedSetting($context->getId());
 
+        $contributorRoleTranslations = Arr::map(
+            ['default.groups.name.author', 'default.groups.name.translator'],
+            fn (string $str): array => Arr::mapWithKeys($context->getData('supportedLocales'), fn (string $l) => [$l => __($str, locale: $l)])
```
bozana: suppoertedFormLocales

api/v1/contributorRoles/ContributorRoleController.php
jardakotesovec: Right and what about having endpoint that gives you all these available role identifiers. And on frontend you do the form specific thing, which is diffing of existing vs available ones?

Reason why I am looking for opportunities to make generally useful API is that ideally anyone can create UI like this just using the API.  So looking for balance to expose the data via API.. but not preprocessing data for the forms neither..

controllers/grid/settings/roles/form/UserGroupForm.php
```diff
@@ -135,6 +135,7 @@ public function initData()
                 'permitSettings' => $userGroup->permitSettings,
                 'recommendOnly' => $userGroup->recommendOnly,
                 'masthead' => $userGroup->masthead,
+                'hideShowTitle' => $userGroup->roleId === Role::ROLE_ID_AUTHOR,
```
bozana: We actually said that this option can be fully removed from the Roles form (Setting > Users & Roles > Roles > Edit a role). If it is not possible for you to do this within this PR maybe I can do it at the end -- when everything is pretty ready for merging...

schemas/contributorRole.json
```diff
@@ -0,0 +1,38 @@
...
+			"description": "The unique id of the role in the database.",
+			"readOnly": true,
+			"apiSummary": true,
+			"origin": "primary"
+		},
+		"identifier": {
+			"type": "string",
+			"description": "The identifier of the role.",
+			"apiSummary": true,
+			"origin": "primary"
```
bozana, line 39: should we list the possible values here?

api/v1/contributorRoles/ContributorRoleController.php
bozana: So I think Jarda means to have the API endpoint getContributorRoleIdentifiers() (returning all possible identifiers) and then we also have getMany() (returning all existing contributor roles) to get existing ContributorRoles. Then what exactly is available for the user to select in the form will be calculated in the frontend/ui-library component, correct @jardakotesovec ? -- The available for selection are all ContributorRoleIdentifiers that are not in existing ContributorRoles.

schemas/contributorRole.json
jyhein: Is this something you do, list all enum choices?

api/v1/contributorRoles/ContributorRoleController.php
jyhein: There can be both the available and all the identifiers in the api. Seems like to better have the list of available ones so that it doesn't need to be calculated every time in the UI.

classes/author/contributorRole/ContributorRole.php
jyhein: If you do such checking here

classes/author/contributorRole/Repository.php
jyhein: Just thinking here that how often UI locales are changed? Forcing to add all ui locales isn't much. It is, probably, done once. It is also a better UX for viewers to have ui elements in correct language than just always relying to fallbacks.

classes/author/creditContributorRole/Repository.php
jyhein: Under the if-check the query deletes roles based on the list, so if the list is empty, all roles are deleted, and nothing are added (below the delete). The comment is true but I add some more text to make it more clear

api/v1/submissions/PKPSubmissionController.php
jyhein: author-object holds array of ContributorRoles not identifier => ContributorRole. Function getContributorRoles returns an assoc array [ identifier => translation(s), ... ] because working with objects just adds extra code (get identifier and translatons). The objects are read-only, so it is easier to make the data more ready in the function. If those objects are really needed as objects, one can use getData(...).

Working with ids is that the ids need to be retrieved from the db first. Of course here is possible that a journal doesn't have an author, but how do you tell what they use as an author equivalent?

jyhein: Also set contributorRoles works with objects directly if needed

classes/author/Author.php
jyhein: The function functions as getLocalizedContributorRoles when locale is not null. As default it returns all the translations, if not needed to return all translations, I can change it to return just one translation

controllers/grid/settings/roles/form/UserGroupForm.php
jyhein: it was unclear, to me, what was the plan, so I left it as it is. The changes can be done in another issue if so.

classes/author/Author.php
bozana: Maybe then to call this function getLocalizedContributorRolesNames -- because it does not return contributor roles, but their localized names...

api/v1/submissions/PKPSubmissionController.php
bozana: Could you please create a new function in ContributoRole Repository to get a role (or roles) by identifier(s), use it here and use the setContributorRoles() without contextId parameter (and remove that parameter from the function)? -- It would be much easier to read and understand... because it is as our usual set functions...

classes/author/contributorRole/Repository.php
bozana: I cannot decide that -- what is better UX -- but we always do like I mentioned, so I do not know why would we now like to do it differently -- we cannot maintain all the special cases and all different implementations -- one needs to understand it easily (which means if we are used to have one logic or solution for something, it is more difficult if there are exceptions/special cases that we need to understand every time we change something -- especially if not needed) and we need to maintain so much code. And if we once decide to change/improve it, everything would be consistent and then easier to change. Also, this kind of decisions/changes -- how to deal with multilingualism, form and ui locales, fall back on default locales -- needs to be discussed and we need to agree on a solution we would then all know how to proceed further and start implementing it everywhere, but I think now it is not time to do that, at least not with this PR. We could start a GitHub discussion where you propose better handling of all of this, but for now it needs to be done in the usual way.

api/v1/submissions/PKPSubmissionController.php
jyhein: I was thinking that setContributorRoles would take identifiers and contextId only. When adding ContributorRole-objects, one would use setData('contributorRoles', ...). Then it is clear. I can still add the function to the repo where it does the query, so it doesn't need to be repeated. Then getData and setData would work directly with the objects

classes/author/Author.php
jyhein: thats's ok.

api/v1/submissions/PKPSubmissionController.php
bozana: Yes, it is just that it is not so clear because we still have getter and setter functions in the code that are used instead of setData() -- also providing better type hinting -- so one would always expect to set just the roles with this setContributorRoles function...

jyhein: I could rename it to e.g. setContributorRolesByIdentifiers (?) and make setContributorRoles just setData('...', ...) if setContributorRoles is needed anymore

classes/author/contributorRole/Repository.php
jyhein: That check for valid identifiers need to be there because there is no check for that elsewhere. Having identifiers listed in schema-json's validation seems not that good choice

api/v1/contributorRoles/ContributorRoleController.php
ewhanson: Sorry the the delay, but if this is still relevant, I agree with Jarda that it would be better to have a more generic data-fetching endpoint rather than a form-specific one, but if you've already settled on a working solution together, please ignore this!

classes/author/Author.php
```diff
@@ -288,18 +268,95 @@ public function getLocalizedAffiliationNamesAsString(?string $preferredLocale =
...
-     * Get contrubutor credit roles and degrees
+     * Get organization name.
+     */
+    public function getOrganizationName(?string $locale = null): string|array|null
+    {
+        return $this->getData('organizationName', $locale);
+    }
+
+    /**
+     * Get the localized given name
```
bozana: organization name

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,86 @@
...
+        ]);
+        $validator = ValidatorFactory::make($props, $rules);
+
+        ValidatorFactory::required(
+            $validator,
+            $role,
+            $this->schemaService->getRequiredProps($schema),
+            $this->schemaService->getMultilingualProps($schema),
+            $locales,
+            Locale::getPrimaryLocale()
```
bozana: I think it should be context->getPrimaryLocale()

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,86 @@
...
+        );
+        ValidatorFactory::allowedLocales($validator, $this->schemaService->getMultilingualProps(PKPSchemaService::SCHEMA_CONTRIBUTOR_ROLE), $locales);
```
bozana: $schema instead of PKPSchemaService::SCHEMA_CONTRIBUTOR_ROLE ?

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,86 @@
...
+
+        return app()->get('schema')->formatValidationErrors($validator->errors());
```
bozana: maybe to add:
 ```
       $errors = [];
        if ($validator->fails()) {
            $errors = $this->schemaService->formatValidationErrors($validator->errors());
        }
        Hook::call('ContributorRole::validate', [&$errors, $object, $props, $locales, $primaryLocale]);
        return $errors;
```
so that plugins can add hooks to this...

File: classes/author/contributorRole/maps/Schema.php [hunk 50]
```diff
@@ -0,0 +1,108 @@
...
+                case 'id':
+                    $output[$prop] = $item->contributor_role_id;
+                    break;
+                case 'identifier':
+                    $output[$prop] = $item->contributor_role_identifier;
+                    break;
+                case 'name':
+                    $output[$prop] = $item->name;
+                    break;
+            }
```
bozana, line 101: This is 'hard coded', so we should use:
```
                case '_href':
                    $output[$prop] = $this->getApiUrl(
                        'contributorRoles' . $item->getKey(),
                        $this->context->getData('urlPath')
                    );
                    break;
                case 'id':
                    $output[$prop] = $item->getKey();
                    break;
                default:
                    $output[$prop] = $item->getAttribute($prop);
                    break;
```
It will also catch contextId.
We also need to rename `identifier` in the contributorRole.json into `contributorRoleIdentifier`, so that that setting can be get with getAttributes().

schemas/contributorRole.json
bozana: I see, we do it somewhere :-)

schemas/contributorRole.json
```diff
@@ -0,0 +1,38 @@
+{
+	"title": "ContributorRole",
+	"description": "Contributor role for contributors",
+	"required": [
+		"contextId",
+		"identifier",
+		"name"
+	],
+	"properties": {
```
bozana, line 9: We can also add _href

"_href": {
			"type": "string",
			"origin": "composed",
			"description": "The URL to this contributor role in the REST API.",
			"format": "uri",
			"readOnly": true,
			"apiSummary": true
		},

classes/author/maps/Schema.php
```diff
@@ -96,17 +91,15 @@ protected function mapByProperties(array $props, Author $item): array
...
-                    $output[$prop] = $userGroup ? $userGroup->name : new stdClass();
-                    break;
                 case 'fullName':
                     $output[$prop] = $item->getFullName();
                     break;
                 case 'hasVerifiedOrcid':
                     $output[$prop] = $item->hasVerifiedOrcid();
                     break;
+                case 'contributorRoles':
+                    $output[$prop] = $item->getContributorRoleNames();
```
bozana: I think contributor roles should be provided via API as normal array and objects as they are, and as we usually do it (s. e.g. submission object containing publications):
`$output[$prop] = Repo::contributorRole()->getSchemaMap()->summarizeMany()->values();`
We would always like to provide the API endpoints and results so that they can be used by any other software, as they would like or need it too -- The API users should get the whole object, as it is, to get all the information and use it as they need.

classes/components/forms/publication/ContributorForm.php
```diff
@@ -68,37 +64,77 @@ public function __construct(string $action, array $locales, ?Submission $submiss
         usort($countries, function ($a, $b) {
             return strcmp($a['label'], $b['label']);
         });
+        array_unshift($countries, "");
 
+        $locale = Locale::getLocale();
```
bozana: no need for this, getLocalizedData() will consider it

classes/components/forms/publication/ContributorForm.php
```diff
@@ -68,37 +64,77 @@ public function __construct(string $action, array $locales, ?Submission $submiss
...
+        $contributorRoles = ContributorRole::query()
+            ->withContextId($context->getId())
+            ->get()
+            ->map(fn (ContributorRole $role): array => ['value' => $role->contributor_role_identifier, 'label' => $role->getLocalizedData('name', $locale)])
```
bozana: only getLocalizedData('name) is enough

classes/migration/install/SubmissionsMigration.php
```diff
@@ -102,17 +105,15 @@ public function up(): void
...
 
             // The foreign key relationship on this table is defined with the publications table.
             $table->bigInteger('publication_id');
 
             $table->float('seq')->default(0);
 
-            $table->bigInteger('user_group_id')->nullable();
-            $table->foreign('user_group_id')->references('user_group_id')->on('user_groups')->onDelete('cascade');
-            $table->index(['user_group_id'], 'authors_user_group_id');
+            $table->enum('contributor_type', ContributorType::getTypes())->nullable();
```
bozana: Can this be NULL?

classes/orcid/PKPOrcidWork.php
```diff
@@ -28,7 +29,7 @@
 abstract class PKPOrcidWork
 {
     public const PUBID_TO_ORCID_EXT_ID = ['doi' => 'doi', 'other::urn' => 'urn'];
-    public const USER_GROUP_TO_ORCID_ROLE = ['Author' => 'AUTHOR', 'Translator' => 'CHAIR_OR_TRANSLATOR', 'Journal manager' => 'AUTHOR'];
+    public const ORCID_MATCHING_CONTRIBUTOR_ROLES = ['AUTHOR', 'EDITOR'];
```
bozana: Maybe then to remove this constant and use it in the match($role) in the function getContributorRolesOrcid() -- it is easier to read/understand if everything is in one place, and this constant does not seem to be needed elsewhere

classes/publication/PKPPublication.php
```diff
@@ -208,40 +208,17 @@ public function getAuthorString(\Traversable $userGroups, $includeInBrowseOnly =
...
-        // If there needs to be a trailing user group title, add it
-        if (isset($author)) {
-            $lastUserGroup = $userGroupMap[$author->getUserGroupId()] ?? null;
-            if ($lastUserGroup && $lastUserGroup->showTitle) {
-                $str .= ' (' . $lastUserGroup->getLocalizedData('name') . ')';
-            }
-        }
-
-        return $str;
+        $locale = Locale::getLocale();
```
bozana: no need for this

classes/publication/PKPPublication.php
```diff
@@ -208,40 +208,17 @@ public function getAuthorString(\Traversable $userGroups, $includeInBrowseOnly =
...
+
+        return collect($authors)
+            ->map(fn (Author $author): string => 
+                $author->getFullName()
+                . " ("
+                . collect($author->getLocalizedContributorRoleNames($locale))
```
bozana: without $locale

classes/submission/Repository.php
```diff
@@ -368,6 +369,7 @@ public function validateSubmit(Submission $submission, Context $context): array
     {
         $locale = $submission->getData('locale');
         $publication = $submission->getCurrentPublication();
+        $submissionLocale = $submission->getData('locale');
```
bozana: this is the same as $locale above, so below $locale can be used everywhere (instead of $submissionLocale)

classes/userGroup/Repository.php
```diff
@@ -469,7 +469,7 @@ public function installSettings(?int $contextId, string $filename): bool
                 'permitSelfRegistration' => $permitSelfRegistration,
                 'permitMetadataEdit' => $permitMetadataEdit,
                 'isDefault' => true,
-                'showTitle' => true,
+                'showTitle' => $roleId !== Role::ROLE_ID_AUTHOR,
```
bozana: Why do we only consider author role here? Maybe to leave this as it is/was, it will be however removed in this new issue: https://github.com/pkp/pkp-lib/issues/11971

controllers/grid/settings/roles/form/UserGroupForm.php
bozana: Maybe then also not to change anything here...

locale/en/manager.po
```diff
@@ -3853,3 +3853,59 @@ msgstr "View Report"
...
+msgstr "Contributor Roles"
+
+msgid "manager.contributorRoles.identifier"
+msgstr "Role Identifier"
+
+msgid "manager.contributorRoles.name"
+msgstr "Role Name"
+
+msgid "manager.contributorRoles.name.description"
+msgstr "Fill name in all of the languages."
```
bozana, line 3867: This seems not to be used

locale/en/manager.po
```diff
@@ -3853,3 +3853,59 @@ msgstr "View Report"
...
+
+msgid "manager.contributorRoles.saved"
+msgstr "Contributor role saved"
+
+msgid "manager.contributorRoles.error.nameRequired"
+msgstr "Role names required in all of the languages."
```
bozana, line 3873: This seems not to be used

locale/en/manager.po
```diff
@@ -3853,3 +3853,59 @@ msgstr "View Report"
...
+
+msgid "manager.contributorRoles.error.delete.inUse"
+msgstr "One or more contributors are using this role. Change the role to another before delete."
+
+msgid "manager.contributorRoles.delete.role"
+msgstr "Delete Role"
+
+msgid "manager.contributorRoles.alert.delete.confirm"
+msgstr "I understand the consequences, delete this role"
```
bozana, line 3882: Is this used?

locale/en/submission.po
```diff
@@ -3277,3 +3277,26 @@ msgstr "Are you sure you want to stop automatically adding this task/discussion
 
 msgid "taskTemplates.confirmEmailTemplate"
 msgstr "Applying this email template will replace the discussion text in the form. The changes will not be saved unless you choose to save. Do you want to continue?"
+msgstr "Applying this template will replace data in related fields on the form. These changes won't be saved unless you choose to save. Continue?"
```
bozana: Is this maybe result of a rebase? It seems the first one is correct...

locale/en/user.po
```diff
@@ -889,3 +889,12 @@ msgstr "Bio Statement"
...
 msgstr "You cannot remove the role. At least one role must be assigned to the user."
+
+msgid "user.organizationName"
+msgstr "Organization Name"
+
+msgid "user.rorId"
+msgstr "ROR ID"
+
+msgid "user.rorId.description"
+msgstr "Enter organization's ROR ID."
```
bozana: I think all these can go into submission.po because they are specific for a contributor, and not used for the user

templates/controllers/grid/settings/roles/form/userGroupForm.tpl
```diff
@@ -54,7 +54,7 @@
 	<div id="userGroupOptionsContainer" class="full left">
 		{fbvFormArea id="userGroupOptions"}
 			{fbvFormSection title="settings.roles.roleOptions" list="true"}
-				{fbvElement type="checkbox" name="showTitle" id="showTitle" checked=$showTitle label="settings.roles.showTitles"}
+				{fbvElement type="{if $hideShowTitle}hidden{else}checkbox{/if}" name="showTitle" id="showTitle" checked=$showTitle label="settings.roles.showTitles"}
```
bozana: This can be left as it is/was

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,193 @@
...
+            $table->bigInteger('contributor_role_setting_id')->autoIncrement();
+            $table->bigInteger('contributor_role_id');
+            $table->string('setting_name', 255);
+            $table->string('setting_value', 255);
+            $table->string('locale', 28);
+            $table->unique(['contributor_role_id', 'setting_name', 'locale'], 'contributor_role_id_setting_name_locale_unique');
+            $table->foreign('contributor_role_id', 'contributor_role_id_settings_foreign')->references('contributor_role_id')->on('contributor_roles')->onDelete('cascade');
+        });
+
+        // Add contributor type column
```
bozana: the comment is not correct here?

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,291 @@
...
+                Role::ROLE_ID_MANAGER,
+            ]),
+        ])->group(function () {
+            Route::get('', $this->getMany(...));
+            Route::get('contributorRoleFormData', $this->getContributorRoleFormData(...));
+            Route::post('', $this->add(...));
+            Route::put('{roleId}', $this->edit(...))
+                ->whereNumber('roleId');
+            Route::delete('{roleId}', $this->delete(...))
+                ->whereNumber('roleId');
```
bozana, line 92: We should also add 
```
Route::get('{roleId}', $this->get(...))
  ->name('contributorRole.getContributorRole')
  ->whereNumber('roleId');
```
 To have consistent API endpoints, and also because we would like `_href` property to be everywhere, also in contributorRole.json.
 
 Also, the routes names are missing, like ` ->name('contributorRole.getContributorRole')` -- s. other API endpoints.

bozana (commented):
Hi Jyrki,
I went through the changes. There are still some things that need to be changed, but it seems not that much any more :-)
I have not reviewed the migration yet -- I wanted to speak with Alec about it -- it seems to me we should go different way... So I will let you know as soon as possible regarding this.
Thanks a lot for all your work!

classes/author/Repository.php
```diff
@@ -305,6 +303,8 @@ public function changePublicationLocale($publicationId, $oldLocale, $newLocale)
                     // copy only the given name with the old locale to the new locale, because the given name is required
                     $author->setGivenName($author->getGivenName($oldLocale), $newLocale);
                 }
+            } else if ($contributorType === ContributorType::ORGANIZATION->getName() && !$author->getOrganizationName($newLocale)) {
+                $author->setOrganizationName($author->ggetOrganizationName($oldLocale), $newLocale);
```
bozana: `getOrganizationName()` instead of `ggetOrganizationName()`

schemas/contributorRole.json
bozana: Yes, please list them here -- they should be displayed here as well, than the validation happen automatically, so maybe there is then no need to have that extra rule in the validate function in the Repository.php

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,291 @@
...
+ *
+ * @class ContributorRoleController
+ *
+ * @brief Handle API requests for contributor role operations.
+ *
+ */
+
+namespace PKP\API\v1\contributorRoles;
+
+use APP\core\Application;
```
bozana: can be removed, it seems it is not used

api/v1/contributorRoles/ContributorRoleController.php
```diff
@@ -0,0 +1,291 @@
...
+use APP\facades\Repo;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Arr;
```
bozana: not used

classes/author/contributorRole/maps/Schema.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @class Schema
+ *
+ * @brief Map contributor roles to the properties defined in the contributorRole schema
+ */
+
+namespace PKP\author\contributorRole\maps;
+
+use APP\facades\Repo;
+use APP\submission\Submission;
```
bozana: These two seem not to be used

classes/author/creditContributorRole/CreditContributorRole.php
```diff
@@ -15,10 +15,12 @@
 namespace PKP\author\creditContributorRole;
 
 use Eloquence\Behaviours\HasCamelCasing;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Casts\Attribute;
-use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
```
bozana: not used

classes/author/creditContributorRole/CreditContributorRole.php
```diff
@@ -15,10 +15,12 @@
...
 use Illuminate\Database\Query\JoinClause;
+use PKP\author\contributorRole\ContributorRole;
```
bozana: not used.

classes/author/creditContributorRole/Repository.php
```diff
@@ -14,46 +14,78 @@
 
 namespace PKP\author\creditContributorRole;
 
+use APP\core\Application;
+use Illuminate\Database\Eloquent\Builder;
```
bozana: not used

classes/author/maps/Schema.php
```diff
@@ -21,7 +21,6 @@
 use PKP\core\PKPRequest;
 use PKP\security\Role;
```
bozana: not used any more

classes/components/forms/publication/ContributorForm.php
```diff
@@ -26,9 +28,10 @@
 use PKP\components\forms\FieldText;
 use PKP\components\forms\FormComponent;
 use PKP\context\Context;
+use PKP\facades\Locale;
+use PKP\facades\Repo;
```
bozana: not used

classes/components/forms/publication/ContributorForm.php
```diff
@@ -26,9 +28,10 @@
...
 use PKP\orcid\OrcidManager;
 use PKP\security\Role;
```
bozana: not used any more

templates/submission/review-contributors.tpl
```diff
@@ -63,7 +63,12 @@
                     >
                         {translate key="author.users.contributor.principalContact"}
                     </badge>
-                    <badge>{{ localize(author.userGroupName) }}</badge>
+                    <badge 
+                        v-for="(role, identifier) in author.contributorRoles"
```
bozana: once you change the author schema map, to have full objects, this will need to be changed too

classes/userGroup/Repository.php
jyhein: yes

locale/en/manager.po
jyhein: This and below not found are used vue-side form

locale/en/submission.po
```diff
@@ -3304,3 +3304,34 @@ msgstr "Are you sure you want to stop automatically adding this task/discussion
 
 msgid "taskTemplates.confirmEmailTemplate"
 msgstr "Applying this email template will replace the discussion text in the form. The changes will not be saved unless you choose to save. Do you want to continue?"
+
+msgid "user.organizationName"
```
bozana: could we please change it to submission.contributor.organizationName or submission.submit.contributor.organizationName
here and for user.rorId and user.rorId.description

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,174 @@
...
+            $table->unique(['contributor_role_id', 'setting_name', 'locale'], 'contributor_role_id_setting_name_locale_unique');
+            $table->foreign('contributor_role_id', 'contributor_role_id_settings_foreign')->references('contributor_role_id')->on('contributor_roles')->onDelete('cascade');
+        });
+
+        // Add foreign key
+        Schema::table('credit_contributor_roles', function (Blueprint $table) {
+            $table->foreign('contributor_role_id', 'contributor_role_id_foreign')->references('contributor_role_id')->on('contributor_roles')->onDelete('cascade');
+        });
+
+        // Add author user group if not exist
```
bozana: No need to do that -- I have just tested: the default user groups cannot be deleted.

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,174 @@
...
+                'userGroupId' => $ugId,
+                'key' => $ug->nameLocaleKey,
+                'name' => $ug->name,
+                'contextId' => $ug->context_id,
+            ]);
+        $roles = collect([
+            'default.groups.name.author' => ContributorRoleIdentifier::AUTHOR->getName(),
+            'default.groups.name.translator' => ContributorRoleIdentifier::TRANSLATOR->getName(),
+            'default.groups.name.chapterAuthor' => ContributorRoleIdentifier::AUTHOR->getName(),
+            'default.groups.name.volumeEditor' => ContributorRoleIdentifier::EDITOR->getName(),
```
bozana, line 75: We need to also consider the roles that has been added by journal manager and that would not have the nameLocaleKey. I think we can map them to AUTHOR identifier.

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,174 @@
...
+                // Save the new author instance to the database
+                $userGroup->save();
+            });
+
+
+        $translations = UserGroup::withRoleIds([Role::ROLE_ID_AUTHOR])
+            ->get()
+            ->map(fn (UserGroup $ug, int $ugId) => [
+                'userGroupId' => $ugId,
+                'key' => $ug->nameLocaleKey,
```
bozana: Just a note, that nameLocaleKey does not need to exists, e.g. for the roles that were additionally created by journal manager

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,174 @@
...
+                            ->insertGetId(['contributor_role_identifier' => $identifier, 'context_id' => $contextId], 'contributor_role_id');
+                        // Insert role settings
+                        DB::table('contributor_role_settings')
+                            ->insert(collect($t['name'])
+                                ->map(fn (string $value, string $locale): array => ['contributor_role_id' => $roleId, 'setting_name' => 'name', 'setting_value' => $value, 'locale' => $locale])
+                                ->values()
+                                ->toArray()
+                            );
+
+                        return [$identifier => $roleId];
```
bozana: I think we also need to keep the user_group_id -- to consider the additionally created roles by journal manager, so maybe user_group_id => roleId

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,174 @@
...
+            ->select(['a.author_id', 'a.user_group_id', 's.context_id'])
+            ->join('publications as p', 'a.publication_id', '=', 'p.publication_id')
+            ->join('submissions as s', 'p.submission_id', '=', 's.submission_id')
+            ->get()
+            ->each(function ($row) use ($translations, $roles, $roleIds) {
+                $ug = $translations->get($row->user_group_id);
+                $contextId = $ug['contextId'];
+                $identifier = $roles->get($ug['key']);
+                DB::table('credit_contributor_roles')
+                    ->insert(['contributor_id' => $row->author_id, 'contributor_role_id' => $roleIds->get($contextId)->get($identifier)]);
```
bozana: here we would need that mapping to user_group_id for the additionally roles created by journal manager

plugins/importexport/native/filter/PKPAuthorNativeXmlFilter.php
```diff
@@ -128,12 +129,19 @@ public function createPKPAuthorNode($doc, $author)
         }
 
         $this->createOptionalNode($doc, $authorNode, 'country', $author->getCountry());
-        $authorNode->appendChild($doc->createElementNS($deployment->getNamespace(), 'email', htmlspecialchars($author->getEmail(), ENT_COMPAT, 'UTF-8')));
+        $this->createOptionalNode($doc, $authorNode, 'email', htmlspecialchars($author->getEmail() ?? '') ?: null);
```
bozana, line 132: if email does not exists
htmlspecialchars($author->getEmail() ?? '') would return empty string
so the last parameter for createOptionalNode is never null.

Maybe to have it something like:
$author->getEmail() ? htmlspecialchars($author->getEmail() : null

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,145 @@
...
+                $translations
+                    ->filter(fn (array $t) => $t['contextId'] === $contextId)
+                    ->mapWithKeys(function (array $t, int $ugId) use ($roles, &$roleIds, $contextId): array {
+                        $identifier = $roles[$t['key']];
+                        // Insert role. Get id for later use
+                        $roleId = DB::table('contributor_roles')
+                            ->insertGetId(['contributor_role_identifier' => $identifier, 'context_id' => $contextId], 'contributor_role_id');
+                        // Insert role settings
+                        DB::table('contributor_role_settings')
+                            ->insert(collect($t['name'])
```
bozana: Maybe to here construct an array and call insert with the whole array -- I think it is better for the performance.

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,145 @@
...
+        // Add contributor roles and types to existing contributors
+        DB::table('authors as a')
+            ->select(['a.author_id', 'a.user_group_id', 's.context_id'])
+            ->join('publications as p', 'a.publication_id', '=', 'p.publication_id')
+            ->join('submissions as s', 'p.submission_id', '=', 's.submission_id')
+            ->get()
+            ->each(function ($row) use ($translations, $roles, $roleIds) {
+                $contextId = $translations->get($row->user_group_id)['contextId'];
+                DB::table('credit_contributor_roles')
+                    ->insert(['contributor_id' => $row->author_id, 'contributor_role_id' => $roleIds->get($contextId)->get($row->user_group_id)]);
```
bozana: Also here, maybe to here construct an array and call insert with the whole array -- I think it is better for the performance -- not to do an insert for every author, it could be so many...

bozana (commented):
Hi @jyhein,
great, I left just a few comments.
I would then test it all, will also take a look what is with the tests, but I believe you can proceed with OMP and OPS

plugins/importexport/native/filter/PKPAuthorNativeXmlFilter.php
jyhein: I tested it with anonymous (no email) it works. empty string is false so it is null in: '' ?: null

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
jyhein: What do you mean?

jyhein: Do you mean outside of the foreach loops?

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,156 @@
...
+
+        Schema::table('authors', function (Blueprint $table) {
+            // Change email to nullable
+            $table->string('email', 90)->nullable()->change();
+            // Remove column 'user_group_id'
+            $table->dropForeign('authors_user_group_id_foreign');
+            $table->dropIndex('authors_user_group_id');
+            $table->dropColumn('user_group_id');
+            // Add the enum column for contributor types
+            $table->enum('contributor_type', ContributorType::getTypes());
```
bozana: I am testing it all right now with PostgreSQL and this line leads to:
`Not null violation: 7 ERROR:  column "contributor_type" of relation "authors" contains null values (Connection: pgsql, SQL: alter table "authors" add column "contributor_type" varchar(255) check ("contributor_type" in ('PERSON', 'ORGANIZATION', 'ANONYMOUS')) not null)`
Thus, maybe to have:
`$table->enum('contributor_type', ContributorType::getTypes())->default($person);`
and then `use($person)` is also needed above:
`Schema::table('authors', function (Blueprint $table) use ($person) {
`
Then we do not need the update below?

jyhein: It worked with mysql

jyhein: I can run the tests now again to see if it's ok

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,79 @@
...
+    public function getSchemaMap(): maps\Schema
+    {
+        return app('maps')->withExtensions($this->schemaMap);
+    }
+
+    public function validate(?ContributorRole $role, array $props, Context $context): array
+    {
+        $locales = $context->getSupportedFormLocales();
+        $primaryLocale = $context->getPrimaryLocale();
+        $identifiers = ContributorRoleIdentifier::getRoles();
```
bozana: This variable seems not to be used any more, so it can be removed...

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,79 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Repository
+ *
+ * @brief A repository to manage actions related to contributor roles
+ */
+
+namespace PKP\author\contributorRole;
+
+use APP\core\Application;
```
bozana: it is not used any more

classes/author/contributorRole/Repository.php
```diff
@@ -0,0 +1,79 @@
...
+use APP\core\Request;
+use Illuminate\Validation\Rule;
```
bozana: it is not used any more

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @class I857_ContributorRolesTypes
+ *
+ * @brief create the contributor roles table
+ */
+
+namespace PKP\migration\upgrade\v3_6_0;
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Arr;
```
bozana: not used any more

classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
```diff
@@ -0,0 +1,152 @@
...
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+use PKP\author\contributorRole\ContributorRoleIdentifier;
+use PKP\author\contributorRole\ContributorType;
+use PKP\facades\Repo;
```
bozana: not used any more

PR #11767, opened 2025-09-02
Prevent password reuse when user is prompted to change password

wopsononock:
fixes issue #11768


wopsononock:
This never allows a reset to succeed. Closing. Copying ChangePasswordForm's implementation would be better anyway. Will resubmit.

PR #11778, opened 2025-09-03
pkp/pkp-lib#10962 Move migration to pkp-lib for reuse across application

Hafsa-Naeem:


classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,61 @@
...
+
+use Illuminate\Support\Facades\DB;
+use PKP\install\DowngradeNotSupportedException;
+use PKP\migration\Migration;
+
+class I10962_UpdateEmailTemplateVariables extends Migration
+{
+    public function up(): void
+    {
+        DB::transaction(function () {
```
touhidurabir: lets not use DB transaction until we reach to resolution on https://github.com/pkp/pkp-lib/discussions/10595 . I personally love to have this approach but as we haven't started to use this consistently .

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,61 @@
...
+            $map = [
+                'journalAcronym' => 'contextAcronym',
+                'journalName' => 'contextName',
+                'journalUrl' => 'contextUrl',
+                'journalSignature' => 'contextSignature',
+            ];
```
touhidurabir: If I understand correctly, this is only considering the case of `OJS` but not `OMP/OPS` where we have like `pressName` or `serverName` , e.g specific to OMP/OPS . I think this is good but this core/shared lib class should have only the core migration define and abstract method for variable mapping so that each app can have it's own variable modifications . As by updating the `classes/migration/upgrade/v3_4_0/InstallEmailTemplates.php`, it will have effect when updating form 3.3.4-x but for installation upgrading form 3.5.0-x (already upgraded to 3.5.0), need to accommodate that also . Please correct me if I am mistaken .

touhidurabir (changes requested):
@Hafsa-Naeem I have added few comments . lets port these for OMP/OPS and we can finalise the PR to merge after one final review .

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,76 @@
...
+            'pressName'        => 'contextName',
+            'pressUrl'         => 'contextUrl',
+            'pressSignature'   => 'contextSignature',
+
+            // OPS
+            'serverAcronym'    => 'contextAcronym',
+            'serverName'       => 'contextName',
+            'serverUrl'        => 'contextUrl',
+            'serverSignature'  => 'contextSignature',
+        ];
```
touhidurabir: I was thinking about having these variable mapping to each of their corresponding apps but this works also . It just that the below loop is going to run 3 times more than it needed to be and have unnecessary DB query .

Hafsa-Naeem:
@touhidurabir Thanks for the approvals. The lone failure is in `SubmissionWizard.cy.js`, timing out on `h1: "Submission complete"`. All prior requests are 200 and the same flow reaches “Submission complete” earlier, so this looks like a timing-related flake and unrelated to this PR (it only changes an email-template migration in pkp-lib).

I'm running the upgrade checks now:

3.3.0-x -> 3.5.0-x , 3.4.0-x -> 3.5.0-x, 3.5.0-0 -> 3.5.0-x 

Once these are green, I'll forward-port to OMP and OPS.

Issue #11772, opened 2025-09-03
ReviewReminder diffInDays Parameter Order Bug in OJS 3.5.0.1


dmcconeghy:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The ReviewReminder scheduled task in OJS 3.5.0.1 has incorrect parameter order in diffInDays() calls for "after due date" reminders, causing them to never be sent.
/lib/pkp/classes/task/ReviewReminder.php

Line 112: $currentDate->diffInDays($dateResponseDue)
Line 141: $currentDate->diffInDays($dateDue)

Problem Description
When calculating days between dates for "after due date" reminders, the parameter order is reversed. This causes diffInDays() to return negative numbers when current date is after the due date, which never satisfies the >= numDaysAfter* conditions.
Example
For a review due 2025-02-03 and current date 2025-09-02:

Buggy code: $currentDate->diffInDays($dateDue) returns -211
Condition: -211 >= 5 evaluates to false (no reminder sent)
Fixed code: $dateDue->diffInDays($currentDate) returns 211
Condition: 211 >= 5 evaluates to true (reminder sent)

### Steps to Reproduce

The bug just is? Our reviewers weren't being sent review reminders on the schedule their journal settings should have allowed. When the code that managed the time calculation was fixed (swapped), the math checks out (no longer negative) and can dutifully send out the notice. 

### Expected Result

We should expect reviewers to continue receiving overdue notices following journal rules. 

### Actual Result

None were being sent/received whose date were caught in this window. "Reviews 60-211 days overdue that previously received no reminders now properly trigger reminder jobs"

### Application Version

OJS3.5.0.1

### Logs

No errors in log produced by this. 

### Additional Information

To debug missing emails I was comparing overdue reviews against our email logs. 
`cd /var/www/ojs && sudo -u www-data php -r "
require_once 'lib/pkp/lib/vendor/autoload.php';
use Carbon\Carbon;
\$currentDate = Carbon::today();
\$dateDue = Carbon::parse('2025-02-03');
\$diff = \$currentDate->diffInDays(\$dateDue);
echo 'Current: ' . \$currentDate . '\n';
echo 'Due: ' . \$dateDue . '\n'; 
echo 'Diff: ' . \$diff . '\n';
echo 'Current > Due: ' . (\$currentDate->gt(\$dateDue) ? 'true' : 'false') . '\n';
"`

produced 
[...]

PR #11780, opened 2025-09-04
pkp/pkp-lib#11772 fixed carbon 3.x diffIn* method negative value issue

touhidurabir:
for pkp/pkp-lib#11772 .

### DIFF
--- a/classes/task/ReviewReminder.php
+++ b/classes/task/ReviewReminder.php
@@ -54,7 +54,7 @@ class ReviewReminder extends ScheduledTask
         foreach ($incompleteAssignments as $reviewAssignment) {
             
             // Fetch the submission
-            if ($submission == null || $submission->getId() != $reviewAssignment->getSubmissionId()) {
+            if ($submission == null || $submission?->getId() != $reviewAssignment->getSubmissionId()) {
                 unset($submission);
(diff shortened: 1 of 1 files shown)


taslangraham (approved):

classes/task/ReviewReminder.php
```diff
@@ -95,7 +95,7 @@ public function executeActions(): bool
                     // need to check should we sent a BEFORE REVIEW REQUEST RESPONSE reminder
                     if ($numDaysBeforeReviewResponseReminderDue > 0 &&
                         $dateResponseDue->gt($currentDate) &&
-                        $dateResponseDue->diffInDays($currentDate) <= $numDaysBeforeReviewResponseReminderDue) {
+                        (int)abs($dateResponseDue->diffInDays($currentDate)) <= $numDaysBeforeReviewResponseReminderDue) {
```
bozana, line 98: Hi @touhidurabir, why do we use abs() ? 
Can the difference be a negative number?
We also check $dateResponseDue->gt($currentDate), so it should not be a negative number, or?

bozana:
All well @touhidurabir. First I was thinking wrong, and added some comments, but then I realized that I am wrong, so I deleted them... :-)
Thanks!

Issue #11754, opened 2025-08-27
Data structure for edit_task_templates and associated user groups


Hafsa-Naeem:
**Description:**
Creation and extension of the data structures required for `edit_task_templates`, specifically to support associating user groups and default roles for automated task assignment based on workflow stages.

**What application are you using?**
  OJS, OMP or OPS main

Hafsa-Naeem:
Ready for review @Vitaliy-1 
pkp-lib: [pkp-pkplib#11754 Added data structure for edit_task_templates and associated user groups #11787](https://github.com/pkp/pkp-lib/pull/11787)
ojs_main: [pkp/pkp-lib#11754 Submodule update ##Hafsa-Naeem/i11754-main-fix #5082](https://github.com/pkp/ojs/pull/5082)

Hafsa-Naeem:
Ready for review @Vitaliy-1 
pkp-lib: [pkp-pkplib#11754 Added data structure for edit_task_templates and associated user groups #11787](https://github.com/pkp/pkp-lib/pull/11787)
ojs_main: [pkp/pkp-lib#11754 Submodule update ##Hafsa-Naeem/i11754-main-fix #5082](https://github.com/pkp/ojs/pull/5082)

Vitaliy-1:
Thanks! Looks good and can be merged

PR #11787, opened 2025-09-05
pkp-pkplib#11754 Added data structure for edit_task_templates and associated user groups

Hafsa-Naeem:
for #11754

### DIFF
--- a/classes/migration/install/SubmissionsMigration.php
+++ b/classes/migration/install/SubmissionsMigration.php
@@ -306,11 +306,11 @@ class SubmissionsMigration extends \PKP\migration\Migration
             // templates are journal/context scoped
             $table->bigInteger('context_id')->comment('Journal/press ID for scoping templates');
             $contextDao = Application::getContextDAO();
-            $table->foreign('context_id', 'edit_task_templates_context_id')
-                ->references('context_id')
+            $table->foreign('context_id', 'edit_task_templates_context_fk')
+                ->references($contextDao->primaryKeyColumn)
                 ->on($contextDao->tableName)
                 ->onDelete('cascade');
-            $table->index(['context_id'], 'edit_task_templates_context_id');
+            $table->index(['context_id'], 'edit_task_templates_context_id_idx');
 
             $table->boolean('include')->default(false);
             $table->bigInteger('email_template_id')->nullable();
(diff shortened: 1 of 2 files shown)


classes/migration/install/SubmissionsMigration.php
```diff
@@ -257,9 +257,11 @@ public function up(): void
             $table->bigInteger('created_by')->nullable()->default(null);
             $table->foreign('created_by')->references('user_id')->on('users');
             $table->unsignedSmallInteger('type')->default(1); // 1 - discussion, 2 - task
-            $table->dateTime('date_started')->nullable();
-            $table->dateTime('date_closed')->nullable();
-            $table->string('title')->nullable();
+            $table->unsignedSmallInteger('status')->default(1); // record about the last activity, default EditorialTask:STATUS_NEW
```
Vitaliy-1: This should be removed

classes/migration/install/SubmissionsMigration.php
```diff
@@ -296,6 +298,18 @@ public function up(): void
             $table->comment('Represents templates for the editorial tasks.');
             $table->unsignedBigInteger('edit_task_template_id')->autoIncrement()->primary();
             $table->unsignedSmallInteger('stage_id');
+            $table->string('title', 255); // template title
+            /*
+            // Pending confirmation from devika to scope templates per context
+
+            $table->bigInteger('context_id')->comment('Journal/press ID for scoping templates');
+            $contextDao = \APP\core\Application::getContextDAO();
```
Vitaliy-1: instead of the call to the DAO, let's just hardcode the column name

classes/migration/upgrade/v3_6_0/I10406_EditorialTasks.php
Vitaliy-1, line 1: You'll need to create a new migration, rather than re-use old

classes/migration/upgrade/v3_6_0/I10406_EditorialTasks.php
```diff
@@ -110,6 +117,7 @@ public function up(): void
             $table->foreign('email_template_id')
                 ->references('email_id')
                 ->on('email_templates');
+            $table->string('title', 255);
```
Vitaliy-1: You can remove this, the title is already added into the edit_task table. In the next migration I'll remove it from the notes table, as a title previously was recorded in the headnote

classes/migration/upgrade/v3_6_0/I11754_TaskTemplateEnhancements.php
```diff
@@ -0,0 +1,68 @@
...
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use PKP\migration\Migration;
+
+class I11754_TaskTemplateEnhancements extends Migration
+{
+    public function up(): void
+    {
+        if (!Schema::hasColumn('edit_tasks', 'edit_task_template_id')) {
```
Vitaliy-1: I think we can remove such checks

Vitaliy-1 (changes requested):
Thanks, I've left a couple of comments

Issue #11791, opened 2025-09-06
User management performs poorly with large databases


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

In OJS/OMP/OPS 3.5.0-1, the user list in `Settings` > `Users & Roles` performs poorly. Almost all of the execution time in generating an API response go to the [`canLoginAs` and `canMergeUsers` properties](https://github.com/pkp/pkp-lib/blob/3_5_0-1/classes/user/maps/Schema.php#L131..L136), which use [very slow individual queries](https://github.com/pkp/pkp-lib/blob/3_5_0-1/classes/user/maps/Schema.php#L289..L326) to check the status.

There was a [SQL formulation drafted for this that performed well](https://github.com/pkp/pkp-lib/issues/10290#issuecomment-2335003555), but it was never used.

### Steps to Reproduce

1. With a large database, navigate to `Settings` > `Users & Roles`.

### Expected Result

The page should perform acceptably.

### Actual Result

The page performs poorly.

### Application Version

OJS 3.5.0-1

Hafsa-Naeem:
@asmecher  Ready for review
pkp-lib (stable-3_5_0): [pkp/pkp-lib#11791 Improve performance of user list by batching permission checks #11794](https://github.com/pkp/pkp-lib/pull/11794)
 OJS (stable-3_5_0):  [pkp/pkp-lib#11791 Submodule update ##Hafsa-Naeem/i11791-stable-3_5_0-fix #5083](https://github.com/pkp/ojs/pull/5083)

asmecher:
@Hafsa-Naeem, I've added a review [here](https://github.com/pkp/pkp-lib/pull/11794#pullrequestreview-3197704817), thanks!

asmecher:
@Hafsa-Naeem, is this ready for another review? If so, please don't forget to flag it here on the issue entry!

Hafsa-Naeem:
@asmecher I've addressed the suggested changes but fixing test. Once it's done i'll flag.

Hafsa-Naeem:
@asmecher Ready for review. 
pkp-lib (stable-3_5_0): [pkp/pkp-lib#11791 Improve performance of user list by batching permission checks #11794](https://github.com/pkp/pkp-lib/pull/11794)
[...]

PR #11794, opened 2025-09-08
pkp/pkp-lib#11791 Improve performance of user list by batching permission checks

Hafsa-Naeem:
for #11791


classes/user/maps/Schema.php
```diff
@@ -97,11 +96,107 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
-            return $this->summarizeReviewer($item);
-        });
+        $this->loadUserContextPermissions($collection); // one batched query per page
+        return $collection->map($perItem);
+    }
+
+    // compute the page wide coverage map
+    private function loadUserContextPermissions(Enumerable $collection): void
+    {
+        $request = Application::get()->getRequest();
```
asmecher: I think the `$auxiliaryData` pattern already exists in this class to provide information like this, which would avoid us having to create another entanglement with `Application::getRequest` from the storage layer, which is a pattern I'd like to avoid. Better for calling code to send in the current user via `$auxiliaryData`. Then we can also avoid the `Schema::$currentUserId` property; it's a little unclear when that can be expected to contain a value.

classes/user/maps/Schema.php
```diff
@@ -97,11 +96,107 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
+        }
+
+        // compute results for this page
+        $this->buildUserContextManagementMap($this->currentUserId, $ids);
+    }
+
+    // ensure cache has entries for ad hoc lookups
+    private function loadUserContextMap(array $targetIds): void
+    {
+        $targetIds = array_values(array_unique(array_map('intval', $targetIds)));
```
asmecher: (It's better to use first-class callable syntax: `intval(...)` rather than `'intval'`)

classes/user/maps/Schema.php
```diff
@@ -97,11 +96,107 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
+                        JOIN user_user_groups uug_m ON ug_m.user_group_id = uug_m.user_group_id
+                        WHERE ug_m.role_id = ?
+                          AND uug_m.user_id = ?
+                          AND ug_m.context_id = ug_t.context_id
+                      )
+                )
+                THEN 0 ELSE 1 END AS manages_all',
+                [ (int) Role::ROLE_ID_MANAGER, (int) $managerId ]
+            )
+            ->whereIn('u.user_id', array_map('intval', $targetIds))
```
asmecher: As above, use first-class callable syntax.

classes/user/maps/Schema.php
```diff
@@ -97,11 +96,107 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
+        $missing = array_values(array_diff($targetIds, $known));
+        if (!empty($missing)) {
+            $this->buildUserContextManagementMap($this->currentUserId, $missing);
+        }
+    }
+
+    /**
+     * extend $this->managesAllMap for the given manager and target ids
+     */
+    private function buildUserContextManagementMap(int $managerId, array $targetIds): void
```
asmecher: `$targetIds` is not clear; call it e.g. `$managedUserIds` instead.

classes/user/maps/Schema.php
```diff
@@ -97,11 +96,107 @@ public function summarizeMany(Enumerable $collection): Enumerable
...
+                      )
+                )
+                THEN 0 ELSE 1 END AS manages_all',
+                [ (int) Role::ROLE_ID_MANAGER, (int) $managerId ]
+            )
+            ->whereIn('u.user_id', array_map('intval', $targetIds))
+            ->get();
+
+        foreach ($rows as $r) {
+            $this->managesAllMap[(int) $r->user_id] = ((int) $r->manages_all) === 1;
```
asmecher: Don't use an exact `===` test; some DBMS configurations return numeric types as strings.

classes/user/maps/Schema.php
```diff
@@ -28,13 +28,18 @@
...
+use Illuminate\Support\Facades\DB;
 
 class Schema extends \PKP\core\maps\Schema
 {
     public Enumerable $collection;
 
     public string $schema = PKPSchemaService::SCHEMA_USER;
 
+    // page scoped cache to check if current user manage all contexts of target user
+    private array $managesAllMap = [];
```
asmecher: Rather than using a class property that can be set in some (but not all) cases, and which might get left around and accidentally affect re-use of the `Schema` object, use a local variable in cases where this is useful. (In other circumstances we use properties, but I think those are appropriate because the values are not as arbitrary as they are here -- where they are only relevant to a single user ID.)

classes/user/maps/Schema.php
```diff
@@ -72,36 +81,112 @@ public function summarizeReviewer(User $item, array $auxiliaryData = []): array
...
+    {
+        if (empty($managedUserIds)) {
+            return [];
+        }
+
+        $rows = DB::table('users as u')
+            ->select('u.user_id')
+            ->selectRaw(
+                'CASE WHEN EXISTS (
+                    SELECT 1
```
Vitaliy-1: Is it correct? At the first glance it should select something like `context_id` column

Vitaliy-1: Also, can we use a Query Builder abstraction here? Either by simplifying the query or, at least, build a subquery separately:
```php
$subquery = DB::table('user_user_groups')
  ->select(...)
  ->where(...)
```
And then inject it into the query?

classes/user/Repository.php
```diff
@@ -446,4 +449,170 @@ public function getAdminUsers(): LazyCollection
...
+                'ug.role_id as roleId',
+                'ug.show_title as showTitle',
+                'ug.permit_self_registration as permitSelfRegistration',
+                'ug.permit_metadata_edit as permitMetadataEdit',
+                'user_user_groups.date_start as dateStart',
+                'user_user_groups.date_end as dateEnd',
+                'user_user_groups.masthead as userMasthead',
+                'ugs_name.setting_value as name',
+                'ugs_abbrev.setting_value as abbrev',
+            ]);
```
asmecher: This is a lot of code that should already be captured in `UserGroup` / `UserUserGroup`; can you work with those classes to get the list of user groups rather than going direct to the database here?

Generally speaking, interactions with a table should be implemented only in that table's DAO class, except where joins are required to e.g filter in complex situations; in those cases, when not using Eloquent, the Collector should take care of those concerns. When using Eloquent, Eloquent relationships should be used.

classes/user/Repository.php
```diff
@@ -446,4 +449,170 @@ public function getAdminUsers(): LazyCollection
...
+        }
+        return $map;
+    }
+
+    /**
+     * Preload interests for a set of users (ids only).
+     */
+    public function preloadInterests(array $userIds): array
+    {
+        $rows = DB::table('user_interests as ui')
```
asmecher: This should already be encapsulated in `lib/pkp/classes/user/interest/UserInterest.php`.

classes/user/DAO.php
```diff
@@ -291,6 +310,37 @@ public function delete(User $user)
         parent::_delete($user);
     }
 
+    /**
+     * Build a lightweight user from a db row without loading settings.
+     */
+    protected function makeUserWithoutSettings(object $row, bool $includeReviewerData = false): User
```
asmecher: This is a lot of code bifurcation to avoid multiple fetches of user settings. It should be possible to add broad support for this in a really lightweight way using something like this:

https://github.com/asmecher/pkp-lib/commit/fce0e1f90ee66f47c7450dc14e03d0aa3b864bfd#diff-5efc8e0460497e833e2cf8292469195d4a4a37ff5f99f89203adff64452922b1

But I don't think this is a performance bottleneck, so I don't think we should include it in `stable-3_5_0`.

classes/user/DAO.php
```diff
@@ -125,10 +125,14 @@ public function exists(int $id): bool
      */
     public function getMany(Collector $query): LazyCollection
     {
+        DB::listen(function ($query) {
```
asmecher: Debug code left in PR.

classes/user/DAO.php
asmecher: @Hafsa-Naeem, this is no longer used and should be removed.

classes/user/Repository.php
```diff
@@ -446,4 +450,162 @@ public function getAdminUsers(): LazyCollection
     {
         return $this->dao->getAdminUsers();
     }
+
+    /**
+     * build a permission map for a manager over a set of users
+     */
+    public function permissionMapForManager(int $managerUserId, array $userIds): array
```
asmecher, line 457: @Hafsa-Naeem, I don't think some of these functions are used now that the PR has been simplified.

Issue #11779, opened 2025-09-03
Add API endpoint to create task templates


Hafsa-Naeem:
**Description:**
Add POST `/api/v1/editTaskTemplates` endpoint to allow creation of task templates.

**What application are you using?**
OJS, OMP and OPS main branch

Hafsa-Naeem:
@Vitaliy-1 Ready for review:
pkp-lib: [pkp/pkp-lib#11779 added API endpoint to create task templates #11807]( https://github.com/pkp/pkp-lib/pull/11807)
ojs: [pkp/pkp-lib#11779 Submodule update ##Hafsa-Naeem/i11779-main-fix #5091](https://github.com/pkp/ojs/pull/5091)

Hafsa-Naeem:
@Vitaliy-1 Ready for review:
pkp-lib: [pkp/pkp-lib#11779 added API endpoint to create task templates #11807]( https://github.com/pkp/pkp-lib/pull/11807)
ojs: [pkp/pkp-lib#11779 Submodule update ##Hafsa-Naeem/i11779-main-fix #5091](https://github.com/pkp/ojs/pull/5091)

Vitaliy-1:
@Hafsa-Naeem, any help required to make it merge? It only requires the fix I've mentioned with missing reference to the migration script `<migration class="PKP\migration\upgrade\v3_6_0\I11754_TaskTemplateEnhancements"/>`

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/11878
OJS: https://github.com/pkp/ojs/pull/5109
OMP: https://github.com/pkp/omp/pull/2129
OPS: https://github.com/pkp/ops/pull/1106

Vitaliy-1:
Merged

PR #11807, opened 2025-09-10
pkp/pkp-lib#11779 added API endpoint to create task templates

Hafsa-Naeem:
for #11779

### DIFF
--- a/api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
+++ b/api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
@@ -7,7 +7,7 @@
  * Copyright (c) 2025 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
- * @class PKPEditTemplateController
+ * @class PKPEditTaskTemplateController
  *
  * @ingroup api_v1_edit_templates
  *
@@ -26,7 +26,6 @@ use PKP\security\authorization\CanAccessSettingsPolicy;
 use PKP\security\authorization\ContextAccessPolicy;
 use PKP\security\Role;
 use PKP\editorialTask\Template;
-use APP\facades\Repo;
 use PKP\API\v1\editTaskTemplates\formRequests\AddEditTaskTemplate;
 use PKP\API\v1\editTaskTemplates\resources\EditTaskTemplateResource;
 
@@ -70,17 +69,12 @@ class PKPEditTaskTemplateController extends PKPBaseController
         $validated = $illuminateRequest->validated();
 
         $template = DB::transaction(function () use ($validated, $context) {
-            $emailTemplateId = null;
-            if (!empty($validated['emailTemplateKey'])) {
-                $et = Repo::emailTemplate()->getByKey($context->getId(), $validated['emailTemplateKey']);
-                $emailTemplateId = $et?->getId();
-            }
             $tpl = Template::create([
                 'stage_id' => $validated['stageId'],
                 'title' => $validated['title'],
                 'context_id' => $context->getId(),
                 'include' => $validated['include'] ?? false,
-                'email_template_id' => $emailTemplateId,
+                'email_template_key' => $validated['emailTemplateKey'] ?? null,
             ]);
 
             $tpl->userGroups()->sync($validated['userGroupIds']);
--- a/api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
+++ b/api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
@@ -38,7 +38,6 @@ class AddEditTaskTemplate extends FormRequest
             ->filter()
             ->values()
             ->all();
- 
         return [
             'stageId' => ['required', 'integer', Rule::in($stageIds)],
             'title' => ['required', 'string', 'max:255'],
@@ -56,22 +55,19 @@ class AddEditTaskTemplate extends FormRequest
 
     protected function prepareForValidation(): void
     {
+        $stageId = $this->input('stageId', null);
         $this->merge([
             'include' => filter_var($this->input('include', false), FILTER_VALIDATE_BOOLEAN),
             'userGroupIds' => array_values(array_map('intval', (array) $this->input('userGroupIds', []))),
+            'stageId' => is_null($stageId) ? $stageId : (int) $stageId,
         ]);
     }
 
-    public function validated($key = null, $default = null)
+    protected function passedValidation(): void
     {
-        $data = parent::validated();
-
-        return [
-            'stageId' => (int) $data['stageId'],
-            'title' => $data['title'],
-            'include' => $data['include'] ?? false,
-            'emailTemplateKey' => $data['emailTemplateKey'] ?? null,
-            'userGroupIds' => $data['userGroupIds'],
-        ];
+        $key = $this->input('emailTemplateKey');
+        if (is_string($key)) {
+            $this->merge(['emailTemplateKey' => trim($key)]);
+        }
     }
 }
--- a/api/v1/editTaskTemplates/resources/EditTaskTemplateResource.php
+++ b/api/v1/editTaskTemplates/resources/EditTaskTemplateResource.php
@@ -28,12 +28,21 @@ class EditTaskTemplateResource extends JsonResource
             'stageId' => (int) $this->stage_id,
             'title' => $this->title,
             'include' => (bool) $this->include,
-            'emailTemplateId' => $this->email_template_id ? (int) $this->email_template_id : null,
-            'userGroupIds' => $this->whenLoaded('userGroups', fn () => $this->userGroups->pluck('user_group_id')->values()->all()),
-            'userGroups' => $this->whenLoaded('userGroups', fn () => $this->userGroups->map(fn ($ug) => [
-                'id' => (int) $ug->user_group_id,
-                'name' => method_exists($ug, 'getLocalizedName') ? $ug->getLocalizedName() : ($ug->name ?? null),
-            ])),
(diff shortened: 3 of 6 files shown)


api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,258 @@
...
+            'emailTemplateId' => $illuminateRequest->input('emailTemplateId'),
+            'userGroupIds' => array_values(array_map('intval', (array) $illuminateRequest->input('userGroupIds', []))),
+            'participants' => (array) $illuminateRequest->input('participants', []),
+        ];
+
+        if ($illuminateRequest->has('dateDue')) {
+            $payload['dateDue'] = $illuminateRequest->input('dateDue');
+        }
+
+        // validation
```
Vitaliy-1: You can use the formRequest for validation https://github.com/pkp/pkp-lib/blob/main/api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
Then, type hint it as an argument in the Controller method https://github.com/pkp/pkp-lib/blob/main/api/v1/reviewers/suggestions/ReviewerSuggestionController.php#L155

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,258 @@
...
+                'context_id' => $context->getId(),
+                'include' => $payload['include'],
+                'email_template_id' => $payload['emailTemplateId'] ?? null,
+            ]);
+
+            // attach user groups via pivot
+            $template->userGroups()->sync($payload['userGroupIds']);
+
+            // store extra template config in settings (non-localized)
+            DB::table('edit_task_template_settings')->insert([
```
Vitaliy-1: I don't think we have got any settings for the Task Template

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,258 @@
...
+                    'locale' => '',
+                    'setting_name' => 'dateDue',
+                    'setting_value'=> $payload['dateDue'] ? (string) $payload['dateDue'] : null,
+                ],
+            ]);
+
+            return $template->getKey();
+        });
+
+        // serialize with TaskResource
```
Vitaliy-1: Regarding Resources, it should have its own resource. You can use task resource as an example: https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/resources/TaskResource.php

api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        return true;
+    }
+
+    public function rules(): array
+    {
+        $contextId = Application::get()->getRequest()->getContext()->getId();
+
+        return [
+            'stageId' => ['required', 'integer', 'min:1'],
```
Vitaliy-1: you can check if the application has the given stage by `Application::getApplicationStages()`

api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
```diff
@@ -0,0 +1,68 @@
...
+            'title' => ['required', 'string', 'max:255'],
+            'include' => ['sometimes', 'boolean'],
+            'emailTemplateId' => ['sometimes', 'nullable', 'integer', Rule::exists('email_templates', 'email_id')],
```
Vitaliy-1: Maybe specifying the key would be better?
And then we can validate with a custom rule: https://laravel.com/docs/12.x/validation#using-rule-objects. Trying to find all templates by their keys in the given context:
```php
$templates = Repo::emailTemplate()->getCollector()->getMany();
```

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,99 @@
...
+    public function add(AddEditTaskTemplate $illuminateRequest): JsonResponse
+    {
+        $context = $this->getRequest()->getContext();
+        $validated = $illuminateRequest->validated();
+
+        $template = DB::transaction(function () use ($validated, $context) {
+            $emailTemplateId = null;
+            if (!empty($validated['emailTemplateKey'])) {
+                $et = Repo::emailTemplate()->getByKey($context->getId(), $validated['emailTemplateKey']);
+                $emailTemplateId = $et?->getId();
```
Vitaliy-1: I think we should make an additional migration and reference from this column email template key, which also is unique. That's how later we can get either the default template, the customised one or a brand new, created specifically for the task

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public function add(AddEditTaskTemplate $illuminateRequest): JsonResponse
+    {
+        $context = $this->getRequest()->getContext();
+        $validated = $illuminateRequest->validated();
+
+        $template = DB::transaction(function () use ($validated, $context) {
+            $emailTemplateId = null;
+            if (!empty($validated['emailTemplateKey'])) {
+                $et = Repo::emailTemplate()->getByKey($context->getId(), $validated['emailTemplateKey']);
```
Vitaliy-1: I think we can use one of the validation hooks instead of the controller for such actions. Not sure but it looks like we can create a method in the class that extends `FormRequest` for this:
```php
protected function passedValidation() 
{
  // transform data after validation is passed
}
```

api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
```diff
@@ -0,0 +1,77 @@
...
+
+    protected function prepareForValidation(): void
+    {
+        $this->merge([
+            'include' => filter_var($this->input('include', false), FILTER_VALIDATE_BOOLEAN),
+            'userGroupIds' => array_values(array_map('intval', (array) $this->input('userGroupIds', []))),
+        ]);
+    }
+
+    public function validated($key = null, $default = null)
```
Vitaliy-1: Can you double check if we need to override this method here? I think we can avoid that...

api/v1/editTaskTemplates/resources/EditTaskTemplateResource.php
```diff
@@ -0,0 +1,39 @@
...
+        return [
+            'id' => (int) $this->id,
+            'stageId' => (int) $this->stage_id,
+            'title' => $this->title,
+            'include' => (bool) $this->include,
+            'emailTemplateId' => $this->email_template_id ? (int) $this->email_template_id : null,
+            'userGroupIds' => $this->whenLoaded('userGroups', fn () => $this->userGroups->pluck('user_group_id')->values()->all()),
+            'userGroups' => $this->whenLoaded('userGroups', fn () => $this->userGroups->map(fn ($ug) => [
+                'id' => (int) $ug->user_group_id,
+                'name' => method_exists($ug, 'getLocalizedName') ? $ug->getLocalizedName() : ($ug->name ?? null),
```
Vitaliy-1: Can you double check if we need to return an assoc array with data in all locales or just a localized string?

Vitaliy-1 (changes requested):
Thanks! A couple of additional comments

api/v1/editTaskTemplates/formRequests/AddEditTaskTemplate.php
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+namespace PKP\API\v1\editTaskTemplates\formRequests;
+
+use APP\core\Application;
+use APP\facades\Repo;
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
+
+class AddEditTaskTemplate extends FormRequest
```
Vitaliy-1, line 23: It can be renamed just to the `AddTaskTemplate` to avoid confusion

classes/migration/upgrade/v3_6_0/I11779_UseEmailTemplateKeyOnEditTaskTemplates.php
```diff
@@ -0,0 +1,72 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+class I11779_UseEmailTemplateKeyOnEditTaskTemplates extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('edit_task_templates', function (Blueprint $table) {
+            if (!Schema::hasColumn('edit_task_templates', 'email_template_key')) {
```
Vitaliy-1, line 25: Personally, I'd avoid using such checks. That's because of something like this goes wrong, it's better to end the migration and give user the ability to restart it from the backup

classes/migration/upgrade/v3_6_0/I11779_UseEmailTemplateKeyOnEditTaskTemplates.php
```diff
@@ -0,0 +1,72 @@
...
+                        DB::table('edit_task_templates')
+                            ->where('edit_task_template_id', $row->edit_task_template_id)
+                            ->update(['email_template_key' => $key]);
+                    }
+                }
+            }, 'edit_task_template_id');
+
+        // drop FK and old column
+        Schema::table('edit_task_templates', function (Blueprint $table) {
+            try { $table->dropForeign(['email_template_id']); } catch (\Throwable $e) {}
```
Vitaliy-1, line 59: The same as above

Issue #11806, opened 2025-09-10
[Recaptcha] | Login - Recaptcha is not enforced in user login


kaitlinnewson:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When Google recaptcha is enabled and `captcha_on_login` is turned on, the checkbox is appearing on the login page but is not being properly enforced for login, so users can login without clicking on it. This may impact other places where recaptcha is used as well (e.g. user sign up).

### Steps to Reproduce

1. In `config.inc.php`, enable recaptcha and `captcha_on_login`, and set the appropriate public and private keys
2. Navigate to the login page
3. Login with any user, and don't click the recaptcha checkbox
4. See that you can login

### Expected Result

Users should have to pass the recaptcha before they can login.

### Actual Result

Recaptcha is not being properly enforced on login.

### Application Version

OJS/OMP/OPS 3.5.0-1

### Additional Information

Forum post: https://forum.pkp.sfu.ca/t/omp-3-5-0-1-recaptcha-not-enforced-on-login-in-omp-3-5-0-1/97042

kaitlinnewson:
Hi @jonasraoni @asmecher, would you be able to review the small PRs above? Thanks!

asmecher:
@kaitlinnewson, all approved, thanks (though I agree with Jonas' comment about variables and scope) -- please go ahead!

kaitlinnewson:
Thanks @jonasraoni and @asmecher, I've made some small adjustments based on the feedback.

Tribunal33:
Ok took a bit of work to get this set up but I believe it no longer allows access without selecting the recaptcha verification dialogue box.  It gives the error when attempting login without being validated by recaptcha.

I have tried it with altcha also turned on and re-tested altcha.  Both appear to work as expected.  

<img width="439" height="607" alt="Image" src="https://github.com/user-attachments/assets/bcb867fa-4ba6-4de9-90e7-480fef53641e" />

<img width="495" height="616" alt="Image" src="https://github.com/user-attachments/assets/8a475dd3-82d5-4b22-88ea-18f05e2314db" />

Tested registration too and error shows up when recaptcha not validated.

<img width="864" height="186" alt="Image" src="https://github.com/user-attachments/assets/d442ae3e-dfef-4d80-b9f8-aa5df2294bfa" />

PR #11808, opened 2025-09-10
pkp/pkp-lib#11806 fix recaptcha validation error handling on user login

kaitlinnewson:


pages/login/LoginHandler.php
```diff
@@ -129,18 +129,22 @@ public function signIn(array $args, PKPRequest $request): void
...
 
-        $error = null;
         $isCaptchaEnabled = Config::getVar('captcha', 'captcha_on_login') && Config::getVar('captcha', 'recaptcha');
         if ($isCaptchaEnabled) {
             $templateMgr->assign('recaptchaPublicKey', Config::getVar('captcha', 'recaptcha_public_key'));
             try {
                 FormValidatorReCaptcha::validateResponse($request->getUserVar('g-recaptcha-response'), $request->getRemoteAddr(), $request->getServerHost());
             } catch (Exception $exception) {
-                $error = 'common.captcha.error.missing-input-response';
+                $recaptchaError = 'common.captcha.error.missing-input-response';
```
jonasraoni: Even though it works, I'm against using variables which are not declared on the same/outer scope, so I'd vote to recover the removed line above (`$error = null`).

pages/login/LoginHandler.php
```diff
@@ -129,18 +129,22 @@ public function signIn(array $args, PKPRequest $request): void
...
             }
         }
 
-        $error = $this->_validateAltchasResponse($request, 'altcha_on_login');
+        $isAltchaEnabled = Config::getVar('captcha', 'altcha_on_login') && Config::getVar('captcha', 'altcha');
+        if ($isAltchaEnabled) {
+            $altchaError = $this->_validateAltchasResponse($request, 'altcha_on_login');
+        }
```
jonasraoni: Assuming `altcha_on_login` and `recaptcha_on_login` work properly together, then if we're not going to show both errors, the validation below can be skipped if the `$recaptchaError` has a value.

jonasraoni (approved):
I've requested minor updates. As it's a small PR I'll mark it as approved to not block further work 👌

pages/login/LoginHandler.php
kaitlinnewson: Since they are both outputting the same error message at the end I think it's okay, but it's unlikely users would be running both (though they could).

Issue #11732, opened 2025-08-24
Add documentation and testing


ewhanson:
The following need to be added for the public commenting feature

- [x] API documentation
- [x] Automated tests
- [x] User-facing documentation on how to use comments as a commenter and moderator. Added at https://github.com/pkp/pkp-lib/issues/11939

### Pull Requests
 
* #### Automated Tests
    ojs: https://github.com/pkp/ojs/pull/5087
    pkp-lib: https://github.com/pkp/pkp-lib/pull/11810

* #### API Documentation
    ojs: https://github.com/pkp/ojs/pull/5093

taslangraham:
The tests are currently [passing](https://github.com/pkp/ojs/actions/runs/17633535827?pr=5087), but they might break once the [changes](https://github.com/pkp/pkp-lib/issues/11327#issuecomment-3279417530) from @jardakotesovec are merged. I’ll revisit and update them as needed after those changes go through.

PR #11810, opened 2025-09-11
pkp/pkp-lib#11732 Add tests for public comments

taslangraham:
For pkp/pkp-lib#11732


cypress/tests/integration/publicComents/PublicComments.cy.js
```diff
@@ -0,0 +1,545 @@
...
+
+	// Full public comments feature is only available in OJS
+	if (Cypress.env('defaultGenre') !== 'Article Text') {
+		return;
+	}
+
+	it('should enable public commenting', () => {
+		cy.login('rvaca');
+
+		cy.visit('http://localhost:8000/index.php/publicknowledge/en/management/settings/website#content');
```
ewhanson, line 33: Should this also just be `index.php/publicknowledge/en/management/settings/website#content`?

cypress/tests/integration/publicComents/PublicComments.cy.js
```diff
@@ -0,0 +1,545 @@
...
+			cy.login(user);
+			cy.visit('index.php/publicknowledge/en/article/view/mwandenga-signalling-theory');
+
+			cy.contains('Version of Record 1.0')
+				.parent();
+			cy.get('.BaseCommentsNewInput').find('textarea')
+				.should('be.visible').type(`${testCommentText} - ${index + 1}`);
+			cy.contains('button.BaseCommentsNewSubmit', 'Submit').click();
+
+			cy.wait(1000);
```
ewhanson: A general comment on these, not this one specifically, but did you find you needed a whole second of waiting here? I noticed there are quite a few and they add up. If they're all necessary, no problem, but just wanted to check.

taslangraham: Yea, they were added in because the test logic would intermittently get executed before the DOM was properly loaded/updated. But, maybe the number of second per wait can be reduced.

PR #11821, opened 2025-09-12
create reviewer invitation

ipula:


classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace PKP\components\forms\invitation;
+
+use PKP\components\forms\FieldOptions;
+use PKP\components\forms\FieldText;
+use PKP\components\forms\FormComponent;
+
+class ReviewerReviewDetailsForm extends FormComponent
```
defstat, line 25: @ipula can all files/classes/code related to a specific invitation be contained withing the `invitations/specificInvitation` folder? So that the files don't appear all over the code base?

classes/components/forms/invitation/ReviewerReviewDetailsForm.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param string $action URL to submit the form to
+     * @param array $locales Supported locales
+     */
+    public function __construct(string $action, array $locales)
+    {
+        $this->action = $action;
+        $this->locales = $locales;
+
+
```
defstat: unnecessary empty space?

classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php
```diff
@@ -99,90 +122,120 @@ public function updateMailableWithUrl(Mailable $mailable): void
...
+    /**
+     * @inheritDoc
+     */
+    public function getValidationMessages(ValidationContext $validationContext = ValidationContext::VALIDATION_CONTEXT_DEFAULT): array
+    {
+        return [];
     }
 
-    public function getInvitationActionRedirectController(): ?InvitationActionRedirectController
+    public function getCreateInvitationController(Invitation $invitation): CreateInvitationController
```
defstat, line 256: Here why do we need to pass the Invitation object? Wouldn't 

```
return new ReviewerAccessInviteCreateController($this);
```

work correctly?

Below also.

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * @param Invitation $invitation
+     */
+    public function __construct(Invitation $invitation)
+    {
+        $this->invitation = $invitation;
+    }
+
+    public function createHandle(Request $request, $userId = null): void
```
defstat, line 46: Here, are you sure that you need the `$userId` as a param?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,142 @@
...
+use PKP\invitation\core\Invitation;
+use PKP\invitation\core\InvitationUIActionRedirectController;
+use PKP\invitation\invitations\userRoleAssignment\resources\UserRoleAssignmentInviteResource;
+use PKP\invitation\stepTypes\SendInvitationStep;
+use PKP\security\Role;
+use PKP\userGroup\UserGroup;
+
+class ReviewerAccessInviteUIController extends InvitationUIActionRedirectController
+{
+    protected Invitation $invitation;
```
defstat, line 36: I am wondering whether here we can be specific about the type of the Invitation we are expecting - probably in the constructor below also.

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,142 @@
...
+                    'settings',
+                    ['access']
+                )
+        ];
+        $breadcrumbs[] = [
+            'id' => 'invitationWizard',
+            'name' => __('reviewerInvitation.wizard.pageTitle'),
+        ];
+        $steps = new SendInvitationStep(
+            null, $context, $user,'reviewerAccess'
```
defstat: Those hard-coded params, like 'reviewerAccess', should be somewhere centrally defined, if they are even necessary.

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,142 @@
...
+        ]);
+        $templateMgr->display('/invitation/userInvitation.tpl');
+    }
+
+    public function editHandle(Request $request): void
+    {
+        // TODO: Implement editHandle() method.
+    }
+
+    public function isUserReviewer($userId): bool
```
defstat: Shouldn't we check if this user is a reviewer given the specific context?

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteUIController.php
```diff
@@ -103,9 +103,14 @@ public function createHandle(Request $request, $userId = null): void
             'id' => 'invitationWizard',
             'name' => __('invitation.wizard.pageTitle'),
         ];
-        $steps = new SendInvitationStep();
+        $steps = new SendInvitationStep(
```
defstat: I am a little bit worried about this one, and generally about the `AcceptInvitationStep`, `SendInvitationStep` and `InvitationStepTypes`

Be the look of the classes and the placement they should be generic classes used or overridden by specific invitations. But, instead, they incorporate specific invitation code inside, by passing the invitation type. 

This does not follow the intended architecture, where all the specific invitation code should be contained in the invitation "context", and outside that, only generic invitation code should reside.

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteUIController.php
```diff
@@ -103,9 +103,14 @@ public function createHandle(Request $request, $userId = null): void
...
+            null,
+            $context,
+            $user,
+            'userRoleAssignment'
```
defstat: Event if we needed to use this call (see comment above for the reason I think we should not use it as is) here the type of the invitation should not be a hard-coded param.

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteUIController.php
```diff
@@ -184,9 +189,14 @@ public function editHandle(Request $request): void
             'id' => 'invitationWizard',
             'name' => __('invitation.wizard.pageTitle'),
         ];
-        $steps = new SendInvitationStep();
+        $steps = new SendInvitationStep(
```
defstat: Same as above

classes/invitation/stepTypes/AcceptInvitationStep.php
```diff
@@ -23,29 +23,39 @@
 
 class AcceptInvitationStep extends InvitationStepTypes
 {
+
+    private ?Invitation $invitation;
+    private Context $context;
+    private ?User $user;
+    private string $invitationType;
+    public function __construct(?Invitation $invitation, Context $context, ?User $user,string $invitationType)
```
defstat: Here, if you can pass the `$invitation` why you also need the $invitationType? Is there a case where you will have a step, and the code's business logic will not have an invitation handy?

classes/invitation/stepTypes/AcceptInvitationStep.php
```diff
@@ -162,7 +172,7 @@ private function userDetailsStep(Context $context): \stdClass
      *
      * @throws \Exception
      */
-    private function acceptInvitationReviewStep(Context $context): \stdClass
+    private function acceptInvitationReviewStep(): \stdClass
```
defstat: Here it is clear that you call a specific invitation code within this generic class. We should find a way to avoid that. Consider overriding the generic class to handle specific invitation "steps" handling

classes/invitation/stepTypes/AcceptInvitationStep.php
```diff
@@ -176,7 +186,7 @@ private function acceptInvitationReviewStep(Context $context): \stdClass
                 'userDetails',
                 __('acceptInvitation.userDetails.form.name'),
                 __('acceptInvitation.userDetails.form.description'),
-                new AcceptUserDetailsForm('accept', $this->getFormLocals($context)),
+                new AcceptUserDetailsForm('accept', $this->getFormLocals($this->context)),
```
defstat: Same as above and in other places also

classes/invitation/stepTypes/AcceptInvitationStep.php
```diff
@@ -201,9 +211,9 @@ private function acceptInvitationReviewStep(Context $context): \stdClass
      * @param Context $context
      * @return array
      */
-    private function getFormLocals(Context $context): array
+    private function getFormLocals(): array
```
defstat: is it "Locals" or "Locales"?

classes/invitation/stepTypes/InvitationStepTypes.php
```diff
@@ -14,22 +14,29 @@
...
      * Get the invitation steps
      * use of the built-in UI for making the invitation
      */
-    abstract public function getSteps(?Invitation $invitation, Context $context, ?User $user): array;
+    abstract public function getSteps(): array;
 
     /** fake invitation for email template
      */
-    protected function getFakeInvitation(): UserRoleAssignmentInvite
+    protected function getFakeInvitation(string $invitationType): UserRoleAssignmentInvite | ReviewerAccessInvite
```
defstat: Same as above

classes/invitation/stepTypes/SendInvitationStep.php
```diff
@@ -16,33 +16,52 @@
...
-        if (!$invitation && !$user) {
+        if ((!$this->invitation  && !$this->user)) {
             $steps[] = $this->invitationSearchUser();
         }
-        $steps[] = $this->invitationDetailsForm($context);
-        $steps[] = $this->invitationInvitedEmail($context);
+        if(self::INVITATION_USER_ROLE_ASSIGNMENT === $this->invitationType || (self::INVITATION_REVIEWER_ACCESS_INVITE === $this->invitationType && !$this->user)) {
+            $steps[] = $this->invitationDetailsForm();
+        }
+        if(self::INVITATION_REVIEWER_ACCESS_INVITE === $this->invitationType) {
```
defstat: Same as above

classes/invitation/stepTypes/SendInvitationStep.php
```diff
@@ -167,11 +189,55 @@ private function invitationInvitedEmail(Context $context): stdClass
         return $step->getState();
     }
 
+    /**
+     * create search user section
+     */
+    private function invitationReviewDetails(): stdClass
```
defstat: Same as above

locale/en/invitation.po
```diff
@@ -379,3 +379,44 @@ msgstr "The user was disabled. You cannot assign them a role while they are disa
...
+msgstr "Review Types"
+
+msgid "reviewerInvitation.responseDueDate"
+msgstr "Review Response Date"
+
+msgid "reviewerInvitation.reviewDueDate"
+msgstr "Review Due Date"
+
+msgid "reviewerInvitation.modal.message"
+msgstr "{$email} has been invited to review the submission "{$articleTitle}"<br><br> You can be updated about the user's decision on the reviewer panel in the review workflow or through email and OJS notifications."
```
defstat: Is this correctly written? I think that a `"` is missing

pages/invitation/InitializeInvitationUIHandler.php
```diff
@@ -87,7 +88,7 @@ public function create(array $args, Request $request): void
             $invitationType = $arg;
             $invitation = app(Invitation::class)->createNew($invitationType);
             $invitationHandler = $invitation->getInvitationUIActionRedirectController();
-            $invitationHandler->createHandle($request);
+            $invitationHandler->createHandle($request,$userId);
```
defstat, line 94: Do we have to define and pass the $userId here? I see the `userId` is in the `$request`, and no other operations are done to it.

But if we need to do that for other reasons (clarity perhaps), leave a space after the `,`, like

```
$invitationHandler->createHandle($request, $userId);
```

classes/mail/mailables/ReviewerAccessInvitationNotify.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        // Invitation User
+        $sendIdentity = $this->invitation->getMailableReceiver($locale);
+
+        // Inviter
+        $inviter = $this->invitation->getInviter();
+
+        //submission
+        $submission = Repo::submission()->get((int) $this->invitation->getPayload()->submissionId);
```
defstat, line 110: Here, are we sure that the $submission exists? Shouldn't we add some checks for that?

classes/mail/mailables/ReviewerAccessInvitationNotify.php
```diff
@@ -0,0 +1,116 @@
...
+        $targetPath = Core::getBaseDir() . '/lib/pkp/styles/mailables/style.css';
+        $emailTemplateStyle = file_get_contents($targetPath);
+
+        $recipientName = !empty($sendIdentity->getFullName()) ? $sendIdentity->getFullName() : $sendIdentity->getEmail();
+
+        // Set view data for the template
+        $this->viewData = array_merge(
+            $this->viewData,
+            [
+                static::$recipientName => $recipientName,
```
defstat, line 141: Couldn't those attributes be CONSTs?

classes/mail/mailables/ReviewerAccessInvitationNotify.php
```diff
@@ -0,0 +1,116 @@
...
+
+        // Invitation User
+        $sendIdentity = $this->invitation->getMailableReceiver($locale);
+
+        // Inviter
+        $inviter = $this->invitation->getInviter();
+
+        //submission
+        $submission = Repo::submission()->get((int) $this->invitation->getPayload()->submissionId);
+        $publication = $submission->getCurrentPublication();
```
defstat, line 115: Same as above.

defstat: And in other places too. Null/validity checks must be added to prevent fatal errors

defstat (changes requested):
@ipula thanks!

I did a preliminary review on the PR's code. 

Besides other comments added, my main concern is with the intended invitation architecture, where all the specific invitation code should reside inside the specific invitation folder, and all the generic invitation code should not have something that point to a specific invitation inside. I have added specific comments for that.

classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
ipula: I moved this to invitation  reviewer access folder

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
ipula: this uses same abstract class `InvitationUIActionRedirectController`  in this class we have `abstract public function createHandle(Request $request, $userId = null): void`. the userId need because for the normal user invitation can happen through the edit user. for that we need userId

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteUIController.php
ipula: I understand what you mention here. Developer able to add new type without changing the `SendInvitation` and `AcceptInvitation`

classes/mail/mailables/ReviewerAccessInvitationNotify.php
ipula: I will check this again

classes/mail/mailables/ReviewerAccessInvitationNotify.php
ipula: I will check this again

classes/invitation/core/InvitationContextFactory.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace PKP\invitation\core;
+
+use PKP\invitation\core\enums\InvitationTypes;
+use PKP\invitation\invitations\reviewerAccess\context\ReviewerAccessInvitationContext;
+use PKP\invitation\invitations\userRoleAssignment\context\UserRoleAssignmentInvitationContext;
+
+class InvitationContextFactory
```
defstat: @ipula why do we need this as a separate Factory class? Isn't the `InvitationFactory` along with the `InvitationServiceProvider` enough for our needs?

In principle you should not need more infrastructure to accommodate the return of an invitation, other than what the `Repository` is providing.

classes/invitation/core/enums/InvitationTypes.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace PKP\invitation\core\enums;
```
defstat, line 3: Same as above - see the `InvitationServiceProvider` which is developed in order to discover all the existing Invitations.

classes/invitation/invitations/reviewerAccess/context/ReviewerAccessInvitationContext.php
```diff
@@ -0,0 +1,402 @@
...
+use PKP\invitation\sections\Email;
+use PKP\invitation\sections\Form;
+use PKP\invitation\sections\Sections;
+use PKP\invitation\steps\Step;
+use PKP\mail\mailables\ReviewerAccessInvitationNotify;
+use PKP\orcid\OrcidManager;
+use PKP\user\User;
+use PKP\userGroup\UserGroup;
+
+class ReviewerAccessInvitationContext implements InvitationContext
```
defstat: I would have named the `InvitationContext` something more relative to the UI steps that it is used for. If you have something more generic in mind, I would avoid the `Context` reference.

classes/invitation/invitations/reviewerAccess/steps/ReviewerAccessInvitationSteps.php
```diff
@@ -0,0 +1,402 @@
...
+                'label' => $name,
+            ];
+        }
+        return $locales;
+    }
+
+    /**
+     * fake invitation for email template
+     */
+    protected function getFakeInvitation(): ReviewerAccessInvite
```
defstat, line 443: What is this used for?

classes/invitation/invitations/reviewerAccess/steps/ReviewerAccessInvitationSteps.php
```diff
@@ -0,0 +1,402 @@
...
+        $step = new Step(
+            'userDetails',
+            __('acceptInvitation.userDetails.stepName'),
+            __('acceptInvitation.userDetails.stepLabel'),
+            __('acceptInvitation.userDetails.nextButtonLabel'),
+            'form',
+            __('acceptInvitation.userDetails.stepDescription'),
+        );
+        $step->addSectionToStep($sections->getState());
+        return $step->getState();
```
defstat, line 338: In `getState()` function of the `Step` class I can see that we are copying all the classes attributes to the exact same new `stdClass()` object. Do we need to do that or we could just return `this`?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
defstat: Is this resolved?

pages/invitation/InitializeInvitationUIHandler.php
defstat: Is this resolved?

classes/invitation/stepTypes/AcceptInvitationStep.php
```diff
@@ -12,205 +12,32 @@
...
 use PKP\context\Context;
 use PKP\invitation\core\Invitation;
-use PKP\invitation\sections\Form;
-use PKP\invitation\sections\Sections;
-use PKP\invitation\steps\Step;
-use PKP\orcid\OrcidManager;
+use PKP\invitation\core\InvitationContext;
 use PKP\user\User;
 
 class AcceptInvitationStep extends InvitationStepTypes
```
defstat: I am wondering whether we are complicating this more by adding `AcceptInvitationStep` and `SendInvitationStep` classes - do they have a different purpose? Can't they be part of the same class?

defstat (changes requested):
@ipula thanks! I think the code is in a good shape. 

Some comments for you. Mainly about the `InvitationContext` and `InvitationContextFactory`. Maybe some renaming could be in order here, and considering using the existing infrastructure for retrieving the Invitation related classes and overrides rather than introducing another "Factory" class.

classes/invitation/invitations/reviewerAccess/steps/ReviewerAccessInvitationSteps.php
ipula: this is for the email template that we show in the UI last step

classes/invitation/core/enums/InvitationTypes.php
```diff
@@ -0,0 +1,9 @@
...
+
+enum InvitationTypes : string
```
defstat, line 5: Why do we need that? isn't the existing InvitationFactory enough? I would prefer to avoid re-assigning those params multiple times in the code

classes/invitation/core/Invitation.php
```diff
@@ -323,13 +327,21 @@ public function invite(): bool
...
             ->byType($this->getType())
             ->byNotId($this->getId())
             ->when(isset($this->invitationModel->userId), fn (Builder $q) => $q->byUserId($this->invitationModel->userId))
             ->when(!isset($this->invitationModel->userId) && $this->invitationModel->email, fn (Builder $q) => $q->byEmail($this->invitationModel->email))
-            ->when(isset($this->invitationModel->contextId), fn (Builder $q) => $q->byContextId($this->invitationModel->contextId))
-            ->delete();
+            ->when(isset($this->invitationModel->contextId), fn (Builder $q) => $q->byContextId($this->invitationModel->contextId));
+
+        // remove the review assignments
+        foreach ($invitations->get() as $invitation) {
```
defstat: Why are we doing that? And why in such a central place of the code?

classes/invitation/core/Invitation.php
```diff
@@ -507,4 +519,28 @@ public function isPending(): bool
...
+
+    public function isInvitationUserReviewer($userId,$contextId): bool
+    {
+        if(!$userId){
+            return false;
+        }
+        $currentUserGroups = Repo::userGroup()->userUserGroups($userId,$contextId);
+        return $currentUserGroups->contains(
+            fn (UserGroup $userGroup) =>
+                $userGroup->roleId == Role::ROLE_ID_REVIEWER
```
defstat: Why this specific invitation code is in the central abstract class?

classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
```diff
@@ -0,0 +1,67 @@
...
+                    'value' => ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN,
+                    'label' => __('reviewerInvitation.reviewMethod.open'),
+                ],
+            ],
+        ]));
+
+        $this->addField(new FieldText('responseDueDate', [
+            'label' => __('reviewerInvitation.responseDueDate'),
+            'inputType' => 'date',
+            'value'=> Core::getCurrentDate(),
```
defstat, line 69: Shouldn't here be a default value assigned other than "current date"?

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
```diff
@@ -0,0 +1,197 @@
+<?php
+
```
defstat, line 2: Shouldn't we add the file signature here?

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
```diff
@@ -0,0 +1,197 @@
...
+    public function invite(Request $illuminateRequest): JsonResponse
+    {
+        $this->invitation->getPayload()->sendEmailAddress = $this->invitation->getEmail();
+
+        $existingUser = $this->invitation->getExistingUser();
+        if (isset($existingUser)) {
+            $this->invitation->getPayload()->sendEmailAddress = $existingUser->getEmail();
+        }
+
+        $this->invitation->getPayload()->inviteStagePayload = $this->invitation->getPayload()->toArray();
```
defstat, line 98: is this `inviteStagePayload` new? There is a specific need for this?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteRedirectController.php
```diff
@@ -15,50 +15,50 @@
...
-                'submissionId' => $reviewAssignment->getSubmissionId(),
-                'reviewId' => $reviewAssignment->getId(),
-            ]
-        );
-
-        $this->getInvitation()->finalize();
-
-        $request->redirectUrl($url);
+        $invitationModel = $this->getInvitation()->invitationModel->toArray();
+        $user = $invitationModel['userId'] ? Repo::user()->get($invitationModel['userId']) : null;
```
defstat: Are we checking here if the user exists?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,262 @@
...
+     * @param Invitation $invitation
+     */
+    public function __construct(Invitation $invitation)
+    {
+        $this->invitation = $invitation;
+    }
+
+    public function createHandle(Request $request, $userId = null): void
+    {
+        if(!$request->getUserVars()['submissionId'] || !$request->getUserVars()['reviewRoundId']) {
```
defstat: In other places you leave a space after the `if` like `if (!$request->getUserVars()['submissionId'] || !$request->getUserVars()['reviewRoundId'])`

and in others you don't like here. 

Can you leave one space after the `if` in all places?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,262 @@
...
+            throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
+        }
+        $submission = Repo::submission()->get((int) $request->getUserVars()['submissionId']);
+        if(!$submission){
+            throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
+        }
+
+        $invitationPayload = [
```
defstat: Why do we have to create this by defining all of the params one by one? Can't we use the invitation's payload instead?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,262 @@
...
+            'orcidValidation' => false,
+            'disabled' => false,
+            'submissionId'=>$request->getUserVars()['submissionId'],
+		    'reviewRoundId'=>$request->getUserVars()['reviewRoundId'],
+            'responseDueDate'=> (new DateTime(Core::getCurrentDate()))->format('Y-m-d'),
+		    'reviewDueDate'=> (new DateTime(Core::getCurrentDate()))->format('Y-m-d'),
+		    'reviewMethod'=> '',
+            'userGroupsToAdd' => [
+                [
+                    'userGroupId' => 16,
```
defstat: We should not use numbers here - we should find the correct named const for that.

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,262 @@
...
+            'pageWidth' => TemplateManager::PAGE_WIDTH_FULL,
+        ]);
+        $templateMgr->display('/invitation/userInvitation.tpl');
+    }
+
+    public function editHandle(Request $request): void
+    {
+        $payload = $this->invitation->getPayload()->toArray();
+        $invitationModel = $this->invitation->invitationModel->toArray();
+        $invitationMode = 'edit';
```
defstat: This 'edit' should not be added here like that. It should be a const that could be used anywhere we need to

classes/invitation/invitations/reviewerAccess/resources/BaseReviewerAccessInviteResource.php
```diff
@@ -0,0 +1,115 @@
...
+                ->toArray();
+            $userGroups[$key] = $userUserGroup;
+            $userGroups[$key]['masthead'] = $userUserGroup['masthead'] === 1;
+            $userGroups[$key]['name'] = $userGroup['name'][Locale::getLocale()];
+            $userGroups[$key]['id'] = $userGroup['userGroupId'];
+        }
+        return $userGroups;
+    }
+
+    public function transformSubmission($submissionId):array
```
defstat, line 103: Maybe submission and user should have their own `JsonResource`

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteManagerDataResource.php
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            $existingUser = $this->getExistingUser();
+
+            if (!isset($existingUser)) {
+                $newUser = $this->createNewUserFromPayload($payload);
+            }
+        }
+
+        // Return specific fields from the UserRoleAssignmentInvite
+        return array_merge($baseData, [
```
defstat, line 47: Couldn't we avoid this duplication of everything?

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php
```diff
@@ -0,0 +1,115 @@
...
+            $newUser->setAffiliation($payload->affiliation, null);
+            $newUser->setFamilyName($payload->familyName, null);
+            $newUser->setGivenName($payload->givenName, null);
+            $newUser->setCountry($payload->userCountry);
+            $newUser->setUsername($payload->username);
+            $newUser->setEmail($payload->sendEmailAddress);
+        }
+
+        // Return specific fields from the UserRoleAssignmentInvite
+        return array_merge($baseData, [
```
defstat, line 41: Again here - can't we avoid the duplication of all fields?

classes/invitation/invitations/userRoleAssignment/handlers/UserRoleAssignmentInviteUIController.php
```diff
@@ -103,9 +106,9 @@ public function createHandle(Request $request, $userId = null): void
             'id' => 'invitationWizard',
             'name' => __('invitation.wizard.pageTitle'),
         ];
-        $steps = new SendInvitationStep();
+        $steps = $this->invitation->buildSendSteps($context, $user);
```
defstat, line 108: I would expect the `buildSendSteps` to be part of the handler more than of the `invitation` business objects

classes/invitation/invitations/userRoleAssignment/steps/UserRoleAssignmentInvitationSteps.php
```diff
@@ -1,37 +1,40 @@
 <?php
-/**
```
defstat, line 2: Why have we removed that?

classes/mail/mailables/ReviewerAccessInvitationNotify.php
defstat: Have you checked this? It seems this comment is unresolved

classes/migration/upgrade/v3_6_0/I9658_UpdateReviewerAssignmentsReviewerId.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Support\Facades\Schema;
+use PKP\migration\Migration;
+
+class I9658_UpdateReviewerAssignmentsReviewerId extends Migration
+{
+
+    public function up(): void
+    {
+        Schema::table('review_assignments', function (Blueprint $table) {
+            $table->dropForeign(['reviewer_id']);
```
defstat, line 27: Why shouldn't we have the foreign key here?

pages/invitation/InitializeInvitationUIHandler.php
```diff
@@ -77,6 +77,7 @@ public function create(array $args, Request $request): void
         }
 
         $this->setupTemplate($request);
+        $userId = $request->getUserVars()['userId'];
```
defstat, line 80: We are getting a `UserId` from the request -> Is this safe? Are we cross-checking with the invitation to be sure that a user does not have access to an invitation that it shouldn't?

defstat (changes requested):
@ipula some comments here - just a note: even though I believe this is already addressed, other parts of the code should also be reviewed. The solution as a whole, and the other invitation types that might be affected, need to be considered as well. Keeping the changes for this invitation as isolated as possible will reduce the risk of introducing new errors in others.

Tagging @jardakotesovec, @Tribunal33, and @asmecher on this.

classes/invitation/core/enums/InvitationTypes.php
ipula: This can be removed. There is separate invitation factory function to get type

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteRedirectController.php
ipula: Yes. If use exists we pass user object if not we pass null

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
ipula: I will replace this with reviewer Id by reviewer roleid

classes/invitation/invitations/userRoleAssignment/steps/UserRoleAssignmentInvitationSteps.php
ipula: Instead of having common steps. I created  invitation type base steps. because steps changed with invitation types

classes/invitation/core/Invitation.php
ipula: I move this to userGroup repository

classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
ipula: I think there is no other use case for this value

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
ipula: this is same as user role assignment

pages/invitation/InitializeInvitationUIHandler.php
ipula: How about we validate this userId?

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php
ipula: We need all the user related fields. Its the same UIs and extra UIs for submission details

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
ipula: there are so many unwanted fields in the initial payload if I used invitation payload as initial payload

classes/mail/mailables/ReviewerAccessInvitationNotify.php
ipula:
```
        $submission = Repo::submission()->get((int) $this->invitation->getPayload()->submissionId);
        if (!$submission) {
            throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
        } 
```
        
I add null validations

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteManagerDataResource.php
ipula: We need all the fields that need to create a user

classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
defstat: Here we should check with @asmecher or @bozana perhaps - to see if there is a default interval that the `responseDueDate` should respect.

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
defstat: @ipula this is tagged as resolved - Shouldn't we add the signature comments like we do for all code files?

something like 

```
/**
 * @file classes/invitation/core/Invitation.php
 *
 * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
 *
 * @class Invitation
 *
 * @brief Abstract class for all Invitations
 */
```

The same applies to all files bellow also

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
```diff
@@ -0,0 +1,214 @@
...
+                }
+                $inviteResult = $this->invitation->invite();
+                if (!isset($inviteResult)) {
+                    return response()->json([
+                        'errors' => $this->invitation->getErrors()
+                    ], Response::HTTP_UNPROCESSABLE_ENTITY);
+                }
+
+                // remove the review assignments
+                foreach ($existingInvitation as $invitation) {
```
defstat: Here you can use something like 

```
$reviewAssignmentId = $this->invitation->getPayload()->reviewAssignmentId

if ($reviewAssignmentId) { ..... **other code**

```

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
```diff
@@ -0,0 +1,214 @@
...
+                    if($invitation->payload['reviewAssignmentId']) {
+                        $reviewAssignment = Repo::reviewAssignment()->get($invitation->payload['reviewAssignmentId']);
+                        $reviewAssignment && Repo::reviewAssignment()->delete($reviewAssignment);
```
defstat: What exactly are we doing here? What are we deleting and why?

I am proposing a code change for the Invitation, that will allow the removal of older invitations taking into account same userId, same submissionId and same reviewRoundId, rather than only the userId as in the context of the previously developed invitation types. 

The PR is here https://github.com/ipula/pkp-lib/pull/5

Maybe this reviewAssignment manipulation should happen there?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
defstat: @ipula is this comment resolved?

classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php
defstat: @ipula what about this one?

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteManagerDataResource.php
```diff
@@ -60,6 +60,8 @@ public function toArray(Request $request)
             'emailBody' => $payload->emailBody,
             'userGroupsToAdd' => $this->transformUserGroups($payload->userGroupsToAdd),
             'username' => $payload->username,
+            'submissionId' => $payload->submissionId,
```
defstat: Why we need to make changes to UserRoleAssignment, reflecting changes causing by the ReviewerAccess invitation? 

I think it is risky to couple the two. 

When we send a Reviewer Role invitation, can we assign a specific submission review to the user? If not, then I think that we should avoid adding those additional fields here.

In principle the different invitations should not depend on each other, but rather, work completely independently.

classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php
```diff
@@ -99,90 +123,131 @@ public function updateMailableWithUrl(Mailable $mailable): void
...
     /**
      * @inheritDoc
      */
-    public function getValidationMessages(ValidationContext $validationContext = ValidationContext::VALIDATION_CONTEXT_DEFAULT): array
+    public function updatePayload(?ValidationContext $validationContext = null): ?bool
     {
-        return [];
+        // Encrypt the password if it exists
+        // There is already a validation rule that makes username and password fields interconnected
+        if (isset($this->getPayload()->username) && isset($this->getPayload()->password) && !$this->getPayload()->passwordHashed) {
```
defstat: @ipula can we re-use some existing code here?

classes/invitation/repositories/Repository.php
```diff
@@ -62,4 +65,28 @@ public function getByKey($key): ?Invitation
 
         return app(Invitation::class)->getExisting($invitationModel->type, $invitationModel);
     }
+
+    public function getInvitationByReviewerAssignmentId(int $reviewerAssignmentId) : ?Invitation
```
defstat: Why do we need this specific invitation code in the core level?

classes/mail/mailables/ReviewerAccessInvitationNotify.php
defstat: Yes - This comment was about the `publication`

classes/userGroup/Repository.php
```diff
@@ -620,4 +620,22 @@ public static function forgetEditorialHistoryCache(int $contextId): void
...
         Cache::forget($cacheKey);
     }
+
+    /**
+     * Check user has an active reviewer role
+     * @param int|null $userId
+     * @param int $contextId
+     * @return bool
+     */
+    public function userIsReviewer(?int $userId, int $contextId): bool
```
defstat: @ipula 

This is a user related function. Never the less, isn't something like this 

```
$user->hasRole([Role::ROLE_ID_REVIEWER], $request->getContext()->getId());
```

enough for what you need to do?

pages/invitation/InitializeInvitationUIHandler.php
```diff
@@ -87,12 +91,12 @@ public function create(array $args, Request $request): void
...
             $invitation = app(Invitation::class)->createNew($invitationType);
             $invitationHandler = $invitation->getInvitationUIActionRedirectController();
-            $invitationHandler->createHandle($request);
+            $invitationHandler->createHandle($request,$userId);
         }
     }
 
     /**
-     * Edit an invitation
+     * Edit an invitation$userId
```
defstat: @ipula is this change intentional?

defstat (changes requested):
@ipula thanks for the changes.

I’ve added some review comments, and there are also a few unresolved comments from previous review rounds that should probably be addressed.

I’ve also made some code additions to improve how this invitation handles **multiple invitations for the same user or the same email**. The goal is for this invitation type to correctly de-duplicate pending invitations based on:

 * `submissionId` and `reviewRoundId`, and
 * `userId` or `email`

 You can find the related changes here:
 [https://github.com/ipula/pkp-lib/pull/5](https://github.com/ipula/pkp-lib/pull/5)

Please review these changes and adjust as needed to align with the intended business logic for this invitation type, which you’re more familiar with than I am.

Finally, I ran into two issues during testing (apologies if these are already known or filed):

1. When cancelling an invitation *before* the final **Invite** action, the invitation does not transition to **Canceled** (nor is it removed).
2. When sending an invitation to a **new reviewer** (no `userId`), the process fails because it attempts to create a `review_assignment` without a `user_id`.

Tagging @asmecher and @Tribunal33 for visibility on the above.

classes/invitation/repositories/Repository.php
ipula: we have to add invitation id into the reviewer list for edit invitation. It will use for review assignment actions like `send reminder(edit invitation and send new accept link)` and also once user decline the invitation `resend the review request(edit and update the review assignment and create new invitation)`

classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteManagerDataResource.php
ipula: That may be a mistake we do not need that

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
ipula: yes this works fine

classes/userGroup/Repository.php
ipula: Yes I can replace with this

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
ipula:
```
public function __construct(Invitation $invitation)
    {
        if (!$invitation instanceof ReviewerAccessInvite) {
            throw new Exception("invalid invitation type");
        }
        $this->invitation = $invitation;
    }
```

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -668,7 +668,15 @@ public function updateUnassignReviewer($args, $request)
 
         // Create mailable and send email
         if ($unassignReviewerForm->execute() && !$request->getUserVar('skipEmail')) {
-            $reviewer = Repo::user()->get($reviewAssignment->getReviewerId());
+
+            if(!$reviewAssignment->getReviewerId()){
+                $reviewer = Repo::user()->newDataObject();
+                $reviewer->setEmail($reviewAssignment->getData('email'));
```
defstat, line 678: is this intentional? To add the email to all the fields? Perhaps because we have nothing else to put in those mandatory DB fields.

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -668,7 +668,15 @@ public function updateUnassignReviewer($args, $request)
 
         // Create mailable and send email
         if ($unassignReviewerForm->execute() && !$request->getUserVar('skipEmail')) {
-            $reviewer = Repo::user()->get($reviewAssignment->getReviewerId());
+
+            if(!$reviewAssignment->getReviewerId()){
```
defstat: Here we could have the `!$reviewAssignment->getReviewerId()` into a new variable outside the `if` and use it inside. 

Also there are inconsistences regarding statement spacings - here and elsewhere - should be `if (!$reviewAssignment->getReviewerId()) {`

classes/invitation/core/Invitation.php
```diff
@@ -507,4 +561,16 @@ public function isPending(): bool
 
         return false;
     }
+
+    public function buildSendSteps(Context $context, ?User $user): array
```
defstat, line 570: Function signature missing - here and elsewhere.

classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php
defstat: This is still an open comment

classes/invitation/invitations/reviewerAccess/forms/ReviewerReviewDetailsForm.php
defstat: Is this resolved?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteRedirectController.php
```diff
@@ -15,50 +15,46 @@
...
+use PKP\invitation\core\Invitation;
 use PKP\invitation\core\InvitationActionRedirectController;
-use PKP\invitation\invitations\reviewerAccess\ReviewerAccessInvite;
+use PKP\invitation\invitations\reviewerAccess\steps\ReviewerAccessInvitationSteps;
+use PKP\user\User;
 
 class ReviewerAccessInviteRedirectController extends InvitationActionRedirectController
 {
-    public function getInvitation(): ReviewerAccessInvite
+    public function getInvitation(): \PKP\invitation\core\Invitation
```
defstat: Shouldn't we return the specific type of Invitation here? We know what kind of object we are returning, being in the context of a specific invitation

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteRedirectController.php
```diff
@@ -15,50 +15,46 @@
...
-            [
-                'submissionId' => $reviewAssignment->getSubmissionId(),
-                'reviewId' => $reviewAssignment->getId(),
-            ]
-        );
-
-        $this->getInvitation()->finalize();
-
-        $request->redirectUrl($url);
+        $invitationModel = $this->getInvitation()->invitationModel->toArray();
```
defstat: Shouldn't we use the `getPayload` of the invitation here?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
defstat: Again here - don't we know what specific invitation type are we returning, being in the context of a specific invitation?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
defstat: Is this resolved? Why can't we specify the payload so that we don't have unwanted fields?

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    public function createHandle(Request $request, $userId = null): void
+    {
+
+        if (!$request->getUserVars()['submissionId'] || !$request->getUserVars()['reviewRoundId']) {
+            throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
+        }
+        $submission = Repo::submission()->get((int) $request->getUserVars()['submissionId']);
+        if (!$submission){
```
defstat: Here is seems that we are relying on the `$request->getUserVars()`, without double-checking on the validity or sanity of the input. Could cause security issues if left unchecked

classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
```diff
@@ -0,0 +1,235 @@
+<?php
+
+/**
+ * @file classes/invitation/invitations/reviewerAccess/handlers/api/ReviewerAccessInviteCreateController.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
```
defstat, line 6: Should we go for `Copyright (c) 2026` for new files?

classes/invitation/invitations/reviewerAccess/payload/ReviewerAccessInvitePayload.php
```diff
@@ -14,14 +14,252 @@
...
         public ?int $reviewAssignmentId = null,
-    ) 
+        public ?array $selectedFileIds = null,
+        public ?array $userInterests = null,
+    )
     {
         parent::__construct(get_object_vars($this));
     }
+
+    public function getValidationRules(ReviewerAccessInvite $invitation, ValidationContext $validationContext = ValidationContext::VALIDATION_CONTEXT_DEFAULT): array
```
defstat, line 67: Don't forget to check validation rules on real scenarios

classes/invitation/invitations/reviewerAccess/resources/BaseReviewerAccessInviteResource.php
defstat: This remains unresolved

classes/invitation/invitations/reviewerAccess/resources/BaseReviewerAccessInviteResource.php
```diff
@@ -0,0 +1,127 @@
...
+            'familyName' => $user->getFamilyName(null),
+            'givenName' => $user->getGivenName(null),
+            'country' => $user->getCountry(),
+            'affiliation' => $user->getAffiliation(null),
+            'orcid' => $user->getOrcid(),
+            'orcidIsVerified' => $user->hasVerifiedOrcid(),
+        ];
+    }
+
+    protected function createNewUserFromPayload(ReviewerAccessInvitePayload $payload): User
```
defstat: I would argue that this is more of a `Payload` function than a `Resource` one

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        // Get all attributes of the invitationModel as an array
+        $baseData = parent::toArray($request);
+        $existingUser = $this->getExistingUser();
+        $newUser = null;
+
+        $payload = $this->getPayload();
+
+        if (!isset($existingUser)) {
+            $newUser = new User();
```
defstat: isn't this duplication of `createNewUserFromPayload`

classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php
defstat: Maybe we could avoid clauses like `'orcidAccessToken' => $payload->orcidAccessToken,` and have only the lines that actually have transformations like `'userGroupsToAdd' => $this->transformUserGroups($payload->userGroupsToAdd),` for example

classes/invitation/repositories/Repository.php
defstat: Right - the question is why this has to be in the core level and not the specific Invitation level

classes/submission/reviewer/ReviewerAction.php
```diff
@@ -166,4 +175,18 @@ public function getResponseEmail(
 
         return $mailable;
     }
+
+    /**
+     * create a fake reviewer for invitations that not accepted yet
+     * @param $reviewAssignment
+     * @return User
+     */
+    private function fakeReviewer($reviewAssignment):User{
```
defstat: I would call him "Temporary" rather that "Fake"...

controllers/grid/users/reviewer/form/ReviewerNotifyActionForm.php
```diff
@@ -80,11 +81,17 @@ public function initData()
             'dateConfirmed' => $reviewAssignment->getDateConfirmed(),
             'reviewerId' => $reviewerId,
         ]);
+        if(!$reviewerId){
```
defstat: Again here, can't we use the `createNewUserFromPayload` function?

controllers/grid/users/reviewer/form/ReviewerNotifyActionForm.php
```diff
@@ -80,11 +81,17 @@ public function initData()
...
+            $tempUser->setEmail($reviewAssignment->getData('email'));
+            $tempUser->setFamilyName($reviewAssignment->getData('email'),null);
+            $tempUser->setGivenName($reviewAssignment->getData('email'),null);
+        }
 
         $context = $request->getContext();
         $mailable = $this->getMailable($context, $submission, $reviewAssignment);
         $mailable->sender($request->getUser());
-        $mailable->recipients([Repo::user()->get($reviewerId)]);
+        $mailable->recipients([$reviewerId?Repo::user()->get($reviewerId):$tempUser]);
```
defstat: Here it seems that we are creating the `$tempUser` inside the `if` clause, and then using it outside of that `if` clause. This has the risk of having an `undefined variable` error.

controllers/grid/users/reviewer/form/ThankReviewerForm.php
```diff
@@ -164,4 +170,17 @@ public function execute(...$functionArgs)
 
         parent::execute(...$functionArgs);
     }
+
+    /**
+     * create a fake reviewer for invitations that not accepted yet
+     * @param $reviewAssignment
+     * @return User
+     */
+    private function fakeReviewer($reviewAssignment):User{
```
defstat: Haven't we declared this function again in the code?

pages/invitation/InitializeInvitationUIHandler.php
defstat: Is this resolved?

defstat (changes requested):
@asmecher @withanage I’ve gone through the codebase again.

Some earlier review comments remain unresolved. The main outstanding concerns are:

- Error-prone payload handling: there are cases where different payload sources interact in ways that could lead to inconsistent or unsafe state.
- Invitation-specific logic in core code: there appears to be invitation-specific handling inside the core layer (possibly limited to a single function), which should be refactored out.
- Invitation typing: in some places, invitation-specific logic returns generic invitation types instead of the expected concrete types.
- Request data handling: there are several usages of `request->getUserVars(...)` without clear, explicit validation or sanity checks at the business-logic level.

I believe these are mostly localized issues, but they should be addressed before this can be merged.

classes/submission/reviewer/ReviewerAction.php
withanage: done

controllers/grid/users/reviewer/form/ReviewerNotifyActionForm.php
withanage: done

controllers/grid/users/reviewer/form/ReviewerNotifyActionForm.php
withanage: done

controllers/grid/users/reviewer/form/ThankReviewerForm.php
withanage: @defstat  I have fixed some of the issues together and @ipula  merged into this  repo for easier review.

FYI @asmecher 


- https://github.com/pkp/pkp-lib/pull/11821/commits/1b667e691f24dbd2992e82e0fbfed22020aee9ac
- https://github.com/pkp/pkp-lib/pull/11821/commits/6e6b2f12c0e214f1013b9a22f5c9a6b1ddef5b61


There  is one  topic , which is unclear to me.


> Invitation-specific logic in core code: there appears to be invitation-specific handling inside the core layer (possibly limited to a single function), which should be refactored out.

This seems  a refacotring needed. Can you elaborate more, what you mean here ?


Rest, I can work on!

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
ipula: For new reviewer the only mandatory field is email. We cannot pass any other details. Here we have to send an email to this new reviewer (did not create account yet) but in email mailable we have to pass names and we do not know any other details other than email

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
ipula: done

classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteRedirectController.php
ipula: I change it to

```$templateMgr->assign('invitation', $this->getInvitation());
        $userId = $this->getInvitation()->invitationModel->userId;
        $user = $userId ? Repo::user()->get($userId) : null;
        $steps = $this->getAcceptSteps($this->invitation, $context, $user);```

File: classes/invitation/repositories/Repository.php (hunk 51 above)
Comment (ipula) [reply classes/invitation/repositories/Repository.php]: I moved this to reviewer repository

File: classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php (hunk 37 above)
Comment (ipula) [reply classes/invitation/invitations/reviewerAccess/handlers/ReviewerAccessInviteUIController.php]: I updated code here check now

File: classes/invitation/invitations/reviewerAccess/resources/BaseReviewerAccessInviteResource.php (hunk 40 above)
Comment (ipula) [reply classes/invitation/invitations/reviewerAccess/resources/BaseReviewerAccessInviteResource.php:103]: this is very tricky. Same UI components use for both reviewer and user invitations. If reviewer invitation response payload changed I have change the UI and user invitation. Also we do not need all the details of the submission . We only need `id, title` and `abstract`. So i kept in one response just to keep UI as same it is

File: classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php (hunk 42 above)
Comment (ipula) [reply classes/invitation/invitations/reviewerAccess/resources/ReviewerAccessInviteResource.php:41]: I made the changes please have a look

File: classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php (hunk 50 above)
Comment (ipula) [reply classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php]: I created a separate function inside `InvitePayload` class

```
public function encryptPasswordIfNeeded(): void
    {
        if (isset($this->username) && isset($this->password) && !($this->passwordHashed ?? false)) {
            $this->password = Validation::encryptCredentials($this->username, $this->password);
            $this->passwordHashed = true;
        }
    }
```

Issue #11701, opened 2025-08-05
Build REST API for the Discussion notes/replies


Vitaliy-1:
**Description:**
Users should be able to create, access, edit and delete notes/replies for tasks and discussions. Add this basic functionality to the API alongside with already existing endpoints for interactions with tasks. 

What to consider:
1. Notes can be retrieved together with associated task through the GET endpoint. Just adding them to the Task Resource could be enough. See https://github.com/pkp/pkp-lib/blob/ecd9073b54dbbd45245cc4e0bb9332e16e14d546/api/v1/submissions/PKPSubmissionController.php#L399 and https://github.com/pkp/pkp-lib/blob/ecd9073b54dbbd45245cc4e0bb9332e16e14d546/api/v1/submissions/resources/TaskResource.php#L21
2. Add, edit and remove operations can be added as a separate endpoints.
3. Probably, auth policies for tasks might be re-used for this case.

**What application are you using?**
  OJS, OMP and OPS

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/11828
OJS: https://github.com/pkp/ojs/pull/5094
OMP: https://github.com/pkp/omp/pull/2126
OPS: https://github.com/pkp/ops/pull/1102

Vitaliy-1:
@touhidurabir, do you have time to review and this part for the new API?

PR #11828, opened 2025-09-15
pkp/pkp-lib#11701 create API for discussion notes

Vitaliy-1:


api/v1/submissions/formRequests/EditTask.php
```diff
@@ -73,13 +73,20 @@ public function rules(): array
...
+                    // No need to check responsible for discussions
+                    if ($this->input('type') == EditorialTaskType::DISCUSSION->value) {
+                        return true;
+                    }
+
                     $responsibles = array_filter(Arr::pluck($this->input('participants'), 'isResponsible'));
-                    if (count($responsibles) > 1) {
-                        return $fail('There should be the only one user responsible for the task');
+                    if (count($responsibles) != 1) {
+                        return $fail('There should be one user responsible for the task');
```
touhidurabir: should we have localized error messages here as this will be return back as API validation error and we do have localized error message API, just wondering .

api/v1/submissions/resources/NoteResource.php
```diff
@@ -0,0 +1,52 @@
...
+
+class NoteResource extends JsonResource
+{
+    use ResourceWithData;
+
+    public function toArray($request)
+    {
+        [$users] = $this->getData('users');
+
+        $user = $users->get($this->user_id); /** @var User $userId */
```
touhidurabir: Is this used here ?

classes/editorialTask/EditorialTask.php
```diff
@@ -230,9 +177,29 @@ public function scopeWithStageId(Builder $query, int $stageId): Builder
...
      */
-    public function scopeWithClosed(Builder $query, bool $closed): Builder
+    public function scopeWithOpen(Builder $query, bool $isOpen): Builder
     {
-        return $query->where('closed', $closed);
+        if (!$isOpen) {
+            return $query;
+        }
+
+        return $query->whereNotNull('date_closed');
```
touhidurabir: If I understand correctly, this is to filter the data based on when `date_closed` not NULL . so do we need this `$isOpen` as it's just return the query ? we can just have it to apply the where rule when added .

classes/editorialTask/EditorialTask.php
```diff
@@ -230,9 +177,29 @@ public function scopeWithStageId(Builder $query, int $stageId): Builder
...
+     * @param string $orderBy One of the ORDERBY_* constants.
+     * @param string $direction One of the ORDER_DIR_* constants.
+     */
+    public function scopeOrderBy(Builder $query, string $orderBy, string $direction = self::ORDER_DIR_ASC): Builder
+    {
+        return match ($orderBy) {
+            self::ORDERBY_DATE_CREATED => $query->orderBy('created_at', $direction),
+            self::ORDERBY_DATE_STARTED => $query->orderBy('date_started', $direction),
+            self::ORDERBY_DATE_DUE => $query->orderBy('date_due', $direction),
+            default => $query,
```
touhidurabir: how about just let it throw exception when no match found as if someone provide some wrong order case, they will know it instead to silently returning ?

classes/editorialTask/EditorialTask.php
```diff
@@ -260,4 +227,143 @@ public function scopeWithAssocType(Builder $query, int $assocType): Builder
...
+    protected function users(): Attribute
+    {
+        return Attribute::make(
+            get: function () {
+                $userIds = $this->participants()
+                    ->pluck('user_id')
+                    ->all();
+                return Repo::user()->getCollector()->filterByUserIds($userIds)->getMany();
+            },
+        );
```
touhidurabir: any downside having `shouldCache` here so that in the same request cycle if this accessed multiple time, only the first time this need to generate from DB . Just a small improvement but if this update in the same request cycle and then needed to refetch on same model instance , better not to do that .

classes/migration/install/SubmissionsMigration.php
```diff
@@ -257,6 +257,8 @@ public function up(): void
             $table->dateTime('date_due')->nullable();
             $table->bigInteger('created_by')->nullable()->default(null);
             $table->foreign('created_by')->references('user_id')->on('users');
+            $table->bigInteger('started_by')->nullable()->default(null);
+            $table->foreign('started_by')->references('user_id')->on('users')->cascadeOnDelete();
```
touhidurabir: Here even though we make it nullable , still defining rule to delete the row as `cascadeOnDelete` . so are we not going to preserve the task when the user who started get deleted ? I am not sure about the intended workflow but is it not possible to have a task ongoing even when the user who started it got removed from the system ?

touhidurabir (commented):
@Vitaliy-1 just few minor comments . Please take a look . 

BTW, one general point, our `PKPSubmissionController` controller now has too many routes, methods and logics and file contains over 2500 lines of code . wondering if it is better to move some of these specific routes like task to a separate controller file . I will have a bit of maintenance overhead like managing the `index.php` for all apps and even though all of these are co-related to submission, perhaps having separate controller file will make it easier to manage ? we have similar implementation for jats , reviewer suggestions and submission files .

classes/migration/install/SubmissionsMigration.php
Vitaliy-1: Thanks, that was mistakenly added

Issue #11833, opened 2025-09-17
Implement API endpoint to list task templates


Hafsa-Naeem:
**Description:**
This complements the POST /editTaskTemplates endpoint

**What application are you using?**
  OJS, OMP or OPS main branch

Hafsa-Naeem:
@Vitaliy-1 Ready for review
OJS: https://github.com/pkp/ojs/pull/5113
pkp-lib: https://github.com/pkp/pkp-lib/pull/11834

Vitaliy-1:
Thanks, @Hafsa-Naeem, added quick comments

Hafsa-Naeem:
@Vitaliy-1 Ready for review
OJS: https://github.com/pkp/ojs/pull/5113
pkp-lib: https://github.com/pkp/pkp-lib/pull/11834

Vitaliy-1:
Thanks, looks good to merge!

PR #11834, opened 2025-09-17
pkp/pkp-lib#11833 added API endpoint to get task templates

Hafsa-Naeem:
for #11833


api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,95 @@
...
+            ]),
+        ])->group(function () {
+            Route::get('', $this->index(...));
+        });
+    }
+
+    /**
+     * GET /api/v1/editTaskTemplates
+     */
+    public function index(ListTaskTemplates $illuminateRequest): JsonResponse
```
Vitaliy-1: It's better to rename to getMany()

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $context = $this->getRequest()->getContext();
+        $validated = $illuminateRequest->validated();
```
Vitaliy-1, line 85: I don't think we need any validation here. It's a get request for the list of templates

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,95 @@
...
+            $q->where('include', (bool) $validated['include']);
+        }
+
+        if (!empty($validated['emailTemplateKey'])) {
+            $et = Repo::emailTemplate()->getByKey($context->getId(), $validated['emailTemplateKey']);
+            $q->where('email_template_id', $et?->getId() ?? 0);
+        }
+
+        $items = $q->orderByPkDesc()->get()
+            ->map(fn ($tpl) => (new EditTaskTemplateResource($tpl))->toArray($illuminateRequest))
```
Vitaliy-1: We need to pass a collection of templates instead of a single resource:
```php
EditTaskTemplateResource::collection(Template::query()...->get())
```

api/v1/editTaskTemplates/formRequests/ListTaskTemplates.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace PKP\API\v1\editTaskTemplates\formRequests;
+
+use APP\core\Application;
+use APP\facades\Repo;
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
+
+class ListTaskTemplates extends FormRequest
```
Vitaliy-1: We can just re-use already created class that extends FormRequest, and get requests usually don't require any validation. It's relevant for operations like create and edit

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * GET /api/v1/editTaskTemplates
+     */
+    public function getMany(Request $request): JsonResponse
+    {
+        $context = $this->getRequest()->getContext();
+
+        $q = Template::query()
```
Vitaliy-1:
```php
It's better to use our already established pattern and scopes to avoid snake case props
$collector = Template::byContextId((int) $context->getId())->with('userGroups');

foreach ($request->query() as $param => $val) {
      switch ($param) {
          case 'stageId':
              $collector = $collector->withStageId((int) $val);
              break;
           ...   
      }
}
```

Vitaliy-1:
@Hafsa-Naeem, is any help here required? Seems like a small tweak an it's ready to go.

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -90,4 +93,50 @@ public function add(AddTaskTemplate $illuminateRequest): JsonResponse
...
+    {
+        $context = $this->getRequest()->getContext();
+
+        // Start with our standard collector/scopes
+        $collector = Template::byContextId((int) $context->getId())
+            ->with('userGroups');
+
+        // Apply supported filters from query params via model scopes
+        foreach ($request->query() as $param => $val) {
+            switch ($param) {
```
Vitaliy-1, line 129: Can you take a look at the example of the params filtering in the submission controller: https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/PKPSubmissionController.php#L473?

api/v1/editTaskTemplates/resources/EditTaskTemplateResource.php
```diff
@@ -0,0 +1,48 @@
...
+ * @brief Transforms an editorial task template into an API response format.
+ *
+ */
+
+namespace PKP\API\v1\editTaskTemplates\resources;
+
+use Illuminate\Http\Request;
+use Illuminate\Http\Resources\Json\JsonResource;
+
+class EditTaskTemplateResource extends JsonResource
```
Vitaliy-1: This is already added in the main: https://github.com/pkp/pkp-lib/blob/main/api/v1/editTaskTemplates/resources/TaskTemplateResource.php

Vitaliy-1 (changes requested):
Thanks! I've added a couple of comments

PR #11846, opened 2025-09-20
ORCID: Handle authors accessing authorize link multiple times

kennydude:
This PR covers a couple of issues we have faced around where authors were clicking multiple times on the "Authorize" link within emails sent for ORCIDs.

* There was a fatal error upon trying again as it would try to still complete the action with no author available
* There was a confusing error message where the author thought the action was not complete, yet it was already completed, so we have added an ability to the page to display a different error message in this situation.


asmecher:
@ewhanson, could you have a look at this one?

ewhanson:
Thanks for the pull request, @kennydude! Looks good.

I'll apply this to 3.5 as well.

Issue #11826, opened 2025-09-14
[Frontend] Settings - Tasks & Discussions Templates


blesildaramirez:
## Description:
Implement UI for managing the tasks and discussions templates in the settings area.

### Table/List
- [x] Group entries by stage: 
  - [x] OJS: Submission, Review, Copyediting, Production
  - [x] OMP: Submission, Internal Review, External Review, Copyediting, Production
  - [x] OPS: Production
- [x] Each group has an Add Template button
- [x] Columns
  - Name
  - Automatically add this task and discussion when a submission reaches a specific stage (toggle, confirm modal when updating)
  - Actions:
    - Edit 
    - Delete (confirm modal before deleting)

### Adding/Editing a Task/Discussion Template
- [x] Clicking Add or Edit opens a side modal form 
- [x] Fields:
  - Details group:
    - Name (required)
    - ~~Participants~~ _removed because automatic task creation often happens before all stage participants are assigned, making the auto-assignment unreliable. Better to let users handle it manually._
    - Mark as unrestricted
      - Mark as unrestricted (default)
      - Limit access to specific roles
    - Limit access to specific roles (required if "Limit access to specific roles" is selected) 🚀 
      -  ~~There's a mapping issue with the selected `userGroups` for older template records from other stages: https://github.com/pkp/pkp-lib/issues/12197~~
  - Task Details:
    - Enter task information checkbox
    - Due Date (required if task information is entered)
  - Discussion: 
    - Message (required)
      - Insert Content (allows selecting variables to use for the email template)
  - Automatically add this task and discussion when a submission reaches a specific stage (checkbox)
- Validations:
  - [x] Cancel/close warns on unsaved changes
  - [x] When creating a submission task/discussion, the template should only appear for users whose roles are included in the “Limit access to specific roles” list.
  - [x] Automatic addition will be handled in the backend: https://github.com/pkp/pkp-lib/issues/11993 🚀

**What application are you using?**
  OJS, OMP or OPS main branch

**Additional information**
Mockups can be found in this link: https://miro.com/app/board/uXjVLjmM3rU=/?moveToWidget=3458764621517837735&cot=14

---
### Pull requests
Development Updates
Branch name used: blesildaramirez:i11826
✅ = merged
⌛ = work in progress
📌 = ready for code review
❌ = tests PR only

**Pull requests**
ui-library:
✅ https://github.com/pkp/ui-library/pull/703 Add task template manager components
[...]

PR #11879, opened 2025-09-29
pkp/pkp-lib#11826 [Frontend] Settings - Tasks & Discussions Templates

blesildaramirez:
- Add translation keys
- Add "Tasks and Discussions" tab to the Workflow Settings tabs

### DIFF
--- a/locale/en/manager.po
+++ b/locale/en/manager.po
@@ -3755,6 +3755,9 @@ msgstr "Validate XML before the export and registration."
 msgid "manager.selectCategories"
 msgstr "Select Categories"
 
+msgid "manager.template.tasksAndDiscussions"
+msgstr "Tasks and Discussions"
+
 msgid "manager.userComment.deleteReport"
 msgstr "Delete Report"
 


locale/en/common.po
```diff
@@ -739,6 +745,12 @@ msgstr "or"
 msgid "common.other"
 msgstr "Other"
 
+msgid "common.oneWeek"
+msgstr "1 week"
```
jardakotesovec, line 749: @asmecher Any suggestion how we go about pluralisation? As its somewhat language specific.

asmecher: @jardakotesovec Here's how it's supposed to work in PO files:

https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

We haven't used these formulations in OJS, but since we're now using Laravel for PO file parsing, it should be supported, I think.

blesildaramirez: I will merge this PR for now, and address the pluralization on a separate PR.

Issue #11690, opened 2025-07-30
[Invitation] | Invitation request auto declined on opening email due to mail client's safe link feature


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The invitation email of user invitation process get auto declined when mail being open first time . This is only happening with mail client with safe link or similar security feature . 

At the opening to invitation email, mail client sent a `HEAD` request to any link within the mail content if it has safe link or similar security feature enable . As the invitation email contains both the accept and decline link and the decline process does not prompt for any confirmation but immediately decline it on loading that url, it get executed and declined immediately . 

### Steps to Reproduce

1. Login as admin/JM .
2. Sent a invitation form `Settings --> Users & Roles --> invite to a role`  and complete the invitation process .
3. Open the invitation email in a mail client or service with safe link or similar security feature
4. Try to accept the invitation and see `404 not found` error as invitation has been declined immediately on opening the mail 
5. Optionally, check the `invitations` table of DB , see that invitation status is set to `DECLINED` . 

### Expected Result

No such request to sent and user can perform the desired action . 

### Actual Result

On opening the invitation email , it get auto declined if the mail client has safe link or similar security feature enabled . 

### Environment Details

To test this behaviour , required a mail client that has safelink or similar security feature enabled . for example , MS mail client . 

### Application Version

OJS/OMP/OPS 3.5/main 

### Additional Information

Similar issue that has been addressed at https://github.com/pkp/pkp-lib/issues/8392 . 

Possible quick solution would to have a quick check in the `PKP\pages\invitation\InvitationHandler::decline` to check if `HEAD` request at the very beginning and return back without running further actions . 

However , the best solution probably would be to have a prompt or confirmation from user to decline the invitation as even in the absence of safe link feature , user may mistakenly click on the decline invitation link . 

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11880
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5111 [TEST ONLY]

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11881
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5112 [TEST ONLY]

Tribunal33:
[...]

PR #11880, opened 2025-09-29
pkp/pkp-lib#11690 resolve invitation auto decline using confirmation landing page

touhidurabir:
for pkp/pkp-lib#11690 .


ewhanson (approved):
Thanks @touhidurabir. Looks good to me!

Issue #11758, opened 2025-08-28
[Upgrade to >= 3.4] | I8866_DispatchRegionCodesFixingJobs - FixRegionCodes Job


bozana:
Update (07.04.26): for the final solution see https://github.com/pkp/pkp-lib/issues/11758#issuecomment-4200433061

Update (31.0326): for the final solution see https://github.com/pkp/pkp-lib/issues/11758#issuecomment-4118566197

<hr>

### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

S. the messages in this issue, from here on: https://github.com/pkp/pkp-lib/issues/8866#issuecomment-3191824860
The lib/pkp/lib/regionMapping.php contains fixes for the old region codes. For some regions, the fix cannot be applied twice, because the fixed region code would actually be seen as the code to be fixed, in some cases.
If the job FixRegionCodes fails, it will be re-tried again, which could then lead to invalid region codes in the DB.
Thus, a new solution is needed, that inserts a marker, e.g. a prefix for the codes to be fixed (so that only this entries with the marker are considered for update) and that additionally splits the lines to update into chunks (so that no timeout happens in huge installations). 

The new solution uses smaller and chained jobs, that can be re-tried. All the region that needs to be fixed are marked with the prefix 'pkp-'. Then the regions containing the prefix are fixed. The jobs are per country and 100000 rows.

### Steps to Reproduce

1. Have entries in the DB table metrics_submission_geo_daily and/or metrics_submission_geo_monthly with some of those region codes, e.g.
country = 'KH', region = '09'
2. run the FixRegionCodes job
3. the region should now be '10'
4. run the FixRegionCodes job
5. the region is now '11', which is wrong, it should be '10'

### Expected Result

After running the job several times the region code should still be correct. From example above: region should be '10'

### Actual Result

After running the job several times the region code is wrongly fixed again. From example above: region is '11'

### Application Version

OJS, OMP, OPS version >= 3.4

<hr>

## stable-3_4_0:
- https://github.com/pkp/pkp-lib/pull/11899
- https://github.com/pkp/ojs/pull/5454 (only submodule update)
- https://github.com/pkp/omp/pull/2290 (only submodule update)
- https://github.com/pkp/ops/pull/1252 (only submodule update)

## stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/12469
- ~~https://github.com/pkp/ojs/pull/5441~~ (only submodule update)

Fix 1:
[...]

PR #11899, opened 2025-10-06
pkp/pkp-lib11758 fix region codes

bozana:
s. https://github.com/pkp/pkp-lib/issues/11758


bozana:
The job RegionMappingTmpClear is maybe not needed, because the table is cleared before insertion in RegionMappingTmpInsert, but I left it for now. Please let me know what you think, if I should remove it...

classes/migration/upgrade/v3_4_0/jobs/CleanTmpChangesForRegionCodesFixes.php
```diff
@@ -0,0 +1,59 @@
...
+            }
+        });
+
+        // return the length of the region columns to varchar(3)
+        Schema::table('metrics_submission_geo_daily', function (Blueprint $table) {
+            $table->string('region', 3)->change();
+        });
+        Schema::table('metrics_submission_geo_monthly', function (Blueprint $table) {
+            $table->string('region', 3)->change();
+        });
```
jonasraoni: I think jobs shouldn't touch the structure of the database.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -31,8 +37,9 @@ class I8866_DispatchRegionCodesFixingJobs extends Migration
      */
     public function up(): void
     {
-        if (DB::table('metrics_submission_geo_monthly')->whereNotNull('region')->exists() ||
-            DB::table('metrics_submission_geo_daily')->whereNotNull('region')->exists()) {
+        if (DB::table('metrics_submission_geo_monthly')->where('region', '<>', '')->exists() ||
+            DB::table('metrics_submission_geo_daily')->where('region', '<>', '')->exists()) {
+
```
jonasraoni, line 35: GitHub doesn't allow me to comment on the lines below, but adding an index to the temporary table's columns might be helpful to performance.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -42,6 +49,14 @@ public function up(): void
                 });
             }
 
+            // temporary change the length of the region columns, becuase we will add prefix 'pkp-'
+            Schema::table('metrics_submission_geo_daily', function (Blueprint $table) {
+                $table->string('region', 7)->change();
+            });
+            Schema::table('metrics_submission_geo_monthly', function (Blueprint $table) {
+                $table->string('region', 7)->change();
+            });
```
jonasraoni: This sounds risky, if the job never gets completed, the database will be left with an unexpected structure.

The same for the indexes below.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -58,35 +73,45 @@ public function up(): void
...
-                        }
-                    });
-
-                    // drop the temporary table
-                    if (Schema::hasTable('region_mapping_tmp')) {
-                        Schema::drop('region_mapping_tmp');
-                    }
+            $jobs[] = new CleanTmpChangesForRegionCodesFixes();
+            Bus::chain($jobs)
+                ->catch(function (Throwable $e) {
```
jonasraoni: There should be some error logging here, the `Throwable` covers even syntax errors at included files.

classes/migration/upgrade/v3_4_0/jobs/CleanTmpChangesForRegionCodesFixes.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/jobs/CleanTmpChangesForRegionCodesFixes.php
+ *
+ * Copyright (c) 2022-2025 Simon Fraser University
+ * Copyright (c) 2022-2025 John Willinsky
```
jonasraoni: For new files we're keeping just the initial year

```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

classes/migration/upgrade/v3_4_0/jobs/RegionMappingTmpClear.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        // clear region_mapping_tmp table
+        DB::table('region_mapping_tmp')->delete();
+    }
+}
```
jonasraoni: This cleanup is covered by the `RegionMappingTmpInsert.php`, so this whole file can be removed.

classes/migration/upgrade/v3_4_0/jobs/RegionMappingTmpInsert.php
```diff
@@ -0,0 +1,61 @@
...
+            $insert = false;
+            if ($fips !== $iso) {
+                $insert = true;
+            }
+            if ($insert) {
+                DB::table('region_mapping_tmp')->insert([
+                    'country' => $this->country,
+                    'fips' => $fips,
+                    'iso' => $iso
+                ]);
```
jonasraoni: You can store the records in an array and do a batch insert (chunks of 1000).

classes/migration/upgrade/v3_4_0/jobs/FixRegionCodesDaily.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Create a new job instance.
+     */
+    public function __construct(int $startId, int $endId)
+    {
+        parent::__construct();
+        $this->startId = $startId;
+        $this->endId = $endId;
+    }
```
jonasraoni: As this isn't going to be extended, it can be `private` and it's possible to simplify the initialization

```suggestion
    /**
     * Create a new job instance.
     * $startId and $endId provide the range of metrics_submission_geo_daily_ids to consider for this update
     */
    public function __construct(private int $startId, private int $endId)
    {
        parent::__construct();
    }
```

classes/migration/upgrade/v3_4_0/jobs/FixRegionCodesDaily.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        // update region code from FIPS to ISP, according to the entries in the table region_mapping_tmp
+        // Laravel join+update does not work well with PostgreSQL, so use the direct SQLs
+        // daily
+        if (substr(Config::getVar('database', 'driver'), 0, strlen('postgres')) === 'postgres') {
```
jonasraoni: I think it should be possible to build this update and the others using the Laravel query builder.

classes/migration/upgrade/v3_4_0/jobs/CleanTmpChangesForRegionCodesFixes.php
asmecher: @jonasraoni, in principle I agree with you -- but I'm not sure I can think of a better alternative. Our options (as I see them) are...
- Move this back into the upgrade script -- at the potential cost of the upgrade taking an extremely long time to run processing old logs that are not critical to the system's ongoing operation. (Shortening the upgrade was the original impetus for doing it here.)
- Keep this as a job, but make sure there are good safeguards against double-execution -- e.g. use the existence of an extra index or prefix as an indication that the job has already been tried (and perhaps failed), like a mutex.
- Keep this as a job, but find a way to do it without modifying the schema. Unfortunately I couldn't see a way to do this.

jonasraoni: Do we know how long it may take to run this script on a really heavy installation? If it's not that long, I think it's ok to keep it as a standard upgrade script.
Otherwise we have to make it very clear to the user that part of the upgrade has been deferred and that they must take a look at the status of post-upgrade jobs.

jnugent: My concern with these queued jobs and background tasks, especially if they are upgrade related, are with installations hosted on potentially less than ideal hosting solutions where users do not have access to server side logs or shell access. Obviously we do not recommend these sorts of scenarios but given how complex this is, even for us, I can imagine users wondering what's up.

As an aside, what does this region code bug cause? If these scripts fail to run, how does one know? Are stats misattributed or not available?

jonasraoni: I started the review with the opinion that this should first be turned into an upgrade script, I'll take a look again and check if I have ideas to share (regarding keeping it as a job).

For now:
1. I'll assume that users with a failed job will probably have to live with the corrupted data, I mean, it's already too late to revert (the file was created in 03/2023), and looks like reconciling is not possible, right?
2. In general, I think we should drop "default timeouts" from jobs... All tasks are supposed to be completed, unless there's an infinite loop (which I prefer to handle with another type of protection)... Following this idea, a job that can be retried as much as it needs (as long as there's progress), until the task is completed sounds nice, instead of just creating one gazillion jobs that will consume the whole processing.
3. While the job is running the updates, more entries might be inserted by the application, so we need to ensure new data isn't going to be affected/corrupted.

bozana: Hi all, this jobs/upgrade script changes the old region codes into new ones. I do not think that lots of users are using regions -- I think they rarely want to have that detailed usage statistics. The tracking of such usage is also resource intensive, so I would even think that the installations with not so much capacity (resources and personal) would not use them. (I am also wondering why don't we remove city and region level for usage stats -- I believe having country is enough for most of users, no?).
I introduced the bug, with not considering that the jobs cannot be re-run. These users, who already applied it and the jobs failed (probably only with big installations running jobs within a web request), will have some corrupt data -- wrong region codes. But, it is maybe not SUCH a big deal.
I now changed the jobs so that they are run only on part of the DB rows (in batches), so that no timeout occurs for huge installations. I also change the jobs so that the prefix is added to the regions that need to be changed, so this will assure that new regions are not affected. And the jobs are chained, so that they cannot run in parallel, so that the temporary table can not grow, and lead to timeouts.

I do not fully understand No. 2 from Jonas. I think this belongs somewhere else, how we generally deal with jobs?

Because this upgrade is probably not so important for most of the users, and the upgrade to 3.4 lasted in some of Alec's tests hours, we decided to put it into jobs back then.

So I believe, these changes make it pretty safe -- that the jobs would not fail (because of the timeout) and can be re-run. So I believe, it would be good to review it all having this in mind -- to be sure that this is so. And, in that case, why not leave it in jobs? Else, the only possibility would be to move it back to the normal process, but because it is so 'unimportant' this maybe does not make soooo much sense now -- but this decision I would need to leave to you...

bozana: > As an aside, what does this region code bug cause? If these scripts fail to run, how does one know? Are stats misattributed or not available?

The metrics will then exist for the region 01 instead for region 10 (not real region codes, just an example). And because we do not do anything with the regions and their metrics, I do not think anybody will notice that difference. Also, not all regions are affected, just some.
Generally, we currently do nothing with Geo statistics. They are only visible in TSV reports.
The numbers for the countries will still be valid.
And the numbers for article usage that we show on the UI, as well for COUNTER reports are all correct, because Geo stats are fully separate, in separate DB tables.

classes/migration/upgrade/v3_4_0/jobs/FixRegionCodesDaily.php
bozana: Back then, some years ago, I tried but it was not working with inner join, I will try again...

classes/migration/upgrade/v3_4_0/jobs/RegionMappingTmpInsert.php
bozana: The entries per country are not big, max 100, s. https://github.com/pkp/pkp-lib/blob/stable-3_4_0/lib/regionMapping.php, but I can prepare the array in PHP and provide it to insert()...

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
bozana: Do you think index would be needed even if the table only contains a few entries, lets say max 100 ?

classes/migration/upgrade/v3_4_0/jobs/FixRegionCodesDaily.php
jonasraoni: Hmm, I remember about something like that... The `updateFrom()` should probably help you, something like:

```php
if (DB::connection() instanceof PostgresConnection) {
    $updateQuery->updateFrom($update);
} else {
    $updateQuery->update($update);
}
```

classes/migration/upgrade/v3_4_0/jobs/RegionMappingTmpInsert.php
jonasraoni: I'm re-reviewing the PR with more attention (the other time I just skimmed), so nevermind, I already have a suggestion for the `regionMapping.php` for example.

bozana:
Hi @jonasraoni and @asmecher,

I considered most of the review comments from Jonas -- they can be seen in the last commit.
The things I have not considered:
1) We are changing the DB here and if a job fails it could be that the DB stays in that state. I increase the VARCHAR value for the column region, and I add temporary indexes. I do not know how to best solve it, if it is solvable at all.
Maybe we can increase that region VARCHAR however, so I do not need to change and revert it here?
Regarding indexes: I am adding an index ['country', 'region']. I think it is not necessarily needed else, because we currently do not search in that matter.
I think I cannot clean it on job failure because admin could re-run it, and this would need the changes. Also, if prefix is already inserted for regions that needs to be updated, the column size should not be resized again.
Would you have any other idea/solution?
2) I have not added an index to region_mapping_tmp because the table does not contain much entries, maybe max. 100, and they change (country by country). But if you think I should -- because it is then joined with a big table -- I will do.
3) Jonas, you wrote that you would have a suggestion for the regionMapping.php?

I would really like to finish this issue, so it can be merge, so I would love to hear your comments how to proceed next.

Thanks a lot!

bozana:
Hi @jonasraoni, could you please take a new look into changes and comments. We should consider this as soon as possible, and for the next 3.5 release.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
bozana: Maybe to have all countries then in the temp table and the index. It would also create less jobs.
I will try this solution...

bozana:
Added a few more improvements, so in total:

**Reduced job chain size**
Previously, the job chain dispatched a RegionMappingTmpInsert + full set of PreFix/Fix chunk jobs for each of the 200 countries separately. All FIPS-ISO mappings are now loaded into the temporary table at once in a single job, followed by one set of PreFix/Fix chunk jobs.

**Pre-computed pkp_fips column**
A pkp_fips column (CONCAT('pkp-', fips)) is now stored in region_mapping_tmp with a composite index on (country, pkp_fips). This allows the Fix jobs to use a direct index lookup instead of computing CONCAT('pkp-', rm.fips) at runtime.

**Reduced batch size (100k → 10k)**
The chunk size was reduced from 100,000 to 10,000 rows to safely stay within the 60-second job timeout defined in BaseJob, even on slower installations.

**Safe retry**
RegionMappingTmpInsert now checks exists() before inserting — since the bulk insert is a single atomic SQL statement, the table is either fully populated or empty, so this check reliably prevents duplicate inserts on retry without needing to clear the table first.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
bozana: We have no better options I think. 
The leftover schema structures seem harmless — varchar(7) still holds all valid ISO codes (max 3 chars), the temporary indexes add minor overhead but cause no errors, and the region_mapping_tmp table is just an extra empty table.
If the chain stops, the admin can retry the failed job directly from the admin interface. Since all jobs are now idempotent, the chain resumes correctly from where it stopped without any manual intervention.

bozana:
maybe these additional things to decide on:
the chunk/batch size of PreFix/Fix jobs -- currently 10000
shell we send an email to the admin if a job fails

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -31,18 +35,30 @@ class I8866_DispatchRegionCodesFixingJobs extends Migration
      */
     public function up(): void
     {
-        if (DB::table('metrics_submission_geo_monthly')->whereNotNull('region')->exists() ||
-            DB::table('metrics_submission_geo_daily')->whereNotNull('region')->exists()) {
+        if (DB::table('metrics_submission_geo_monthly')->where('region', '<>', '')->exists() ||
+            DB::table('metrics_submission_geo_daily')->where('region', '<>', '')->exists()) {
```
jonasraoni: I think early returns are easier to read, so I'd modify it to:

```suggestion
        if (!DB::table('metrics_submission_geo_monthly')->where('region', '<>', '')->exists() &&
            !DB::table('metrics_submission_geo_daily')->where('region', '<>', '')->exists()) {
            return;
        }
```

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -31,18 +35,30 @@ class I8866_DispatchRegionCodesFixingJobs extends Migration
...
-                    $table->string('iso', 3)->nullable();
+                    $table->string('pkp_fips', 7);
+                    $table->string('iso', 3);
+                    $table->index(['country', 'fips']);
+                    $table->index(['country', 'pkp_fips']);
                 });
             }
 
-            // temporary create index on the column country and region, in order to be able to update the region codes in a reasonable time
+            // temporary change the length of the region columns, because we will add prefix 'pkp-'
```
jonasraoni: Instead of increasing the field length to include a prefix (mark which records need to be updated), we can:
- Add a not null column `needs_review` with a default value of 1 (all records will instantly contain the value 1)
- Right after creating the column, change its default value to 0. This way, possible new records that get inserted, while the system is running, can be easily filtered out by the jobs.

Why:
- I think it's less disruptive (we won't modify user data)
- Changing a status column should be faster than updating strings
- We'll decrease the work by half (records will be updated once instead of twice)

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -58,42 +74,51 @@ public function up(): void
...
+            $chunkSize = 10000;
+            $geoDailyChunksNo = ceil($geoDailyIdMax / $chunkSize);
+            $geoMonthlyChunksNo = ceil($geoMonthlyIdMax / $chunkSize);
 
-                    // drop the temporary table
-                    if (Schema::hasTable('region_mapping_tmp')) {
-                        Schema::drop('region_mapping_tmp');
-                    }
+            // load all FIPS-ISO mappings into the temporary table, then dispatch chunk jobs
+            $jobs = [new RegionMappingTmpInsert()];
```
jonasraoni: I've checked the underlying code of the Laravel jobs and it's possible to emulate loops with something like:

- Setup the $tries attribute to 0
- Do some work (e.g. update 1000 records)
- Call `$this->release()` to restart the job if the work isn't over (it will send the job back to the end of the queue, and optionally accepts a delay for extra "breathing"). Alternatively we could `dispatch()` the same job, which should have the same effect...

Then the creation of these series of jobs (given the number of records is undefined, perhaps series of jobs will clog the queue) in a chain could be possibly turned into a single job, that would probably have something like this:

```sql
UPDATE metrics_submission_geo_daily gd
INNER JOIN region_mapping_tmp rm ON ...
SET
  needs_review = 0, 
  region = rm.iso
WHERE gd.needs_review = 1
LIMIT 1000 -- I guess this doesn't work in PostgreSQL, to limit the number of updated records, anyway it's feasible to build a query that works for both
```

And...

```sql
UPDATE metrics_submission_geo_monthly gm
INNER JOIN region_mapping_tmp rm ON ...
SET
  needs_review = 0, 
  region = rm.iso
WHERE gm.needs_review = 1
LIMIT 1000
```

The job must be redispatched whenever at least one of the queries finds something to update, as we have the `needs_review` column, the job can be retried... Once there's no more records to process, we can finally do the cleanup, which doesn't need to be in a separated job (it should run instantly).

jonasraoni (changes requested):
@bozana Sorry for the delay, I was lost here 👀

I just needed to do a small test, to check the feasibility of having a job running in a loop, I see it works fine, so I left some suggestions based on it and some extra observations, check if it makes sense.

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
bozana: There seem to be one problem here: PostgreSQL 9.5–10 and MySQL 5.7 rewrite the entire table (require a full table rewrite) when adding a NOT NULL DEFAULT 1 column, which could be very slow for huge metrics DB tables.

bozana: So maybe:
Use NULL as the "needs processing" marker instead of 1:                                                                                                                                                                                                     
                                                                                                                                                                                                                                                              
1. Migration:                                                                                                                                                                                                                                               
    - ADD COLUMN needs_review smallInteger NULL                                                                                                                                                           
    - ALTER COLUMN needs_review SET DEFAULT 0                                                                                                                                                                                 
    - Existing rows: NULL                                                                                                                                                                                                              
    - New rows inserted during processing: get 0 automatically                                                                                                                                                                        
                                                                                                                                                                                                                                                              
2. Then dispatch two chained jobs:                                                                                                                                                                                                                          
    - RegionMappingTmpInsert (as now, populates the mapping table)
    - Self-requeuing FixRegionCodes job: processes WHERE needs_review IS NULL LIMIT 1000, sets needs_review = 0 after converting, calls release() if more rows remain. When done, runs cleanup inline (drops column, temp table, indexes)                     
                                                                                                                                                                                                                                                            
Re-runnability: converted records have needs_review = 0, failed mid-batch records still have NULL and get picked up on retry.

bozana:
Hi @jonasraoni, I implemented your suggestions. Could you please take a look?
In the meantime I will test it...
Thanks a lot!

classes/migration/upgrade/v3_4_0/I8866_DispatchRegionCodesFixingJobs.php
```diff
@@ -31,69 +30,61 @@ class I8866_DispatchRegionCodesFixingJobs extends Migration
...
-            $mappings = include Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/regionMapping.php';
-            $jobs = [];
-            foreach (array_keys($mappings) as $country) {
-                $jobs[] = new FixRegionCodes($country);
+        // add needs_review column: NULL marks existing records that need processing,
+        // default 0 ensures new records inserted during processing are excluded
+        foreach (['metrics_submission_geo_daily', 'metrics_submission_geo_monthly'] as $table) {
+            if (!Schema::hasColumn($table, 'needs_review')) {
+                Schema::table($table, function (Blueprint $t) {
+                    $t->smallInteger('needs_review')->nullable();
```
jonasraoni, line 54: I think the `needs_review` should be created as `NOT NULL` and with default value of `1`, then all records will receive this initial value.
Right after that, the column should be turned into `NULLABLE`.

I think this should be done to cover new records that might be inserted while the application is online, then the old records will have `1`, the new records will have `NULL` and the reviewed records can be turned into `0`.

bozana: I decided to do this this way because: "PostgreSQL 9.5–10 and MySQL 5.7 rewrite the entire table (require a full table rewrite) when adding a NOT NULL DEFAULT 1 column, which could be very slow for huge metrics DB tables." (s. my comment above).

jonasraoni: I see, but did you check the condition I've mentioned? New rows being inserted while the jobs are being executed?

Then an alternative that probably is going to work is to go with a `NULL DEFAULT 0` to mark the new rows as ok.

PR #11908, opened 2025-10-07
Resolve duplicated validation response from new endpoints (task&discu…

jardakotesovec:
…ssions)


touhidurabir (approved):
Need back ported to stable-3.5.0 once tests passes .

PR #11915, opened 2025-10-10
I11902 fixes

GaziYucel:
@bozana

### DIFF
--- a/locale/en/submission.po
+++ b/locale/en/submission.po
@@ -609,7 +609,10 @@ msgid "submission.citations.structured.label.rawCitation"
 msgstr "Edit Raw Citation"
 
 msgid "submission.citations.structured.label.doi"
-msgstr "DOI, e.g. 10.1000/182, doi:10.1000/182, https://doi.org/10.1000/182"
+msgstr "DOI"
+
+msgid "submission.citations.structured.description.doi"
+msgstr "10.1000/182, doi:10.1000/182, https://doi.org/10.1000/182"
 
 msgid "submission.citations.structured.label.title"
 msgstr "Title"
@@ -657,10 +660,16 @@ msgid "submission.citations.structured.label.urn"
 msgstr "URN"
 
 msgid "submission.citations.structured.label.arxiv"
-msgstr "Arxiv, e.g. 1234.123456v2, arxiv:1234.123456v2, https://arxiv.org/abs/1234.123456v2"
+msgstr "Arxiv"
+
+msgid "submission.citations.structured.description.arxiv"
+msgstr "1234.123456v2, arxiv:1234.123456v2, https://arxiv.org/abs/1234.123456v2"
 
 msgid "submission.citations.structured.label.handle"
-msgstr "Handle, e.g. 20.1000/100, handle:20.1000/100, https://hdl.handle.net/20.1000/100"
+msgstr "Handle"
+
+msgid "submission.citations.structured.description.handle"
+msgstr "20.1000/100, handle:20.1000/100, https://hdl.handle.net/20.1000/100"
 
 msgid "submission.citations.structured.label.openAlex"
 msgstr "OpenAlex"
--- a/classes/components/forms/citation/CitationStructuredEditForm.php
+++ b/classes/components/forms/citation/CitationStructuredEditForm.php
@@ -49,14 +49,31 @@ class CitationStructuredEditForm extends FormComponent
         ]));
 
         // Article Information
-        foreach (['doi', 'url', 'urn', 'arxiv', 'handle'] as $key) {
-            $this->addField(new FieldText($key, [
-                'label' => __('submission.citations.structured.label.' . $key),
-                'description' => '',
-                'value' => null,
-            ]));
-        }
-
+        $this->addField(new FieldText('doi', [
+            'label' => __('submission.citations.structured.label.doi'),
+            'description' => __('submission.citations.structured.description.doi'),
+            'value' => null,
+        ]));
+        $this->addField(new FieldText('url', [
+            'label' => __('submission.citations.structured.label.url'),
+            'description' => '',
+            'value' => null,
+        ]));
+        $this->addField(new FieldText('urn', [
+            'label' => __('submission.citations.structured.label.urn'),
+            'description' => '',
+            'value' => null,
+        ]));
+        $this->addField(new FieldText('arxiv', [
+            'label' => __('submission.citations.structured.label.arxiv'),
+            'description' => __('submission.citations.structured.description.arxiv'),
+            'value' => null,
+        ]));
+        $this->addField(new FieldText('handle', [
+            'label' => __('submission.citations.structured.label.handle'),
+            'description' => __('submission.citations.structured.description.handle'),
+            'value' => null,
+        ]));
         $this->addField(new FieldText('title', [
             'label' => __('submission.citations.structured.label.title'),
             'description' => '',


classes/citation/pid/Arxiv.php
```diff
@@ -19,33 +19,20 @@
...
-
-    /** @copydoc AbstractPid::prefix */
-    public const prefix = 'https://arxiv.org/abs';
-
-    /** @copydoc AbstractPid::prefixInCorrect */
-    public const prefixInCorrect = [
-        'arxiv:'
+    public const regexes = [
+        '/arxiv:\s*\d{4}\.\d{4,5}(v\d+)?/i', // arxiv:2025.12345v2
+        '/https?:\/\/arxiv.org\/(?:abs|pdf)\/(?:\d+.\d+|[a-z-]+.\d+)/i' // https://arxiv.org/abs/2025.12345
```
GaziYucel: @bozana

In Wikidata I found this:
^https?://arxiv.org/(?:pdf|abs)/((?:\d{4}.\d{4,5}|[a-z-]+(?:.[A-Z]{2})?/\d{7})(?:v\d+)?)
For URLs.

It seems there were old identifier, that could contain letters and then 7 digits, s. https://info.arxiv.org/help/arxiv_identifier.html#old

Also there could be pdf instead of abs in the URL -- if someone wants to access/cite PDF directly

Maybe not need to consider www

@GaziYucel

This doesn't match for example this: https://arxiv.org/pdf/cond-mat.02534

I changed the regex a bit, this would work:
/https?://(?:www.)?arxiv.org/(?:abs|pdf)/(?:\d+.\d+|[a-z-]+.\d+)/i

Try with https://regex101.com/

arxiv:cond-mat.02534
arxiv:1902.02534
https://arxiv.org/abs/1902.02534
https://arxiv.org/pdf/cond-mat.02534
doi: 10.1111/j.1365-277X.2011.01184.x.
doi: 10.1002/tox.20155.
doi: 10.7748/phc.2016.e1162.
https://doi.org/10.3390/s23218689.
https://doi.org/10.7717/peerj.1990
https://doi.org/10.1037/arc0000014.
hdl:11366/1249
handle:11366/1249
http://hdl.handle.net/11366/1249
https://www.bildung-forschung.digital/files/BAnz%20AT%2017.06.2020%20B3-1.pdf

@bozana

shell we ignore www ?

classes/citation/pid/Doi.php
```diff
@@ -22,14 +22,23 @@
 class Doi extends BasePid
 {
     /** @copydoc AbstractPid::regex */
-    public const regex = '(10[.][0-9]{4,}[^\s"/<>]*/[^\s"<>]+)';
+    public const regexes = [
+        '/doi:\s*10[.][0-9]{4,}[^\s"\/<>]*\/[^\s"<>]+/i',
+        '/https?:\/\/doi\.org\/10[.][0-9]{4,}[^\s"\/<>]*\/[^\s"<>]+/i'
+    ];
```
GaziYucel, line 28: @bozana

s. https://www.crossref.org/blog/dois-and-matching-regular-expressions/

/^10.\d{4,9}/[-._;()/:A-Z0-9]+$/i

I am not sure if we need to consider the following as explained on that page:

/^10.1002/[^\s]+$/i

/^10.\d{4}/\d+-\d+X?(\d+)\d+<[\d\w]+:[\d\w]*>\d+.\d+.\w+;\d$/i

/^10.1021/\w\w\d++$/i

/^10.1207/[\w\d]+&\d+_\d+$/i

Why do we not consider URL (https://doi.org/) and name (DOI:) as for other identifiers?

classes/citation/pid/Handle.php
```diff
@@ -19,33 +19,22 @@
...
-    /** @copydoc AbstractPid::prefix */
-    public const prefix = 'https://hdl.handle.net';
-
-    /** @copydoc AbstractPid::prefixInCorrect */
-    public const prefixInCorrect = [
-        'handle:'
+    public const regexes = [
+        '/hdl:\s*\d+\/[a-zA-Z0-9\-_]+/i', // hdl:12345/abcde
+        '/handle:\s*\d+\/[a-zA-Z0-9\-_]+/i', // handle:12345/abcde
+        '/https?:\/\/hdl\.handle\.net\/\d+\/[a-zA-Z0-9\-_]+/i', // https://hdl.handle.net/12345/abcde
```
GaziYucel: @bozana 

According to Wikipedia: Handles consist of a prefix which identifies a "naming authority" and a suffix which gives the "local name" of a resource. ... The local name may consist of any characters from the Unicode UCS-2 character set. The prefix also consists of any UCS-2 characters, other than "/". The prefixes consist of one or more naming authority segments, separated by periods, representing a hierarchy of naming authorities.

The regex from Wikidata (https://www.wikidata.org/wiki/Property:P1184):
^https?://hdl.handle.net/([0-9a-z]+(?:.[0-9a-z]+)*/.{2,})

This regex does not consider all Unicode characters for the prefix, but the separation by periods and slightly more characters than the one you use.
However I think it can be only one character in the suffix, like in the example https://hdl.handle.net/10419/2.

I do not think we should consider www in the URL

classes/citation/pid/Urn.php
```diff
@@ -19,5 +19,7 @@
 class Urn extends BasePid
 {
     /** @copydoc AbstractPid::regex */
-    public const regex = '/urn:([a-z0-9][a-z0-9-]{1,31}):((?:[-a-z0-9()+,.:=@;$_!*\'&~\/]|%[0-9a-f]{2})+)(?:(\?\+)((?:(?!\?=)(?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2}))*))?(?:(\?=)((?:(?!#).)*))?(?:(#)((?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2})*))?$/i';
+    public const regexes = [
+        '/urn:([a-z0-9][a-z0-9-]{1,31}):((?:[-a-z0-9()+,.:=@;$_!*\'&~\/]|%[0-9a-f]{2})+)(?:(\?\+)((?:(?!\?=)(?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2}))*))?(?:(\?=)((?:(?!#).)*))?(?:(#)((?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2})*))?$/i'
+    ];
```
GaziYucel, line 24: @bozana

I found some simpler ones, but lets keep your complicated one :-)
e.g. from https://regex101.com/r/ddkvjd/1
urn:[A-Za-z0-9][A-Za-z0-9-]{0,31}:([A-Za-z0-9()+,-.:=@;$_!*']|%[0-9A-Fa-f]{2})+

From https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s06.html
^urn:[a-z0-9][a-z0-9-]{0,31}:[a-z0-9()+,-.:=@;$_!*'%/?#]+$

Maybe to leave comments everywhere where from do all the regex come

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2463,17 +2448,23 @@ protected function deleteCitationsByPublicationId(Request $illuminateRequest): J
...
+        foreach ($citations as &$citation) {
+            $citation->setIsProcessed(false);
+            Repo::citation()->edit($citation, []);
         }
+        unset($citation);
 
-        Repo::citation()->deleteByPublicationId($publication->getId());
+        foreach ($citations as &$citation) {
+            Repo::citation()->reprocessCitation($citation);
+            $citation = Repo::citation()->getSchemaMap()->map($citation);
```
bozana: This can be done in one, the previous foreach loop, I think

classes/citation/pid/Arxiv.php
bozana: Shouldn't the regex be the same, just either arxiv: or https://arxiv.org/abs ?
The first pattern would not match arxiv:cond-mat.02534

classes/citation/pid/BasePid.php
```diff
@@ -98,46 +87,17 @@ public static function normalize(?string $string): string
...
+            [$class::prefix, $class::prefix . ' '],
+            [$class::urlPrefix],
+            $class::alternatePrefixes,
+            array_map(fn($value) => trim($value) . ' ', $class::alternatePrefixes)
+        );
+        $prefixes = array_filter($prefixes, fn($value) => !empty(trim($value)));
+        $prefixes = array_unique($prefixes);
+        usort($prefixes, fn($a, $b) => strlen($b) - strlen($a));
+
+        return trim(str_ireplace($prefixes, '', $string), $class::defaultTrimCharacters);
```
bozana: why do we trim these default characters `.` and `/` ?
Would this DOI 10.123/tib123 then look like this 10123tib123 ?

classes/citation/pid/Handle.php
bozana: The regex does not consider e.g. [20.500.11794/34239](https://hdl.handle.net/20.500.11794/34239).

Maybe to use `[0-9a-z]+(?:.[0-9a-z]+)*\/.+`

classes/citation/pid/Doi.php
bozana: It seems this part in the prefix is not needed `[^\s"\/<>]*` -- why do we have it?

classes/citation/pid/BasePid.php
GaziYucel: Trim removes those only from start and the end. Sometimes there are raw citations which end with a `.` or `/`. It's to remove those. Thus `10.123/tib123.` would be `10.123/tib123` or `https://doi.org/10.123/tib123/`would be `10.123/tib123`

locale/en/submission.po
```diff
@@ -600,7 +609,7 @@ msgid "submission.citations.structured.label.rawCitation"
 msgstr "Edit Raw Citation"
 
 msgid "submission.citations.structured.label.doi"
-msgstr "DOI"
+msgstr "DOI, e.g. 10.1000/182, doi:10.1000/182, https://doi.org/10.1000/182"
```
jardakotesovec: could we have the examples as field description and not part of the label? I think it will be more readable.

bozana:
closed because moved to https://github.com/pkp/pkp-lib/pull/12000

PR #11924, opened 2025-10-14
pkp/pkp-lib#9968 Additional functions to support smarty->blade migration

jardakotesovec:


jardakotesovec:
Hi @touhidurabir and @asmecher 
small additions I needed for smooth smarty->blade migration of our current templates.

Suggestions on naming/namespacing these is appraciated.

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +117,24 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
             ?>";
         });
 
-        // Register the sanitizeHtml macro
-        Str::macro('sanitizeHtml', function (string $input, string $configKey = 'allowed_html') {
+       Str::macro('sanitizeHtml', function (?string $input, string $configKey = 'allowed_html') {
```
asmecher: Looks like maybe an inadvertent indentation change here?

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +117,24 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
...
+            if ($input === null || $input === '') {
```
asmecher: Lately PHP has been tightening up its requirements for functions that operate on strings, and soon will report fatal errors when a `null` is passed in. (I guess passing in an empty string will work just fine here, if maybe a little inefficiently.) It makes sense to me that our functions behave in line with PHP built-ins -- so callers would need to make sure that `null` values were cast to strings first, or that this function wasn't called in that case. So I'd be in favour of removing this conditional entirely, and changing `?string $input` to `string $input`, unless you feel strongly about it.

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +117,24 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
...
+                return '';
+            }
             $sanitized = PKPString::stripUnsafeHtml($input, $configKey);
-            $sanitized = str_replace('{{', '<span v-pre>{{</span>', $sanitized);
```
asmecher, line 122: Wasn't this a measure to guard against VueJS injections?

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +117,24 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
...
-            $sanitized = str_replace('}}', '<span v-pre>}}</span>', $sanitized);
             return $sanitized;
         });
+
+
+        Str::macro('bladeDateFormat', function ($string, $format = null, $default_date = '', $formatter = 'auto') {
+            return (new \Carbon\Carbon($string))->locale(Locale::getLocale())->translatedFormat($format);
+        });
```
touhidurabir: As this is a manipulation of data formatting, better to add this to `Illuminate\Support\Carbon` which extends the core `Carbon\Carbon` .  Also as this is not just tied to view (for now but I think will have more usage at other places) , better to moved it to `pkp-lib` `AppServiceProvider::boot` . And to have easier use like `Str` , we need to register an alias . So basically something like this in `AppServiceProvider`
```php
use Illuminate\Foundation\AliasLoader; // import namespace

// within boot method 

// register the alias loader
AliasLoader::getInstance()->alias('Carbon', \Illuminate\Support\Carbon::class);

Carbon::macro('localisedDataFormat', function (....) {
     // implementations
})
```

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +117,24 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
...
+
+        Str::macro('bladeUrl', function (array $parameters) {
+            return \PKP\template\PKPTemplateManager::getManager()->smartyUrl($parameters);
+        });
```
touhidurabir: this works . but if the use case only for template , perhaps adding the macro to `Str` does not make much sense . For view/template only usage, we can do something like following 
```php
AliasLoader::getInstance()->alias('View', \Illuminate\Support\Facades\View::class);
View::macro('url', function (array $parameters = []) {
    return \PKP\template\PKPTemplateManager::getManager()->smartyUrl($parameters);
});
```
and use it like `{{ View::url([....]) }}` while keep it in the view service provider .

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: Thanks, thats helpful. I was also iterating on some ideas this morning.

And my hesitation of extending Carbon is basically that I am not sure whether Carbon is strong long term bet as its 3rd party library. And on JS ecosystem I can see how the datetime libraries are evolving constantly.

So instead of I tried to introduce BladeHelper class, which can abstract any change without impacting in templates.


So we could support both directive and BladeHelper when used within translation or @php block:

```
                                <span>{{ __('submission.updatedOn', [
                                    'datePublished' => BladeHelper::dateFormat($firstPublication->getData('datePublished'), $dateFormatShort),
                                    'dateUpdated' => BladeHelper::dateFormat($publication->getData('datePublished'), $dateFormatShort)
                                ]) }}</span>

```

```
                                <span>@dateFormat($firstPublication->getData('datePublished'), $dateFormatShort)</span>

```

And same approach I could take for the url.

What you think? I am happy to follow your lead there. Just trying to make sure its going to be stable.. so we don't have to update templates in future much (especially because eventually plugins will start using it).

touhidurabir: @jardakotesovec  If I understand correctly , you want to put these kind of methods to a `Helper` class as static methods, right ? If that is the case , I will be supportive of this . One big benefit of this is to have a abstraction over the implementation as you have mentioned so that the change of underneath implementation does not need the change of the usage . Another one I see is the less use of macros which is in many cases is bit confusing and IDEs does not quite cover code completion. I am not sure about the name as `BladeHelper` or should it be `ViewHelper` , so leaving that up to you . 

then one issue I can see as these are going to be `static` method, it will be hard to mock/swap during a unit test specially . but if the usage is limited only within the template scope, I think that not going to cause an issue .

jardakotesovec: @touhidurabir Thank you, ViewHelper sounds good. Will update the PR and request one more review afterwards.

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +114,15 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
             ?>";
         });
 
-        // Register the sanitizeHtml macro
-        Str::macro('sanitizeHtml', function (string $input, string $configKey = 'allowed_html') {
-            $sanitized = PKPString::stripUnsafeHtml($input, $configKey);
-            $sanitized = str_replace('{{', '<span v-pre>{{</span>', $sanitized);
-            $sanitized = str_replace('}}', '<span v-pre>}}</span>', $sanitized);
-            return $sanitized;
+        Blade::directive('dateFormat', function ($expression) {
```
touhidurabir: if we are having a helper , then do we need this directive ?

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +114,15 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
...
+            return "<?php echo ViewHelper::dateFormat({$expression}); ?>";
         });
+
+
+        // Create a global alias so ViewHelper can be used without full namespace in templates
+        if (!class_exists('ViewHelper', false)) {
+            class_alias(\PKP\template\ViewHelper::class, 'ViewHelper');
+        }
```
touhidurabir: rather than register alias this way , lets use the same approach we have above for `Js` like 
```php
AliasLoader::getInstance()->alias('ViewHelper', \PKP\template\ViewHelper::class);
```

classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -116,13 +114,15 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
             ?>";
         });
 
-        // Register the sanitizeHtml macro
-        Str::macro('sanitizeHtml', function (string $input, string $configKey = 'allowed_html') {
```
touhidurabir, line 120: As this is getting moved to `ViewHelper`, need to update the basic doc at https://github.com/pkp/pkp-lib/issues/9968 in issue description .

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: Thanks, at first I thought we would not have this problem for vue.js & blade due syntax differences, but thinking more about it - I think same thing applies.

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: We don't.. I was considering it to keep it as shortcut..  but happy to drop it actually.. so its one way to format date.

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: Oki

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: And what about url? Do we also keep it only as ViewHelper:url ? And remove @url? 

I would be supportive of that..  as the @url does not work well in some scenarios as discussed and its nice having just one way to do things..

touhidurabir: > And what about url? Do we also keep it only as ViewHelper:url ? And remove @url?
> 
> I would be supportive of that.. as the @url does not work well in some scenarios as discussed and its nice having just one way to do things..

I agree , lets remove the @url directive and only url as helper . If we feel that we really need `@url` directive, we can add it back later . This will also need an update to the basic doc at https://github.com/pkp/pkp-lib/issues/9968 in issue description .

classes/core/PKPBladeViewServiceProvider.php
jardakotesovec: Yes, once we would merge this I am happy to update the issue.

I also intend to add there example for every smarty filter to make sure how to transition these as well. Often its just to use plain php functions or syntax.

Issue #11917, opened 2025-10-11
API to create a task/discussion from a template


Vitaliy-1:
**Description:**
One of the core task template functionality is a facilitation of a process to create a task. This should be done through a separate API endpoint. The endpoint should return a pre-filled task/discussion but shouldn't save it into the database 

**Solution:** 
1. Create an API endpoint with the same access as adding the task
2. Enhance Task template with data - date interval, task type.
3. The approach from a technological perspective should be applicable and for automatic task creation.

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/11928
OJS: https://github.com/pkp/ojs/pull/5137
ui-library: https://github.com/pkp/ui-library/pull/727
OMP: https://github.com/pkp/omp/pull/2148
OPS: https://github.com/pkp/ops/pull/1119

Vitaliy-1:
@jardakotesovec, can you take a look at this PR? It reflects our latest discussion regarding templates and adds a new endpoint to create a task from a template. It's a `POST` request with a following path: 
```
http://localhost:8000/publicknowledge/api/v1/submissions/{submissionId}/stages/{stageId}/tasks/fromTemplate/{templateId}
```

PR #11928, opened 2025-10-15
pkp/pkp-lib#11917 API to create a task from a template

Vitaliy-1:


api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -110,6 +111,10 @@ public function getGroupRoutes(): void
             ->name('submission.task.start')
             ->whereNumber(['submissionId', 'taskId']);
 
+        Route::post('{submissionId}/stages/{stageId}/tasks/fromTemplate/{templateId}', $this->fromTemplate(...))
```
jardakotesovec: If its not changing anything GET might be better fit?

Issue #11922, opened 2025-10-13
Expose Publication's Open Peer Reviews via API


taslangraham:
**Description:**
Implement necessary API endpoints that exposes the open/public peers reviews data associated the publications(versions) of a published item(article, etc)

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
1. Using the site '...' 
3. Using a DB query '...'
4. This table should have '...'
5. POST Request '...'
6. Should return '...'

**Solution:** What the proposed solutions are and additional affected areas

**What application are you using?**
  OJS, OMP or OPS version 3.6

**Additional information**
Please add any screenshots, videos or other information we can use to understand this dev task report.

**Pull Requests**
~~pkp-lib: https://github.com/pkp/pkp-lib/pull/11930~~
~~ojs: https://github.com/pkp/ojs/pull/5138~~

- `pkp-lib`: pkp/pkp-lib#11998
- `ojs`: pkp/ojs#5168
- `ops`: pkp/ops#1130
- `omp`: pkp/omp#2160

**Changes Introduced**

Work done for this issue introduced the following endpoints to OJS:

**[GET] `peerReviews/open/publications/{publicationId}`**
Returns review information for publication with the given `publicationId`

<details>
<summary>Example Response</summary>

```json
{
  "publicationId": 11,
  "datePublished": "2026-02-11",
  "reviewRounds": [
    {
      "displayText": "Version of Record 1.0",
      "roundId": 8,
      "originalPublicationId": 11,
      "status": "complete",
      "dateStarted": "2026-02-10 06:27:10",
      "dateInProgress": "2026-02-10 06:27:23",
      "dateCompleted": "2026-02-10 06:27:36",
      "reviews": [
        {
          "id": 15,
          "reviewerId": 9,
          "reviewerFullName": "Aisla McCrae",
          "reviewerAffiliation": "University of Manitoba",
          "dateCompleted": "2026-02-10 06:27:26",
          "isReviewOpen": true,
          "reviewerRecommendationDisplayText": "Revisions Required",
          "reviewerRecommendationId": 2,
          "reviewerRecommendationTypeId": 3,
          "reviewerRecommendationTypeLabel": "Revisions requested",
          "reviewForm": null,
          "reviewerComments": [
            "<p>Here are my review comments</p>"
          ]
        },
        {
          "id": 16,
          "reviewerId": 10,
          "reviewerFullName": "Adela Gallego",
          "reviewerAffiliation": "State University of New York",
[...]

Issue #11913, opened 2025-10-09
Associate review rounds with publications


ewhanson:
**Description:**
As an initial step for supporting open peer review, review rounds need to be able to be associated with publications/publication IDs, not just submissions/submission IDs. This first step will add support for publication IDs alongside submission IDs.

**Steps to test the changes:**
1. Run migration for `PKP\migration\upgrade\v3_6_0\I11914_Reviews`
2. Create a new review round for an existing or new submission.
3. Check `review_rounds` database table to see successfully saved publication ID.

**Solution:** What the proposed solutions are and additional affected areas

**What application are you using?**
  3.6

**Additional information**
N/A

### Round 1

- `pkp-lib`: pkp/pkp-lib#11966
- `ojs`: pkp/ojs#5159
- `ops`: pkp/ops#1124
- `omp`: pkp/omp#2153

### Round 2

- `pkp-lib`: pkp/pkp-lib#11985
- `ojs`: pkp/ojs#5165
- `ops`: pkp/ops#1126
- `omp`: pkp/omp#2155

ewhanson:
Hey @taslangraham, I'm still waiting for tests to pass, but could you have a look at this? Thanks!

taslangraham:
Hey @ewhanson, I've reviewed and left a few questions and suggestions for you on both PRs

Tribunal33:
@ewhanson not sure if this is ready for me but I went ahead since it was moved to "Ready for QA".  

<img width="1533" height="895" alt="Image" src="https://github.com/user-attachments/assets/a9bd02ee-fa92-4e4f-8ce6-bd907e04e33b" />

ewhanson:
Thanks @Tribunal33. I have a PR in for review to fix an issue @Vitaliy-1 ran into, but after he has a look at it, I'll close the issue.

ewhanson:
@Tribunal33, I've updated this with a fix for the migration. Not sure if you want to re-look at it, but it's otherwise done and can be closed.

asmecher:
@ewhanson, this caused a knock-on problem when deleting submissions; if there's a publication ID associated with the review round, and OJS is attempting to delete the publication, then the constraint fails. (There's code to delete the review rounds associated with the submission, but that only runs after the publications have been deleted).

Example log:
```
[...]

PR #11930, opened 2025-10-15
pkp/pkp-lib#11922 Expose Publication's Open Peer Reviews via API

taslangraham:
For pkp/pkp-lib#11922

### DIFF
--- a/classes/publication/Collector.php
+++ b/classes/publication/Collector.php
@@ -31,6 +31,7 @@ class Collector implements CollectorInterface
     public \APP\publication\DAO $dao;
     public ?array $contextIds;
     public ?array $submissionIds;
+    public ?array $publicationIds;
     public ?array $doiIds = null;
     public ?string $versionStage = null;
     public ?int $versionMajor = null;
@@ -83,6 +84,17 @@ class Collector implements CollectorInterface
     public function filterBySubmissionIds(?array $submissionIds): self
     {
         $this->submissionIds = $submissionIds;
+        return $this;
+    }
+
+    /**
+     * Filter by publication Ids
+     * 
+     * @param ?int[] $publicationIDs Publication IDs
+     */
+    public function filterByPublicationIds(?array $publicationIds): self
+    {
+        $this->publicationIds = $publicationIds;
         return $this;
     }
 
@@ -153,6 +165,9 @@ class Collector implements CollectorInterface
         if (isset($this->submissionIds)) {
             $qb->whereIn('p.submission_id', $this->submissionIds);
         }
+        if (isset($this->publicationIds)) {
+            $qb->whereIn('p.publication_id', $this->publicationIds);
+        }
 
         $qb->when($this->doiIds !== null, function (Builder $qb) {
             $qb->whereIn('p.doi_id', $this->doiIds);
@@ -165,7 +180,7 @@ class Collector implements CollectorInterface
         $qb->when($this->versionMajor !== null, function (Builder $qb) {
             $qb->where('p.version_major', $this->versionMajor);
         });
-        
+
         if (isset($this->statuses)) {
             $qb->whereIn('p.status', $this->statuses);
         }
--- a/classes/publication/Repository.php
+++ b/classes/publication/Repository.php
@@ -24,6 +24,7 @@ use APP\publication\Publication;
 use APP\submission\Submission;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Enumerable;
+use PKP\API\v1\publicationPeerReviews\resources\PublicationPeerReviewResource;
 use PKP\context\Context;
 use PKP\core\Core;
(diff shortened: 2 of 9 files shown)


classes/publication/Collector.php
```diff
@@ -86,6 +87,15 @@ public function filterBySubmissionIds(?array $submissionIds): self
         return $this;
     }
 
+    /**
+     * Filter by publication Ids
```
ewhanson, line 91:
```suggestion
     * Filter by publication Ids
     * 
     * @param ?int[] $publicationIDs Publication IDs
```

api/v1/publicationOpenPeerReviews/PublicationOpenPeerReviewController.php
```diff
@@ -0,0 +1,117 @@
...
+use PKP\core\PKPBaseController;
+
+class PublicationOpenPeerReviewController extends PKPBaseController
+{
+    /**
+     * @inheritdoc
+     */
+    public function getHandlerPath(): string
+    {
+        return 'publicationOpenPeerReviews';
```
ewhanson: To keep things a bit more open-ended for future changes, I think we should call this something like `publicationPeerReviews` instead and have open be a separate thing so the end point would be something like `/api/v1/publicationPeerReviews/open` since there may be something other than open in the future. Same goes for naming conventions for all the files, etc. What do you think?

classes/publication/Repository.php
```diff
@@ -977,4 +978,20 @@ public function getMinorVersionsSettingValues(int $submissionId, string $version
     {
         return $this->dao->getMinorVersionsSettingValues($submissionId, $versionStage, $versionMajor, $settingName);
     }
+
+    /**
+     * Get public peer review data for publications.
+     *
+     * @param Enumerable $publications - The publications to get peer review data for.
+     */
+    public function getPeerReviews(Enumerable $publications): Enumerable
```
ewhanson: Does `$publications` need to be an `Enumerable` since you are collecting it first thing in the method below? It might be easier to accept an array here since you're collecting it directly below. There's an [instance above](https://github.com/pkp/pkp-lib/pull/11930/files#diff-bb80c34eaa5c838c06651c0309281673e7466076755e42c03698546006be8a8cR113) where you collect an array of a single publication ID above and collect it again. I don't think it makes a huge difference, but thought I'd point it out.

api/v1/publicationOpenPeerReviews/resources/PublicationOpenPeerReviewResource.php
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * Get public peer review specific data for a list of review assignments.
+     *
+     * @param Enumerable $assignments - The review assignments to get data for.
+     *
+     */
+    private function getReviewAssignmentPeerReviews(Enumerable $assignments): Enumerable
+    {
+        $context = Application::get()->getRequest()->getContext();
```
ewhanson: Is this the right context? Should we be getting the context from the publication/submission rather than assuming the request submission will always match? Or is that not actually a problem?

classes/publication/Repository.php
taslangraham: You're right, having `$publications` as an `Enumerable` isn't necessary here. Thanks!

api/v1/publicationOpenPeerReviews/PublicationOpenPeerReviewController.php
taslangraham: I don't have any objections to this

api/v1/publicationOpenPeerReviews/resources/PublicationOpenPeerReviewResource.php
taslangraham: hmm, yea, getting it from the publication is the better choice here

taslangraham:
Thanks for the review @ewhanson. I've made updates based on your feedback.

classes/submission/reviewRound/ReviewRoundDAO.php
```diff
@@ -204,6 +204,21 @@ public function getBySubmissionId($submissionId, $stageId = null, $round = null)
         return new DAOResultFactory($q->get(), $this, '_fromRow', [], $q);
     }
 
+    /**
+     * Get an iterator of review round objects associated with a set of publication IDs
+     */
+    public function getByPublicationIds($publicationIds, $stageId = null, $round = null): DAOResultFactory
```
ewhanson, line 210: Nothing that needs to be done now, but this method will be in conflict with pkp/pkp-lib#11913, but should be an easy fix once that's merged.

ewhanson:
Thanks @taslangraham. This will be ready to merge as soon as the publication ID migration PRs get merged.

taslangraham:
These changes were merged in via https://github.com/pkp/pkp-lib/pull/11998

PR #11942, opened 2025-10-17
pkp/pkp-lib#9345 added ability to retrive context for jobs on CLI session

touhidurabir:
for pkp/pkp-lib#9345 .

### DIFF
--- a/classes/core/PKPQueueProvider.php
+++ b/classes/core/PKPQueueProvider.php
@@ -17,9 +17,14 @@
 namespace PKP\core;
 
 use APP\core\Application;
+use Illuminate\Database\PostgresConnection;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Contracts\Debug\ExceptionHandler;
-use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Queue\Events\JobFailed;
+use Illuminate\Queue\Events\JobProcessed;
+use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\QueueServiceProvider as IlluminateQueueServiceProvider;
 use Illuminate\Queue\Worker;
 use Illuminate\Queue\WorkerOptions;
@@ -29,6 +34,8 @@ use PKP\config\Config;
 use PKP\job\models\Job as PKPJobModel;
 use PKP\queue\JobRunner;
 use PKP\queue\WorkerConfiguration;
+use PKP\queue\PKPQueueDatabaseConnector;
+use Throwable;
 
 class PKPQueueProvider extends IlluminateQueueServiceProvider
 {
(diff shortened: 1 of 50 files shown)


classes/core/PKPQueueProvider.php
```diff
@@ -131,17 +190,35 @@ public function boot()
...
+                if (app()->runningInConsole()) {
+                    return;
+                }
+
+                // Not to run in unit test mode as part of the application lifecycle
+                if (app()->runningUnitTests()) {
+                    return;
+                }
+
+                error_log('Shutdown function started at: ' . microtime(true));
```
asmecher: This should be removed (also below); it'll clutter the logs.

classes/core/PKPQueueProvider.php
```diff
@@ -155,6 +232,123 @@ public function boot()
...
+                $context = $contextDao->getById($contextId);
+                
+                if (!$context) {
+                    $jobName = $event->job->payload()['displayName'] ?? 'Unknown';
+
+                    // Fail the job immediately with a meaningful exception
+                    throw new \RuntimeException(
+                        "Job execution failed: Invalid context_id {$contextId}. The context does not exist in the database for job: {$jobName}."
+                    );
+                }
```
asmecher, line 279: I'd be tempted to just delete a job that referred to a nonexistent context ID without logging a message. The only way this could happen is if the context got deleted, and in that (drastic) case, we can consider anything attached to the context to be left-over clutter. We have cases like this in the upgrade scripts; in a situation as drastic as this, we just clean up the leftover data and move on without complaining.

classes/core/PKPQueueProvider.php
```diff
@@ -155,6 +232,123 @@ public function boot()
...
+
+                Application::get()->setCliContext($context);
+            }
+
+            if (!app()->runningUnitTests()) {
+                // Initialize the locale and load generic plugins for context or no context
+                // But will not load when running unit tests as part of the application lifecycle 
+                // to avoid any unintended side effect on tests that are not related to queue jobs
+                \PKP\plugins\PluginRegistry::loadCategory('generic', false, $contextId);
```
asmecher, line 290: I think this will cause hook registrations to pile up in an unpredictable way.

Consider our normal hook registration pattern:
```php
public function register($category, $path, $mainContextId = null): bool
{
    if (parent::register($category, $path, $mainContextId)) {
        if ($this->getEnabled($mainContextId)) {
            Hook::add('SomeHookName', $this->someHookCallback(...));
        }
        return true;
    }
    return false;
}
```
Unless I recall wrong, with multiple contexts, each call to `loadCategory` will result in the hooks being registered again; there is no de-duplication of hook registrants.

asmecher (changes requested):
@touhidurabir, I've added a couple of comments; if my understanding of the plugin load / hook registration cycle is not missing something, we've got a conceptual problem to solve to prevent multiple registrations of a plugin on a hook.

This won't affect most plugins, since few currently use the job runner tools -- but I think an approach that duplicates registrations, or expects the hook to somehow support de-registration, will lead to tricky downstream problems on long-running threads.

I like the "ContextAwareJob" interface allowing a job to specify how its context ID should be derived from its contents; it adds smarts without adding a lot of code.

classes/core/PKPQueueProvider.php
touhidurabir: It is a tempted idea but I think not that straight forward . We only get to know this information at this point when the job has been picked to process, deleting job at the point may cause unexpected behaviour by internal core job processor . That is why it's immediately being failed and pushed to failed jobs list which is the right approach.

classes/core/PKPQueueProvider.php
touhidurabir: > with multiple contexts, each call to loadCategory will result in the hooks being registered again; there is no de-duplication of hook registrants.

@asmecher it is right about the theoretical mechanism, e.g. if `PluginRegistry ::loadCategory` were called with different `$contextId` values in the same process, the cache key would differ `("generic\0\05" vs "generic\0\07")`, creating separate cache entries. And `PluginRegistry::register()` guard only protects against same-named plugins, not against the same plugin being loaded as a different object instance from a different cache entry.

But this scenario cannot happen because of the `Looping::class` listener. Before the worker even pops a job from a different context, it detects the mismatch and sets `shouldQuit = true`. The worker exits, supervisor restarts fresh. So `PluginRegistry::loadCategory` is never called with two different $contextId values in the same process.

This is the kind of similar concern @jonasraoni has which pointed at https://github.com/pkp/pkp-lib/pull/9463#discussion_r2723064775 and I explain in that thread . 

so basically the `$this->app['events']->listen(Looping::class, function (Looping $event) {...});` here is doing a peak-ahead for each current job at the next pickable one to see if that going to have a different `contextId` set at `static $lockedContextId = null;` . and if so, it will set the `app('queue.worker')->shouldQuit = true;` which will quit the worker before it get to pick the next job right after complete processing the current one . The the supervisor (or any daemon monitoring service/process) will re-initiate the worker and it's start fresh with fresh app boot in the memory .

touhidurabir: Anyway it's better to add a guard to reduce the subsequent redundant calls, so I added it .

touhidurabir:
outdated PRs, closing in favour of latest PRs .

Issue #11933, opened 2025-10-16
[Login] | Multiple Browser or devices - all pages, being logged in on one browser logs you out of the other browser


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

This is true for cross platforms too (like phones and PCs).

When you login on one browser you will be logged out of the other browser after reload. Currently this is how it works on OJS 3.4.  There must be a regression somewhere but I couldn't find the original requirement.

There were similar fixed issues here https://github.com/pkp/pkp-lib/issues/4595 and here: https://github.com/pkp/pkp-lib/issues/3624 

### Steps to Reproduce

1. Login with any user on one browser
2. Open up another browser (or incognito works too)
3. Login with the same user on this browser
4. Flip back to the first browser
5. Reload the page

### Expected Result

User's logged in status should be maintained.  

### Actual Result

User is logged out and redirected back to the login page.

### Application Version

OJS 3.5.0-x and main

touhidurabir:
@asmecher can you review the PRs at https://github.com/pkp/pkp-lib/issues/11933#issue-3520117532 . The root cause to issue is the `cost` of password hash which is by default to 10 up to PHP 8.3 but our upgrade of laravel which is using 12 . Using PHP 8.4+ does not to have this issue as this has been upgraded to 12 as default .

asmecher:
@touhidurabir PRs approved, thanks!

touhidurabir:
@bozana can you please review the PRs are https://github.com/pkp/pkp-lib/issues/11933#issue-3520117532 . Need another approval .

touhidurabir:
Merged the `pkp-lib` PRs for `3.5.0` and `main` branch . Confirming that before merge, all tests were passing for both 3.5.0/main branch .

touhidurabir:
@Tribunal33 ready for QA check .

Tribunal33:
@touhidurabir or @asmecher how should we be handling logout when multiple browsers are signed in?  What I see now is that when I logout on one device/browser I'm not being logged out everywhere.  

_As it stands this ticket passes QA.  I am able to be logged in on several browsers/tabs/devices.  Tested Chrome tabs, incognito same browser, Edge, Safari (mobile device).  All were logged into same user and refreshing the page did not close any other user session._

Previously multiple logouts wouldn't be an issue because you couldn't have simultaenous logins.  

[...]

PR #11952, opened 2025-10-19
pkp/pkp-lib#11933 Proper hashing cost applied at login time

touhidurabir:
for pkp/pkp-lib#11933 .


asmecher:
Both PRs approved, thanks!

Issue #11832, opened 2025-09-17
[Email Templates] | Edit Templates - Review Email, Editorial signature is not being displayed correctly in the templates


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Bug from community forum : https://forum.pkp.sfu.ca/t/ojs-3-3-review-request-email-shows-wrong-signature/97076
Appears on ~~3.5 as well as the~~ 3.3.0-17 in original post.  Essentially profile signature isn't showing up even when there is a profile signature.
**Code is suggested for fix at the bottom** 

<img width="1807" height="1092" alt="Image" src="https://github.com/user-attachments/assets/15123284-5e4c-4240-bf77-d1b6ef9c8b1e" />

### Steps to Reproduce

1. Preconditions.  As an Editor you must have a signature in the Profile > Contact > Signature
2. Login with Editor
3. Navigate to Workflow settings > Email > Email Templates
4. Find template for 3.3 REVIEW_REQUEST and REVIEW_REQUEST_ONECLICK
5. Scroll down to the bottom of the template
6. Observe the template emails

To see the email
1. Login as Editor
2. Find a submission in Review > View
3. Add a reviewer
4. Find the Review Request template (default)
5. Observe the email signature

### Expected Result

Should be the signature from the profile when it is filled out and full name when it is not filled out

### Actual Result

It displays the full name & e-mail instead of their saved signature 

Refer to original forum post : https://forum.pkp.sfu.ca/t/ojs-3-3-review-request-email-shows-wrong-signature/97076

### Application Version

OJS 3.3.0-17 (found) but still in 3.5.0-2 OJS

### Additional Information

Here is the community member's work to trace the root cause of the bug.  Not sure if the fix is good.  It looks like it returns if there is a signature but skips some other things?

```php
function getContactSignature() {
    $signature = htmlspecialchars($this->getFullName());
    AppLocale::requireComponents(LOCALE_COMPONENT_PKP_USER);
    if ($a = $this->getLocalizedAffiliation()) $signature .= '<br/>' . htmlspecialchars($a);
    if ($p = $this->getPhone()) $signature .= '<br/>' . __('user.phone') . ' ' . htmlspecialchars($p);
    $signature .= '<br/>' . htmlspecialchars($this->getEmail());
    return $signature;
}
```

Suggested fix:
```php
function getContactSignature() {
        $signature = $this->getLocalizedSignature();
        if (!empty($signature)) {
[...]

PR #11955, opened 2025-10-20
pkp/pkp-lib#11832 Use User's custom signature when available

taslangraham:
For pkp/pkp-lib#11832

### DIFF
--- a/classes/user/User.inc.php
+++ b/classes/user/User.inc.php
@@ -337,7 +337,7 @@ class User extends Identity {
 	}
 
 	function getContactSignature() {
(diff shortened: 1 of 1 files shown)


Hafsa-Naeem (approved):

taslangraham:
> Looks good to me . just one question , is it properly working with different locales depending on signature available on that locale and what is the targeted locale ?

Yea, it will return the signature available for the current locale.

>what is the targeted locale
The signature is multilingual, so the user should be able enter a signature for any of the active locales

PR #11956, opened 2025-10-20
pkp/pkp-lib#11957 Move handling of Content Menu to app's template manager

blesildaramirez:


classes/template/PKPTemplateManager.php
```diff
@@ -1266,17 +1266,26 @@ public function setupBackendPage()
...
+                                'name' => __('manager.userComment.comments'),
+                                'url' => $router->url($request, null, 'management', 'settings', ['userComments']),
+                                'isCurrent' => $router->getRequestedPage($request) === 'management' && in_array('userComments', (array) $router->getRequestedArgs($request)),
+                            ];
+                        }
 
+                        $isOPS = Application::getName() === 'ops';
+
+                        // In OPS, only show Content menu to users with settings access
+                        if (!$isOPS || $hasSettingsAccess) {
```
taslangraham: Could we accomplish this without doing an app specific check inside pkp-lib? Maybe introduce a method like, `populateContentMenu` or similar, that each app can override with app specific content. What do you think?

blesildaramirez: Hi @taslangraham , yes this can be done through the app's template manager. I moved the handling of Content Menu there instead, so these are now handled in ojs/omp/ops. Could you check the PRs please?
(these are also linked on the issue's [description](https://github.com/pkp/pkp-lib/issues/11957#issue-3533919494).
ojs: https://github.com/pkp/ojs/pull/5153
omp: https://github.com/pkp/omp/pull/2144
ops: https://github.com/pkp/ops/pull/1117

taslangraham (approved):

classes/template/PKPTemplateManager.php
taslangraham: All reviewed @blesildaramirez. Just the one [comment](https://github.com/pkp/ops/pull/1117#discussion_r2453676765) on the OPS PR. Once that's resolved then everything will be good. Thanks!

jardakotesovec (approved):

Issue #11697, opened 2025-08-05
[ORCID] | Edit contributor deletes authenticated ORCID


Bernardo-f:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

While testing version 3.5 of OJS, we noticed that when removing an article from publication to complete the metadata of its contributors (e.g., adding the ROR of their affiliations), the ORCID ID of the contributor was being deleted.

After further investigation, we confirmed that this happens regardless of whether any of the contributor's fields are edited or not.

### Steps to Reproduce

1. Log in as Admin or Editor of a journal.
2. Go to Published Submissions: `/dashboard/editorial?currentViewId=published`
3. Select an article that has contributors with ORCID IDs.
4. Go to the "Contributors" section.
5. Select a contributor with an authenticated ORCID.
6. Click the "Save" button without making any changes.

### Expected Result

The contributor should be editable without removing their ORCID, especially if it has already been authenticated.

### Actual Result

The contributor's ORCID is deleted, even if no information was modified.

### Application Version

3.5.0-1

### Additional Information

While investigating this issue, we reviewed some components and made a change to the UI-library to store the ORCID value and send it along during updates, since it was not being included in the request payload.

We are not entirely sure if this is the correct behavior, as the ORCID is checked for verification before submission. However, this change helped preserve the ORCID in our case.

Here is the commit we made in a forked version of the UI-library, in case it is useful:

🔗 https://github.com/Bernardo-f/ui-library/commit/307ba7c50c423ac2005f675984112be8c0692683

asmecher:
@ewhanson, assigning to you, thanks!

taslangraham:
Hey @ewhanson, I added PRs for this at https://github.com/pkp/pkp-lib/issues/11697#issue-3291481636 for stable 3.5. Could you take a look at it, please? If all looks good, then I'll port to main.

The solution I went with was to essentially prevent the contributor endpoints from updating an Author's ORCiD.

ewhanson:
Thanks @taslangraham. Just a small question about the locale string. After you address that, happy to merge. This should also be forward ported to `main`.

taslangraham:
Thanks for the review @ewhanson and @asmecher. I've made the necessary changes based on your reviews, and this should be ready to merge now.

asmecher:
Thanks, @taslangraham, I've merged to `stable-3_5_0` and cherry-picked forward to `main`. I'll manage the submodule updates manually.

[...]

PR #11963, opened 2025-10-22
pkp/pkp-lib#11697 Fix unintended deletion of ORCiD on edit contributor

taslangraham:
For pkp/pkp-lib#11697

### DIFF
--- a/locale/en/api.po
+++ b/locale/en/api.po
@@ -167,6 +167,9 @@ msgstr "The highlight you requested was not found."
 msgid "api.orcid.403.orcidNotEnabled"
 msgstr "You cannot perform this operation with ORCID functionality disabled."
 
+msgid "api.orcid.403.cannotUpdateAuthorOrcid"
+msgstr "Assigning an author's ORCiD ID via the Edit/Add Author endpoints is not permitted. To request that an author update their ORCID iD, please use the \"orcid/requestAuthorVerification/{authorId}\" endpoint."
+
 msgid "api.orcid.404.authorNotFound"
 msgstr "The author requested was not found"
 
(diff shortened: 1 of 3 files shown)


locale/en/api.po
```diff
@@ -167,6 +167,9 @@ msgstr "The highlight you requested was not found."
 msgid "api.orcid.403.orcidNotEnabled"
 msgstr "You cannot perform this operation with ORCID functionality disabled."
 
+msgid "api.orcid.403.cannotUpdateAuthorOrcid"
+msgstr "Assigning an author's ORCiD ID via the Edit/Add Author endpoints is not permitted. To request that an author update their ORCID iD, please use the \"orcid\\requestAuthorVerification\{authorId}\" endpoint."
```
ewhanson: The escaping isn't quite right for the endpoint in the string. I don't remember off hand which is correct, but `requestAuthorVerificaiton` has 2 `//` before it and `{authorId}` has only 1 `/` before it.

asmecher: This looks correct to me -- `\"` to encode a `"` character, and `\\` to encode a `\` character. See:
http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html#sec-poescapes

> The plain double quote is written as `\"`

...and...

> Another two escape sequences, usually of much lower frequency than the double quote and the newline, are the tabulator `\t` and the backslash itself `\\` (because single backslash always starts an escape sequence).

ewhanson: Shouldn't the one after requestAuthorVerification be doubled then too? Or am I just completely missing something?

asmecher: ...actually, shouldn't those just be forward slashes?
```
msgstr "Assigning an author's ORCiD ID via the Edit/Add Author endpoints is not permitted. To request that an author update their ORCID iD, please use the \"orcid/requestAuthorVerification/{authorId}\" endpoint."
```
(sorry for adding to the confusion)

ewhanson: Yes, probably! So in conclusion, the escaped quotes are okay `\"` but the path seprates should be `/`. Thanks!

Issue #11914, opened 2025-10-10
[CRediT] - Migrate data saved by the old plugin into the core


bozana:
Now that CRediT is part of the core, as correctly marked here: https://github.com/pkp/pkp-lib/issues/857#issuecomment-3378697979, the data that existed by using the old plugin should now be migrated to the new data model in the core.

ajnyga:
ojs: https://github.com/pkp/ojs/pull/5536
ops: https://github.com/pkp/ops/pull/1336
pkp-lib: https://github.com/pkp/pkp-lib/pull/12704

ajnyga:
Restarted the test. Local migration tests from 3.5 and Credit plugin to 3.6 and core feature seemed to work ok.

ping @bozana tests passed

PR #11966, opened 2025-10-22
pkp/pkp-lib#11913 Associate review rounds with publications

ewhanson:


classes/migration/upgrade/v3_6_0/I11914_Reviews.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I11914_Reviews.php
```
taslangraham: File name, class name, and documentation is referencing [incorrect issue](https://github.com/pkp/pkp-lib/issues/11914) number.

taslangraham: The migration name could be a bit more descriptive to better reflect the changes it introduces.

classes/migration/upgrade/v3_6_0/I11913_AssociatePublicationsWithReviewRounds.php
```diff
@@ -0,0 +1,75 @@
...
+                    // so we should get the oldest existing publication for each submission
+                    $publicationId = DB::table('publications')
+                        ->where('submission_id', '=', $reviewRound->submission_id)
+                        ->orderBy('created_at')
+                        ->pluck('publication_id')
+                        ->first();
+
+                    DB::table('review_rounds')
+                        ->where('submission_id', '=', $reviewRound->submission_id)
+                        ->update(['publication_id' => $publicationId]);
```
taslangraham, line 75: This will update all rows where `submission_id` has the value from`$reviewRound->submission_id`, and as such if the process encounters another review round with same submission ID, then those rows would be updated again, resulting in multiple updates of the same rows.

Since we only need to find the first publication for each submission and update the associated review rounds with that publication ID, we could chunk by `submission_id` instead. This way, we process each submission once and perform a single query to update all rows with a given submission ID. Haven't tested, but something like the following should work:

```php
DB::table('review_rounds')
->select(['submission_id'])
->distinct()
->chunk(100, function (Collection $submissions) {
    foreach ($submissions as $submission) {
        $publicationId = DB::table('publications')
            ->where('submission_id', '=', $submission->submission_id)
            ->orderBy('created_at')
            ->pluck('publication_id')
            ->first();
        
        DB::table('review_rounds')
            ->where('submission_id', '=', $submission->submission_id)
            ->update(['publication_id' => $publicationId]);
    }
}, 'submission_id');
```

classes/submission/reviewRound/ReviewRound.php
```diff
@@ -85,6 +85,16 @@ public function setSubmissionId($submissionId)
...
 
+    public function getPublicationId(): int
+    {
+        return (int) $this->getData('publicationId');
+    }
+
+    public function setPublicationId(int $publicationId): void
+    {
+        $this->setData('publicationId', $publicationId);
+    }
```
taslangraham, line 102: Missing PHPDoc for methods

classes/submission/reviewRound/ReviewRoundDAO.php
```diff
@@ -204,6 +210,27 @@ public function getBySubmissionId($submissionId, $stageId = null, $round = null)
         return new DAOResultFactory($q->get(), $this, '_fromRow', [], $q);
     }
 
+
+    /**
+     * Get an iterator of review round objects associated with this submission
```
taslangraham:
```suggestion
     * Get an iterator of review round objects associated with this publication.
```

classes/submission/reviewRound/ReviewRoundDAO.php
```diff
@@ -256,6 +302,32 @@ public function getLastReviewRoundBySubmissionId($submissionId, $stageId = null)
         return $row ? $this->_fromRow($row) : null;
     }
 
+    /**
+     * Get the last review round for a give stage (or for the latest stage)
```
taslangraham:
```suggestion
     * Get the last review round for a given stage (or for the latest stage).
```

classes/submission/reviewRound/ReviewRoundDAO.php
```diff
@@ -256,6 +302,32 @@ public function getLastReviewRoundBySubmissionId($submissionId, $stageId = null)
...
+     */
+    public function getLastReviewRoundByPublicationId(int $publicationId, ?int $stageId = null): ?ReviewRound
+    {
+        $q = DB::table('review_rounds', 'rr')
+            ->where('rr.publication_id', '=', $publicationId)
+            ->when($stageId, fn (Builder $q) => $q->where('rr.stage_id', '=', $stageId))
+            ->orderByDesc('rr.stage_id')
+            ->orderByDesc('rr.round');
+
+        $resultsFactory = new DAOResultFactory($q->first(), $this, '_fromRow', [], $q);
```
taslangraham: Just double checking if `$q->first()` is intentional here?

taslangraham (changes requested):
Mostly looks good @ewhanson! Just a few questions and suggestions.

classes/migration/upgrade/v3_6_0/I11914_Reviews.php
ewhanson: Oops, thanks!

classes/migration/upgrade/v3_6_0/I11913_AssociatePublicationsWithReviewRounds.php
ewhanson: That's a good idea. I'll give that a try. 👍

classes/submission/reviewRound/ReviewRoundDAO.php
ewhanson: No, I double checked and I think we do want `get()` here after all since the `DAOResultFactory` was the first argument to be a generator or Enumerable rather than a single record. Thanks!

PR #11967, opened 2025-10-23
pkp/pkp-lib#9968 Add missing Blade directive for search page

jardakotesovec:


classes/core/PKPBladeViewServiceProvider.php
```diff
@@ -100,6 +100,26 @@ function (\Illuminate\View\Component $component, string $viewPath): string {
             ?>";
         });
 
+        // use as @html_select_date_a11y(['legend' => $dateFromLegend, 'prefix' => 'dateFrom', 'time' => $dateFrom, 'start_year' => $yearStart, 'end_year' => $yearEnd])
```
touhidurabir: as blade directive is registered as `htmlSelectDateA11y`, the usage is also as `@htmlSelectDateA11y` . here and below .

Issue #11885, opened 2025-09-30
Implement Edit, Delete and Search endpoints


Hafsa-Naeem:
**Description:**
Implement Edit, Delete, and Search/List endpoints for editorial task templates.
This complements the existing POST /api/v1/editTaskTemplates endpoint.

**What application are you using?**
OJS, OMP or OPS main branch

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib:[ pkp/pkp-lib#11885 Implemented edit, delete, and search endpoints for task templates and fixed stageId on add #11968](https://github.com/pkp/pkp-lib/pull/11968)
ojs:[ updated swagger-source.json and Submodule update ##Hafsa-Naeem/i11885-main-fix #5160](https://github.com/pkp/ojs/pull/5160)

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib:[ pkp/pkp-lib#11885 Implemented edit, delete, and search endpoints for task templates and fixed stageId on add #11968](https://github.com/pkp/pkp-lib/pull/11968)
ojs:[ updated swagger-source.json and Submodule update ##Hafsa-Naeem/i11885-main-fix #5160](https://github.com/pkp/ojs/pull/5160)

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib:[ pkp/pkp-lib#11885 Implemented edit, delete, and search endpoints for task templates and fixed stageId on add #11968](https://github.com/pkp/pkp-lib/pull/11968)
ojs:[ updated swagger-source.json and Submodule update ##Hafsa-Naeem/i11885-main-fix #5160](https://github.com/pkp/ojs/pull/5160)

Vitaliy-1:
@Hafsa-Naeem, can you go through all Bozana's comments. I see some of the aren't resolved. E.g., https://github.com/pkp/pkp-lib/pull/11968/files#r2489440186 (duplicated `type` property, I see it remains in your branch: https://github.com/Hafsa-Naeem/pkp-lib/blob/i11885-main-fix/api/v1/editTaskTemplates/PKPEditTaskTemplateController.php#L94)

Hafsa-Naeem:
@bozana @Vitaliy-1 Ready for review
[...]

PR #11968, opened 2025-10-23
pkp/pkp-lib#11885 Implemented edit, delete, and search endpoints for task templates and fixed stageId on add

Hafsa-Naeem:
for #11885

### DIFF
--- a/api/v1/editTaskTemplates/formRequests/TaskTemplateRequestTrait.php
+++ b/api/v1/editTaskTemplates/formRequests/TaskTemplateRequestTrait.php
@@ -27,13 +27,7 @@ trait TaskTemplateRequestTrait
 
     protected function getStageIds(): array
     {
-        $stages = Application::getApplicationStages();
-
-        return array_values(
-            array_unique(
-                array_map('intval', array_values((array) $stages))
-            )
-        );
+        return Application::getApplicationStages();
     }
 
     protected function userGroupIdsItemRules(int $contextId): array
--- a/classes/editorialTask/Template.php
+++ b/classes/editorialTask/Template.php
@@ -245,17 +245,17 @@ class Template extends Model
 
             $query->where(function (Builder $outer) use ($keywords, $settingsTable, $pk, $selfTable) {
                 foreach ($keywords as $tok) {
-                    $like = '%' . str_replace(['%', '_'], ['\\%', '\\_'], mb_strtolower($tok, 'UTF-8')) . '%';
+                    $like = '%' . addcslashes($tok, '%_') . '%';
 
                     $outer->where(function (Builder $q) use ($like, $settingsTable, $pk, $selfTable) {
-                        $q->whereRaw('LOWER(title) LIKE ?', [$like])
-                            ->orWhereRaw('LOWER(description) LIKE ?', [$like])
+                        $q->whereRaw('LOWER(title) LIKE LOWER(?)', [$like])
+                            ->orWhereRaw('LOWER(description) LIKE LOWER(?)', [$like])
                             ->orWhereExists(function ($sub) use ($like, $settingsTable, $pk, $selfTable) {
                                 $sub->select(DB::raw(1))
                                     ->from($settingsTable . ' as ets')
                                     ->whereColumn("ets.$pk", "$selfTable.$pk")
                                     ->whereIn('ets.setting_name', ['name', 'description'])
-                                    ->whereRaw('LOWER(ets.setting_value) LIKE ?', [$like]);
+                                    ->whereRaw('LOWER(ets.setting_value) LIKE LOWER(?)', [$like]);
                             });
                     });
                 }


api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -162,5 +183,67 @@ public function getMany(Request $request): JsonResponse
...
+    /**
+     * UPDATE /api/v1/editTaskTemplates/{templateId}
+     */
+    public function update(UpdateTaskTemplate $request): JsonResponse
+    {
+        $contextId = (int) $this->getRequest()->getContext()->getId();
+        $id = (int) $request->route('templateId');
+
+        $template = Template::query()
+            ->where('context_id', $contextId)
```
Vitaliy-1: It's better to use scope here as we avoid using column names besides the Model itself

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -162,5 +183,67 @@ public function getMany(Request $request): JsonResponse
...
+            ->findOrFail($id);
```
Vitaliy-1: We would need the explicit response error code (404) and message

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -162,5 +183,67 @@ public function getMany(Request $request): JsonResponse
...
+
+        $data = $request->validated();
+
+        DB::transaction(function () use ($template, $data) {
+            $updates = [];
+            if (array_key_exists('stageId', $data)) $updates['stage_id'] = (int) $data['stageId'];
+            if (array_key_exists('title', $data)) $updates['title'] = $data['title'];
+            if (array_key_exists('include', $data)) $updates['include'] = (bool) $data['include'];
+            if (array_key_exists('emailTemplateKey', $data)) $updates['email_template_key'] = $data['emailTemplateKey'];
```
Vitaliy-1: Just removed the email template key association. The current idea is to hard code the message into the template using `description` attribute

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -162,5 +183,67 @@ public function getMany(Request $request): JsonResponse
...
+
+        $template = Template::query()
+            ->where('context_id', $contextId)
+            ->findOrFail($id);
+
+        $data = $request->validated();
+
+        DB::transaction(function () use ($template, $data) {
+            $updates = [];
+            if (array_key_exists('stageId', $data)) $updates['stage_id'] = (int) $data['stageId'];
```
Vitaliy-1: I think we don't need these checks and go directly to the update

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -162,5 +183,67 @@ public function getMany(Request $request): JsonResponse
...
+            if (array_key_exists('title', $data)) $updates['title'] = $data['title'];
+            if (array_key_exists('include', $data)) $updates['include'] = (bool) $data['include'];
+            if (array_key_exists('emailTemplateKey', $data)) $updates['email_template_key'] = $data['emailTemplateKey'];
+            if (array_key_exists('type', $data)) $updates['type'] = (int) $data['type'];
+
+            if ($updates) {
+                $template->update($updates);
+            }
+            if (array_key_exists('userGroupIds', $data)) {
```
Vitaliy-1: Maybe something like the following would be better:
```php
$userGroupIds = Arr::pull($data, 'userGroupIds');
$template->update($data);
$template->userGroups()->sync($userGroupIds);
```

api/v1/editTaskTemplates/formRequests/AddTaskTemplate.php
```diff
@@ -30,7 +30,14 @@ public function authorize(): bool
     public function rules(): array
     {
         $contextId = Application::get()->getRequest()->getContext()->getId();
-        $stageIds = array_keys(Application::getApplicationStages());
+
+        // build a list of allowed stage IDs from values
+        $stages = Application::getApplicationStages();
+        $stageIds = array_values(array_unique(array_filter(
+            array_map('intval', array_merge(array_keys((array)$stages), array_values((array)$stages))),
+            fn ($id) => $id > 0
```
Vitaliy-1: probably we don't need this check

api/v1/editTaskTemplates/formRequests/AddTaskTemplate.php
```diff
@@ -43,6 +50,7 @@ public function rules(): array
             'title' => ['required', 'string', 'max:255'],
             'include' => ['boolean'],
             'emailTemplateKey' => ['sometimes', 'nullable', 'string', 'max:255', Rule::in($emailKeys)],
+            'type' => ['required','integer','in:1,2'],
```
Vitaliy-1: I think we can re-use our EditorialTaskType enum here instead of plain numbers

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,85 @@
...
+
+    public function rules(): array
+    {
+        $contextId = Application::get()->getRequest()->getContext()->getId();
+
+        // Build allowed stage IDs (values)
+        $stages = Application::getApplicationStages();
+        $stageIds = array_values(array_unique(array_filter(
+            array_map('intval', array_merge(array_keys((array)$stages), array_values((array)$stages))),
+            fn ($id) => $id > 0
```
Vitaliy-1: Probably this check isn't needed

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,85 @@
...
+        )));
+
+        // Context-scoped email keys
+        $emailKeys = \APP\facades\Repo::emailTemplate()
```
Vitaliy-1: This can be removed as we don't rely on email templates anymore

classes/editorialTask/Template.php
```diff
@@ -33,13 +34,17 @@ class Template extends Model
 
     public $timestamps = true;
 
+    public const TYPE_DISCUSSION = 1;
```
Vitaliy-1: We can just re-use EdirotialTaskType enum

Vitaliy-1 (changes requested):
Thanks, I left a couple of comments. Can you also do a rebase, this would also require additional modifications because of a recent merge.

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -156,4 +179,78 @@ public function getMany(Request $request): JsonResponse
...
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $data = $request->validated();
+
+        DB::transaction(function () use ($template, $data) {
+            $userGroupIds = Arr::pull($data, 'userGroupIds');
+            $updates = collect($data)
```
Vitaliy-1: Not anything critical but this should already be done during validation

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace PKP\API\v1\editTaskTemplates\formRequests;
+
+use APP\core\Application;
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
+use PKP\editorialTask\enums\EditorialTaskDueInterval;
+use PKP\editorialTask\enums\EditorialTaskType;
+
+class UpdateTaskTemplate extends FormRequest
```
Vitaliy-1, line 11: I think this class and `AddTaskTemplate` could have a common parent to avoid code duplication. Or `AddTaskTemplate` could extend this class. Or have a common trait...

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    public function authorize(): bool
```
Vitaliy-1: This won't work as we haven't yet integrated/adopted Laravel auth policies (it uses Laravel gates and policies according to the documentation: https://laravel.com/docs/12.x/validation#authorizing-form-requests)

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+     */
+    public function scopeFilterBySearch(Builder $query, string $phrase): Builder
+    {
+        $phrase = trim($phrase);
+        if ($phrase === '') {
+            return $query;
+        }
+
+        $tokens = preg_split('/\s+/', $phrase) ?: [];
+        $tokens = array_values(array_filter($tokens, fn ($t) => $t !== ''));
```
Vitaliy-1, line 210: We also should do a special treatment for the task or discussion keyword within search phrase. E.g., the search phrase
```
task ethics
```
The code should perform the search by the keyword `ethics` within the tasks only.

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+     * name, description
+     */
+    public function scopeFilterBySearch(Builder $query, string $phrase): Builder
+    {
+        $phrase = trim($phrase);
+        if ($phrase === '') {
+            return $query;
+        }
+
+        $tokens = preg_split('/\s+/', $phrase) ?: [];
```
Vitaliy-1, line 209: Not sure but maybe we already got a helper to break the phrase onto tokens somewhere in our codebase.

Vitaliy-1:
@bozana, could you take a look at the search implementation for task templates? I'm not much familiar with our patterns and best practices, e.g., breaking the string into tokens/keywords. Also, should we use vocabulary for this? The idea is to search matches in the title and description/headnote. Ideally grouping and showing matches in the title first.

api/v1/editTaskTemplates/formRequests/AddTaskTemplate.php
```diff
@@ -31,7 +31,12 @@ public function authorize(): bool
     public function rules(): array
     {
         $contextId = Application::get()->getRequest()->getContext()->getId();
-        $stageIds = array_keys(Application::getApplicationStages());
+
+        // build a list of allowed stage IDs from values
+        $stages = Application::getApplicationStages();
+        $stageIds = array_values(array_unique(array_map('intval', array_values((array) $stages))));
```
bozana: $stages is array, so no need for extra cast?
Why do we need array_unique? 
Why do we need array_values again?
(here and below)

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,71 @@
...
+                ? filter_var($this->input('include'), FILTER_VALIDATE_BOOLEAN)
+                : $this->input('include', null),
+
+            'userGroupIds' => $this->has('userGroupIds')
+                ? array_values(array_map('intval', (array) $this->input('userGroupIds', [])))
+                : $this->input('userGroupIds', null),
+
+            'stageId' => $this->has('stageId')
+                ? (is_null($stageId) ? null : (int) $stageId)
+                : $this->input('stageId', null),
```
bozana: This looks the same as $stageId

api/v1/editTaskTemplates/formRequests/UpdateTaskTemplate.php
```diff
@@ -0,0 +1,71 @@
...
+
+            'type' => $this->has('type')
+                ? (is_null($type) ? null : (int) $type)
+                : $this->input('type', null),
```
bozana: This looks the same as $type

api/v1/editTaskTemplates/resources/TaskTemplateResource.php
```diff
@@ -31,6 +31,7 @@ public function toArray(Request $request)
             'include' => (bool) $this->include,
             'dueInterval' => $this->dueInterval,
             'description' => $this->description,
+            'type' => (int) $this->type,
```
bozana: The same line exists above in the line 28

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -93,6 +98,7 @@ public function add(AddTaskTemplate $illuminateRequest): JsonResponse
                 'include' => $validated['include'] ?? false,
                 'description' => $validated['description'] ?? null,
                 'dueInterval' => $validated['dueInterval'] ?? null,
+                'type' => (int) $validated['type'],
```
bozana: Then the type in the line 94 above is not needed any more

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -128,6 +137,20 @@ public function getMany(Request $request): JsonResponse
...
+                case 'search':
+                    $collector->filterBySearch((string) $val);
+                    break;
+                case 'title':
+                    $collector->filterByTitleLike((string) $val);
+                    break;
+                case 'type':
+                    if (is_numeric($val)) {
+                        $type = (int) $val;
+                        if (in_array($type, [EditorialTaskType::DISCUSSION->value, EditorialTaskType::TASK->value], true)) {
```
bozana: Should maybe `array_column(EditorialTaskType::cases(), 'value')` be used -- as above -- to stay flexible -- for example in case additional task type is added in the future?

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -128,6 +137,20 @@ public function getMany(Request $request): JsonResponse
...
         foreach ($queryParams as $param => $val) {
             switch ($param) {
+                case 'search':
+                    $collector->filterBySearch((string) $val);
+                    break;
+                case 'title':
+                    $collector->filterByTitleLike((string) $val);
+                    break;
+                case 'type':
+                    if (is_numeric($val)) {
```
bozana: Is this needed?

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -156,4 +179,78 @@ public function getMany(Request $request): JsonResponse
...
+        DB::transaction(function () use ($template, $data) {
+            $userGroupIds = Arr::pull($data, 'userGroupIds');
+            $updates = collect($data)
+                ->only(['stageId', 'title', 'include', 'type', 'description', 'dueInterval'])
+                ->mapWithKeys(function ($v, $k) {
+                    return match ($k) {
+                        'stageId' => ['stageId' => (int) $v],
+                        'include' => ['include' => (bool) $v],
+                        'type' => ['type' => (int) $v],
+                        'dueInterval' => ['dueInterval' => $v],
```
bozana: dueInterval seems could also go into default below

api/v1/editTaskTemplates/PKPEditTaskTemplateController.php
```diff
@@ -156,4 +179,78 @@ public function getMany(Request $request): JsonResponse
...
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        DB::transaction(function () use ($template) {
+            // Pivot/settings rows cascade via FKs defined in migration
+            $template->delete();
+        });
+
+        return response()->json([], Response::HTTP_OK);
```
bozana: Other API delete endpoints always return the deleted object's properties. I am just asking so I know for the future how should it be actually?

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+     * Scope: filter by title LIKE
+     */
+    public function scopeFilterByTitleLike(Builder $query, string $title): Builder
+    {
+        $title = trim($title);
+        if ($title === '') {
+            return $query;
+        }
+        $needle = '%' . str_replace(['%', '_'], ['\\%', '\\_'], mb_strtolower($title)) . '%';
+        return $query->whereRaw('LOWER(title) LIKE ?', [$needle]);
```
bozana: So here 'lower(title) LIKE lower('%{$needle}%') needs to be done.
I am not sure, but I think also here for `whereRaw` the same security consideration needs to be done as here for DB:raw: https://github.com/pkp/pkp-lib/commit/35e3c87441dc9742779178165636353b30d78313 -- to escape the search string properly using pdo->quote()

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+    /**
+     * Scope: filter by title LIKE
+     */
+    public function scopeFilterByTitleLike(Builder $query, string $title): Builder
+    {
+        $title = trim($title);
+        if ($title === '') {
+            return $query;
+        }
+        $needle = '%' . str_replace(['%', '_'], ['\\%', '\\_'], mb_strtolower($title)) . '%';
```
bozana: `addcslashes($title, '%_');` can be used.
Also 'to lower' should be done by the DB function so that is the same logarithm for both title column values and search string

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+        }
+
+        $settingsTable = $this->getSettingsTable();
+        $pk = $this->getKeyName();
+        $selfTable = $this->getTable();
+
+        return $query->where(function (Builder $outer) use ($tokens, $settingsTable, $pk, $selfTable) {
+            foreach ($tokens as $tok) {
+                // escape % and _
+                $like = '%' . str_replace(['%', '_'], ['\\%', '\\_'], mb_strtolower($tok, 'UTF-8')) . '%';
```
bozana: the same as above for title

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+
+                $outer->where(function (Builder $q) use ($like, $settingsTable, $pk, $selfTable) {
+                    $q->whereRaw('LOWER(title) LIKE ?', [$like])
```
bozana: $like should be converted to lower here, also probably escaped
(here and below)

classes/editorialTask/Template.php
```diff
@@ -172,4 +167,67 @@ public function promote(Submission $submission): Task
...
+                    ->orWhereRaw('LOWER(description) LIKE ?', [$like])
+                    ->orWhereExists(function ($sub) use ($like, $settingsTable, $pk, $selfTable) {
```
bozana: `Builder $sub`  -- type Builder?

bozana (commented):
Hi @Hafsa-Naeem and @Vitaliy-1,
I went through the changes and left some comments.
Thanks!

bozana:
Hi @Vitaliy-1, oh I first now see your message above.
I think we have never used vocabularies, so I think this is not needed.
Regarding "The idea is to search matches in the title and description/headnote. Ideally grouping and showing matches in the title first.": hmmm... I am not sure how to implement ranking, would need to think about it...

classes/editorialTask/Template.php
Hafsa-Naeem: @Vitaliy-1 I checked the PKP codebase for existing helpers around tokenizing search phrases, but didn't find a reusable helper for this. Other search-related code appears to split phrases inline in a similar way.
For now I've kept the simple `preg_split('/\s+/', ...)` implementation here to stay consistent. If we later introduce a shared helper for free-text search tokens, I can switch this scope to use it.

api/v1/editTaskTemplates/formRequests/TaskTemplateRequestTrait.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    protected function getStageIds(): array
+    {
+        $stages = Application::getApplicationStages();
+
+        return array_values(
+            array_unique(
+                array_map('intval', array_values((array) $stages))
+            )
```
bozana: $stages is array, so no need for extra cast?
Why do we need array_unique?
Why do we need array_values again?

classes/editorialTask/Template.php
```diff
@@ -172,4 +168,100 @@ public function promote(Submission $submission): Task
...
+        }
+
+        if ($keywords) {
+            $settingsTable = $this->getSettingsTable();
+            $pk = $this->getKeyName();
+            $selfTable = $this->getTable();
+
+            $query->where(function (Builder $outer) use ($keywords, $settingsTable, $pk, $selfTable) {
+                foreach ($keywords as $tok) {
+                    $like = '%' . str_replace(['%', '_'], ['\\%', '\\_'], mb_strtolower($tok, 'UTF-8')) . '%';
```
bozana: addcslashes()  and LOWER from DB?

classes/editorialTask/Template.php
```diff
@@ -172,4 +168,100 @@ public function promote(Submission $submission): Task
...
+
+                    $outer->where(function (Builder $q) use ($like, $settingsTable, $pk, $selfTable) {
+                        $q->whereRaw('LOWER(title) LIKE ?', [$like])
```
bozana: LIKE LOWER(?) here and below

classes/editorialTask/Template.php
```diff
@@ -172,4 +168,100 @@ public function promote(Submission $submission): Task
...
+        $title = trim($title);
+        if ($title === '') {
+            return $query;
+        }
+
+        // escape LIKE wildcards in the user input, then wrap with %
+        $needle = '%' . addcslashes($title, '%_') . '%';
+
+        // use LOWER() on both sides so DB applies the same case-folding
+        return $query->whereRaw('LOWER(title) LIKE LOWER(?)', [$needle]);
```
bozana, line 193: just for me to know for the future: does this whereRaw needs the same security consideration as here for DB:raw: https://github.com/pkp/pkp-lib/commit/35e3c87441dc9742779178165636353b30d78313 -- to escape the search string properly using pdo->quote()

Vitaliy-1: I'd say not, as we are passing binding separately and don't perform any SQL operations with them inside the bindings scope. If we passed something like 
```php
$query->whereRaw('LOWER(title) LIKE LOWER(?)', DB:raw("LOWER ('%{$needle}%')");
```
then yes, we are performing a query without bindings before passing it as a binding itself

api/v1/editTaskTemplates/formRequests/TaskTemplateRequestTrait.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    protected function getContextId(): int
+    {
+        return Application::get()->getRequest()->getContext()->getId();
+    }
+
+    protected function getStageIds(): array
+    {
+        $stages = Application::getApplicationStages();
+        return array_map('intval', array_values($stages));
```
Vitaliy-1: I think that there already was a comment regarding this, do we  need `array_values` and `array_map` here. I mean  `Application::getApplicationStages()` already returns non-associative array

Hafsa-Naeem: @Vitaliy-1 Yes! I'll update this accordingly.

Issue #11975, opened 2025-10-27
[Templating] Add mix use to Blade/Smarty templates for view templating


touhidurabir:
**Description:**
With the merge of https://github.com/pkp/pkp-lib/issues/9968, now it's possible to use blade template view/component along with existing smarty templates . 

However as most of the templates are already in smarty, effective usage of blade template require to convert all smarty template to blade template. While this the best approach, it is also quite time consuming task and a blocker of using blade for new template views where it need to include an existing one without duplication . To address this issue, need a solution so that blade or smarty can use each other . 

Most important factor of this usage is for the existing theme plugin to use a newly added blade view in the existing smarty template . Same also goes for the main apps . 

The addition of this feature will prevent the duplication of every single template (both in blade and smarty) and provide a smoother transition from smarty to blade with backward compatibilities. 

**What application are you using?**
  OJS, OMP or OPS version main(upcoming 3.6)

### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/11980
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5162 [TEST ONLY]

touhidurabir:
@jardakotesovec can you check the PRs at https://github.com/pkp/pkp-lib/issues/11975#issue-3555951348 to see if this cover up all the requirements .

jardakotesovec:
@touhidurabir Hi Touhidur, 
that looks like being able to include smarty file from blade file, right?

That might be useful as well. But to be able to start migrating out existing tpl files in the ojs/pkp-lib I think it would be more helpful if its possible to reference blade from smarty. Typical use case would be custom theme which is using some of the tpl files from the ojs/pkp-lib.  And if we could migrate these to blade and still make possible from someone to easily update his theme just to use the blade files that I think would make possible to convert tpl to blade files and not keeping both at the same time.

touhidurabir:
> [@touhidurabir](https://github.com/touhidurabir) Hi Touhidur, that looks like being able to include smarty file from blade file, right?
> 
[...]

PR #11980, opened 2025-10-28
pkp/pkp-lib#11975 Allow mix of blade/smarty templates

touhidurabir:
for pkp/pkp-lib#11975 .

### DIFF
--- a/classes/template/PKPTemplateManager.php
+++ b/classes/template/PKPTemplateManager.php
@@ -2542,29 +2542,41 @@ class PKPTemplateManager extends Smarty
      *
      * Custom Smarty function for injecting a blade view in a smarty template
      *
-     * @param array $params associative array
+     * IMPORTANT: The file path must follow strict Blade view format:
+     * - Use dot notation: 'path.to.template' or 'namespace::path.to.template'
+     * - NO slashes, extensions, or path traversal patterns
+     *
+     * Valid: 'frontend.pages.article', 'app::submissions.view'
+     * Invalid: 'frontend/pages/article', 'article.blade.php', '../config'
+     *
+     * @param array $params associative array with 'file' key
      * @param Smarty|null $smarty
      *
      * @return string The compiled content of the blade view
-     * 
-     * @throws Exception If the file parameter is missing or the blade view does not exist  
+     *
+     * @throws Exception If validation fails or view does not exist
      */
     public function smartyIncludeBlade($params, $smarty = null): string
     {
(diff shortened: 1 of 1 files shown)


classes/template/PKPTemplateManager.php
```diff
@@ -410,6 +410,9 @@ public function initialize(PKPRequest $request)
         // load NavigationMenu Areas from context
         $this->registerPlugin('function', 'load_menu', $this->smartyLoadNavigationMenuArea(...));
 
+        // load a blade view in smarty templates
+        $this->registerPlugin('function', 'load_blade_view', $this->smartyLoadBladeView(...));
```
jardakotesovec: Maybe just last name adjustments - include_blade ? I think it helps to have expectation that it does work same as `{include ...}`, just loading blade template instead.

touhidurabir: Done .

classes/template/PKPTemplateManager.php
```diff
@@ -2504,6 +2507,57 @@ public function smartyLoadNavigationMenuArea($params, $smarty = null)
...
+            throw new Exception('file parameter is missing in {include_blade}');
+        }
+
+        $file = $params['file'];
+
+        if (str_contains($file, '..') || str_contains($file, '\\') || str_starts_with($file, '/')) {
+            throw new Exception("Invalid file path: path traversal or absolute paths not allowed in given blade file {$file}");
+        }
+
+        $file = str_replace(['/', '.blade.php', '.blade'], ['.', '', ''], $file);
```
asmecher: We shouldn't be coding for `.blade.php` files anywhere; they should always be just `.blade` (at least until we move the codebase outside the web root).

touhidurabir: @asmecher this is not providing any support for `.blade.php` but sanitizing the path in proper format by removing any extension and `/` . But now that you mentioning this, I think rather than we sanitizing, we should be strict and throws exception so that developer is forced to maintain/pass proper path format . 

@jardakotesovec do you think it will be too strict if we force developer to have like this 
```tpl
{ include_blade file='path.to.blade.template' params1=$params1 params2=$params2 ...}
```
e.g. no `/` and no extension for `file` like we have for `.tpl` files. I was trying to be bit relax here but probably better to be strict here . 

BTW, if we fully stop the support of `.blade.php` , we need to handled it in service provider and have throws exception with file extension that has `.blade.php` . I will see what I can do .

PR #11985, opened 2025-10-29
pkp/pkp-lib#11913 Update migration for postgres

ewhanson:


ewhanson:
Hey @Vitaliy-1, could you review this? It addresses the issue you had with the migration with postgres. Thanks!

classes/migration/install/ReviewsMigration.php
```diff
@@ -34,7 +34,7 @@ public function up(): void
             $table->index(['submission_id'], 'review_rounds_submission_id');
 
             // Foreign key constraint and index depend on publication table, added in app-specific migration
-            $table->bigInteger('publication_id');
+            $table->bigInteger('publication_id')->default(0);
```
Vitaliy-1: I think we can avoid adding this to the installation script

classes/migration/upgrade/v3_6_0/I11913_AssociatePublicationsWithReviewRounds.php
```diff
@@ -27,7 +27,7 @@ class I11913_AssociatePublicationsWithReviewRounds extends Migration
     public function up(): void
     {
         Schema::table('review_rounds', function (Blueprint $table) {
-            $table->bigInteger('publication_id');
+            $table->bigInteger('publication_id')->default(0);
```
Vitaliy-1, line 30: If data is consistent, it should be safe to remove the default value once publication ID migration is complete

Issue #11986, opened 2025-10-31
Right click menu on TinyMCE fields only has 'link'


MrRob100:
## Issue Description
When right clicking in a TinyMCE field for example on an email body, only 'link' is available in the toolset. It would be easier to construct and edit text if the right click menu had other tools such as copy, paste, select all etc.

3.6 https://github.com/pkp/pkp-lib/pull/11989
3.5 https://github.com/pkp/pkp-lib/pull/11988, https://github.com/pkp/ui-library/pull/792
3.4 https://github.com/pkp/pkp-lib/pull/11987, https://github.com/pkp/ui-library/pull/790

<img width="416" height="324" alt="Image" src="https://github.com/user-attachments/assets/38f9db9f-cf26-4652-a8d3-2200987dd2c0" />

asmecher:
Thanks, @MrRob100! @jardakotesovec, could you review this one?

jardakotesovec:
@MrRob100 Hi Rob, sorry for delay.
We discussed this internally and agree that have just "Link" on right-click menu is not useful (its also available in toolbar). But rather than adding more - we would like to go in direction of disabling it - so users have access to their native menu with copy/paste functionality but also things like spellchecking etc..    Same approach we will follow with new Editor in 3.6, which will be for the article body.

So technically thats just adjusting tinyMce configuration with
```
  contextmenu: false,
```

The tedious part of the changes is to update pkp.min.js with the lib/pkp/tools/buildjs.sh so it picks up the change from SiteHandler.js. And thats needed for ojs/omp/ops repos.

Let me know if you would have some capacity to help with that. Otherwise I will get it on my list and hopefully get to it sometime soon. 

Best might be create PRs for one version first so I can just double check the PRs before spending time on the other versions.

jardakotesovec:
And now I noticed that we have issue opened with the same request of removing context menu - https://github.com/pkp/pkp-lib/issues/10862

Closing this one..  will make the change now (of disabling context menu) while I am in submodule spree..

PR #11989, opened 2025-10-31
Right click menu tools added for TinyMCE

MrRob100:
https://github.com/pkp/pkp-lib/issues/11986 

Screenshot 2025-10-31 at 12 42 47


asmecher:
@MrRob100, there are two places where TinyMCE is initialized:
- `lib/pkp/js/controllers/SiteHandler.js` (our legacy JS stuff)
- `lib/ui-library/src/components/Form/fields/FieldRichTextarea.vue` (newer parts of the UI)

I would think that the adjustment has to be made to both places. Could you check and adjust? You can try the old-style TinyMCE on the user profile forms; the new-style one is used e.g. in journal setup.

Issue #12009, opened 2025-11-10
Create an API endpoint to get a list of task participants


Vitaliy-1:
**Description:**
The request to the API endpoint should return a list of potential participants of the task/discussion. It includes users that have stage assignments and review assignments. Below several important points

- Reviewers should be included only in tasks/discussions in the review stage
- Anonimize (don't include) reviewers for authors in case of blinded review assignments. 
- Anonimize (don't include) authors for reviewers in case of double-blind reviews
- Don't include other reviewers for users with a single reviewer role. 
- Reviewers can participate in a new task or discussion if they have accepted or completed review assignment, even if associated submission is already published. 

**Solution:** 
Create a new endpoint in the current Task API controller, which will handle described above edge cases

**What application are you using?**
  OJS and OMP main branch

Vitaliy-1:
PR pkp-lib: https://github.com/pkp/pkp-lib/pull/12010
OJS: https://github.com/pkp/ojs/pull/5171

Vitaliy-1:
@ewhanson, can you take a look at the PR?

PR #12010, opened 2025-11-10
pkp/pkp-lib#12009 Add API endpoint to get a list of potential task pa…

Vitaliy-1:
…rticipants

### DIFF
--- a/api/v1/submissions/tasks/EditorialTaskController.php
+++ b/api/v1/submissions/tasks/EditorialTaskController.php
@@ -591,6 +591,23 @@ class EditorialTaskController extends PKPBaseController
             ], Response::HTTP_FORBIDDEN);
         }
 
+        // In a review stage, if the user is a reviewer and has no other roles, check if there is at least 1 accessible review assignment
+        if ($isReviewStage && $roleReviewer && $reviewerIsOnlyRole) {
+            $accessibleReviewAssignments = Repo::reviewAssignment()->getCollector()
+                ->filterBySubmissionIds([$submission->getId()])
+                ->filterByStageId($stageId)
+                ->filterByReviewerIds([$currentUser->getId()])
+                ->filterByIsAccessibleByReviewer(true)
+                ->getMany();
+
+            // The reviewer has only review assignments that were declined or cancelled
+            if ($accessibleReviewAssignments->isEmpty()) {
+                return response()->json([
+                    'error' => __('api.403.forbidden'),
+                ], Response::HTTP_FORBIDDEN);
+            }
+        }
+
         $currentUserHasDoubleBlindReview = false;
         $reviewUserGroup = UserGroup::with('userUserGroups')->where('role_id', Role::ROLE_ID_REVIEWER)
             ->where('context_id', $submission->getData('contextId'))
@@ -604,7 +621,6 @@ class EditorialTaskController extends PKPBaseController
                 ->filterBySubmissionIds([$submission->getId()])
                 ->filterByStageId($stageId)
                 ->filterByReviewerIds([$currentUser->getId()])
-                ->filterByActive(true)
                 ->getMany();
 
             if ($currentUserReviewAssignments->isNotEmpty()) {
@@ -699,7 +715,7 @@ class EditorialTaskController extends PKPBaseController
         $reviewAssignments = Repo::reviewAssignment()->getCollector()
             ->filterBySubmissionIds([$submission->getId()])
             ->filterByStageId($stageId)
-            ->filterByActive(true)
+            ->filterByIsAccessibleByReviewer(true)
             ->getMany();
 
         // Filter review assignments by latest review round by reviewer
--- a/classes/submission/reviewAssignment/Collector.php
+++ b/classes/submission/reviewAssignment/Collector.php
@@ -47,6 +47,7 @@ class Collector implements CollectorInterface, ViewsCount
     public ?array $reviewerIds = null;
     public ?bool $isLastReviewRoundByReviewer = false;
     public bool $isLastReviewRound = false;
+    public bool $isAccessibleByReviewer = false;
     public ?int $count = null;
     public ?int $offset = null;
     public ?array $reviewMethods = null;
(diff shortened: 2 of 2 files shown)


api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -547,4 +555,193 @@ public function deleteNote(Request $illuminateRequest): JsonResponse
...
+            $currentUserReviewAssignments = Repo::reviewAssignment()->getCollector()
+                ->filterBySubmissionIds([$submission->getId()])
+                ->filterByStageId($stageId)
+                ->filterByReviewerIds([$currentUser->getId()])
+                ->filterByActive(true)
+                ->getMany();
+
+            if ($currentUserReviewAssignments->isNotEmpty()) {
+                $currentUserHasDoubleBlindReview = (bool) $currentUserReviewAssignments->search(fn (ReviewAssignment $reviewAssignment) =>
+                    $reviewAssignment->getReviewMethod() == ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS);
```
ewhanson, line 628: I don't know if it matters since they should both be `int`s but should this be checked with `===` instead?

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -547,4 +555,193 @@ public function deleteNote(Request $illuminateRequest): JsonResponse
...
+        $users = collect([$currentUser->getId() => $currentUser]);
+        $stageAssignments = StageAssignment::with(['userGroup', 'userGroup.userUserGroups'])
+            ->withSubmissionIds([$submission->getId()])
+            ->withStageIds([$stageId])
+            ->get();
+
+        foreach ($stageAssignments as $stageAssignment) {
+            if ($isReviewStage && $roleReviewer && $currentUserHasDoubleBlindReview) {
+                // Anonymize author if the current user is a reviewer with double-blind review
+                if ($stageAssignment->userGroup->roleId == Role::ROLE_ID_AUTHOR) {
```
ewhanson, line 645: Same with this `==`.

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -547,4 +555,193 @@ public function deleteNote(Request $illuminateRequest): JsonResponse
...
+        // Form the list of all user participants, stage assignments, review assignments and corresponding user groups for the Participant resource
+        $users = collect([$currentUser->getId() => $currentUser]);
+        $stageAssignments = StageAssignment::with(['userGroup', 'userGroup.userUserGroups'])
+            ->withSubmissionIds([$submission->getId()])
+            ->withStageIds([$stageId])
+            ->get();
+
+        foreach ($stageAssignments as $stageAssignment) {
+            if ($isReviewStage && $roleReviewer && $currentUserHasDoubleBlindReview) {
+                // Anonymize author if the current user is a reviewer with double-blind review
```
ewhanson, line 644: It may not matter, but it's not completely clear to me going through the code if the authors are being anonymized (i.e. they're included but with no identifiable information) or if they're rather completely excluded (i.e. not available as a participant at all). It seems like the code is excluding them outright, but wanted to double-check because the reference to anonymizing the author were a little confusing to follow.

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -547,4 +555,193 @@ public function deleteNote(Request $illuminateRequest): JsonResponse
...
+            if ($reviewAssignment->getReviewRoundId() > $filteredReviewAssignments[$reviewerId]->getReviewRoundId()) {
+                $filteredReviewAssignments[$reviewerId] = $reviewAssignment;
+            }
+        }
+
+        // Retrieve participants from review assignments
+        $includedReviewAssignments = collect();
+        foreach ($filteredReviewAssignments as $reviewAssignment) { /** @var ReviewAssignment $reviewAssignment */
+
+            // Check whether participant should be anonymized
```
ewhanson, line 741: Same comment about "anonymized" as above.

api/v1/submissions/tasks/resources/EditorialTaskParticipantResource.php
```diff
@@ -78,23 +78,33 @@ public function toArray(Request $request, array $data = []): array
...
+
+            if ($roleId == Role::ROLE_ID_REVIEWER) {
+                $role['reviewMethod'] = $reviewMethod;
+            }
+            $groupedRoles[] = $role;
         }
 
         return [
-            'id' => $this->id,
+            'id' => $this?->id,
```
ewhanson, line 107: In which cases would `$this` be potentially null?

ewhanson (changes requested):
Thanks @Vitaliy-1. Just a few questions, mostly around how things are anonymized. Otherwise, looks good!

api/v1/submissions/tasks/EditorialTaskController.php
Vitaliy-1: I recall Alec's comment regarding problems with strict comparisons in some configurations...

api/v1/submissions/tasks/EditorialTaskController.php
Vitaliy-1: Yes, completely excluding from this endpoint. I don't think we need to expose their presence at all.

api/v1/submissions/tasks/resources/EditorialTaskParticipantResource.php
Vitaliy-1: I use the same resource for expose data that isn't yet recorded in the database. Task Participant object is created but not yet saved, so, doesn't have an ID.

api/v1/submissions/tasks/EditorialTaskController.php
ewhanson: Ah, I don't recall that. No problem to keep it that way then. @asmecher, what was the problem with strict comparisons?

asmecher: @ewhanson, there are some deployments where a combo of MySQL/MariaDB and possibly PHP configuration result in numeric data coming back from the DB (and into our `DataObject`s) as `string` data rather than `int`s. On servers in that config, wherever we have a `===` test, it fails to match. I wasn't 100% able to pin down the cause, but there are some details here:

https://github.com/pkp/pkp-lib/issues/10034

Generally speaking, this will go away as we make better use of typehinted function declarations and Eloquent, but that'll be a gradual change. Meanwhile we'll have to be careful not to use strict comparisons that fall afoul of this (unusual but exists-in-the-wild) configuration.

Issue #12002, opened 2025-11-05
[Multi-Journal] | Make a new submission - Dashboard, when a user doesn't have a role in a journal they are unable to make a new submission


Tribunal33:
**Description:**
- Include specific context details that have been discovered that surround this developer task.
For users without assigned roles in multi-journal installs there should be a way for them to create a new submission.  The simplest way for a user without a role to start a new submission is with the About > Submission > Make a new submission

After creating this submission they will automatically be assigned a role as Author.  This is also accessible via the Edit Profile > Roles page.  The problem is that some users are unable to figure out that they need to assign a role first and becomes a burden.  

**History:**
- OJS 3.2 this used to work
- OJS 3.3 this still works but did get an error at the end of the submission process.  Appears related to email.  Didn't block being added as an author role and didnt block creating the submission.
<img width="356" height="116" alt="Image" src="https://github.com/user-attachments/assets/8dea4277-6509-487f-b103-892aa9c8f227" />

- OJS 3.4 **There is an issue** when a users selects About > Submission > Start new submission link (aka the submission form process) will not allow the user to proceed beyond the first step.

<img width="1015" height="92" alt="Image" src="https://github.com/user-attachments/assets/b395ffea-ab39-4275-a5c5-826c4f9c80e2" />

- In OJS 3.5 no error seen and roleless user is able create a submission.  (might need more investigation into why it started to work and if the 3.4 error was intentional.)
- **Update for 3.5.**  There is a niche use case with roleless user but it is not seen initially when you first submit to one of the other journals.  You have to have to self-remove the role and afterwards will not be able to submit.  Additionally there is a query that seems to rely on other users having an author role to create a ghost author role. 

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
1. Prerequisites: Have a multi-journal install
2. Login to the journal that the user has no roles in
3. Select About > Submissions > Make a new submission
4. Fill out the new submission form correctly
5. Complete the new submission
6. Under Edit profile > Roles you should see the author role selected

**Solution:** What the proposed solutions are and additional affected areas
I believe this fix is for 3.4 only but there should be some investigation into why this was initially locked down (GDPR initiative perhaps?) Or some other reason.  And then figure out why in 3.5 there was no limitation. 

**A/C:**
- [x] The roleless user should be able to complete a new submission and be marked as an author
[...]

PR #12012, opened 2025-11-11
pkp/pkp-lib#12002 Relax role requirements for adding new submission

jardakotesovec:


api/v1/submissions/PKPSubmissionHandler.php
```diff
@@ -271,7 +272,13 @@ public function authorize($request, &$args, $roleAssignments)
 
         $this->addPolicy(new UserRolesRequiredPolicy($request), true);
 
-        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+        if ($routeName === 'add') {
+            // For 'add' endpoint, mark role assignments as checked since the add() method
```
Vitaliy-1, line 276: I see that we have a `PKPSiteAccessPolicy` class, I'm wondering if it suits here

jardakotesovec: @Vitaliy-1 Thanks for checking.

That PKPSiteAccessPolicy seems to be specifically designed for endpoints which are outside of the context.

First thing I was double checking was whether it still requires for user to be logged in and thats covered in the UserRolesRequiredPolicy, which is applied above.

So I think we should be covered ok here.

PR #12016, opened 2025-11-12
pkp/pkp-lib#11583 strict mode update and removed PKP_STRICT_MODE const

touhidurabir:
for pkp/pkp-lib#11583  .

### DIFF
--- a/classes/metadata/MetadataDataObjectAdapter.php
+++ b/classes/metadata/MetadataDataObjectAdapter.php
@@ -176,10 +176,8 @@ class MetadataDataObjectAdapter extends PersistableFilter
 
     /**
      * Get the supported application entity (class) name
-     *
-     * @return string
      */
-    public function getDataObjectName()
+    public function getDataObjectName(): string
     {
         return $this->_dataObjectName;
     }
@@ -194,7 +192,6 @@ class MetadataDataObjectAdapter extends PersistableFilter
     {
         if (is_null($this->_dataObjectClass)) {
             $dataObjectName = $this->getDataObjectName();
-            assert(!is_null($dataObjectName));
             $dataObjectNameParts = explode('\\', $dataObjectName);
             $this->_dataObjectClass = array_pop($dataObjectNameParts);
         }
@@ -340,7 +337,6 @@ class MetadataDataObjectAdapter extends PersistableFilter
     public function &instantiateDataObject()
     {
         $dataObjectName = $this->getDataObjectName();
-        assert(!is_null($dataObjectName));
         $dataObject = new $dataObjectName;
         return $dataObject;
     }
--- a/classes/metadata/MetadataTypeDescription.php
+++ b/classes/metadata/MetadataTypeDescription.php
@@ -83,7 +83,6 @@ class MetadataTypeDescription extends ClassTypeDescription
     {
         // Configure the parent class type description
         // with the expected meta-data class.
-        // parent::parseTypeName('lib.pkp.classes.metadata.MetadataDescription');
         parent::parseTypeName(MetadataDescription::class);
 
         // Split the type name into class name and assoc type.


classes/core/PKPContainer.php
```diff
@@ -671,11 +684,51 @@ public function unsetRunningUnitTests(): void
...
         return 'PKP\\';
     }
+
+    /**
+     * Register class consts as global consts
+     */
+    public function registerGlobalConstants(string $classNamespacePath, array $constants): void
+    {
+        if ($this->getApplicationStrictModeStatus()) {
+            throw new Exception('Registering class const as global const in strict mode in now allowed');
```
asmecher:
```suggestion
            throw new Exception('Registering class const as global const in strict mode is not allowed');
```

classes/core/PKPContainer.php
```diff
@@ -671,11 +684,51 @@ public function unsetRunningUnitTests(): void
...
+        }
+
+        if (!class_exists($classNamespacePath)) {
+            throw new Exception(sprintf('Given class %s does not exists', $classNamespacePath));
```
asmecher:
```suggestion
            throw new Exception(sprintf('Given class %s does not exist', $classNamespacePath));
```

classes/plugins/PluginRegistry.php
```diff
@@ -202,29 +202,50 @@ private static function instantiatePlugin(string $category, string $pluginName,
...
-                ? new $pluginClassName()
-                : static::deprecatedInstantiatePlugin($category, $pluginName);
+            if (class_exists($pluginClassName)) {
+                $plugin = new $pluginClassName(); /** @var Plugin $plugin */
+            } else {
+                $pluginIndexPath = app()->basePath()
+                    . DIRECTORY_SEPARATOR
+                    . 'plugins'
+                    . DIRECTORY_SEPARATOR
+                    . "{$category}"
```
asmecher:
```suggestion
                    . $category
```

classes/plugins/PluginRegistry.php
```diff
@@ -202,29 +202,50 @@ private static function instantiatePlugin(string $category, string $pluginName,
...
+                    . DIRECTORY_SEPARATOR
+                    . "{$pluginName}"
```
asmecher:
```suggestion
                    . $pluginName
```

classes/plugins/PluginRegistry.php
```diff
@@ -202,29 +202,50 @@ private static function instantiatePlugin(string $category, string $pluginName,
...
+                    . DIRECTORY_SEPARATOR
+                    . 'index.php';
+
+                if (!file_exists($pluginIndexPath)) {
+                    throw new Exception("fallback to plugin wrapper failed as not found at path {$pluginIndexPath} for plugin {$category}/{$pluginName}");
+                }
+
```
asmecher: This whole block is already wrapped in a `try`, so this error (or the `Error` that `require` throws just below) would both be caught and dealt with below anyway. I'd suggest just letting the `try`/`catch` handle a failure from `require`, and removing this check.

classes/metadata/MetadataDataObjectAdapter.php
```diff
@@ -341,7 +341,7 @@ public function &instantiateDataObject()
     {
         $dataObjectName = $this->getDataObjectName();
         assert(!is_null($dataObjectName));
```
asmecher: I'd suggest adding a strict `string` return typehint to the `getDataObjectName` function; then you can remove this assertion.

classes/metadata/MetadataTypeDescription.php
```diff
@@ -82,7 +83,8 @@ public function parseTypeName(string $typeName): bool
     {
         // Configure the parent class type description
         // with the expected meta-data class.
-        parent::parseTypeName('lib.pkp.classes.metadata.MetadataDescription');
+        // parent::parseTypeName('lib.pkp.classes.metadata.MetadataDescription');
```
asmecher: Remove commented code

Issue #12017, opened 2025-11-12
UserGroup::scopeWithUserUserGroupStatus does not work correctly, fix getting user that is active in a role


bozana:
The function UserGroup::scopeWithUserUserGroupStatus does not consider the user ID -- e.g. if the user has an active role. There is no relationship to the user there.
S. https://github.com/pkp/pkp-lib/blob/stable-3_5_0/pages/submission/PKPSubmissionHandler.php#L620-L624
which leads to this SQL:
```
SELECT *
FROM "user_groups"
WHERE "context_id" IN (1)
  AND EXISTS (
    SELECT *
    FROM "user_user_groups"
    WHERE "user_groups"."user_group_id" = "user_user_groups"."user_group_id"
      AND "user_id" IN (21)
  )
  AND EXISTS (
    SELECT *
    FROM "user_group_stage"
    WHERE "user_groups"."user_group_id" = "user_group_stage"."user_group_id"
      AND "stage_id" IN (1)
  )
  AND EXISTS (
    SELECT *
    FROM "user_user_groups"
    WHERE "user_groups"."user_group_id" = "user_user_groups"."user_group_id"
      AND (
        "date_start" <= '2025-11-10T18:44:51.030404Z'
        OR "date_start" IS NULL
      )
      AND (
        "date_end" > '2025-11-10T18:44:51.030404Z'
        OR "date_end" IS NULL
      )
  );
```

This function will be removed and instead of this function this pattern should be used:
```
$userGroups = UserGroup::query()
  ->withContextIds([$context->getId()])
  ->whereHas('userUserGroups', function (Builder $query) use ($user) {
    $query->withUserId($user->getId())->withActive();
  })
  ->get();
```
Also, the function Repo::userGroup()::userUserGroups(), that also needs to be fixed, can be used for the same purpose.

<hr>

AC to test this
- Test if there exists author roles in journal a new roleless user can still create a new submission and be granted author role after privacy step.
Previously to the fix, if there are no other users with an author role then a ghost author role is given to the first user to create a new submission but any other user without author role would not be able to complete a new submission.  Any user should be able to add a new submission even if they don't have the author role.  It will be given to them after the privacy consent step is completed.

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/12018
- https://github.com/pkp/ojs/pull/5175 (only submodule update)
- https://github.com/pkp/omp/pull/2165 (only submodule update)
[...]

PR #12018, opened 2025-11-12
pkp/pkp-lib#12017 fix getting user active roles, consider only active users at some places in the code

bozana:
s. https://github.com/pkp/pkp-lib/issues/12017

This PR for 3.5 is kept minimal. The PR for main contains some additional changes, one more commit, to standardize how active user roles are get.

### DIFF
--- a/controllers/grid/settings/roles/form/UserGroupForm.php
+++ b/controllers/grid/settings/roles/form/UserGroupForm.php
@@ -20,6 +20,7 @@ use APP\core\Application;
 use APP\core\Request;
 use APP\facades\Repo;
 use APP\template\TemplateManager;
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use PKP\core\JSONMessage;
 use PKP\db\DAORegistry;
 use PKP\facades\Locale;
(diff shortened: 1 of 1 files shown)


bozana:
Hi @Vitaliy-1, there are lots of places in the code that look for the active user role(s). In this PR, there are only those that need to be fixed. The other places are correct, but using different possibilities to get the same result, which I tried to standardize in the PR for main.
So, yes, maybe it would make sense for a global scope... If so, how would one look for ended user groups?
Else, maybe also in the way Jarda suggested, having scopeWithActiveUserIds()?
Thanks!

classes/userGroup/Repository.php
```diff
@@ -209,7 +213,6 @@ public function getUserUserGroupsContextIds(int $userId): Collection
     {
         return UserGroup::query()
             ->withUserIds([$userId])
-            ->withUserUserGroupStatus(UserUserGroupStatus::STATUS_ALL->value)
             ->masthead(true)
```
jardakotesovec, line 216: Here its removed because its only used for user merging, where we want to remove all?

bozana: Yes. This getUseruserGroupsContextIds should get all context IDs for all masthead user groups a user is in, no matter if active or ended.
It is actually only be used when merging users, and there also only in order to know if the masthead cache needs to be cleared. Theoretically I could remove this function too and move the code to that place where it is called... :thinking: ... sometimes I cannot remember why I did something in the way I did... :sweat_smile:

Issue #12035, opened 2025-11-17
[New Review Round] | Workflow Section - Review - Recomand-only users cannot add a new round


salmanm2003:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

OJS editorial roles have the option for This role is only allowed to recommend a review decision and will require an authorised editor to record a final decision. When enabled, the relevant editor will only be able to make recommendations to other assigned editors rather than make the decision themselves.

From 3.4 onwards, if a role/user is marked as This role is only allowed to recommend a review decision and will require an authorised editor to record a final decision. then the relevant editor is NOT permitted to start a new round of review. This was not the case in 3.3.

### Steps to Reproduce

1- Go to Submission > Workflow > Review.
2- Assign an editor and enable Assignment privileges: “This participant is only allowed to recommend an editorial decision and will require an authorised editor to record editorial decisions.”
3- Add a new review round.
4- OJS does not allow the assigned editor (with recommend-only privileges) to access this feature.

### Expected Result

The editor with recommend-only privileges should be able to create a new review round, rather than needing permission from a senior editor.

### Actual Result

OJS 3.4 and later versions do not allow recommend-only editors to add a new review round.

### Application Version

OJS 3.4 and later

salmanm2003:
Hi @Vitaliy-1,
Regarding our recent conversation about allowing recommend-only editors to add a new review round, I’ve submitted above the PR for OJS 3.4 for review. Once it’s approved, I will apply the same changes to versions 3.5 and 3.6.

Vitaliy-1:
@salmanm2003, thanks, looks good! Could you also make a PR with a submodule update to OJS stable branch to ensure that tests are passing?

salmanm2003:
Thanks @Vitaliy-1, the PR has already been made.

asmecher:
Thanks, @salmanm2003 and @Vitaliy-1; I've [added a review](https://github.com/pkp/pkp-lib/pull/12037#pullrequestreview-3545535889).

Vitaliy-1:
Merged to 3.4 but requires also porting to 3.5 and main

salmanm2003:
> Merged to 3.4 but requires also porting to 3.5 and main

Thanks @Vitaliy-1, @asmecher  would it be possible to create a new tag that includes all these changes as well as the ones since 3.4.0-10?

asmecher:
[...]

PR #12037, opened 2025-11-17
pkp/pkp-lib#12035 (3.4.0) Fix issue where recommend-only users cannot add a new review round

salmanm2003:
This is for: [https://github.com/pkp/pkp-lib/issues/12035](https://github.com/pkp/pkp-lib/issues/12035)

### DIFF
--- a/classes/security/authorization/internal/DecisionAllowedPolicy.php
+++ b/classes/security/authorization/internal/DecisionAllowedPolicy.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file classes/security/authorization/internal/DecisionAllowedPolicy.php
  *
@@ -16,6 +17,7 @@
 
 namespace PKP\security\authorization\internal;
 
+use APP\decision\Decision;
 use APP\core\Application;
 use APP\facades\Repo;
 use PKP\db\DAORegistry;
@@ -69,16 +71,21 @@ class DecisionAllowedPolicy extends AuthorizationPolicy
                 if (!in_array($userGroup->getRoleId(), [Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR])) {
                     continue;
                 }
-                if (Repo::decision()->isRecommendation($decisionType->getDecision()) && $stageAssignment->getRecommendOnly()) {
+
+                $decision = $decisionType->getDecision();
+                if (
+                    !$stageAssignment->getRecommendOnly() ||
+                    $decision === Decision::NEW_EXTERNAL_ROUND ||
+                    Repo::decision()->isRecommendation($decision)
+                ) {
                     $isAllowed = true;
-                } elseif (!$stageAssignment->getRecommendOnly()) {
-                    $isAllowed = true;
+                    break;
                 }
 
                 // Check whether there is a decision that a recommending role can make on the stage the submission is in.
                 $recommendatorsAvailableDecisions = Repo::decision()
                     ->getDecisionTypesMadeByRecommendingUsers($submission->getData('stageId'));
-                
+
                 // if there is any decision that the recommending role is allowed to make, check if the current decision is within the allowed ones 
                 if (!empty($recommendatorsAvailableDecisions)) {
                     $matches = array_filter($recommendatorsAvailableDecisions, function ($decisionInArray) use ($decisionType) {
--- a/classes/security/authorization/internal/DecisionAllowedPolicy.php
+++ b/classes/security/authorization/internal/DecisionAllowedPolicy.php
@@ -17,8 +17,8 @@
 
 namespace PKP\security\authorization\internal;
 
-use APP\decision\Decision;
 use APP\core\Application;
+use APP\decision\Decision;
 use APP\facades\Repo;
 use PKP\db\DAORegistry;
 use PKP\security\authorization\AuthorizationPolicy;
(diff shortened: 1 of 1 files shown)


Vitaliy-1 (approved):

classes/security/authorization/internal/DecisionAllowedPolicy.php
```diff
@@ -73,6 +74,10 @@ public function effect()
                     $isAllowed = true;
                 } elseif (!$stageAssignment->getRecommendOnly()) {
                     $isAllowed = true;
+                } elseif ($stageAssignment->getRecommendOnly()) {
+                    if ($decisionType->getDecision() === Decision::NEW_EXTERNAL_ROUND) {
+                        $isAllowed = true;
+                    }
```
jonasraoni: I think all these `ifs` can be simplified into:

```suggestion
$decision = $decisionType->getDecision();
if (!$stageAssignment->getRecommendOnly() || 
    $decision === Decision::NEW_EXTERNAL_ROUND ||
    Repo::decision()->isRecommendation($decision)) {
    $isAllowed = true;
    break;
}
```

asmecher (changes requested):
I'm ambivalent about this change -- if you were asking me about whether recommend-only users should be able to start a new round, I'd probably say no (not that I'm particularly informed about this aspect of workflow); however, this is a regression from 3.3.0, probably an accidental one made during refactor of editorial decisions. So I'm OK either way.

However, we probably should stay consistent in this decision between internal and external reviews (for OMP).

classes/security/authorization/internal/DecisionAllowedPolicy.php
jonasraoni: In fact, after checking the other nearby lines, I think a lot of improvements should be added here:

- The only purpose of the `foreach` loop seems to **perhaps** set the `$isAllowed` to true, therefore, a `break` can be added once the value is set to true... 
- On the lines 62 and 66: the `else` isn't needed and can be removed, a decreased indentation simplifies reading code.
- Looks like the block ranging from line 84 until 96 has no dependency with the inner variables of the `foreach` loop, therefore, it doesn't need to be called many times (there's a query involved and this is expensive) and should be cached... (e.g. `$cached ??= (function (){...})();` or using an extra private method)

jonasraoni (changes requested):
I've never touched this piece of code, but I've left a couple of suggestions, which I think are good to handle while there's someone touching this.

salmanm2003:
@asmecher Since this is working fine in 3.4, could you please apply it to 3.4 for now and consider further improvements in later versions? I’ve just simplified the if statements as @jonasraoni  suggested, while the remaining enhancements can be addressed for 3.5 and beyond.

asmecher:
@salmanm2003, I'll leave that decision to @Vitaliy-1 as the workflow lead.

Vitaliy-1:
@salmanm2003, I'm ok with this change. For consistency, this also should be applied for the 3.5 and main branch. Can you make necessary PRs? Also, I see some additional changes in this PRs which refer to another issue. Should they be here?

salmanm2003:
@Vitaliy-1 The PR for 3.4 is now ready to be merged, and I have removed the additional commits that were not related to this issue.
While testing 3.5 and main, I found that both contain a bug that allows a recommend-only editor to add a new review round when editing editor details.
To reproduce: assign a new editor and set them as recommend-only. Then log in as that editor, go to Participants, click the three dots, and choose Edit. This changes the permissions and allows the editor to add a new review round.
I believe this issue should be fixed in 3.5 and main before enabling recommend-only editors to add new rounds. Therefore, could you please merge the stable-3_4_0 changes for now? We can address 3.5 and main later when it’s more convenient and after further discussion.

Screenshot 2026-01-23 at 01 01 33

salmanm2003:
@Vitaliy-1  Would it be possible to merge this into 3.4 soon please? I can then look into the bug separately for 3.5 and main.

classes/security/authorization/internal/DecisionAllowedPolicy.php
```diff
@@ -69,16 +71,21 @@ public function effect()
...
                     continue;
                 }
-                if (Repo::decision()->isRecommendation($decisionType->getDecision()) && $stageAssignment->getRecommendOnly()) {
-                    $isAllowed = true;
-                } elseif (!$stageAssignment->getRecommendOnly()) {
+
+                $decision = $decisionType->getDecision();
+                if (
+                    !$stageAssignment->getRecommendOnly() ||
+                    $decision === Decision::NEW_EXTERNAL_ROUND ||
```
Vitaliy-1, line 78: Can we treat OMP specific internal review the same way? See `NEW_INTERNAL_ROUND`.

Vitaliy-1 (changes requested):
Thanks, @salmanm2003. 
As per Alec's comment, I think we need one more minor change

classes/security/authorization/internal/DecisionAllowedPolicy.php
salmanm2003: Hi @Vitaliy-1 , The changes have been added for OMP as well.
Here is the PR for OMP: https://github.com/pkp/omp/pull/2312
Could you review it please and let me know if any issues?

salmanm2003: @Vitaliy-1 could you check this please and let me know if any issues?

Vitaliy-1 (approved):

classes/security/authorization/internal/DecisionAllowedPolicy.php
Vitaliy-1: Sorry, missed it, looks good!

salmanm2003: Thanks @Vitaliy-1, is it possible to have these changes released as a new tagged version?

PR #12038, opened 2025-11-17
pkp/pkp-lib#12002 Add author role on submit, not on submission form d…

jardakotesovec:
…isplay

### DIFF
--- a/api/v1/submissions/PKPSubmissionController.php
+++ b/api/v1/submissions/PKPSubmissionController.php
@@ -346,11 +346,8 @@ class PKPSubmissionController extends PKPBaseController
         });
 
         Route::post('', $this->add(...))
-            ->name('submission.add')
-            ->middleware([
-                self::roleAuthorizer(Role::getAllRoles()),
-            ]);
-    }
+            ->name('submission.add');
+        }
 
     /**
      * @copydoc \PKP\core\PKPBaseController::authorize()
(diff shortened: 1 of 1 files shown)


pages/submission/PKPSubmissionHandler.php
```diff
@@ -620,17 +620,15 @@ protected function getSubmitUserGroups(Context $context, User $user): Collection
...
+                ->whereHas('userUserGroups', function ($query) use ($user) {
+                    $query->withUserId($user->getId())->withActive();
+                })
                 ->get(); // For non-admin users, query for the groups tht give them access to the submission stage
 
         // Users without a submitting role or access to submission stage can submit as an
         // author role that allows self registration.
         // They are also assigned the author role
         if ($userGroups->isEmpty()) {
-            Repo::userGroup()->assignUserToGroup(
```
Vitaliy-1, line 630: Hmm, if the code already assigned the author permission in the API, why this part was added here? Is it just a permission aspect to the submission wizard?

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -604,17 +607,26 @@ public function add(Request $illuminateRequest): JsonResponse
...
+            $submitterUserGroupsQuery->withStageIds([WORKFLOW_STAGE_ID_SUBMISSION]);
+        }
+
+        $submitterUserGroups = $submitterUserGroupsQuery->get();
 
 
         $userGroupIdPropName = 'userGroupId';
 
         if (isset($params[$userGroupIdPropName])) {
+            error_log('add 1');
```
Vitaliy-1: Don't forget to remove debugging output

pages/submission/PKPSubmissionHandler.php
jardakotesovec: My understanding how things evolved is that originally (3.3, 3.4) we added Author role when the 'Add' api endpoints was hit. But than this issue came about https://github.com/pkp/pkp-lib/issues/10929 , where the issue was with our favourite production editor, which is used only be @Tribunal33 :-D.  And that was fixed here (loading create submission page) instead of in the original place.

So we have two places where author role is added in 3.5. This PR is trying to adjust it so its again just in one place, but it handles also production editor and also making it work for users who don't have any role.

Issue #12050, opened 2025-11-20
Custom API endpoint for plugins


touhidurabir:
**Description:**
Details and related discussions at https://github.com/pkp/pkp-lib/discussions/11991

**What application are you using?**
  OJS, OMP or OPS version 3.5.0 and main

touhidurabir:
@ewhanson can you please review the PRs at https://github.com/pkp/pkp-lib/issues/12050#issue-3645211202 . we can port to main after we finalise the approach for the 3.5.0 branch .

asmecher:
@ewhanson, now that 3.5.0-3 has been released, could you give this one another once-over and see if it's ready for merge?

ewhanson:
@asmecher, looks good to me. Ready for merge and porting to `main`.

asmecher:
@touhidurabir, please go ahead!

touhidurabir:
Merged the `pkp-lib` PRs for 3.5 and main . Tested with OJS 3.5/main and confirmed tests were passing for both branch .

PR #12051, opened 2025-11-20
pkp/pkp-lib#12050 Allow custom API routes for plugin

touhidurabir:
for pkp/pkp-lib#12050 .

### DIFF
--- a/classes/core/APIRouter.php
+++ b/classes/core/APIRouter.php
@@ -108,25 +108,23 @@ class APIRouter extends PKPRouter
         // Give the plugin a chance to register its API controllers
         Hook::run('APIHandler::endpoints::plugin', [$this]);
 
+        // If any plugin API controllers were registered and request path match with it
+        // that plugin api controller will be used to handle the request
+        if (!empty($this->registeredPluginApiControllers)) {
+            $requestPath = $request->getRequestPath();
+            foreach ($this->registeredPluginApiControllers as $handlerPath => $apiController) {
+                if ($this->matchesPluginHandlerPath($requestPath, $handlerPath)) {
+                    $handler = new APIHandler($apiController);
+                    $this->setHandler($handler);
+                    $handler->runRoutes();
+                    return;
+                }
+            }
+        }
+
         $sourceFile = $this->getSourceFilePath();
 
         if (!file_exists($sourceFile)) {
-
-            // If no match found for any core route,
-            // check for any plugin API controllers were registered and request path match with it
-            // that plugin api controller will be used to handle the request
-            if (!empty($this->registeredPluginApiControllers)) {
-                $requestPath = $request->getRequestPath();
-                foreach ($this->registeredPluginApiControllers as $handlerPath => $apiController) {
-                    if ($this->matchesPluginHandlerPath($requestPath, $handlerPath)) {
-                        $handler = new APIHandler($apiController);
-                        $this->setHandler($handler);
-                        $handler->runRoutes();
-                        return;
-                    }
-                }
-            }
-
             response()->json([
                 'error' => 'api.404.endpointNotFound',
                 'errorMessage' => __('api.404.endpointNotFound'),


classes/core/APIRouter.php
```diff
@@ -96,6 +102,20 @@ public function getEntity(): string
      */
     public function route(PKPRequest $request): void
     {
+        Hook::run('APIHandler::endpoints::plugin', [$this]);
```
touhidurabir, line 109: should we have a different hook name here as in future we may get a `plugin` entity which will conflict with this one  ?

ewhanson: What kind of future hook are you concerned this might conflict with? I think this should be okay.

ewhanson, line 109: This will need the corresponding `@hook` annotation in the doc block.

classes/core/APIRouter.php
```diff
@@ -116,6 +140,24 @@ public function route(PKPRequest $request): void
         $handler->runRoutes();
     }
 
+    /**
+     * Register API controllers from plugin
+     */
+    public function registerPluginApiControllers(array $apiControllers): void
```
ewhanson, line 155: Is the idea that this method would be called by the plugin that gets this class from the hook? If so, I think this information should be a bit more explicit in the function description, e.g. "when receiving the instance of this class from the hook, the plugin should call this method and add in any custom api controllers." Something like that. We will also need to be sure to add this to the pkp docs as well, but I can handle that after this is merged.

classes/core/APIRouter.php
```diff
@@ -169,6 +211,39 @@ public function url(
...
+     * @param string $requestPath Full request path (e.g., '/index.php/testContext/api/v1/custom-plugin-path/data')
+     * @param string $handlerPath Plugin handler path (e.g., 'custom-plugin-path')
+     *
+     * @return bool True if the request should be handled by this plugin controller
+     */
+    protected function matchesPluginHandlerPath(string $requestPath, string $handlerPath): bool
+    {
+        // Extract the path after /api/v{version}/
+        // Example: '/index.php/testContext/api/v1/custom-plugin-path/data' -> 'custom-plugin-path/data'
+        $pattern = '~/api/v\d+/([^?#]+)~';
```
ewhanson, line 239: It looks like there are a few things off about the regex. Could you have another look? 

I think the `/` characters need to be escaped like so `\/` and I'm not sure what the `~` characters are catching. Even with the `~` and updating the `/` characters, I still wasn't able to get it to match the example you gave. Unless there's something I'm missing.

ewhanson (changes requested):
Thanks @touhidurabir, it's looking good. Just a few quesitons, mostly about documentation.

classes/core/APIRouter.php
touhidurabir: Here `~` used as delimiter to escape so both the following is same as 
1. `~/api/v\d+/([^?#]+)~`
2. `/\/api\/v\d+\/([^?#]+)/`

> I still wasn't able to get it to match the example you gave.

not sure what I am missing as I test with following code snippets as 
```php
$pattern = "~/api/v\d+/([^?#]+)~";
$requestPath = "/index.php/testContext/api/v1/custom-plugin-path/data";

preg_match($pattern, $requestPath, $matches);

dump($matches);
```
which provide result as 
```
[
    "/api/v1/custom-plugin-path/data",
    "custom-plugin-path/data",
]
```
basically what we want to match here . 

The purpose there is to match the request path with controller handler path to determine if that controller should be used . 

or perhaps I am misunderstanding something here ?

classes/core/APIRouter.php
touhidurabir: > Is the idea that this method would be called by the plugin that gets this class from the hook? 

right . 

updated the description .

classes/core/APIRouter.php
touhidurabir: So currently we have API hooks as `APIHandler::endpoints::<ENTITY>` where we get the entity from the `APIRouter::getEntity` . for example , for USER entity we get like `APIHandler::endpoints::users`, we can inject routes to `users/` endpoint like `users/new-route` . so if we get a plugins entity in future (currently we don't have anything such), then it will conflict with that one . 

Not sure if this is something we should think about  for future possibility or I am just over thinking

ewhanson: Ah, I see. I think it's okay as we are unlikely to have a plugin entity that will behave in a similar way to the other entities like users, so I think the name is fine.

classes/core/APIRouter.php
ewhanson: Thanks @touhidurabir. I'm not sure what I had been doing when testing it but it's working as you say now when I test it again. So that's okay. Sorry for the confusion!

ewhanson (approved):
Looks good, thanks @touhidurabir!

ewhanson:
@asmecher, could you have a second look at this before it gets merged into `stable_3_5_0`? Then @touhidurabir, would you forward-port it to `main`? Thanks!

asmecher:
@ewhanson Approved, thanks!

classes/core/APIRouter.php
```diff
@@ -91,11 +97,31 @@ public function getEntity(): string
...
+            $requestPath = $request->getRequestPath();
+            foreach ($this->registeredPluginApiControllers as $handlerPath => $apiController) {
+                if ($this->matchesPluginHandlerPath($requestPath, $handlerPath)) {
+                    $handler = new APIHandler($apiController);
+                    $this->setHandler($handler);
+                    $handler->runRoutes();
+                    return;
+                }
+            }
+        }
```
touhidurabir, line 123: Hi @ewhanson , How about moving the block just below inside of `if (!file_exists($sourceFile)) {` check ? 

pros : 
1. most of the core API will get the initial priority to complete first 
2. If there are many plugins (say over 100) with custom routes, this will reduce the check for those 100 plugins for each API request which I think a performance issue for every API request . 
3. Can not accidentally over write core entity based API end points e.g. submissions, users, etc

Cons : 
1. Plugin can not replace a existing entity API path. For example, `submissions` as core API path are given higher priority to run first . They can do extends using other hook .
2. Plugin to fully replace an existing entity API route, need to use `Dispatcher::dispatch` hook with a different implementation . 

would like to know what you think .

asmecher: (Touhidur and I talked about this today; I have no objection. Plugins needing to override a full API route do have an alternative in other hooks -- and my preference for plugins that need to alter built-in behaviour is that they use something much less drastic than this, closer to the map or entity level.)

touhidurabir: Now that I am thinking about this a bit more, I think the initial approach was fine which also allow a full override if needed . Plugin developer can use alternative workaround like `Dispatcher::dispatch` so better to have this via what we initially had and properly document it. Adding/Overriding API means plugin developer already intervening into the core at very low level and they should understand the risk of mistake or untested behaviour implementation .

PR #12053, opened 2025-11-20
Submission Emails: Use author email link when emailing authors

kennydude:
We have come across an issue when using discussions on a submission:

The email link sends the author to a dead end as it assumes they are an editor. This PR checks if they are the author and swaps for the "my submissions" url instead allowing them to view the conversation and correct context,


asmecher:
Thanks, @kennydude! @Vitaliy-1, could you have a look at this one (or delegate it)?

Issue #11996, opened 2025-11-04
[Structured Citations] -- possibility to define that citations should not be structured on import


bozana:
Implement a possibility to say that citations should not be structured on import even if the metadata lookup is enabled for journal

bozana:
Hi  @asmecher, could you maybe take a look at the PR above? 
Once we improved the jobs processing with the PRs Touhidur works on, this is maybe not needed, but also maybe not wrong to have it. (For example, if we would extend the XML import/export schema to support structured citations, it would be needed.)
Thanks a lot!

asmecher:
@bozana, I think the choice of whether or not to look up the metadata should be an option to the import/export tools -- probably with a "no" default, so that folks who don't know about it don't DDOS themselves. I would rather we used an option to the import/export toolset, like we do with [the `--no-embed` option](https://github.com/pkp/pkp-lib/issues/5372#issuecomment-588019936).

bozana:
hi @asmecher, do you think that option would then be enough, only for the CLI usage?

asmecher:
Yes, I think so, thanks! It's much less likely that folks will be doing large-scale import operations on the web-based interface; they might go an issue at a time, which is manageable with jobs being kicked off.

bozana:
Hi @asmecher, I considered the option citation-metadata-lookup on CLI import. The PR includes a few other fixes, e.g. assuming that we can get the context from the request, which does not work on such an import.
Could you please take a look?
Thanks a lot!

asmecher:
Thanks, I've added a review! https://github.com/pkp/pkp-lib/pull/12072#pullrequestreview-3590405911

bozana:
Hi @Tribunal33, this issue considers citations processing in native import via CLI:
If citation metadata lookup is enabled and a native import is done via CLI, then the citations in the import XML will not be processed per default. If the new option `citation-metadata-lookup` is used then they will be processed.
This is an example now to start a native import via CLI:
php tools/importExport.php NativeImportExportPlugin usage
php tools/importExport.php NativeImportExportPlugin import path/to/test.xml publicknowledge --admin
php tools/importExport.php NativeImportExportPlugin import --citation-metadata-lookup path/to/test.xml publicknowledge --admin

If you would like to test and close the issue...
Thanks a lot!

Tribunal33:
[...]

PR #12072, opened 2025-11-27
pkp/pkp-lib#11996 add attribute whether metadata lookup for citations…

bozana:
… should be done on import

s. https://github.com/pkp/pkp-lib/issues/11996


bozana:
Thanks a lot @asmecher! I added a note into that issue for release notebook.

PR #12075, opened 2025-11-27
pkp/pkp-lib#10263 fix to allow editors to add/edit contributors on published publications

Hafsa-Naeem:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -2472,8 +2466,23 @@ protected function removeIrrelevantContributorTypeData(array &$params, Context $
     /**
      * Get contributor role objects from ids and context id
      */
-    protected function contributorRolesToParams(array $roleIds, int $contextId): array
+    protected function contributorRolesToParams(array $roles, int $contextId): array
```
asmecher: This change makes the `$roles` parameter ambiguous where I don't think that's justified. It would be better to keep it `$roleIds`, but make sure that calling code is using it consistently.

One day [PHP will get generics](https://thephp.foundation/blog/2024/08/19/state-of-generics-and-collections/), at which time we'll be able to define what types an `array` can contain. When that happens, ambiguous arrays will stand out even further.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1672,7 +1675,12 @@ public function addContributor(Request $illuminateRequest): JsonResponse
         }
 
         if (isset($params['contributorRoles']) && is_array($params['contributorRoles'])) {
-            $params['contributorRoles'] = $this->contributorRolesToParams($params['contributorRoles'], $submissionContext->getId());
+            $roleIds = array_values(array_unique(array_filter(
+                array_map('intval', $params['contributorRoles']),
+                static fn (int $id) => $id > 0
+            )));
+
+            $params['contributorRoles'] = $this->contributorRolesToParams($roleIds, $submissionContext->getId());
```
asmecher: @Hafsa-Naeem, does this change (and the near-duplicate block at line 1827) add any protection? It looks like noise to me; the `roleIds` parameter to `contributorRolesToParams` is passed into `ContributorRole::scopeWithRoleIds`, which uses a `whereIn` on the numeric `contributor_role_id` column. There will never be negative or non-numeric values there, and the `IN` SQL clause doesn't mind if there are duplicates or if non-numeric string values are passed in.

In fact, I think applying `intval` will introduce a risk -- when passed non-numeric content it will return a `1`, which might match something unexpected.

I'm fine with defensive programming, but that doesn't mean adding checks to new code without a clear purpose. Our enemy is volume of code and ambiguity of purpose, with a large codebase and a small team.

api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1642,10 +1642,13 @@ public function addContributor(Request $illuminateRequest): JsonResponse
...
 
-        // Publications can not be edited when they are published
-        if ($publication->getData('status') === PKPPublication::STATUS_PUBLISHED) {
+        $userRoles = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
+        if (
+            !in_array(Role::ROLE_ID_SITE_ADMIN, $userRoles)
+            && !Repo::submission()->canEditPublication($submission->getId(), $currentUser->getId())
+        ) {
             return response()->json([
-                'error' => __('api.publication.403.cantEditPublished'),
```
asmecher, line 1648: @Hafsa-Naeem, I think this locale key is no longer used after this change, and should be removed from the locale files.

Issue #12045, opened 2025-11-19
Add peer review public visibility options


ewhanson:
The existing model of peer review methods (anonymous, double anonymous, open) all currently refer to the visibility of author/review identity and not whether the review contents are made publicly visible to readers. We will need to support public visibility of not only fully open peer reviews but also varying degrees of anonymity. 

To support this, we will need:
- an option to specify whether reviews will be made publicly visible or not
- the ability to set this at a journal level
- the ability to change/opt-out on an individual review level

Journal-level implementation:
- Under `Settings -> Workflow -> Review -> Setup` below `Default Review Mode`:
> Heading - Publicly Show Reviewer Comments
> Explanation text below it: Enable this setting if you’d like the review process to be made publicly visible alongside published submissions. This supports transparent peer review practices and can help foster greater trust and accountability.
> 
> Checkbox Label: ☑ Make reviewer comments publicly visible with published content

Other considerations:
- Review identity should be either exposed or anonymized based on the review mode type (anonymized for both anonymous types, visible for the open type)
- Reviews should be publicly visible based on this new visibility option.

Unanswered questions:
- [x] Implementation approach -- (see implementation above)

### Pull Requests
~~pkp-lib: https://github.com/pkp/pkp-lib/pull/12076~~
~~ojs: https://github.com/pkp/ojs/pull/5202~~
~~ops: https://github.com/pkp/ops/pull/1144~~
~~omp: https://github.com/pkp/omp/pull/2176~~

- `pkp-lib`: pkp/pkp-lib#12083
- `ojs`: pkp/ojs#5208
- `ops`: pkp/ops#1148
- `omp`: pkp/omp#2180

taslangraham:
@ewhanson Could you give the PRs listed at https://github.com/pkp/pkp-lib/issues/12045#issue-3643900796 a review, please.

### Changes
- Added settings to control peer review public visibility
- Manage public peer review API access based on current visibility setting
- Hide/Show(anonymized for both anonymous types, visible for the open type) reviewer identity on public peer reviews based on review mode

ewhanson:
Thanks @taslangraham. Looks pretty good. Just a few questions and one bug to fix. Happy to have another look once those are in if you'd like, otherwise, if the OJS tests are passing, feel free to merge it.

ewhanson:
Thanks @taslangraham, I've merged your commits via separate PRs because of merge conflicts.

Tribunal33:
[...]

PR #12076, opened 2025-11-27
pkp/pkp-lib#12045 Add peer review public visibility options

taslangraham:
For pkp/pkp-lib#12045

### DIFF
--- a/api/v1/publicationPeerReviews/resources/PublicationPeerReviewResource.php
+++ b/api/v1/publicationPeerReviews/resources/PublicationPeerReviewResource.php
@@ -111,15 +111,18 @@ class PublicationPeerReviewResource extends JsonResource
             } else {
                 $reviewerComments = $this->getReviewAssignmentComments($assignment);
             }
+
+            $isReviewOpen = $assignment->getReviewMethod() === ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN;
             return [
                 'id' => $assignment->getData('id'),
-                'reviewerId' => $assignment->getReviewerId(),
-                'reviewerFullName' => $assignment->getReviewerFullName(),
-                'reviewerAffiliation' => Repo::user()->get($assignment->getReviewerId())->getLocalizedAffiliation(),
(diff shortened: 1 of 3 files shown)


api/v1/publicationPeerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -111,11 +111,13 @@ private function getReviewAssignmentPeerReviews(Enumerable $assignments, Context
             } else {
                 $reviewerComments = $this->getReviewAssignmentComments($assignment);
             }
+
+            $isReviewOpen = $assignment->getReviewMethod() === ReviewAssignment::SUBMISSION_REVIEW_METHOD_OPEN;
```
ewhanson, line 115: I wonder if we'd like to include this in the data that's returned. I know it will be obvious in context, but it may be helpful to have this codified in the data returned. Something like `'isReviewOpen' => true`.

classes/context/Context.php
```diff
@@ -70,6 +70,15 @@ public function areDoisEnabled(): bool
         return (bool) $this->getData(Context::SETTING_ENABLE_DOIS);
     }
 
+    /**
+     * Whether peer reviews are publicly visible for this context
+     *
+     */
+    public function arePeersReviewPublic(): bool
+    {
+        return (bool) $this->getData('enablePublicPeerReviews');
```
ewhanson: Could you create a constant above for this string. Something like `SETTING_ENABLE_PUBLIC_REVIEWS`?

ewhanson (changes requested):

classes/context/Context.php
taslangraham: done

api/v1/publicationPeerReviews/resources/PublicationPeerReviewResource.php
taslangraham: Good idea -- I've added it to the data returned.

ewhanson:
Merged via separate PR. See pkp/pkp-lib#12045 for commits.

Issue #11937, opened 2025-10-16
Implement getWriteDisabledErrors() for every API controller


bozana:
We use `writeDisabledInApi` attribute in our schemas, and we also need a validation check for those properties in API endpoints. It is currently only considered in PKPSubmissionController and a few other controllers, where function getWriteDisabledErrors() is used. This however should be considered in every API controller where objects that contain those properties are edited i.e. as a function in the base API controller class.

taslangraham:
@bozana I’ve been thinking about a way to make this functionality available to all controllers without requiring each one to manually call `getWriteDisabledErrors()`. Below is the approach I’m proposing. Let me know what you think.


##### 1. Add Schema Property to `PKPBaseController`

Introduce:

```php
protected ?string $apiWriteSchema = null;
```

Controllers that handle write operations will set this property to the relevant schema. If `null`, no check is performed.

---

##### 2. Move `getWriteDisabledErrors()` to `PKPBaseController`

Implement the `getWriteDisabledErrors` method in `PKPBaseController` as outlined in the issue description.

---

##### 3. Introduce Middleware: `CheckAPIWriteSchema`

Create middleware that:

* Runs on write requests (POST/PUT/PATCH)
* Checks if the resolved controller has `$apiWriteSchema` set
* Calls `getWriteDisabledErrors()`
* Returns necessary error if write-disabled properties are present
---

##### 4. Register the middleware globally
Add the middleware to `PKPRoutingProvider::$globalMiddleware`

---

With this in place, once a controller defines `$apiWriteSchema`, the middleware automatically enforces validation checks for `writeDisabledInApi` properties in the schema. This removes the need to have code like the following in every handler that requires the check.
```php
   $readOnlyErrors = $this->getWriteDisabledErrors(PKPSchemaService::SCHEMA_CITATION, $params);
        if (!empty($readOnlyErrors)) {
            return response()->json($readOnlyErrors, Response::HTTP_BAD_REQUEST);
        }
```

bozana:
Hi @taslangraham, maybe @ewhanson (as API lead) would be the best person to ask, so maybe to include him as well.
What if a controller would handle different objects i.e. different schemas? Maybe this should never be the case, from our API design strategy?
Else, yes, that all sounds good to me.
Thanks!

PR #12079, opened 2025-11-28
Implement Data Citations support

ajnyga:
https://github.com/pkp/pkp-lib/issues/6278


api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,224 @@
+<?php
+
+/**
+ * @file api/v1/dataCitations/PKPDataCitationController.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
```
bozana: 2026 everywhere :-)

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,224 @@
...
+
+        return parent::authorize($request, $args, $roleAssignments);
+    }
+
+    /**
+     * Get a single data dataCitation.
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $announcement = DataCitation::find((int) $illuminateRequest->route('dataCitationId'));
```
bozana: $dataCitation instead of $announcement

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,224 @@
...
+
+    }
+
+    /**
+     * Add a data citation.
+     */
+    public function add(Request $illuminateRequest): JsonResponse
+    {
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_DATA_CITATION, $illuminateRequest->input());
+        $params['publicationId'] = (int) $illuminateRequest->route('publicationId');
```
bozana: Why do we need this, shouldn't this happen in convertStringsToSchema above?

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
+{
+	"title": "DataCitation",
+	"description": "A Data Citation in a Publication.",
+	"properties": {
+		"id": {
+            "type": "integer",
+            "origin": "primary",
+            "apiSummary": true
```
bozana: aslo readOnly

classes/dataCitation/DataCitation.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        return $this->settingsTable;
+    }
+
+    /**
+     * @return bool
+     *
+     * @hook DataCitation::add [[$this]]
+     */
+    public function save(array $options = [])
```
bozana: return type missing

classes/dataCitation/DataCitation.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function getSettingsTable(): string
+    {
+        return $this->settingsTable;
+    }
+
+    /**
+     * @return bool
```
bozana: function description, inheritDoc?

classes/dataCitation/Repository.php
```diff
@@ -0,0 +1,91 @@
...
+use PKP\plugins\Hook;
+use PKP\services\PKPSchemaService;
+use PKP\validation\ValidatorFactory;
+
+class Repository
+{
+    /** @var string $schemaMap The name of the class to map this entity to its schema */
+    public string $schemaMap = maps\Schema::class;
+
+    /**  */
```
bozana: can be removed

classes/dataCitation/Repository.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        Hook::call('DataCitation::validate', [&$errors, $dataCitation, $props]);
+
+        return $errors;
+    }
+
+    /**
+     * Get an instance of the map class for mapping
+     * announcements to their schema
```
bozana: data citations instead of announcements

classes/components/forms/dataCitation/DataCitationEditForm.php
```diff
@@ -0,0 +1,99 @@
...
+    public const FORM_DATA_CITATION_EDIT = 'data_citation';
+    public $id = self::FORM_DATA_CITATION_EDIT;
+    public $method = 'POST';
+
+    /**
+     * Constructor
+     *
+     * @param string $action URL to submit the form to
+     */
+    public function __construct(string $action) {
```
bozana: { should go in the new line

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+            "description": "The title of the dataset.",
+            "apiSummary": true
+        },
+        "identifierType": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The identifier type for the dataset.",
+            "apiSummary": true,
+            "validation": [
```
bozana, line 49: Could you add the allowed types here, so that validation can happen here too?

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+				"nullable"
+			]
+        },
+        "relationshipType": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The type of relationship between the data and the publication.",
+            "apiSummary": true
```
bozana: Maybe to also add the allowed relationship types or validation

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
+{
+	"title": "DataCitation",
+	"description": "A Data Citation in a Publication.",
+	"properties": {
+		"id": {
```
bozana: could you please the properties alphabetically -- that way one can find/read them much easier...

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+			]
+        },
+        "year": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The year the dataset was published.",
+            "apiSummary": true,
+            "validation": [
+				"nullable"
```
bozana: Do we have a expected form this date to have, for validation? Maybe "digit:4"?

classes/components/forms/dataCitation/DataCitationEditForm.php
```diff
@@ -0,0 +1,99 @@
...
+        ]));
+
+        $this->addField(new FieldText('url', [
+            'label' => __('submission.dataCitations.label.url'),
+            'description' => '',
+            'value' => null
+        ]));
+
+    }
+}
```
bozana: add empty line after }

classes/dataCitation/DataCitation.php
bozana: do we need this function only because of hooks? :thinking:

classes/publication/Repository.php
```diff
@@ -398,6 +399,20 @@ public function version(Publication $publication, ?VersionStage $versionStage =
             }
         }
 
+        Repo::citation()->importCitations(
+            $newPublication->getId(),
+            $newPublication->getData('citationsRaw')
+        );
```
bozana: This can be removed -- it was removed, because no need to re-import and re-process citations on versioning. They are just copied above. This is probably result of a rebase?

classes/publication/Repository.php
```diff
@@ -398,6 +399,20 @@ public function version(Publication $publication, ?VersionStage $versionStage =
...
+
+        // Clone data citations if any
+        $dataCitations = DataCitation::where('publication_id', $publication->getId())->get();
```
bozana: we have dataCitations in the publication itself, so maybe no need to query the DB here?

classes/security/authorization/internal/DataCitationRequiredPolicy.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
```
bozana: Is this policy needed and used?

classes/security/authorization/internal/DataCitationRequiredPolicy.php
```diff
@@ -0,0 +1,82 @@
...
+            return AuthorizationPolicy::AUTHORIZATION_DENY;
+        }
+
+        // Save the dataCitation to the authorization context.
+        $this->addAuthorizedContextObject(Application::ASSOC_TYPE_DATA_CITATION, $dataCitation);
+        return AuthorizationPolicy::AUTHORIZATION_PERMIT;
+    }
+}
+
+if (!PKP_STRICT_MODE) {
```
bozana: this is not needed

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        Hook::run('API::dataCitations::params', [$dataCitations, $illuminateRequest]);
+
+        return response()->json([
+            'itemsMax' => $dataCitations->count(),
+            'items' => Repo::dataCitation()->getSchemaMap()->summarizeMany($dataCitations->get())->values(),
+        ], Response::HTTP_OK);
+
+
```
bozana: maybe to remove the empty lines?

classes/dataCitation/DataCitation.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Filter by publication ID
+     */
+    protected function scopeWithPublicationId(EloquentBuilder $builder, int $publicationId): EloquentBuilder
+    {
+        return $builder->where('publication_id', $publicationId);
+    }
+
+
```
bozana: empty lines can be removed

classes/publication/DAO.php
```diff
@@ -471,6 +474,23 @@ protected function deleteCategories(int $publicationId): void
         PublicationCategory::where('publication_id', $publicationId)->delete();
     }
 
+    /**
+     * Set a publication's Data Citations
+     */
+    protected function setDataCitations(Publication $publication)
```
bozana: return type = void

classes/publication/DAO.php
```diff
@@ -471,6 +474,23 @@ protected function deleteCategories(int $publicationId): void
...
+    {
+        $dataCitations = DataCitation::where('publication_id', $publication->getId())->get()->values()->all();
+        $publication->setData('dataCitations', $dataCitations);
+    }
+
+    /**
+     * Delete a publication's Data Citations
+     */
+    protected function deleteDataCitations(int $publicationId)
```
bozana: return type = void

pages/submission/PKPSubmissionHandler.php
```diff
@@ -752,14 +763,51 @@ protected function getDetailsStep(
...
+
         return [
             'id' => 'details',
             'name' => __('common.details'),
             'reviewName' => __('common.details'),
             'sections' => $sections,
             'reviewTemplate' => '/submission/review-details.tpl',
         ];
+
+
```
bozana: empty lines, here and below

plugins/importexport/native/pkp-native.xsd
```diff
@@ -423,4 +424,12 @@
 			<attribute name="locale" type="string" />
 		</complexType>
 	</element>
+	<element name="dataCitations">
+		<complexType>
+			<sequence>
+				<element name="dataCitation" type="string" minOccurs="0" maxOccurs="unbounded" />
```
bozana: shouldn't we define the elements: title, identifier, identifierType, etc. ?

classes/submission/Repository.php
```diff
@@ -432,6 +432,7 @@ public function validateSubmit(Submission $submission, Context $context): array
             if ($metadata === 'citations') {
                 $metadata = 'citationsRaw';
             }
+
```
bozana: empty line can be removed

classes/services/PKPSchemaService.php
```diff
@@ -398,6 +399,7 @@ public function coerce($value, $type, $schema)
                     }
                     $newObject[$propName] = $this->coerce($value[$propName], $propSchema->type, $propSchema);
                 }
+
```
bozana: empty line can be removed

classes/migration/install/MetadataMigration.php
```diff
@@ -55,6 +55,34 @@ public function up(): void
...
+            $table->foreign('data_citation_id', 'data_citation_settings_data_citation_id')->references('data_citation_id')->on('data_citations')->onDelete('cascade');
+            $table->index(['data_citation_id'], 'data_citation_settings_data_citation_id');
+
+            $table->string('locale', 28)->default('');
+            $table->string('setting_name', 255);
+            $table->mediumText('setting_value')->nullable();
+
+            $table->unique(['data_citation_id', 'locale', 'setting_name'], 'data_citation_settings_unique');
+        });
+
```
bozana, line 85: Also an upgrade migration script, that adds these tables, is needed

bozana:
- Data Availability Statement now appears both, on the Metadata page as well as on the Data page. The one on Metadata page should be removed, right?
- Sequencing is not implemented -- seq column seems to be always 0

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+			}
+		},
+        "url": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The URL of the dataset.",
+            "apiSummary": true,
+            "validation": [
+				"nullable"
```
bozana: maybe also "url" validation here

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+				"type": "object",
+				"properties": {
+					"givenName": {
+						"type": "string"
+					},
+					"familyName": {
+						"type": "string"
+					},
+					"orcid": {
+						"type": "string"
```
bozana: we can also add "orcid" validation

schemas/dataCitation.json
```diff
@@ -0,0 +1,113 @@
...
+			]
+        },
+        "identifier": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The identifier for the dataset.",
+            "apiSummary": true,
+            "validation": [
+				"nullable"
```
bozana: it would be great to have a validation here as well, but I do not know if we can do it for all types we support :sweat_smile:

pages/authorDashboard/PKPAuthorDashboardHandler.php
```diff
@@ -240,9 +241,11 @@ public function setupTemplate($request)
         );
 
         $titleAbstractForm = $this->getTitleAbstractForm($latestPublicationApiUrl, $locales, $latestPublication, $submissionContext);
+        $dataAvailabilityAndCitationsForm  = new PKPDataAvailabilityAndCitationsForm($latestPublicationApiUrl, $locales, $latestPublication);
```
bozana: it seems the 4. parameter, dataAvailabilitySetting, is missing here

bozana: Now I see: I think this function is not used any more. Instead the one in PKPDashboardHandler is used. Thus, you could remove changes from here. 
Could you please also remove $citationForm::FORM_CITATIONS (2 times)?

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: check, note that the whole pkp-lib still has 2025 and some other years as well, but updated everything in this commit to 2026

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: I mean without that you would get:
`Next Illuminate\Database\QueryException: SQLSTATE[HY000]: General error: 1364 Field 'publication_id' doesn't have a default value (Connection: mysql, SQL: insert into `data_citations` () values ()) in`

If there is another way of setting the publicationId let me know

ajnyga: Announcements does something similar:
https://github.com/pkp/pkp-lib/blob/main/api/v1/announcements/PKPAnnouncementController.php#L196-L198

classes/dataCitation/DataCitation.php
ajnyga: I probably started with the Announcement code in this one as well. It has this but it does also functionality with image uploads there https://github.com/pkp/pkp-lib/blob/main/classes/announcement/Announcement.php#L111

So maybe we could just remove this from here and add if there is a request for hooks here, maybe there is no request.

classes/publication/Repository.php
ajnyga: yeah probably just a leftover from an older version of the code

classes/publication/Repository.php
ajnyga: check

classes/security/authorization/internal/DataCitationRequiredPolicy.php
ajnyga: I maybe wrote this early on already last summer and since it is not used have not encountered any problems with it either. I am thinking that do we need this related to the API? Maybe not

pages/authorDashboard/PKPAuthorDashboardHandler.php
ajnyga: yes I think this was a leftover from my endeavour to add the vue manager component to the wizard

plugins/importexport/native/pkp-native.xsd
ajnyga: Yes so the native XML support is something I have not added yet. 

What kind of support do you think we need here? Full implementation or something else?

schemas/dataCitation.json
ajnyga: yes, not even sure what the rules would be for some of the identifiers

schemas/dataCitation.json
ajnyga: Do we have a validator called "orcid", Laravel does not have this?

classes/migration/install/MetadataMigration.php
ajnyga: added

classes/dataCitation/DataCitation.php
bozana: Yes, lets remove it...

api/v1/dataCitations/PKPDataCitationController.php
bozana: Oh, this is my bad -- I thought it was coming as parameter. Sorry!
I think we should define `publicationId` property with `"writeDisabledInApi": true`, and add the function `getWriteDisabledErrors()` in this class, similarly how it is in ContributorRoleController.
Here, in this function add(), you assign the publication ID directly, so it will override any existing parameter publicationId if given/provided. But, in the function edit, you do not do it, so there it seems the publication ID could be set to another value with the list of parameters. Or you should override it in the function edit() as well.
However we should define the property with `"writeDisabledInApi": true` because once we will handle it central (s. https://github.com/pkp/pkp-lib/issues/11937)

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,222 @@
...
+    {
+        $dataCitation = DataCitation::find((int)$illuminateRequest->route('dataCitationId'));
+
+        if (!$dataCitation) {
+            return response()->json([
+                'error' => __('api.dataCitations.404.dataCitationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_DATA_CITATION, $illuminateRequest->input());
```
bozana, line 205: here the publicationId could be overridden with the given/received parameter?

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,222 @@
...
+                    ->whereNumber('dataCitationId');
+            });
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::authorize()
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
```
bozana, line 106: I think we also need 
`$this->addPolicy(new PublicationAccessPolicy($request, $args, $roleAssignments));`.

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,222 @@
...
+
+        $rolePolicy = new PolicySet(PolicySet::COMBINING_PERMIT_OVERRIDES);
+
+        $this->addPolicy(new ContextRequiredPolicy($request));
```
bozana: Could you move this policy above, on the first place?

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,222 @@
...
+
+        return parent::authorize($request, $args, $roleAssignments);
+    }
+
+    /**
+     * Get a single data dataCitation.
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $dataCitation = DataCitation::find((int) $illuminateRequest->route('dataCitationId'));
```
bozana, line 124: either we can provide $publicationId to find(), or we check if the dataCitation->getData('publicationId') is the same as the given publication ID and return the error, similarly to how some check if the context is the same.

api/v1/dataCitations/PKPDataCitationController.php
bozana: Hmmm... but this policy would need submission to exist and we do not have it in our API path... 🤔

bozana: I see the same problem exist for citations :-(

bozana: I need to see with Alec and Erik what is the right way to solve it. Either to implement a new policy -- which I would rather not do, because we need to copy and consider some rules from already existing policies.
Or to adapt the API URL. But also here I would need help from Erik, what he thinks would be best.

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,222 @@
...
+     *
+     * @hook API::dataCitations::params [[$collector, $illuminateRequest]]
+     */
+    public function getMany(Request $illuminateRequest): JsonResponse
+    {
+        $dataCitations = DataCitation::limit(self::DEFAULT_COUNT)->offset(0);
+
+        if ($illuminateRequest->route('publicationId')) {
+            $dataCitations->withPublicationId($illuminateRequest->route('publicationId'));
+        }
```
bozana: I think the publication object should always exist.
Later, when we consider that PublicationAccessPolicy we can get it with:
`$publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION);`
We can thus always i.e. immediately filter by publication ID.
I do not think that the user should ever see all existing dataCitations, right?

classes/security/authorization/internal/DataCitationRequiredPolicy.php
bozana: I think we rather need PublicationAccessPolicy, which again requires SubmissionAccessPolicy, which again requires submission ID in the request...
And then in each api controller function we should check that the data citation belongs to the given publication.
So I think we will not need this class, but let me ask Alec and Erik what they think we should solve the problem... If we would need additional policy, then another rules should be considered here...

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: yes, the edit form does not pass this but of course someone could pass that with the API.
So is using the getWriteDisabledErrors method and "writeDisabledInApi": true here enough?

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: you mean something like?

```
        if ($illuminateRequest->route('publicationId') !== $dataCitation->publicationId) {
            return response()->json([
                'error' => __('api.dataCitations.400.publicationsNotMatched'),
            ], Response::HTTP_FORBIDDEN);
        }
```

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: yes at least with the current UI there is no overall view of data citations and maybe there will never be a requirement for that, ie. getting data citations without a publication context, I will remove that check and just use withPublicationId directly

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: ok let me know

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: not 100% sure I understood what you would like to see here, but let me know if my latest pr is what you were thinking

ajnyga:
the new version now removes the Data Availability Statement from the metadata page and introduces Sequencing.

Note that I could not find any eloquent + api examples for the sorting. I did find a sorting composable used in Galleys grid and used the same but in galleys it interacts with the legacy gridhandler. My version does use the API now but not sure if this is the correct approach.

api/v1/dataCitations/PKPDataCitationController.php
bozana: I see now you have the property in the schema. 
Now you would also need something like this: https://github.com/pkp/pkp-lib/blob/main/api/v1/contributorRoles/ContributorRoleController.php#L220-L223.

api/v1/dataCitations/PKPDataCitationController.php
bozana: yes.
Just that once we consider PublicationAccessPolicy, we can get the publication this way:
`$publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION);`
And then get the ID $publication->getId().

p.s. I am still waiting for the answer from Erik, regarding the API URLs...

api/v1/dataCitations/PKPDataCitationController.php
bozana: as for add() above also this part is needed: Now you would also need something like this: https://github.com/pkp/pkp-lib/blob/main/api/v1/contributorRoles/ContributorRoleController.php#L220-L223.

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: You mean this https://github.com/pkp/pkp-lib/pull/12079/changes#diff-bc8c4665159672721adc06aa04db684397ec6610841c2ea1d32da6c3fc4527daR162-R164

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: It should be there in line 195 onwards

api/v1/dataCitations/PKPDataCitationController.php
bozana: Hi @ajnyga,
we realized that the API endpoint would need to be: api/v1/submissions/<submissionId>/publications/<publicationId>/dataCitations. As example see [https://github.com/pkp/pkp-lib/blob/main/api/v1/jats/PKPJatsController.php](https://github.com/pkp/pkp-lib/blob/main/api/v1/jats/PKPJatsController.php). I think the same policies could apply here as well 🤔

ajnyga: ok, I will rewrite the handler and the vue stuff for that

ajnyga: New pr now includes the rewritten API paths and authorization

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: this is probably resolved now with the new pr as well? I now use $publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION); instead of the illuminateRequest one

api/v1/dataCitations/PKPDataCitationController.php
bozana: Yes, sorry, I have not seen it before :-)

plugins/importexport/native/pkp-native.xsd
bozana: I have spoken with Alec, and we do not need to consider data citations in the native import/export at all.

api/v1/dataCitations/PKPDataCitationController.php
bozana: Shell we do the check here, in this function, as well?
So that we cannot get a data citation from another publication...

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,292 @@
...
+     */
+    public function edit(Request $illuminateRequest): JsonResponse
+    {
+        $dataCitation = DataCitation::find((int)$illuminateRequest->route('dataCitationId'));
+
+        if (!$dataCitation) {
+            return response()->json([
+                'error' => __('api.dataCitations.404.dataCitationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
```
bozana, line 196: Here as well, there should be a check if the data citation belong to the given publication, I think

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,292 @@
...
+    public function delete(Request $illuminateRequest): JsonResponse
+    {
+        $dataCitation = DataCitation::find((int) $illuminateRequest->route('dataCitationId'));
+
+        if (!$dataCitation) {
+            return response()->json([
+                'error' => __('api.dataCitations.404.dataCitationNotFound')
+            ], Response::HTTP_OK);
+        }
+
```
bozana, line 247: also here, check if the data citation belongs to the given publication ?

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,292 @@
...
+        return response()->json(
+            Repo::dataCitation()->getSchemaMap()->map($dataCitation), Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Save the order of data citations for a publication.
+     */
+    public function saveOrder(Request $illuminateRequest): JsonResponse
```
bozana, line 258: Lets see what Jarda says, if this is needed, or if we should ignore the seq for now, until an UI component is there...

classes/core/PKPApplication.php
```diff
@@ -91,6 +91,7 @@ abstract class PKPApplication implements PKPApplicationInfoProvider
     public const ASSOC_TYPE_DECISION_TYPE = 0x010000f;
     public const ASSOC_TYPE_COMMENT = 0x0100010;
     public const ASSOC_TYPE_COMMENT_REPORT = 0x0100011;
+    public const ASSOC_TYPE_DATA_CITATION = 0x0100010;
```
bozana: this is the same as ASSOC_TYOE_COMMENT above, maybe next would be 0x0100012 ?

classes/security/authorization/internal/DataCitationRequiredPolicy.php
bozana: I think we can remove this now. Submission and Publication objects are checked in the PublicationAccessPolicy, so we only need to ensure that the data citation belongs to the given publication, which we can do in the API controller.

plugins/importexport/native/filter/NativeXmlPKPPublicationFilter.php
```diff
@@ -204,6 +204,7 @@ public function handleChildElement($n, $publication)
                 case 'authors':
                     $this->parseAuthors($n, $publication);
                     break;
+// DATACITATIONS TODO
```
bozana: can be removed here and below...

api/v1/dataCitations/PKPDataCitationController.php
bozana: Oh, first now I see that you have implemented the ordering... sorry...

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,292 @@
...
+     * Get a single data dataCitation.
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $dataCitation = DataCitation::find((int) $illuminateRequest->route('dataCitationId'));
+
+        if (!$dataCitation) {
+            return response()->json([
+                'error' => __('api.dataCitations.404.dataCitationNotFound')
+            ], Response::HTTP_OK);
```
bozana: This should not be Response::HTTP_OK, but Response::HTTP_NOT_FOUND

api/v1/dataCitations/PKPDataCitationController.php
bozana: Also here Response::HTTP_NOT_FOUND

api/v1/dataCitations/PKPDataCitationController.php
```diff
@@ -0,0 +1,292 @@
...
+        $dataCitation = DataCitation::find((int)$illuminateRequest->route('dataCitationId'));
+
+        if (!$dataCitation) {
+            return response()->json([
+                'error' => __('api.dataCitations.404.dataCitationNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_DATA_CITATION, $illuminateRequest->input());
+        $params['id'] = $dataCitation->id;
```
bozana: maybe to set this after getWriteDisabledErrors...

schemas/dataCitation.json
```diff
@@ -0,0 +1,124 @@
...
+        "relationshipType": {
+            "type": "string",
+            "origin": "setting",
+            "multilingual": false,
+            "description": "The type of relationship between the data and the publication.",
+            "apiSummary": true,
+            "validation": [
+                "in:supporting,generated,analyzed,non-analyzed"
+            ]
+        },
```
bozana, line 79: is relationshipType required ?

schemas/dataCitation.json
```diff
@@ -0,0 +1,124 @@
+{
+	"title": "DataCitation",
+	"description": "A Data Citation in a Publication.",
+	"properties": {
```
bozana, line 9: I think we need to add this too:
	"type": "object",
	"required": [
		"relationshipType",
		"title"
	],

schemas/dataCitation.json
```diff
@@ -0,0 +1,124 @@
...
+            "multilingual": false,
+            "description": "The URL of the dataset.",
+            "apiSummary": true,
+            "validation": [
+                "nullable",
+                "url"
+            ]
+        },
+        "year": {
+            "type": "string",
```
bozana: "type": "integer" ?

bozana:
Hi @ajnyga, great!
it looks really good, great work! I left only a few comments...
Thanks a lot!

plugins/importexport/native/filter/NativeXmlPKPPublicationFilter.php
ajnyga: so no support for native xml at this point?

ajnyga: ok saw the other comment now

classes/security/authorization/internal/DataCitationRequiredPolicy.php
ajnyga: check, removed

api/v1/dataCitations/PKPDataCitationController.php
ajnyga: added to get, getMany should not need it I think

```
        $publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION);
        if ($publication->getId() !== $dataCitation->publicationId) {
            return response()->json([
                'error' => __('api.dataCitations.400.publicationsNotMatched'),
            ], Response::HTTP_FORBIDDEN);
        }
```

schemas/dataCitation.json
ajnyga: yes, i guess overall it is still a bit debatable that which fields should be required. For example do we want to allow data citations without any kind of identifier?

plugins/importexport/native/pkp-native.xsd
```diff
@@ -289,6 +289,7 @@
 			<element ref="pkp:authors" minOccurs="0" maxOccurs="1" />
 			<element ref="pkp:representation" minOccurs="0" maxOccurs="unbounded" />
 			<element ref="pkp:citations" minOccurs="0" maxOccurs="unbounded"/>
+			<element ref="pkp:dataCitations" minOccurs="0" maxOccurs="unbounded"/>
```
bozana: You can remove this. and the <element name="dataCitations"> below

PR #12087, opened 2025-12-01
Body text

jardakotesovec:


api/v1/bodyText/PKPBodyTextController.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Save body text content for a publication
+     */
+    public function save(Request $illuminateRequest): JsonResponse
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+        $publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION);
+
+        Repo::bodyText()->setBodyText(
```
ewhanson, line 143: Is it possible this could fail in a way that the API needs to return an error?

classes/bodyText/BodyTextFile.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+ * @file classes/bodyText/BodyTextFile.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
```
ewhanson: Will need update for copyright years.

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Map the BodyTextFile to an array including body text content and dependent files
+     */
+    public function map(BodyTextFile $bodyTextFile): array
+    {
+        $fileProps = [];
+
+        if ($bodyTextFile->submissionFile) {
+            $submission = Repo::submission()->get($bodyTextFile->submissionId);
+            $schemaMap = Repo::submissionFile()->getSchemaMap($submission, []);
```
ewhanson, line 37: I'm not sure if this will be an issue in practice, but because `$bodyTextFile->submissionId` could be `null`, it's possible `$submission` here could be empty and therefore cause an exception when calling `getSchemaMap()`, no? If it is missing, it could be presumably set (either here or in `BodyTextFile`) based on the submission ID from the publication that is required.

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Create or update the body text file for a publication
+     */
+    public function setBodyText(
+        string $bodyText,
+        int $publicationId,
+        ?int $submissionId = null,
+    ): BodyTextFile {
+        $publication = Repo::publication()->get($publicationId, $submissionId);
```
ewhanson: Do we need to be passing around the `$submissionId` here since it could be null, and it should always be available form the `$publication`. It looks like, in this method at least, this is the only place the original argument `$submissionId` is used, but shouldn't the `$publicationId` be unique enough to retrieve the correct publication and not need the `$submissionId` to be passed in?

ewhanson: As a more general comment on this approach throughout this PR and this class, do we need to be passing in an optionally null `$submissionId` rather than getting it from the required publication? It could still be set in constructors, etc. so it's always available, but it seems like there could be room for error in the future, but maybe I'm missing why the `$submissionId` is being passed around separately and is optional.

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+        $existingBodyTextFile = $this->getBodyTextFile($publicationId, $submission->getId());
+
+        if ($existingBodyTextFile->submissionFile) {
+            return $this->updateBodyTextFile($existingBodyTextFile->submissionFile, $bodyText, $submission);
+        }
+
+        $context = Application::get()->getRequest()->getContext();
+        $user = Application::get()->getRequest()->getUser();
+
+        return $this->createBodyTextFile($bodyText, $publication, $submission, $context, $user);
```
ewhanson, line 90: The thrown exceptions in this method don't look like they're caught anywhere here. Is that on purpose or should they be caught here or at the API controller level?

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Store body text content to file storage
+     */
+    protected function storeContent(string $bodyText, Submission $submission): int
+    {
+        $temporaryFileManager = new TemporaryFileManager();
+        $temporaryFilename = tempnam($temporaryFileManager->getBasePath(), 'bodyText');
+
+        if (!file_put_contents($temporaryFilename, $bodyText)) {
+            throw new Exception('Unable to save body text!');
```
ewhanson: It also looks like this exception is not explicitly caught/handled when this method is called.

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Update existing body text file content
+     */
+    protected function updateBodyTextFile(
+        SubmissionFile $submissionFile,
+        string $bodyText,
+        Submission $submission,
+    ): BodyTextFile {
+        $newFileId = $this->storeContent($bodyText, $submission);
```
ewhanson, line 118: I know `storesContent` throws an exception if it fails, but should the whole process fail, or should it be caught and somehow handled here?

templates/layouts/backend.tpl
```diff
@@ -18,6 +18,9 @@
 	{load_header context="backend"}
 	{load_stylesheet context="backend"}
 	{load_script context="backend"}
+		  <link rel="stylesheet"
+    href="https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@20..48,400,0..1,0" />
```
ewhanson: I don't know if this will always be here or what the specific reasons are, but just wanted to flag that this was being loaded directly with a hard-coded URL. Will this eventually be included or loaded elsewhere? I don't have a good sense of whether this is good or bad practice, but it caught my eye.

classes/bodyText/BodyTextFile.php
```diff
@@ -0,0 +1,65 @@
...
+
+class BodyTextFile
+{
+    public ?string $loadingContentError = null;
+    public ?string $bodyTextContent = null;
+    public bool $isDefaultContent = true;
+
+    public function __construct(
+        public int $publicationId,
+        public ?int $submissionId = null,
```
asmecher: The `$submissionId` appears to be unused. (We can derive it, when needed, from the `$publicationId` anyway -- and I prefer if we avoid redundant and possibly conflicting IDs if we can.)

classes/bodyText/BodyTextFile.php
```diff
@@ -0,0 +1,65 @@
...
+                    'attrs' => ['level' => 1],
+                    'content' => [['type' => 'text', 'text' => 'Introduction']],
+                ],
+                [
+                    'type' => 'paragraph',
+                    'content' => [['type' => 'text', 'text' => 'Start writing your article content here...']],
+                ],
+            ],
+        ]);
+    }
```
asmecher, line 53: (This was just for example purposes, I think, and the hard-coded text can be removed.)

classes/bodyText/Repository.php
```diff
@@ -0,0 +1,198 @@
...
+    public function getBodyTextFile(int $publicationId, ?int $submissionId = null): BodyTextFile
+    {
+        $submissionFileQuery = Repo::submissionFile()
+            ->getCollector()
+            ->filterByFileStages([SubmissionFile::SUBMISSION_FILE_BODY_TEXT])
+            ->filterByAssoc(Application::ASSOC_TYPE_PUBLICATION, [$publicationId]);
+
+        if ($submissionId) {
+            $submissionFileQuery = $submissionFileQuery->filterBySubmissionIds([$submissionId]);
+        }
```
asmecher: Is there a good reason for this optional parameter to be used? I'd prefer if we picked one approach or the other -- either the submission ID is consistently specified, or not -- unless we have a concrete reason for making both modes operate. Again, my preference is to reduce the number of redundant and potentially mismatching IDs as much as possible. I see Erik is picking up on some of the same things in his comments :)

classes/bodyText/Repository.php
asmecher: Someday I'd like to consistently delegate filesystem access through [Flysystem](https://flysystem.thephpleague.com/docs/) (which is Laravel's preference too). This will one day allow us to swap out the filesystem-based implementation for something cloud-backed, for example.

classes/template/PKPTemplateManager.php
```diff
@@ -1493,7 +1493,8 @@ public function display($template = null, $cache_id = null, $compile_id = null,
             'helpUrl' => Application::get()->getHelpUrl(),
             'timeZone' => Config::getVar('general', 'time_zone'),
             'featureFlags' => [
-                'enableNewDiscussions' => Config::getVar('features', 'enable_new_discussions')
+                'enableNewDiscussions' => Config::getVar('features', 'enable_new_discussions'),
+                'enableBodyTextEditor' => Config::getVar('features', 'enable_body_text_editor'),
```
asmecher: I'd be OK if this feature was always enabled, rather than hidden behind the feature flag.

asmecher (changes requested):
Thanks, @jardakotesovec! Some minor comments, aligned with what Erik has suggested.

classes/bodyText/BodyTextFile.php
jardakotesovec: Oki. It was used in Repository.php - but we can get submissionId there from the submissionFile object instead, which I think is fine as well.

classes/bodyText/BodyTextFile.php
jardakotesovec: I think its actually useful having non-empty starting point. Is that ok to keep it?

classes/bodyText/Repository.php
jardakotesovec: On Alec suggestion I removed submissionId from BodyTextFile. And now getting the submissionId from `$bodyTextFile->submissionFile->getData('submissionId')`. Let me know if that looks ok.

classes/bodyText/Repository.php
jardakotesovec: This is really just carry over from the jats implementation which was starting point. I will remove it.

classes/bodyText/Repository.php
jardakotesovec: Oki. Removing $submissionId. This is mostly carry over from the jats implementation which I used as starting point..   But agree it make sense to reduce possible code paths :-)

classes/bodyText/Repository.php
jardakotesovec: This is another jats carry over.  My understanding is that even validation is unnecessary as pretty much all params are defined on backend, so they should be accurate. So this would be more to catch programming error than user error.  But if there is better practice to follow for such use case please let me know.

classes/bodyText/Repository.php
jardakotesovec: @ewhanson So this should not happen unless storage is full or something. So don't think we need to communicate that to the user as validation error or something. I would expect this to end up with 500 error. 

I did see similar scenario in PKPFileService.php..   And this is similar case to me - when it means some serious failure - out of storage etc..

```
    public function add($from, $to)
    {
        $stream = fopen($from, 'r+');
        if (!$stream) {
            throw new Exception("Unable to copy {$from} to {$to}.");
        }

```

In the UI it results in such error.  Which seems reasonable to me. But if there is better practice around that - let me know.
Screenshot 2025-12-08 at 14 33 58

classes/bodyText/Repository.php
jardakotesovec: Commented above.. if you still think there is more that should be done, let me know.

classes/bodyText/Repository.php
jardakotesovec: @asmecher So should I use that - or we don't use it yet at all?

classes/template/PKPTemplateManager.php
jardakotesovec: Oki, it will be visible by default.

templates/layouts/backend.tpl
jardakotesovec: We are avoiding external sources. This is temporary - I added comment there to make it more obvious. Frederik is looking into how to embed the icons with the sciflow editor.

classes/bodyText/Repository.php
ewhanson: Seems good to me!

classes/bodyText/Repository.php
ewhanson: In that case, that sounds good. I agree and think it's totally fine to have catastrophic failure during programming.

classes/bodyText/Repository.php
ewhanson: @jardakotesovec, to my mind the error you show in the UI seems plenty sufficient. Not sure if @asmecher has any other opinion, but otherwise 👍 from me.

classes/bodyText/Repository.php
ewhanson: Nope, seems fine now.

classes/bodyText/BodyTextFile.php
ewhanson: I agree on having some example text, and this may be getting too complicated, but if we keep it, should it be localized?

ewhanson (approved):

classes/bodyText/Repository.php
asmecher: I'm OK with this treatment for serious system errors -- throw an appropriate exception, let the application's exception handling catch it (or not). No need to localize error messages when they're this profound. (There are [some early steps](https://github.com/pkp/pkp-lib/issues/10027#issuecomment-2574187814) towards proper exception handling already in `main`.)

(One of the advantages of Flysystem over the low-level PHP file I/O tools is that they take care of throwing the exceptions, rather than requiring return values to be checked like we were writing C in 1988.)

classes/bodyText/BodyTextFile.php
asmecher: Can we use a placeholder on the control in the UI, rather than actual text potentially in the database? (Yes, it would need to be localized.)

jardakotesovec: On second thought..  word also does not have placeholder document when creating new document, so I am removing it..

Issue #12046, opened 2025-11-19
Add machine-readable options for reviewer recommendations


ewhanson:
In addition to the semantic and configurable text included as options for each of the potential reviewer recommendations, we need to support a machine-readable subset that can be used as part of the open peer review display on article pages/other summaries as well as support conveying peer review results in other machine readable ways (e.g. in downstream metadata services, JATS metadata, OAI-PMH formats, etc.).

The options to be included:
- Accept
- Revisions requested
- Decline
- Comments

These will need to be added as a required option for each reviewer recommendation type.

Unanswered questions:
- [x] How to map existing defaults to these?
- [ ] How to handle these in migrations where we cannot know how users have configured these since they are essentially free text fields.

### Pull Request
pkp-lib: https://github.com/pkp/pkp-lib/pull/12092
ojs: https://github.com/pkp/ojs/pull/5213
omp: https://github.com/pkp/omp/pull/2185
ops:https://github.com/pkp/ops/pull/1153

### Notes

Remaining unanswered question to be addressed and tested in sub-issue.

taslangraham:
Ready for review @ewhanson. PRs listed here https://github.com/pkp/pkp-lib/issues/12046#issue-3643907096

### Changes
- Add a migration introducing a new `type` column on the `reviewer_recommendations` table to map recommendations to their machine readable status.

- Require selecting one of the newly introduced machine readable types when creating a recommendation in the journal settings UI.

- [OJS] Add a migration that assigns the appropriate type to the default recommendations only when the title has not been customised within the journal. Recommendations with customised titles are skipped because the new title may no longer reflect the general meaning of the default recommendation. So applying a type could result in a mismatch between the type applied and the message conveyed by the title. Further work is needed to determine a safe approach for updating recommendations whose titles have been modified.

**Additional notes**:
- There are no default recommendations in OMP and OPS

ewhanson:
Thanks @taslangraham, just a few questions on OJS mainly. It looks like there's a test failing for OJS at least as well around reviewer recommendations that probably needs to be updated as well. Otherwise looks really good!

taslangraham:
Thanks for the review @ewhanson. I've made updates based on your suggestions and fixed the OJS tests. Please take a look.

ewhanson:
Thanks @taslangraham. Looks good and all the tests passed. Merged!

Tribunal33:
This passes QA.  Recommendation Type has been added as required field for reviewer recommendation.

[...]

PR #12092, opened 2025-12-03
pkp/pkp-lib#12046 Introduce machine readable recommendation types

taslangraham:


classes/migration/upgrade/v3_6_0/I12046_AssignMachineReadableRecommendationType.php
```diff
@@ -0,0 +1,46 @@
...
+     * @inheritDoc
+     */
+    public function up(): void
+    {
+        Schema::table('reviewer_recommendations', function (Blueprint $table) {
+            if (!Schema::hasColumn('reviewer_recommendations', 'type')) {
+                $table
+                    ->enum('type', array_column(ReviewerRecommendationType::cases(), 'value'))
+                    ->nullable();
+            }
```
taslangraham, line 33: This check was included because I've added the `type` column to  the `reviewer_recommendations` table in [`classes/migration/install/ReviewerRecommendationsMigration.php`](https://github.com/pkp/pkp-lib/pull/12092/files#diff-84fe5e61752b3242391fbb77ac827a925a43816058ec5605e98786aefba4c790)  install script. However, I noticed that the same install script is also used inside the [upgrade.xml ](https://github.com/pkp/ojs/blob/08cf538f2ec2aad75ce990bceb556998e0fb9b94/dbscripts/xml/upgrade.xml#L195) prior to this migration running, therefore in some cases, the column would have already existed at the time this migration is being executed, resulting in an error. 

The check prevents that error, but I’m not sure an install script should be invoked from `upgrade.xml`. Maybe @touhidurabir can confirm whether that was intentional, since he originally added it.

ewhanson (approved):

classes/migration/upgrade/v3_6_0/I12046_AssignMachineReadableRecommendationType.php
taslangraham: Spoke with Touhidur, and this was intentionally done as a part of the work for https://github.com/pkp/pkp-lib/issues/1660

Issue #12088, opened 2025-12-01
[Templating] Allow mix use of blade/smarty template override for plugin


touhidurabir:
**Description:**
With the merge of https://github.com/pkp/pkp-lib/issues/9968, now it's possible to use blade template view/component along with existing smarty templates . It also allow plugin to have templates or override existing one via blade along with existing smarty capability.

However, for plugin overriding, it's not possible to override an child template to be overridden using the blade template when the parent template is the different one . For example, plugin can override the `templates/frontend/pages/article.tpl` with a blade template when there is a corresponding one at the `PLUGIN_PATH/templates/frontend/pages/article.blade` . But plugin can not override the `templates/frontend/objects/article_details.tpl` with a `PLUGIN_PATH/templates/frontend/objects/article_details.blade` when the parent template in smarty( as `.tpl`) . To do that plugin first need to override the parent smarty template with a blade one . 

This cause some issue with smoother transition from smarty to blade as plugin need to override the parent one even if intend to override the child and may want to use blade template . 

**Task:**
Introduce mechanism so that plugin can override any child templates that is different of parent template . That is as follow 
- [x] Plugin should be able to override a top level Smarty template (parent template) with a blade template
- [x] Plugin should be able to override a child template with a blade template when the top level template (parent template) is Smarty
- [x] For backward compatibility : Replacing a core parent smarty template with a blade one should not break a plugin where it override it with a smarty 
- [x] For backward compatibility : Replacing a core child smarty template with a blade one should not break a plugin where it override it with a smarty 
- [x] For backward compatibility : Allow to override core parent/top level or child level blade/smarty template via non theme plugins (e.g. generic, importexport, ect) using the hook `TemplateResource::getFilename` 

**What application are you using?**
  OJS, OMP or OPS version main (upcoming 3.6)

# Implementation details
# Normalisation
To make the override logic same for both smarty and template. Always the smarty syntax is converted to laravel view:
```
     * - "frontend/pages/article.tpl"      → "frontend.pages.article" (implicit, allows override)
     * - "app:frontend/pages/article.tpl"  → "app::frontend.pages.article" (explicit, skips override)
     * - "core:frontend/pages/article.tpl" → "pkp::frontend.pages.article" (explicit, skips override)
```

Also `getTemplateResource` has been updated to directly generate the laravel view syntax with `pluginName::templatePathWithDots`. 

[...]

PR #12093, opened 2025-12-03
pkp/pkp-lib#12088 Mix override of blade/smarty template via plugin

touhidurabir:
for pkp/pkp-lib#12088 .

### DIFF
--- a/classes/plugins/Plugin.php
+++ b/classes/plugins/Plugin.php
@@ -684,9 +684,6 @@ abstract class Plugin
             . str_replace('.tpl', '.blade', $fullPath);
 
         if ($this->cachedFileExists($bladeViewFilePath)) {
-            // Alternatively we can return view namespaced path as pluginViewNamespace::path.to.blade.template via
-            // `$this->resolveBladeViewPath(str_replace('templates/', '', $path))` but will cause double hook call
-            // return $bladeViewFilePath;
             return $this->resolveBladeViewPath(str_replace('templates/', '', $path));
         }
 
--- a/classes/template/PKPTemplateResource.php
+++ b/classes/template/PKPTemplateResource.php
@@ -169,8 +169,8 @@ class PKPTemplateResource extends \Smarty_Resource_Custom
             return $cache[$path];
         }
 
-        // OPTIMIZATION: Skip view()->exists() for obvious Smarty paths
         // 1. Paths ending with .tpl are Smarty templates
+        //    e.g. Skip view()->exists() for obvious Smarty paths
         if (str_ends_with($path, '.tpl')) {
             $cache[$path] = false;
             return false;
@@ -248,7 +248,7 @@ class PKPTemplateResource extends \Smarty_Resource_Custom
 
         $bladeName = str_replace('/', '.', $bladeName);
 
-        // Register the mapping so callComposer() can do reverse lookup
+        // Register the mapping so callComposer()/callCreator() can do reverse lookup
         $finder->addViewOverride($bladeName, $bladeNamespace);
     }
 }
--- a/classes/core/blade/Factory.php
+++ b/classes/core/blade/Factory.php
@@ -1,5 +1,31 @@
 <?php
 
+/**
+ * @file classes/core/blade/Factory.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Factory
+ *
+ * @brief Custom Laravel View Factory with plugin template override support
+ *
+ * Extends Laravel's View Factory to ensure view composers and creators fire correctly
+ * when plugins override core templates. Key responsibilities:
+ *
+ * 1. Call find() before getOverriddenView() so viewOverrides mapping is populated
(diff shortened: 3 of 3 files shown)


classes/plugins/Plugin.php
```diff
@@ -543,7 +542,7 @@ public function _overridePluginTemplates($hookName, $args)
      *
      * @return string|null
      */
-    private function _findOverriddenTemplate($path)
+    public function _findOverriddenTemplate($path)
```
asmecher: No problems with moving this to `public`, but please also remove the `_` prefix from the function name.

classes/plugins/Plugin.php
```diff
@@ -552,12 +551,10 @@ private function _findOverriddenTemplate($path)
...
-            if (view()->exists($bladePath)) {
-                return $bladePath;
-            }
+        // Fallback to blade view if exists for theme plugins
+        $bladePath = $this->resolveBladeViewPath(str_replace('templates/', '', $path));
+        if (view()->exists($bladePath)) {
+            return $bladePath;
         }
 
         // Backward compatibility for OJS prior to 3.1.2; changed path to templates for plugins.
```
asmecher: This clause is prehistoric and should be removed.

asmecher: Also, it's below the scope of this PR, so I can't comment directly -- but watch out for [line 572](https://github.com/pkp/pkp-lib/pull/12093/files#diff-0cff6285dcf4edb36419ecacd8026325dbb8c807a337ad5c18a96e54bb60372aL572). `ThemePlugin` is not specified in a `use` statement, so when aliases go, the check will break!

classes/plugins/ThemePlugin.php
```diff
@@ -1081,4 +1084,32 @@ protected function requiresVueRuntime()
...
+     * - 'frontend/pages/article' → 'templates/frontend/pages/article.tpl'
+     * - 'frontend/pages/article.blade' → 'templates/frontend/pages/article.tpl'
+     * - 'frontend/pages/article.tpl' → 'templates/frontend/pages/article.tpl'
+     *
+     * @param string $templatePath Template path in various formats
+     * @return string Smarty template path
+     */
+    public function convertToSmartyTemplatePath(string $templatePath): string
+    {
+        $path = str_replace(['.blade.php', '.blade', '.tpl'], '', $templatePath);
```
asmecher: Please remove the `.blade.php` suffix from this -- I'd rather we didn't accidentally let support for these slip into the codebase until we've properly separated the web root from the installation (code) path.

classes/template/PKPTemplateResource.php
```diff
@@ -49,6 +52,20 @@ public function __construct($templateDir)
     public function fetch($name, &$source, &$mtime)
     {
         $filename = $this->_getFilename($name);
+
+        if ($this->isBladeViewPath($filename)) {
+            try {
+                $mtime = time();
+                $templateManager = TemplateManager::getManager(Application::get()->getRequest());
+                // $templateManager->shareTemplateVariables($templateManager->getTemplateVars());
```
asmecher: (remove commented code)

classes/template/PKPTemplateResource.php
```diff
@@ -49,6 +52,20 @@ public function __construct($templateDir)
...
+                $source = view($filename, $templateManager->getTemplateVars())->render();
+                return true;
+            } catch (Throwable $exceptin) {
+                error_log("Error rendering Blade view '{$filename}': " . $exceptin->getMessage());
+                throw $exceptin;
```
asmecher: *exception; also, is it necessary to include both an `error_log` and a `throw`? The `Throwable` should have enough detail in it already, I think.

classes/template/PKPTemplateResource.php
```diff
@@ -67,7 +84,14 @@ public function fetch($name, &$source, &$mtime)
      */
     protected function fetchTimestamp($name)
     {
-        return filemtime($this->_getFilename($name));
+        $filename = $this->_getFilename($name);
+
+        // Check if this is a Blade view path (namespace notation)
+        if ($this->isBladeViewPath($filename)) {
+            return time(); // Return current time for Blade views (always fresh)
```
asmecher, line 124: Have you worked through the Blade compilation behaviour? This looks wrong to me, but I'm not sure what the appropriate behaviour would be at a glance.

classes/template/PKPTemplateResource.php
```diff
@@ -89,4 +113,21 @@ protected function _getFilename($template)
         Hook::call('TemplateResource::getFilename', [&$filePath, $template]);
         return $filePath;
     }
+
+    /*
+     * Detect Blade view namespace
+     */
+    private function isBladeViewPath($path)
```
asmecher: Needs typehints.

classes/template/PKPTemplateResource.php
touhidurabir: This is intentional . here at this point for case of `SMARY --> BLADE` override, as we can not pass a blade template to smarty to processing, we are just taking a complied/rendered blade content . As also this is a HTML content as this point , no point in caching as smarty and blade has it's own caching . but Smarty still expects a cache time so we just pass one even if it has no impact for specific case of `SMARY --> BLADE` override .

asmecher: Thanks!

PR #12095, opened 2025-12-03
pkp/pkp-lib#12949 Support publish-review-curate model for review round creation

ewhanson:


classes/publication/Repository.php
```diff
@@ -1003,4 +1003,17 @@ public function getPeerReviews(array $publications): Enumerable
                 return (new PublicationPeerReviewResource($publication))->resolve();
             })->values();
     }
+
+    /**
+     * Returns the provided publication ID as well as any other publications
+     * that reference this publication via the `source_publication_id`.
+     */
+    public function getWithSourcedPublications(int $publicationId): Collection
```
bozana: maybe to call it getWithSourcedPublicationIds -- because it returns IDs?

Issue #11590, opened 2025-07-02
[Implementation] Consider different DOIs for different versions in Crossref


bozana:
Now that we have introduced JAV and versions, and that we are introducing different DOIs for different versions, we need to consider this different versions when depositing DOIs to Crossref.
AO will be considered as preprint -- the XML element would need to be posted_content (and not journal_article).
AM and VoR will be considered as journal_article element. We need to see with Crossref if PMUR could be added to the list of accepted content types too.
Currently it seems that SM and PF would not be considered.

ewhanson:
Thanks @bozana. Just a few questions really. 

I wasn't able to directly validate and test the exported XML because I was running into seemingly unrelated issues with my OJS install and environment, but I would test the export XML for a few different scenarios against https://www.crossref.org/02publishers/parser.html if you haven't already. Otherwise looks good!

bozana:
Hi @ewhanson, thanks a lot!
I addressed a few comments, and answered some questions.
I do not know how to solve the reloading page on DOI setting changes now -- I would then maybe ask Jarda later, if you do not know.
I have validated and tested the export and deposit, and now I also validated the XMLs with that Crossref parser, for ca. following cases:
- submission with a publication without an issue
- submission with two publications without an issue
- submission with several publications with an issue

and the export of
- one and several selected submissions above with one DOI per submission
- one and several selected submissions above with different DOIs for different versions.

I rebased, so if you would like/could take another look?

Thanks a lot!

bozana:
Hi @Tribunal33, this is ready for QA :-)

Crossref Management Plugin needs to be enabled, DOIs should be set up, and Crossref should be chosen as registration agency for DOIs and set up.
- If no DOI versioning is selected in DOI settings, then the resulting Crossref XML should be as till now, it will contain the same elements as till now (except that now it also contains the citations without the need to use CrossrefReferencePlugin) -- the current published publication that has a DOI is considered.
- If DOI versioning is selected -- different DOIs for different versions -- one needs to also enter journal update policy DOI in Crossref settings. (If the versioning settings changes, a reload is needed to see the updated Crossref setting form -- I need to ask Jarda if this can be changed).
Then all submission's exportable versions (the latest minor version of a major version, that is published and has a DOI) should be exported within the same XML. See also the issue description -- what should be there and how. 

Thanks a lot!

bozana:
Maybe a note that this still needs to be ported to crossref-ops as well...

[...]

PR #12098, opened 2025-12-04
pkp/pkp-lib#11590 consider submissions with exportable DOIs, fix citation type

bozana:
s. https://github.com/pkp/pkp-lib/issues/11590


classes/publication/DAO.php
```diff
@@ -490,4 +495,31 @@ public function getMinorVersionsSettingValues(int $submissionId, string $version
...
+    public function getExportableDOIsSubmissionIds(int $contextId, bool $doiVersioning): array
+    {
+        return DB::table('publications as p')
+            ->select(['p.submission_id'])
+            ->join(
+                'submissions as s',
+                function (JoinClause $join) use ($doiVersioning) {
+                    $join
+                        ->on('p.submission_id', '=', 's.submission_id')
+                        ->when((!$doiVersioning), function (Builder $qb) {
```
ewhanson, line 514: This all looks right, I just want to check my understanding. To be clear, this `when` clause is saying, if there is no DOI versioning, include to this join clause `s.current_publication_id = p.publication_id` in addition to `p.submisison_id = s.submission_id`.

bozana: Hi @ewhanson, yes, correct, so in that case only current publications should be filtered

ewhanson (approved):

Issue #11634, opened 2025-07-11
CI build for OJS 3.5.0 has issue with weblateLanguages/languages.json


bdcaf:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

I am seeing a weird issue with gitlab ci docker images for ojs 3.5.0. they would miss ./lib/pkp/lib/weblateLanguages/languages.json and ojs would stop with error:

```
[11-Jul-2025 09:39:40 Europe/Vienna] PHP Warning:  require_once(/var/www/html/plugins/generic/citationStyleLanguage/lib/vendor/autoload.php): Failed to open stream: No such file or directory in /var/www/html/plugins/generic/citationStyleLanguage/CitationStyleLanguagePlugin.php on line 19
[11-Jul-2025 09:39:40 Europe/Vienna] Instantiation of the plugin generic/citationStyleLanguage has failed
Error: Failed opening required '/var/www/html/plugins/generic/citationStyleLanguage/lib/vendor/autoload.php' (include_path='.:/usr/share/php83') in /var/www/html/plugins/generic/citationStyleLanguage/CitationStyleLanguagePlugin.php:19
Stack trace:
#0 /var/www/html/lib/pkp/lib/vendor/composer/ClassLoader.php(576): include()
#1 /var/www/html/lib/pkp/lib/vendor/composer/ClassLoader.php(427): Composer\Autoload\{closure}()
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass()
#3 /var/www/html/lib/pkp/classes/plugins/PluginRegistry.php(209): class_exists()
#4 /var/www/html/lib/pkp/classes/plugins/PluginRegistry.php(263): PKP\plugins\PluginRegistry::instantiatePlugin()
#5 /var/www/html/lib/pkp/classes/plugins/PluginRegistry.php(128): PKP\plugins\PluginRegistry::loadFromDisk()
#6 /var/www/html/lib/pkp/classes/cliTool/CommandLineTool.php(72): PKP\plugins\PluginRegistry::loadCategory()
#7 /var/www/html/lib/pkp/classes/cliTool/UpgradeTool.php(40): PKP\cliTool\CommandLineTool->__construct()
#8 /var/www/html/tools/upgrade.php(21): PKP\cliTool\UpgradeTool->__construct()
#9 {main}
```

### Steps to Reproduce

I realize this might depend on the runner config. But I can reliable reproduce it in our setup.

minimal dockerfile:

```
[...]

PR #12118, opened 2025-12-10
ComposerScript - for CI use cURL for weblate download by composer

bdcaf:
This is my workaround to issue https://github.com/pkp/pkp-lib/issues/11634

To recap: in a sandboxed CI build (that requires use of a proxy) the composer install script fails to  download a file using stream. 
During debugging it was odd that composer was downloading its other files fine. it turns out that composer relies on curl, which automagically configures itself through environment variables. It may be possible to reconfigure the stream as well but it turns out quite a complicated matter. As curl is required anyway by composer to be configured correctly, I decided to reuse that.

### DIFF
--- a/classes/dev/ComposerScript.php
+++ b/classes/dev/ComposerScript.php
@@ -48,11 +48,16 @@ class ComposerScript
                 mkdir($dirPath);
             }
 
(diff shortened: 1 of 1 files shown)


asmecher:
Simple enough and works well -- thanks, @bdcaf! I'll cherry-pick this forward to `main` too.

Issue #11887, opened 2025-10-01
[DOIs] | Submissions published in early stages (Review/Submission) are not putting the article in DOI management because it hasn't update the stage to published


Tribunal33:
**Description:**
Not sure what the requirement would be for this.  There are two potential options (pick one).
- [ ] Update the stage status to published even in early stages (this might fix the issue)
- [x] Update the DOI management to include those submissions published in early stages and leave the stage alone

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
1. Precondition: Must have the Distribution Setting > DOIs > Automatic DOI Assignment set to `Upon Publication` + DOI Versioning set to `Yes, assign a unique DOI to every version...`
2. Precondition step (rest of the settings are at default and DOI prefix is filled in)
<img width="370" height="114" alt="Image" src="https://github.com/user-attachments/assets/c2afecde-0d1a-48b6-99d3-07148934610b" />

<img width="729" height="196" alt="Image" src="https://github.com/user-attachments/assets/73080a28-f5e1-4dd7-99fb-a04c6159c0ac" />

3. Either create a new submission or pick an existing one that is in the submission or review stage
4. Click on `schedule for publication`
5. Complete form and select Publish
6. Navigate to the DOI management page (use side nav `DOIs`)

**What application are you using?**
OJS/OMP/OPS 3.6(main)

**Additional information**
Please add any screenshots, videos or other information we can use to understand this dev task report.

<img width="2109" height="959" alt="Image" src="https://github.com/user-attachments/assets/ad78dbd5-807c-45fb-b31e-10e1b347e174" />

<hr>

PRs:
- https://github.com/pkp/pkp-lib/pull/12119
- https://github.com/pkp/ojs/pull/5135
- https://github.com/pkp/omp/pull/2197
- https://github.com/pkp/ops/pull/1164

bozana:
I think I need to ask @ewhanson again: with our new workflows, shell the DOI management page list all submissions, no matter in which stage they are. Currently only submissions in WORKFLOW_STAGE_ID_EDITING and WORKFLOW_STAGE_ID_PRODUCTION are considered, s. https://github.com/pkp/ojs/blob/main/pages/dois/DoisHandler.php#L45?
I think yes, right?

bozana:
Hi @ewhanson, could you please review the changes here?

bozana:
[...]

PR #12119, opened 2025-12-10
pkp/pkp-lib#11887 filter submissions that should be listed on the DOI…

bozana:
… management page

s. https://github.com/pkp/pkp-lib/issues/11887

### DIFF
--- a/classes/submission/Collector.php
+++ b/classes/submission/Collector.php
@@ -454,9 +454,11 @@ abstract class Collector implements CollectorInterface, ViewsCount
         // search phrases starting with number followed by a '.'  will be interpreted as a DOI identifier. E.g: 10.1
         $isSearchPhraseDoi = Doi::beginsWithDoiPrefixPattern($this->searchPhrase ?: '');
         // Prepare keywords (allows short and numeric words)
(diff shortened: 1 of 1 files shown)


classes/submission/Collector.php
```diff
@@ -156,6 +157,19 @@ public function filterByHasDois(?bool $hasDois, ?array $enabledDoiTypes = null):
...
 
+    /**
+     * Limit results to submissions that should be listed on the DOI management page
+     *
+     * @param array|null $enabledDoiTypes TYPE_* constants to consider when checking submission has DOIs
+     */
+    public function filterByOnDoiPage(?bool $onDoiPage, ?array $enabledDoiTypes = null): AppCollector
+    {
+        $this->onDoiPage = $onDoiPage;
+        $this->enabledDoiTypes = $enabledDoiTypes === null ? [Repo::doi()::TYPE_PUBLICATION] : $enabledDoiTypes;
```
ewhanson: If there are no enabled DOI types, should we not default to not showing them at all? Or does this break the logic somehow?

ewhanson (changes requested):
Looks good, just one question about the default enabled DOI types in the filter function.

Issue #6851, opened 2021-03-10
Let OJS Read HTTP_X_FORWARDED_PROTO instead of trying to discover it


marcbria:
**Describe the problem you would like to solve**
Example: Our editors need a way to [...]

**Describe the solution you'd like**
Tell us how you would like this solution to be solved.

**Who is asking for this feature?**
Tell us what kind of users are requesting this feature. Example: Journal Editors, Journal Administrators, Technical Support, Authors, Reviewers, etc.

**Additional information**
Add any other information or screenshots about the feature request here.

asmecher:
Note that any solution that relies on request headers to determine the protocol will need to consider URL generation from command-line tools e.g. when a CRON job sends off review reminder emails containing URLs.

marcbria:
Hummm... didn't though in this. :_-(

I'm unsure if getProtocol() will be able to get the right protocol based on SCRIPT_URI, but in worst of cases cron will send a mail with an http petition that will be redirected to httpS when it arrives to the server.

But jumping to a different approach... why "force_ssl" is not forcing the protocol in getProtocol()?

asmecher:
> Jumping to a different approach... why "force_ssl" is not forcing the protocol in getProtocol()?

The `force_ssl` setting is checked in an authorization policy written in `PKPHandler`:

https://github.com/pkp/pkp-lib/blob/stable-3_3_0/classes/handler/PKPHandler.inc.php#L275..L279

If `force_ssl` is set, and `PKPRequest::getProtocol()` doesn't report `https` protocol (based on `$_SERVER['HTTPS']`), the authorization policy recommends redirecting to `https`.

marcbria:
I really thought this was going to be an easy one. :-(

Tomorrow I will take a look but because I suspect command-line tools will take the protocol properly even without headers, or we won't mind much if they are going through http... 

What other features of ojs are not reading headers apart of command-line tools?

Trying to evaluate if a a non perfect implementation would be better than nothing.

marcbria:
Checking what other CMS are doing:

In WP they recognize the issue and offer a couple of workarrounds:

![image](https://user-images.githubusercontent.com/2779168/110756872-03a16900-824b-11eb-98e3-5d2456e496d9.png)

- Their "FORCE_SSL_ADMIN" is like our "force_ssl"... to check authorization, not protocol.
- They **don't implement** "HTTP_X_FORWARDED_PROTO".

[...]

PR #12132, opened 2025-12-15
Enhance getProtocol method to handle X-Forwarded-Proto

marcbria:
Prioritize X-Forwarded-Proto header for protocol detection. Implementing what is proposed at #6851

### DIFF
--- a/classes/core/PKPRequest.php
+++ b/classes/core/PKPRequest.php
@@ -395,8 +395,13 @@ class PKPRequest
     public function getProtocol(): string
     {
         if (!isset($this->_protocol)) {
-            $this->_protocol = (!isset($_SERVER['HTTPS']) || strtolower($_SERVER['HTTPS']) != 'on') ? 'http' : 'https';
-            Hook::call('Request::getProtocol', [&$this->_protocol]);
+            // If defined, priorizes X-Forwarded-Proto header (#6851)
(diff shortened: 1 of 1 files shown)


classes/core/PKPRequest.php
```diff
@@ -395,8 +395,13 @@ public function getServerHost(mixed $default = null, bool $includePort = true):
...
     {
         if (!isset($this->_protocol)) {
-            $this->_protocol = (!isset($_SERVER['HTTPS']) || strtolower($_SERVER['HTTPS']) != 'on') ? 'http' : 'https';
-            Hook::call('Request::getProtocol', [&$this->_protocol]);
+            // If defined, priorizes X-Forwarded-Proto header (#6851)
+            if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
+                $this->_protocol = strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']);
+            } else {
+                $this->_protocol = (!isset($_SERVER['HTTPS']) || strtolower($_SERVER['HTTPS']) != 'on') ? 'http' : 'https';
+                Hook::call('Request::getProtocol', [&$this->_protocol]);
```
asmecher, line 403: This removes the hook's ability to override the returned value in all cases, which might be important. I'd suggest promoting this outside of this conditional.

asmecher (commented):
I'll add a more general comment on the issue, but here's a detail...

Issue #11645, opened 2025-07-16
Upgrade process interrupted due to invalid context IDs at the access_keys table


jonasraoni:
**Description:**
The migration `lib\pkp\classes\migration\upgrade\v3_5_0\I9197_MigrateAccessKeys.php` is attempting to use existing context IDs from the `access_keys.context` column (notice it's a `varchar` column that might also have string values). The entries are migrated to the new `invitations` table, and it might fail the foreign key validation if the context ID doesn't exist (i.e. orphaned entries).

A cleanup of invalid context IDs is performed at the `lib\pkp\classes\migration\upgrade\v3_4_0\PreflightCheckMigration.php`, but due to the unexpected behavior of this column (a varchar that might have strings or context IDs), its cleanup wasn't addressed and no foreign keys were introduced.

**Solution:**
Clear the entries before running the upgrade

**What application are you using?**
OJS 3.5

---

**PRs**
OJS: pkp/ojs#5232 **TESTS ONLY**
pkp-lib: #12143

asmecher:
@defstat, could you have a look at this one?

asmecher:
Deferring to 3.5.0-3; this will need a look before the LTS stamp goes on.

defstat:
@asmecher @jonasraoni I have added a proposed fix for that in the PR found in the initial comment. @jonasraoni could you review the code? Still waiting for the tests to pass.

asmecher:
@jonasraoni, do you have time to code review this one?

asmecher:
@jonasraoni, a reminder about this code review, thanks! Also for 3.5.0-4 possibly LTS.

jonasraoni:
Ok, I'll check 👀
Edit: Done 😁

Tribunal33:
@jonasraoni This does not require QA approval.  Please validate and provide evidence that it has been fixed.  Then feel free to close 😄

asmecher:
@jonasraoni I agree with your review comments, however this is migration code (which I hold to a lower syntactical standard) and none of the comments are show-stoppers, so I've taken the liberty of merging this (and forward-porting to `main`) in case we get 3.5.0-4 released tomorrow.

@defstat, if you could add some evidence of the PR fixing the problem (e.g. before-and-after screenshots of the upgrade failing then passing), I'd be grateful!

asmecher:
[...]

PR #12143, opened 2025-12-16
[pkp-lib][stable-3_5_0] #11645 Prevent FK violations when migrating access_keys to invitations

defstat:


classes/migration/upgrade/v3_5_0/I9197_MigrateAccessKeys.php
```diff
@@ -81,25 +81,40 @@ public function up(): void
...
         foreach ($accessKeys as $accessKey) {
             $invitation = null;
 
             if ($accessKey->context == 'RegisterContext') { // Registered User validation Invitation
                 $invitation = new RegistrationAccessInvite();
                 $invitation->initialize($accessKey->user_id, null, null, null);
             } elseif (isset($accessKey->context) && is_numeric($accessKey->context)) { // Reviewer Invitation
+                $contextExists = DB::table($contextDao->tableName)
+                    ->where($contextDao->primaryKeyColumn, (int) $accessKey->context)
+                    ->exists();
```
jonasraoni, line 95: As there's an outer loop, I think this should be cached.

classes/migration/upgrade/v3_5_0/I9197_MigrateAccessKeys.php
```diff
@@ -81,25 +81,40 @@ public function up(): void
...
-                ]));
+                $this->warnAndDelete(
+                    $accessKey,
+                    'Unsupported access key in access_keys table'
+                );
+
+                continue;
             }
 
             if (isset($invitation)) {
```
jonasraoni, line 120: As you've added a `continue` above, then this check can be removed, `$invitation` should be always set at this point.

classes/migration/upgrade/v3_5_0/I9197_MigrateAccessKeys.php
```diff
@@ -117,6 +132,26 @@ public function up(): void
         Schema::dropIfExists('access_keys');
     }
 
+    /**
+     * Log a warning and delete an invalid access key record.
+     */
+    protected function warnAndDelete(object $accessKey, string $reason): void
+    {
+        error_log(
```
jonasraoni, line 140: I think we should be using `$this->_installer->log()`

classes/migration/upgrade/v3_5_0/I9197_MigrateAccessKeys.php
```diff
@@ -117,6 +132,26 @@ public function up(): void
...
+            json_encode([
+                'id' => $accessKey->access_key_id,
+                'context' => $accessKey->context,
+                'assoc_id' => $accessKey->assoc_id,
+                'expiry_date' => $accessKey->expiry_date,
+            ])
+        );
+
+        DB::table('access_keys')
+            ->where('access_key_id', $accessKey->access_key_id)
```
jonasraoni, line 151: Is is possible for the `$accessKey->access_key_id` to be `null`? If yes, then this needs to be addressed here.

jonasraoni (changes requested):
I left minor suggestions for the PR.

Regarding the migration, I see we're using application classes, and this breaks the resilience of the migration (small changes to the code might cause unexpected side-effects), if it's not tough to address it, I think it's better to fix while we still have time.

PR #12153, opened 2025-12-17
I12088 Attempt to have unified file/view resolution for smarty&blade

jardakotesovec:


classes/core/blade/SmartyTemplate.php
```diff
@@ -0,0 +1,249 @@
...
+        $cache_lifetime,
+        $data,
+        $scope,
+        $forceTplCache,
+        $uid = null,
+        $content_func = null
+    ): void {
+        // Check if template has a Smarty built-in resource prefix (file:, string:, eval:)
+        // These are special Smarty resources that should be handled directly by Smarty
+        if ($this->isSmartyBuiltinResource($template)) {
```
jardakotesovec: Not sure if this scenario that someone would be using directly file: would be ever needed.. but keeping for compatibility

classes/template/PKPTemplateManager.php
```diff
@@ -1602,26 +1712,8 @@ public function display($template = null, $cache_id = null, $compile_id = null,
...
-            echo $template instanceof \Illuminate\View\View
-                ? $template->render()
-                : view($template)->render();
-            return;
-        }
-
-        // Actually display the template.
-        parent::display($template, $cache_id, $compile_id, $parent);
+        // Use fetch() for unified template rendering, then output
+        echo $this->fetch($template, $cache_id, $compile_id, $parent);
```
touhidurabir, line 1678: This is causing issue with trying to render a blade template as 
```php
$templateMgr->display('frontend.pages.article');
```
somehow it's still taking into the existed `.tpl` . It will work if the .tpl is deleted or renamed or passed as `$templateMgr->display(view('frontend.pages.article'))`

Also interestingly , if we have a article.blade and article.tpl, and calling as `$templateMgr->display(view('frontend.pages.article'))` still taking the article.tpl which seems like having implicit higher priority on smarty based on search and availability even when specifically passing a view instance .

classes/template/PKPTemplateManager.php
```diff
@@ -1332,41 +1338,145 @@ public function setupBackendPage()
...
+
+        // Parse the resource prefix to extract plugin name
+        // Format: plugins-{contextId}-{pluginPath with / as -}-{category}-{pluginName}
+        // Example: plugins-0-plugins-generic-funding-generic-funding
+        // The last segment is the plugin name (used as Laravel namespace)
+        $segments = explode('-', $resourcePrefix);
+        $pluginName = array_pop($segments);  // Last segment is plugin name
+
+        // Return Laravel namespaced view: "funding::listFunders"
+        return "{$pluginName}::{$templatePath}";
```
touhidurabir: Here we are assuming that plugin name is registered as view namespace but plugin can also over write it . so any plugin overriding the `Plugin::getTemplateViewNamespace` method will have a different view namespace .

classes/plugins/Plugin.php
```diff
@@ -344,25 +344,30 @@ public function getDirName()
...
+            $viewPath = str_replace(['/', '\\'], '.', $viewPath);
 
-            if (view()->exists($bladeTemplatePath)) {
-                // share the template variables to the blade view
-                $templateManager = TemplateManager::getManager(Application::get()->getRequest());
-                $templateManager->shareTemplateVariables($templateManager->getTemplateVars());
-                
-                return $bladeTemplatePath;
-            }
+            return $this->getTemplateViewNamespace() . '::' . $viewPath;
```
touhidurabir: this assumes that a corresponding blade view is always available but that may not be case for plugin not migrating to blade yet .

touhidurabir (changes requested):
@jardakotesovec added few review comments and will add more later . There are also few general case as follow 
1. with this mechanism, we lost the ability to auto inject plugin view namespace as defined in the Plugin::_registerTemplateViewNamespace as 
```php
// Make sure to set the view namespace always available for plugins view
        View::composer(
            "{$viewNamespace}::*",
            fn (\Illuminate\View\View $view) => $view->with('viewNamespace', $viewNamespace)
        );
```
I think this will cause plugin to unable to use the auto resolve to view component namespace with the use of macro `resolvePluginComponentViewPath` defined in the PKPBladeViewServiceProvider . 
the same way plugin can not use the view composer to have something available through the view namespace only .
2. The view composer and creator does not work with child/included level blade views . so the following will work 
```php
\Illuminate\Support\Facades\View::creator('frontend.pages.article', fn($v) => $v->with('creator', 'creator'));
                \Illuminate\Support\Facades\View::composer('frontend.pages.article', fn($v) => $v->with('composer', 'composer'));
```
but not the following 
```php
\Illuminate\Support\Facades\View::creator('frontend.objects.article_details', fn($v) => $v->with('creator', 'creator'));
                \Illuminate\Support\Facades\View::composer('frontend.objects.article_details', fn($v) => $v->with('composer', 'composer'));
```
and to make it work , I think we have to override core view factory and also keep a mapping of overrides in our overridden view finder . 
3. I feel like we are mixing the smarty and blade templating together here . So even for a smarty template, it will pass to smarty templating and then again to our custom smarty engine . Just not sure how this will unfold in long run . 

Will add more details as I test it more .

touhidurabir:
BTW, the updated PR of https://github.com/pkp/pkp-lib/pull/12093 now contains a different mechanism to handle proper view composer/creator feature (with the override of view factory and using the view namespace path instead of absolute path) and fix of double hook call.

classes/template/PKPTemplateManager.php
jardakotesovec: I think its actually good thing that it does not matter if you use smarty file name syntax or blade view syntax. Its intentionally normalising everything to laravel view names and than it goes from there.  I think there is benefit to control what will be rendered only by the extensions - so whatever file is available (tpl/blade) it will use that one.

jardakotesovec:
I tried to reproduce 2) case and it seems to be working fine for me..  hope I understood the concern.

I created article.blade and article_details.blade in the defaultManuscript theme..  and than used composer like this:

```
            View::composer('frontend.pages.article', function ($view) {
                $view->with('testComposerData', 'ARTICLE COMPOSER DATA ✅');
            });

            // Register composer for article details included view
            View::composer('frontend.objects.article_details', function ($view) {
                $view->with('testDetailsComposerData', 'ARTICLE DETAILS COMPOSER DATA ✅');
            });


```

And both `testComposerData` and `testDetailsComposerData` were available in the correct files. What am I missing?

jardakotesovec:
@touhidurabir 1) Case also works for me..  If plugins is using
```
            $output .= $templateMgr->fetch($this->getTemplateResource('listFunders.tpl'));

```
getTemplateResource adds the namespace..  so the following works as expected for me. Is that the scenario you had in mind?
```
  View::composer(
            "{$viewNamespace}::*",
            fn (\Illuminate\View\View $view) => $view->with('viewNamespace', $viewNamespace)
        );
```

touhidurabir:
> @touhidurabir 1) Case also works for me.. If plugins is using
> 
> ```
>             $output .= $templateMgr->fetch($this->getTemplateResource('listFunders.tpl'));
> ```
> 
> getTemplateResource adds the namespace.. so the following works as expected for me. Is that the scenario you had in mind?
> 
> ```
>   View::composer(
>             "{$viewNamespace}::*",
>             fn (\Illuminate\View\View $view) => $view->with('viewNamespace', $viewNamespace)
>         );
> ```

My test case is that try to use the default theme plugin and no need for `$output .= $templateMgr->fetch($this->getTemplateResource('listFunders.tpl'));` . just have what you have e.g. have the article.blade or article_details.blade in the theme and it should not work.

touhidurabir:
> I tried to reproduce 2) case and it seems to be working fine for me.. hope I understood the concern.
> 
> I created article.blade and article_details.blade in the defaultManuscript theme.. and than used composer like this:
> 
> ```
>             View::composer('frontend.pages.article', function ($view) {
>                 $view->with('testComposerData', 'ARTICLE COMPOSER DATA ✅');
>             });
> 
>             // Register composer for article details included view
>             View::composer('frontend.objects.article_details', function ($view) {
>                 $view->with('testDetailsComposerData', 'ARTICLE DETAILS COMPOSER DATA ✅');
>             });
> ```
> 
> And both `testComposerData` and `testDetailsComposerData` were available in the correct files. What am I missing?

sorry , I should have been more detailed , as I have mentioned as 

> The view composer and creator does not work with child/included level blade views 

it is when the parent in smarty , e.g. article.tpl and have a article_details.blade . my test case was the again the default theme and have only a article_details.blade and use the core's article.tpl . 

basically the mixup of blade/smarty mixup for parent/child case .

jardakotesovec:
@touhidurabir Thanks! Keep referencing the issues by number.. so hopefully its clear what I am responding to.

1) When adopting laravel views. Its always  `view name ` -> `view file path` resolution. This is resolution which I am currently using for overrides. So if `frontend.objects.article_details` is passed - it keeps the name and just find the file path inside the plugin. As result as you pointed out you can't really specifically reference it via namespace. I am thinking whether its good thing or bad thing. Alternative is instead of overriding file path would be overriding view name.  THan the file resolution should work by default mechanism.

With that we could have mechanism where `frontend.objects.article_details` is requested, but it gets overridden (via hook) to `defaultManuscript::frontend.objects.article_details` for example.

Only downside I can think of is that the view name gets changed. So if the composer is referencing to it as `frontend.objects.article_details`.. is suddenly becomes `defaultManuscript::frontend.objects.article_details`, so it would need to be generally referenced as `*::frontend.objects.article_details` in the composer.

2) I see..  good catch. That seems to be easy fix. Problem why this happened is because smarty forced me to configure some default namespace, which is "app"..  Therefore it ends up requesting app::frontend.bojects.article_details. That can be adjusted and this namespace coming from smarty can be removed.

classes/plugins/Plugin.php
jardakotesovec: Not really.. Its just embracing Laravel View system..   so always first thing what I do is to normalise whatever template is requested to Laravel view name with namespace and dots. 
That basically removes the challenge of thinking about different template names for smarty vs blade. It gets normalised to the laravel view name (namespace::a.b.c) and than it goes through laravel view system to look up the file..  and extension is used to pick the correct rendering engine.

classes/template/PKPTemplateManager.php
jardakotesovec: Fair point - but is not pluginName unique enough for namespacing - we could just not let people override that.. don't see harm in that.

touhidurabir:
@jardakotesovec thanks . 

But for no 1, I am not sure I understand the path issue here as view composer/creator can attach view data to view namespace(not to specific view path only) without any path reference which is very common in blade view templates via `{$viewNamespace}::*` . So a plugin can generally define some bit of view data to it's view namespace so that it's auto available to plugins every view as plugin templates are do needed to be namespaced . 

This mechanism allow plugin to get specific data available to it's all views or even utilise it in view component without passing specific data to parent --> child1 --> child2 which for reason the composer/creator view data binding to view namespace has been introduced .

jardakotesovec:
@touhidurabir How is this particular case handled in your implementation?

My point was that if the view name changes (from: `frontend.objects.article_details` to `defaultManuscript::frontend.objects.article_details`), because its being overridden in plugin - that could break other plugins or directly ojs if its using `View::composer('frontend.objects.article_details',...)`

So I get the convenience you are mentioning. But my concern is that it would cause other troubles. But maybe you have some way to have alias to the view name - so both works?

They still could use the `FunderPlugin::` namespacing if they override article_details.. but it would need to be more hardcoded. Which could be also ok.

classes/template/PKPTemplateManager.php
touhidurabir: > Fair point - but is not pluginName unique enough for namespacing - we could just not let people override that.. don't see harm in that.

true . I dont have any problem with that . just seems bit too restrictive . so if dont allow the `{$viewNamespace}::*` feature, then plugin need to pass around the plugin name or plugin instance to views to get a hold of plugin namespace which has some need in some case of view component usage of plugin .

touhidurabir:
> My point was that if the view name changes (from: `frontend.objects.article_details` to `defaultManuscript::frontend.objects.article_details`), because its being overridden in plugin - that could break other plugins or directly ojs if its using `View::composer('frontend.objects.article_details',...)`

I think we both are misunderstanding each other . my point to make data available to specific view namespace , not to specific path(e.g. `some.path.to.template`) or specific view namespace path(e.g. `pluginViewNamespace::some.path.to.template`) . so for the default theme plugin, if it's view namespace is like `defaultthemeplugin`, we can have specific view to available for all of it's  template via `View::composer(defaultthemeplugin::*, fn ($view) => ...)` . Should and does not have any impact to `defaultthemeplugin::some.path.to.specific.tempplate` or just `some.path.to.template` .  this case is applicable for any other and other kind of plugins also . 

> But my concern is that it would cause other troubles. 

Not sure about what kind of trouble you are thinking ? perhaps some example ? 

my implementation case is very simple, extend the core view factory and swap the container bindings . then extending is very minimal, does not change core behaviour but add extra ability to that .

jardakotesovec:
@touhidurabir Example I had in mind was that lets say in ArticleHandler you have:

```
            View::composer('frontend.pages.article', function ($view) {
                $view->with('testComposerData', 'ARTICLE COMPOSER DATA ✅');
            });

```

Than if you add article.blade to the theme plugin it would override the view name to the `defaultManuscript::frontend.pages.article`, right? So that would effectively changed the view name, and the composer statement would not work anymore.

But when checking your last changes - it seems that its not suffering from that - from quick look it seems that laravel somehow keep the original name as well.. so it can be referenced both by `frontend.pages.article` and `defaultManuscript::frontend.pages.article`

touhidurabir:
> @touhidurabir Example I had in mind was that lets say in ArticleHandler you have:
> 
> ```
>             View::composer('frontend.pages.article', function ($view) {
>                 $view->with('testComposerData', 'ARTICLE COMPOSER DATA ✅');
>             });
> ```

ahh , I see. no it will and should continue to work that way . that is one added ability added to extended view factory that I added and swapped in the container .

touhidurabir:
@jardakotesovec on thing I forgot to point is that we need to register the previously used hook `TemplateResource::getFilename` as the unsupported hook as `Hook::addUnsupportedHooks` in the `PKPApplication` .

jardakotesovec:
@touhidurabir Great, thanks for review! On it (tests, and hook update)

Issue #12151, opened 2025-12-17
ojs 3.5 release does not bundle altcha.js


bdcaf:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Hi, I noticed the release for 3.5 on the download page https://pkp.sfu.ca/software/ojs/download/ - currently https://pkp.sfu.ca/ojs/download/ojs-3.5.0-2.tar.gz does not include the altcha.js file (or alternatively instructions do not mention how/which to obtain).

I suppose it is related with altcha used from the path `/node_modules/altcha/dist/altcha.js` (which appears to be cleaned from the build along package.json)

### Steps to Reproduce

1. Install from release bundle as described in doc
2. configure to use altcha
3. try to create account

### Expected Result

altcha should be present.

### Actual Result

registration fails (and in network inspector see altcha.js fails to load)

### Application Version

OJS 3.5

asmecher:
@Godoy0722, if you have time, could you take a look at this?

Godoy0722:
Hello @asmecher ! I hope you're doing well. I made some tests regarding this issue and for the release to work, I added the minified javascript for the altcha, so it now renders on all configured pages. Would you mind having a look on the PR I created? Thanks a lot in advance!

asmecher:
For posterity's sake -- I've merged https://github.com/pkp/pkp-lib/pull/12159 to `stable-3_5_0`, which should get Altcha up to date and working in `stable-3_5_0` for 3.5.0-3, but @Godoy0722 is working on a different approach that doesn't require committing 3rd party code to our git repo. That'll get merged for 3.5.0-4 (and forward-ported to `main`).

Godoy0722:
Hello @asmecher . I just created PRs to handle with this issue in our new approach. In a nutshell, now all systems will work by copying the file from node_modules when installing and then lib-pkp will use this file to make ALTCHA work Independently of third-party code.

You can find the PRs for:

## stable-3_5_0

 - [OJS](https://github.com/pkp/ojs/pull/5244)
 - [OMP](https://github.com/pkp/omp/pull/2199)
 - [OPS](https://github.com/pkp/ops/pull/1166)
 - [pkp-lib](https://github.com/pkp/pkp-lib/pull/12167)

## main

 - [pkp-lib](https://github.com/pkp/pkp-lib/pull/12168)
 - [OJS](https://github.com/pkp/ojs/pull/5245)
[...]

PR #12158, opened 2025-12-18
Update Altcha Integration

Godoy0722:
Change script path to minified version and add altcha.min.js file. Related to issue #12151


asmecher:
Mooted by https://github.com/pkp/pkp-lib/pull/12159

Issue #12074, opened 2025-11-27
Security | Adopt Laravel encryption tools for internal credential storage


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

While user passwords are stored securely hashed, various parts of PKP software store other credentials in the database plaintext. This should be improved via the Laravel Crypt facade.

The Swordv3 plugin provides an example for how this can be done (see [the settings handler](https://github.com/NateWr/swordv3/blob/fe39d297056ec1e7299525ee4a20dffae213f7cb/classes/SettingsHandler.php#L138..L148) and [service code](https://github.com/NateWr/swordv3/blob/fe39d297056ec1e7299525ee4a20dffae213f7cb/Swordv3Plugin.php#L213..L224)). (Most use cases won't need this distinction between API and basic auth modes.)

Review the code for credentials (API keys, usernames, passwords) and modify them to use the Crypt toolset instead. (This would only apply to credentials stored in the database; the configuration file can continue to use plaintext.)

There are two migration strategies that I can suggest considering:
1. A migration to apply the encryption to credentials stored in the database during the upgrade to 3.6. This would require us to build it into the upgrade process for the core code plus any affected plugins (CrossRef, ORCiD, etc).
2. A check-and-migrate process that is applied e.g. when the service is used. This would require a heuristic or flag to check whether a stored password is plaintext or hashed, and some code to convert it to hashed when it's found to be in plaintext. I'm less enthusiastic about this approach, because it would require us to maintain this code long-term.

### Steps to Reproduce

(n/a)

### Expected Result

(n/a)

### Actual Result

(n/a)

### Application Version

OJS, OMP, OPS pre-3.6.0

#### Credentials require storage(database) Encryption/Decryption
 - [x] ORCID Client Secret
 - [x] CrossRef Password
 - [x] Datacite password (both prod and test)

### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12170
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5248
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2203
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1174
[crossref-ojs](https://github.com/pkp/crossref-ojs) --> https://github.com/pkp/crossref-ojs/pull/82
[...]

PR #12170, opened 2025-12-24
pkp/pkp-lib#12074 Use laravel encryption tools for internal credential storage

touhidurabir:
for pkp/pkp-lib#12074 .

### DIFF
--- a/classes/core/EntityDAO.php
+++ b/classes/core/EntityDAO.php
@@ -113,7 +113,7 @@ abstract class EntityDAO
                         value: $row->{$column},
                         type: $schema->properties->{$propName}->type,
                         nullable: true,
-                        encrypt: $schema->properties->{$propName}->encrypt ?? false
+                        decrypt: $schema->properties->{$propName}->encrypt ?? false
                     )
                 );
             }
@@ -131,7 +131,7 @@ abstract class EntityDAO
                         $this->convertFromDB(
                             value: $row->setting_value,
                             type: $schema->properties->{$row->setting_name}->type,
-                            encrypt: $schema->properties->{$row->setting_name}->encrypt ?? false
+                            decrypt: $schema->properties->{$row->setting_name}->encrypt ?? false
                         ),
                         empty($row->locale) ? null : $row->locale
                     );
@@ -291,9 +291,9 @@ abstract class EntityDAO
     /**
      * @copydoc DAO::convertFromDB()
      */
-    protected function convertFromDB($value, string $type, bool $nullable = false, bool $encrypt = false)
+    protected function convertFromDB($value, string $type, bool $nullable = false, bool $decrypt = false)
     {
-        return $this->deprecatedDao->convertFromDB($value, $type, $nullable, $encrypt);
+        return $this->deprecatedDao->convertFromDB($value, $type, $nullable, $decrypt);
     }
 
     /**
--- a/classes/core/PKPAppKey.php
+++ b/classes/core/PKPAppKey.php
@@ -242,7 +242,7 @@ class PKPAppKey
      * Refresh the Laravel container's config and encrypter after key change
      *
      * This is critical when generating a new key in CLI tools that may run
-     * migrations afterward in single CLI request lift cycpe. Without this, the
+     * migrations afterward in single CLI request life cycle. Without this, the
      * Encrypter singleton retains the old key and encrypts data incorrectly.
      */
     public static function refreshEncrypter(): void
--- a/classes/db/DAO.php
+++ b/classes/db/DAO.php
@@ -243,9 +243,9 @@ class DAO
      * @param $value Value from the database
      * @param $type Type from the database, eg `string`
      * @param $nullable True iff the value is allowed to be null
-     * @param $encrypt True if the value is in encrypted format need decryption at pulling from DB
+     * @param $decrypt True if the value is in encrypted format need decryption at pulling from DB
      */
-    public function convertFromDB(mixed $value, ?string $type, bool $nullable = false, bool $encrypt = false): mixed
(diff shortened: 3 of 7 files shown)


classes/core/EntityDAO.php
```diff
@@ -272,17 +291,17 @@ protected function getPrimaryDbProps(DataObject $object): array
     /**
      * @copydoc DAO::convertFromDB()
      */
-    protected function convertFromDB($value, string $type, bool $nullable = false)
+    protected function convertFromDB($value, string $type, bool $nullable = false, bool $encrypt = false)
```
asmecher: Should be `decrypt`, not `encrypt`.

classes/core/PKPAppKey.php
```diff
@@ -234,4 +237,26 @@ public static function parseKey(string $key): string
 
         return $key;
     }
+
+    /**
+     * Refresh the Laravel container's config and encrypter after key change
+     *
+     * This is critical when generating a new key in CLI tools that may run
+     * migrations afterward in single CLI request lift cycpe. Without this, the
```
asmecher:
```suggestion
     * migrations afterward in single CLI request life cycle. Without this, the
```

classes/db/DAO.php
```diff
@@ -243,8 +243,9 @@ public function dateFromDB(?string $d): ?string
      * @param $value Value from the database
      * @param $type Type from the database, eg `string`
      * @param $nullable True iff the value is allowed to be null
+     * @param $encrypt True if the value is in encrypted format need decryption at pulling from DB
      */
-    public function convertFromDB(mixed $value, ?string $type, bool $nullable = false): mixed
+    public function convertFromDB(mixed $value, ?string $type, bool $nullable = false, bool $encrypt = false): mixed
```
asmecher: Should be `decrypt` in this function, not `encrypt`.

classes/migration/upgrade/v3_6_0/I12074_EncryptCredentials.php
```diff
@@ -0,0 +1,217 @@
...
+     *
+     * @throws DowngradeNotSupportedException
+     */
+    public function down(): void
+    {
+        try {
+            $this->decryptContextSettings();
+            $this->decryptPluginSettings();
+        } catch (DecryptException $e) {
+            throw new DowngradeNotSupportedException($e->getMessage());
```
asmecher: This should probably be a regular Exception, not a DowngradeNotSupportedException. Downgrades are supported here; an exception would indicate a problem with the decryption.

classes/migration/upgrade/v3_6_0/I12074_EncryptCredentials.php
```diff
@@ -0,0 +1,217 @@
...
+            ->where('setting_name', $settingName)
+            ->whereNotNull('setting_value')
+            ->where('setting_value', '<>', '')
+            ->get([$primaryKey, 'setting_value']);
+
+        foreach ($rows as $row) {
+            // Skip if already encrypted (starts with eyJ which is base64 for {"iv":)
+            // if (str_starts_with($row->setting_value, 'eyJ')) {
+            //     continue;
+            // }
```
asmecher: Remove dead code

classes/migration/upgrade/v3_6_0/I12074_EncryptCredentials.php
```diff
@@ -0,0 +1,217 @@
...
+            ->whereIn('setting_name', $settingNames)
+            ->whereNotNull('setting_value')
+            ->where('setting_value', '<>', '')
+            ->get(['context_id', 'setting_name', 'setting_value']);
+
+        foreach ($rows as $row) {
+            // Skip if already encrypted (starts with eyJ which is base64 for {"iv":)
+            // if (str_starts_with($row->setting_value, 'eyJ')) {
+            //     continue;
+            // }
```
asmecher: Remove dead code

classes/migration/upgrade/v3_6_0/I12074_EncryptCredentials.php
```diff
@@ -0,0 +1,217 @@
...
+            ->where('setting_name', $settingName)
+            ->whereNotNull('setting_value')
+            ->where('setting_value', '<>', '')
+            ->get([$primaryKey, 'setting_value']);
+
+        foreach ($rows as $row) {
+            // Skip if not encrypted (doesn't start with eyJ)
+            if (!str_starts_with($row->setting_value, 'eyJ')) {
+                continue;
+            }
```
asmecher: I'm not sure we want to do this. What if the setting is a key that's already encrypted by the 3rd party service using the same encryption? Or a password that actually starts with `eyJ`? I'd rather we just treated the migrations as atomic and didn't try to second-guess them. (Also below.)

classes/plugins/Plugin.php
```diff
@@ -591,6 +591,15 @@ public function addLocaleData(): void
         }
     }
 
+    /**
+     * Contains and return the list of plugin setting name which should be encrypted/decrypted
+     * at the time to DB store/retrive.
+     */
+    public function getEncryptedSettingFieldNames(): array
```
asmecher: I'd prefer `getEncryptedSettingNames` rather than `...SettingFieldNames`.

asmecher (changes requested):
Some minor changes to suggest, thanks!

classes/migration/upgrade/v3_6_0/I12074_EncryptCredentials.php
touhidurabir: make sense , in that case we have no need for the `try...catch` block and let the exception bubble up if it arise .

Issue #12162, opened 2025-12-18
Improve password strength requirements


asmecher:
**Description:**
Password strength requirements in OJS/OMP/OPS are extremely low. Currently you can set a minimum length, and further requirements are set only by external tooling (through plugins like "[better password](https://github.com/ulsdevteam/pkp-betterPassword)").

There is a general discussion of related issues at https://github.com/pkp/pkp-lib/discussions/11084.

For OSS ORE (OJS/OMP/OPS 3.6), we are required to comply with EC's [IT technical standard on passwords](https://www.scribd.com/document/733071341/Password-technical-standard).  This sets out numerous requirements with two classifications: "traditional" (representing traditional common practices) and "user centric" (representing newer best practices). There is significant overlap, and we must support the required features of at last one approach.

The following are the changes required to meet baseline requirements for "traditional" handling:

- [ ] 1. and 1.a.: Passwords shall be validated against a blacklist of unauthorised passwords based on the Commission’s password blacklist. The blacklist check shall be performed as part of the immediate password validity control, i.e. whenever a password is created or modified.
- #4265 needs to be resolved in order to permit the minimum password length to be customized.

**Additional requirements:**

Section 6.1.3 specifies throttling. For the moment, this can/should be achieved via the use of a general-purpose security tool like [Fail2ban](https://github.com/fail2ban/fail2ban). (Laravel does provide tools for this, but those cannot be easily integrated into OJS until middleware options are better integrated.)

Requirements 8. and 9. in section 6.1.2 specify minimum lengths of 8 or 10 characters, depending on whether 2FA is being used. This can be achieved by configuration (once #4265 is resolved).

**Implementation:**

The [Laravel password validation rule object](https://laravel.com/docs/12.x/validation#validating-passwords) should be used if possible in order to bring OJS/OMP/OPS closer to Laravel for password management, and to make additional rules (for cases where a blacklist isn't desired or enough) easy to implement in a standard way.

### PRs (WIP)
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12174
[ui-library](https://github.com/pkp/ui-library) --> https://github.com/pkp/ui-library/pull/768
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5251
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1195 [TEST ONLY]
[...]

PR #12174, opened 2026-01-01
pkp/pkp-lib#12162 better password validation rule

touhidurabir:
for pkp/pkp-lib#12162

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -13,7 +13,7 @@ import 'cypress-wait-until';
 import 'cypress-iframe'
 
 /**
- * Get a standardized password for a test user
+ * Generate and return a standardized password for a test user
  * @param {string} username - The username
  * @returns {string} The password
  */
@@ -27,7 +27,7 @@ function getPassword(username) {
 		return username;
 	}
 
-	return username + '1!#';
+	return username + username;
 }
 
 // Export for use in commands and tests
--- a/classes/components/forms/site/PKPSiteSecurityForm.php
+++ b/classes/components/forms/site/PKPSiteSecurityForm.php
@@ -67,6 +67,19 @@ class PKPSiteSecurityForm extends FormComponent
                 'size' => 'small',
                 'value' => $site->getData('minPasswordLength'),
                 'groupId' => static::PASSWORD_POLICY_GROUP,
+            ]))
+            ->addField(new FieldOptions('passwordUncompromisedEnabled', [
+                'label' => __('admin.settings.security.passwordUncompromised'),
+                'description' => __('admin.settings.security.passwordUncompromised.description'),
+                'type' => 'checkbox',
+                'options' => [
+                    [
+                        'value' => true,
+                        'label' => __('admin.settings.security.passwordUncompromised.enable.label'),
+                    ],
+                ],
+                'value' => $site->getData('passwordUncompromisedEnabled') ?? false,
+                'groupId' => static::PASSWORD_POLICY_GROUP,
             ]));
 
         return $this;
--- a/classes/core/ValidationServiceProvider.php
+++ b/classes/core/ValidationServiceProvider.php
@@ -14,6 +14,7 @@
 
 namespace PKP\core;
 
+use APP\core\Application;
 use Illuminate\Contracts\Validation\UncompromisedVerifier;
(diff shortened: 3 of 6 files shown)


classes/security/LocalPasswordBlacklistVerifier.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $filePath = $this->getBlacklistFilePath();
+
+        if (!file_exists($filePath)) {
+            return [];
+        }
+
+        $cacheKey = $this->getCacheKey($filePath);
+
+        return Cache::remember($cacheKey, static::CACHE_LIFETIME_SECONDS, function () use ($filePath) {
```
asmecher: I don't think this caching code is worth the complexity. You could just implement `getBlacklist` with `array_filter` wrapped around `file()` and it's a pretty efficient operation. It's not a high-frequency operation, so I think we can just keep it simple.

classes/security/RateLimitingService.php
```diff
@@ -0,0 +1,265 @@
...
+
+    /**
+     * Check if login attempts are rate limited for the given username and IP.
+     *
+     * @param string $username The username attempting to log in
+     * @param string $ip The IP address of the request
+     *
+     * @return bool True if rate limited, false if attempts are allowed
+     */
+    public function isLoginLimited(string $username, string $ip): bool
```
asmecher, line 81: Will the IP handling work with IPv6 addresses?

classes/security/LocalPasswordBlacklistVerifier.php
touhidurabir: removed caching

classes/security/RateLimitingService.php
touhidurabir: it was but not fully but now added extra normalisation check to only consider the /64 prefix (e.g. first 8 bytes) so all addresses in the same /64 network share the same rate limit counter.

cypress/support/commands.js
```diff
@@ -12,6 +12,27 @@ import 'cypress-file-upload';
...
+	if (username === undefined || username === null) {
+		throw new Error('must provide a username');
+	}
+	
+	// admin keeps original password
+	if (username === 'admin') {
+		return username;
+	}
+
+	return username + '1!#';
```
touhidurabir: Due to default usage of `uncompromise` rule, we get hit by the leaked password for our usual password convention (e.g. usernameusername). I can think of following options 
1. have a setting option in the security tab to have it controlled and have it unset by default 
2. have a evn for cypress test and have it checked where we bind the singleton to the container
3. update all the password to a new convention which the above code does .

classes/security/RateLimitingService.php
```diff
@@ -0,0 +1,347 @@
...
+     * @param string $ip The IP address
+     *
+     * @return string The rate limit key
+     */
+    protected function getPasswordResetRateLimitKey(string $ip): string
+    {
+        $normalizedIp = $this->normalizeIp($ip);
+
+        return 'password_reset:' . $normalizedIp;
+    }
```
touhidurabir: The password reset rate liming is using only the IP unlike the login (which use IP+username) and that is standard/common practice. but I am wondering if that is going to cause issue given the institutional nature of OJS specifically as in that case, many user are using shared IP or behind NAT. that can cause following scenario as 
```
With IP-only (current):                                                                                                                                       
                                                                                                                                                               
 10:00:00 - User A requests reset → attempt 1/5                                                                                                                
 10:01:00 - User B requests reset → attempt 2/5                                                                                                                
 10:02:00 - User C requests reset → attempt 3/5                                                                                                                
 10:03:00 - User D requests reset → attempt 4/5                                                                                                                
 10:04:00 - User E requests reset → attempt 5/5 → ALL USERS BLOCKED                                                                                            
 10:04:30 - User F needs reset → BLOCKED (must wait until 10:05:00) 
```

we can choose an alternative(but a bit week) approach of IP+email . However at the same time, the chances of above scenario to happened for a short span of time very low . but as the time is configurable, this can happen for a long time span like 24 hours .

touhidurabir: After reading though the laravel's rate limiting mechanism for password reset, updated it using IP+Email instead of only using IP .

Issue #12175, opened 2026-01-01
Allow file uploads to tasks discussions through the API


Vitaliy-1:
**Description:**
The functionality should include:
1. Upload a new file
2. Attach existing submission file

**Steps to test the changes:**
To upload a new file:
1. Make an API request to upload a temporary file. E.g., on a local instance http://localhost:8000/publicknowledge/api/v1/temporaryFiles
2. Include a new parameter `temporaryFileIds` to the post request when creating a new task or discussion, e.g. http://localhost:8000/publicknowledge/api/v1/submissions/21/tasks
```json
{
    "type": 1,
    "title": "This is a task",
    "assocId": 21,
    "stageId": 1,
    "status": 1,
    "createdBy": 1,
    "dateStarted": "2025-07-01",
    "dateClosed": null,
    "temporaryFileIds": [46,47],
    "participants": [
        {
            "userId": 3
        },
        {
            "userId": 2
        },
        {
            "userId": 1
        }
    ]
}
```
To upload existing submission file, pass the `submissionFileIds` parameter to the post request when creating a new task or discussion

**Solution:** 
Use API for uploading a temporary file, which then can be recorded to the `files` and `submission_files` tables upon saving the task. That should be available for all roles.
Only managers as well as editors and assistants can attach an existing submission file to the task or discussion.

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/12176
tests: https://github.com/pkp/ojs/pull/5252

Vitaliy-1:
@blesildaramirez, can you take a look at the shape of the API data (request/response)? The main question is whether this is enough for front-end? Don't worry about the backend part.

Vitaliy-1:
@ewhanson, do you have time to look at the PR?

ewhanson:
Yes, I'll have a look

PR #12176, opened 2026-01-01
pkp/pkp-lib#12175 Allow files upload to tasks and discussions

Vitaliy-1:


classes/note/SaveNoteWithFiles.php
```diff
@@ -0,0 +1,216 @@
...
+
+        return $attributes;
+    }
+
+    /**
+     * Handle temporary files associated with the task.
+     *
+     * @return array<int> IDs of created submission files
+     */
+    protected function saveTemporaryFiles(Note $note = null): array
```
ewhanson: Implicitly marking argument as nullable is deprecated. Should probably make the type `?Note`.

Vitaliy-1: Thanks!

Issue #12177, opened 2026-01-05
Refactor Navigation Menu Drag-and-Drop UI


blesildaramirez:
**Description:**
The navigation menu editor currently uses jQuery and supports dragging items between **Assigned** and **Unassigned** menu items lists, with nesting and a visual drop preview.

This task is to refactor that UI to a **Vue-based draggable tree**, replacing the existing implementation without changing how it works for users.

**Requirements:**
- [ ] Use Vue 3 (Composition API)
- [ ] Replace jQuery with:
  - [ ] Reka UI – Draggable Sortable Tree
  - [ ] @atlaskit/pragmatic-drag-and-drop
  - [ ] Vue version behaves the same as the current UI
- [ ] Support up to 3 levels of nesting for Assigned items
- [ ] Keep Unassigned items flat
- [ ] Moving a parent item to Unassigned should flatten all children
- [ ] Moving a multi-level parent item should not be allowed if the targeted item tree would exceed 3 levels
- [ ] Keep a clear drag preview (drop position / nesting intent)
- [ ] Hook up the UI to the existing navigation menu APIs
  - [ ] Menu data loads and saves via the API
- [ ] Storybook examples are added

**What application are you using?**
  OJS, OMP or OPS version main

**Additional information**
<img width="1890" height="894" alt="Image" src="https://github.com/user-attachments/assets/92b65713-f55b-46c2-a458-b76ab7da501c" />

Tribunal33:
Hi @blesildaramirez this looks great.  I did noticed few issues.  I will mark them with a ❌ 
3 - When you try to drag and drop a nested group onto another nested group it just flattens the dropped group.  With this new impelmentation it has a red 🛑 icon and will not allow.  This seems acceptable to me
10 - Does not give the success message but displays checkmark.  Not sure if it is bubbling up messaging to other window.
[Missing this]
<img width="510" height="129" alt="Image" src="https://github.com/user-attachments/assets/ac840e63-fa16-40b0-97ca-bb3f53b7cb75" />

Is this good enough?

<img width="424" height="126" alt="Image" src="https://github.com/user-attachments/assets/f50fb0f8-6796-4073-a5e2-bba87011be8e" />

11 - This should not be allow on older version it gives a warning and blocks this from happening but new menu allows this.  You should not be allowed to have two 'User' navigation locations
[Missing this and shouldn't allow]

[...]

PR #12178, opened 2026-01-05
pkp/pkp-lib#12177 Add apis to integrate the new navigation menu component

blesildaramirez:


blesildaramirez:
Hi @ewhanson , may I ask for a code review for this PR? The APIs cover adding, updating, and retrieving the list of existing navigation menu data. These APIs are needed to refactor the UI form with drag and drop to use a Vue component.

Let me know if you have any questions. Many thanks!

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+
+        // Get item types for reference
+        $navigationMenuService = app(PKPNavigationMenuService::class);
+        $itemTypes = $navigationMenuService->getMenuItemTypes();
+
+        return response()->json([
+            'assigned' => [],
+            'unassigned' => $unassignedItems,
+            'itemTypes' => $itemTypes,
+            'maxDepth' => 3,
```
ewhanson: Is the max depth a fixed constant that can never change? Or is it configurable? Either way, I would maybe make the default a constant as part of this class so it's more explicit. That way it's not a magic number that's reused in several places, and if it ever changes, it can be changed in only one place.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+     */
+    public function getAreas(Request $illuminateRequest): JsonResponse
+    {
+        $request = $this->getRequest();
+        $context = $request->getContext();
+
+        $areas = [];
+
+        if ($context) {
+            $themePlugins = \PKP\plugins\PluginRegistry::loadCategory('themes', true);
```
ewhanson: Very minor and feel free to ignore, but I would use a `use` import statement above for the namespace instead for consistency.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+            $activeThemeNavigationAreas = [];
+
+            foreach ($themePlugins as $themePlugin) {
+                if ($themePlugin->isActive()) {
+                    $themeAreas = $themePlugin->getMenuAreas();
+                    foreach ($themeAreas as $area) {
+                        // Use the area name directly as the label (matches old form behavior)
```
ewhanson: Could you be a little more specific about what old behaviour this matches and why it's necessary?

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+    public function getAreas(Request $illuminateRequest): JsonResponse
+    {
+        $request = $this->getRequest();
+        $context = $request->getContext();
+
+        $areas = [];
+
+        if ($context) {
+            $themePlugins = \PKP\plugins\PluginRegistry::loadCategory('themes', true);
+            $activeThemeNavigationAreas = [];
```
ewhanson: Could the `$areas` array above not be used directly here? Not a huge deal, just seems like a duplication of variables.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    /**
+     * Create a new navigation menu
+     */
+    public function add(Request $illuminateRequest): JsonResponse
+    {
+        $request = $this->getRequest();
+        $context = $request->getContext();
+        $contextId = $context?->getId() ?? \PKP\core\PKPApplication::SITE_CONTEXT_ID;
```
ewhanson: Again, I would use a `use` statement for the namespace, especially since this class is referenced throughout.

classes/components/listPanels/NavigationMenuEditorListPanel.php
```diff
@@ -0,0 +1,43 @@
...
+class NavigationMenuEditorListPanel extends ListPanel
+{
+    /** @var string URL to the API endpoint for navigation menus */
+    public string $apiUrl = '';
+
+    /** @var int|null Initial menu ID to load */
+    public ?int $initialMenuId = null;
+
+    /** @var int Maximum nesting depth */
+    public int $maxDepth = 3;
```
ewhanson: This `3` could refer to the same default constant I mentioned in `PKPNavigationMenuController`

classes/components/listPanels/NavigationMenuEditorListPanel.php
```diff
@@ -0,0 +1,43 @@
...
+ * @class NavigationMenuEditorListPanel
+ *
+ * @ingroup classes_components_list
+ *
+ * @brief A ListPanel component for editing navigation menus with drag and drop
+ */
+
+namespace PKP\components\listPanels;
+
+class NavigationMenuEditorListPanel extends ListPanel
```
ewhanson: This all looks okay on the PHP side, but just wanted to check on using a ListPanel for this since I know we've been moving away from using the ListPanel suite of components in Vue. Just wanted to double-check.

controllers/grid/navigationMenus/NavigationMenusGridHandler.php
```diff
@@ -44,8 +44,6 @@ public function __construct()
             Role::ROLE_ID_MANAGER,
             $ops = [
                 'fetchGrid', 'fetchRow',
```
ewhanson: For future editing/diffing, it would probably be easier if each of these array items were on their own line.

locale/en/manager.po
```diff
@@ -2562,6 +2562,18 @@ msgstr "Assigned Menu Items"
 msgid "manager.navigationMenus.unassignedMenuItems"
 msgstr "Unassigned Menu Items"
 
+msgid "manager.navigationMenu.noAssignedItems"
+msgstr "No items assigned to this menu. Drag items from the right panel."
```
ewhanson: Will the panel always be on the right? Or does the panel have a name it could be referred to? Maybe it's fine to have this localized since RTL languages would also have a different localized string of where the user should drag items from.

pages/management/ManagementHandler.php
```diff
@@ -804,6 +807,30 @@ protected function getHighlightsListPanel(): HighlightsListPanel
...
+            $context->getPath(),
+            'navigationMenus'
+        );
+
+        return new NavigationMenuEditorListPanel(
+            'navigationMenuEditor',
+            __('manager.navigationMenus'),
+            [
+                'apiUrl' => $apiUrl,
+                'maxDepth' => 3,
```
ewhanson: Another `maxDepth` magic number.

schemas/navigationMenu.json
```diff
@@ -0,0 +1,59 @@
...
+			"validation": [
+				"nullable"
+			]
+		},
+		"menuTree": {
+			"type": "object",
+			"description": "The menu tree structure for assigned items. Format: menuTree[menuItemId] = {seq: number, parentId: number|null}.",
+			"writeOnly": true,
+			"apiSummary": false,
+			"additionalProperties": {
```
ewhanson: I'm not sure where the `"additionalProperties"` key came from, but see the [announcement schema file](https://github.com/pkp/pkp-lib/blob/e297c24714744d078258910074ca3d27609afe09/schemas/announcement.json#L79-L105) for an example of an object within a schema file.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+
+        /** @var NavigationMenuItemDAO $menuItemDao */
+        $menuItemDao = DAORegistry::getDAO('NavigationMenuItemDAO');
+
+        // Get all items for this context
+        $allItems = $menuItemDao->getByContextId($contextId);
+        $unassignedItems = [];
+
+        while ($item = $allItems->next()) {
```
ewhanson: `DAOResultFactory`s include a `toArray()` method you could use instead of manually writing the while loop in these cases throughout.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+     *
+     * This is the single source of truth for menu item serialization.
+     * Used by getAllItems, getItems (assigned), and getItems (unassigned).
+     *
+     * @param NavigationMenuItem $menuItem The menu item
+     * @param NavigationMenuItemAssignment|null $assignment The assignment (null for unassigned items)
+     * @param array $children Pre-built children array for assigned items
+     * @return array Mapped item data
+     */
+    protected function mapMenuItem(
```
ewhanson: This and any associated methods should be part of a resource rather than included in the scope of the API controller itself.

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,668 @@
...
+            'unassigned' => $unassignedItems,
+            'itemTypes' => $itemTypes,
+            'maxDepth' => 3,
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Get assigned items as a hierarchical tree
+     */
+    protected function getAssignedItemsTree(int $navigationMenuId, NavigationMenu $navigationMenu): array
```
ewhanson: This and any other methods about organizing the NavigationMenu data rather than actually handling the API calls should be in a different class. 

A lot of the methods that aren't directly related to the API calls in here seem like they would fit with our exising `Repository` pattern, though I know this doesn't use the newer (but not new) `EntityDAO` pattern. I'll leave it up to you how you'd like to organize the methods based on the needs you see, but it should either be part of the resource if it's resource specific for presentation or in a more general data manipulation class. 

Ideally, the only methods in the API controller are directly related to handling and processing API calls, not actually preparing and manipulating the data. This is something that hasn't been perfect throughout the codebase to date, but with anything new, this is an approach I'd like to more strictly enforce.

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: Good catch - the comment was a bit vague, so I removed it. Themes provide menu area names (e.g. primary, user) via `getMenuAreas()`, which returns an array of strings. Since themes don't provide separate translated labels for menu areas, we use the area name as both the key and the display label.

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: Good catch! Yeah I agree that this should be put as a constant for this class, I updated this now. It's a fixed value atm but could change in the future. On current implementation, the max level for menu items is 2, but here on the refactored UI we want up to 3 levels. So that might change again in the future.

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: This has been fixed, thank you!

schemas/navigationMenu.json
blesildaramirez: Good point, I updated this to match the schema structure from announment.json file. Thanks!

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: This has been fixed, thanks!

classes/components/listPanels/NavigationMenuEditorListPanel.php
blesildaramirez: You're right, this file has been removed and the configuration has been moved to `ManagementHandler` instead.

controllers/grid/navigationMenus/NavigationMenusGridHandler.php
blesildaramirez: Good catch, I updated this, `fetchRow` now is in its own line.

locale/en/manager.po
blesildaramirez: I updated this to use "No items assigned to this menu. Drag items from Unassigned Menu Items."

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: Thanks a lot for the detailed explanation. I have moved the other methods to the resource file and service depending on their use case.

blesildaramirez:
Hi @ewhanson , I updated this PR based on your feedback. This is now ready for your review. Thanks a lot!

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,291 @@
...
+     * @copydoc \PKP\core\PKPBaseController::getGroupRoutes()
+     */
+    public function getGroupRoutes(): void
+    {
+        // Get all available menu items (for new menu creation)
+        Route::get('items', $this->getAllItems(...))
+            ->name('navigationMenu.getAllItems');
+
+        // Get navigation menu areas for the active theme
+        Route::get('areas', $this->getAreas(...))
```
jardakotesovec, line 72: I see that areas endpoint is returning: 
```
{"areas":{"primary":"primary","user":"user"}}
```

@ewhanson  It looks like unusual data shape to me for our API. I am not sure actually about which convention to follow in this case. Just array of areas for not-paginated endpoint?

api/v1/navigationMenus/PKPNavigationMenuController.php
ewhanson: That sounds good, especially in the context of some of the discussions we've been having about what the max depth should be in general vs what's required for ORE so this make sense.

api/v1/navigationMenus/PKPNavigationMenuController.php
ewhanson: @jardakotesovec, that makes sense to me. In this case (in PHP) would that just an associated array, in this case something like `['primary' => 'primary', 'user' => 'user']`?

api/v1/navigationMenus/PKPNavigationMenuController.php
```diff
@@ -0,0 +1,291 @@
...
+
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Facades\Route;
+use PKP\core\PKPApplication;
+use PKP\core\PKPBaseController;
+use PKP\core\PKPRequest;
+use PKP\db\DAORegistry;
+use PKP\navigationMenu\NavigationMenu;
```
ewhanson: No longer used.

classes/navigationMenu/resources/NavigationMenuItemResource.php
```diff
@@ -0,0 +1,193 @@
...
+    public function toArray($request): array
+    {
+        /** @var NavigationMenuItem $menuItem */
+        $menuItem = $this->resource;
+        $assignment = $this->assignment;
+
+        $conditionalInfo = $this->getItemConditionalInfo($menuItem);
+
+        // For assigned items, use assignment ID; for unassigned, use menu item ID
+        $id = $assignment ? $assignment->getId() : $menuItem->getId();
```
ewhanson: Does this ambiguity in which ID is used cause any issues where this data is consumed? Does it matter? Is there any possibility of an overlap in IDs in places this is shown or is that not a possibility?

classes/navigationMenu/resources/NavigationMenuItemResource.php
```diff
@@ -0,0 +1,193 @@
...
+    protected function getItemTitle(NavigationMenuItemAssignment $assignment, NavigationMenuItem $menuItem): string
+    {
+        $navigationMenuService = app(PKPNavigationMenuService::class);
+        $navigationMenuService->setAllNMILocalizedTitles($menuItem);
+
+        // Try assignment-specific title first (custom override)
+        $assignmentTitles = $assignment->getTitle(null);
+        if (is_array($assignmentTitles) && !empty($assignmentTitles)) {
+            $locale = Locale::getLocale();
+            if (isset($assignmentTitles[$locale]) && !empty($assignmentTitles[$locale])) {
```
ewhanson: Do you need both `isset()` and `!empty()` to get the information you need for this?

classes/navigationMenu/resources/NavigationMenuItemResource.php
```diff
@@ -0,0 +1,193 @@
...
+
+        // Final fallback to type-based title
+        $itemTypes = $navigationMenuService->getMenuItemTypes();
+        $type = $menuItem->getType();
+
+        if (isset($itemTypes[$type]['title'])) {
+            return $itemTypes[$type]['title'];
+        }
+
+        return $type ?? '';
```
ewhanson: This should always be a string since `$menuItem->getType()` can only return a string, but the underlying `getData()` call could return null. Is the menu type always set, and therefore not a problem or should the return type fo `$menuItem->getType()` change?

classes/navigationMenu/resources/NavigationMenuItemResource.php
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        $itemTypes = $navigationMenuService->getMenuItemTypes();
+        $type = $menuItem->getType();
+
+        if (isset($itemTypes[$type]['title'])) {
+            return $itemTypes[$type]['title'];
+        }
+
+        return $type ?? '';
```
ewhanson: Same goes for this instance as noted above.

classes/navigationMenu/resources/NavigationMenuItemResource.php
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Transform template variables in title
+     */
+    protected function transformTitleVariables(string $title): string
+    {
+        if (strpos($title, '{$loggedInUsername}') !== false) {
```
ewhanson: This works fine, but may  be better practice to use the more modern `str_contains()` to make this check.

api/v1/navigationMenus/PKPNavigationMenuController.php
blesildaramirez: Thanks both for the feedback. I updated this to return `{"areas":["primary","user"]}` since this is how the theme plugins return the data.

classes/navigationMenu/resources/NavigationMenuItemResource.php
blesildaramirez: Good catch - you're right. Using the assignment ID for assigned items and the menu item ID for unassigned items could overlap since both tables use auto-incrementing IDs. However, this doesn't cause issues because assigned and unassigned items are not mixed within the same list.

I still updated this to always use `menuItemId` as the id field, which is clearer since every item has a menu item ID.

classes/navigationMenu/resources/NavigationMenuItemResource.php
blesildaramirez: You're right, the isset() is redundant. I removed it. Thanks!

classes/navigationMenu/resources/NavigationMenuItemResource.php
blesildaramirez: Good point. Updated `NavigationMenuItem::getType()` to return `$this->getData('type') ?? ''` so the return type matches the declared string.

classes/navigationMenu/resources/NavigationMenuItemResource.php
blesildaramirez: Thank you, I updated this.

Issue #11993, opened 2025-11-03
Automatically create task from template when submission reaches specific stage


Hafsa-Naeem:
**Description:**
Automatically create an editorial task from a template when a submission enters a certain stage (e.g. Review, Copyediting, Production, etc.).

**What application are you using?**
OJS, OMP or OPS main branch

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib (main): [pkp/pkp-lib#11993 Automatically create task from template when submission reaches specific stage #12186
](https://github.com/pkp/pkp-lib/pull/12186)
OJS (main): [pkp/pkp-lib#11993 Submodule update ##Hafsa-Naeem/ i11993-main-fix #5262](https://github.com/pkp/ojs/pull/5262)

Vitaliy-1:
Thanks @Hafsa-Naeem, looks good, I've added couple of comments. I see there are changes regarding authorisation. Could you take a look where policies fail? I've added comments to the PR regarding that

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib (main): [pkp/pkp-lib#11993 Automatically create task from template when submission reaches specific stage #12186
](https://github.com/pkp/pkp-lib/pull/12186)
OJS (main): [pkp/pkp-lib#11993 Submodule update ##Hafsa-Naeem/ i11993-main-fix #5262](https://github.com/pkp/ojs/pull/5262)

Vitaliy-1:
@Hafsa-Naeem, thanks for the update. I've added a review

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib (main): [pkp/pkp-lib#11993 Automatically create task from template when submission reaches specific stage #12186
](https://github.com/pkp/pkp-lib/pull/12186)
OJS (main): [pkp/pkp-lib#11993 Submodule update ##Hafsa-Naeem/ i11993-main-fix #5262](https://github.com/pkp/ojs/pull/5262)

Vitaliy-1:
Thanks, @Hafsa-Naeem. I've added a review

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib (main): [pkp/pkp-lib#11993 Automatically create task from template when submission reaches specific stage #12186
](https://github.com/pkp/pkp-lib/pull/12186)
[...]

PR #12186, opened 2026-01-07
pkp/pkp-lib#11993 Automatically create task from template when submission reaches specific stage

Hafsa-Naeem:
for issue [#11993](https://github.com/pkp/pkp-lib/issues/11993)

### DIFF
--- a/api/v1/submissions/tasks/EditorialTaskController.php
+++ b/api/v1/submissions/tasks/EditorialTaskController.php
@@ -45,6 +45,7 @@ use PKP\security\authorization\QueryWorkflowStageAccessPolicy;
 use PKP\security\authorization\QueryWritePolicy;
 use PKP\security\authorization\SubmissionAccessPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
+use PKP\security\authorization\NoteAccessPolicy;
 use PKP\security\Role;
 use PKP\stageAssignment\StageAssignment;
 use PKP\submission\GenreDAO;
@@ -147,20 +148,29 @@ class EditorialTaskController extends PKPBaseController
         $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
         $this->addPolicy(new SubmissionAccessPolicy($request, $args, $roleAssignments));
 
-        // For operations to retrieve task(s), we just ensure that the user has access to it
-        if ($actionName === 'getTask') {
+        // For operations to retrieve task or work with its notes, ensure that the user has access to the task itself
+        if (in_array($actionName, ['getTask', 'addNote', 'deleteNote'])) {
             $stageId = $request->getUserVar('stageId');
             $this->addPolicy(new QueryAccessPolicy($request, $args, $roleAssignments, !empty($stageId) ? (int)$stageId : null, 'taskId'));
         }
 
-        // To modify a task or its notes, need to check read and write access policies
-        if (in_array($actionName, ['addNote', 'deleteNote', 'editTask', 'deleteTask', 'closeTask', 'openTask', 'startTask'])) {
+        // deleting a note additionally requires write access to the note itself
+        if ($actionName === 'deleteNote') {
+            $this->addPolicy(new NoteAccessPolicy(
+                $request,
+                (int) $illuminateRequest->route('noteId'),
+                NoteAccessPolicy::NOTE_ACCESS_WRITE
+            ));
+        }
+
+        // To modify a task, need to check read and write access policies
+        if (in_array($actionName, ['editTask', 'deleteTask', 'closeTask', 'openTask', 'startTask'])) {
             $stageId = $request->getUserVar('stageId');
             $this->addPolicy(new QueryAccessPolicy($request, $args, $roleAssignments, !empty($stageId) ? (int)$stageId : null, 'taskId'));
             $this->addPolicy(new QueryWritePolicy($request));
         }
 
-        // To create a task or get a list of tasks, check if the user has access to the workflow stage; note that controller must ensure to get a list of tasks where user is a participant
+        // To create a task or get a list of tasks, check if the user has access to the workflow stage; note that controller still ensures that getTasks only returns tasks where user is a participant
         if (in_array($actionName, ['addTask', 'getTasks', 'fromTemplate', 'getParticipants'])) {
             $this->addPolicy(new QueryWorkflowStageAccessPolicy($request, $args, $roleAssignments, (int)$request->getUserVar('stageId')));
         }
--- a/classes/editorialTask/Repository.php
+++ b/classes/editorialTask/Repository.php
@@ -228,13 +228,6 @@ class Repository
 
             $task->save();
 
-            $task->updateSettings(
-                [
-                    'templateId' => $templateId,
-                    'autoCreated' => true,
-                ],
-                (int) $task->id
-            );
         }
     }
 


api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -198,6 +198,10 @@ public function getTask(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_NOT_FOUND);
         }
 
+        if ($response = $this->forbidIfUnassignedAndNotManager($submission, $editTask)) {
```
Vitaliy-1: I'm trying to replicate the problem on the current main branch. I've created a task, assigned participants and trying to access by a section editor unassigned to the submission. I'm receiving:
```
{
    "error": "user.authorization.roleBasedAccessDenied",
    "errorMessage": "The current role does not have access to this operation."
}
```

Vitaliy-1: Can you double check which policies don't apply as should. I'm referencing them here: https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/tasks/EditorialTaskController.php#L155-L164

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -478,13 +494,21 @@ protected function getTaskData(Submission $submission, EditorialTask $editTask):
...
-        $creatorId = $editTask->createdBy;
-        if (!in_array($creatorId, $participantIds)) {
+        $participantIds = $editTask->participants
+            ->map(fn (Participant $participant) => $participant->userId)
+            ->filter()
+            ->unique()
+            ->values()
+            ->all();
+
+        $creatorId = (int) ($editTask->createdBy ?? 0);
```
Vitaliy-1: By default it should be null, e.g., if it's created by the system

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -505,11 +529,41 @@ protected function getTaskData(Submission $submission, EditorialTask $editTask):
...
     public function addNote(AddNote $illuminateRequest): JsonResponse
     {
+        $submission = $this->getAuthorizedContextObject(PKPApplication::ASSOC_TYPE_SUBMISSION); /** @var Submission $submission */
+        $task = EditorialTask::find((int) $illuminateRequest->route('taskId')); /** @var EditorialTask $task */
+
+        if (!$task) {
+            return response()->json(['error' => __('api.404.resourceNotFound')], Response::HTTP_NOT_FOUND);
+        }
+
+        if ($response = $this->forbidIfUnassignedAndNotManager($submission, $task)) {
```
Vitaliy-1: The same as for notes, if it's related to authorization, it should be reflected in policies

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -547,10 +607,10 @@ public function deleteNote(Request $illuminateRequest): JsonResponse
         }
 
         $user = $this->getRequest()->getUser();
-        $submission = $this->getAuthorizedContextObject(PKPApplication::ASSOC_TYPE_SUBMISSION); /** @var Submission $submission */
 
         // Allow removing the note for its creator or a manager/admin
-        if ($note->userId !== $user->getId() || !$user->hasRole([Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER], $submission->getData('contextId'))) {
+        $canDelete = $note->userId === $user->getId() || $user->hasRole([Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER], $submission->getData('contextId'));
```
Vitaliy-1: the same as above

classes/editorialTask/services/AutoCreateTasksFromTemplates.php
```diff
@@ -0,0 +1,79 @@
...
+namespace PKP\editorialTask\services;
+
+use APP\core\Application;
+use APP\submission\Submission;
+use Illuminate\Support\Facades\DB;
+use PKP\core\PKPApplication;
+use PKP\editorialTask\EditorialTask;
+use PKP\editorialTask\Template;
+
+class AutoCreateTasksFromTemplates
```
Vitaliy-1: According to our pattern, it's better to create a Repository with correspondent methods

api/v1/submissions/tasks/EditorialTaskController.php
Vitaliy-1: Can you double check if QueryWritePolicy doesn't work in this case, I'm checking the action here https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/tasks/EditorialTaskController.php#L160? If not, can you make necessary changes in the Policy?

classes/decision/DecisionType.php
```diff
@@ -203,9 +204,10 @@ public function runAdditionalActions(Decision $decision, Submission $submission,
             Repo::submission()->updateStatus($submission, $this->getNewStatus());
         }
 
-        $newStageId = $this->getNewStageId($submission, (int)$decision->getData('reviewRoundId'));
+        $oldStageId = (int) $submission->getData('stageId');
+        $newStageId = $this->getNewStageId($submission, (int) $decision->getData('reviewRoundId'));
 
-        if ($newStageId) {
+        if ($newStageId && $newStageId !== $oldStageId) {
```
Vitaliy-1: Have you encountered cases when this additional check is needed?

classes/editorialTask/Repository.php
```diff
@@ -192,6 +193,63 @@ public function addCommentsForEditorsQuery(Submission $submission): int
...
+                [
+                    'templateId' => $templateId,
+                    'autoCreated' => true,
+                ],
+                (int) $task->id
+            );
+        }
+    }
+
+    private function taskAlreadyCreatedFromTemplate(int $submissionId, int $templateId): bool
```
Vitaliy-1, line 234: In `edit_tasks` table, we have a column `edit_task_template_id`. Can we use it for this purpose?

classes/editorialTask/services/AutoCreateTasksFromTemplates.php
Vitaliy-1, line 1: It looks like that this class should be removed

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -148,13 +148,13 @@ public function authorize(PKPRequest $request, array &$args, array $roleAssignme
...
-        if (in_array($actionName, ['getTask', 'addNote', 'deleteNote'])) {
+        if ($actionName === 'getTask') {
             $stageId = $request->getUserVar('stageId');
             $this->addPolicy(new QueryAccessPolicy($request, $args, $roleAssignments, !empty($stageId) ? (int)$stageId : null, 'taskId'));
         }
 
-        // To modify a task, need to check read and write access policies
-        if (in_array($actionName, ['editTask', 'deleteTask', 'closeTask', 'openTask', 'startTask'])) {
+        // To modify a task or its notes, need to check read and write access policies
+        if (in_array($actionName, ['addNote', 'deleteNote', 'editTask', 'deleteTask', 'closeTask', 'openTask', 'startTask'])) {
```
Vitaliy-1: The `QueryWritePolicy` doesn't match expectations - task participants should have an option to add a note. Can you double check if `NoteAccessPolicy` could work here or create a new Policy with needed permissions?

classes/editorialTask/Repository.php
```diff
@@ -192,6 +193,61 @@ public function addCommentsForEditorsQuery(Submission $submission): int
...
+            $maxSeq = (float) (EditorialTask::query()
+                ->where('assoc_type', PKPApplication::ASSOC_TYPE_SUBMISSION)
+                ->where('assoc_id', $submission->getId())
+                ->max('seq') ?? 0);
+
+            $task->seq = $maxSeq + 1;
+
+            $task->save();
+
+            $task->updateSettings(
```
Vitaliy-1: Do we need this settings?

Issue #12024, opened 2025-11-13
[Eloquent] | Saving arrays to settings tables with Eloquent breaks the values


ajnyga:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

While working on the in-development Data Citations feature, saving array-type settings through an Eloquent model behaves inconsistently between create and update.

On initial creation, Eloquent stores the array correctly as JSON in the setting_value column:

`[{"givenName":"Antti-Jussi","familyName":"Nygård","orcid":"https://orcid.org/0000-0002-1534-109X"}]`

However, updating the same setting results in corrupted data. During the update, Eloquent has already converted the array into a JSON string before PKP’s schema processing runs. As a result, PKPSchemaService::coerce() receives a JSON string instead of an array, falls through to the default string-handling case, and serializes the JSON string. This produces invalid, nested, serialized data such as:

`["s:104:\"[{\"givenName\":\"NEW NAME\",\"familyName\":\"Nyg\\u00e5rd\", ... }]\";"]`

The stored value can no longer be decoded into a proper array, breaking the feature logic.

The development branch for datacitations now includes a fix 
https://github.com/ajnyga/pkp-lib/blob/f6278/classes/services/PKPSchemaService.php#L380-L382

https://github.com/ajnyga/pkp-lib/tree/f6278
https://github.com/ajnyga/ui-library/tree/f6278

ping @touhidurabir 

### Steps to Reproduce

1. Use the in-development Data Citations branches (e.g. ajnyga/pkp-lib f6278 and ajnyga/ui-library f6278). 
2. Undo the included fix in `coerce` method
3. Create a new citation and save author information using Vue form components.
4. Check the settings table: array data is correctly stored as JSON.
5. Edit the author information and save again.
6. Inspect the database: the value is now serialized incorrectly.

### Expected Result

Array-type settings should behave the same on create and update.

When the schema defines a property as array, PKP’s settings update pipeline (sanitize → coerce) should receive a PHP array, not a pre-encoded JSON string.

The final value in the database should remain valid JSON, e.g.:

`[{"givenName":"NEW NAME","familyName":"Nygård","orcid":"https://orcid.org/0000-0002-1534-109X"}]`

### Actual Result

On update, Eloquent’s cast converts the array into a JSON string before schema sanitization.

coerce() does not detect an array and serializes the string instead.

The database ends up with invalid nested serialization:

[...]

PR #12201, opened 2026-01-13
pkp/pkp-lib#12024 Proper handling of JSON data with array type using eloquent schema based entity at update

touhidurabir:
for pkp/pkp-lib#12024 .

### DIFF
--- a/classes/services/PKPSchemaService.php
+++ b/classes/services/PKPSchemaService.php
@@ -374,6 +374,10 @@ class PKPSchemaService
                     foreach ($value as $i => $v) {
                         $newArray[$i] = $this->coerce($v, $schema->items->type, $schema->items);
                     }
+                } elseif (isValidJson($value)) {
+                    // Value is already a JSON-encoded string (e.g., from Eloquent cast)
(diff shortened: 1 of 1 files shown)


classes/services/PKPSchemaService.php
```diff
@@ -374,6 +374,10 @@ public function coerce($value, $type, $schema)
                     foreach ($value as $i => $v) {
                         $newArray[$i] = $this->coerce($v, $schema->items->type, $schema->items);
                     }
+                } elseif (isValidJson($value)) {
```
bozana, line 377: I think this function only exist from PHP 8.3 on, right? And I think Alec said that we are currently supporting 8.2, or has that changed? I can double check...

touhidurabir: is it about the `isValidJson` ? if so, this is a custom function in `functions.php` and already used for the `object` case .

bozana: Because it is so in the composer, I assume we cannot use this function now...

bozana: Oh, now I see it...

Issue #12206, opened 2026-01-13
Disallow creating tasks/discussions with both anonymous reviewers as participants


blesildaramirez:
**Description:**
Currently, it’s possible to create a task or discussion where both participants are anonymous reviewers using the new Tasks/Discussions. This setup should not be allowed, as it can result in an invalid setup and potentially break anonymity expectations.

**Current behavior:**
- The UI allows selecting anonymous reviewers for both participants when creating a task/discussion.
- The task/discussion is created successfully without any validation.

**Expected behavior:**
- The system should prevent creating a task or discussion when both participants are anonymous reviewers.
- A clear validation message should explain why this combination is not allowed (like in old discussions implementation).

Vitaliy-1:
PR pkp-lib: https://github.com/pkp/pkp-lib/pull/12218
tests: https://github.com/pkp/ojs/pull/5277

Vitaliy-1:
Thanks, @blesildaramirez, I've made a PR to address this. Could you check if it works as expected?

PR #12218, opened 2026-01-19
pkp/pkp-lib#12206 Fix reviewer anonymization on task/discussion assig…

Vitaliy-1:
…nment


blesildaramirez:
Hi @Vitaliy-1 , I tested this locally. The validation works now if I assign both double blind reviewers, so our form is not allowing it anymore. However, when I assign a Disclosed Author and Open reviewer, it allowed it. It was not allowed in the previous implementation. Can you have a look? Thanks!

Issue #12197, opened 2026-01-12
Template user groups reset/missing when saving templates across stages (find() vs get() mismatch)


Hafsa-Naeem:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When creating task/discussion templates for multiple stages (Submission, then Review, etc.), the user group selections saved earlier appear empty when revisiting a previous stage. Also, eager-loading userGroups behaves inconsistently: Template::with('userGroups')->find($id) returns groups, but Template::with('userGroups')->get()->where('editTaskTemplateId', $id) loses some.

### Steps to Reproduce
1. Open the Tasks and Discussions.
2. Under Submission Stage, create one or more templates and select at least one User Group (e.g. Journal Editor). Save.
3. Under Review Stage, create a template and select the SAME User Group and save.
4. Go back to Submission stage -> previously selected user groups are now empty.

### Expected Result

- Previously saved user group selections remain intact per template/stage.
- Both retrieval methods should return the same userGroups.

### Actual Result

Earlier stage templates show empty selections / missing userGroups.

get()->where() path drops related userGroups.

### Application Version

OJS 3.6.0

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib(main): [pkp/pkp-lib#12197 Fix template user groups reset/missing when saving templates across stages #12221](https://github.com/pkp/pkp-lib/pull/12221)
OJS (main): [pkp/pkp-lib#12197 Submodule update ##Hafsa-Naeem/i12197-main-fix #5280](https://github.com/pkp/ojs/pull/5280)

touhidurabir:
Hi @Hafsa-Naeem can you please describe what was happening here and how the PR solve it , I am just curious . If this is a hidden bug of a specific case with our settings builder (which support the settings data via eloquent), I think we need to back port it to 3.5 as well .

Hafsa-Naeem:
[...]

PR #12221, opened 2026-01-19
pkp/pkp-lib#12197 Fix template user groups reset/missing when saving templates across stages

Hafsa-Naeem:
for #12197

### DIFF
--- a/classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
+++ b/classes/migration/upgrade/v3_6_0/I857_ContributorRolesTypes.php
@@ -59,14 +59,15 @@ class I857_ContributorRolesTypes extends Migration
         /**
          * Add contributor roles and types to current contributors
          */
-
         $translations = UserGroup::withRoleIds([Role::ROLE_ID_AUTHOR])
             ->get()
-            ->map(fn (UserGroup $ug, int $ugId): array => [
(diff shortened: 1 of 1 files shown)


classes/core/SettingsBuilder.php
```diff
@@ -316,33 +316,43 @@ protected function getModelWithSettings(array|string $columns = ['*']): array
...
-        $settings->each(function (\stdClass $setting) use (&$rows, $primaryKey, $columns) {
+        $settings->each(function (stdClass $setting) use (&$rows, $rowIndexesById, $primaryKey) {
             $settingModelId = $setting->{$primaryKey};
 
-            // Even for empty('') locale, the multilingual props need to be an array
-            if (isset($setting->locale) && $this->isMultilingual($setting->setting_name)) {
-                $rows[$settingModelId]->{$setting->setting_name}[$setting->locale] = $setting->setting_value;
-            } else {
-                $rows[$settingModelId]->{$setting->setting_name} = $setting->setting_value;
+            foreach ($rowIndexesById[$settingModelId] ?? [] as $index) {
```
Vitaliy-1: According to the prior code, we should anyway execute the content of the loop at least once. But I see that in the modification there can be the case `... ?? []`, thus it won't be executed. Is this correct?

Issue #12181, opened 2026-01-05
Add support for a configuration option to lock down plugin installation


asmecher:
**Description:**
Site administrators are currently allowed to install and/or upload 3rd party plugins through the Plugin Gallery and plugin upload features.

However, server administrators may want to lock down those features to prevent the introduction of new code in production deployments. This could be done easily using a `config.inc.php` parameter:

```
[security]
; This setting allows control over how new plugins can be installed via the web interface. Supported options (from most to least permissive):
;  on: Plugins can be installed or upgraded by upload and through the Plugin Gallery. (This is the default.)
;  gallery_only: Plugins can only be installed and upgraded only through the Plugin Gallery.
;  upgrade_only: Plugins can be upgraded only through the Plugin Gallery; no new plugins can be installed
;  off: Plugins cannot be installed or upgraded through the web interface.
allow_plugin_install = gallery_only
```

This is motivated by the EC Web Application Security Standard, requirement 77:

> The web application shall not execute files uploaded by the user.

---

**PRs**

OJS: pkp/ojs#5281
OMP: pkp/omp#2266
OPS: pkp/ops#1231
PKP-LIB: #12227

defstat:
@asmecher could you review those PRs?

asmecher:
I added a comment, thanks: https://github.com/pkp/pkp-lib/pull/12227#pullrequestreview-3683454937

defstat:
> I added a comment, thanks: [#12227 (review)](https://github.com/pkp/pkp-lib/pull/12227#pullrequestreview-3683454937)

Thanks @asmecher. I added support for `upgrade_only` also, and some comments/questions for the implementation

asmecher:
Thanks, @defstat, looks good! I added [a review](https://github.com/pkp/pkp-lib/pull/12227#pullrequestreview-3712215946) to the pkp-lib PR.

asmecher:
All merged! @Tribunal33, over to you for QA.

mreiko:
@Tribunal33 Can this be closed or would you like to verify?

Tribunal33:
Passes QA.  Played around with it a bit.  Not sure what is expected.  No Acceptance Criteria but was able to edit the config setting and it will hide/show the upload a plugin button.  

I tried it on journal with site level plugins as well. It affects both site level plugins and journal level plugin areas.

PR #12227, opened 2026-01-20
[PKP-LIB][main] #12181 add support for installation admins to allow control over how new plugins can be installed via the web interface

defstat:


defstat:
Thanks @asmecher!

Two design points to consider:

**1. Plugin Gallery visibility**
I currently keep the Plugin Gallery tab and grid visible for all modes, off included, and enable/disable Install and Upgrade actions independently based on the selected configuration.
I’ve added a flag so this behavior can easily be changed if we prefer to hide the gallery entirely.

**2. Placement of backend exception logic**
I limited PluginHelper to centralized upload checks and capability resolution, and enforce gallery install vs upgrade rules directly in the PluginGalleryGridHandler (where $isUpgrade is already known).
If you’d prefer all install/upgrade rules to be entirely in PluginHelper, I can refactor accordingly.

Happy to adjust either approach.

classes/plugins/PluginHelper.php
```diff
@@ -102,8 +205,13 @@ private function extractPlugin(string $filePath, string $originalFileName, calla
      *
      * @return Version Version of installed plugin on success
      */
-    public function installPlugin(string $path, string $originalFileName): Version
+    public function installPlugin(string $path, string $originalFileName, bool $gallerySource = false): Version
```
asmecher, line 209: New parameter should be in the self-documentation above.

classes/plugins/PluginHelper.php
```diff
@@ -175,8 +283,13 @@ protected function _getConnectionParams(): array
      * @param string $originalFileName Original filename of plugin archive
      *
      */
-    public function upgradePlugin(string $category, string $plugin, string $path, string $originalFileName): Version
+    public function upgradePlugin(string $category, string $plugin, string $path, string $originalFileName, bool $gallerySource = false): Version
```
asmecher, line 288: New parameter needs self-documentation.

locale/en/manager.po
```diff
@@ -1646,6 +1646,18 @@ msgstr ""
 msgid "manager.plugins.uploadFailed"
 msgstr "Please ensure a file was selected for upload."
 
+# fuzzy
+msgid "manager.plugins.galleryInstallsDisabled"
+msgstr "Plugin installs through the gallery are disabled by the installation owner."
```
asmecher:
```suggestion
msgid "manager.plugins.galleryInstallsDisabled"
msgstr "Plugin installs through the Plugin Gallery are disabled by the Site Administrator."
```

locale/en/manager.po
```diff
@@ -1646,6 +1646,18 @@ msgstr ""
...
+
+# fuzzy
+msgid "manager.plugins.galleryUpgradesDisabled"
+msgstr "Plugin upgrades through the gallery are disabled by the installation owner."
```
asmecher:
```suggestion
msgid "manager.plugins.galleryUpgradesDisabled"
msgstr "Plugin upgrades through the Plugin Gallery are disabled by the Site Administrator."
```

locale/en/manager.po
```diff
@@ -1646,6 +1646,18 @@ msgstr ""
...
+
+# fuzzy
+msgid "manager.plugins.uploadPluginsDisabled"
+msgstr "Plugin installs and upgrades through file upload if disabled by the installation owner"
```
asmecher:
```suggestion
msgid "manager.plugins.uploadPluginsDisabled"
msgstr "Plugin uploads are disabled by the Site Administrator."
```

classes/plugins/PluginHelper.php
```diff
@@ -41,6 +41,109 @@ class PluginHelper
...
+            'security',
+            'allow_plugin_install',
+            self::PLUGIN_INSTALL_ON
+        );
+
+        // boolean config values
+        if (is_bool($raw)) {
+            return $raw ? self::PLUGIN_INSTALL_ON : self::PLUGIN_INSTALL_OFF;
+        }
+
```
asmecher, line 61: Ah, I can see we're running afoul of the config loader's treatment of `On` and `Off` as booleans rather than strings. Up to you, but I'd be tempted to use `all` and `none` instead of `on` and `off` for the possible setting values -- it would mean that everything is consistently a string and we could get rid of this special case.

classes/plugins/PluginHelper.php
defstat: Documentation added

classes/plugins/PluginHelper.php
defstat: Documentation added

locale/en/manager.po
defstat: Changed

locale/en/manager.po
defstat: Changed

locale/en/manager.po
defstat: Changed

Issue #12204, opened 2026-01-13
Peer review public visibility options should set defaults


ewhanson:
Currently the "Publicly Show Reviewer Comments" option is a global toggle that either enables or disables the public visibility of reviews. Instead, it should set the default that is configurable when a review is assigned in the same way the default review mode is set. This will also require adding the public visibility as an option upon review assignment. Public visibility should be separate from the review mode, being able to display public reviews even when the review mode is not open (i.e. reviewer names should be anonymized).

Tasks:

- [ ] Add public visibility option to review assignment
- [ ] Have "Publicly Show Reviewer Comments" set the default public visibility option
- [ ] ~~Change how public reviews are displayed based on the assigned review mode (either anonymous or open)~~ This was already existing.

**Pull Request**
pkp-lib: https://github.com/pkp/pkp-lib/pull/12228
ojs: https://github.com/pkp/ojs/pull/5282
omp: https://github.com/pkp/omp/pull/2213
ops: https://github.com/pkp/ops/pull/1183

taslangraham:
Ready for review @ewhanson. PRs are listed at https://github.com/pkp/pkp-lib/issues/12204

ewhanson:
Thanks @taslangraham. Just a few small things in pkp-lib, but feel free to merge after addressing those.

taslangraham:
All merged @ewhanson. Thanks for the review!

Tribunal33:
Hi @taslangraham I can confirm that the default state that is set at the journal level is acurately setting the state of the individual reviewer mode.  The new public visibility is also being set.  This is working.  However, adding the reviewer is not working.  It throws an error.

<img width="742" height="152" alt="Image" src="https://github.com/user-attachments/assets/33b5a363-5d3a-4029-9bad-a9e9143ca640" />

Assuming this might need a migration?

`[Thu Jan 22 22:00:27.191622 2026] [php:error] [pid 524276] [client 64.180.254.142:62791] PHP Fatal error:  Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'is_review_publicly_visible' in 'field list' in `

Tribunal33:
This is now working and is passing QA.

Here is the updated testing scenarios for #12045 
1. Test for Setting field Default review mode are displayed in correct spot
2. Test for Setting field Publicly show reviewer comments in correct spot
3. Test that the default reviewer options are being set and maintained when reviewer is added
[...]

PR #12228, opened 2026-01-20
pkp/pkp-lib#12204 Allow peer review visibility to be set when creating review assignment

taslangraham:
For pkp/pkp-lib#12204

### DIFF
--- a/classes/context/Context.php
+++ b/classes/context/Context.php
@@ -45,7 +45,7 @@ abstract class Context extends \PKP\core\DataObject
     public const SETTING_DOI_CREATION_TIME = 'doiCreationTime';
     public const SETTING_DOI_AUTOMATIC_DEPOSIT = 'automaticDoiDeposit';
     public const SETTING_DOI_VERSIONING = 'doiVersioning';
-    public const SETTING_ENABLE_PUBLIC_PEER_REVIEWS = 'enablePublicPeerReviews';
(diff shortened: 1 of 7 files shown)


classes/submission/reviewAssignment/ReviewAssignment.php
```diff
@@ -228,6 +228,22 @@ public function setReviewRoundId($reviewRoundId)
...
+    {
+        return (bool)$this->getData('isReviewPubliclyVisible');
+    }
+
+    /**
+     * Set the public visibility of the review.
+     */
+    public function setIsReviewPubliclyVisible(bool $isReviewPubliclyVisible): void
+    {
+        $this->setData('isReviewPubliclyVisible', 'isReviewPubliclyVisible');
```
ewhanson: The second argument is suppose to be the variable `$isReviewPubliclyVisible`, right?

classes/context/Context.php
```diff
@@ -75,7 +75,7 @@ public function areDoisEnabled(): bool
      * Whether peer reviews are publicly visible for this context
      *
      */
-    public function arePeersReviewPublic(): bool
+    public function arePeerReviewsPublic(): bool
```
ewhanson: I think this functions role has changed it would be more clear if the named reflected that this is setting a default for individual revieweres, rather than a global "on/off" setting.

ewhanson (changes requested):
Looks good! Just two small questions.

Issue #12044, opened 2025-11-18
Add API endpoint for review summary


ewhanson:
A summary of peer review elements should be available via the API to be included on the article landing page as well as displayable as a widget elsewhere. This should be a public endpoint  that does not require authentication.

Required elements:
- Number of accepted/approved, revisions requested, and not approved/rejected
	- Duplicates/changed decisions should not be counted, e.g. if there are two review rounds and reviewer A accepts in round 1, that counts once. If reviewer B rejects in round 1 and round 2, that counts as rejected once. If reviewer C requests revisions in round 1 and accepts in round 2, that counts as accept. 
- Total number of peer reviewed versions published
- Current version info (and date)
- Review model (open, anonymous, double anonymous)

### Pull Requests
pkp-lib: https://github.com/pkp/pkp-lib/pull/12057
ojs: https://github.com/pkp/ojs/pull/5228
omp(submodule update): https://github.com/pkp/omp/pull/2191
ops(submodule update): https://github.com/pkp/ops/pull/1159

_Details of changes made: https://github.com/pkp/pkp-lib/issues/12044#issuecomment-3638361125_

**Changes Introduced**

The following endpoints were added to OJS:

Returns review summary information for submission with the given `submissionId`.

**[GET] `peerReviews/open/submissions/{submissionId}/summary`**
<details>
<summary>Example Response</summary>

```json
{
  "submissionId": 17,
  "reviewerRecommendations": [
    {
      "recommendationTypeId": 1,
      "recommendationTypeLabel": "Approved",
      "count": 2
    },
    {
      "recommendationTypeId": 3,
      "recommendationTypeLabel": "Revisions Requested",
      "count": 0
    },
    {
      "recommendationTypeId": 2,
      "recommendationTypeLabel": "Not Approved",
      "count": 1
    },
    {
      "recommendationTypeId": 4,
      "recommendationTypeLabel": "With Comments",
      "count": 1
    }
  ],
  "submissionPublishedVersionsCount": 1,
  "reviewerCount": 4,
  "submissionCurrentVersion": {
    "versionString": "Version of Record 1.0",
    "datePublished": "2026-02-18"
  }
}
```
</details>

**[GET] `peerReviews/open/submissions/summary`**
Returns peer review summaries for the submissions specified in the `submissionIds` query parameter.

<details>
<summary>Example Response</summary>

```json
[
  {
    "submissionId": 1,
    "reviewerRecommendations": [
      {
[...]

PR #12230, opened 2026-01-20
adding api tests for open review #12044

Tribunal33:
adding api tests for #12044


cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+		cy.request('index.php/publicknowledge/api/v1/submissions?status=3')
+			.then(response => {
+				expect(response.status).to.eq(200);
+				expect(response.body.items).to.have.length.greaterThan(0);
+				const submission = response.body.items[0];
+				submissionId = submission.id;
+				publicationId = submission.publications[0].id;
+				cy.log(`Using submission ID: ${submissionId}, publication ID: ${publicationId}`);
+			});
+	});
```
taslangraham, line 75: Best to select a submission that is not just in review stage, but also has completed review assignments. Then in the tests, you can validate that those reviews are being returned by the endpoint.
Important to note, though, that by default the review assignments are not publicly visible, which means we will either have to add publicly visible ones to the test dataset, create public ones when running these api test or edit existing ones to be public.

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+
+	it('Tests peer reviews endpoint without public reviews enabled', function() {
+		// Test that the endpoint returns 401 when public reviews are not enabled
+		cy.request({
+			url: `index.php/publicknowledge/api/v1/peerReviews/open/publications/${publicationId}`,
+			failOnStatusCode: false
+		}).then(response => {
+			expect(response.status).to.eq(401);
+		});
+	});
```
taslangraham: This test case is no longer relevant as the workflow setting `Publicly show reviewer comments` no longer acts as a global flag (see https://github.com/pkp/pkp-lib/issues/12204) enabling or disabling public peer reviews. Instead, it now acts as the default option for the public visibility of individual review assignments -- the visibility a review assignment can be set when creating it.

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+
+	it('Enables public peer reviews', function() {
+		cy.login('dbarnes', null, 'publicknowledge');
+		cy.visit('index.php/publicknowledge/management/settings/workflow');
+		cy.get('#review-button').click();
+		
+		// Enable public peer reviews
+		cy.get('input[name="enablePublicPeerReviews"]').check();
```
taslangraham:
```suggestion
		cy.get('input[name="defaultReviewPublicVisibility"]').check();
```

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+		// Test that the endpoint returns 401 when public reviews are not enabled
+		cy.request({
+			url: `index.php/publicknowledge/api/v1/peerReviews/open/publications/${publicationId}`,
+			failOnStatusCode: false
+		}).then(response => {
+			expect(response.status).to.eq(401);
+		});
+	});
+
+	it('Enables public peer reviews', function() {
```
taslangraham:
```suggestion
	it('Make review assignment default visibility public', function() {
```

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+		cy.request({
+			url: 'index.php/publicknowledge/api/v1/peerReviews/open/submissions/summary?submissionIds[]=invalid',
+			failOnStatusCode: false
+		}).then(response => {
+			expect(response.status).to.eq(400);
+			expect(response.body).to.have.property('error');
+		});
+	});
+
+	it('Disables public peer reviews', function() {
```
taslangraham:
```suggestion
	it('Set review assignment default visibility to not publicly visible', function() {
```

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+		cy.login('dbarnes', null, 'publicknowledge');
+		cy.visit('index.php/publicknowledge/management/settings/workflow');
+		cy.get('#review-button').click();
+		
+		// Disable public peer reviews
+		cy.get('input[name="enablePublicPeerReviews"]').uncheck();
```
taslangraham:
```suggestion
		cy.get('input[name="defaultReviewPublicVisibility"]').uncheck();
```

cypress/tests/integration/API.cy.js
```diff
@@ -56,7 +53,192 @@ describe('API tests', function() {
...
+		// Get a publication ID from an existing published article
+		cy.login('dbarnes', null, 'publicknowledge');
+		cy.request('index.php/publicknowledge/api/v1/submissions?status=3')
+			.then(response => {
+				expect(response.status).to.eq(200);
+				expect(response.body.items).to.have.length.greaterThan(0);
+				const submission = response.body.items[0];
+				submissionId = submission.id;
+				publicationId = submission.publications[0].id;
+				cy.log(`Using submission ID: ${submissionId}, publication ID: ${publicationId}`);
```
taslangraham, line 73: Is this intentionally left in?

taslangraham (changes requested):
Hey @Tribunal33,
Just a few questions and suggestions.
Overall, these tests do a good job of validating that data and error responses are returned for the expected scenarios. One possible improvement would be to also validate the actual contents of the data returned by the endpoints. If interested, example data for the endpoints' response can be found at these comments:
- data for summary endpoints: https://github.com/pkp/pkp-lib/issues/12044#issuecomment-3638361125
- data for publication peer reviews endpoint: https://github.com/pkp/pkp-lib/issues/11922#issuecomment-3408690209

cypress/tests/integration/API.cy.js
Tribunal33: Yes it was.  it's for if it fails can get good info.  Not sure what the policy for logs are here though.  If you see an accidental log it will be less structured :).  More like "you are here" or something like that.

cypress/tests/integration/API.cy.js
Tribunal33: I c.  Hmm I'd liked to strictly keep these api tests focused on api.  Dataset should have a valid entry to test against.  Are there any ways to use the API to update a submission and add reviews that way?

Tribunal33: Also this api call ( index.php/publicknowledge/api/v1/submissions?status=3 ) is checking for articles in the published stage not review stage.  I'm just grabbing the first one that is published.  Any issues with this?

cypress/tests/integration/API.cy.js
Tribunal33: Yeah I can see that now.  I will remove or is there a way to check if the flag is correct with api?  So If I was to disable it then it would be disabled in the article?

Issue #12245, opened 2026-01-26
Workflow | Implement PMUR workflow in submission workflow queues


asmecher:
The PMUR workflow has been implemented in the OJS/OMP/OPS `main` branch via https://github.com/pkp/pkp-lib/issues/4860 and others. However, a "Publish" action currently moves the submission to the Published queue, even if the submission is currently a preprint being put through Publish-Review-Curate style review.

Currently, there are two sources of truth about "status" -- the `submissions.status` and `publications.status` columns in the database. These are used to determine a mixture of things:
 - Whether the submission is considered to be part of one of the active workflow queues (as opposed to "Published" or "Declined")
 - Whether a reader can access the submission through the reader interface
 - Whether the submission is included in the OAI-PMH interface
 - Whether a publication is listed as ready to be deposited in various deposit workflows (e.g. PubMed Central and Zenodo).

These two status columns should be differentiated to serve two different purposes:
 - The `submissions.status` column should contain the workflow status, i.e. whether the editor is in some way active on the submission. This is determined on a per-submission basis, i.e. a submission can be active on a PMUR even though it already has a published preprint.
 - The `publication.status` column should determine the publication status, i.e. whether it is posted for readers to view. This is determined on a per-publication basis, i.e. a preprint can be posted and viewable by readers even though the submission is still under active review.

Acceptance criteria:

- [ ] Check to ensure that submission status behaves appropriately for the Editor as the submission moves through a PMUR-style workflow:
  - [ ] When a PMUR version is published, ensure that the submission enters the appropriate Review workflow queue.
  - [ ] When a non-PMUR version is published, ensure that the submission moves to the Published workflow queue.

(This replaces https://github.com/pkp/pkp-lib/discussions/11762#discussioncomment-15221831)

PRs:
- https://github.com/pkp/pkp-lib/pull/12247
- https://github.com/pkp/ojs/pull/5308
- https://github.com/pkp/omp/pull/2242
- https://github.com/pkp/ops/pull/1210

asmecher:
In 3.5 and below, there is a lot of duplicate updating being done when workflow actions are performed, and in the case of submission status, this shows up as multiple calls to `Submission::updateStatus` when certain workflow actions are performed (e.g. a Publish action). It would be nice to clean this up, though possibly a rat's nest.

`\APP\submission\Repository::updateStatus` (and `\PKP\submission\Repository::updateStatus`) are called (as of pre-3.6) by
[...]

PR #12247, opened 2026-01-26
pkp/pkp-lib#12245 Move PMUR submissions into review on PMUR flag

asmecher:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1454,10 +1458,14 @@ public function publishPublication(Request $illuminateRequest): JsonResponse
         $shouldCreatePMURReviewRound = filter_var($illuminateRequest->input('createPMURReviewRound'), FILTER_VALIDATE_BOOL);
         if ($shouldCreatePMURReviewRound) {
             try {
-                $this->createPMURReviewRound($submissionContext, $submission, $publication);
+                $this->createPMURReviewRound($submissionContext, $submission, $publication, Submission::STATUS_QUEUED);
```
jardakotesovec, line 1461: @asmecher @ewhanson  I understand the intention to keep it in review queue. But as soon as the submission status gets recalculated it jumps to published state without any control from user. 
That would happen if
- VoR draft is deleted
- New version is created
- Also when issue is deleted/unpublished - but thats more of the edge case

And also it means that this checkbox is not only shortcut (which I think was original intention), but key for the submission status being correct.

Issue #11902, opened 2025-10-06
[Structured Citations] - Part 2 enhancements


Tribunal33:
**Description:**
- Include specific context details that have been discovered that surround this developer task.
Carried over from previous ticket:
Still to finish:
- [ ] double check catch exception for guzzle requests
- [ ] test native import export
- [ ] adapt article_details.tpl for each theme
- [ ] adapt jatsTemplate
- [ ] adapt crossrefReferenceLinking 

During testing some improvements were suggested to facilitate a more fluid user experience that were not in scope for the initial feature.  It, additionally, addresses some requirements that were unclear.
- [ ] Async functionality with jobs processing. (Currently there is a noticible slow down in response times for other users navigating the site when a large job like this is being processed)
- [ ] Arxiv does not need URL validation - arXiv:2401.12345 is valid.  (Unclear what should be validated for this field)
- [ ] Increase the size of the citation input field.  Feels too small for the amount of information and spacing between citations.
- [ ] Disable Metadata lookup if it is enabled site-wide in settings for author.
- [ ] Editing citation should have more informative validation warnings for Type and source type or have auto complete or dropdown list
- [ ] Editing citation fields should limit the max number of characters for these free text fields. (Security general best practice is to limit user input)
https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#:~:text=When%20designing%20regular%20expression%2C%20be,Allow%20List%20Regular%20Expression%20Examples
- [ ] Should limit to 100 citations at a time (for both security and navigating the site while it is processing is sluggish).  Similar to max text length except for separated based on new lines.  You shouldn't be able to add more until it is done processing the first batch.

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
Precondidtion: Workflow Settings > Submission > Metadata > References > Enable References > Enable Metadata lookup.

- Not a single set of steps to cover all the extra requested improvements.  (I could break it out into more tickets if need be)

1.  Async - Process 100+ citations then swap to another browser and navigate the site.
2. After processing the citation select Edit and add the Arxiv reference without a URL.  
3. After processing the citation select 3 dots > Edit and there are a few fields which difficult to understand what is expected.  Like type and source.  Kept getting an warning when I added something wrong.
4. Add whatever the MaxLength +1 of the field is and observe the warning and is won't save the form.

**Solution:** What the proposed solutions are and additional affected areas
Solutions described above.

**What application are you using?**
  OJS 3.6

[...]

PR #12268, opened 2026-01-29
pkp/pkp-lib#11902 strip tags from citation, improve URL regex

bozana:
s. https://github.com/pkp/pkp-lib/issues/11902, Fix 3.

### DIFF
--- a/classes/citation/pid/Url.php
+++ b/classes/citation/pid/Url.php
@@ -20,6 +20,6 @@ class Url extends BasePid
 {
     /** @copydoc AbstractPid::regex */
     public const regexes = [
-        '#(http|https|ftp)://[\d\w\.-]+\.[\w\.]{2,6}[^\s\]\[\<\>]*/?#'
+        '#https?://(www\.)?[-a-zA-Z0-9@:%._\+~\#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\+.~\#?&//=]*)#'
     ];
 }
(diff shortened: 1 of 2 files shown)


bozana:
Hi @asmecher, could you review these changes?
The tags are stripped from the raw citation when it is send to Crossref, to get/find the DOI.
Also, previous URL regex would not remove `"` at the end of e.g. `<a href="https://pubmed.ncbi.nlm.nih.gov/31382596">`, so now a different URL regex is used.
Thanks a lot!

bozana:
Hi @kaitlinnewson, would you maybe be able to review this PR? -- I think Alec is very busy with other tasks...

bozana:
I am going to merge this today, even without review -- it would be good to have for ORE...

classes/citation/pid/Url.php
```diff
@@ -20,6 +20,6 @@ class Url extends BasePid
 {
     /** @copydoc AbstractPid::regex */
     public const regexes = [
-        '#(http|https|ftp)://[\d\w\.-]+\.[\w\.]{2,6}[^\s\]\[\<\>]*/?#'
+        '#https?://(www\.)?[-a-zA-Z0-9@:%._\+~\#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\+.~\#?&//=]*)#'
```
kaitlinnewson, line 23: Though it's rare, should we support the possibility for TLDs longer than 4 characters?

bozana: yes, it could actually be 63 char long...

Issue #10496, opened 2024-10-02
Null pointer in the Editorial Reminder if the user has blocked notifications


nwoodward:
**Describe the bug**
We see several errors in the OJS logs about a null pointer in EditorialReminder.php, and it's due to the $notification being null. In the handle() method below the $notification variable returned from `$notificationManager->createNotification()` could potentially be null. So when it's used in the `allowUnsubscribe()` method of `$mailable` it throws a null pointer error.

https://github.com/pkp/pkp-lib/blob/09388ce9f4f3feef99d52cdcf95517620bf50378/jobs/email/EditorialReminder.php#L143-L162

This happens when the user has blocked notifications. In the `createNotification()` method of `PKPNotificationOperationManager` there's a check at the beginning to see if the user exists and has blocked notifications, in which case `null` is returned.

https://github.com/pkp/pkp-lib/blob/09388ce9f4f3feef99d52cdcf95517620bf50378/classes/notification/PKPNotificationOperationManager.php#L143-L167

I don't understand the code well enough to say if the fix should be a null check on `$notification` after it gets created in `EditorialReminder.php`, and if so then return from `handle()`. Or if `$mailable` should still be called without the `allowUnsubscribe($notification)` method.

**What application are you using?**
 OJS 3.4.0-5

**PRs**

#### Stable-3.4.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12275
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5306 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2224 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1193 [TEST ONLY]

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12298
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5318 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2225 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1194 [TEST ONLY]

#### main
[...]

PR #12275, opened 2026-01-30
pkp/pkp-lib#10496 Better check of notification setting for editorial reminder

touhidurabir:
for pkp/pkp-lib#10496 .

### DIFF
--- a/jobs/email/EditorialReminder.php
+++ b/jobs/email/EditorialReminder.php
@@ -192,6 +192,19 @@ class EditorialReminder extends BaseJob
     {
         /** @var NotificationSubscriptionSettingsDAO $notificationSubscriptionSettingsDao  */
         $notificationSubscriptionSettingsDao = DAORegistry::getDAO('NotificationSubscriptionSettingsDAO');
+
+        // Check if the editor/user has blocked this notification type itself
+        // If so, email notification should not be sent anyway
+        $blockedNotifications = $notificationSubscriptionSettingsDao->getNotificationSubscriptionSettings(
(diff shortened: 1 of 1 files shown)


jobs/email/EditorialReminder.php
kaitlinnewson, line 180: Since the return type of `createNotification` is `?Notification`, should there be a null check before it is passed here as well?

jobs/email/EditorialReminder.php
```diff
@@ -192,6 +192,19 @@ protected function isSubscribed(): bool
...
+        $blockedNotifications = $notificationSubscriptionSettingsDao->getNotificationSubscriptionSettings(
+            NotificationSubscriptionSettingsDAO::BLOCKED_NOTIFICATION_KEY,
+            $this->editorId,
+            $this->contextId
+        );
+        if (in_array(Notification::NOTIFICATION_TYPE_EDITORIAL_REMINDER, $blockedNotifications)) {
+            return false;
+        }
+
+        // Check if the editor/user has blocked ONLY the email notifaction for this notification type
```
kaitlinnewson:
```suggestion
        // Check if the editor/user has blocked ONLY the email notification for this notification type
```

taslangraham (approved):
Approved. Nothing further to add beyond @kaitlinnewson comments.

jobs/email/EditorialReminder.php
touhidurabir: that is one way to do that but it should not be necessary . the logic is  `isSubscribed` should be responsible to properly determine the subscription logic to the notification and return early .

Issue #10405, opened 2024-09-12
Permit file revisions of galley files and XML


asmecher:

mreiko:
@Devika008 Can you please let me know your timing to complete design on this and any other remaining ORE related scope?

Devika008:
Hello @mreiko,

I am chipping away at this as I have some other things to close. Will get to it by mid-september

touhidurabir:
Hi @asmecher about the galley file revision, we have galley for OJS and OPS but it is bit different as OPS galley are actually submission file unlike OJS, so does this galley revision also applied for OPS or just OJS only ?

touhidurabir:
@asmecher aside from [question above](https://github.com/pkp/pkp-lib/issues/10405#issuecomment-3845967216), PRs are review for review at https://github.com/pkp/pkp-lib/issues/10405#issue-2523195734 .

asmecher:
> Hi @asmecher about the galley file revision, we have galley for OJS and OPS but it is bit different as OPS galley are actually submission file unlike OJS, so does this galley revision also applied for OPS or just OJS only?

OMP sends files that authors upload directly into a `file_stage` of `SubmissionFile::SUBMISSION_FILE_PROOF` (10), just like OJS and OMP. I don't think you'll have to do anything different; you can support OPS the same as the other apps. Managing galleys happens in the same place ("Galleys") in the editorial interface:

<img width="1151" height="476" alt="Image" src="https://github.com/user-attachments/assets/2362a6e3-53b4-4c24-b5d9-3ec7efea2640" />

asmecher:
(@touhidurabir, tagging you re: the comment above in case you don't see it otherwise)

touhidurabir:
@asmecher PRs are ready to review at https://github.com/pkp/pkp-lib/issues/10405#issue-2523195734 .

asmecher:
Thanks, @touhidurabir, and sorry for the wait. I've added some review comments:
- https://github.com/pkp/pkp-lib/pull/12286#pullrequestreview-3899515640
- https://github.com/pkp/ojs/pull/5313#pullrequestreview-3899425598

I haven't reviewed the ui-library part of it; let's reach readiness on the back end first.

touhidurabir:
@asmecher PRs at https://github.com/pkp/pkp-lib/issues/10405#issue-2523195734 are ready for review again .

asmecher:
Thanks, @touhidurabir, this looks and tests well for me!

[...]

PR #12286, opened 2026-02-03
pkp/pkp-lib#10405 file revision history of jats and galley

touhidurabir:
for pkp/pkp-lib#10405 .

### DIFF
--- a/api/v1/jats/PKPJatsController.php
+++ b/api/v1/jats/PKPJatsController.php
@@ -36,6 +36,7 @@ use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\authorization\internal\SubmissionCompletePolicy;
 use PKP\security\authorization\internal\SubmissionRequiredPolicy;
 use PKP\security\authorization\ContextRequiredPolicy;
+use PKP\security\authorization\internal\PublicationIsSubmissionPolicy;
 use PKP\security\authorization\internal\PublicationRequiredPolicy;
 use PKP\security\Role;
 use PKP\services\PKPSchemaService;
@@ -89,7 +90,7 @@ class PKPJatsController extends PKPBaseController
 
         })->whereNumber(['submissionId', 'publicationId']);
 
-        // Public route for JATS download which requried no authentication
+        // Public route for JATS download which requires no authentication
         Route::get('download', $this->publicDownload(...))
             ->name('publication.jats.publicDownload')
             ->whereNumber(['submissionId', 'publicationId']);
@@ -109,8 +110,9 @@ class PKPJatsController extends PKPBaseController
 
         if ($actionName === 'publicDownload') {
             $this->addPolicy(new PublicationRequiredPolicy($request, $args, 'publicationId'));
-            
-            // For the public api endpoint, we don't need to role/access based authorization
+            $this->addPolicy(new PublicationIsSubmissionPolicy(__('api.publications.403.submissionsDidNotMatch')));
+
+            // For the public api endpoint, we don't need role/access based authorization
             return parent::authorize($request, $args, $roleAssignments);
         }
 
@@ -270,35 +272,33 @@ class PKPJatsController extends PKPBaseController
 
     /**
      * Public endpoint to download JATS XML.
-     * This endpoint does not require authentication.
-     * Response is cached as per defined in \PKP\jats\Repository::JATS_FILE_CACHE_LIFETIME to prevent DDOS.
      */
     public function publicDownload(Request $illuminateRequest): Response|JsonResponse
     {
         $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION); /** @var \APP\submission\Submission $submission */
         $publication = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_PUBLICATION); /** @var \APP\publication\Publication $publication */
 
-        // Verify publication is published and has public visibility enabled
+        // Verify public visibility is enabled
         if (!$publication->getData('jatsPublicVisibility')) {
             return response()->json([
                 'error' => __('api.403.unauthorized'),
             ], Response::HTTP_FORBIDDEN);
         }
 
-        // If publication is not published yet,
-        // still woule be available for public download if public visibility enabled
-        // but only for authorised roles such as Admin, JM and & Editor.
-        // This is to support the non published submission preview mode only.
-        if ($publication->getData('status') != PKPPublication::STATUS_PUBLISHED) {
+        $isPublished = $publication->getData('status') == PKPPublication::STATUS_PUBLISHED;
+
+        // If publication is not published yet, allow download only for users
+        // who can preview (e.g. managers, sub-editors, assistants, assigned authors).
+        // This supports the article preview mode.
+        if (!$isPublished) {
             $user = $this->getRequest()->getUser();
-            if (!$user || !$user->hasRole([Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR], $submission->getData('contextId'))) {
+            if (!$user || !Repo::submission()->canPreview($user, $submission)) {
                 return response()->json([
                     'error' => __('api.403.unauthorized'),
                 ], Response::HTTP_FORBIDDEN);
             }
         }
 
-        // Get cached JATS content
         $jatsContent = Repo::jats()->getPublicJatsContent($publication->getId(), $submission->getId());
 
         if (!$jatsContent) {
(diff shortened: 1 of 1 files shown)


api/v1/jats/PKPJatsController.php
```diff
@@ -89,6 +99,11 @@ public function authorize(PKPRequest $request, array &$args, array $roleAssignme
         $illuminateRequest = $args[0]; /** @var \Illuminate\Http\Request $illuminateRequest */
         $actionName = static::getRouteActionName($illuminateRequest);
 
+        // no authorization check for public access api request
+        if ($actionName === 'publicDownload') {
+            return true;
```
asmecher: This totally shortcuts the call to `parent::authorize`; are you sure this is necessary?

api/v1/jats/PKPJatsController.php
```diff
@@ -214,4 +228,91 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+
+        return response()->json([
+            'jatsPublicVisibility' => $newVisibility,
+            'publicationId' => $publication->getId(),
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Public endpoint to download JATS XML.
+     * This endpoint does not require authenticatio.
```
asmecher:
```suggestion
     * This endpoint does not require authentication.
```

api/v1/jats/PKPJatsController.php
```diff
@@ -214,4 +228,91 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+        $publication = Repo::publication()->get($publicationId, $submissionId);
+
+        if (!$publication) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        // Verify publication is published
+        if ($publication->getData('status') !== PKPPublication::STATUS_PUBLISHED) {
```
asmecher:
```suggestion
        if ($publication->getData('status') != PKPPublication::STATUS_PUBLISHED) {
```

api/v1/jats/PKPJatsController.php
```diff
@@ -214,4 +228,91 @@ public function delete(Request $illuminateRequest): JsonResponse
...
+
+    /**
+     * Public endpoint to download JATS XML.
+     * This endpoint does not require authenticatio.
+     * Response is cached as per defined in \PKP\jats\Repository::JATS_FILE_CACHE_LIFETIME to prevent DDOS.
+     */
+    public function publicDownload(Request $illuminateRequest): Response|JsonResponse
+    {
+        $submissionId = (int) $illuminateRequest->route('submissionId');
+        $publicationId = (int) $illuminateRequest->route('publicationId');
```
asmecher: These two are checked for internal consistency, but it's never checked to ensure that the submission is in the current journal!

classes/jats/Repository.php
```diff
@@ -154,51 +159,86 @@ public function addJatsFile(
...
+        return $this->getJatsFile($publication->getId(), $submission->getId(), $genres->toArray());
+    }
+
+    /**
+     * Delete the JATS file for a publication.
+     *
+     * @param int $publicationId The publication ID
+     * @param int|null $submissionId Optional submission ID filter
+     */
+    public function deleteJatsFile(int $publicationId, ?int $submissionId = null): void
```
asmecher: I see you're just following existing patterns, but this Repository doesn't follow our usual conventions -- it should be `function delete`, not `function deleteJatsFile`. (The latter is redundant.)

classes/jats/Repository.php
```diff
@@ -154,51 +159,86 @@ public function addJatsFile(
...
+    {
+        $query = Repo::submissionFile()
+            ->getCollector()
+            ->filterByFileStages([SubmissionFile::SUBMISSION_FILE_JATS])
+            ->filterByAssoc(Application::ASSOC_TYPE_PUBLICATION, [$publicationId]);
+
+        if ($submissionId) {
+            $query = $query->filterBySubmissionIds([$submissionId]);
+        }
```
asmecher: Most stylistic but:
```suggestion
            ->filterByAssoc(Application::ASSOC_TYPE_PUBLICATION, [$publicationId])
            ->filterBySubmissionIds($submissionId !== null ? [$submissionId] : null);
```
Note that we're being strict about the `null` here! If we're not careful, someone could pass in a `null` to bypass the test that the publication and submission match up.

classes/jats/Repository.php
```diff
@@ -235,4 +275,56 @@ public function getFileStages(): array
...
+        $cacheKey = $this->getPublicJatsCacheKey($publicationId);
+
+        return Cache::remember($cacheKey, static::JATS_FILE_CACHE_LIFETIME, function () use ($publicationId, $submissionId) {
+            $submission = Repo::submission()->get($submissionId);
+            
+            /** @var \PKP\context\Context $context */
+            $context = app()->get('context')->get($submission->getData('contextId'));
+            
+            $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
+            $genres = $genreDao->getEnabledByContextId($context->getId());
```
asmecher:
```suggestion
            $genreDao = DAORegistry::getDAO('GenreDAO'); /** @var GenreDAO $genreDao */
            $genres = $genreDao->getEnabledByContextId($submission->getData('contextId'));
```

classes/jats/Repository.php
```diff
@@ -235,4 +275,56 @@ public function getFileStages(): array
...
+     * Get JATS content for public download with given cache life time as defined
+     *
+     * This method caches the JATS XML content to prevent server overload
+     * from repeated public requests. Cache is invalidated when:
+     * - JATS file is uploaded (addJatsFile)
+     * - JATS file is deleted (deleteJatsFile)
+     * - Visibility setting changes (via controller)
+     *
+     * @param int $publicationId The publication ID
+     * @param int $submissionId Optional submission ID filter
```
asmecher: This description is wrong; the `$submissionId` is not optional.

controllers/grid/eventLog/GalleyFileEventLogGridHandler.php
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class GalleyFileEventLogGridHandler
+ *
+ * @brief Grid handler presenting the galley files event log grid.
+ */
+
```
asmecher: This file (and more below) basically copy-paste the submission file event log stuff. I don't think that's necessary -- the JATS files make use of the submission file toolset so it shouldn't be necessary to replicate all this code. It should be possible to call on the existing submission file event log code.

api/v1/jats/PKPJatsController.php
touhidurabir: Weird , I am pretty sure I added the following 3 policies the at the beginning of the authorize method 
```php
$this->addPolicy(new ContextRequiredPolicy($request));
        $this->addPolicy(new SubmissionRequiredPolicy($request, $args));
        $this->addPolicy(new SubmissionCompletePolicy($request, $args));
```

wondering did I made a mess with some rebase at some points .

api/v1/jats/PKPJatsController.php
touhidurabir: Truthfully not needed and we can still return from parent::authorize, updated .

api/v1/jats/PKPJatsController.php
touhidurabir: Anyway updated it .

controllers/grid/eventLog/GalleyFileEventLogGridHandler.php
touhidurabir: @asmecher with a quick look it seems like a copy paste but there are some significant changes like 
1. we need galley instead of submission file
2. we need different policy for authorize but also dont need the SubmissionFileAccessPolicy policy 
3. the load data is different 

The identical ones are 
1. getFilterForm
2. getFilterSelectionData

but then extending the SubmissionFileEventLogGridHandler will force us to call grandparent class authorrize like `SubmissionEventLogGridHandler::authorize` so that we can skip the SubmissionFileAccessPolicy in the `SubmissionFileEventLogGridHandler::authorize` . 

Or perhaps I am misunderstanding something here .

asmecher: We can rely on the `submission_file_id` in the `publication_galleys` table and open the information center for that file. No need to implement a duplicate for the galley. We'll be rewriting the history view shortly anyway, so need to make sure we're not adding technical debt to the current implementation.

touhidurabir: Thanks for the hint . this works . updated .

Issue #12251, opened 2026-01-27
API for Media Files


jardakotesovec:
### 1. Get Media Files

**GET** `/api/v1/submissions/{submissionId}/mediaFiles`

Returns a list of media files for a submission.

**Query Parameters:**
- `variantGroupIds` (optional) - Array of integers to filter by variant group IDs
- `variantTypes` (optional) - Array of variant type strings (`web` or `high_resolution`)

**Response:**
```json
{
  "itemsMax": 2,
  "items": [
    {
      "id": 1,
      "fileId": 100,
      "fileStage": 23,
      "name": {
        "en": "example-image.png"
      },
      "submissionId": 5,
      "variantGroupId": null,
      "variantType": "web"
    }
  ]
}
```

---

### 2. Add Media Files

**POST** `/api/v1/submissions/{submissionId}/mediaFiles`

Upload one or more media files from temporary file uploads.

**Request Body:**
```json
{
  "files": [
    {
      "temporaryFileId": 123,
      "variantType": "web"
    },
    {
      "temporaryFileId": 124,
      "variantType": "high_resolution"
    }
  ]
}
```

**Required Fields:**
- `files` - Array of file objects (minimum 1)
- `files[].temporaryFileId` - Integer ID of the temporary file
- `files[].variantType` - String, must be `web` or `high_resolution`

**Optional Fields:**
- Any submission file properties (e.g., `name`, `description`, `caption`, etc.)

**Response:**
```json
[
  {
    "id": 1,
    "fileId": 100,
    "fileStage": 23,
    "name": {
      "en": "example-image.png"
    },
    "submissionId": 5,
    "variantGroupId": null,
    "variantType": "web"
  }
]
```

---

### 3. Batch Link Media Files

**POST** `/api/v1/submissions/{submissionId}/mediaFiles/link`

Link or unlink multiple media file pairs in a single request.

**Request Body:**
```json
{
  "links": [
    {
      "primarySubmissionFileId": 1,
      "secondarySubmissionFileId": 2
    },
    {
      "primarySubmissionFileId": 3,
      "secondarySubmissionFileId": null
    }
  ]
}
```

**Required Fields:**
- `links` - Array of link objects (minimum 1)
- `links[].primarySubmissionFileId` - Integer ID of the primary submission file

**Optional Fields:**
- `links[].secondarySubmissionFileId` - Integer ID of the secondary file to link, or `null` to unlink

**Validation:**
- Files cannot be linked to themselves
[...]

PR #12306, opened 2026-02-05
pkp/pkp-lib#12251 API for Media Files

ewhanson:


api/v1/submissions/MediaFilesController.php
```diff
@@ -0,0 +1,463 @@
...
+     */
+    public function getMany(Request $illuminateRequest): JsonResponse
+    {
+        /** @var Submission $submission */
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        $collector = Repo::submissionFile()
+            ->getCollector()
+            ->filterBySubmissionIds([$submission->getId()])
+            ->filterByFileStages([SubmissionFile::SUBMISSION_FILE_MEDIA]);
```
blesildaramirez: Hi @ewhanson , apologies - I missed this during testing: we should also filter by `publicationId`, or `assocId` & `assocType`, since the media section sits within each publication version. With that, I think the `getMany` api may need to be updated to support these params, or possibly as query params.

api/v1/submissions/MediaFilesController.php
```diff
@@ -0,0 +1,463 @@
...
+                    $fileId = app()->get('file')->add(
+                        $temporaryFile->getFilePath(),
+                        $submissionDir . '/' . uniqid() . '.' . $extension
+                    );
+
+                    $params = collect($fileEntry)->except(['temporaryFileId'])->toArray();
+                    $params['fileId'] = $fileId;
+                    $params['submissionId'] = $submission->getId();
+                    $params['fileStage'] = SubmissionFile::SUBMISSION_FILE_MEDIA;
+                    $params['uploaderUserId'] = (int) $user->getId();
```
blesildaramirez, line 268: Related to https://github.com/pkp/pkp-lib/pull/12306/changes#r3135144388, we also need to store the `assocId` and `assocType` here..

classes/submissionFile/Repository.php
```diff
@@ -929,4 +930,26 @@ public function getSubmissionFileContent(SubmissionFile $submissionFile): string
...
+     *
+     * @return string[]
+     */
+    public function getCommonMediaFileFields(): array
+    {
+        return [
+            'caption',
+            'copyrightOwner',
+            'creator',
+            'credit',
```
blesildaramirez, line 945: I think the `dateCreated` and `language` are missing metadata fields here on this list.

classes/submissionFile/Repository.php
```diff
@@ -929,4 +930,26 @@ public function getSubmissionFileContent(SubmissionFile $submissionFile): string
...
+            'description',
+            'publisher',
+            'source',
+            'sponsor',
+            'subject',
+            'terms',
+            'viewable',
```
blesildaramirez: The `viewable` is not a metadata field for "artwork" or "supplementary" genre category. Maybe we should delete this?

blesildaramirez (approved):
Hi @ewhanson , while reviewing this, I noticed that media files aren't being filtered by `publicationId`, which is important since each media section sits within a specific publication version. ~~It might be better to address this in a separate PR or issue~~. I think it's important to include this to this PR. Please check my two initial comments regarding this.

I also have a couple of minor suggestions related to syncing the metadata fields for grouped media files.

Let me know if you have any questions regarding this. Thanks!

blesildaramirez (approved):
Tested all the changes and everything worked as expected.

Issue #12278, opened 2026-01-30
Enforce 1-hour edit window for discussion messages (including replies)


Hafsa-Naeem:
**Description:**
Limit editing of discussion/task notes so that a user can edit only their own message/reply, and only within 1 hour (3600 seconds) of creation. After that, edits should be rejected.

**Expected behaviour:**

- User can edit their own discussion message or reply if it was created < 1 hour ago.

- Editing is rejected if the message/reply is more than1 hour old.

- Editing is rejected if the user is not the author of the message.

**What application are you using?**
OJS 3.6.0

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib https://github.com/pkp/pkp-lib/pull/12313

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib https://github.com/pkp/pkp-lib/pull/12313

Vitaliy-1:
Thanks, @Hafsa-Naeem. The review is done!

Hafsa-Naeem:
@Vitaliy-1 Ready for review
pkp-lib https://github.com/pkp/pkp-lib/pull/12313
OJS: https://github.com/pkp/ojs/pull/5328

Hafsa-Naeem:
@Vitaliy-1 Reminder to review this PR. Thanks.

Vitaliy-1:
@Hafsa-Naeem, thanks, approved!

PR #12313, opened 2026-02-09
pkp/pkp-lib#12278 Enforced 1-hour edit window for discussion headnotes

Hafsa-Naeem:
for #12278

### DIFF
--- a/api/v1/submissions/tasks/formRequests/EditTask.php
+++ b/api/v1/submissions/tasks/formRequests/EditTask.php
@@ -78,19 +78,35 @@ class EditTask extends FormRequest
                 'sometimes',
                 'string',
                 function (string $attribute, string $value, Closure $fail) {
-                    $headnote = $this->task->notes()->where('is_headnote', true)->first();
+                    if (!$this->task) {
+                        return;
+                    }
+
+                    $currentUser = Application::get()->getRequest()?->getUser();
+                    if (!$currentUser) {
+                        return;
+                    }
+
+                    if (
+                        $currentUser->hasRole([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR], $this->submission->getData('contextId'))
+                        || $currentUser->hasRole([Role::ROLE_ID_SITE_ADMIN], Application::CONTEXT_SITE)
+                    ) {
+                        return;
+                    }
+
+                    $headnote = $this->task->notes()->withHeadnote()->first();
 
                     if (!$headnote) {
                         return;
                     }
 
-                    if ($headnote->userId != Application::get()->getRequest()?->getUser()?->getId()) {
-                        $fail(__('api.403.forbidden'));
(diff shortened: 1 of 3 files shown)


Vitaliy-1 (commented):
Thanks, I think we should also add a validation check to the `EditTask` form request

api/v1/submissions/tasks/formRequests/EditTask.php
```diff
@@ -74,7 +74,26 @@ public function rules(): array
                 Rule::date()->format('Y-m-d'),
                 'after_or_equal:today',
             ],
-            EditorialTask::ATTRIBUTE_HEADNOTE => ['sometimes', 'string'],
+            EditorialTask::ATTRIBUTE_HEADNOTE => [
+                'sometimes',
+                'string',
+                function (string $attribute, string $value, Closure $fail) {
+                    $headnote = $this->task->notes()->where('is_headnote', true)->first();
```
Vitaliy-1: Could we hide the call to the column behind the scope?

api/v1/submissions/tasks/formRequests/EditTask.php
```diff
@@ -74,7 +74,26 @@ public function rules(): array
...
+
+                    if (!$headnote) {
+                        return;
+                    }
+
+                    if ($headnote->userId != Application::get()->getRequest()?->getUser()?->getId()) {
+                        $fail(__('api.403.forbidden'));
```
Vitaliy-1: It needs localized validation message

api/v1/submissions/tasks/formRequests/EditTask.php
```diff
@@ -74,7 +74,26 @@ public function rules(): array
...
+                        return;
+                    }
+
+                    if (time() - strtotime($headnote->dateCreated) >= 3600) {
+                        $fail(__('api.403.forbidden'));
```
Vitaliy-1: The same

classes/security/authorization/NoteAccessPolicy.php
```diff
@@ -90,9 +90,19 @@ public function effect(): int
...
         // Notes can only be edited by their original creators
-        if ($this->_accessMode === self::NOTE_ACCESS_WRITE
-                && $note->userId != $this->_request->getUser()->getId()) {
-            return AuthorizationPolicy::AUTHORIZATION_DENY;
+        if ($this->_accessMode === self::NOTE_ACCESS_WRITE) {
+            if ($note->userId != $this->_request->getUser()->getId()) {
+                return AuthorizationPolicy::AUTHORIZATION_DENY;
+            }
+
+            // 1 hour edit window for the discussion head note
```
Vitaliy-1: Hmm, is the logic here correct?
What about notes, e.g., that aren't a headnote?
Maybe it's better to permit under certain parameters and deny in all other cases?

PR #12316, opened 2026-02-09
Fix N+1 query performance issue in Validation::getAdministrationLevel

mariuszsienkiewicz:
The `getAdministrationLevel` method was suffering from a classic N+1 query problem. Previously, the code retrieved a list of `UserGroup` objects and then iterated over them. Inside the foreach loop, the code accessed `$userGroup->userUserGroups`. Since this relationship was not preloaded, the ORM was forced to execute a separate SQL query against the database for every single user group iteration.

**Solution**

I refactored the query to use Eager Loading via the ->with() method. This ensures that the userUserGroups relationship is retrieved in a single accompanying query at the very beginning.

For example, this drastically improves submission list loading performance (reducing it from **19s to 3s** in my test case)

### DIFF
--- a/classes/security/Validation.php
+++ b/classes/security/Validation.php
@@ -504,6 +504,10 @@ class Validation
 
         // single query to fetch user groups assigned to either user
         $allUserGroups = UserGroup::query()
+            ->with(['userUserGroups' => function ($query) use ($administratorUserId, $administeredUserId) {
+                $query->withUserIds([$administratorUserId, $administeredUserId])
+                    ->withActive();
+            }])
             ->whereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
                 $query->withUserIds([$administratorUserId, $administeredUserId])
                     ->withActive();
--- a/classes/security/Validation.php
+++ b/classes/security/Validation.php
@@ -504,11 +504,7 @@ class Validation
 
         // single query to fetch user groups assigned to either user
         $allUserGroups = UserGroup::query()
-            ->with(['userUserGroups' => function ($query) use ($administratorUserId, $administeredUserId) {
-                $query->withUserIds([$administratorUserId, $administeredUserId])
-                    ->withActive();
-            }])
-            ->whereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
+            ->withWhereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
                 $query->withUserIds([$administratorUserId, $administeredUserId])
                     ->withActive();
             })


asmecher:
Thanks, @mariuszsienkiewicz!

@touhidurabir, would you mind reviewing this one?

classes/security/Validation.php
```diff
@@ -504,6 +504,10 @@ public static function getAdministrationLevel(
 
         // single query to fetch user groups assigned to either user
         $allUserGroups = UserGroup::query()
+            ->with(['userUserGroups' => function ($query) use ($administratorUserId, $administeredUserId) {
+                $query->withUserIds([$administratorUserId, $administeredUserId])
+                    ->withActive();
+            }])
             ->whereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
                 $query->withUserIds([$administratorUserId, $administeredUserId])
                     ->withActive();
```
touhidurabir: Here we are doing eager load with constrain + data filter based on relation with same logic so we can use the `withWhereHas` like 
```php
->withWhereHas(['userUserGroups' => function ($query) use ($administratorUserId, $administeredUserId) {
    $query->withUserIds([$administratorUserId, $administeredUserId])
        ->withActive();
}])
```
and don't need two separate `with` and `whereHas` if the eager load and filter logic is identical which the case here. Laravel doc that demonstrate this at https://laravel.com/docs/11.x/eloquent-relationships#constraining-eager-loads-with-relationship-existence

touhidurabir (changes requested):
Hi @mariuszsienkiewicz , great catch . just one change to look for

classes/security/Validation.php
mariuszsienkiewicz: Thanks for the suggestion, @touhidurabir, updated the query to use ‎`withWhereHas` as you proposed :)

touhidurabir (approved):

asmecher (approved):

asmecher:
Merged and cherry-picked back to `stable-3_5_0` -- thanks, @mariuszsienkiewicz (and @touhidurabir)!

darius-v:
Hello, is it safe to apply change only for getAdministrationLevel method without raising whole OJS application version from 3.5.0.3 to 3.5.0.4 . Asking cause applying the fix only for this method would be much faster. And then later we could raise whole version if needed.

I mean replacing
```
$allUserGroups = UserGroup::query()
            ->whereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
                $query->withUserIds([$administratorUserId, $administeredUserId])
                    ->withActive();
            })
            ->get();
```

with

```
$allUserGroups = UserGroup::query()            
            ->withWhereHas('userUserGroups', function ($query) use ($administratorUserId, $administeredUserId) {
                $query->withUserIds([$administratorUserId, $administeredUserId])
                    ->withActive();
            })
            ->get();
```

touhidurabir:
@darius-v thats a one line change (e.g. see at https://github.com/pkp/pkp-lib/pull/12316/changes) so I dont see any issue .

darius-v:
Since it is related, asking here:
I can apply safely changes in those 2 files also without raising version from 3.5.0.3 to 3.5.0.4

image
?

https://github.com/pkp/pkp-lib/commit/2e01f530c5bee2498747f25c4b97e0692409367c#diff-fe4b90f23fff57b4a5b3882e0e211d39e7858658b18ff4166b821f2aca99d224

asmecher:
@darius-v, I suspect so -- but we don't provide support for changes like this, so you'd have to try it and see.

Issue #12319, opened 2026-02-09
Ensure correct elements returned for review and review summary endpoints


ewhanson:
**Description:**
Some items in the returned results from review and review summary endpoints require refinement. They are:

- `PublicationPeerReviewResource` returns some fields that are unnecessary for a public endpoint, e.g. `dateAssigned`, `dateConfirmed`, `declined`.
  - `dateCompleted` is required. 
- Peer review status (complete, in progress, not started) is needed for summary along with relevant date, e.g. "In progress since Dec 12, 2025"
- `authorResponse` for a review round should be singular rather than `authorResponses` and return a singular object rather than an array.
- An `isPublic` attribute was added to the `AuthorResponse` model.

#### Review status

The following has been added to the returned data. All statuses and dates start from an assumption of excluding declined and canceled reviews.
- `status`:
	-  `notStarted`: No review assignments are present, or no present review assignments have been accepted yet
	- `inProgress`: At least one review assignment has been accepted
	- `complete`: All review assignments in round that were not declined or canceled have been completed
-`dateStarted`: Oldest date a reviewer was assigned (helpful fallback)
- `dateInProgress`: Oldest date a reviewer has accepted to complete a review (when in progress starts)
- - `dateComplete`: When the most recent review assignment has been completed if all assignments have been completed

### Data shape summary

Peer review data shape example:
```json
{
  "publicationId": 11,
  "datePublished": "2026-02-11",
  "reviewRounds": [
    {
      "displayText": "Version of Record 1.0",
      "roundId": 8,
      "originalPublicationId": 11,
      "status": "complete",
      "dateStarted": "2026-02-10 06:27:10",
      "dateInProgress": "2026-02-10 06:27:23",
      "dateCompleted": "2026-02-10 06:27:36",
      "reviews": [
        {
          "id": 15,
          "reviewerId": 9,
          "reviewerFullName": "Aisla McCrae",
          "reviewerAffiliation": "University of Manitoba",
          "dateCompleted": "2026-02-10 06:27:26",
          "isReviewOpen": true,
          "reviewerRecommendationDisplayText": "Revisions Required",
          "reviewerRecommendationId": 2,
          "reviewerRecommendationTypeId": 2,
          "reviewerRecommendationTypeLabel": "Not Approved",
          "reviewForm": null,
          "reviewerComments": [
            "<p>Here are my review comments</p>"
          ]
        },
        {
          "id": 16,
          "reviewerId": 10,
[...]

PR #12320, opened 2026-02-10
pkp/pkp-lib#12319 Ensure correct elements returned for review and review summary endpoints

ewhanson:


api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -40,7 +45,23 @@ class PublicationPeerReviewResource extends JsonResource
...
+    private ?Collection $reviewFormsCache = null;
+
+    /** @var Collection<int, ReviewFormElement>|null Caches review form elements to avoid redundant fetches */
+    private ?Collection $reviewFormElementsCache = null;
+
+    /** @var Collection<int, ReviewFormResponse>|null Caches review form responses to avoid redundant fetches */
+    private ?Collection $reviewFormResponsesCache = null;
+
+    /** @var Collection<int, array<string>>|null Caches reviewer comments to avoid redundant fetches */
+    private ?Collection $reviewerCommentsCache = null;
```
taslangraham, line 62: Good idea to cache these!

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -103,10 +130,16 @@ private function getPublicationPeerReview(Publication $publication): Enumerable
...
 
+            $publicStatus = $reviewRound->getPublicReviewStatus($assignments);
+
             $roundsData->add([
                 'displayText' => $roundDisplayText,
                 'roundId' => $reviewRound->getData('id'),
                 'originalPublicationId' => $reviewRound->getPublicationId(),
+                'status' => $publicStatus['status']->value,
+                'dateStarted' => $publicStatus['dateStarted'],
+                'dateInProgress' => $publicStatus['dateInProgress'],
```
taslangraham, line 141: What is the difference between `dateStarted` and `dateInProgress`?

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -150,54 +186,139 @@ private function getReviewAssignmentPeerReviews(Enumerable $assignments, Context
...
+
+        // Fetch all review forms
+        $this->reviewFormsCache = collect();
+        foreach ($reviewFormIds as $formId) {
+            $form = $reviewFormDao->getById($formId, Application::getContextAssocType(), $context->getId());
+
+            if ($form) {
+                $this->reviewFormsCache->put($formId, $form);
+            }
+        }
```
taslangraham, line 243: Since we are trying to reduce trips to the DB, further improvement could be to introduce a new method like `getByIds` or `getByFormIds` on the `ReviewFormDAO` that allows fetching of multiple forms at once, thus reducing the number of DB calls made here. What do you think?

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -150,54 +186,139 @@ private function getReviewAssignmentPeerReviews(Enumerable $assignments, Context
...
+        foreach ($reviewFormIds as $formId) {
+            $elements = $reviewFormElementDao->getByReviewFormId($formId);
+            $elementsList = collect();
+
+            while ($element = $elements->next()) {
+                $elementsList->push($element);
+            }
+
+            $this->reviewFormElementsCache->put($formId, $elementsList);
+        }
```
taslangraham, line 256: Similar comment to the one above, https://github.com/pkp/pkp-lib/pull/12320/changes#r2793512686, but for `ReviewFormElementDAO`. So the new method would be named something like `getByReviewFormIds`.

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -150,54 +186,139 @@ private function getReviewAssignmentPeerReviews(Enumerable $assignments, Context
...
+                /** @var Collection<SubmissionComment> $commentsList */
+                $commentsList = collect();
+
+                /** @var SubmissionComment $comment */
+                while ($comment = $comments->next()) {
+                    $commentsList->push($comment->getComments());
+                }
+                $this->reviewerCommentsCache->put($assignment->getId(), $commentsList->all());
+            }
+        }
```
taslangraham, line 287: Similar to previous comments (https://github.com/pkp/pkp-lib/pull/12320/changes#r2793512686 & https://github.com/pkp/pkp-lib/pull/12320/changes#r2793526450), I think we can improve this by adding a new method to the `SubmissionCommentDAO` that allows fetching of multiple comments at once.

api/v1/reviews/resources/ReviewRoundAuthorResponseResource.php
```diff
@@ -25,26 +25,20 @@ class ReviewRoundAuthorResponseResource extends JsonResource
...
         $associatedAuthors = $response->associatedAuthors;
         return [
+            'id' => $response->id,
             'reviewRoundId' => $response->reviewRoundId,
             'response' => $response->authorResponse,
-            'id' => $response->id,
-            'submittedByUser' => [
-                'id' => $user->getId(),
-                'fullName' => $user->getFullName(),
-            ],
```
taslangraham, line 42: Not sure if this is something you are already looking into, but the ui-library uses the  `submittedByUser` property.  See https://github.com/search?q=repo%3Apkp%2Fui-library%20submittedByUser&type=code. If we are removing it here, then the ui-library will need to be updated.

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
ewhanson: Good idea. I had started down a road of some more significant refactoring to reduce DB calls, but decided to shelve it for now to keep things in scope. A good easy win here, though.

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
ewhanson: I'm not sure if it will ever really come up, but `dateStarted` is basically "when was the first reviewer reached out to" and `dateInProgress` reflects when the first reviewer accepts to take on the review. I'll make sure this is documented within the code (and issue) as well as double check that `$reviewAssignment->getDateConfirmed()` is the right value to be getting.

api/v1/reviews/resources/ReviewRoundAuthorResponseResource.php
ewhanson: I did not know that, thanks! I'll get that changed there as well.

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
ewhanson: On second thought, I don't think this type of database fetching code change this late is a good idea. None of the legacy DAOs have the equivalent of a `getMany` or `getByIds` so I think this needs addressing before release, I'm hesitant to do it right before the ORE deadline.

Issue #12322, opened 2026-02-10
Re-add notifications to tasks and discussions


Vitaliy-1:
**Description:**
Re-add missing functionality to notify users regarding the task/discussion related activity.

**Solution:** 
The ability to send system and email notifications to subscribed users should be added to the API when
1. New task/discussion is created.
2. New participants are added when editing task/discussion
3. When new note/reply is posted

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PR: https://github.com/pkp/pkp-lib/pull/12323
tests OJS: https://github.com/pkp/ojs/pull/5331

Vitaliy-1:
@touhidurabir, do you have time to look at the PR?

touhidurabir:
Hi @Vitaliy-1 , review added at https://github.com/pkp/pkp-lib/pull/12323#pullrequestreview-3783656469

Vitaliy-1:
PR: https://github.com/pkp/pkp-lib/pull/12381
tests: https://github.com/pkp/ojs/pull/5366

(The fix for the creation of a new task on the copyediting or production stage and the error during submission deletion)

PR #12323, opened 2026-02-10
pkp/pkp-lib#12322 Re-add task related notifications

Vitaliy-1:


controllers/grid/queries/traits/StageMailable.php
```diff
@@ -26,21 +26,21 @@
...
 {
-    abstract public function getStageId();
-
     /**
      * @return Mailable which corresponds to the given workflow stage
      */
-    protected function getStageMailable(Context $context, Submission $submission): Mailable
+    protected function getStageMailable(Context $context, Submission $submission, ?int $stageId = null): Mailable
     {
+        $stageId = $stageId ?? $this->getStageId();
```
touhidurabir, line 34: Here in the method we added a new method param `$stageId` that can be null but later trying to get it from method `getStageId` if not passed . But above removed the abstract method `getStageId` which does not force anymore . 

So in case the class using this trait does not have a method `getStageId` and value of param `stageId` is not passed or passed null, the call of non existence method `getStageId` will cause exception . 

quick check point that currently(only checked for OJS) have these 2 following class have this trait being used 
1. PKP\controllers\grid\users\stageParticipant\StageParticipantGridHandler
2. PKP\controllers\grid\users\stageParticipant\form\PKPStageParticipantNotifyForm

and both of them have the implementation of method `getStageId`

But some other class may not have it . so is this an intended changed and we are sure that any  class using this trait will always have the `getStageId` method implementation ?

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -357,6 +374,12 @@ public function editTask(EditTask $illuminateRequest): JsonResponse
             ], Response::HTTP_CONFLICT);
         }
 
+        $newParticipantIds = $editTask->participants->pluck('userId')->toArray();
```
touhidurabir: If I am not mistaken and remember correctly, after an update of relational data (in this case `participants`) on update does not update the relational collection on previously fetched model instance as laravel lazy load the relational data and cache the collection on model instance . so if there is some update of participants also, it probably will not reflect on the previously fetched model instance as relational data unless laravel has fixed this behaviour in newer version . 

so to be safe, we can do something like 
```php
// will refresh the model along with reload the participants relation
$editTask = $editTask->fresh('participants');
```
or 
```php
$editTask->refresh();   // refreshes attributes only
$editTask->unsetRelation('participants'); // clear the cached collection
$newParticipantIds = $editTask->participants->pluck('userId')->toArray();
```

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -898,4 +924,86 @@ public function getReviewers(Submission $submission, int $stageId, array $curren
...
+                $editorialTask->id,
+                Notification::NOTIFICATION_LEVEL_TASK
+            );
+
+            // Check if the user is unsubscribed
+            $notificationSubscriptionSettings = $notificationSubscriptionSettingsDao->getNotificationSubscriptionSettings(
+                NotificationSubscriptionSettingsDAO::BLOCKED_EMAIL_NOTIFICATION_KEY,
+                $user->getId(),
+                $context->getId()
+            );
```
touhidurabir, line 997: Here I think we need also check the `BLOCKED_NOTIFICATION_KEY` as if that is disable, `BLOCKED_EMAIL_NOTIFICATION_KEY` will have no impact . something similar we have faced , see at https://github.com/pkp/pkp-lib/issues/10496#issuecomment-3823325131 . 

I see that we also check if the notification we get is null in below check and perhaps that would mitigate the above concern but should we just check proper notification permission and then create the notification instance from the notification manager ?

api/v1/submissions/tasks/EditorialTaskController.php
Vitaliy-1: Thanks!

controllers/grid/queries/traits/StageMailable.php
Vitaliy-1: abstract method was just a trick  and you are right, only those 2 classes has it. Let's leave it for as is for now. I don't think other classes would need that trait

Issue #9921, opened 2024-04-29
In OJS 3.4, only default reminder template can be used when sending a reminder to a reviewer manually.


pilasou:
**Describe the bug**
As an editor, I can create multiple templates for manual reviewer reminder as shown below: 
![image](https://github.com/pkp/pkp-lib/assets/19291880/b22597c8-c020-4e0f-b3ce-b3650e60f164)

However, the new template created cannot be used when trying to send a reminder to the reviewer: 
![image](https://github.com/pkp/pkp-lib/assets/19291880/6e25ea65-f646-4d39-a838-8dbf07774da1)

**To Reproduce**
1. As an editor, go to email management page (Parameter, Workflow, Emails) and search for Reminder
2. Clic Edit on the Review Reminder template.
3. In the new window, clic on Add template, then add all necessary information and Clic Save
4. Open a submission for which at least one of the reviewer is late. 
5. Clic on Send Reminder
6. In the new window, only the default email is displayed, the new one cannot be choosen/selected.

**What application are you using?**
 OJS 3.4.0.5

**Additional information**
Correcting this would address a problem that has been reported on PKP forum, see : https://forum.pkp.sfu.ca/t/problem-with-workflow-emails-templates-review-remind/71545 or https://forum.pkp.sfu.ca/t/different-templates-for-automated-review-request-reminder-and-automated-review-reminder/4540 or more recently: https://forum.pkp.sfu.ca/t/same-email-template-for-review-reminder-and-review-request-reminder/71547

pilasou:
Thanks @asmecher! This would definetly address this issue (use 3.4 new editorial decision workflow for reviewer reminder).

pilasou:
@asmecher We will work on this issue. Our idea is to simply add a drop down list the the Reminder window so that editors can select the email appropriate to the reminder context (invitation or recommandation). We will use the reviewer invitation window as a model: 

<img width="1610" height="353" alt="Image" src="https://github.com/user-attachments/assets/ed0373a9-d357-4c96-9335-9d76d6d9ab28" />

asmecher:
Thanks, @nibou230! 

@defstat, would you be able to review the two PRs for this?

jardakotesovec:
@pilasou Hi, could you list the PRs link in the top post on this issue? THanks!

pilasou:
@jardakotesovec Hi! I have just added the 2 PRs to the issue.

defstat:
[...]

PR #12331, opened 2026-02-12
pkp/pkp-lib#9921 Dropdown to select the email template when sending a reminder

nibou230:
For #9921

Added a dropdown in the review reminder form.

There is a new JS handler to make the template modification possible by the user.


js/controllers/grid/users/reviewer/form/ReviewReminderFormHandler.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * @file js/controllers/grid/users/reviewer/form/ReviewReminderFormHandler.js
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
```
defstat: This should be 2026 for new files?

js/controllers/grid/users/reviewer/form/ReviewReminderFormHandler.js
```diff
@@ -0,0 +1,102 @@
...
+	/**
+	 * @constructor
+	 * @extends $.pkp.controllers.form.AjaxFormHandler
+	 *
+	 * @param {jQueryObject} $form
+	 * @param {Object} options
+	 */
+	$.pkp.controllers.grid.users.reviewer.form.ReviewReminderFormHandler =
+		function ($form, options) {
+		this.parent($form, options);
```
defstat, line 32: should that be indented?

controllers/grid/users/reviewer/form/ReviewReminderForm.php
```diff
@@ -111,6 +114,26 @@ public function fetch($request, $template = null, $display = false)
...
             'passwordResetUrl' => __('common.url'),
         ]);
+        // Provide templates for the <select>
+        $context = $request->getContext();
+        $reviewAssignment = $this->getReviewAssignment();
+        $submission = Repo::submission()->get($reviewAssignment->getSubmissionId());
+        $mailable = new ReviewRemind($context, $submission, $reviewAssignment);
+        $data = $mailable->getData(Locale::getLocale());
+        $templateKey = $mailable::getEmailTemplateKey();
+        $defaultTemplate = Repo::emailTemplate()->getByKey($context->getId(), $templateKey);
```
defstat: Should this be checked for `null` value?

controllers/grid/users/reviewer/form/ReviewReminderForm.php
```diff
@@ -142,7 +166,8 @@ public function execute(...$functionArgs)
 
         // Create ReviewRemind email and populate with data
         $mailable = new ReviewRemind($context, $submission, $reviewAssignment);
-        $template = Repo::emailTemplate()->getByKey($context->getId(), $mailable::getEmailTemplateKey());
+        $templateKey = $this->getData('template') ?: $mailable::getEmailTemplateKey();
```
defstat: I think there is an ambiguity regarding what we are calling `template` and what `templateKey` - Is there a way to standardize that?

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -969,6 +971,42 @@ public function sendReminder($args, $request)
...
+     * Fetch the compiled body for a review reminder template (AJAX).
+     *
+     * @param array $args
+     * @param PKPRequest $request
+     *
+     * @return JSONMessage JSON object
+     */
+    public function fetchReviewReminderTemplateBody(array $args, PKPRequest $request): JSONMessage
+    {
+        $templateKey = $request->getUserVar('template');
```
defstat, line 996: Here we may need to check against the set of "allowed" template keys, to avoid the user adding "any key" in the request variable.

controllers/grid/users/reviewer/form/ReviewReminderForm.php
```diff
@@ -111,6 +114,26 @@ public function fetch($request, $template = null, $display = false)
         $templateMgr->assign('emailVariables', [
             'passwordResetUrl' => __('common.url'),
         ]);
+        // Provide templates for the <select>
+        $context = $request->getContext();
```
defstat: It seems that we are duplicating parts of the same code to different functions (initData, fetch, execute etc) 

Is there something preventing us from adding the duplicated code to a method?

defstat (changes requested):
Thanks @nibou230! 

Just some comments to consider here.

nibou230:
Hello @defstat, I addressed all your review recommendations. Let me know if there's something else.

defstat:
> Hello @defstat, I addressed all your review recommendations. Let me know if there's something else.

Thanks @nibou230! I propose we wait for @jardakotesovec to comment on the `pkp.min.js` changing before merging it.

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -986,6 +988,50 @@ public function sendReminder($args, $request)
...
     }
 
+    /**
+     * Fetch the compiled body for a review reminder template (AJAX).
+     *
+     * @param array $args
+     * @param PKPRequest $request
+     *
+     * @return JSONMessage JSON object
+     */
```
asmecher, line 993: This is just stylistic, but when there are typehints and the self-documentation comments don't add anything to them, it's best to just leave the typehints:
```suggestion
    /**
     * Fetch the compiled body for a review reminder template (AJAX).
     */
```

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
asmecher: ^ Agreed, I think this is important.

controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -0,0 +1,13 @@
...
+ * @file controllers/grid/users/reviewer/PKPReviewerGridHandler.php
+ *
+ * Copyright (c) 2014-2026 Simon Fraser University
+ * Copyright (c) 2000-2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ * 
+ * Make PKPReviewerGridHandler resolvable by PKPComponentRouter.
+ */
+
+require_once(PKP_LIB_PATH . '/classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php');
```
asmecher: `PKPReviewerGridHandler` is extended in the applications that have a review workflow (OJS and OMP) by `ReviewerGridHandler`. The `pkp-lib` version is not intended for direct access, which is why it's in `classes` rather than `controllers`. (It should probably be marked `abstract`, but it was added before that was available in PHP.)

It would be better to use the application-level one, and remove this effective alias.

asmecher (changes requested):
Thanks, @nibou230, I have a couple additional comments!

controllers/grid/users/reviewer/PKPReviewerGridHandler.php
nibou230: I removed the alias and used the ReviewerGridHandler instead. I wasn't aware that I could simply use this one!

asmecher:
Manually merged

Issue #12347, opened 2026-02-16
[Editorial Workflow] | File revision history has confusing/wrong event log of previous file at new revision upload


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

After a submission submitted, it is possible to upload a revision of submission file from the editorial workflow section . However the revision log history accessible from the `More Information` of specific file action menu presents a confusing/wrong log history . 

<img width="1910" height="656" alt="Image" src="https://github.com/user-attachments/assets/a380d82a-5881-4066-b471-b3a023316ef5" />

Without any revision upload, the log history is ok and well preserved showing proper entry and download link of the last uploaded file as 

<img width="2716" height="666" alt="Image" src="https://github.com/user-attachments/assets/11a7ee40-5564-4338-ba54-5cff15d36a9a" />

However once a new revision is uploaded for that specific file, the new entry still pointing to the previous file instead of the new file that just uploaded as revision . This is confusing (or seems wrong). 

<img width="2800" height="670" alt="Image" src="https://github.com/user-attachments/assets/fb912508-7104-48f7-b001-fe5142eb385f" />

What should be the one in this case is that the new log entries should point to the newly upload file not the previous one for which the revision has been uploaded 

### Steps to Reproduce

1. Login with Editorial role
2. Navigate to workflow dashboard
3. Click on any submission
4. First check the file log history from the file's `More Information` action menu
5. Now upload a new revision of that file
6. See that the new logs are still pointing to the old file for which revision has been just added, not the one that just uploaded

### Expected Result

the revision log history should properly points to the updated files for each actions that been performed . 

### Actual Result

The revision log history still pointing to the previous file instead of the new file that just uploaded as revision . This is confusing (or seems wrong).

### Application Version

OJS/OMP/OPS 3.5/main 

### Additional Information

This seems like a very trivial fix where for the event log store, we should passed newly created submission file instance instead of previous one at https://github.com/pkp/pkp-lib/blob/main/classes/submissionFile/Repository.php#L372

Also NOTE that, it is actually a regression which has been introduced in 3.4 and any 3.4 pre data is ok . 

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12350
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5349 [TEST ONLY]
[...]

PR #12350, opened 2026-02-16
pkp/pkp-lib#12347 Added missing log entry for submission at submission file edit

touhidurabir:
for pkp/pkp-lib#12347 .

### DIFF
--- a/classes/submissionFile/Repository.php
+++ b/classes/submissionFile/Repository.php
@@ -385,17 +385,18 @@ abstract class Repository
 
         $submission = Repo::submission()->get($submissionFile->getData('submissionId'));
 
-        Repo::eventLog()->newDataObject(array_merge(
+        $submissionLogEntry = Repo::eventLog()->newDataObject(array_merge(
             $logData,
             [
                 'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION,
                 'assocId' => $submission->getId(),
                 'eventType' => $newFileUploaded ? SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_REVISION_UPLOAD : SubmissionFileEventLogEntry::SUBMISSION_LOG_FILE_EDIT,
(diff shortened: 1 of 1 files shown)


classes/submissionFile/Repository.php
Vitaliy-1, line 1: Makes sense to fix inserting a log entry like here https://github.com/pkp/pkp-lib/pull/12352/changes#r2845592533

Vitaliy-1 (approved):

classes/submissionFile/Repository.php
touhidurabir: @Vitaliy-1 do you think we should add this fix only for 3.5 aside from migration and fix of updated submission file .  that is only fix the missing log entry of submission ?

Vitaliy-1: Sorry, I don't completely understand. That's only for missing log entry.

touhidurabir: > Sorry, I don't completely understand. That's only for missing log entry.

Yes, thats what I mean . thanks for the confirmation , I will update the 3.5 PR with fix of missing log entry only .

touhidurabir:
@Vitaliy-1 can you quick check this PR which just add the missing log ? if all ok, merge it as I cant without 2 approval and does not seems like this one need another review .

classes/submissionFile/Repository.php
Vitaliy-1: @touhidurabir, sorry, the change regarding logging `$newSubmissionFile` is also necessary

touhidurabir: > @touhidurabir, sorry, the change regarding logging `$newSubmissionFile` is also necessary

@Vitaliy-1 if we added the same fix as main where we put the `$newSubmissionFile` as file event, does not that create a asymmetry for the files history ? that would require to include the migration(to fox the older data) which is the reason I intentionally skipped it .

Vitaliy-1: Agree!

Issue #12349, opened 2026-02-16
[Editorial Workflow] | Galley file allow unintended revision upload of past files


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

For the galley upload, we only have `Change File` option but it still allow upload revision of previous upload file . this is confusing/wrong as galley at a time only points to one single file and uploading a new one will/should be counted as the revision of that galley file . 

<img width="3044" height="972" alt="Image" src="https://github.com/user-attachments/assets/1a89520d-f185-4664-945c-f2e80c38af91" />

### Steps to Reproduce

1. Login with Editorial Role
2. Select a submission Navigate to `Galley` tab
3. Create a new one with a file
4. Now perform `Change File` action on that galley
5. See that a revision of a pervious upload file can be uploaded

### Expected Result

Should be able only upload a new file for galley, not revision of any previous file 

### Actual Result

Possible to upload the revision of a previous file . 

### Application Version

OJS 3.5/main

#### Stable-3.5.0
~~[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12351~~
~~[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5350 [TEST ONLY]~~
~~[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2239 [TEST ONLY]~~
~~[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1207 [TEST ONLY]~~
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12441
[ui-library](https://github.com/pkp/ui-library) --> https://github.com/pkp/ui-library/pull/825
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5408
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1225
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2260 [TEST ONLY]

#### main
~~[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12353~~
~~[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5352 [TEST ONLY]~~
[...]

PR #12353, opened 2026-02-17
pkp/pkp-lib#12349 Restrict galley upload only to new file and no revision

touhidurabir:
for pkp/pkp-lib#12349 .


touhidurabir:
cherry picked from https://github.com/pkp/pkp-lib/pull/12351

Issue #12060, opened 2025-11-25
[Discussion Boards] | Any Workflow Section - Cannot reorder discussions


busspatrick:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When attempting to reorder the list of discussions in any workflow section, the page freezes, and a PHP stack trace is generated. It appears the AJAX callback is trying to update the database but fails. This appears to be the relevant SQL error:

```
Illuminate\\Database\\QueryException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause' (Connection: mysql, SQL: select * from `queries` where `id` = <id> and `assoc_type` = <assoc_type> and `assoc_id` = <assoc_id> limit 1) in <project_root_directory>/ojs/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php:825\nStack trace:\n#0
```

### Steps to Reproduce

1. Create a new submission.
2. Create at least two discussions.
3. Click the 'Order' button.
4. The widget allows you to reorder the items.
5. Click the 'Done' button.
6. The page freezes.

### Expected Result

The new discussion order should be saved.

### Actual Result

Page freezes.

### Application Version

OJS 3.5.0-1

### Logs

```
[...]

PR #12358, opened 2026-02-17
pkp/pkp-lib#12060 fix discussion ordering

kaitlinnewson:


controllers/grid/queries/QueriesGridHandler.php
blesildaramirez, line 336: It seems that we need to set `orderBy` here so that, after saving the order, the list is sorted by `seq`.
```suggestion
            ->orderBy('seq')
            ->lazy();
```

kaitlinnewson:
Thanks @blesildaramirez, good catch! Clearly I did not test that fix thoroughly enough - should be fixed now.

Issue #12307, opened 2026-02-07
[Author Response] | Include author response as part of review round - Part 2 enhancements


Tribunal33:
**Description:**
- Include specific context details that have been discovered that surround this developer task.
- Link related tickets #12048 (part 1)
During testing there were a few things that were noticed as being a bit odd for the workflow.  These are enhancements to existing functionality that went beyond the scope of the original ticket.  

1. if there were no reviews completed the option to request an author response is enabled and status is changed. The status changes and button should only enable if there is a review for the author to respond to. Currently an editor can click the response button and have an author respond but even though it won't be displayed.

<img width="951" height="250" alt="Image" src="https://github.com/user-attachments/assets/8a5efd47-0f87-4ea3-b602-686db1f70430" />

2. One of the triggers is if all of the reviewer have completed their reviews. If all of them accept then the email template should be changed to cover the situation a bit better (ie Submission accepted template). Or we could update the template to cover a more generic situation.
a. Currently it uses the request revisions template and says "we would like to encourage you to submit revisions that address the reviewers' comments. An editor will review these revisions and if they address the concerns adequately, your submission may be accepted for publication". Which doesn't make sense if either the Accept button or all reviews are recommending to be accepted.
3. Not sure if this is expected or a feature not a bug but the contributor who does not have a user account in the system can't access the Request Response until they have created an account.  (if this is needed we can figure out the A/C)

Acceptance Criteria:
- [ ]  If a review round does not have the necessary amount(all reviews completed, or at least the minimum number of completed reviews required per submission, by the Journal, is completed) of completed reviews the Response Request button is disabled and status is "Awaiting Reviews" even after the Accept main action workflow is completed or Request revisions Button is completed.  Reason: Not sure if there should be a response from an author when there is nothing to respond to.
~- [ ] When all reviewers complete the review with Recommendation : Accept then the Request Response template would be the 'Submission accepted' template instead of Request Revisions template.~
- [ ] Do not use the Revision Requested email template when sending an email to request author's response. Instead, introduce a dedicated email template specifically created for requesting a review response from authors.

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
**Part 1 - Disable button if no reviewers completed a review.  Currently it is enabled even when there are no reviews complete**
1. Prereqs: Must have a submission in the review stage with reviewers added but no reviews completed.
2. Login with any editor role with access
3. Find submission that is in review stage in prereqs.
4. Select Accept from main action
5. Record decision
[...]

PR #12364, opened 2026-02-18
pkp/pkp-lib#12307 Author review response enhancements

taslangraham:
For pkp/pkp-lib#12307


api/v1/reviews/PKPReviewController.php
```diff
@@ -882,6 +883,40 @@ public function requestAuthorResponse(RequestAuthorResponse $illuminateRequest):
 
         /** @var ReviewRound $reviewRound */
         $reviewRound = $validated['reviewRound'];
+
+        $assignments = Repo::reviewAssignment()->getCollector()
```
ewhanson, line 887: I'm not sure if it's better to include it here or below, but this should also exclude canceled or declined reviews, otherwise they'll be considered in the `$areAllReviewsCompleted` check below.

Issue #11825, opened 2025-09-12
[Frontend] Tasks & Discussions in Submission Workflow Page


blesildaramirez:
## Description:
Implement UI for managing tasks and discussions in the submission workflow page.
This feature is behind the config flag: `enable_new_discussions = On`

## Tasks & Discussions
Headers to use for each stages: 🚀 
- Submission: **Desk Review Tasks & Discussions**
- Internal Review: **Review Tasks & Discussions**
- External Review: **Review Tasks & Discussions**
- Copyediting: **Copyediting Tasks & Discussions**
- Production: **Production Tasks & Discussions**

### Table/List
- [x] Group entries by: **Yet to Begin, In Progress, Closed**, sort by **oldest to newest**.
- [ ] Columns ⚠️ 
  - [x] Name: Type (Task/Discussion), clickable title, **Created By:** for discussion, **Task Owner:** for tasks (the participant responsible to complete the task) 🚀 
  - [x] Activity: Most recent activity log or thread activity logs 🚀 
  - [x] Due Date: Tasks only 
  - [x] Started: Checkbox for tasks (confirm modal to start, disabled if started/user doesn't have access writes), hidden for discussions 
  - [x] Closed: Checkbox (confirm modal to complete for tasks/discussions or reopen for discussions), disabled if task is closed / user doesn't have access writes. 
  - [ ] Actions:
    - [x] Edit (disabled for closed task/discussions)
    - [x] Add Task Details (discussions only), opens the edit modal with "Add task information" checkbox ticked.
    - [ ] History (full list of history will be handled by https://github.com/pkp/pkp-lib/issues/12248)
    - [x] Delete (confirm modal to delete the task/discussion)

### Access Rules: 
  - Anyone can create tasks/discussions for as long as they have access to the workflow stage.
  - Edit, delete, viewing the history, and starting and closing the task/discussion can only be accessed by the following:
    - Journal Managers
    - Task/discussion owner [Issue: owners can't access their created task/discussion if they are not a participant: issue #5 stated on https://github.com/pkp/pkp-lib/issues/11912]
    - The person responsible to complete the task

### Adding a new Task/Discussion
- [x] "Add" button opens side modal form
- [x] Fields:
  - Details group: 
    - Templates selection - see the **Selecting a template** section below for details. 
    - Name (required)
    - Participants (required), options include:
      - all the participants for the submission
      - reviewers assigned to that stage (if in Review stage) 
        - all reviewers are available for journal managers & section editors
[...]

PR #12382, opened 2026-02-24
pkp/pkp-lib#11825 Update discussions table header depeding on the submission stage

blesildaramirez:


locale/en/submission.po
```diff
@@ -2340,13 +2340,13 @@ msgid "submission.license.cc.by-sa3.footer"
 msgstr "<a rel=\"license\" href=\"https://creativecommons.org/licenses/by-sa/3.0/\"><img alt=\"Creative Commons License\" src=\"//i.creativecommons.org/l/by-sa/3.0/88x31.png\" /></a><p>This work is licensed under a <a rel=\"license\" href=\"https://creativecommons.org/licenses/by-sa/3.0/\">Creative Commons Attribution-ShareAlike 3.0 Unported License</a>.</p>"
 
 msgid "submission.queries.submission"
-msgstr "Pre-Review Discussions"
+msgstr "Desk Review Tasks & Discussions"
```
blesildaramirez, line 2343: Hi @asmecher , can you please advise if deleting old keys and creating new ones is better than reusing the same keys and updating just the content? The old `.queries` is not used elsewhere in the codebase so maybe creating new ones is better in this case? Also not sure how Weblate handles this but as I understand - if we reuse the keys, we need to tag all the keys with `#, fuzzy` for non-English files, so the content are retranslated. But after that, will these `#, fuzzy` tags be removed automatically? If not, what would happen if we need to change them again? Thanks!

asmecher: In this case I think adding a fuzzy tag for non-english translations is OK. It would be better for folks to see the old translation, rather than a missing locale key, if the translator doesn't get to it in time.

When the translation gets updated in Weblate, it'll automatically remove the tag.

Be careful about adding fuzzy tags to the English translation, though -- that'll result in Weblate preventing translators from updating the translation until the source translation is no longer marked fuzzy.

Issue #12392, opened 2026-02-26
Introduce Funder Data to the core


ajnyga:
This issue is for introducing Funder Data to the application core. This metadata has been handled before mainly with the [Funding Plugin](https://github.com/ajnyga/funding).

The new feature will introduce a manager component for handling Funder Data. It will use ROR's as data source like we use already with affiliations. This is a change from the Funding Plugin which uses the Open Funder Registry. However, it will be soon deprecated, see [https://ror.readme.io/docs/funder-registry](https://ror.readme.io/docs/funder-registry)

- [x] Workflow Metadata settings, allow editors to enable Funder Data field
- [x] Eloquent class for Funders
- [x] API endpoint for Funders
- [x] UI-library Vue components
  - [x] Funder Manager component for listing Funders 
  - [x] Funder Form component for adding/editing Funders. Separate FieldFunder and FieldFunderGrant vue fields.
- [x] Submission Wizard allows adding Funders
- [x] Editorial dashboard allows editing/adding Funders
- [x] Landing page shows Funders
- [x] Grant ID validation based on Zenodo API (**BUT** maybe this should be a validation plugin and not in the core. It maybe does not make sense that the core has EU specific code and this would be easy to solve with a plugin with validation hooks already existing OR maybe add this validation to the existin Zenodo plugin?)
 
ojs: https://github.com/pkp/ojs/pull/5378
pkp-lib: https://github.com/pkp/pkp-lib/pull/12397
ui-library: https://github.com/pkp/ui-library/pull/813

crossref-ojs: https://github.com/pkp/crossref-ojs/pull/88
jatsTemplate: https://github.com/pkp/jatsTemplate/pull/102
pflPlugin: https://github.com/pkp/pflPlugin/pull/65
zenodo: https://github.com/pkp/zenodo/pull/5

### Things to consider

- [ ] Review of Manager component and funder form UI
- [ ] Review of publication landing page layout for Funder Data 
- [ ] Support for Funder Data in Native import and export
- [ ] Support for Funder Data in exports like JATS, ONIX, Crossref, Datacite
- [ ] Support for Funder Data in plugins that are dependent on Funding Plugin, like pfl
- [ ] Migration of Funding plugin data to the new Funder Data model

### Migration considerations

[...]

Issue #1232, opened 2016-03-07
[OMP] update es_ES locale


mtub:
Update OMP locale: es_ES (Español (España)). 

Files provided by @marcbria, credits: see https://pkp.sfu.ca/wiki/index.php?title=OMP:_Spanish_(es_ES)

Update OMP repository and OMP plugins.

mtub:
Closing… thanks @marcbria, @asmecher.

PR #12397, opened 2026-02-28
pkp/pkp-lib#1232 Support Funder Data

ajnyga:
https://github.com/pkp/pkp-lib/issues/12392

### DIFF
--- a/api/v1/funders/PKPFunderController.php
+++ b/api/v1/funders/PKPFunderController.php
@@ -1,0 +1,339 @@
+<?php
+
+/**
+ * @file api/v1/funders/PKPFunderController.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class PKPFunderController
+ *
+ * @ingroup api_v1_funders
+ *
+ * @brief Controller class to handle API requests for funder operations.
+ * 
+ */
+
+namespace pkp\api\v1\funders;
+
+use APP\core\Application;
+use APP\facades\Repo;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Facades\Route;
+use PKP\funder\Funder;
+use PKP\core\PKPBaseController;
+use PKP\core\PKPRequest;
+use PKP\plugins\Hook;
+use PKP\security\authorization\ContextAccessPolicy;
+use PKP\security\authorization\PublicationAccessPolicy;
+use PKP\security\authorization\PublicationWritePolicy;
+use PKP\security\authorization\SubmissionAccessPolicy;
+use PKP\security\authorization\UserRolesRequiredPolicy;
+use PKP\security\Role;
+use PKP\services\PKPSchemaService;
+
+class PKPFunderController extends PKPBaseController
+{
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getHandlerPath()
+     */
+    public function getHandlerPath(): string
+    {
+        return 'submissions/{submissionId}/publications/{publicationId}/funders';
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getRouteGroupMiddleware()
+     */
+    public function getRouteGroupMiddleware(): array
+    {
+        return [
+            'has.user',
+            'has.context',
+        ];
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getGroupRoutes()
+     */
+    public function getGroupRoutes(): void
+    {
+
+        Route::middleware([
+            self::roleAuthorizer([
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SITE_ADMIN,
+                Role::ROLE_ID_SUB_EDITOR,
+                Role::ROLE_ID_ASSISTANT,
+                Role::ROLE_ID_AUTHOR,
+            ]),
+        ])->group(function () {
+
+            Route::get('', $this->getMany(...))
+                ->name('funders.getMany');
+
+            Route::get('{funderId}', $this->get(...))
+                ->name('funders.getFunder')
+                ->whereNumber('funderId');
+
+            Route::post('', $this->add(...))
+                ->name('funders.add');
+
+            Route::put('{funderId}', $this->edit(...))
+                ->name('funders.edit')
+                ->whereNumber('funderId');
+
+            Route::delete('{funderId}', $this->delete(...))
+                ->name('funders.delete')
+                ->whereNumber('funderId');
+
+            Route::put('order', $this->saveOrder(...))
+                ->name('funders.order');
+
+        })->whereNumber(['submissionId', 'publicationId']);
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::authorize()
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+
+        $illuminateRequest = $args[0]; /** @var \Illuminate\Http\Request $illuminateRequest */
+        $actionName = static::getRouteActionName($illuminateRequest);
+
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
+
+        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+        $this->addPolicy(new SubmissionAccessPolicy($request, $args, $roleAssignments));
+
+        if (in_array($actionName, ['get', 'getMany'], true)) {
+            $this->addPolicy(new PublicationAccessPolicy($request, $args, $roleAssignments));
+        } else {
+            $this->addPolicy(new PublicationWritePolicy($request, $args, $roleAssignments));
+        }
+
+        return parent::authorize($request, $args, $roleAssignments);
+    }
+
+    /**
+     * Get a single funder.
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $funder = Funder::find((int) $illuminateRequest->route('funderId'));
+
+        if (!$funder) {
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.403.submissionsNotMatched'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        return response()->json(Repo::funder()->getSchemaMap()->summarize($funder), Response::HTTP_OK);
+    }
+
+    /**
+     * Get a collection of funders.
+     *
+     * @hook API::funders::params [[$collector, $illuminateRequest]]
+     */
+    public function getMany(Request $illuminateRequest): JsonResponse
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+        $funders = Funder::withSubmissionId($submission->getId())->orderBySeq();
+
+        Hook::run('API::funders::params', [$funders, $illuminateRequest]);
+
+        return response()->json([
+            'itemsMax' => $funders->count(),
+            'items' => Repo::funder()->getSchemaMap()->summarizeMany($funders->get())->values(),
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Add a funder.
+     */
+    public function add(Request $illuminateRequest): JsonResponse
+    {
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? [] : ($input['funder']['name'] ?? []),
+            'grants' => array_values(array_filter(                                                                                                                                                                                          
+                $input['grants'] ?? [],                                                                                                                                                                                                     
+                fn($g) => !empty($g['grantNumber']) || !empty($g['grantDoi']) || !empty($g['grantName'])                                                                                                                                    
+            )),
+            'seq' => 0,
+        ];    
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_FUNDER, $params);
+        $readOnlyErrors = $this->getWriteDisabledErrors(PKPSchemaService::SCHEMA_FUNDER, $params);
+        if ($readOnlyErrors) {
+            return response()->json($readOnlyErrors, Response::HTTP_BAD_REQUEST);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+        $params['submissionId'] = (int) $submission->getId();
+
+        $errors = Repo::funder()->validate(null, $params);
+        if (!empty($errors)) {
+            return response()->json($errors, Response::HTTP_BAD_REQUEST);
+        }
+
+        $funder = Funder::create($params);
+
+        return response()->json(Repo::funder()->getSchemaMap()->map($funder), Response::HTTP_OK);
+    }
+
+    /**
+     * Edit a funder.
+     */
+    public function edit(Request $illuminateRequest): JsonResponse
+    {
+        $funder = Funder::find((int)$illuminateRequest->route('funderId'));
+
+        if (!$funder) {
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.403.submissionsNotMatched'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? [] : ($input['funder']['name'] ?? []),
+            'grants' => array_values(array_filter(                                                                                                                                                                                          
+                $input['grants'] ?? [],                                                                                                                                                                                                     
+                fn($g) => !empty($g['grantNumber']) || !empty($g['grantDoi']) || !empty($g['grantName'])                                                                                                                                    
+            )),
+        ];
+
+        $params = $this->convertStringsToSchema(PKPSchemaService::SCHEMA_FUNDER, $params);
+
+        $readOnlyErrors = $this->getWriteDisabledErrors(PKPSchemaService::SCHEMA_FUNDER, $params);
+        if (!empty($readOnlyErrors)) {
+            return response()->json($readOnlyErrors, Response::HTTP_BAD_REQUEST);
+        }
+
+        $params['id'] = $funder->id;
+
+        $errors = Repo::funder()->validate($funder, $params);
+        if (!empty($errors)) {
+            return response()->json($errors, Response::HTTP_BAD_REQUEST);
+        }
+
+        $funder->update($params);
+
+        $funder = Funder::find($funder->id);
+
+        return response()->json(
+            Repo::funder()->getSchemaMap()->map($funder), Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Delete a funder.
+     */
+    public function delete(Request $illuminateRequest): JsonResponse
+    {
+        $funder = Funder::find((int) $illuminateRequest->route('funderId'));
+
+        if (!$funder) {
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.403.submissionsNotMatched'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $funder->delete();
+
+        return response()->json(
+            Repo::funder()->getSchemaMap()->map($funder), Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Save the order of funders for a publication.
+     */
+    public function saveOrder(Request $illuminateRequest): JsonResponse
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        $submissionId = (int) $submission->getId();
+        $sequence = $illuminateRequest->json()->all();
+
+        foreach ($sequence as $index => $funderId) {
+            Funder::where('funder_id', (int) $funderId)
+                ->where('submission_id', $submissionId)
+                ->update(['seq' => $index + 1]);
+        }
+
+        return response()->json(['status' => true], Response::HTTP_OK);
+    }
+
+    /**
+     * This method returns errors for any params that match
+     * properties in the schema with writeDisabledInApi set to true.
+     *
+     * This is used for properties that can not be edited through
+     * the API, but which otherwise can be edited by the entity's
+     * repository.
+     */
+    protected function getWriteDisabledErrors(string $schemaName, array $params): array
+    {
+        $schema = app()->get('schema')->get($schemaName);
+
+        $writeDisabledProps = [];
+        foreach ($schema->properties as $propName => $propSchema) {
+            if (!empty($propSchema->writeDisabledInApi)) {
+                $writeDisabledProps[] = $propName;
+            }
+        }
+
+        $errors = [];
+
+        $notAllowedProps = array_intersect(
+            $writeDisabledProps,
+            array_keys($params)
+        );
+
+        if (!empty($notAllowedProps)) {
+            foreach ($notAllowedProps as $propName) {
+                $errors[$propName] = [__('api.400.propReadOnly', ['prop' => $propName])];
+            }
+        }
+
+        return $errors;
+    }
+}
--- a/classes/components/forms/FieldFunder.php
+++ b/classes/components/forms/FieldFunder.php
@@ -1,0 +1,44 @@
+<?php
+
+/**
+ * @file classes/components/form/FieldFunder.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class FieldFunder
+ *
+ * @ingroup classes_controllers_form
+ *
+ * @brief A field for funder information.
+ */
+
+namespace PKP\components\forms;
+
+class FieldFunder extends Field
+{
+    /** @copydoc Field::$component */
+    public $component = 'field-funder';
+
+    /** @copydoc Field::$component */
+    public $default = [];
+
+    /**
+     * Submission ID associated with the funder
+     */
+    public int $submissionId = 0;
+
+    /**
+     * @copydoc Field::getConfig()
+     */
+    public function getConfig()
+    {
+        $config = parent::getConfig();
+
+        $config['submissionId'] = $this->submissionId;
+        $config['value'] = $this->value ?? $this->default ?? null;
+
+        return $config;
+    }
+}
--- a/classes/components/forms/FieldFunderGrants.php
+++ b/classes/components/forms/FieldFunderGrants.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * @file classes/components/form/FieldFunderGrants.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class FieldFunderGrants
+ *
+ * @ingroup classes_controllers_form
+ *
+ * @brief A field for funder grants.
+ */
+
+namespace PKP\components\forms;
+
+class FieldFunderGrants extends Field
+{
+    /** @copydoc Field::$component */
+    public $component = 'field-funder-grants';
+
+    /** @copydoc Field::$component */
+    public $default = [];
+
+    /**
+     * Primary language
+     */
+    public string $primaryLocale = '';
+
+    /**
+     * Supported locales for forms
+     */
+    public array $supportedFormLocales = [];
+
+    /**
+     * This value overrides the common.add button text if given.
+     */
+    private string $addButtonLabel = '';
+
+    /**
+     * @copydoc Field::getConfig()
+     */
+    public function getConfig()
+    {
+        $config = parent::getConfig();
+        $config['value'] = $this->value ?? $this->default ?? null;
+        $config['addButtonLabel'] = $this->addButtonLabel;
+        return $config;
+    }
+}
--- a/classes/components/forms/funder/FunderEditForm.php
+++ b/classes/components/forms/funder/FunderEditForm.php
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * @file classes/components/form/funder/FunderEditForm.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class FunderEditForm
+ *
+ * @ingroup classes_controllers_form
+ *
+ * @brief A preset form for setting a publication's single funder
+ */
+
+namespace PKP\components\forms\funder;
+
+use PKP\components\forms\FieldFunder;
+use PKP\components\forms\FieldFunderGrants;
+use PKP\components\forms\FormComponent;
+
+class FunderEditForm extends FormComponent
+{
+    public const FORM_FUNDER_EDIT = 'funder';
+    public $id = self::FORM_FUNDER_EDIT;
+    public $method = 'POST';
+
+    /**
+     * Constructor
+     *
+     * @param string $action URL to submit the form to
+     */
+    public function __construct(string $action)
+    {
+        $this->action = $action;
+
+        $this->addField(new FieldFunder('funder', [
+            'label' => __('submission.funders.funder'),
+            'isMultilingual' => false,
+        ]));
+        $this->addField(new FieldFunderGrants('grants', [
+            'label' => __('submission.funders.funder.grants.label.name'),
+            'description' => __('submission.funders.funder.grants.label.description'),
+            'value' => null,
+        ]));
+    }
+}
--- a/classes/funder/Funder.php
+++ b/classes/funder/Funder.php
@@ -1,0 +1,237 @@
+<?php
+
+/**
+ * @file classes/funder/Funder.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Funder
+ *
+ * @brief Basic class describing Funders existing in the system.
+ */
+
+namespace PKP\funder;
+
+use APP\core\Application;
+use APP\facades\Repo;
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Queue\Events\Looping;
+use PKP\core\traits\ModelWithSettings;
+use PKP\i18n\LocaleConversion;
+use PKP\ror\Ror;
+use PKP\services\PKPSchemaService;
+use PKP\submission\PKPSubmission;
+use PKP\context\Context;
+
+class Funder extends Model
+{
+    use ModelWithSettings;
+
+    protected $table = 'funders';
+    protected $primaryKey = 'funder_id';
+    public const CREATED_AT = null;
+    public const UPDATED_AT = null;
+    protected string $settingsTable = 'funder_settings';
+
+    protected $guarded = ['funderId', 'id'];
+
+    /** Per-request caches keyed by id — shared across all Funder instances. */
+    private static array $submissionCache = [];
+    private static array $contextCache    = [];
+    private static array $rorCache        = [];
+    private static array $localesCache    = [];
+
+    /**
+     * @inheritDoc
+     */
+    public static function getSchemaName(): ?string
+    {
+        return PKPSchemaService::SCHEMA_FUNDER;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getSettingsTable(): string
+    {
+        return $this->settingsTable;
+    }
+
+    /**
+     * Filter by submission ID
+     */
+    protected function scopeWithSubmissionId(EloquentBuilder $builder, int $submissionId): EloquentBuilder
+    {
+        return $builder->where('submission_id', $submissionId);
+    }
+
+    /**
+     * Order by seq
+     */
+    protected function scopeOrderBySeq(EloquentBuilder $builder): EloquentBuilder
+    {
+        return $builder->orderBy('seq')->orderBy('funder_id');
+    }
+
+    /**
+     * The parent Submission, lazily loaded and memoized.
+     */
+    protected function submission(): Attribute
+    {
+        return Attribute::make(
+            get: function (): ?PKPSubmission {
+                $id = (int) $this->getRawOriginal('submission_id');
+                if (!$id) {
+                    return null;
+                }
+                if (!array_key_exists($id, self::$submissionCache)) {
+                    self::$submissionCache[$id] = Repo::submission()->get($id);
+                }
+                return self::$submissionCache[$id];
+            },
+        )->shouldCache();
+    }
+
+    /**
+     * The submission's Context, lazily loaded and memoized.
+     */
+    protected function context(): Attribute
+    {
+        return Attribute::make(
+            get: function (): ?Context {
+                $submission = $this->submission;
+                if (!$submission) {
+                    return null;
+                }
+                $contextId = (int) $submission->getData('contextId');
+                if (!$contextId) {
+                    return null;
+                }
+                if (!array_key_exists($contextId, self::$contextCache)) {
+                    self::$contextCache[$contextId] = Application::getContextDAO()->getById($contextId);
+                }
+                return self::$contextCache[$contextId];
+            },
+        )->shouldCache();
+    }
+
+    /**
+     * The resolved ROR object, lazily loaded and memoized.
+     */
+    protected function rorObject(): Attribute
+    {
+        return Attribute::make(
+            get: function (): ?Ror {
+                $ror = $this->getRawOriginal('ror');
+                if (empty($ror)) {
+                    return null;
+                }
+                if (!array_key_exists($ror, self::$rorCache)) {
+                    self::$rorCache[$ror] = Repo::ror()
+                        ->getCollector()
+                        ->filterByRor($ror)
+                        ->getMany()
+                        ->first();
+                }
+                return self::$rorCache[$ror];
+            },
+        )->shouldCache();
+    }
+
+    /**
+     * Funder name.
+     * - If a multilingual value is stored in settings, return it unchanged.
+     * - Else, if the funder has a ROR, derive a multilingual array by mapping
+     *   the ROR's names onto the submission's publication languages.
+     */
+    protected function name(): Attribute
+    {
+        return Attribute::make(
+            get: function (mixed $value, array $attributes) {
+                if (!empty($value) || empty($attributes['ror'])) {
+                    return $value;
+                }
+                $rorObject = $this->rorObject;
+                if (!$rorObject) {
+                    return $value;
+                }
+                $locales = $this->resolvedPublicationLanguages();
+                if (empty($locales)) {
+                    return $value;
+                }
+                $names = [];
+                foreach ($locales as $locale) {
+                    $rorLocale = LocaleConversion::getIso1FromLocale($locale);
+                    $names[$locale] = $rorObject->getName($rorLocale)
+                        ?? $rorObject->getName($rorObject->getDisplayLocale());
+                }
+                return empty($names) ? $value : $names;
+            },
+        )->shouldCache();
+    }
+
+    /**
+     * Used by ModelWithSettings::getLocalizedData() as the last-resort locale
+     * in its fallback chain. Returns the submission's primary locale.
+     */
+    public function getDefaultLocale(): ?string
+    {
+        return $this->submission?->getData('locale');
+    }
+
+    /**
+     * Compute the full locale set for this funder's submission, memoized.
+     *
+     * @return string[]
+     */
+    private function resolvedPublicationLanguages(): array
+    {
+        $submissionId = (int) $this->getRawOriginal('submission_id');
+        if (!$submissionId) {
+            return [];
+        }
+        if (isset(self::$localesCache[$submissionId])) {
+            return self::$localesCache[$submissionId];
+        }
+        $submission = $this->submission;
+        $context = $this->context;
+        if (!$submission || !$context) {
+            return self::$localesCache[$submissionId] = [];
+        }
+        return self::$localesCache[$submissionId] = $submission->getPublicationLanguages(
+            $context->getSupportedSubmissionMetadataLocales() ?? []
+        );
+    }
+
+    /**
+     * Reset per-process caches.
+     */
+    public static function clearResolverCaches(): void
+    {
+        self::$submissionCache = [];
+        self::$contextCache    = [];
+        self::$rorCache        = [];
+        self::$localesCache    = [];
+    }
+
+    /**
+     * Between each worker loop iteration (CLI mode only), flush the per-request
+     * static caches so data from the previous job can't bleed into the next one.
+     */
+    protected static function booted(): void
+    {
+        app('events')->listen(
+            Looping::class,
+            function (Looping $event): void {
+                if (!app()->runningInConsole() || app()->runningUnitTests()) {
+                    return;
+                }
+                self::clearResolverCaches();
+            }
+        );
+    }
+}
--- a/classes/funder/Repository.php
+++ b/classes/funder/Repository.php
@@ -1,0 +1,181 @@
+<?php
+/**
+ * @file classes/funder/Repository.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Repository
+ *
+ * @brief A helper class to handle operations with Funder data
+ */
+
+namespace PKP\funder;
+
+use APP\core\Application;
+use APP\core\Request;
+use PKP\plugins\Hook;
+use PKP\services\PKPSchemaService;
+use PKP\validation\ValidatorFactory;
+
+class Repository
+{
+    /** @var string $schemaMap The name of the class to map this entity to its schema */
+    public string $schemaMap = maps\Schema::class;
+
+    protected Request $request;
+
+    /** @var PKPSchemaService<Funder> $schemaService */
+    protected PKPSchemaService $schemaService;
+
+    // Funders with awards that can be validated via the Zenodo API, and their corresponding ROR IDs.
+    const AWARD_FUNDERS = [
+        '05k73zm37', // Research Council of Finland
+        '00rbzpz17', // French National Research Agency
+        '05mmh0f86', // Australian Research Council
+        '03zj4c476', // Aligning Science Across Parkinson's
+        '01gavpb45', // Canadian Institutes of Health Research
+        '00k4n6c32', // European Commission
+        '02k4b9v70', // European Environment Agency
+        '00snfqn58', // Portuguese Science and Technology Foundation
+        '013tf3c58', // Austrian Science Fund
+        '03m8vkq32', // The French National Cancer Institute
+        '03n51vw80', // Croatian Science Foundation
+        '02ar66p97', // Latvian Council of Science
+        '01znas443', // Ministry of Education, Science and Technological Development of the Republic of Serbia
+        '011kf5r70', // National Health and Medical Research Council
+        '01cwqze88', // National Institutes of Health
+        '01h531d29', // Natural Sciences and Engineering Research Council of Canada
+        '021nxhr62', // National Science Foundation
+        '04jsz6e67', // Dutch Research Council
+        '00dq2kk65', // Research Councils UK
+        '0271asj38', // Science Foundation Ireland
+        '00yjd3n13', // Swiss National Science Foundation
+        '006cvnv84', // Social Science Research Council
+        '04w9kkr77', // Scientific and Technological Research Council of Turkey
+        '00x0z1472', // Templeton World Charity Foundation
+        '001aqnf71', // UK Research and Innovation
+        '029chgv08', // Wellcome Trust
+    ];
+
+    public function __construct(Request $request, PKPSchemaService $schemaService)
+    {
+        $this->request = $request;
+        $this->schemaService = $schemaService;
+    }
+
+    /**
+     * Validate properties for a funder
+     *
+     * Perform validation checks on data used to add or edit a funder
+     *
+     * @param Funder|null $funder Funder being edited. Pass `null` if creating a new funder
+     * @param array $props A key/value array with the new data to validate
+     *
+     * @return array A key/value array with validation errors. Empty if no errors
+     *
+     * @hook Funder::validate [[&$errors, $funder, $props]]
+     */
+    public function validate(?Funder $funder, array $props): array
+    {
+        $schema = Funder::getSchemaName();
+
+        $validator = ValidatorFactory::make(
+            $props,
+            $this->schemaService->getValidationRules($schema, [])
+        );
+
+        // Check required fields
+        ValidatorFactory::required(
+            $validator,
+            $funder,
+            $this->schemaService->getRequiredProps($schema),
+            $this->schemaService->getMultilingualProps($schema),
+            [],
+            ''
+        );
+
+        // Validate grant numbers if funder grant validation is enabled and a ROR ID is provided
+        $context = $this->request->getContext();
+        $funderGrantValidationSetting = (bool) $context?->getData('funderGrantValidation');
+
+        if ($funderGrantValidationSetting) {
+            $ror = $props['ror'] ?? null;
+
+            if ($ror) {
+                $ror = preg_replace('#^https?://ror\.org/#i', '', trim($ror));
+
+                if (in_array($ror, self::AWARD_FUNDERS, true)) {
+
+                    $validator->after(function ($validator) use ($props, $ror) {
+
+                        $grants = $props['grants'] ?? [];
+                        $httpClient = Application::get()->getHttpClient();
+
+                        foreach ($grants as $key => $grant) {
+                            $grantNumber = $grant['grantNumber'] ?? null;
+
+                            if (!$grantNumber) {
+                                continue;
+                            }
+
+                            try {
+                                $awardResponse = $httpClient->request(
+                                    'GET',
+                                    "https://zenodo.org/api/awards?funders={$ror}&q=" . urlencode($grantNumber),
+                                    ['timeout' => 5]
+                                );
+
+                                $body = json_decode($awardResponse->getBody(), true);
+
+                                if ($body === null) {
+                                    throw new \Exception('Zenodo returned a non-JSON response');
+                                }
+
+                                $success = array_reduce(
+                                    $body['hits']['hits'] ?? [],
+                                    fn($carry, $item) => $carry || $item['number'] == $grantNumber,
+                                    false
+                                );
+
+                                if (!$success) {
+                                    $validator->errors()->add("grants.{$key}.grantNumber", __('submission.funders.grantNumberInvalid'));
+                                }
+                            } catch (\Exception $e) {
+                                // Zenodo unavailable or timed out — skip validation silently
+                                continue;
+                            }
+                        }
+                    });
+                }
+            }
+        }
+
+        // Validate that either a ror or funder name exists
+        $validator->after(function ($validator) use ($props) {
+            if (empty($props['ror']) && empty(array_filter($props['name'] ?? []))) {
+                $validator->errors()->add('funder', __('submission.funders.funderNameOrRorRequired'));
+            }
+        });
+
+        $errors = [];
+
+        if ($validator->fails()) {
+            $errors = $this->schemaService->formatValidationErrors($validator->errors());
+        }
+
+        Hook::call('Funder::validate', [&$errors, $funder, $props]);
+
+        return $errors;
+    }
+
+    /**
+     * Get an instance of the map class for mapping
+     * data citations to their schema
+     */
+    public function getSchemaMap(): maps\Schema
+    {
+        return app('maps')->withExtensions($this->schemaMap);
+    }
+}
--- a/classes/funder/maps/Schema.php
+++ b/classes/funder/maps/Schema.php
@@ -1,0 +1,113 @@
+<?php
+/**
+ * @file classes/funder/maps/Schema.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class Schema
+ *
+ * @brief Map Funder data to the properties defined in the Funder schema.
+ */
+
+namespace PKP\funder\maps;
+
+use APP\facades\Repo;
+use Illuminate\Support\Enumerable;
+use PKP\funder\Funder;
+use PKP\services\PKPSchemaService;
+
+class Schema extends \PKP\core\maps\Schema
+{
+    /** @copydoc \PKP\core\maps\Schema::$collection */
+    public Enumerable $collection;
+
+    /** @copydoc \PKP\core\maps\Schema::$schema */
+    public string $schema = PKPSchemaService::SCHEMA_FUNDER;
+
+    /**
+     * Map a Funder
+     */
+    public function map(Funder $item): array
+    {
+        return $this->mapByProperties($this->getProps(), $item);
+    }
+
+    /**
+     * Summarize a Funder
+     */
+    public function summarize(Funder $item): array
+    {
+        return $this->mapByProperties($this->getSummaryProps(), $item);
+    }
+
+    /**
+     * Map a collection of Funder
+     */
+    public function mapMany(Enumerable $collection): Enumerable
+    {
+        $this->collection = $collection;
+        return $collection->map(function ($item) {
+            return $this->map($item);
+        });
+    }
+
+    /**
+     * Summarize a collection of Funder
+     */
+    public function summarizeMany(Enumerable $collection): Enumerable
+    {
+        $this->collection = $collection;
+        return $collection->map(function ($item) {
+            return $this->summarize($item);
+        });
+    }
+
+    /**
+     * Map schema properties of a Funder to an assoc array
+     */
+    protected function mapByProperties(array $props, Funder $item): array
+    {
+        $grantModel = $this->getGrantDataModel();
+        $output = [];
+
+        foreach ($props as $prop) {
+            switch ($prop) {
+                case 'grants':
+                    $grants = [];
+                    foreach (is_array($item->getAttribute($prop)) ? $item->getAttribute($prop) : [] as $grant) {
+                        $grants[] = array_merge($grantModel, $grant);
+                    }
+                    $output[$prop] = $grants;
+                    break;
+
+                case 'rorObject':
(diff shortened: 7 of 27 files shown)


classes/funder/Funder.php
```diff
@@ -0,0 +1,109 @@
...
+    use ModelWithSettings;
+
+    protected $table = 'funders';
+    protected $primaryKey = 'funder_id';
+    public const CREATED_AT = null;
+    public const UPDATED_AT = null;
+    protected string $settingsTable = 'funder_settings';
+
+    /**
+     * The attributes that are mass assignable.
```
kaitlinnewson:
```suggestion
     * The attributes that are not mass assignable.
```

classes/funder/maps/Schema.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        $schemaService = new PKPSchemaService();
+        $schema = $schemaService->get($this->schema);
+        $grantModel = [];
+        foreach (array_keys((array)$schema->properties->grants->items->properties) as $property) {
+            $grantModel[$property] = '';
+        }
+        return $grantModel;
+    }
+}
```
bozana, line 130: missing empty line at the end of the file

classes/funder/Funder.php
```diff
@@ -0,0 +1,109 @@
...
+                    ->filterByRor($attributes['ror'])
+                    ->getMany()
+                    ->first();
+
+                if (!$ror) {
+                    return $value;
+                }
+
+                $names = [];
+                foreach (Application::get()->getRequest()->getContext()->getSupportedLocales() as $allowedLocale) {
```
bozana, line 100: I think those are wrong locales, we would need to use submission's publication languages, i.e. something like:
`$submission->getPublicationLanguages($this->context->getSupportedSubmissionMetadataLocales());`.

So, here we would also need to get the submission (additionally to context object) here...

And because we would duplicate the queries here and in Schema.php (at least), I believe it would be good to have a more centralized solution, also because of performance... if possible...

classes/funder/Funder.php
```diff
@@ -0,0 +1,109 @@
...
+                    $rorLocale = LocaleConversion::getIso1FromLocale($allowedLocale);
+                    $names[$allowedLocale] = $ror->getName($rorLocale) ?? $ror->getName($ror->getDisplayLocale());
+                }
+
+                return empty($names) ? $value : $names;
+            }
+        );
+    }
+}
```
bozana, line 109: empty line missing

classes/submission/DAO.php
```diff
@@ -291,6 +292,10 @@ public function deleteById(int $id): int
         $submissionCommentDao = DAORegistry::getDAO('SubmissionCommentDAO'); /** @var SubmissionCommentDAO $submissionCommentDao */
         $submissionCommentDao->deleteBySubmissionId($id);
 
+        // Delete a submissions's Funders
+        Funder::withSubmissionIds([$id])
+            ->delete();
+
```
bozana, line 298: This is maybe not needed because we have the cascade defined in the DB migration?

bozana: Also, we have defined `scopeWithSubmissionId` i.e. singular so `withSubmissionIds` does not exist.

classes/components/forms/FieldFunder.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Primary language
+     */
+    public string $primaryLocale = '';
+
+    /**
+     * Supported locales for forms
+     */
+    public array $supportedFormLocales = [];
```
bozana, line 40: How do we provide primaryLocale and supportedFormLocales  to the Vue component?
supportedFormLocales should then be = $this->submission->getPublicationLanguages($this->context->getSupportedSubmissionMetadataLocales()); ?
primaryLocale is submission locale?

classes/components/forms/FieldFunderGrants.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Supported locales for forms
+     */
+    public array $supportedFormLocales = [];
+
+    /**
+     * This value overrides the common.add button text if given.
+     */
+    private string $addButtonLabel = '';
```
bozana, line 40: Should this be rather protected or public?

classes/publication/DAO.php
```diff
@@ -247,6 +249,7 @@ public function deleteById(int $publicationId): int
         $this->deleteControlledVocab($publicationId);
         $this->deleteDataCitations($publicationId);
         Repo::citation()->deleteByPublicationId($publicationId);
+        // Deleting Funders takes place in Submission DAO because Funders are linked to Submissions
```
bozana, line 252: Deletion is handled by the database cascade on submission_id. So, maybe no comment is needed — the foreign key definition in the migration is the documentation for that behavior.

classes/funder/Repository.php
```diff
@@ -0,0 +1,172 @@
...
+                                "https://zenodo.org/api/awards?funders={$ror}&q=" . urlencode($grantNumber)
+                            );
+
+                            $body = json_decode($awardResponse->getBody(), true);
+
+                            $success = array_reduce(
+                                $body['hits']['hits'] ?? [],
+                                fn($carry, $item) => $carry || $item['number'] == $grantNumber,
+                                false
+                            );
```
bozana, line 135: This seems to be a little bit 'dangerous' to validate against 3rd party synchronously. 
If the Zenodo call fails for any reason, the validation should rather be skipped than returning an error to the user, I think.

I am not sure if we should use it at all... It is just a small amount or RORs we can validate against... and are they changing?... I maybe need to take a detailed look at Zenodo... 🤔 

If we keep it then maybe to:
- Wrap the HTTP call in a try/catch — if Zenodo is down or times out, the entire form submission should not fail. So maybe to catch the exception and  skip validation silently (or add a warning but don't block).
- Add a timeout as option to the Guzzle client — maybe 5 seconds ?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I12392_Funders.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
bozana, line 7: 2026

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
...
+            $legacyFunders = DB::table('funders_legacy')->get();
+
+            foreach ($legacyFunders as $legacyFunder) {
+
+                // If we want to try mapping old Funder Registry values to RORs, this is where that would happen.
+                // We could use a csv file for mapping containing ROR's and matching Funder Registry DOI suffixes 
+
+                $newFunderId = DB::table('funders')->insertGetId([
+                    'submission_id' => $legacyFunder->submission_id,
+                    'ror' => $legacyFunder->funder_identification,
```
bozana, line 80: Here only RORs should be saved, when the mapping is added, right?

classes/funder/Funder.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return $builder->where('submission_id', $submissionId);
+    }
+
+    /**
+     * Order by seq
+     */
+    protected function scopeOrderBySeq(EloquentBuilder $builder): EloquentBuilder
+    {
+        return $builder->orderBy('seq');
```
bozana, line 75: Maybe to add `->orderBy('funder_id'); ` at the end?
When all seq values are equal, the database is free to return rows in any order, so that different databases can return different orderings, I think.

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
...
+            $table->string('setting_name', 255);
+            $table->mediumText('setting_value')->nullable();
+            $table->foreign('funder_id', 'funder_settings_funder_id')->references('funder_id')->on('funders')->onDelete('cascade');
+            $table->index(['funder_id'], 'funder_settings_funder_id');
+            $table->unique(['funder_id', 'locale', 'setting_name'], 'funder_settings_unique');
+        });
+
+        // Migrate legacy funder data if it exists
+        if (Schema::hasTable('funders_legacy')) {
+            $legacyFunders = DB::table('funders_legacy')->get();
```
bozana, line 71: shell we maybe use `->orderBy('funder_id')` here -- just to be safe -- to keep the ordering as it was, if that was important back then?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
...
+
+                $settings = DB::table('funder_settings_legacy')
+                    ->where('funder_id', $legacyFunder->funder_id)
+                    ->get();
+
+                foreach ($settings as $setting) {
+                    DB::table('funder_settings')->insert([
+                        'funder_id' => $newFunderId,
+                        'locale' => $setting->locale,
+                        'setting_name' => $setting->setting_name,
```
bozana, line 92: We set the setting_name as it was, but now we use 'name' and earlier it was 'funderName'.

bozana: Also, if we round the ROR, we should not insert the name -- as the current logic is.

bozana: Earlier there was no `locale`, so maybe we need to map it to the submission primary locale?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
...
+
+                    DB::table('funder_settings')->insert([
+                        'funder_id' => $newFunderId,
+                        'locale' => '',
+                        'setting_name' => 'grants',
+                        'setting_value' => json_encode($grants),
+                    ]);
+                }
+            }
+        }
```
bozana, line 119: Maybe this can be implemented with a slightly better performance:
Get all data at the beginning, then loop through legacy funders, then construct $newSettings, then batch insert all settings at once:
```
     $legacyFunders = DB::table('funders_legacy')->orderBy('funder_id')->get();
                                                                                                                                                                         
      $legacySettings = DB::table('funder_settings_legacy')                                                                                                              
          ->get()                                                                                                                                                        
          ->groupBy('funder_id');                                                                                                                                        
                  
      $legacyAwards = Schema::hasTable('funder_awards_legacy')
          ? DB::table('funder_awards_legacy')->get()->groupBy('funder_id')
          : collect();

      $submissionLocales = DB::table('submissions')                                                                                                                          
            ->whereIn('submission_id', $legacyFunders->pluck('submission_id'))
            ->pluck('locale', 'submission_id');

     $newSettings = []; 

     foreach ($legacyFunders as $legacyFunder) {

          // mapping
          // maybe copy the CSV to lib/pkp/lib/funder-ror-map.csv and use it here
          $ror = ...

          $newFunderId = DB::table('funders')->insertGetId([                                                                                                         
                  'submission_id' => $legacyFunder->submission_id,                                                                                                       
                  'ror' => $ror,                                                                                                                                         
                  'seq' => 0,
              ]);

         $locale = $submissionLocales[$legacyFunder->submission_id];

             // Migrate settings, remapping funderName -> name                                                                                                          
              foreach ($legacySettings[$legacyFunder->funder_id] ?? [] as $setting) {
                  if ($setting->setting_name === 'funderName') {                                                                                                         
                      // Only migrate name if we have no ROR (ROR provides name at runtime)
                      if (!$ror) {                                                                                                                                       
                          $newSettings[] = [                                                                                                                             
                              'funder_id' => $newFunderId,                                                                                                               
                              'locale' => $locale,                                                                                            
                              'setting_name' => 'name',
                              'setting_value' => $setting->setting_value,                                                                                                
                          ];
                      }                                                                                                                                                  
                  }
                  // other settings (if any) migrated as-is
              }        

             // Collapse awards into grants JSON                                                                                                                        
              $awardNumbers = ($legacyAwards[$legacyFunder->funder_id] ?? collect())
                  ->pluck('funder_award_number')                                                                                                                         
                  ->unique()                                                                                                                                             
                  ->values();                                                                                                                                            
                                                                                                                                                                         
              if ($awardNumbers->isNotEmpty()) {
                  $newSettings[] = [
                      'funder_id' => $newFunderId,
                      'locale' => '',
                      'setting_name' => 'grants',                                                                                                                        
                      'setting_value' => json_encode(
                          $awardNumbers->map(fn($n) => [                                                                                                                 
                              'grantDoi' => null,                                                                                                                        
                              'grantNumber' => $n,
                              'grantName' => null,                                                                                                                       
                          ])->all()
                      ),                                                                                                                                                 
                  ];
              }                

         // Batch insert all settings at once
          foreach (array_chunk($newSettings, 100) as $chunk) {
              DB::table('funder_settings')->insert($chunk);                                                                                                              
          }
                                              
```

bozana: eventually have one DB transaction around all insertions.... 🤔

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,135 @@
...
+        });
+
+        // Migrate legacy funder data if it exists
+        if (Schema::hasTable('funders_legacy')) {
+            $legacyFunders = DB::table('funders_legacy')->get();
+
+            foreach ($legacyFunders as $legacyFunder) {
+
+                // If we want to try mapping old Funder Registry values to RORs, this is where that would happen.
+                // We could use a csv file for mapping containing ROR's and matching Funder Registry DOI suffixes
```
bozana, line 76: I think we should migrate it.
I think your idea was good -- to have the CSV in a GitHub repo, updated by a GitHubAction, and get the local copy of it (either as a submodule or just a file in lib/pkp/lib?).

classes/funder/Funder.php
```diff
@@ -0,0 +1,109 @@
...
+     */
+    protected function name(): Attribute
+    {
+        return Attribute::make(
+            get: function (mixed $value, array $attributes) {
+                if (!empty($value) || empty($attributes['ror'])) {
+                    return $value;
+                }
+
+                $ror = Repo::ror()
```
bozana, line 89: This fires a live Repo::ror()->getCollector()->filterByRor() query every time name is accessed on a funder that has a ROR...
In Funder/maps/Schema.php, the rorObject case does a separate ROR lookup too...
hmmm... 🤔
Maybe we should think about another solution here... 🤔

classes/funder/maps/Schema.php
```diff
@@ -0,0 +1,130 @@
...
+     * Map schema properties of a Funder to an assoc array
+     */
+    protected function mapByProperties(array $props, Funder $item): array
+    {
+        $grantModel = $this->getGrantDataModel();
+        $output = [];
+
+        foreach ($props as $prop) {
+            switch ($prop) {
+                case 'grants':
```
bozana, line 85: case '_href' is missing

schemas/funder.json
```diff
@@ -0,0 +1,88 @@
+{
+	"title": "Funder",
+	"description": "A Funder in a Publication.",
+    "type": "object",
+    "required": [
+        "submissionId"
+    ],
+	"properties": {
```
bozana, line 8: "_href" is missing

bozana: I see it is also missing for data citations... :-)

classes/funder/Funder.php
bozana: Hi @touhidurabir, we have been reviewing the funder implementation and would love your input on the best way to centralize funder name resolution.

The problem: 
A funder can have either a ROR or manually entered names in the submission metadata languages. We need a centralized way to get the correct name in both cases, without duplicating the logic in every export (DataCite, JATS, templates, etc.).
For ROR-based funders, the name comes from the existing ROR classes, but ROR names don't necessarily exist in the submission metadata locales — so we need a mapping: use the ROR name in the requested locale if available, otherwise fall back to the ROR display locale name.

Since Funder is an Eloquent model using ModelWithSettings, one approach would maybe be to use the booted() method with a retrieved listener:

```
protected static function booted(): void
  {                                                                                                                                                                      
      static::retrieved(function (Funder $funder) {
          if (!$funder->ror) {
              return;
          }                                                                                                                                                              
          $rorObject = Repo::ror()->getCollector()->filterByRor($funder->ror)->getMany()->first();
          if (!$rorObject) {                                                                                                                                             
              return;
          }                                                                                                                                                              
          $funder->setAttribute('rorObject', $rorObject);
          if (empty($funder->getRawOriginal('name'))) {
              $submission = Repo::submission()->get($funder->submission_id);                                                                                             
              $context = Application::getContextDAO()->getById($submission->getData('contextId'));                                                                       
              $locales = $submission->getPublicationLanguages(                                                                                                           
                  $context->getSupportedSubmissionMetadataLocales()                                                                                                      
              );                                                                                                                                                         
              $names = [];
              foreach ($locales as $locale) {                                                                                                                            
                  $rorLocale = LocaleConversion::getIso1FromLocale($locale);
                  $names[$locale] = $rorObject->getName($rorLocale)                                                                                                      
                      ?? $rorObject->getName($rorObject->getDisplayLocale());                                                                                            
              }                                                                                                                                                          
              $funder->setAttribute('name', $names);                                                                                                                     
          }       
          $funder->setAttribute('defaultLocale', $submission->getData('locale'));
      });                                                                                                                                                                
  }

```

This would:     
- Set rorObject on every retrieved funder that has a ROR — so $item->rorObject just works in chema.php without an extra DB query
- Map ROR names to submission metadata locales and set them on name — so every consumer can simply use $funder->getLocalizedData('name', $locale) regardless of whether the funder has a ROR or manually entered names 
- Set defaultLocale to the submission locale so that getDefaultLocale() works correctly as a fallback in getLocalizedData() without needing to query the submission again, e.g.:
```
 public function getDefaultLocale(): ?string                                                                                                                            
  {                                          
      return $this->getAttribute('defaultLocale');                                                                                                                       
  }        
```
All DB queries (ROR, submission, context) happen once in one place. Schema.php needs no special name case (falls through to default) and no extra query for rorObject.
Every export and template uses `$funder->getLocalizedData('name', $locale)` consistently. 
Every plugin could get the funders for a submission with:
`Funder::withSubmissionId($submissionId)->orderBySeq()->orderBy('funder_id')->get();`

Does this approach look good to you, or do you see a better way to handle this within the Eloquent model pattern?

bozana (commented):
Hi @ajnyga, great work! Well structured, well done and thought through!
There is maybe one or two important things to consider here. 
One of them is the solution to get all needed objects and funder attributes at one place, that I am not sure how to best solve with Eloquent. Thus, I asked/will ask Touhidur too.
The other is making sure Zenodo validation do not 'break' the system.
And maybe also to slightly try to improve the migration script performance.
Thanks a lot!

classes/funder/Funder.php
touhidurabir: Hi @bozana using the eloquent model event `retrieved` is bit risky and need a very careful consideration when to use it and then also the cases how/when to fetch model data from DB in a single request life cycle. Otherwise it may/can infinite recursion and break the app . 

So follow this case
 1. from one block we have code that retrieve the target model 
 2. in the callback of this event, we have some codes that execute another different block
 3. that different block again fetch a model instance of same model as it can be also used as standalone call for other purpose
 4. that would cause to fire the `retrieved` event again of the step 1 target model
 5. we go back to step 1 which cause an infinite recursion loop
 
it is not impossible to handle it but always need very careful checking and consideration which can break anytime. so I consider it bit fragile .

bozana: Hi @touhidurabir, would you maybe have another suggestion? Should we then rather leave it all in the name attribute function and Schema.php?

bozana: I think in our scenario the recursion risk is real -- we get the submission object, that gets publications, that sets funders... 🤔

bozana: hmmm... maybe we could implement a Repo method, that retrieves funders, but then other developers would need to know that only that function should be used, and not something like Funder::withSubmissionId($submissionId)... and $funder->getFunderName() instead of $funder->name... 🤔 
For the schema map we need to provide submission in constructor... so name can there be handled correctly for API calls... 🤔

bozana: I do not know what is the good/best way to solve it 😅

bozana: maybe this solution would be then best:

- retrieved — sets rorObject
- getDefaultLocale() — needs a DB query to get the submission locale
- getFunderName(?string $locale, array $allowedLocales) — for full locale-keyed array -- currently only used in Schema.php
- getLocalizedFunderName(?string $preferredLocale) — for single best value, uses getLocalizedData() with correct defaultLocale fallback in both cases (when name entered manually or for rorObject) -- used in most exports
- no extra name() Attribute function
- for the schema map we need to provide submission in constructor... so allowedLocales can be provided to getFunderName()...

schemas/publication.json
```diff
@@ -210,6 +210,15 @@
 			"readOnly": true,
 			"apiSummary": true
 		},
+		"funders": {
+			"type": "array",
+			"description": "Optional metadata that contains an array of funders for this publication.",
+			"apiSummary": true,
```
bozana, line 216: Could we remove funders from apiSummary -- I think no need?
Also dataCitations, please?

classes/funder/Funder.php
bozana: 🎉 

Touhidur found a really elegant solution, that I think we can use:
```

/** Per-request caches keyed by id — shared across all Funder instances. */
     private static array $submissionCache = []; // int submissionId => PKPSubmission|null
     private static array $contextCache    = []; // int contextId    => Context|null
     private static array $rorCache        = []; // string ror       => Ror|null
     private static array $localesCache    = []; // int submissionId => string[]

/** The parent Submission, lazily loaded and memoized (per-instance + per-request). */
     protected function submission(): Attribute
     {
         return Attribute::make(
             get: function (): ?PKPSubmission {
                 $id = (int) $this->getRawOriginal('submission_id');
                 if (!$id) {
                     return null;
                 }
                 if (!array_key_exists($id, self::$submissionCache)) {
                     self::$submissionCache[$id] = Repo::submission()->get($id);
                 }
                 return self::$submissionCache[$id];
             },
         )->shouldCache();
     }

/** The submission's Context, lazily loaded and memoized. */
     protected function context(): Attribute
     {
         return Attribute::make(
             get: function (): ?Context {
                 $submission = $this->submission;
                 if (!$submission) {
                     return null;
                 }
                 $contextId = (int) $submission->getData('contextId');
                 if (!$contextId) {
                     return null;
                 }
                 if (!array_key_exists($contextId, self::$contextCache)) {
                     self::$contextCache[$contextId] = Application::getContextDAO()->getById($contextId);
                 }
                 return self::$contextCache[$contextId];
             },
         )->shouldCache();
     }

/** The resolved ROR object, lazily loaded and memoized. */
     protected function rorObject(): Attribute
     {
         return Attribute::make(
             get: function (): ?Ror {
                 $ror = $this->getRawOriginal('ror');
                 if (empty($ror)) {
                     return null;
                 }
                 if (!array_key_exists($ror, self::$rorCache)) {
                     self::$rorCache[$ror] = Repo::ror()
                         ->getCollector()
                         ->filterByRor($ror)
                         ->getMany()
                         ->first();
                 }
                 return self::$rorCache[$ror];
             },
         )->shouldCache();
     }

/**
      * Funder name.
      * - If a multilingual value is stored in settings, return it unchanged.
      * - Else, if the funder has a ROR, derive a multilingual array by mapping
      *   the ROR's names onto the submission's publication languages (falling
      *   back to the ROR's display locale where a locale isn't available).
      */
     protected function name(): Attribute
     {
         return Attribute::make(
             get: function (mixed $value, array $attributes) {
                 if (!empty($value) || empty($attributes['ror'])) {
                     return $value;
                 }
                 $rorObject = $this->rorObject;
                 if (!$rorObject) {
                     return $value;
                 }
                 $locales = $this->resolvedPublicationLanguages();
                 if (empty($locales)) {
                     return $value;
                 }

                 $names = [];
                 foreach ($locales as $locale) {
                     $rorLocale = LocaleConversion::getIso1FromLocale($locale);
                     $names[$locale] = $rorObject->getName($rorLocale)
                         ?? $rorObject->getName($rorObject->getDisplayLocale());
                 }
                 return empty($names) ? $value : $names;
             },
         )->shouldCache();
     }

/**
      * Used by ModelWithSettings::getLocalizedData() as the last-resort locale
      * in its fallback chain. Returns the submission's primary locale.
      */
     public function getDefaultLocale(): ?string
     {
         return $this->submission?->getData('locale');
     }

     /**
      * Compute the full locale set for this funder's submission, memoized.
      * Uses $submission->getPublicationLanguages() directly — safe here
      * because no `retrieved` handler exists; the publication cascade runs
      * silently.
      *
      * @return string[]
      */
     private function resolvedPublicationLanguages(): array
     {
         $submissionId = (int) $this->getRawOriginal('submission_id');
         if (!$submissionId) {
             return [];
         }
         if (isset(self::$localesCache[$submissionId])) {
             return self::$localesCache[$submissionId];
         }
         $submission = $this->submission;
         $context = $this->context;
         if (!$submission || !$context) {
             return self::$localesCache[$submissionId] = [];
         }
         return self::$localesCache[$submissionId] = $submission->getPublicationLanguages(
             $context->getSupportedSubmissionMetadataLocales() ?? []
         );
     }

     /**
      * Reset per-process caches. Safe to call repeatedly.
      * Hooked automatically into the queue worker loop via booted() below.
      */
     public static function clearResolverCaches(): void
     {
         self::$submissionCache = [];
         self::$contextCache    = [];
         self::$rorCache        = [];
         self::$localesCache    = [];
     }

     /**
      * Self-contained queue-worker integration: between each worker loop
      * iteration (CLI mode only), flush the per-request static caches so
      * data from the previous job can't bleed into the next one.
      */
     protected static function booted(): void
     {
         app('events')->listen(
             \Illuminate\Queue\Events\Looping::class,
             function (\Illuminate\Queue\Events\Looping $event): void {
                 if (!app()->runningInConsole() || app()->runningUnitTests()) {
                     return;
                 }
                 self::clearResolverCaches();
             }
         );
     }
```

What do you think @ajnyga?

schemas/funder.json
bozana: Would you mind adding it to data citations as well, maybe in a separate commit or PR?

classes/funder/Funder.php
ajnyga: Thanks, I trust Touhidur in these, I do not have that good of a grasp on Eloquent especially when we have limited examples of it in the code at the moment. I will try to implement this! Thanks!

classes/funder/Repository.php
ajnyga: Yes agreed, I would feel more comfortable to have this in a plugin for anyone who needs it.

classes/submission/DAO.php
ajnyga: ok, I will remove it

classes/components/forms/FieldFunder.php
ajnyga: The locales are fetched from the submission using `setLocalesForSubmission` in `useForm`

https://github.com/ajnyga/ui-library/blob/00ed84de6efc330f6a5e97a7a5314fa1d31d24c0/src/managers/FunderManager/useFunderManagerActions.js#L27-L30

https://github.com/pkp/ui-library/blob/b32e250eb91df2edab31f4f861803051618a23e4/src/composables/useForm.js#L359

classes/components/forms/FieldFunderGrants.php
ajnyga: I took example from FieldAuthors.php and seems to be private there as well.

classes/publication/DAO.php
ajnyga: removed

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: time flyes

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: yes I will review the whole migration now that there is a decision on what to use as the data

classes/funder/Funder.php
ajnyga: the fact is that these names are i a different table and not directly in the funder settings

what kind of other solution are you thinking? Affiliations was limited help because it is not using Eloquent - my understanding has been that we do want to move towards using it though?

classes/funder/Funder.php
ajnyga: Added and seems to work very well, thanks for this!

classes/funder/Repository.php
ajnyga: Let me know what you want to do with this:
- do we want to move it to a Zenodo plugin - this might be a more flexible solution if the ROR's that are validated against change due time.
- Or keep it here and just add the safeguards you mention

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: So loading it now as a submodule and reading from the file.

Was thinking that we can of course have that submodule under PKP as well if you want. It is now in tsv-fi account in Github.

classes/funder/Repository.php
bozana: Lets keep it and safeguard it...

bozana: I need to double check if validation is a setting still...

bozana: I see, the validation is a setting, so all fine... lets keep it with those additional safeguards...

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,180 @@
...
+                    ? DB::table('funder_awards_legacy')->get()->groupBy('funder_id')
+                    : collect();
+
+                $submissionLocales = DB::table('submissions')
+                        ->whereIn('submission_id', $legacyFunders->pluck('submission_id'))
+                        ->pluck('locale', 'submission_id');
+
+                $newSettings = [];
+
+                $rorMap = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/rorFundrefData/data/ror_fundref.csv';
```
kaitlinnewson: I think there is a missing import for `use PKP\core\Core;` for this.

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,180 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+
+        // Detect and rename Funding plugin tables by column structure
+        if (Schema::hasTable('funders') && Schema::hasColumn('funders', 'funder_identification')) {
+            Schema::rename('funders', 'funders_legacy');
```
kaitlinnewson, line 53: On my local install this is not dropping the empty table (maybe due to foreign keys), leading to errors.

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,180 @@
...
+        }
+        if (Schema::hasTable('funder_settings') && Schema::hasColumn('funder_settings', 'setting_type')) {
+            Schema::rename('funder_settings', 'funder_settings_legacy');
```
kaitlinnewson, line 56: Same as above, the empty table was not dropped.

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,180 @@
...
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Facades\DB;
+use PKP\migration\Migration;
+
+class I12392_Funders extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
```
kaitlinnewson, line 44: Should data from the previous funding plugin in `plugin_settings` and `versions` also be removed in a migration?

kaitlinnewson: And similarly, if the old plugin was enabled, it would make sense to enable funders metadata automatically as well for that context.

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,341 @@
...
+                'error' => __('api.funders.400.submissionsNotMatched'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? null : ($input['funder']['name'] ?? []),
```
kaitlinnewson: I think instead of null these should be an empty string - otherwise an error is logged (and also above in `add`).

I think this also doesn't properly handle the case where you have an existing funder without a ROR, then delete the funder name inside of "Edit Funder", and re-enter it with a ROR - the old name data persists in the DB and appears in the display.

(I was testing this locally to try things out for zenodo)

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: do you mean that funders_legacy is not dropped after migration?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: Yes these make sense thanks, I will add this

api/v1/funders/PKPFunderController.php
ajnyga: Ok, so this seems to be a problem in using Eloquent. Passing and empty array does not empty the existing setting values like it should.

ajnyga:
Hi @bozana and @touhidurabir 

There seems to be a challenge with using Eloquent and passing an empty array to delete existing values in settings table. This was noticed by @kaitlinnewson.

Scenario:
1. Add a new funder with a custom name without a ROR
2. Edit that funder, delete the custom name and add a new one with a ROR
3. See that the ROR has been saved, but at the same time the old custom name is still there.

So it seems that when we pass an empty array for the name setting, it does not remove existing values. 

I think it should happen here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L131 but probably that is not executed when using Eloquent?

So `$deleteSettings` is set correctly but the check there inside if `is_a($this, EntityDAO::class)` prevents the deletion of being executed.


**Also additional related problem.**

`$deleteSettings` also inclused `grants`even if there is actual data set for grants. it means that if the settings deletion code would run, it would also wipe the grants data.

1. Add new funder with grants data
2. Edit the funder and log around here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L74

So because `grants` has not been changed when the data is edited, it is not present in props there and ends up being subject to deletion. The only reason why it is not deleted though is that `$deleteSettings` is not passed for deletion because of the issue above.

api/v1/funders/PKPFunderController.php
ajnyga: So this was more complex and related to how settings are handled in the code, this seems to be not working as expected with Eloquent. https://github.com/pkp/pkp-lib/pull/12397#issuecomment-4376984368

classes/migration/upgrade/v3_6_0/I12392_Funders.php
kaitlinnewson: My mistake - I checked again and this was due to an problem in my data (a non-existent `submission_id`) which resulted in a failed migration - then it was left in an incomplete state with an empty `funders` table.

touhidurabir:
> Hi @bozana and @touhidurabir
> 
> There seems to be a challenge with using Eloquent and passing an empty array to delete existing values in settings table. This was noticed by @kaitlinnewson.
> 
> Scenario:
> 
> 1. Add a new funder with a custom name without a ROR
> 2. Edit that funder, delete the custom name and add a new one with a ROR
> 3. See that the ROR has been saved, but at the same time the old custom name is still there.
> 
> So it seems that when we pass an empty array for the name setting, it does not remove existing values.
> 
> I think it should happen here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L131 but probably that is not executed when using Eloquent?
> 
> So `$deleteSettings` is set correctly but the check there inside if `is_a($this, EntityDAO::class)` prevents the deletion of being executed.
> 
> **Also additional related problem.**
> 
> `$deleteSettings` also inclused `grants`even if there is actual data set for grants. it means that if the settings deletion code would run, it would also wipe the grants data.
> 
> 1. Add new funder with grants data
> 2. Edit the funder and log around here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L74
> 
> So because `grants` has not been changed when the data is edited, it is not present in props there and ends up being subject to deletion. The only reason why it is not deleted though is that `$deleteSettings` is not passed for deletion because of the issue above.

@ajnyga @bozana As I have tested, this is exactly the sitiation which described above . This will require to have some extra cases to consider but should be handled as different issue . 

Any problem if we consider and assume that passing an empty array as `[]` for a setting prop will have all those related data be deleted for that model instance form the setting table and proceed on ? So we will not stuck here until this issue being resolved as I suspect, this may need to back port for 3.5 as well . I will register an issue . 

Update
issue filed at https://github.com/pkp/pkp-lib/issues/12658

locale/en/submission.po
```diff
@@ -962,6 +962,66 @@ msgstr "For the Editors"
...
+msgstr "Funding"
+
+msgid "submission.funders"
+msgstr "Funders"
+
+msgid "submission.funders.description"
+msgstr "This table allows users to add formal funding information, ensuring funders are properly credited and appear in the publication metadata."
+
+msgid "submission.funders.action.addFunder"
+msgstr "Add a new Funder"
```
kaitlinnewson:
```suggestion
msgstr "Add Funder"
```

kaitlinnewson: (Similarly "Add a new Data Citation" could be simplified to "Add Data Citation" and would be more aligned with other UI labels).

locale/en/submission.po
```diff
@@ -962,6 +962,66 @@ msgstr "For the Editors"
...
 msgstr "For Reviewer Suggestion"
 
+msgid "submission.funding"
+msgstr "Funding"
+
+msgid "submission.funders"
+msgstr "Funders"
+
+msgid "submission.funders.description"
+msgstr "This table allows users to add formal funding information, ensuring funders are properly credited and appear in the publication metadata."
```
kaitlinnewson:
```suggestion
msgstr "Add funding information, ensuring funders are properly credited and appear in the publication metadata."
```

classes/funder/Repository.php
```diff
@@ -0,0 +1,172 @@
...
+ * @class Repository
+ *
+ * @brief A helper class to handle operations with Funder data
+ */
+
+namespace PKP\funder;
+
+use APP\core\Application;
+use APP\core\Request;
+use PKP\context\Context;
```
kaitlinnewson: It looks like this can be removed.

classes/publication/maps/Schema.php
```diff
@@ -21,6 +21,7 @@
 use Illuminate\Support\Enumerable;
 use PKP\context\Context;
 use PKP\dataCitation\DataCitation;
+use PKP\funder\Funder;
```
kaitlinnewson: I think this (and DataCitation above) can be removed.

classes/components/forms/funder/FunderEditForm.php
```diff
@@ -0,0 +1,49 @@
...
+ * @class FunderEditForm
+ *
+ * @ingroup classes_controllers_form
+ *
+ * @brief A preset form for setting a publication's single funder
+ */
+
+namespace PKP\components\forms\funder;
+
+use PKP\components\forms\FieldText;
```
kaitlinnewson: I think this can be removed and is not used.

kaitlinnewson: This does not seem to be used below and can be removed.

classes/components/forms/funder/FunderEditForm.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/**
+ * @file classes/components/form/funder/FunderEditForm.php
+ *
+ * Copyright (c) 2014-2026 Simon Fraser University
```
kaitlinnewson:
```suggestion
 * Copyright (c) 2026 Simon Fraser University
```

(for both lines)

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,180 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I12392_Funders.php
+ *
+ * Copyright (c) 2014-2026 Simon Fraser University
```
kaitlinnewson:
```suggestion
 * Copyright (c) 2026 Simon Fraser University
```

(for both lines)

classes/migration/upgrade/v3_6_0/I12392_Funders.php
ajnyga: I followed an example from orcid plugin migration. I did add the part for removing the versions table entry but begun to think whether the ugrade does this in any case when the Funding plugin is not present anymore?

locale/en/submission.po
ajnyga: "Add a new Data Citation" originates from a mockup I got from Devika and used the same naming convention here as well. But agreed that the shorter form would align with things like Add Galley or Add Contributor

locale/en/submission.po
ajnyga: So same here that it comes from an UI mockup for Data Citations and I have just altered the explanation a bit for Funders.

classes/components/forms/FieldFunder.php
bozana: We can remove supportedFormLocales and primaryLocale from here

classes/migration/upgrade/v3_6_0/I12392_Funders.php
kaitlinnewson: I don't think it happens automatically, and can see some examples in past migrations where plugins are removed in the `versions` table, e.g. for crossref and usage stats. I don't think it causes any issues to be there though.

bozana:
Hi @touhidurabir and @ajnyga, yes, I think we can proceed assuming the two problems will be fixed in that separate issue:
- similarly to EntitiDAO also Eloquent empty arrays for multilingual properties should lead to deleting the settings
- different to EntitiDAO that sends all _data, Eloquent+ModelWithSettings sends only updated properties/data, so this needs to be checked differently, so that missing properties (grants in this example) are not removed

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,257 @@
...
+                    ->insert($mappedResultsChunk->toArray());
+            }            
+
+            // Remove old plugin settings
+            DB::table('plugin_settings')
+                ->where('plugin_name', '=', 'fundingplugin')
+                ->delete();
+
+            // Remove old plugin version entry
+            DB::table('versions')
```
ajnyga, line 292: Hi @kaitlinnewson and @bozana 

For me the upgrade works with OJS 3.5 dataset + some funding plugin data.

However, I begun to think that maybe we should not add this sql for removing the data from versions table? I do not think for example orcid upgrade does this and probably the upgrade process + loading the existing plugin checks if something needs to be removed? 

What I am not sure about is how to best handle people not trying to upgrade while having an outdated funding plugin in their plugins/generic folder?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,257 @@
...
+            // If funding plugin was enabled, enable new core funder support
+            $q = DB::table('plugin_settings')
+                ->where('plugin_name', '=', 'fundingplugin')
+                ->where('context_id', '<>', 0)
+                ->select(['context_id', 'setting_name', 'setting_value']);
+
+            $results = $q->get();
+
+            $enabledContexts = $results
+                ->where('plugin_name', '=', 'fundingplugin')
```
bozana: the plugin_name column is not in the $results

classes/migration/upgrade/v3_6_0/I12392_Funders.php
bozana: Hmmm... Let me double check this in the meeting, but I think that we should also remove the plugin from the file system. I am wondering why have we not done it yet for any other plugin... 🤔

bozana: I double checked and then I also remembered :-) -- The users who upgrade get the new code without old plugins first, so no need to remove the the plugin code from the file system :-) But I think we can remove it from plugin_settings and versions.

ajnyga: Ok, just a note that the orcid migration does not remove the version

schemas/funder.json
```diff
@@ -0,0 +1,94 @@
...
```
bozana, line 8: Could you please sort the properties alphabetically -- so that they can be easier found?

schemas/funder.json
```diff
@@ -0,0 +1,94 @@
...
+			"validation": [
+				"nullable"
+			],
+			"items": {
+				"type": "object",
+				"properties": {
+					"grantDoi": {
+						"type": "string",
+                        "validation": [
+                            "regex:/^\\d+(.\\d+)+\\//"
```
bozana: The . is unescaped. (.\d+)+ matches any character followed by digits, not a literal dot followed by digits.
How about using the regex that we use in pid/Doi.php:
"nullable",
"regex:#^10[.][0-9]{4,}/\\S+$#i"
I think it should also be nullable -- it is not a required field?

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,341 @@
...
+    public function saveOrder(Request $illuminateRequest): JsonResponse
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        $submissionId = (int) $submission->getId();
+        $sequence = $illuminateRequest->json()->all();
+
+        if (!is_array($sequence)) {
+            return response()->json(
+                ['error' => __('api.funders.404.invalidOrderFormat')],
```
bozana: maybe to rename the locale key to be api.funders.**400**.invalidOrderFormat -- bad request is 400.

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /**
+     * Get a single funder.
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $funder = Funder::find((int) $illuminateRequest->route('funderId'));
+
+
```
bozana: double blank line

classes/components/forms/funder/FunderEditForm.php
```diff
@@ -0,0 +1,48 @@
...
+     * Constructor
+     *
+     * @param string $action URL to submit the form to
+     */
+    public function __construct(string $action)
+    {
+        $this->action = $action;
+
+        $this->addField(new FieldFunder('funder', [
+            'label' => __('submission.funders.funder.label.name'),
```
bozana: This locale key does not exist. Maybe you mean `submission.funders.funder`?

classes/migration/install/AffiliationsMigration.php
```diff
@@ -27,6 +27,7 @@ public function up(): void
     {
         Schema::create('author_affiliations', function (Blueprint $table) {
             $table->comment('Author affiliations');
+
```
bozana: this empty line can be removed

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+            $table->index(['funder_id'], 'funder_settings_funder_id');
+            $table->unique(['funder_id', 'locale', 'setting_name'], 'funder_settings_unique');
+        });
+
+        // Migrate legacy funder data if it exists
+        if (Schema::hasTable('funders_legacy')) {
+        
+            DB::transaction(function () {
+    
+                $legacyFunders = DB::table('funders_legacy')->orderBy('funder_id')->get();
```
bozana: Maybe to have something like this here:
```
$legacyFunders = DB::table('funders_legacy')
      ->join('submissions', 'funders_legacy.submission_id', '=', 'submissions.submission_id')
      ->orderBy('funders_legacy.funder_id')
      ->select('funders_legacy.*', 'submissions.locale')
      ->get();
```
In case funders exist even if a submission does not exist any more -- the JOIN would skip orphaned funders. 
Even if this case is rather rare or not possible, this would need to be safe guarded below for `$submissionLocales[$legacyFunder->submission_id]` below, not to get an undefined index error and abort the migration.

Then maybe also to replace `$submissionLocales[$legacyFunder->submission_id]` with `$legacyFunder->locale`, and remove the separate `$submissionLocales` query entirely.

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+                $fundrefToRor = [];
+
+                if (!file_exists($rorMap)) {
+                    $this->_installer->log('WARNING: Funder Registry to ROR mapping file not found at ' . $rorMap . '. Funders will be migrated without ROR mappings.');
+                } else {
+                    $handle = fopen($rorMap, 'r');
+                    fgetcsv($handle); // skip first line
+                    while (($row = fgetcsv($handle)) !== false) {
+                        [$rorId, $fundrefId] = $row;
+                        $fundrefToRor[$fundrefId] = $rorId;
```
bozana: Maybe to get the list of existing fundrefIds and only consider them -- there will probably be less existing fundrefIds than the whole CSV has -- so not to load it all...
Maybe to have something like this above:
```
  // funder_identification was stored as a full Fundref URL e.g. http://dx.doi.org/10.13039/501100001780;
  // basename() extracts the numeric ID that matches the fundref_id column in ror_fundref.csv
  $neededIds = $legacyFunders
      ->pluck('funder_identification')
      ->map(fn($id) => basename($id))
      ->filter()
      ->unique()
      ->flip()
      ->toArray();
```

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+                $submissionLocales = DB::table('submissions')
+                        ->whereIn('submission_id', $legacyFunders->pluck('submission_id'))
+                        ->pluck('locale', 'submission_id');
+
+                $newSettings = [];
+
+                $rorMap = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/rorFundrefData/data/ror_fundref.csv';
+                $fundrefToRor = [];
+
+                if (!file_exists($rorMap)) {
```
bozana, line 117: Maybe to move this out of DB transaction.
Also, because the migration should fail, if the file is not found or fopen fails (returns false). I think in those cases we should throw an Exception and stop the migration, right? -- No point to continue migration if the mapping is missing, right?

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+
+            });
+
+            // Drop legacy funding plugin tables
+            Schema::dropIfExists('funder_award_settings_legacy');
+            Schema::dropIfExists('funder_awards_legacy');
+            Schema::dropIfExists('funder_settings_legacy');
+            Schema::dropIfExists('funders_legacy');
+
+            // If funding plugin was enabled, enable new core funder support
```
bozana, line 247: Instead of fetching all funding plugin settings into memory and then mapping/filtering them in PHP, we could use two queries: one to get the context IDs where the plugin was enabled, and one to get their grant validation setting. The results can be inserted directly without any PHP-level transformation. This makes the intent explicit, eliminates the intermediate $enabledContexts array and in_array() checks, and removes unnecessary chunking — the number of contexts is always small enough, so a single INSERT per setting type is probably safe and simpler. 
So maybe something like:
```
     // If funding plugin was enabled, enable new core funder support
      DB::transaction(function () {
          $enabledContextIds = DB::table('plugin_settings')
              ->where('plugin_name', 'fundingplugin')
              ->where('setting_name', 'enabled')
              ->where('setting_value', '1')
              ->where('context_id', '<>', 0)
              ->pluck('context_id');

          if ($enabledContextIds->isNotEmpty()) {
              // Enable core funder support for contexts that had the plugin enabled
              DB::table($this->settingsTableName)->insert(
                  $enabledContextIds->map(fn($contextId) => [
                      $this->settingsTableKey => $contextId,
                      'locale' => '',
                      'setting_name' => 'funders',
                      'setting_value' => 'enable',
                  ])->all()
              );

              // Migrate grant validation setting only for contexts where the plugin was enabled
              $grantValidationSettings = DB::table('plugin_settings')
                  ->where('plugin_name', 'fundingplugin')
                  ->where('setting_name', 'enableGrantIdValidation')
                  ->whereIn('context_id', $enabledContextIds)
                  ->select(['context_id', 'setting_value'])
                  ->get();

              if ($grantValidationSettings->isNotEmpty()) {
                  DB::table($this->settingsTableName)->insert(
                      $grantValidationSettings->map(fn($row) => [
                          $this->settingsTableKey => $row->context_id,
                          'locale' => '',
                          'setting_name' => 'funderGrantValidation',
                          'setting_value' => $row->setting_value,
                      ])->all()
                  );
              }
          }

```

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+            $table->string('locale', 28)->default('');
+            $table->string('setting_name', 255);
+            $table->mediumText('setting_value')->nullable();
+            $table->foreign('funder_id', 'funder_settings_funder_id')->references('funder_id')->on('funders')->onDelete('cascade');
+            $table->index(['funder_id'], 'funder_settings_funder_id');
+            $table->unique(['funder_id', 'locale', 'setting_name'], 'funder_settings_unique');
+        });
+
+        // Migrate legacy funder data if it exists
+        if (Schema::hasTable('funders_legacy')) {
```
bozana, line 91: I would have a few suggestions how to maybe refactor the code here. Please see also the comments below, that eventually contain more information -- here is only a summary:
1. Fail fast on missing CSV -- throw Exception, don't warn. Since the CSV is essential for the migration's main purpose, a missing file should halt the upgrade. fopen() failure should also be checked explicitly.
2. Thus, move CSV reading before the transaction, with early termination.
3. Get funders using JOIN with submissions -- it eliminates orphaned funders (those whose submission was deleted) and delivers the submission locale in the same row — no extra query needed. We also check if eventually all funders are orphaned for the funders migration.
4. chunkById(500) also by getting funders from DB instead of load-all -- just in case a huge installation would have huge amount of funders, so that we do not keep them all in memory. 
5. Preserve Fundref ID when no ROR mapping is found -- When a funder can't be mapped to a ROR, the original Fundref ID is stored as a fundrefId setting.
6. funder_settings_legacy: throw exception if missing — if funders_legacy exists but  funder_settings_legacy does not, throw immediately before doing any other work, since this indicates an unexpected schema state. Without it, funders that have no ROR mapping would be migrated with no name.
7. Skip funders with no ROR and no name, with a warning. The skipped funders are a data quality issue that existed before the migration — keeping the legacy tables doesn't fix them, it just defers the decision, so I think also here, no need to keep the legacy tables because of them?

So maybe something like:
```
        // Migrate legacy funder data if it exists
        if (Schema::hasTable('funders_legacy')) {

            if (!Schema::hasTable('funder_settings_legacy')) {
                throw new \Exception(
                    'Upgrade failed: could not prepare the funder settings data for migration. ' .
                    'The funder_settings database table may be missing or has an unexpected structure. ' .
                    'Please check your database schema before running this migration.'
                );
            }

            // funder_identification was stored as a full Fundref URL e.g.
            // http://dx.doi.org/10.13039/501100001780; basename() extracts the
            // numeric ID that matches the fundref_id column in ror_fundref.csv.
            // JOIN with submissions excludes orphaned funders from the set of needed mappings.
            $neededIds = DB::table('funders_legacy')
                ->join('submissions', 'funders_legacy.submission_id', '=', 'submissions.submission_id')
                ->pluck('funders_legacy.funder_identification')
                ->map(fn($id) => basename($id))
                ->filter()
                ->unique()
                ->flip()
                ->toArray();

            $fundrefToRor = [];
            if (!empty($neededIds)) {
                $rorMap = Core::getBaseDir() . '/' . PKP_LIB_PATH . '/lib/rorFundrefData/data/ror_fundref.csv';
                if (!file_exists($rorMap)) {
                    throw new \Exception(
                        'ROR-Fundref mapping file not found at ' . $rorMap . '. Please initialize the rorFundrefData submodule.'
                    );
                }
                $handle = fopen($rorMap, 'r');
                if ($handle === false) {
                    throw new \Exception('Could not open ROR-Fundref mapping file at ' . $rorMap . '.');
                }
                fgetcsv($handle); // skip header line
                while (($row = fgetcsv($handle)) !== false) {
                    [$rorId, $fundrefId] = $row;
                    if (isset($neededIds[$fundrefId])) {
                        $fundrefToRor[$fundrefId] = $rorId;
                        if (count($fundrefToRor) === count($neededIds)) {
                            break; // all needed mappings found
                        }
                    }
                }
                fclose($handle);

                $hasAwardsTable = Schema::hasTable('funder_awards_legacy');

                DB::transaction(function () use ($fundrefToRor, $hasAwardsTable) {
                    // JOIN with submissions excludes orphaned funders and provides the submission
                    // locale directly, avoiding a separate query per funder.
                    DB::table('funders_legacy')
                        ->join('submissions', 'funders_legacy.submission_id', '=', 'submissions.submission_id')
                        ->select('funders_legacy.*', 'submissions.locale')
                        ->chunkById(500, function ($chunk) use ($fundrefToRor, $hasAwardsTable) {

                            $funderIds = $chunk->pluck('funder_id');

                            $legacySettings = DB::table('funder_settings_legacy')
                                ->whereIn('funder_id', $funderIds)
                                ->get()
                                ->groupBy('funder_id');

                            $legacyAwards = $hasAwardsTable
                                ? DB::table('funder_awards_legacy')
                                    ->whereIn('funder_id', $funderIds)
                                    ->get()
                                    ->groupBy('funder_id')
                                : collect();

                            $newSettings = [];

                            foreach ($chunk as $legacyFunder) {
                                $fundrefId = basename($legacyFunder->funder_identification);
                                $ror = $fundrefToRor[$fundrefId] ?? null;

                                $funderName = null;
                                foreach ($legacySettings[$legacyFunder->funder_id] ?? [] as $setting) {
                                    if ($setting->setting_name === 'funderName') {
                                        $funderName = $setting->setting_value;
                                        break;
                                    }
                                }

                                if (!$ror && empty($funderName)) {
                                    $this->_installer->log(
                                        'WARNING: Funder (legacy ID ' . $legacyFunder->funder_id .
                                        ', submission ID ' . $legacyFunder->submission_id .
                                        ') has no ROR and no name — skipping migration of this funder.'
                                    );
                                    continue;
                                }

                                $newFunderId = DB::table('funders')->insertGetId([
                                    'submission_id' => $legacyFunder->submission_id,
                                    'ror' => $ror,
                                    'seq' => 0,
                                ]);

                                if (!$ror) {
                                    // No ROR mapping found; preserve name and original Fundref ID
                                    // so it can still be exported to Crossref etc.
                                    $newSettings[] = [
                                        'funder_id' => $newFunderId,
                                        'locale' => $legacyFunder->locale,
                                        'setting_name' => 'name',
                                        'setting_value' => $funderName,
                                    ];
                                    if ($fundrefId) {
                                        $newSettings[] = [
                                            'funder_id' => $newFunderId,
                                            'locale' => '',
                                            'setting_name' => 'fundrefId',
                                            'setting_value' => $fundrefId,
                                        ];
                                    }
                                }
                                // ROR found: name is provided at runtime from ROR data; skip migrating it

                                // Collapse awards into grants JSON
                                $awardNumbers = ($legacyAwards[$legacyFunder->funder_id] ?? collect())
                                    ->pluck('funder_award_number')
                                    ->unique()
                                    ->values();

                                if ($awardNumbers->isNotEmpty()) {
                                    $newSettings[] = [
                                        'funder_id' => $newFunderId,
                                        'locale' => '',
                                        'setting_name' => 'grants',
                                        'setting_value' => json_encode(
                                            $awardNumbers->map(fn($n) => [
                                                'grantDoi' => null,
                                                'grantNumber' => $n,
                                                'grantName' => null,
                                            ])->all()
                                        ),
                                    ];
                                }
                            }

                            foreach (array_chunk($newSettings, 500) as $settingsChunk) {
                                DB::table('funder_settings')->insert($settingsChunk);
                            }

                        }, 'funders_legacy.funder_id', 'funder_id');
                });
            }
            ...
```

classes/migration/upgrade/v3_6_0/I12392_Funders.php
```diff
@@ -0,0 +1,255 @@
...
+        
+            DB::transaction(function () {
+    
+                $legacyFunders = DB::table('funders_legacy')->orderBy('funder_id')->get();
+                
+                $legacySettings = DB::table('funder_settings_legacy')
```
bozana: Theoretically funder_settings_legacy could not exist if for some reason the DB is corrupted, so maybe to check it too, soon in the code, so that admin can investigate.

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? [] : ($input['funder']['name'] ?? []),
+            'grants' => $input['grants'] ?? [],
+            'seq' => 0,
```
kaitlinnewson: Should this consider the existing `seq` value for the funder or be removed?

bozana (commented):
Hi @ajnyga,

A few comments, mostly some to improve the migration script.
Also, I think you would maybe need to add submodule update commit for the ror-fundref-data in the same way you do in OJS for pkp-lib and ui-library.

After the migration script is fully finished, I would also try to test it.

The legacy tables are most valuable in the window right after upgrade, when an admin might want to verify the migration. After that they're just clutter. Thus, I think we should log clearly -- warn about everything needed -- and still drop the tables.
We can preserve the fundrefId if it could not be mapped. 

Thanks a lot for the great and huge work!

api/v1/funders/PKPFunderController.php
ajnyga: @bozana  not quite sure what the purpose of _href here is. If it needs to be an actual working api path, then there is the problem of funder object not having the publicationId. Or can I just have

```
                case '_href':
                    $output[$prop] = $this->getApiUrl(
                        'funders/' . $item->getId(),
                        $this->context->getData('urlPath')
                    );
                    break;
```

classes/migration/install/AffiliationsMigration.php
ajnyga: I do not think this is in my pr :) but removed the line.

api/v1/funders/PKPFunderController.php
bozana: Hi @ajnyga, hmmm... lets leave `_href` out for now -- Erik can then later think about it. Please remove it then also from funders.json.
p.s. this comment actually belongs to my comment above -- just so you do not miss this comment here... :-)

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,339 @@
...
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.400.submissionsNotMatched'),
```
bozana: minor but maybe to change the key to 403 because HTTP_FORBIDDEN is 403

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,339 @@
...
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.400.submissionsNotMatched'),
```
bozana: here as well

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,339 @@
...
+            return response()->json([
+                'error' => __('api.funders.404.funderNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION);
+
+        if ($submission->getId() !== $funder->submissionId) {
+            return response()->json([
+                'error' => __('api.funders.400.submissionsNotMatched'),
```
bozana: and here

classes/funder/Repository.php
```diff
@@ -0,0 +1,177 @@
...
+            $funder,
+            $this->schemaService->getRequiredProps($schema),
+            $this->schemaService->getMultilingualProps($schema),
+            [],
+            ''
+        );
+
+        // Validate grant numbers if funder grant validation is enabled and a ROR ID is provided
+        $context = $this->request->getContext();
+        $funderGrantValidationSetting = (bool) $context->getData('funderGrantValidation');
```
bozana: Maybe to use `$context?` -- getContext() can return null.

classes/funder/Repository.php
```diff
@@ -0,0 +1,177 @@
...
+                            }
+
+                            try {
+                                $awardResponse = $httpClient->request(
+                                    'GET',
+                                    "https://zenodo.org/api/awards?funders={$ror}&q=" . urlencode($grantNumber),
+                                    ['timeout' => 5]
+                                );
+
+                                $body = json_decode($awardResponse->getBody(), true);
```
bozana, line 130: Maybe to add this check here:
```
  if ($body === null) {
      throw new \Exception('Zenodo returned a non-JSON response');
  }
```
This would be catched by the catch block and the validation skipped.
Currently, in that case the grant number will be invalid -- a Zenodo failure becomes a validation error.

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,339 @@
...
+
+        $submissionId = (int) $submission->getId();
+        $sequence = $illuminateRequest->json()->all();
+
+        if (!is_array($sequence)) {
+            return response()->json(
+                ['error' => __('api.funders.400.invalidOrderFormat')],
+                Response::HTTP_BAD_REQUEST
+            );
+        }
```
bozana: Hmmm... Do we need this code? -- `$illuminateRequest->json()->all();` always returns an array? 🤔

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,339 @@
...
+     * Add a funder.
+     */
+    public function add(Request $illuminateRequest): JsonResponse
+    {
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? null : ($input['funder']['name'] ?? []),
```
bozana: this function passes `null` for the name if ROR is set. `edit()` function passes `[]`. It seems then it is handled with `foreach (null as $locale => $localizedValue)` --> PHP warning in `SettingBuilder::getSettingRows()`. Thus, maybe to use `[]` instead ?

api/v1/funders/PKPFunderController.php
bozana: It seem it is the same in PKPDataCitationController

classes/funder/maps/Schema.php
```diff
@@ -0,0 +1,114 @@
...
+        ksort($output);
+        return $this->withExtensions($output, $item);
+    }
+
+    /**
+     * Get a funder grant data model as defined in schemas/funder.json.
+     */
+    public function getGrantDataModel(): array
+    {
+        $schemaService = new PKPSchemaService();
```
bozana: I think we can use `$schema = $this->schemaService->get($this->schema);` -- `$this->schemaService` is already available from the base class — no need for `new PKPSchemaService()`.

schemas/publication.json
bozana, line 149: Shell we remove apiSummary for dataCitations?

bozana (commented):
Hi @ajnyga, I have a few new comments 😅 
And now I will also test the migration...
Huge thanks! Great work!

api/v1/funders/PKPFunderController.php
ajnyga: ok, I am preparing a separate pr to fix these small issues in data citations...

schemas/publication.json
ajnyga: I will do a separate pr for datacitations, I have gathered the small fixes

api/v1/funders/PKPFunderController.php
```diff
@@ -0,0 +1,333 @@
...
+    public function add(Request $illuminateRequest): JsonResponse
+    {
+        $input = $illuminateRequest->input();
+
+        $ror = $input['funder']['ror'] ?? null;
+
+        $params = [
+            'ror' => $ror,
+            'name' => $ror ? [] : ($input['funder']['name'] ?? []),
+            'grants' => $input['grants'] ?? [],
```
bozana: Hi @ajnyga, I have now realized, that if a user clicks `Add` to add a grant, but inserts nothing it is saved as an empty `[]`. This then leads to strange exports.
So maybe to have this here and below in `edit()`:
```
'grants' => array_values(array_filter(                                                                                                                                                                                          
    $input['grants'] ?? [],                                                                                                                                                                                                     
    fn($g) => !empty($g['grantNumber']) || !empty($g['grantDoi']) || !empty($g['grantName'])                                                                                                                                    
)),
```

bozana: I hope then we do not need to check it in all exports

Issue #3477, opened 2018-03-13
Permit editors to see competing interests for declining reviewers


asmecher:
From Laval:

> So, the potential conflict of interest is not available to the editor before the reviewer sent his/her review? If the conflict of interest is the reason why the review would decline the invitation to review, where will the editor see the conflict of interest?

If reviewers are declining requests because of competing interests, it would be nice for the editor to be able to see the statement. Currently reviewer competing interests are only shown with the completed review.

pilasou:
@pkp-dev-workflow Hi, we would like to claim this issue. Before we do, we need some information. Two years ago, we provided a code that solved the problem: https://github.com/pkp/pkp-lib/issues/9400 However, before it was merged, a change was made that prevented the display of the reviewer competing interest declaration (ab83618a113acfda53553a6f54e247da38f81145) by moving the code "in a protected area". What was the reason of this change? What would be a better approach to display the competing interest information? We wlcom any clues or advices. Thank you very much!

asmecher:
@jonasraoni, I think the commits that @pilasou is referring to you were yours... Was the removal of the competing interests in those commits intentional?

pilasou:
Hi @asmecher hi @jonasraoni Just to clarify: the code @nibou230 committed only add the display of the conflct of interest for reviewers that declare a conflit but accept to review. **It does not solve the issue for declining reviewers**. For reviewers that decline an invitation, the problem remains and reveals itselft to be more complex than it seems at first sight. Basicaly, when declining, a new form is called that only contains the email template content as input (not the ones from the previous form, i.e the conflit of interest statement). In other words, all inputs made in the previous form (step 1 form) are not copied to the new form, thus, when confirming decline, all the infirmations (conflit of interest statment, etc.) the reviewer wrote in the previous form are lost.

asmecher:
This is the pathway where the reviewer's Competing Interests aren't saved to the database:

https://github.com/user-attachments/assets/01980007-2f05-41ab-bf6f-c9a17ba40752

This is in fact intentional; when choosing Decline, the reviewer is expected to explain their reasons in the resulting email compose window, and their competing interests are not considered relevant.

I could see value in either...
- Moving the CI field to the next page of the reviewer interaction (so it's only possible to enter after the review is accepted), or
- Changing the code so that the competing interests are stored when the review is declined

...and I would prefer the second, as it seems useful at least for Laval. However, one of these needs to be addressed before we can merge the attached PR.

[...]

PR #12400, opened 2026-03-02
pkp/pkp-lib#3477 Show competing interests for declining reviewers

nibou230:
Placed the competing interests outside the restricted area, which was requiring a date completed value.

Resolves the ticket #3477.

### DIFF
--- a/templates/reviewer/review/modal/regretMessage.tpl
+++ b/templates/reviewer/review/modal/regretMessage.tpl
@@ -13,11 +13,18 @@
 	$(function() {ldelim}
 		// Attach the form handler.
 		$('#declineReviewForm').pkpHandler('$.pkp.controllers.form.AjaxFormHandler');
+		// Copy competing interests from the still-live Step 1 form into this decline form.
+		$('#declineCompetingInterestOption').val($('input[name="competingInterestOption"]:checked').val() || 'noCompetingInterests');
+		var ciTextarea = $('textarea[name="reviewerCompetingInterests"]');
+		var ciEditor = (typeof tinymce !== 'undefined' && ciTextarea.length) ? tinymce.get(ciTextarea.attr('id')) : null;
+		$('#declineReviewerCompetingInterests').val(ciEditor ? ciEditor.getContent() : ciTextarea.val());
 	{rdelim});
 </script>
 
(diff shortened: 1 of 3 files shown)


templates/reviewer/review/modal/regretMessage.tpl
```diff
@@ -13,11 +13,18 @@
 	$(function() {ldelim}
 		// Attach the form handler.
 		$('#declineReviewForm').pkpHandler('$.pkp.controllers.form.AjaxFormHandler');
+		// Copy competing interests from the still-live Step 1 form into this decline form.
+		$('#declineCompetingInterestOption').val($('input[name="competingInterestOption"]:checked').val() || 'noCompetingInterests');
+		var ciTextarea = $('textarea[name="reviewerCompetingInterests"]');
+		var ciEditor = (typeof tinymce !== 'undefined' && ciTextarea.length) ? tinymce.get(ciTextarea.attr('id')) : null;
```
asmecher, line 19: At a glance, it's not clear why both of these form styles are necessary... Is this working around a potential race condition for form handling and modal loading or something?

templates/reviewer/review/modal/regretMessage.tpl
```diff
@@ -13,11 +13,18 @@
...
+		$('#declineReviewerCompetingInterests').val(ciEditor ? ciEditor.getContent() : ciTextarea.val());
 	{rdelim});
 </script>
 
 <form class="pkp_form" id="declineReviewForm" method="post" action="{url op="saveDeclineReview" path=$submissionId|escape}">
 	{csrf}
+	<input type="hidden" id="declineCompetingInterestOption" name="competingInterestOption" value="">
```
asmecher, line 26: Is this input necessary? It looks like it's set but not used.

nibou230: It is used in the handler (PKPReviewerHandler) to make sure we save the competing interests only when the option is checked.

`$request->getUserVar('competingInterestOption') === 'hasCompetingInterests'`

templates/reviewer/review/modal/regretMessage.tpl
nibou230: From what I remember, yes TinyMCE is messing with the input (hiding the original textarea) when the component is initialized. The part `ciTextarea.val()` is a fallback, just in case something weird happens.

Issue #12332, opened 2026-02-12
[OJS] | Account Invites - Microsoft auto declines invitations


amenbreak:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Microsoft (Safe Links?) automatically declines reviewer invitations before the intended recipient can accept them. The email security scanner loads both the accept and decline invitation URLs when scanning emails, renders the decline confirmation page, and automatically clicks the "Confirm Decline Invitation" button via GET request.

This occurs despite the fix implemented in [#11690](https://github.com/pkp/pkp-lib/issues/11690) which added a two-step decline confirmation process. While that fix prevented immediate auto-decline from initial email scanning, it didn't protect the confirmDecline endpoint from automated clicking.

### Steps to Reproduce

I don't have a Safe Links environment to reproduce the exact steps, but can confirm the following in server logs:

1. Send "Invitation to New Role" email
2. Safelinks (Microsoft ip):
 Loads both /invitation/accept and /invitation/decline URLs (same second)
 30-120 seconds later, automatically clicks the "Confirm Decline Invitation" button
 GET request to /invitation/confirmDecline?id=X&key=Y

4. Invitation status changes to DECLINED
5. Real user later clicks "Accept" link → receives 404 error

### Expected Result

Only human users with genuine intent should be able to decline invitations.

### Actual Result

Safelinks scanner automatically declines invitations by following the GET-based confirmDecline link and user sees 404 upon clicking "Accept Invitation"

### Environment Details

OS platform | Linux
PHP version | 8.3.6
Apache version | nginx/1.24.0

### Application Version

OJS 3.5.0.3

### Logs

Multiple instances showing the pattern (IPs in 135.232.20.x range are Microsoft infrastructure):

```
# Invitation ID 62 - Email scanned by Microsoft (IP: 135.232.20.13)
[10/Feb/2026:13:13:44] 135.232.20.13 GET /invitation/accept?id=62 200
[10/Feb/2026:13:13:44] 135.232.20.13 GET /invitation/decline?id=62 200  ← Both same second
[10/Feb/2026:13:13:45] 135.232.20.13 GET /$$$call$$$/page/page/css?v=3.5.0.3 200  ← Decline page rendering
[10/Feb/2026:13:13:45] 135.232.20.13 GET /api/v1/_i18n/ui.js 307
[10/Feb/2026:13:13:48] 135.232.20.13 GET /api/v1/invitations/62 307
[...]

Issue #12208, opened 2026-01-15
Invitation accept link shows generic 404 after it has been used/expired (should show “link no longer valid” message)


Denzellee08:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When a user receives an invitation link (ex: role assignment or reviewer invitation) like:

/index.php/<journalPath>/invitation/accept?id=97&key=RA4353

The link works the first time, but if the user clicks it again (or returns later after it has been accepted/declined/expired), it shows a generic 404 page.

This is confusing for end users because it looks like the journal site is broken.

### Steps to Reproduce

As an admin, invite an existing user to a new role or invite a reviewer.

Open the invitation email and click the “Accept” link.

Click the same “Accept” link again (or refresh / open it in a second browser).

### Expected Result

Instead of a generic 404, show a dedicated page/message such as:

“This invitation link is no longer valid (already accepted/declined/expired). Please log in to your account or request a new invitation.”

### Actual Result

A generic 404 page is shown.

### Environment Details

OS platform: Linux
PHP version:8.3.28
Database: mysql

### Application Version

OJS version:3.5.0.2

### Update
A fix has been applied as part of https://github.com/pkp/pkp-lib/issues/12332

touhidurabir:
@Tribunal33 ready for QA check .

Tribunal33:
@touhidurabir, to be clear this wasn't a bug and 404 was the expected outcome.  It was descoped due to time.  This needs to be in a format for a minor feature enhancement with acceptance criteria.

I have some questions.

When logged in as a different user (like admin) should this email prompt a warning to logout first?  Currently it doesn't.
What should the login button do when you are already logged in?
What should the register button do when you are already registered?

<img width="2557" height="483" alt="Image" src="https://github.com/user-attachments/assets/23bc789a-547b-438e-abe4-cd132d129016" />

touhidurabir:
@Tribunal33 

> When logged in as a different user (like admin) should this email prompt a warning to logout first? Currently it doesn't.

It can but I intentionally skip the complexity of that behaviour as it's an url that no longer is valid . So instead of a generic 404 which mislead to false error, it's just a generic message to let them know it's no longer is valid . Same for logged/non-logged in user . 

> What should the login button do when you are already logged in?

It should redirect to submission lists basically what is default behaviour if user already logged in and try to visit the login page again .  And it does that . 

> What should the register button do when you are already registered?

[...]

PR #12403, opened 2026-03-03
pkp/pkp-lib#12332 MS aggressive safe link prevention by POST reqeust

touhidurabir:
for pkp/pkp-lib#12332 .

### DIFF
--- a/templates/invitation/declineInvitation.tpl
+++ b/templates/invitation/declineInvitation.tpl
@@ -18,9 +18,7 @@
         </p>
         <form method="post" action="{$declineUrl}">
             {csrf}
-            <button type="submit" class="pkpButton inline-flex relative items-center gap-x-1 text-lg-semibold text-primary border-light hover:text-hover bg-secondary py-[0.4375rem] px-3 border rounded">
-                {translate key="invitation.decline.confirm"}
-            </button>
+            {fbvElement type="submit" id="declineInvitationSubmit" label="invitation.decline.confirm"}
         </form>
     </div>
 {/block}


pages/invitation/InvitationHandler.php
```diff
@@ -72,10 +88,21 @@ private function getInvitationByKey(Request $request): Invitation
...
+        if ($id && $key) {
+            $expiredInvitation = Repo::invitation()->getById($id)?->invitationModel;
+
+            if ($expiredInvitation && password_verify($key, $expiredInvitation->keyHash)) {
+                $this->displayInvitationNotAvailablePage($request);
+            }
         }
-        return $invitation;
+
+        throw new NotFoundHttpException();
```
touhidurabir, line 105: this provide a solution for https://github.com/pkp/pkp-lib/issues/12208

templates/invitation/declineInvitation.tpl
```diff
@@ -16,11 +16,11 @@
...
         </p>
-        <pkp-button
-            element="a"
-            href="{$declineUrl}"
-        >
-            {translate key="invitation.decline.confirm"}
-        </pkp-button>
+        <form method="post" action="{$declineUrl}">
+            {csrf}
+            <button type="submit" class="pkpButton inline-flex relative items-center gap-x-1 text-lg-semibold text-primary border-light hover:text-hover bg-secondary py-[0.4375rem] px-3 border rounded">
```
defstat: @touhidurabir Not sure about this hard-coded css button instead of `pkp-button`. But I leave it to your desission

touhidurabir: problem with our vue component `pkp-button` is that it does not support submit action and need modification to ui lib which I am trying to avoid for this very small case . But you are right so I updated it with our form builder which works here .

Issue #12415, opened 2026-03-05
My Queue submission search by author not working


MrRob100:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

My Queue submission search by author not working when the user has active review assignments. 

### Steps to Reproduce

1. Login with lbarbieri on the 3.4 test site
2. look at the submissions list: https://ojs34.testdrive.publicknowledgeproject.org/index.php/testdrive-journal/submissions
3. Search for 'Willinsky'
4. Observe that the submission with author: Willinsky does not show 

### Expected Result

The search works

### Actual Result

Searching by author doesn't work for this user

### Environment Details

<img width="1473" height="719" alt="Image" src="https://github.com/user-attachments/assets/9a2e1c57-97e4-48b4-8c72-7f69ec0eab59" />

<img width="1471" height="609" alt="Image" src="https://github.com/user-attachments/assets/b9bb1786-3de3-4645-9890-f1a0af02b4cf" />

### Application Version

OJS 3.4 (the logic been re-written and fixed in 3.5)

PR: https://github.com/pkp/pkp-lib/pull/12416

Vitaliy-1:
The problem appears in dual assignments, when a user having an editorial assignment has also any review assignment. E.g., user with a `SUB_EDITOR` role has a stage assignment and review assignment.

The current behaviour is not allowing search by author names in case of user having any review assignment. I think, we can add following exceptions
1. If user has any stage assignment, allow the search. We disclose author names anyway
2. if the review assignment is not double blind, allow the search. The reasoning is the same

@Devika008, @jardakotesovec do you have any considerations on this?

Tribunal33:
@MrRob100 I noticed that the screenshots you provided there is an incorrect spelling of the author "Willinksy" not "Willinsky" so there is a mismatch between the search and author.  When I logged in and used the search on the test drive site it works as expected.  I could find both `Willinsky` and `Willinksy` as authors from the list.

<img width="1555" height="668" alt="Image" src="https://github.com/user-attachments/assets/d2006029-2383-4f7b-868a-b3c833453a09" />

MrRob100:
Good spot @Tribunal33 . Bad example from me. I have recreated another one. By logging in as lbarbieri again, there should be 2 submissions in their Queue. None show up in the search. They do when logged in as admin

As admin:
[...]

PR #12416, opened 2026-03-05
fix(search): allow author name search in editorial dashboard for user…

MrRob100:
This affects authors with review assignments

The Collector's searchPhrase logic blocked author name search for any user with assignedTo set who also had any review_assignment row, even historically. This was overly broad for the editorial dashboard where editors need to find submissions by author name.

Adds an allowAuthorSearch flag (default false, preserving existing behaviour everywhere) that PKPBackendSubmissionsHandler opts into, bypassing the reviewer restriction only for the /_submissions endpoint.

https://github.com/pkp/pkp-lib/issues/12415


MrRob100:
@Vitaliy-1 Thank you for the review, changes are in

PR #12423, opened 2026-03-06
Rename privacy consent id on reviewer's page

blesildaramirez:


cypress/support/commands.js
```diff
@@ -629,7 +629,7 @@ Cypress.Commands.add('performReview', (username, password, title, recommendation
 	cy.contains('table tr', title).within(() => {
 		cy.get('button').click()
 	})
-	cy.get('input[id="privacyConsent"]').click();
+	cy.get('input[id="dataCollectionCheck"]').click();
```
jardakotesovec: Looks good.. just minor suggestion - if the id is problem and its used only in cypress test - we could go with name="privacyConsent" in the cypress test and basically just change id to name.

Issue #12339, opened 2026-02-12
Support secure connections to the database server


asmecher:
OJS/OMP/OPS currently do not support secure connections to the database server.

The European Commission's Web Application Security Standard require this:

> 55. Communication of sensitive non-classified information between applications and underlying services shall be encrypted.

It has also been requested on the support forum, e.g.: https://forum.pkp.sfu.ca/t/ojs-3-5-and-mysql-ssl-tls-database-connections/97795

It should be trivial to add support for secure connections, as we currently just forward our `config.inc.php` configuration into Laravel's expected form:

https://github.com/pkp/pkp-lib/blob/stable-3_5_0/classes/core/PKPContainer.php#L359..L372

Add support for the additional options required for secure connections via Laravel's toolset, and document them in `config.TEMPLATE.inc.php`.

#### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12429
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5397
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1238
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2271

touhidurabir:
@asmecher ready for look at https://github.com/pkp/pkp-lib/issues/12339#issue-3934391614 . once finalised, I will forward port to OMP and OPS .

asmecher:
I've added a review [here](https://github.com/pkp/pkp-lib/pull/12429#pullrequestreview-3946744055).

touhidurabir:
@asmecher ready for final look of PRs at https://github.com/pkp/pkp-lib/issues/12339#issue-3934391614 . One specific note about adding the config in the install form at https://github.com/pkp/pkp-lib/pull/12429#issuecomment-4068303853 .

asmecher:
All merged, thanks! Skipping QA -- this is setup intensive and fairly simple, and I consider it adequately tested.

PR #12429, opened 2026-03-09
pkp/pkp-lib#12339 Database SSL/TLS config options

touhidurabir:
for pkp/pkp-lib#12339 .


classes/plugins/PluginHelper.php
```diff
@@ -164,7 +164,9 @@ protected function _getConnectionParams(): array
             'unixSocket' => Config::getVar('database', 'unix_socket'),
             'databaseUsername' => Config::getVar('database', 'username'),
             'databasePassword' => Config::getVar('database', 'password'),
-            'databaseName' => Config::getVar('database', 'name')
+            'databaseName' => Config::getVar('database', 'name'),
```
asmecher: I don't think these parameters are ever used. In fact, I'd be surprised if this whole function isn't dead code (since we removed ADODB for schema management). Did you find a use for these that I'm not aware of?

classes/install/PKPInstall.php
```diff
@@ -100,6 +100,15 @@ public function preInstall()
             'charset' => 'utf8',
             'collation' => 'utf8_general_ci',
         ];
+
+        // Apply SSL/TLS options from config.inc.php if enabled
+        if (Config::getVar('database', 'secure', false)) {
+            $config['connections'][$driver] = PKPContainer::addDatabaseSslOptionsToConnection(
```
asmecher: `PKPContainer::addDatabaseSslOptionsToConnection` gets parameters from `Config::getVar`, but this class should use `$this->getParam` as above.

touhidurabir:
> I think these new parameters need to be added to the installation form and `lib/pkp/classes/cliTool/InstallTool.php` so that users requiring an SSL connection on install can proceed.

@asmecher  I initially thought so but then intentionally back out form that as that make the scope much bigger and bit more complex . Also DB SSL/TLS connection is kind of an advanced mechanism which is likely to have use case for only a handful installation, so perhaps we can document that to have this feature set up at install phase, one needed to update the copied config.inc.php before start the installation phase . Also it's not mandatory unless DB server is set up only to support SSL/TLS . This idea does comes from the WP . 

> Edit: on further thought, I'm not so sure about the installation form. Maybe check Wordpress or Drupal to see what they do?

Yes, keeping it as it is now does comes from my experience from WP and Drupal also seems to does not provide any options in their web installer to configure such options . If needed, one need to do a manual update of corresponding file before initiating the web installer . But we can do it anyway with some modification which will increase the scope of this task .

classes/plugins/PluginHelper.php
touhidurabir: you are right . upon checking it more, it not needed but we need to update the PKPInstall also and wrap the whole DB config under `if ($this->isPlugin)` check but that seems be an enhancement beyond the scope of this task . But doable .

classes/install/PKPInstall.php
touhidurabir: It depends on if we allow the options in the install form, otherwise with current setup, we need to get those from the config anyway given the how the plugin install/upgrade set up unless we configure the PKPInstall based on https://github.com/pkp/pkp-lib/pull/12429#discussion_r2933557188

Issue #12309, opened 2026-02-07
Authors cannot access workflow on some servers


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

On some servers, authors cannot access the workflow after completing the submission process. They see a message on the workflow page: "You don't currently have access to that stage of the workflow."

See:
- https://forum.pkp.sfu.ca/t/ojs-3-5-0-1-you-dont-currently-have-access-to-that-stage-of-the-workflow/97479
- https://forum.pkp.sfu.ca/t/author-unable-to-add-revised-version-to-original-submission-after-revisions-requested/97311

This is similar to https://github.com/pkp/pkp-lib/issues/10034 and can be emulated following the steps below.

### Steps to Reproduce

1. Emulate the problematic environment:
   ```diff
   diff --git a/classes/core/PKPApplication.php b/classes/core/PKPApplication.php
   index 7e92aad6ad..7decb35c8b 100644
   --- a/classes/core/PKPApplication.php
   +++ b/classes/core/PKPApplication.php
   @@ -275,6 +275,9 @@ abstract class PKPApplication implements iPKPApplicationInfoProvider
                    error_log($count++ . ": Database query\n{$query->sql}\n" . json_encode($query->bindings));
                });
            }
   +\Illuminate\Support\Facades\Schema::getConnection()->getPdo()->setAttribute(\PDO::ATTR_EMULATE_PREPARES,true);
   +\Illuminate\Support\Facades\Schema::getConnection()->getPdo()->setAttribute(\PDO::ATTR_STRINGIFY_FETCHES,true);
   +
        }
    
        /**
   ```
2. Log into a submission as author
3. In the submission workflow, see the message: "You don't currently have access to that stage of the workflow."

### Expected Result

The author should be able to view the submission workflow for their submissions.

### Actual Result

The author sees only the message: "You don't currently have access to that stage of the workflow."

### Environment Details

It's not clear what server configuration is responsible; however, the configuration can be emulated per the instructions above.

### Application Version

OJS 3.5.0-x

### Additional Information

The problem server appears to be using the mysqlnd driver.

asmecher:
Confirmed fixed (both on local dev machine and problem remote). Committed to `stable-3_5_0` and `main`; closing without QA.

asmecher:
[...]

PR #12430, opened 2026-03-09
pkp/pkp-lib#4142 Process email replies into discussions

Hafsa-Naeem:


classes/editorialTask/Repository.php
```diff
@@ -113,8 +124,24 @@ public function addQuery(int $submissionId, int $stageId, string $title, string
...
+
+        $mailableMap = [
+            WORKFLOW_STAGE_ID_SUBMISSION => DiscussionSubmission::class,
+            WORKFLOW_STAGE_ID_INTERNAL_REVIEW => DiscussionReview::class,
+            WORKFLOW_STAGE_ID_EXTERNAL_REVIEW => DiscussionReview::class,
+            WORKFLOW_STAGE_ID_EDITING => DiscussionCopyediting::class,
+            WORKFLOW_STAGE_ID_PRODUCTION => DiscussionProduction::class,
+        ];
+
+        $mailableClass = $mailableMap[$stageId] ?? null;
```
asmecher, line 141: Use `match` to ensure an exception is thrown if no match is made:
```suggestion
        $mailableMap = match($stageId) {
            WORKFLOW_STAGE_ID_SUBMISSION => DiscussionSubmission::class,
            WORKFLOW_STAGE_ID_INTERNAL_REVIEW => DiscussionReview::class,
            WORKFLOW_STAGE_ID_EXTERNAL_REVIEW => DiscussionReview::class,
            WORKFLOW_STAGE_ID_EDITING => DiscussionCopyediting::class,
            WORKFLOW_STAGE_ID_PRODUCTION => DiscussionProduction::class,
        };
```

classes/editorialTask/Repository.php
```diff
@@ -250,6 +287,149 @@ public function deleteBySubmissionId(int $submissionId): void
...
+            }
+
+            $submission = Repo::submission()->get($task->assocId);
+            if (!$submission) {
+                return;
+            }
+
+            $application = Application::get();
+            $request = $application->getRequest();
+            $context = $request->getContext();
```
asmecher, line 315: As above, don't rely on the request to get the context.

classes/editorialTask/Repository.php
```diff
@@ -113,8 +124,24 @@ public function addQuery(int $submissionId, int $stageId, string $title, string
         /** @var NotificationSubscriptionSettingsDAO $notificationSubscriptionSettingsDao */
         $notificationSubscriptionSettingsDao = DAORegistry::getDAO('NotificationSubscriptionSettingsDAO');
 
+        // need submission + context + stage mailables to send capturable email
+        $submission = Repo::submission()->get($submissionId);
+        $application = Application::get();
+        $request = $application->getRequest();
+        $context = $request?->getContext();
```
asmecher, line 131: We shouldn't be relying on `Request::getContext` in model code; this might be a task runner or a job related to another context. You're given the `$contextId` as input to the function; use that.

classes/editorialTask/Repository.php
```diff
@@ -123,7 +150,8 @@ public function addQuery(int $submissionId, int $stageId, string $title, string
                 Notification::NOTIFICATION_LEVEL_TASK
             );
 
-            if (!$sendEmail) {
+            if (
+                !$sendEmail|| !$notification || !$mailableClass || !$submission || !$context) {
```
asmecher, line 154: There is a mixture of defensive programming (what if the submission or context don't exist) and business logic (what if `$sendEmail` is false or the notification is disabled).
- Move the checks on the `$submission` and `$context` object out of the loop, to where they're fetched in the code above. Personally, I think we should be using exceptions rather than stubbing out the function with returns -- but either would be OK.
- As above, use `match` to handle the `mailableClass` check
- Keep the business logic checks here: `if (!$sendEmail || !$notification)` here in the loop where they should be.

classes/editorialTask/Repository.php
```diff
@@ -250,6 +287,149 @@ public function deleteBySubmissionId(int $submissionId): void
...
+            if (!$sender) {
+                return;
+            }
+
+            $headNote = Repo::note()->getHeadNote($task->id);
+            $threadAnchorMessageId = $headNote?->messageId;
+            $title = $headNote?->title ?: $task->title;
+            $subject = $title
+                ? __('common.re') . ' ' . $title
+                : __('common.re');
```
asmecher, line 330: The order of string concatenation might be different in other languages, so it's better to do this in the locale file using variables.

The `common.re` locale key also appears to be unused -- going all the way back to OJS 2.4.x!

I'd suggest...
- Edit the English `common.re` key to make it use a variable, e.g.:
   ```
   msgid "common.re"
   msgstr "Re: {$subject}"
   ```
- Specify the title as a parameter to the translate call
- Edit non-English locales to add the `{$subject}` variable as best you can, but mark them `#, fuzzy`
- If you're not sure about where to put the variable for any non-English locales (e.g. Chinese), delete the key to force re-translation.

classes/editorialTask/Repository.php
```diff
@@ -250,6 +287,149 @@ public function deleteBySubmissionId(int $submissionId): void
...
+                WORKFLOW_STAGE_ID_EXTERNAL_REVIEW => DiscussionReview::class,
+                WORKFLOW_STAGE_ID_EDITING => DiscussionCopyediting::class,
+                WORKFLOW_STAGE_ID_PRODUCTION => DiscussionProduction::class,
+            ];
+
+            if (!isset($mailableMap[$task->stageId])) {
+                return;
+            }
+
+            $mailableClass = $mailableMap[$task->stageId];
```
asmecher, line 363: Again, I suggest using `match` to make this more concise (and it'll throw an exception when there's no match).

classes/editorialTask/Repository.php
```diff
@@ -250,6 +287,149 @@ public function deleteBySubmissionId(int $submissionId): void
...
+                ? __('common.re') . ' ' . $title
+                : __('common.re');
+
+            $participantIds = $task->participants()
+                ->pluck('user_id')
+                ->all();
+
+            if (empty($participantIds)) {
+                return;
+            }
```
asmecher, line 338: `$participantIds` is only used in a `foreach` below; this isn't necessary.

classes/editorialTask/Repository.php
```diff
@@ -250,6 +287,149 @@ public function deleteBySubmissionId(int $submissionId): void
...
+            ];
+
+            if (!isset($mailableMap[$task->stageId])) {
+                return;
+            }
+
+            $mailableClass = $mailableMap[$task->stageId];
+
+            foreach ($participantIds as $userId) {
+                if ($userId === $sender->getId()) {
```
asmecher, line 366: Be careful using `===` comparisons on numeric data fetched from the DB. On some server configurations -- we haven't been able to pin it down -- the DB sends numeric data back to PHP as `string` data rather than `int`. Depending on whether we're using typehints, or Eloquent vs. old DAOs vs. Repository patterns, it's hard to know which type will appear on that environment. Until we know more about the cause, and while we have mixed patterns, it's safest to use `==` comparisons in cases like this to avoid rare and hard-to-find problems.

See e.g.: https://github.com/pkp/pkp-lib/issues/12309

classes/mail/traits/CapturableReply.php
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        return new Headers(
+            messageId: $this->messageId ?: null,
+            references: $this->references,
+            text: $textHeaders,
+        );
+    }
+
+    public function setupCapturableReply()
```
asmecher, line 46: Needs self-documentation

classes/mail/traits/CapturableReply.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        if ($replyTo = Config::getVar('email', 'reply_to_address')) {
+            $this->replyTo($replyTo);
+        }
+    }
+
+    public function allowCapturableReply(
```
asmecher, line 53: Needs self-documentation

classes/migration/upgrade/v3_6_0/I4142_AddMessageIdToNotesAndReviewAssignments.php
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        if (!Schema::hasColumn('review_assignments', 'message_id')) {
+            Schema::table('review_assignments', function (Blueprint $table) {
+                $table->string('message_id', 255)->nullable();
+                $table->index(['message_id'], 'review_assignments_message_id');
+            });
+        }
+    }
+
```
asmecher, line 28: Since this PR doesn't implement review invitations, this column should be removed (and the migration renamed).

classes/note/Note.php
```diff
@@ -57,11 +59,25 @@ protected function casts(): array
...
-            DB::table(Repo::submissionFile()->dao->table)->where('assoc_type', '=', PKPApplication::ASSOC_TYPE_NOTE)
+            DB::table(Repo::submissionFile()->dao->table)
+                ->where('assoc_type', '=', PKPApplication::ASSOC_TYPE_NOTE)
                 ->where('assoc_id', '=', $note->id)
                 ->delete();
         });
+
+        static::created(function (Note $note) {
+            Repo::editorialTask()->notifyParticipantsOnNote($note);
+        });
```
asmecher, line 80: I'm not sure we should be implementing business logic in the storage (model) code. This leads to cases like [this](https://github.com/pkp/pkp-lib/pull/12430/changes#diff-4801fbe579119e2b682a4a458f6d3312cd88a936b16a69c30c2a6beeacfe553cR298), where we need to double-check whether this is a head note or not; if we just left the notifications to the controller code, it wouldn't be necessary. Instead, any calling code creating a note would need to call the business logic required to send the notifications. I think that would probably be cleaner.

classes/note/Note.php
```diff
@@ -122,6 +138,22 @@ public function getData(?string $field)
...
+    {
+        $suffix = (string) Config::getVar('email', 'message_id_suffix');
+
+        if ($suffix === '') {
+            $suffix = '@localhost';
+        } elseif ($suffix[0] !== '@') {
+            $suffix = '@' . ltrim($suffix, '@');
+        }
+
+        return 'note-' . (new \Random\Randomizer())->getInt(0, PHP_INT_MAX) . $suffix;
```
asmecher, line 154: You can use the 3rd parameter to `Config::getVar` to specify a default. Also, I prefer clarity over what should be used in a configuration option (i.e. it should not include the `@` part), rather than code-space work-arounds to standardize things.

```suggestion
        $suffix = Config::getVar('email', 'message_id_suffix', 'localhost');
        return 'note-' . (new \Random\Randomizer())->getInt(0, PHP_INT_MAX) . '@' . $suffix;
```

classes/editorialTask/Repository.php
```diff
@@ -46,7 +51,7 @@ public function countOpenPerStage(int $submissionId, ?array $participantIds = nu
     {
         $counts = EditorialTask::withAssoc(Application::ASSOC_TYPE_SUBMISSION, $submissionId)
             ->when($participantIds !== null, function ($q) use ($participantIds) {
-                $q->withUserIds($participantIds);
+                $q->withParticipantIds($participantIds);
```
asmecher, line 54: This looks like a bug fix, not part of the feature; if so, could you fix this directly in `main` and bring it out of the PR?

classes/task/CheckEmail.php
```diff
@@ -0,0 +1,445 @@
...
+        return __('admin.scheduledTask.checkEmail');
+    }
+
+    private function dbg(string $msg): void
+    {
+        $line = '[' . date('Y-m-d H:i:s') . '] [CheckEmail] ' . $msg . PHP_EOL;
+        // Always write to a temp file – ignores whatever scheduler does with output
+        @file_put_contents(sys_get_temp_dir() . '/checkemail_debug.log', $line, FILE_APPEND);
+    }
+
```
asmecher, line 44: It looks like this is always called, and will always leave a file at a hard-coded location. If this is temporary (debugging code), remove it; if it's intended to be left in place, it should be possible to disable, and should use `error_log` rather than a `file_put_contents` call. (I see there are also some `error_log` calls that are not related to error conditions below; these should also be made possible to disable, if kept in the PR.)

classes/task/CheckEmail.php
```diff
@@ -0,0 +1,445 @@
...
+    
+    private function normalizeMessageId(?string $id): ?string
```
asmecher, line 46: Needs self-documentation (also below)

classes/task/CheckEmail.php
```diff
@@ -0,0 +1,445 @@
...
+                error_log("CheckEmail: exception for message {$messageId}: " . $e->getMessage());
+            }
+        }
+
+        return true;
+    }
+
+
+
+    protected function handleReviewConfirmation(Imap $imap, string $messageId, array $headers, array &$messageParts = null): bool
```
asmecher, line 127: This should be removed from the PR if reviewer invites are not included in the feature.

composer.json
```diff
@@ -29,15 +29,18 @@
 		"mpdf/mpdf": "^8.2",
 		"opensearch-project/opensearch-php": "^2.4",
 		"phpmailer/phpmailer": "6.*",
+		"popphp/pop-mail": "^4.0",
 		"ralouphie/getallheaders": "*",
 		"sentry/sentry": "^4.16",
+		"sabre/vobject": "^4.5",
```
asmecher, line 35: I don't think this is used unless reviewer invites are implemented.

Issue #12374, opened 2026-02-20
[Users & Roles] | Invite new user - Email link redirect new user, with Site Access options setting for users who must be registered enabled user cannot be registered


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

From forum post: https://forum.pkp.sfu.ca/t/problem-inviting-new-user-unless-journal-is-public-ojs-3-5-0-3/97812 
With a site setting enabled the new invite user workflow doesn't work.  The email invitation will instead go to a login page and can't access the invited user registration workflow.

<img width="697" height="157" alt="Image" src="https://github.com/user-attachments/assets/289d712e-66b9-498c-924c-cfc1bda3a7af" />

### Steps to Reproduce

1. Precondition: must have setting enabled for Users & Roles > Site Access Options > Site Access > Users must be registered and log in to view the journal site.
2. Login with JM/Admin role
3. Invite a user - Users & Roles > Invite to a role and complete this process for a new user
4. Either logout of current user or use Incognito mode when accepting the invite.
5. Go to email client and find invite.
6. Accept the Invite (incognito mode)

### Expected Result

Should be redirected to a invitation process

### Actual Result

Are redirected back to login page.

### Application Version

OJS 3.5.0-3 and main

**Pull Requests**
(main)
pkp-lib: https://github.com/pkp/pkp-lib/pull/12433
ojs(submodule update): https://github.com/pkp/ojs/pull/5403
omp(submodule update): https://github.com/pkp/omp/pull/2258
ops(submodule update): https://github.com/pkp/ops/pull/1223

(stable-3.5)
pkp-lib: https://github.com/pkp/pkp-lib/pull/12484
ojs: https://github.com/pkp/ojs/pull/5436
omp: https://github.com/pkp/omp/pull/2280
ops: https://github.com/pkp/ops/pull/1246

asmecher:
I think, at a glance, it should be possible to resolve this by adding `invitation` to the list of exemptions in `lib/pkp/classes/security/authorization/RestrictedSiteAccessPolicy.php`. However, this needs testing.

taslangraham:
[...]

PR #12433, opened 2026-03-10
pkp/pkp-lib#12374 Fix issue preventing users from accepting invitations while logged out

taslangraham:
For pkp/pkp-lib#12374

### DIFF
--- a/api/v1/_i18n/I18nController.php
+++ b/api/v1/_i18n/I18nController.php
@@ -23,6 +23,7 @@ use Illuminate\Support\Facades\Route;
 use PKP\core\PKPBaseController;
 use PKP\core\PKPRequest;
 use PKP\facades\Locale;
+use PKP\security\authorization\PublicAccessPolicy;
 
 class I18nController extends PKPBaseController
 {
@@ -56,8 +57,9 @@ class I18nController extends PKPBaseController
      */
     public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
     {
-        // No authorization required for publicly accessible endpoint
-        return true;
+        $this->setEnforceRestrictedSite(false);
+        $this->addPolicy(new PublicAccessPolicy());
+        return parent::authorize($request, $args, $roleAssignments);
     }
 
     /**
--- a/api/v1/comments/UserCommentController.php
+++ b/api/v1/comments/UserCommentController.php
@@ -31,6 +31,7 @@ use PKP\API\v1\comments\resources\UserCommentReportResource;
 use PKP\API\v1\comments\resources\UserCommentResource;
 use PKP\core\PKPBaseController;
 use PKP\core\PKPString;
+use PKP\security\authorization\PublicAccessPolicy;
 use PKP\security\Role;
 use PKP\userComment\relationships\UserCommentReport;
 use PKP\userComment\UserComment;
@@ -119,8 +120,8 @@ class UserCommentController extends PKPBaseController
      */
     public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
     {
-        // No authorization required for public endpoint
-        return true;
+        $this->addPolicy(new PublicAccessPolicy());
+        return parent::authorize($request, $args, $roleAssignments);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


api/v1/_i18n/I18nController.php
```diff
@@ -49,6 +50,16 @@ public function getGroupRoutes(): void
         Route::get('ui.js', $this->getTranslations(...))->name('_i18n.getTranslations');
     }
 
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::authorize()
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+        // No authorization required for publicly accessible endpoint
```
asmecher: This will bypass e.g. `HttpsPolicy`, which should apply. Could you see if there's another way that doesn't require short-cutting the `authorize` function? I worry this will have other knock-on effects.

taslangraham: @asmecher 
Maybe something like the following would suffice.

```php
    // api/v1/_i18n/I18nController.php::authorize()
    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
    {
        $this->addPolicy(new class extends \PKP\security\authorization\AuthorizationPolicy {
            public function effect(): int {
                return self::AUTHORIZATION_PERMIT;
            }
        });
        return parent::authorize($request, $args, $roleAssignments);
    }
```

This adds an anonymous policy class which grants access to the endpoints, but still allows the authorization checks in the base `PKPBaseController` class to take effect. What do you think?

If we go with this approach then that anonymous policy could be moved to an actual policy class.

asmecher: @taslangraham, that looks promising, but before you go that route -- can you check whether a call to `markRoleAssignmentsChecked` would be enough? If I remember correctly, the auth code doesn't like it when code is run without an role-checking auth policy, because there's a risk that someone will forget to add the auth policy. This function will let the auth code know that it's not a mistake, we actually do want to run without a policy added.

taslangraham: @asmecher 

I checked, and the `markRoleAssignmentsChecked` method alone isn't enough. Based on what you described though, if we go the route I suggested above, then it likely should be updated to also call `markRoleAssignmentsChecked`.

ewhanson: Hey @asmecher and @taslangraham, I think we're at the point where we need an explicit way to allow public, unauthenticated endpoints. 

My preference for handling this case would be twofold:

1. Create an explicit PublicAccessPolicy to be used in cases like this (example below).
2. For the `I18NController` specifically, since this needs to work regardless of whether the site access is restricted to logged in users, it should have `$this->setEnforceRestrictedSite(false)` set to false.

There are a few different places with public API endpoints that should be looked at as well, though I don't think any others need to bypass the `setEnforceRestrictedSite`.

As an example, the `PublicAccessPolicy` could look something like:

```php
class PublicAccessPolicy extends AuthorizationPolicy
{
    public function applies(): bool
    {
        return true;
    }

    public function effect(): int
    {
        return AuthorizationPolicy::AUTHORIZATION_PERMIT;
    }
}
```

with the `authorize` method for the `I18NController` looking something like:

```php
public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
{
    $this->setEnforceRestrictedSite(false);
    $this->addPolicy(new PublicAccessPolicy($request));
    return parent::authorize($request, $args, $roleAssignments);
}
```

This will definitely apply to this controller, but we should check any other public API controllers as well (I know offhand that the `StatsSushiController` should be looked at, but there may be others as well. 

This approach should allow the public endpoints to pass through the authorization process without skipping any of the other global middleware.

Let me know your thoughts or if anything is missing here.

taslangraham: @ewhanson This looks great, it’s along the same lines I was suggesting earlier, and fleshes things out nicely. 

I did a quick audit and found a few more API controllers that should be reviewed, namely `PKPInterestController`, `UserCommentController`, and `PeerReviewController`. These are all publicly accessible and are currently bypassing the global middleware checks.

ewhanson: Thanks @taslangraham. When you address this controller, could you add the new PublicAccessPolicy to the other controllers you mentioned?

Issue #12354, opened 2026-02-17
[Data Citations] - validation of PIDs in data citations


bozana:
Currently, the users can enter whatever they want as persistent identifier -- there is no validation. Also, it would be good to only save the identifier itself, without resolver/URL part or similar.

PR: 
- https://github.com/pkp/pkp-lib/pull/12455

Fix1 (PID namespace used for structured citations): 
- https://github.com/pkp/pkp-lib/pull/12544
- ~~https://github.com/pkp/ojs/pull/5480~~ (only submodule update)

ajnyga:
The challenge here of course is that there are so many different types of identifiers out there. Of course we could also approach this by introducing at least checks for some types like DOI

ajnyga:
pkp-lib: https://github.com/pkp/pkp-lib/pull/12455
ojs (just tests): https://github.com/pkp/ojs/pull/5417

Should validate the following examples:

## DOI
Valid:
- `10.1234/5a6b-7c8d`
- `doi:10.1234/5a6b-7c8d` (saved as `10.1234/5a6b-7c8d`)
- `https://doi.org/10.1234/5a6b-7c8d` (saved as `10.1234/5a6b-7c8d`)
- `https://dx.doi.org/10.1234/5a6b-7c8d` (saved as `10.1234/5a6b-7c8d`)

Invalid:
- `10.123/too-short-prefix`
- `just-random-text`

## ARXIV
Valid:
- `2301.00001`
- `cs.AI/0601001`
- `arxiv:2301.00001` (saved as `2301.00001`)
- `https://arxiv.org/abs/2301.00001` (saved as `2301.00001`)
- `https://arxiv.org/pdf/2301.00001` (saved as `2301.00001`)

Invalid:
- `random-text`

## Handle
Valid:
- `11366/1249`
- `20.1000/xyz123`
- `handle:11366/1249` (saved as `11366/1249`)
- `hdl:11366/1249` (saved as `11366/1249`)
- `https://hdl.handle.net/11366/1249` (saved as `11366/1249`)

Invalid:
- `11366`
- `random-text`

## ISSN
Valid:
- `2167-8359`
- `1234-567X`
- `issn:2167-8359` (saved as `2167-8359`)
- `https://portal.issn.org/resource/ISSN/2167-8359` (saved as `2167-8359`)

Invalid:
- `21678359`
- `1234-56789`

## ISBN
Valid:
- `9780306406157`
- `978-0-306-40615-7`
[...]

PR #12455, opened 2026-03-15
pkp/pkp-lib#12354 Add validation for data citation identifiers

ajnyga:
https://github.com/pkp/pkp-lib/issues/12354


classes/pid/Url.php
```diff
@@ -0,0 +1,30 @@
...
+
+class Url extends BasePid
+{
+    /** @copydoc BasePid::regexes */
+    public const regexes = [
+        '#https?://(www\.)?[-a-zA-Z0-9@:%._\+~\#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\+.~\#?&//=]*)#'
+    ];
+
+    /** @copydoc BasePid::validationRegexes */
+    public const validationRegexes = [
```
bozana, line 27: Maybe only for this class it would be better to overwrite the method isValid and use the php `filter_var($url, FILTER_VALIDATE_URL)` ?

classes/citation/pid/ExtractPidsHelper.php
```diff
@@ -17,6 +17,11 @@
 namespace PKP\citation\pid;
 
 use PKP\citation\Citation;
+use PKP\pid\doi;
```
bozana: Doi instead of doi

classes/pid/PidResolver.php
```diff
@@ -0,0 +1,48 @@
...
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class PidResolver
+ *
+ * @ingroup pid
+ *
+ * @brief Resolves identifier type strings to their corresponding PID class.
+ */
+
+namespace PKP\pid;
```
bozana: It seems this class is currently only used in/for data citations. It also does not contain orcid and urn -- because those are not supported in data citations, I assume.
So maybe to move it to the dataCitation namespace? 

This is also assumed in Repository, with:
```
$pidClass = PidResolver::resolveByIdentifierType($identifierType);
if (!$pidClass) {
  return;
}
```

classes/dataCitation/Repository.php
```diff
@@ -51,6 +53,7 @@ public function __construct(Request $request, PKPSchemaService $schemaService)
     public function validate(?DataCitation $dataCitation, array $props): array
     {
         $schema = DataCitation::getSchemaName();
+        $errors = [];
```
bozana: I think you can leave this below, adding it shortly before it is used -- it is slightly easier to read/understand. Also, so it does not get initialized even before the validator runs.

classes/pid/BasePid.php
```diff
@@ -143,6 +146,35 @@ public static function extractFromString(?string $string): string
...
+        if (!empty($class::validationRegexes)) {
+            foreach ($class::validationRegexes as $regex) {
+                if (preg_match($regex, $value)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+ 
+        return $class::removePrefix($value) !== '';
```
bozana, line 175: empty lines in this function contain an empty space :-)

classes/pid/Purl.php
```diff
@@ -0,0 +1,26 @@
...
+ * to the current location of a resource. PURLs are validated as URLs.
+ *
+ * @see https://purl.archive.org/
+ */
+
+namespace PKP\pid;
+
+class Purl extends Url
+{
+}
```
bozana: a new line missing

classes/pid/Url.php
ajnyga: Yes why not

classes/pid/Ecli.php
```diff
@@ -0,0 +1,42 @@
...
+        '/^ECLI:[A-Z]{2}:[A-Z0-9.]+:\d{4}:[A-Z0-9.]+$/i'
+    ];
+
+    /** @copydoc BasePid::urlPrefix */
+    public const urlPrefix = 'https://eur-lex.europa.eu/legal-content/redirect/?urn=ecli:';
+
+    /** @copydoc BasePid::alternatePrefixes */
+    public const alternatePrefixes = [
+        'https://eur-lex.europa.eu/legal-content/redirect/?urn=ecli:',
+        'https://eur-lex.europa.eu/legal-content/redirect/?urn=ECLI:',
```
bozana: I think we need to remove ecli: and ECLI: from these prefixes -- they should stay in the bare identifier we save, correct? -- when removing the prefix the whole string will be removed.
The regexes above seem to handle it correctly for extraction.

classes/pid/Urn.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace PKP\pid;
+
+class Urn extends BasePid
+{
+    /** @copydoc BasePid::regexes */
+    public const regexes = [
+        // urn:nbn:de:101:1-2019072802401757702913
+        '/urn:([a-z0-9][a-z0-9-]{1,31}):((?:[-a-z0-9()+,.:=@;$_!*\'&~\/]|%[0-9a-f]{2})+)(?:(\?\+)((?:(?!\?=)(?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2}))*))?(?:(\?=)((?:(?!#).)*))?(?:(#)((?:[-a-z0-9()+,.:=@;$_!*\'&~\/\?]|%[0-9a-f]{2})*))?$/i'
```
bozana: this is a bug from earlier -- there should be no $ in the regex here -- the regex is used for extraction from the text and there could be further text after the urn -- the text does not need to end with urn. Could you please remove it?

classes/pid/Arxiv.php
```diff
@@ -1,36 +1,41 @@
...
+namespace PKP\pid;
 
 class Arxiv extends BasePid
 {
-    /** @copydoc AbstractPid::regex */
+    /** @copydoc BasePid::regexes */
     public const regexes = [
         // arxiv:2025.12345v2 https://arxiv.org/abs/2025.12345
         '/(?:arxiv:\s*|https?:\/\/arxiv.org\/(?:abs|pdf)\/)(?:\d+.\d+|[a-z-]+.\d+)/i'
     ];
```
bozana, line 25: could you please improve this regex as well, to use:
`'/(?:arxiv:\s*|https?:\/\/arxiv\.org\/(?:abs|pdf)\/)(?:\d+\.\d+|[a-zA-Z.-]+\/\d+)/i' `                                                                              
                                                                                                                                                                         
I.e. to escape the dot (.) and consider A-Z, . and - correctly:
  - `arxiv.org` → `arxiv\.org`                                                                                                                                               
  - `\d+.\d+` → `\d+\.\d+`                                                                                                                                                   
  - `[a-z-]+.\d+` → `[a-zA-Z.-]+\/\d+`

schemas/dataCitation.json
bozana, line 63: shell we maybe consider required_with:identifier here?

bozana (commented):
Hi @ajnyga, great work!!! Thanks a lot!
I left just a few comments.
Also, I would maybe double check with @touhidurabir if overriding the fill() is the best solution here. I think it is good, but just in case...

classes/dataCitation/Repository.php
ajnyga: ok, saw that it was here in the DOI class so moved it because of that

touhidurabir:
> if overriding the fill() is the best solution here.

@bozana @ajnyga this is not bad and we have already an existing use case of `Model::fill` overwrite in the `PKP\editorialTask\EditorialTask` . 

But I have few concerns here for the proposed implementation of `fill` overwrite . 

if I understand correctly, the purpose of this overwrite of `fill` is to do some sort of sanitization/modification of the identifier by some sort of standards . I am not sure of all the possible use case but `fill`  will not help for the case of prop update/saviing like 
```php
$model = new Model(); 
$model->identifier = '...'; 
$model->save(); // or $model->update()
// NOTE that there are differences between new Model($data) and above approach
```

so basically it will be completely bypasses at the case of property assignment but works for  mass assignment usage like `create`, `fill` , `update` etc . 

I suggest using the `saving` eloquent model event if that works(need testing) . like following code (not tested but something similar should work)
```php
protected static function boot()
{
    parent::boot();

    static::saving(function (self $model) {
        // Only sanitize/modiyf if the identifier is actually being changed
        // AND we have a valid identifierType (new or existing record).
        if ($model->isDirty('identifier') && !empty($model->identifierType)) { // Here dirty to handle case of updating existing record
            $pidClass = PidResolver::resolveByIdentifierType($model->identifierType);

            if ($pidClass) {
                $identifier = $pidClass::extractFromString($model->identifier);

                if ($identifier !== '') {
                    $model->identifier = $identifier;
                } else {
                    $model->identifier = $pidClass::removePrefix($model->identifier);
                }
            }
        }
    });
}
```

touhidurabir:
@bozana @ajnyga just a note regarding the bulk inset via `insert`, `upsert`, `insertOrIgnore` etc will not be solved via either `fill` overwrite or `saving` eloquent model event as that use query builder, not eloquent methods . either need to use `createMany` or create via loop , or need custom sanitization/modification before bulk insert .

classes/pid/PidResolver.php
ajnyga: I was maybe thinking that this could be reused somewhere, what if we just also orcid and urn? If the class in under data citations it is maybe likely to be ignored. But for me moving it to data citations is fine as well.

classes/pid/Url.php
ajnyga: FILTER_VALIDATE_URL seems to allow things like mailto: and ftp:. I do not think it is a good idea to use that here since we probably just want http(s) links?

ajnyga:
> I suggest using the `saving` eloquent model event if that works(need testing) . like following code (not tested but something similar should work)

It seems that with the given code a test case like this does not work:
1. Add new data citation, identifier type = URI and identifier a valid DOI with https 
2. Works as expected and the whole DOI url is saved
3. Edit the data citation and change type to DOI
4. The result is that although the new type is saved, the prefix of the identifier is not removed. It works if I edit again.

This is the code that I rewrote to deal with that problem, it takes into account also the situation where just the type is changing.

```
        static::saving(function (self $model) {
            $dirty = $model->getDirty();
            $identifierType = $dirty['identifier_type'] ?? $model->identifierType;

            if ((isset($dirty['identifier']) || isset($dirty['identifier_type'])) && !empty($model->identifier) && !empty($identifierType)) {
                $pidClass = PidResolver::resolveByIdentifierType($identifierType);

                if ($pidClass) {
                    $identifier = $pidClass::extractFromString($model->identifier);

                    if ($identifier !== '') {
                        $model->identifier = $identifier;
                    } else {
                        $model->identifier = $pidClass::removePrefix($model->identifier);
                    }
                }
            }
        });
```

touhidurabir:
> This is the code that I rewrote to deal with that problem, it takes into account also the situation where just the type is changing.

@ajnyga no wonder some modification are needed there as I wrote that code as an example on fly here without any testing and thats why mention that `not tested but something similar should work` . If it works for all the cases with modification, thats great .

ajnyga:
yes, thanks for pointing the saving model. I think I tried that as well at some point but was not familiar with isDirty or getDirty and probably had this same problem and ended up with fill while not knowing how to solve it in saving. But like you said that was problematic as well.

classes/pid/Url.php
bozana: Then maybe a stricter regex? 
e.g. 
`/^https?:\/\/(www\.)?[-a-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,}([-a-z0-9@:%_\+.~#?&\/=]*)$/i`

classes/pid/PidResolver.php
bozana: lets keep it now as it is... it can then be moved later, to be used generally, when needed...

classes/pid/BasePid.php
bozana: still there, in two empty lines above, in this function :-)

classes/pid/Arxiv.php
bozana: ah, now I have realized that we use 'i' at the end, so that A-Z is not needed, here and below.

bozana:
Hi @ajnyga, it looks great! And together with @touhidurabir, you find a good solution for the validation! Thanks a lot!
There is maybe one thing to do -- to maybe better specify the URL validation regex. There are also two other, very minor comments, if you would like to address them too. Else, it is then ready to be merged.
Thanks a lot!

classes/pid/BasePid.php
ajnyga: how do you catch these? do you have a setting in your editor that marks these somehow?

classes/pid/Url.php
ajnyga: should we then use {2,} also in the other regex above the allow longer domains?

bozana: yes, thanks!

classes/pid/Arxiv.php
bozana: thanks a lot!

classes/pid/BasePid.php
bozana: this was accidentally when I was looking in the file and new code... :sweat_smile:

bozana:
Hi @ajnyga, this seems to be ready for merge, once the tests pass, correct?

Issue #12402, opened 2026-03-03
Restrict starting a task when there are no participants or assignee


blesildaramirez:
**Description:**
Currently, a task can be started even if it has no participants or no assignee. This occurs for tasks that are automatically created.

**Solution:**
-  Starting a task without participants is not allowed.
- Starting a task without an assignee is not allowed.
- A validation message is displayed to the user.
- A task can still be closed even if there are no participants  or no assignee.

**What application are you using?**
  OJS, OMP or OPS version 3.6 (main)

Hafsa-Naeem:
@Vitaliy-1  Ready for review
pkp-lib: https://github.com/pkp/pkp-lib/pull/12456
OJS (main): https://github.com/pkp/ojs/pull/5418

Vitaliy-1:
Thanks, @Hafsa-Naeem! All looks good but could you also update API documentation?

Hafsa-Naeem:
@Vitaliy-1  Ready for review
OJS (main): https://github.com/pkp/ojs/pull/5418

pkp-lib: https://github.com/pkp/pkp-lib/pull/12456 (merged)

Hafsa-Naeem:
@Vitaliy-1 I've updated the API documentation as well. Could you please review this so I can close out the issue? Thanks.

PR #12456, opened 2026-03-16
pkp/pkp-lib#12402 Restrict starting a task when there are no participants or assignee

Hafsa-Naeem:
for #12402


Vitaliy-1 (approved):
Thanks, @Hafsa-Naeem!

Can you also add these responses to the swagger source?

Issue #11803, opened 2025-09-09
User invitation email modifications are lost


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When sending a role invitation email, it's possible to customize the message text. However, those customizations are not included in the delivered email; it is sent with a stock invitation.

### Steps to Reproduce

With an installation that's capable of sending and receiving emails...

1. As Journal Manager, go to Users & Roles.
2. Use the "Invite to a role" tool to create an invitation to a user account. (Details not important.)
3. Customize the email template with e.g. a new sentence.
4. Send the invitation.
5. Check the received email.

https://github.com/user-attachments/assets/5f7b9546-2e71-42fa-af8b-6edaea10df0f

### Expected Result

The email should be received with the customizations that were made during composition.

### Actual Result

The email will be received without any customizations.

### Application Version

OJS 3.5.0-x

---

**PRs**

**Branch `stable-3_5_0`**
pkp-lib: #11921
ui-library: pkp/ui-library#721

defstat:
@asmecher could you please review the PR found in the description comment of the issue?

asmecher:
(Just for the sake of recording -- we discussed this today during my open office and decided to try a different approach to the styling.)

defstat:
@asmecher @jardakotesovec if you can check those PRs - I tried the least possible change as this is for `stable-3_5_0` along with `main`. Also I, more or less, used the same "structure" as Decision mailables work.

I am not happy for the server-side handling of the `<span>` value for the TinyMCE Editor that my code introduces. This should happen client-side by something like this in the `Composer.vue`.

```
localizedVariables() {
			return (this.variables && this.variables[this.locale]) ? this.variables[this.locale] : [];
		},
		compiledVariables() {
			const variables = [...this.localizedVariables].map(v => {
				const hasValue = v.value && String(v.value).trim() !== '';
				const wrapped = hasValue
					? `<span class="mceNonEditable" data-symbolic="${v.key}" contenteditable="false">${this.escapeHtml(String(v.value))}</span>`
					: `<span class="pkpTag mceNonEditable" data-symbolic="${v.key}" contenteditable="false">${v.key}</span>`;
				return { ...v, value: wrapped };
			});

			const recipientName = variables.find(v => v.key === 'recipientName');
			if (recipientName) {
				if (this.separateEmails) {
					recipientName.value = '{$recipientName}';
				} else if (this.canChangeRecipients) {
[...]

PR #12462, opened 2026-03-17
pkp/pkp-lib#11803 [main] User invitation email modifications are lost

blesildaramirez:


classes/mail/traits/AddsStyleToSymfonyMessage.php
```diff
@@ -0,0 +1,54 @@
...
+    protected function registerMailCss(): void
+    {
+        $path = $this->mailCssPath();
+        if (!$path || !is_file($path)) {
+            return;
+        }
+
+        static $cssCache = [];
+        $css = $cssCache[$path] ??= (string) @file_get_contents($path);
+
```
asmecher: I don't recommend using a static cache here. As we move towards using long-running threads/daemons for jobs etc., we may find these living longer than originally expected, and there is no mechanism here to expire the cache e.g. based on the file date. Beyond that, I don't think a cache is justified; `file_get_contents` is extremely quick, and these won't be large files.

Issue #11800, opened 2025-09-09
Allow manager to edit user setting Apper on the masthead


bozana:
Because there are cases, when a user does not want to be displayed on the masthead, we should allow the journal manager to edit/change this setting for that user.
However, if a user opts out to be displayed, we should display the following message on the masthead: "One or more editors have opted not to be listed on the masthead of this journal."

jardakotesovec:
We discussed some options how to incorporate it to current workflow with @Devika008 .

As reference - this is our current screen for editing user:

<img width="1496" height="905" alt="Image" src="https://github.com/user-attachments/assets/394b8ca0-fb98-4a94-8097-cd8c38b855f7" />


### Options
Question is how to make the Journal Masthead option editable:

1) We can follow same workflow as with Delete Role, which presents confirmation dialog and after confirming it removes it immediately and it sends information email to the user that the role was removed. In this case we would have dropdown (select element) and if it gets change it would present dialog and send email.  Therefore this change would not require user confirmation
2) Alternative is to create invitation. The workflow would be same when for example adding new role. Therefore user would get email about this setting being changed for his role and he/she would need to confirm it.

asmecher:
Deferring to 3.5.0-3 -- but this is probably necessary to fix before the LTS designation goes on, so please don't forget.

bozana:
Hi @ewhanson and @taslangraham, would one of you be able to review my backend changes for this issue, s. PR for 3.5 and main above? This should go to the next 3.5.0.4 release.
Thanks a lot!

jardakotesovec:
I added changes in ui-library and pkp-lib (combined PR of my and Bozana changes). I will do additional testing on monday as I also changed the logic how the `currentUserGroups` are handled - so they can be properly reloaded after the change from API, instead of just changed optimistically client side.

taslangraham:
~@bozana I see that @jardakotesovec has opened a PR built on top of your changes, I'll give the review there.~

@bozana I've left a few comments and suggestions on your [3.5 pkp-lib PR](https://github.com/pkp/pkp-lib/pull/12471). Those comments will also need to be addressed in the [PR for main](https://github.com/pkp/pkp-lib/pull/12472).

@jardakotesovec Your UI-library PR looks good, but I'll hold on off approving until you're through with the testing you mentioned above. I've left a comment on the pkp-lib PR. Happy to give another review once everything is tested and finalised.

bozana:
[...]

PR #12471, opened 2026-03-19
pkp/pkp-lib#11800 Allow manager to edit user setting Appear on the ma…

bozana:
…sthead

s. https://github.com/pkp/pkp-lib/issues/11800

### DIFF
--- a/api/v1/users/PKPUserController.php
+++ b/api/v1/users/PKPUserController.php
@@ -29,14 +29,17 @@ use Illuminate\Support\LazyCollection;
 use PKP\context\Context;
 use PKP\core\PKPBaseController;
 use PKP\core\PKPRequest;
+use PKP\core\PKPString;
 use PKP\facades\Locale;
 use PKP\mail\mailables\UserRoleEndNotify;
+use PKP\mail\mailables\UserRoleMastheadUpdateNotify;
 use PKP\plugins\Hook;
 use PKP\security\authorization\ContextAccessPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
-use PKP\userGroup\UserGroup;
 use PKP\security\Validation;
+use PKP\userGroup\relationships\UserUserGroup;
+use PKP\userGroup\UserGroup;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class PKPUserController extends PKPBaseController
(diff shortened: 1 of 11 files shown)


api/v1/users/PKPUserController.php
```diff
@@ -353,6 +358,54 @@ public function endRole(Request $request): JsonResponse
...
+        $userId = $request->route('userId');
+        $user = Repo::user()->get($userId, true);
+        if (!$user) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound')
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        // Ensure user group exists and belongs to the current context
+        $context = $request->attributes->get('context'); /** @var Context $context */
```
taslangraham, line 378: Any reason(or benefits) to getting the Context this way as opposed a more common `$this->getRequest()->getContext()` pattern?

api/v1/users/PKPUserController.php
```diff
@@ -353,6 +358,54 @@ public function endRole(Request $request): JsonResponse
...
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $masthead = $request->input('masthead');
+        if (isset($masthead) && !is_bool($masthead)) {
+            $masthead = PKPString::strictConvertToBoolean($masthead);
+        }
+        if (!is_bool($masthead)) {
+            return response()->json([
+                'error' => __('api.400.invalidRequest')
```
taslangraham: Could you double check that this locale exists. I did a search but nothing turned up. Did you intend on adding it?

api/v1/users/PKPUserController.php
```diff
@@ -353,6 +358,54 @@ public function endRole(Request $request): JsonResponse
...
+            ], Response::HTTP_BAD_REQUEST);
+        }
+
+        Repo::userGroup()->updateUserUserGroupMasthead($userId, $userGroupId, $masthead, activeOnly: false);
```
taslangraham: I’d suggest sticking to either positional arguments or named arguments for consistency, rather than mixing both.

api/v1/users/PKPUserController.php
bozana: I see we now use it this way everywhere in this class

bozana: Added/changed with this commit: https://github.com/pkp/pkp-lib/commit/71e79e31e3d5c827e4bfa2443bbe81e5ec4c1dba#diff-bab50ec5ded0dcc8adbaa27c6ac3428934b6b254145f1340a34be5db63e36eeb

api/v1/users/PKPUserController.php
bozana: forget, and I renamed it to invalidMastheadParameter

api/v1/users/PKPUserController.php
bozana: I would then leave it as it is...

bozana: It seems to be the Laravel HTTP request pipeline, which would already have the context, so I am wondering why we do not use it everywhere... Maybe we should...

taslangraham: I see it is used in only a handful of other places. But I don't have any strong opinions on whether or not we should adapt it more.

PR #12473, opened 2026-03-19
I11800 3 5

jardakotesovec:


api/v1/users/PKPUserController.php
```diff
@@ -353,6 +360,87 @@ public function endRole(Request $request): JsonResponse
...
+
+        // Update via repository to ensure cache invalidation is handled
+        Repo::userGroup()->setUserUserGroupMasthead($userUserGroup, $masthead);
+
+        // Send email notification
+        $mailable = new UserRoleMastheadUpdateNotify($context, $userUserGroup);
+        $emailTemplate = Repo::emailTemplate()->getByKey($context->getId(), $mailable::getEmailTemplateKey());
+
+        if (!$emailTemplate) {
+            $message = 'Email template ' . $mailable::getEmailTemplateKey() . ' not found. The migration script I11800_AddUserRoleMastheadUpdateEmail needs to be run.';
```
taslangraham, line 427: I understand the idea behind this, but I don't feel like we should be sharing internal details like migration names to the user. 

Secondly, whichever message is used here should be added to appropriate locale (`.po`) file.

api/v1/users/PKPUserController.php
```diff
@@ -353,6 +360,87 @@ public function endRole(Request $request): JsonResponse
...
+            error_log($message);
```
taslangraham: Should be removed

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public static function getDataDescriptions(): array
+    {
+        $variables = parent::getDataDescriptions();
+        $variables[static::$roleNameAndDates] = __('emailTemplate.variable.userRoleMastheadUpdateNotify.roleNameAndDates');
+        $variables[static::$appearOnMasthead] = __('emailTemplate.variable.userRoleMastheadUpdateNotify.appearOnMasthead');
+        return $variables;
+    }
+
+    public function setData(?string $locale = null): void
```
taslangraham, line 83: Missing documentation for the method

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
```diff
@@ -0,0 +1,105 @@
...
+        $roleName = $userGroup->getLocalizedData('name', $locale);
+        $startDate = $this->userUserGroup->dateStart?->format('Y-m-d');
+        $endDate = $this->userUserGroup->dateEnd?->format('Y-m-d');
+        $role = ($startDate || $endDate)
+            ? __('emailTemplate.variable.userRoleMastheadUpdateNotify.roleNameAndDates.value', [
+                'roleName' => $roleName,
+                'startDate' => $startDate ?? '',
+                'endDate' => $endDate ?? '',
+            ])
+            : $roleName;
```
taslangraham: Curious to know under which circumstances start date would be missing.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $userGroup = Repo::userGroup()->get($this->userUserGroup->userGroupId);
+        $roleName = $userGroup->getLocalizedData('name', $locale);
+        $startDate = $this->userUserGroup->dateStart?->format('Y-m-d');
+        $endDate = $this->userUserGroup->dateEnd?->format('Y-m-d');
+        $role = ($startDate || $endDate)
+            ? __('emailTemplate.variable.userRoleMastheadUpdateNotify.roleNameAndDates.value', [
+                'roleName' => $roleName,
+                'startDate' => $startDate ?? '',
+                'endDate' => $endDate ?? '',
```
taslangraham: A possible improvement: when end date is missing we display `"present"` instead of an empty string. So the text when the end date is missing would look something like `"Author March 31, 2026 - present"`.  If start date can truly be missing then a similar placeholder text should be considered. What do you think? 

If my suggestion at https://github.com/pkp/pkp-lib/pull/12473/changes#r3017302386 is implemented then this comment can be ignored.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
```diff
@@ -0,0 +1,105 @@
...
+        parent::setData($locale);
+        if (is_null($locale)) {
+            $locale = $this->getLocale() ?? Locale::getLocale();
+        }
+
+        $userGroup = Repo::userGroup()->get($this->userUserGroup->userGroupId);
+        $roleName = $userGroup->getLocalizedData('name', $locale);
+        $startDate = $this->userUserGroup->dateStart?->format('Y-m-d');
+        $endDate = $this->userUserGroup->dateEnd?->format('Y-m-d');
+        $role = ($startDate || $endDate)
```
taslangraham: Given that the locale `emailTemplate.variable.userRoleMastheadUpdateNotify.roleNameAndDates.value` expects both `startDate` and `endDate` to be present and will always contain a dash (`-`), shouldn't this be an `AND` condition (`$role = ($startDate && $endDate)`) instead of an `OR` condition?

api/v1/users/PKPUserController.php
jardakotesovec: @taslangraham  We discussed options on dev leads meeting.. and there is no great solution for migrations being added to the stable release. Unfortunately. This is message that Journal Managers will see and can notify admin to do the update. 
This is not intended to be localised..  without adding this they would see php cryptic error instead. So we at least made the error actionable.
We also intend to add note to the release notes.. but thats not commonly read.. 

Other options that we considered
- apply migration when the template is not there..   but thats also not great pattern to encourage mixing runtime code and migration code.. 
- make 3.5.1 release, thats just way too much work to justify it

So from the options this was @asmecher preference.

Eventually solution is probably just have standard tooling to track in database which migrations are applied. Which makes it quite easy for apply missing migrations. I would love to see that being worked on.. as it would also help with the automatic deployments to the VM..

bozana: This code will/should be removed in 3.5.1. It is just for now, so that those users that do not run the DB upgrade when upgrading from 3.5.0.x to 3.5.0.4 see that they should do it. With 3.5.1 we assume/require the users to do the DB upgrade, so this will not be needed any more.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
bozana: The existing users -- the existing journals that upgrade to 3.5 -- will have no start date.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
bozana: I actually left the text in the same way as those information are displayed on the masthead pages -- with dash and empty start or end date, if they are missing. If both are missing then nothing is displayed. Just that on mastehad pages only year is displayed.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
bozana: On the public masthead page it is displayed that way, for example:
```
Section editor
– 2026
Ramiro Vaca
Universidad Nacional Autónoma de México

```
or:

```
Section editor
2025 – 
Daniel Barnes
University of Melbourne 
```

That is why I left it so in the email as well.

api/v1/users/PKPUserController.php
bozana: I left this so that the system admins can see it -- those can only run the migration script -- and they will probably notice it if it is in the server's php error log.
If journal manager sees the message, it is not sure that they will communicate it to system admins -- the communication between editorial team and system admins is not always straight. I am not sure if journal manager will even understand the message correclty.
Should we then leave it?

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
bozana: Or do you think we should rather have 'present' (no end date) and 'unknown' (no start date)?

bozana: I can ask Devika -- I asked about the current solution but still have not answer, so I will ask about the solution suggested here...

bozana: Also, I now realized that I should then rather use either date_format_short or date_format_long from the config.inc.php, to format the start and end date...

taslangraham: Yes, let's see what @Devika008 thinks about this.

api/v1/users/PKPUserController.php
taslangraham: Okay, I'm fine with leaving it it based on what you've explained.

classes/mail/mailables/UserRoleMastheadUpdateNotify.php
bozana: This is the summary from Devika:

> Both dates present:
> Section Editor (April 1, 2025 – April 1, 2026)
> Only start date:
> Section Editor (since April 1, 2025)
> Only end date:
> Section Editor (until April 1, 2026)
> Neither date present:
> Section Editor

I will change accordingly...

api/v1/users/PKPUserController.php
bozana: Oh, my bad. It seems this Exception below will be logged, because it is not catched. So I can remove error_log()...
Sorry!

bozana:
Hi @jardakotesovec and @taslangraham,
I updated my branch/PR (https://github.com/pkp/pkp-lib/pull/12471), considering Taslan's comments. 
Also:
- date format is taken from the config.inc.php date_format_long -- as Devika suggested
- the email is now sent in the journal primary locale instead of users UI locale.

OK so?
When Jarda merges it, then maybe Taslan you can take a look again?
Thanks a lot!

Issue #7531, opened 2021-12-06
Checking the version of a plugin before loading it


marcbria:
This is intended to summarise the conversation we have had about the problem of "OJS breaking after an update due to incompatible plugins".

Currently an updated OJS does not check if plugins are compatible and this can result in errors that can lead to an inoperative platform. Same will happen with OMP and OPS.

We have discussed three approaches:

- **(a) WP style:** for this it is not clear if we will need to enclose the plugin code in a try/catch or resolve them by test (activate a plugin, check the response code, and then a separate request to disable the plugin if the response code is not 200)... the point is that WP loads the plugin, checks for a fatal exception/error, and if one is found automatically disables the plugin and displays a warning.
 
- **(b) Based on a centralized file:** The plugin gallery is based on a [centralized xml file](http://pkp.sfu.ca/ojs/xml/plugins.xml) which, among other things, includes compatibility information. This information can be used during the upgrade to decide whether plugins should be deactivated or for OJS to refuse to activate/install an incompatible plugin. 

- **(c) Based on plugin's own metadata** (version.xml): Each plugin must include compatibility information. As the file will not be able to describe the compatibility of future versions, the compatibility version should be based on the semantic versioning contract of the product. For example, a plugin should state 3.3.*.* to make the plugin load up to version 3.4.0-1.

Any, even a combination of the different techniques could be used, although (c) looks like the easiest and fastest to be adopted. 

Ideas for future improvements
- plugins automatically updated during an upgrade when a compatible version exists in the plugin gallery.

References:
- WP's check for requirements:  https://wptavern.com/how-to-prevent-wordpress-plugins-from-activating-on-sites-with-incompatible-hosting-environments
- Drupal's compatibility metadata: https://www.drupal.org/docs/creating-custom-modules/let-drupal-know-about-your-module-with-an-infoyml-file

marcbria:
@asmecher, @ctgraham, @NateWr, @henriqueramos, @jonasraoni and @Vitaliy-1 we can keep talking about this here.

jonasraoni:
@ewhanson I remember you were working on this during the sprint, did you commit the code somewhere? If yes, then we can close this and some other related things :)

JhonathanLepidus:
@thiagolepidus and I just implemented a PR which tackles the version.xml changes request in this feature: [PR 12480](https://github.com/pkp/pkp-lib/pull/12480)

[...]

PR #12480, opened 2026-03-19
Adds checking of plugin compatibility

JhonathanLepidus:
This pull request adds a plugin compatibility check before installation or update.

Thus, each plugin **must specify** in its version.xml file which applications and their respective versions it is compatible with (from 3.6 and beyond).

This is done through a new node called “compatibility.” It works similarly to the method used in the plugin gallery to indicate plugin compatibility.

Example of version.xml of a plugin compatible with OJS and OPS versions 3.6.0-x:

```xml
<version>
	<application>pluginTemplate</application>
	<type>plugins.generic</type>
	<release>1.0.0.0</release>
	<date>2023-05-15</date>
 	<class>PluginTemplatePlugin</class>
 	<compatibility application="ojs2">
		<release>~3.6.0.0</release>
	</compatibility>
 	<compatibility application="ops">
		<release>~3.6.0.0</release>
	</compatibility>
</version>
```

This pull request addresses a feature requirement described in the issue #7531 and was implemented in the context of the 2026's Plugin Sprint :)

### DIFF
--- a/classes/site/VersionCheck.php
+++ b/classes/site/VersionCheck.php
@@ -162,11 +162,9 @@ class VersionCheck
     /**
      * Checks whether the given plugin version is compatible with the current application version.
      *
-     * @param array $versionInfo
-     *
      * @throws Exception if the plugin version is not compatible
      */
-    public static function checkPluginVersionCompatibility($versionInfo): void
+    public static function checkPluginVersionCompatibility(array $versionInfo): void
     {
         $application = Application::get();
         $compatibility = $versionInfo['compatibility'];
@@ -181,17 +179,14 @@ class VersionCheck
 
         $compatibleVersions = $compatibility[$application->getName()];
         $applicationVersion = $application->getCurrentVersion();
-        $hasCompatibleVersion = false;
         foreach ($compatibleVersions as $compatibleVersion) {
             if ($applicationVersion->isCompatible($compatibleVersion)) {
-                $hasCompatibleVersion = true;
-                break;
+                return; // A compatible version was found
             }
         }
 
-        if (!$hasCompatibleVersion) {
-            throw new Exception(__('manager.plugins.incompatiblePlugin.version'));
-        }
+        // There was no compatible version found.
+        throw new Exception(__('manager.plugins.incompatiblePlugin.version'));
     }
 
     /**


classes/site/VersionCheck.php
```diff
@@ -144,9 +154,46 @@ public static function getValidPluginVersionInfo($versionFile): Version
...
     }
 
+    /**
+     * Checks whether the given plugin version is compatible with the current application version.
+     *
+     * @param array $versionInfo
+     *
+     * @throws Exception if the plugin version is not compatible
+     */
+    public static function checkPluginVersionCompatibility($versionInfo): void
```
asmecher: This is just stylistic, but unless the self-doc adds something, it's best to just use a typehint.
```suggestion
    /**
     * Checks whether the given plugin version is compatible with the current application version.
     * @throws Exception if the plugin version is not compatible
     */
    public static function checkPluginVersionCompatibility(array $versionInfo): void
```

classes/site/VersionCheck.php
```diff
@@ -144,9 +154,46 @@ public static function getValidPluginVersionInfo($versionFile): Version
...
+
+        if (empty($compatibility)) {
+            throw new Exception(__('manager.plugins.versionFileInvalid'));
+        }
+
+        if (!isset($compatibility[$application->getName()])) {
+            throw new Exception(__('manager.plugins.incompatiblePlugin.application'));
+        }
+
+        $compatibleVersions = $compatibility[$application->getName()];
```
asmecher, line 180: I think you'll need to prevent a warning here when there are none for this application:
```suggestion
        $compatibleVersions = $compatibility[$application->getName()] ?? [];
```

classes/site/VersionCheck.php
```diff
@@ -144,9 +154,46 @@ public static function getValidPluginVersionInfo($versionFile): Version
...
+        foreach ($compatibleVersions as $compatibleVersion) {
+            if ($applicationVersion->isCompatible($compatibleVersion)) {
+                $hasCompatibleVersion = true;
+                break;
+            }
+        }
+
+        if (!$hasCompatibleVersion) {
+            throw new Exception(__('manager.plugins.incompatiblePlugin.version'));
+        }
```
asmecher: Just stylistic, but you can use an early return:
```suggestion
        foreach ($compatibleVersions as $compatibleVersion) {
            if ($applicationVersion->isCompatible($compatibleVersion)) {
                return; // A compatible version was found
            }
        }

        // There was no compatible version found.
        throw new Exception(__('manager.plugins.incompatiblePlugin.version'));
```

classes/site/VersionCheck.php
asmecher: (Never mind, I see this is checked above)

PR #12482, opened 2026-03-20
fix: use scalar correlated subqueries for name ORDER BY to fix MySQL error

ambs:
## Summary

- MySQL does not allow references to outer query tables (`u`) in `JOIN ON` clauses
   inside subqueries, causing a fatal `Unknown column 'u.user_id' in 'on clause'`
   error on the Editorial Masthead page.
- Replaced the `orderBy(function...)` block that used `fromSub` + `leftJoin` with
   `orderByRaw` using scalar correlated subqueries, moving the `u.user_id` reference
   from an `ON` clause into `WHERE` clauses — which MySQL allows.
- Semantically equivalent: produces the same `CONCAT(COALESCE(...), COALESCE(...))`
   ordering and handles multiple locales correctly.

## Reproduction

Load `/about/editorialMasthead` on a MySQL instance. Fatal PDO exception:
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'u.user_id' in 'on clause'
Stack trace: `Repository::getMastheadUserIdsByRoleIds` → `Collector::buildOrderBy`

## Notes:

- Discussion: https://forum.pkp.sfu.ca/t/bug-editorialmasterhead-500-internal-server-error/97909/3
- PR prepared with AI, as my PHP and SQL are currently too rusty.

### DIFF
--- a/classes/user/Collector.php
+++ b/classes/user/Collector.php
@@ -806,44 +806,29 @@ class Collector implements CollectorInterface
             );
 
             $sortedSettings = array_values(
-                $this->orderBy === self::ORDERBY_GIVENNAME 
-                    ? $nameSettings 
+                $this->orderBy === self::ORDERBY_GIVENNAME
+                    ? $nameSettings
                     : array_reverse($nameSettings)
             );
 
-            $query->orderBy(
-                function (Builder $query) use ($sortedSettings, $locales): void {
-                    $aliasesBySetting = [];
+            $coalesceExpressions = [];
+            foreach ($sortedSettings as $i => $setting) {
+                $parts = [];
+                foreach ($locales as $j => $locale) {
+                    $alias = "us_{$i}_{$j}";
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @ambs! Before I review this, could you try to use the query builder rather than string concatenation to build this part of the query? Also, could you confirm whether this is tested? We don't have a firm policy on AI yet, but are finding that there needs to be a human in the loop to vet and test AI work before we receive it. Otherwise it can quickly overload us with additional QA.

ambs:
Hi @asmecher, here is the best code I've achieved that works properly. 
You can see it working here: https://linguamatica.com/index.php/linguamatica/pt/about/editorialMasthead

ambs:
@asmecher can we include this fix in the next release?
Anything else I can do to help merging it?

bozana:
Thanks a lot @ambs! Everything merged and it will be in the next 3.5.0.4 release.

Issue #12446, opened 2026-03-12
Improve ROR dataset update: check schema version, use JSON, improve error reporting, remove temporary table, schedule monthly


bozana:
From the suggestions from the ROR team we should implement the following in this issue:
- [x] use JSON instead of CSV -- JSON is ROR's canonical data format, used internally for all operations. The CSV, by contrast, is a derived convenience format, intended for users doing manual review of the data or who are otherwise unable to work with the JSON. ROR's data fields are also inherently hierarchical and the CSV flattens these structures into a best approximation.
- [x] handle Zenodo API failures gracefully -- the Zenodo API has had in the past undocumented instability and behavior changes, so this dependency needs safeguarding.
- [x] read admin.schema_version from JSON records, and use that schema version to determine compatibility and to find the right JSON file, instead of relying on file names.
- [x] ensure that ROR data update script never breaks OJS functionality, in this case installation.

Additionally:
- [x] Improve error reporting with specific log messages for each failure point 
- [x] Remove the temporary database table — the update now upserts directly into rors and replaces ror_settings per batch, reducing DB overhead and complexity -- and make ror_settings updates atomic using a DB transaction (DELETE + INSERT) per batch, preventing partial/inconsistent state
- [x] Changed schedule from twice-monthly to monthly

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/12522
- ~~https://github.com/pkp/ojs/pull/5467~~ [TEST ONLY]

main:
- https://github.com/pkp/pkp-lib/pull/12526

halaxa/json-machine PR, that was not so performant:
- ~~https://github.com/pkp/pkp-lib/pull/12491~~ (halaxa/json-machine)
- ~~https://github.com/pkp/ojs/pull/5440~~ (only submodule udpate)
- ~~https://github.com/pkp/omp/pull/2285~~ (only submodule udpate)
- ~~https://github.com/pkp/ops/pull/1249~~ (only submodule udpate)

old not finished PRs, used to compare the different solutions:
- ~~https://github.com/pkp/pkp-lib/pull/12459 (json_decode full json file)~~
- ~~https://github.com/pkp/pkp-lib/pull/12490 (symfony/json-streamer)~~

bozana:
Hi @touhidurabir,
Could you please review the PR above. It brings the following changes:

classes/task/UpdateRorRegistryDataset.php — core rewrite:                                                                                                                                                                                                   
[...]

PR #12491, opened 2026-03-23
12446 3 5 0 halaxa json machine

bozana:
s. https://github.com/pkp/pkp-lib/issues/12446

test halaxa/json-machine:
JSON streaming: 59s 
DB operations: 25s

### DIFF
--- a/classes/task/UpdateRorRegistryDataset.php
+++ b/classes/task/UpdateRorRegistryDataset.php
@@ -11,55 +11,19 @@
  *
  * @ingroup tasks
  *
- * @brief Class responsible for bi-weekly update of the ROR Registry tables in the database.
+ * @brief Class responsible for monthly update of the ROR Registry tables in the database.
  */
 
 /**
- * Mapping Ror data dump and OJS
+ * Mapping Ror JSON schema v2 data dump and OJS
  *
- *  0 id                      rors.ror
- * 29 names.types.label       ror_settings['locale' => 'en', 'setting_name' => 'name',           'setting_value' => 'name']
- * 30 names.types.ror_display
- * 31 ror_display_lang        ror_settings['locale' =>   '', 'setting_name' => 'default_locale', 'setting_value' => 'en']
- * 33 status                  ror_settings['locale' =>   '', 'setting_name' => 'is_active',      'setting_value' => '1']
- *
- * | index | ror registry                                        | ojs          |
- * |-------|-----------------------------------------------------|--------------|
- * | 0     | id                                                  | rors.ror     |
- * | 1     | admin.created.date                                  |              |
- * | 2     | admin.created.schema_version                        |              |
- * | 3     | admin.last_modified.date                            |              |
- * | 4     | admin.last_modified.schema_version                  |              |
- * | 5     | domains                                             |              |
- * | 6     | established                                         |              |
- * | 7     | external_ids.type.fundref.all                       |              |
- * | 8     | external_ids.type.fundref.preferred                 |              |
- * | 9     | external_ids.type.grid.all                          |              |
- * | 10    | external_ids.type.grid.preferred                    |              |
- * | 11    | external_ids.type.isni.all                          |              |
- * | 12    | external_ids.type.isni.preferred                    |              |
- * | 13    | external_ids.type.wikidata.all                      |              |
- * | 14    | external_ids.type.wikidata.preferred                |              |
- * | 15    | links.type.website                                  |              |
- * | 16    | links.type.wikipedia                                |              |
- * | 17    | locations.geonames_id                               |              |
- * | 18    | locations.geonames_details.continent_code           |              |
- * | 19    | locations.geonames_details.continent_name           |              |
- * | 20    | locations.geonames_details.country_code             |              |
- * | 21    | locations.geonames_details.country_name             |              |
- * | 22    | locations.geonames_details.country_subdivision_code |              |
- * | 23    | locations.geonames_details.country_subdivision_name |              |
- * | 24    | locations.geonames_details.lat                      |              |
- * | 25    | locations.geonames_details.lng                      |              |
- * | 26    | locations.geonames_details.name                     |              |
- * | 27    | names.types.acronym                                 |              |
- * | 28    | names.types.alias                                   |              |
- * | 29    | names.types.label                                   | ror_settings |
- * | 30    | names.types.ror_display                             | ror_settings |
- * | 31    | ror_display_lang                                    | ror_settings |
- * | 32    | relationships                                       |              |
- * | 33    | status                                              | ror_settings |
- * | 34    | types                                               |              |
+ * | json path                                  | ojs                                                                               |
+ * |--------------------------------------------|-----------------------------------------------------------------------------------|
+ * | id                                         | rors.ror                                                                          |
+ * | status                                     | rors.is_active                                                                    |
+ * | names[].value where types[] = ror_display  | ror_settings['locale' => lang, 'setting_name' => 'name', 'setting_value' => '…'] |
+ * | names[].value where types[] = label        | ror_settings['locale' => lang, 'setting_name' => 'name', 'setting_value' => '…'] |
+ * | names[].lang  where types[] = ror_display  | rors.display_locale                                                               |
  */
 
 namespace PKP\task;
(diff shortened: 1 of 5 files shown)


classes/task/UpdateRorRegistryDataset.php
```diff
@@ -239,102 +244,128 @@ private function downloadAndExtract(string $downloadUrl, string $pathZipFile, st
...
+            );
+            return false;
+        }
+    }
+
+    /**
+     * Process JSON file.
      */
-    private function processCsv(string $pathCsv): void
+    private function processJson(string $pathJson): void
```
touhidurabir: Here it returns void, and `executeActions()` unconditionally returns true after calling it . If every record fails the `empty()` check, the task reports success while writing zero rows. There's no count-based or error-based feedback. 

should we return a processed count from processJson() and log a warning (or return false) if zero records were processed ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -239,102 +244,128 @@ private function downloadAndExtract(string $downloadUrl, string $pathZipFile, st
...
-                    $searchPhrase .= ' ' . $namesOut[$name[0]];
-                }
+        // Process names array to extract display name, locale, and all name variants
+        $displayLocale = $this->noLocale;
+        $namesByLocale = [];
+        $firstLabelName = null;
+
+        foreach ($record['names'] as $nameEntry) {
+            // according to the ROR schema, 'value' is required for each name entry
+            $name = $nameEntry['value'];
```
touhidurabir, line 372: The above `empty()` guard checks that `$record['names']` is non-empty, but processRow() accesses `$nameEntry['value']` without a null check. Per the ROR schema, value is required, but defensive code should handle a malformed record without crashing the entire remaining dataset. 

If any single record has a name entry missing value, an Undefined index notice (or in strict mode, an exception) would propagate up to the top-level `catch(Throwable)`, aborting all remaining records.

So should we validate the `$nameEntry['value']` just in case to complete other ones if the dataset at ROR get is messed up ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -192,21 +171,35 @@ private function getDownloadData(): array
...
+     * Download and extract zip file.
      */
-    private function downloadAndExtract(string $downloadUrl, string $pathZipFile, string $pathZipDir, string $checksum): bool
+    private function downloadAndExtract(string $downloadUrl, string $pathZipFile, string $pathZipDir, string $checksumAlgorithm, string $checksum): bool
     {
         try {
             $client = Application::get()->getHttpClient();
 
             // Download the file
             $client->request('GET', $downloadUrl, ['sink' => $pathZipFile]);
```
touhidurabir: Should we add a timeout here for max download time ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -121,66 +71,95 @@ public function executeActions(): bool
...
     }
 
     /**
-     * Get download url
+     * Get download URL and checksum from the Zenodo API.
      */
     private function getDownloadData(): array
     {
         try {
             $client = Application::get()->getHttpClient();
```
touhidurabir, line 131: Here a timeout for max time to get the json data instead of having it run for as long as possible ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -344,118 +375,94 @@ private function processRow(array $row): array
...
-
-        // truncate temporary table
-        DB::table($this->temporaryTable)->truncate();
+            foreach ($staleByRorId as $rorId => $staleLocales) {
+                DB::table('ror_settings')
+                    ->where('ror_id', $rorId)
+                    ->where('setting_name', 'name')
+                    ->whereIn('locale', $staleLocales)
+                    ->delete();
+            }
```
touhidurabir: instead of deleting one by one which kind of introduce N+1 issue, can we not collected all the collect all stale `(ror_id, locale)` pairs and delete in fewer queries ?

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -239,102 +244,128 @@ private function downloadAndExtract(string $downloadUrl, string $pathZipFile, st
...
+
+    /**
+     * Process JSON file.
      */
-    private function processCsv(string $pathCsv): void
+    private function processJson(string $pathJson): void
     {
-        $isHeader = true;
-        $batchCounter = 0;
         $batchSize = 5000;
```
touhidurabir: A very edge case but this may cause an issue for PostgreSQL with its MVCC where it manage data consistencey but with slow performance . So we have like 5000 ROR records × ~2/4 locales each = ~10000/20000 ror_settings rows per upsert + the SELECT for stale detection which may turn into quite big in that case . perhaps reduce it to 2500 to 3000 ?

classes/task/UpdateRorRegistryDataset.php
bozana: I will then consider timeout = 30 seconds here

classes/task/UpdateRorRegistryDataset.php
bozana: Here maybe: 'connect_timeout' => 30, 'read_timeout' => 3600 -- 30 seconds to establish the connection, and abort if no data is received for 60 minutes (detecting a stalled transfer without killing a legitimately slow one). What do you think?

bozana:
Hi @touhidurabir, I considered all comments. Could you please take another look?
Thanks a lot!

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -239,102 +251,161 @@ private function downloadAndExtract(string $downloadUrl, string $pathZipFile, st
...
-                    $namesOut[$name[0]] = trim($name[1]);
-                    $searchPhrase .= ' ' . $namesOut[$name[0]];
-                }
+        // Process names array to extract display name, locale, and all name variants
+        $displayLocale = $this->noLocale;
+        $namesByLocale = [];
+        $firstLabelName = null;
+        $firstValidName = null;
+
+        foreach ($record['names'] as $nameEntry) {
```
touhidurabir, line 360: better to do it as `foreach ($record['names'] ?? [] as $nameEntry)` so that if there is `name` index, no warning being logged, but it will not cause to crash so nothing critical .

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -192,21 +178,35 @@ private function getDownloadData(): array
...
      */
-    private function downloadAndExtract(string $downloadUrl, string $pathZipFile, string $pathZipDir, string $checksum): bool
+    private function downloadAndExtract(string $downloadUrl, string $pathZipFile, string $pathZipDir, string $checksumAlgorithm, string $checksum): bool
     {
         try {
             $client = Application::get()->getHttpClient();
 
             // Download the file
-            $client->request('GET', $downloadUrl, ['sink' => $pathZipFile]);
+            $client->request('GET', $downloadUrl, ['sink' => $pathZipFile, 'connect_timeout' => 30, 'read_timeout' => 3600]);
```
touhidurabir: Here we are setting a very high `read_timeout` and no `timeout` . So `read_timeout` is  max time to wait between receiving data chunks from the server that is reset every time data arrives . and `timeout` set the max total wall-clock time for the entire request (connect + send + receive) . So my suggestion is that we set the `read_timeout` to `300` (5 mins) which is still very generous for inter-chunk silence on a 30MB file. And set a `timeout` to `600` (10 mins) as overall wall-clock limit for the entire download.

classes/task/UpdateRorRegistryDataset.php
```diff
@@ -192,21 +178,35 @@ private function getDownloadData(): array
...
+                    ScheduledTaskHelper::SCHEDULED_TASK_MESSAGE_TYPE_ERROR
+                );
                 return false;
             }
 
             // Verify the checksum of the downloaded file
-            $fileChecksum = md5_file($pathZipFile);
+            $fileChecksum = match($checksumAlgorithm) {
+                'md5' => md5_file($pathZipFile),
+                'sha256' => hash_file('sha256', $pathZipFile),
```
touhidurabir, line 201: just a suggestion to add a `default` as `throw new Exception("Unsupported checksum algorithm: {$checksumAlgorithm}")` that will logged a more meaningful reason , but nothing critical .

classes/task/UpdateRorRegistryDataset.php
bozana: because we check that $record['names'] is not empty in line 313 i.e. ensure that the name index always exists here, I will leave it as it is...

bozana:
Hi @touhidurabir, I have considered the comments and will merge as soon as tests pass.
Thanks a lot! Very helpful!!!

PR #12492, opened 2026-03-23
I12489 - PluginSettingsController

jardakotesovec:


classes/plugins/PluginSettingsController.php
```diff
@@ -0,0 +1,52 @@
...
+        return 'plugins/' . $this->plugin->getName() . '/settings';
+    }
+
+    public function getRouteGroupMiddleware(): array
+    {
+        $middleware = ['has.user'];
+        $roles = [Role::ROLE_ID_SITE_ADMIN];
+
+        if (!$this->plugin->isSitePlugin()) {
+            $middleware[] = 'has.context';
```
touhidurabir, line 50: This is correct but partially . To make the API route site wide, we need to have the `PKPBaseController::isSiteWide()` return `true` so we need to do something like 

```php
public function isSiteWide(): bool
{
    return $this->plugin->isSitePlugin();
}
```

otherwise it will not handle the sitewide api urls like `BASE_URL/index.php/index/api/v1/...` . 

so we should have default implementation here .

classes/plugins/PluginSettingsController.php
```diff
@@ -0,0 +1,52 @@
...
+
+abstract class PluginSettingsController extends PKPBaseController
+{
+    public function __construct(
+        protected Plugin $plugin
+    ) {}
+
+    public function getHandlerPath(): string
+    {
+        return 'plugins/' . $this->plugin->getName() . '/settings';
```
touhidurabir: we can do `"plugins/{$this->plugin->getName()}/settings"` . here in below for routes names. just a suggestion, not important .

classes/plugins/PluginSettingsController.php
```diff
@@ -0,0 +1,52 @@
...
+
+        $middleware[] = self::roleAuthorizer($roles);
+
+        return $middleware;
+    }
+
+    public function getGroupRoutes(): void
+    {
+        Route::get('', $this->get(...))->name('plugin.' . $this->plugin->getName() . '.settings.get');
+        Route::put('', $this->edit(...))->name('plugin.' . $this->plugin->getName() . '.settings.edit');
```
touhidurabir: This make sense as we can not have a default or abstract method for this 2 methods due to PHP's parameter type contravariance rules . Either we have it this way to nothing at all for the `getGroupRoutes` method . How about add some comments above to let plugin developer know that if they want to use these 2 method, they need to implement those in the extending class or if not needed, just override the `getGroupRoutes` method .

touhidurabir (changes requested):
@jardakotesovec I added few comments, most important one is related to the site wide api routes . 

As I can not add a non changed file in review, I am thinking that as our target is to reduce boilerplate code for plugin developer, how about we also simplify the setting api controller registration by adding following in the PKP\plugins\Plugin as 

in the PKP\plugins\Plugin::register add at the of `__construct` method
```php
$this->registerPluginSettingApiRoute();
```

then as following implementation as 
```php
public function getSettingsController(): ?PluginSettingsController
{
    return null;
}

protected function registerPluginSettingApiRoute(): void
{
    if ($settingsController = $this->getSettingsController()) {
     Hook::add('APIHandler::endpoints::plugin', function (string $hookName, \PKP\core\APIRouter $apiRouter) use ($settingsController): bool {
         $apiRouter->registerPluginApiControllers([$settingsController]);
         return Hook::CONTINUE;
     });
 }
}
```

Then plugin developer does not need to implement the hook callback but only over write the getSettingsController method and have an extended class of PluginSettingsController like in their plugin class 
```php
public function getSettingsController(): ?PluginSettingsController
{
    // plugin setting api controller 
    return new PluginSettiingApiControllerOrWhateverTheyNamed($this);
}
```

Something like above(not tested code but something similar can work), however not 100% sure if that make it simple or complex to handle in future .

classes/plugins/PluginSettingsController.php
jardakotesovec: @touhidurabir I did test this and worked fine without this setting (screenshot attached) There is some additional magic happening with `matchesPluginHandlerPath` - so maybe this is not needed for plugin controllers? 

I also now tried with `isSiteWide` returning true. And does not make difference.  So might be good to double check what we need to follow there for plugins?

I duplicated the allowedUploads plugin with different name and its configured to be site plugin - just for testing this. 

So feel free to clone this plugin to check it yourself - https://github.com/jardakotesovec/allowedUploadsSite 

Screenshot 2026-03-26 at 14 26 41

classes/plugins/PluginSettingsController.php
jardakotesovec: Make sense, will add comment for that

jardakotesovec:
@touhidurabir Thanks, regarding this simplification I will consider your suggestion. This would reduce the boilerplate, but on the other hand it would make the registration settings controller using unique pattern. So maybe showcasing how to register plugin controller is good thing as they can reuse that pattern for additional APIs they might need. And that PluginSettingsController base class which covers good example for permissions etc might be good balance.

classes/plugins/PluginSettingsController.php
touhidurabir: Interesting, I will check it as last time I checked, it was necessary if I remember correctly .

touhidurabir: @jardakotesovec sorry for the confusion as I failed to explain it (because I kind of forgot it as it's been quite some time since I worked on https://github.com/pkp/pkp-lib/issues/7698) the proper use of `isSiteWide`. 

So the purpose of it not to lockdown the API route but restrict the generation of API routes itself . as with `isSiteWide` retuning
 - `true`, having a path like `BASE_URL/index.php/index/api/v1/custom-admin-plugin-path` is ok but it will not work for `BASE_URL/index.php/CONTEXT_PATH/api/v1/custom-admin-plugin-path` so that API routes will not be available at context level but only at Administrator level . 
 - `false`, both the `BASE_URL/index.php/index/api/v1/custom-admin-plugin-path` and `BASE_URL/index.php/CONTEXT_PATH/api/v1/custom-admin-plugin-path` will works , e.g. the API path will be available both the Context and Administrator level . 
 
I think for plugin that is site level have these default as 
```php
public function isSiteWide(): bool
{
    return $this->plugin->isSitePlugin();
}
```
but that really depends on the plugin's purpose and requirement . Now specifically for settings url, I think a site level plugin's settings path should not be accessible at context level so it should have the above one by default . And plugin devs can still override . 

But I leave that decision to you , we need to have a good doc anyway for all these cases and API level routing starting form 3.5 as a lot of stuff have changed and added .

jardakotesovec: @touhidurabir I see, thats great explanation. I will add it. I found couple hints in code base that assume that plugin either works on site level or context level.. therefore I think for settings its appropriate just to link to one or the other.

Testing it now

Issue #12493, opened 2026-03-23
[Reviewer Suggestion] Locales to render & validate form does not match


jardakotesovec:
### Description

Currently Reviewer Suggestion are using `Metadata locales` to render the form. But on the validation step its using the `Submission locales` list.

Therefore if these locales are configured differently it leads to this error.

I think probably best fit would be 'Form' locales. As thats for example whats used for the reviewer forms. Instead of submission specific locales.

<img width="1097" height="622" alt="Image" src="https://github.com/user-attachments/assets/067e653a-946c-4d19-98c8-b97491f5f9ee" />

### Steps to Reproduce

1. Enable reviewer Suggestion from Settings --> Workflow --> Review --> Setup --> Reviewer Suggestion at Submission
2. Start new submission
3. Add new reviewer suggestion - enter data for locale which is not part of the  Submission locales, but its in Metadata locales
4. Submit & see error

### Expected Result

I believe this form should be presented using the 'Form locales',  not submission specific locales (neither submission nor metadata)

### Actual Result

Reviewer suggestion form is presented with metadata locales, which leads to the error when saving via api.

### Application Version

OJS,OMP  3.5, main

### Additional Information

<img width="1166" height="500" alt="Image" src="https://github.com/user-attachments/assets/21bce6da-bbc5-4126-8220-f3ffe206f6a1" />
<img width="1258" height="744" alt="Image" src="https://github.com/user-attachments/assets/4d12f67b-d3c8-41f0-a26f-6a164b16ba81" />

**PRs**
**3.5**
ojs: https://github.com/pkp/ojs/pull/5459
pkp-lib: https://github.com/pkp/pkp-lib/pull/12494
ui-library: https://github.com/pkp/ui-library/pull/865

**PRs**
**Main**
ojs: https://github.com/pkp/ojs/pull/5489
pkp-lib: https://github.com/pkp/pkp-lib/pull/12560
ui-library: https://github.com/pkp/ui-library/pull/872

jardakotesovec:
@bozana @touhidurabir Could you please review? Especially review the conclusion which locales should be used for Reviewer Suggestions.

touhidurabir:
@jardakotesovec the above PR does fix the issue but I think not the right way . So for the following locale setting

<img width="2014" height="1228" alt="Image" src="https://github.com/user-attachments/assets/44c303b1-9b0e-4dac-8087-4dc7c6a8b235" />

[...]

PR #12494, opened 2026-03-23
pkp/pkp-lib#12493 Use Form locales for ReviewerSuggestions

jardakotesovec:


api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
```diff
@@ -72,8 +75,7 @@ public function rules(): array
                 Rule::exists('users', 'user_id'),
             ],
             'familyName' => [
-                'required',
-                // 'multilingual:en,fr_CA', // Alternative way to do multilingual validation, do not use HasMultilingualRule trait with this approach
+                'sometimes',
```
touhidurabir, line 76: as this can be non mandatory, better to add a `nullable` rule with `sometimes` . Due the global API middleware `ConvertEmptyStringsToNull::class`, any empty string as `''` passed will be converted to null but without `nullable` will fail .

api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
```diff
@@ -72,8 +75,7 @@ public function rules(): array
                 Rule::exists('users', 'user_id'),
             ],
             'familyName' => [
-                'required',
-                // 'multilingual:en,fr_CA', // Alternative way to do multilingual validation, do not use HasMultilingualRule trait with this approach
```
touhidurabir, line 76: The purpose of this commented out message is to demonstrate another alternative way to run MultilingualInput validation from the FormRequest . Better to move this comment for `givenName` for future reference unless we add a better doc about these . The whole new mechanism has been introduced as part of initial PR of reviewer suggestion (e.g. Form request validation )

api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
```diff
@@ -38,20 +37,24 @@ public function multilingualInputs(): array
...
+    public function allowedLocales(): array
     {
-        return Application::get()->getRequest()->getContext()->getSupportedSubmissionLocales();
+        $locales = Application::get()->getRequest()->getContext()->getSupportedFormLocales();
+        $sitePrimaryLocale = Application::get()->getRequest()->getSite()->getPrimaryLocale();
+
+        if (!in_array($sitePrimaryLocale, $locales)) {
+            $locales[] = $sitePrimaryLocale;
+        }
+
```
touhidurabir: dont think we need to include the primary locale in the optional allowed locale so we can just do `return Application::get()->getRequest()->getContext()->getSupportedFormLocales();` .

touhidurabir (commented):
@jardakotesovec I added few suggestions, please take a look . Otherwise it looks good . 

Edit 1: 
As we make the `familyName` optional , better to update the `PKP\submission\reviewer\suggestion\ReviewerSuggestion::fullName` as the line `->map(fn ($givenName, $locale) => $givenName . ' ' . $this->familyName[$locale])` should cause issue with null familyName. perhaps we can do is 
```php
return Attribute::make(
    get: fn () => collect($this->givenName)
        ->map(fn ($givenName, $locale) => $givenName . ' ' . ($this->familyName ? ($this->familyName[$locale] ?? '') : ''))
        ->toArray()
)->shouldCache();
```

Edit 2 : 
Both the OJS and OMP has a app level (not at the shared `pkp-lib`) `submission.json` where we need to update `reviewerSuggestions --> items --> properties --> familyName` with gave 
```json
"validation": [
	"nullable"
]
```

pages/submission/PKPSubmissionHandler.php
```diff
@@ -228,7 +229,20 @@ protected function showWizard(array $args, Request $request, Submission $submiss
...
 
         if ($context->getData('reviewerSuggestionEnabled')) {
-            $reviewerSuggestionsListPanel = $this->getReviewerSuggestionsListPanel($request, $submission, $publication, $formLocales);
+            $supportedFormLocaleNames = $context->getSupportedFormLocaleNames();
+
+            // Ensure the site primary locale is available for reviewer name fields
+            $sitePrimaryLocale = $request->getSite()->getPrimaryLocale();
+            if (!isset($supportedFormLocaleNames[$sitePrimaryLocale])) {
+                $supportedFormLocaleNames[$sitePrimaryLocale] = Locale::getMetadata($sitePrimaryLocale)?->getDisplayName() ?? $sitePrimaryLocale;
+            }
```
bozana, line 237: I think it should rather be something like:
```
$supportedFormLocales = $context->getSupportedFormLocales();
$sitePrimaryLocale = $request->getSite()->getPrimaryLocale();
if (!isset($supportedFormLocales[$sitePrimaryLocale])) {
    $supportedFormLocales[] = $sitePrimaryLocale;
}
$supportedFormLocaleNames = Locale::getFormattedDisplayNames($supportedFormLocales, null, LocaleMetadata::LANGUAGE_LOCALE_WITHOUT);

```

Only to have country displayed in case there are two same languages from different countries.
For example: if context supported form locales are en and fr_CA, and site primary locale is fr.
Locale::getFormattedDisplayNames handles it properly. 

(I see this use case is not correct in those legacy forms, but in the new forms and also in new suggest review form)

api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
jardakotesovec: Its the **site** primary locale. 

As for user details we require site primary locale. getSupportedFormLocales considers only **context locales**.

api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
jardakotesovec: Oki, moved that to firstName

api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
touhidurabir: Yes, that is correct . I think the name `allowedLocale` is not clear enough . It's the additional optional but allowed ones . As we set the primary ones above, it's not needed to be in the list of allowed optional ones . Does not break anything as there are checks but should not be in that list .

Issue #11760, opened 2025-08-29
When errors occurr while submitting a review, the way OJS displays errors is misleading.


pilasou:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

If reviewers try to submit an incomplete review, OJS will only display a temporary notice on the top right of the screen, but it will also stay at the bottom of the page and reset the recommendation selected. If the reviewers do not notice the message, they may leave thinking that their review has been submitted…when it is not and their review is lost.

### Steps to Reproduce

- Connect as an editor create a form with mandatory answers
- Invite a reviewer and select the new form
- Log in as Reviewer, accept the invitation and move to the form page (step 3)
- Go the bottom of the page, select a recommendation
- Clic on Submit button. A window display asking to confirm submission

<img width="518" height="314" alt="Image" src="https://github.com/user-attachments/assets/aec1cd63-36ea-45e8-af7d-086307cece6c" />

- Clic OK
- A notice will display on the top right, the recommendation is reset. 

<img width="1143" height="562" alt="Image" src="https://github.com/user-attachments/assets/df2649b8-58ee-4ece-9b17-43b521c1ab95" />

### Expected Result

When reviewers try to submit an incomplete review: 

- The recommendation must not be reset
- The way to handle errors may be align with the one already in place in OJS 3.5 for authors submissions: 

1.  Filed required are clearing identifier with a message: 

<img width="459" height="80" alt="Image" src="https://github.com/user-attachments/assets/08e84b36-3d56-4344-b5de-3ef7ea74b245" />

2. There is a count on the error that is displayed directly to authors, with the ability to “jump” to one: 

<img width="991" height="146" alt="Image" src="https://github.com/user-attachments/assets/501f322a-c63e-4ebc-9536-7be838d4c572" />

### Actual Result

The review's recommendation is reset. The screen does not display on the page the errors/required fields. 

### Application Version

OJS 3.3, .3.4 and 3.5 (can be reproduced on OJS Testdrive).

Tribunal33:
@pilasou I would not consider this to be a bug.  This seems like an improvement to existing functionality.  I have updated the ticket and included the milestone for 3.6.

pilasou:
@Tribunal33 Thanks! I agree, more an improvment than a bug. How should improvment to existing features be reported ? I used the issue template because I cannot find one for improvments (there is one for issues/bugs, and for feature, it redirects to the forum).

Tribunal33:
[...]

PR #12496, opened 2026-03-23
pkp/pkp-lib#11760 Add client side validations for reviewer custom forms

nibou230:
For https://github.com/pkp/pkp-lib/issues/11760

Added JS to handle the validation on required custom form fields.

Also, there is a new general validation message (french and english only) at the bottom of the form.

### DIFF
--- a/locale/en/reviewer.po
+++ b/locale/en/reviewer.po
@@ -195,6 +195,9 @@ msgstr ""
 msgid "reviewer.submission.reviewRound.reviewNotCompleted"
 msgstr "The review was not completed."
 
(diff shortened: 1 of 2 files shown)


asmecher:
@jardakotesovec, this is a work-around adding validation to the custom review forms, which are still old code; I think we can consider this a necessary evil until we refactor that code to the new UI tools. However, can you take a quick look from a front end perspective and see if it works for you?

asmecher:
Resolved per https://github.com/pkp/pkp-lib/issues/11760#issuecomment-4167047826

Issue #12154, opened 2025-12-17
Check ROR API response handling


bozana:
Update: after discussion and feedback from ROR, we will be leaving the v2 API URL but need to check that all error responses are being handled properly. 

- 410 Error: the linked ROR service is no longer available - the user should contact their system administrator, who will need to patch or upgrade the installation.
- 500-level Error: the ROR service is not available.

After the error is encountered, the user should still be able to enter a free text affiliation that is not linked to a ROR.

---
~~If ROR API version changes, and v2 is deprecated, all installation will need to patch. However it could be that our code would still work with the new version and that actually no changes are needed. In this case it would be better not to use v2 in the ROR API URL, as we currently use.~~

## Pull Requests

- ui-library 3.5: https://github.com/pkp/ui-library/pull/852
- pkp-lib 3.5: https://github.com/pkp/pkp-lib/pull/12499
- ui-library main: https://github.com/pkp/ui-library/pull/855
- pkp-lib main: https://github.com/pkp/pkp-lib/pull/12500

### Submodule updates/tests

- OJS 3.5: https://github.com/pkp/ojs/pull/5443
- OJS main (tests only): https://github.com/pkp/ojs/pull/5445

asmecher:
(related: https://github.com/pkp/pkp-lib/issues/12144)

kaitlinnewson:
Hi @bozana, I was reviewing the ROR API practices and found [this document linked](https://docs.google.com/document/d/18nl6pq0kdCU5ApcdbNjKnV7xHIw9eEY7DJG1WHjaLSs/edit?tab=t.0) on their website which notes the following about the unversioned API:

> Requests that do not include a version in the path portion will default to the current (unversioned) schema until that schema is sunset. At that point a version will be required, and requests without a version in the path portion will return a 410 Gone error, with a detailed message.

It sounds like removing the version in the path could lead to future problems. Maybe a config file setting for the API version or API url could work instead, then we could consider it if it exists?

bozana:
Hi @kaitlinnewson, thanks a lot! I am currently in connection with ROR team, to see how to best proceed. There are some things that we will need to consider differently, it seems :-) Once we defined it all, I will post the summary here or so.

asmecher:
Deferring -- but if it's ready in time for LTS, that's great too!

kaitlinnewson:
Hi @blesildaramirez and @bozana, one more for your review - thanks!

[...]

PR #12499, opened 2026-03-24
pkp/pkp-lib#12154 improve handling of ROR API error responses

kaitlinnewson:


locale/en/user.po
```diff
@@ -51,10 +51,10 @@ msgid "user.affiliations.translationsSomeAvailable"
 msgstr "{$count} of {$total} languages completed"
 
 msgid "user.affiliations.typeTranslationNameInLanguageLabel"
-msgstr "Type the institute name in {$language}"
+msgstr "Type the institution name in {$language}"
```
blesildaramirez, line 54: Since this is a change in an existing locale string, I wonder if we should flag this key with `#, fuzzy` ?
If yes, we need to add this flag to all non-English po files. Your commit would look like [this](https://github.com/pkp/pkp-lib/pull/12382/changes/d4debf78180e99e5a663a73e5a5660cd615b50ec).

Or maybe the translation to other languages are okay to leave as is, since this change looks more of an issue just for English locale.

kaitlinnewson: I thought it's likely an issue in the english strings only, but I'll mark them fuzzy just to be safe.

asmecher: I don't think you're talking about marking the English text fuzzy, but just in case, be careful about that; it'll mean nobody will be able to translate those strings to other languages in Weblate, and we've had cases before where they've been left fuzzy without anyone noticing.

The fuzzy tag is great to use for other languages when you want a translation review, though!

Issue #12497, opened 2026-03-24
[Main Feature] | Workflow Section - Recommend only option


Vitaliy-1:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Editors can change own recommend only flag. 

### Steps to Reproduce

1. Login as a user with journal manager role.
2. Navigate to the submission where section editor is assigned.
3. Click to add recommend only flag to the section editor
4. Login as the section editor
5. Click on the `Edit` button under the own stage assignment
6. Click on the `Ok` button
7. See the message "The stage assignment has been changed."  
8. Login back as a journal manager
9. See that the section editor doesn't have recommend only option checked anymore. 

### Expected Result

Editors shouldn't be able to change own recommend only status

### Actual Result

Editors are able to change own recommend only status

### Application Version

OJS 3.4, 3.5, main 

### Additional Information

<img width="805" height="421" alt="Image" src="https://github.com/user-attachments/assets/97758bfa-fb4b-41c8-8cff-02a5cd70f320" />

<img width="801" height="237" alt="Image" src="https://github.com/user-attachments/assets/3322fe00-1be2-4e7f-b162-48c1175b7fc2" />

<img width="325" height="73" alt="Image" src="https://github.com/user-attachments/assets/b4033b31-953f-4e7b-99ab-8aa1b1ac7b65" />

Vitaliy-1:
It looks like the fix of the current logic would be to disallow the user with recommend only status to change it for other users as well.

Vitaliy-1:
PRs

stable 3.4.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/12503
tests: https://github.com/pkp/ojs/pull/5448

stable 3.5.0
pkp-lib: https://github.com/pkp/pkp-lib/pull/12525
ui-library: https://github.com/pkp/ui-library/pull/868
tests: https://github.com/pkp/ojs/pull/5468

main
pkp-lib: https://github.com/pkp/pkp-lib/pull/12527
ui-library: https://github.com/pkp/ui-library/pull/871
tests: https://github.com/pkp/ojs/pull/5469

Vitaliy-1:
@bozana, can you take a look?

bozana:
[...]

PR #12503, opened 2026-03-26
pkp/pkp-lib#12497 Editors shouldn't change own recommend-only status

Vitaliy-1:


classes/security/Validation.php
```diff
@@ -598,6 +602,63 @@ public static function getAdministrationLevel(int $administeredUserId, int $admi
...
+            return false;
+        }
+
+        // Editors aren't allowed to edit managers' assignments
+        $editableUser = Repo::user()->getCollector()->filterByUserIds([$stageAssignment->getUserId()])->getMany()->first();
+        if ($editableUser->hasRole([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $submission->getContextId())) {
+            return false;
+        }
+
+        return true;
```
bozana, line 660: Hi @Vitaliy-1, should this function also check if the user has a recommend-only status?

bozana: Oh, I think I understand: the recommend-only can edit, but it cannot change the status (which is checked in AddParticipantFrom::execute())?

bozana (approved):

Issue #12338, opened 2026-02-12
Assess and implement reauthorization for sensitive areas of the application


asmecher:
Currently, OJS/OMP/OPS do not distinguish authentication levels for content vs. administration areas of the system -- once logged in, a user with administration access may wander between editorial and administrative areas.

The European Commission requires differentiated access in its Web Application Security Standard:

> 10. Administration pages of an application containing sensitive non-classified information shall have stricter security requirements which may justify the need for a re-authentication and/or for a stronger authentication.

(This is tracked in our IT Security Plan as WASS.10.)

Recommended: implement re-authorization when entering Administration if this has not been done within a short time window (similar to Github's challenge for settings areas); consider the same for context settings.

taslangraham:
Implementing an approach similar to Github's [Sudo Mode](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/sudo-mode).

taslangraham:
@asmecher Would we want to consider making this feature configurable at the Context level? Therefore allowing individual Contexts to decide whether or not reauthorization should be required when accessing Context settings?


To require reauthorization for Administration, we can configure that via and entry in `config.inc.php`

asmecher:
@taslangraham, I think we can consider this only relevant to the Administration area, which is only available to the Site Admin and is not context-dependent. Therefore, it would only need to be configured in `config.inc.php` if anywhere -- a single setting for the amount of time a reauthentication action is valid, with a `0` to make it forever (as before).

taslangraham:
@touhidurabir Could you review the PRs listed at https://github.com/pkp/pkp-lib/issues/12338#issue-3934371867, please? Tagging you as well @asmecher, as you might be interested in taking a look too. Thanks in advance!

touhidurabir:
@taslangraham review added at https://github.com/pkp/pkp-lib/pull/12505#pullrequestreview-4025518109 and https://github.com/pkp/ui-library/pull/858#pullrequestreview-4025579307 .

touhidurabir:
@taslangraham thinking a bit more about this, this is very similar to https://laravel.com/docs/11.x/authentication#password-confirmation but we can not use that directly as our page router is not based on laravel routing system like the API . So comparing to this, do we need a vue component for this confirmation page or does it bring any extra benefit comparing to a simple backend tpl/blade template ?

taslangraham:
[...]

PR #12505, opened 2026-03-26
pkp/pkp-lib#12338 Require re-authentication when accessing administration area

taslangraham:
For pkp/pkp-lib#12338


pages/admin/AdminHandler.php
```diff
@@ -786,4 +795,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
             ]
         );
     }
+
+    /**
+     * Display the page for admin to re-authenticate and confirm access to the Administration area of the site.
+     */
+    public function confirmAccess(array $args, $request): void
```
taslangraham: Added this and the `confirmAccessSubmit` method directly to the `adminHandler` since re-authentication is only relevant to admins, but it could also be moved to the `loginHandler` and made accessible to only admins.

classes/security/authorization/AdminReauthenticationRequiredPolicy.php
```diff
@@ -0,0 +1,83 @@
...
+use PKP\security\Role;
+use PKP\security\Validation;
+
+class AdminReauthenticationRequiredPolicy extends AuthorizationPolicy
+{
+    private PKPRequest $request;
+
+    /**
+     * @param PKPRequest $request
+     * @param array $roleIds - Array of Role IDs that are allowed to access the app area that requires sudo mode
```
touhidurabir: This description is wrong as roles id are not passed here .

classes/core/PKPSessionGuard.php
```diff
@@ -45,6 +46,8 @@ class PKPSessionGuard extends SessionGuard
      */
     protected $provider;
 
+    // Session key to store the timestamp when a user reauthenticated.
+    private const SESSION_KEY_REAUTHENTICATED_AT = 'reauthenticated_at';
```
touhidurabir: No new line after this const been defined . I suggest that move this const at the very beginning of class before any props .

classes/security/authorization/AdminReauthenticationRequiredPolicy.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @ingroup security_authorization
+ *
+ * @brief Policy to require admin to reauthenticate when accessing Administration pages.
+ */
+
+namespace PKP\security\authorization;
+
+use APP\core\Application;
+use PKP\config\Config;
```
touhidurabir: This and few other class import are not being used .

classes/core/PKPSessionGuard.php
```diff
@@ -301,4 +304,63 @@ protected function getCookieExpirationDate(array $config): DateTimeInterface|int
             Carbon::now()->addRealMinutes($config['lifetime'])
         );
     }
+
+    /**
+     * Check if user currently has access to sensitive area(Administration) of the app.
+     * This is only applicable for site admins.
+     */
+    public static function isElevatedSessionActive(): bool
```
touhidurabir: do these methods must be `static` as this feature and check only applicable for logged in user so we should be able to use public methods. this and other methods below .

classes/security/Validation.php
```diff
@@ -587,4 +587,13 @@ public static function canUserLoginAs(
         }
         return self::getAdministrationLevel($targetUserId, $currentUserId, $contextId) === self::ADMINISTRATION_FULL;
     }
+
+    /**
+     * Check if Admins are required to reauthenticate when accessing the admin area.
+     */
+    public static function isAdminReauthenticationRequired(): bool
```
touhidurabir: lets call it `isReauthenticationRequired` as for now this is only for admin section but it can be applied for any parts later .

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+    {
+        $this->setupTemplate($request);
+        $templateMgr = TemplateManager::getManager($request);
+        $submitUrl = $request->url(null, 'admin', 'confirmAccessSubmit');
+
+        if (Config::getVar('security', 'force_login_ssl')) {
+            $submitUrl = preg_replace('/^http:/', 'https:', $submitUrl);
+        }
```
touhidurabir: Do we need it specifically here as our PKPHandler::authorize as a SSL policy applied ?

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+
+    /**
+     * Handle submission of form to confirm access to the Administration area of the site.
+     */
+    public function confirmAccessSubmit($args, Request $request): void
+    {
+        if (Config::getVar('security', 'force_login_ssl') && $request->getProtocol() != 'https') {
+            // Force SSL connections for auth verification
+            $request->redirectSSL();
+        }
```
touhidurabir, line 867: same question as above, do we need this check as our parent authorizer have a SSL policy ?

classes/core/PKPSessionGuard.php
```diff
@@ -301,4 +304,63 @@ protected function getCookieExpirationDate(array $config): DateTimeInterface|int
...
+            return false;
+        }
+
+        $timeoutSeconds = $timeoutMinutes * 60;
+        $elapsedSeconds = time() - $lastReauthenticationTimestamp;
+
+        $isWithinWindow = $elapsedSeconds < $timeoutSeconds;
+        // Reset window time while user is actively navigating around the sensitive area of app.
+        // This prevents the elevated access from expiring while user is actively using that sensitive area.
+        self::startElevatedSession();
```
touhidurabir: if I understand correctly, this is a critical miss as it allows to bypass the required re-auth in following flow 
1. Admin authenticates, timeout expires (e.g. 1 min)
2. Visit /admin/settings → isElevatedSessionActive() runs: $isWithinWindow = false, but startElevatedSession() resets timestamp to now, returns false
3. Policy denies, redirects to confirmAccess
4. Admin cancels (or presses Back)
5. Visit /admin/settings again within 1 min → timestamp was just set in step 2, so $isWithinWindow = true
6. Visit /admin/settings again within 1 min → timestamp was just set in step 2, so $isWithinWindow = true

I think it should be like 
```php
if ($isWithinWindow) {
     self::startElevatedSession();
 }
```

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+            'pageWidth' => TemplateManager::PAGE_WIDTH_NARROW,
+        ]);
+
+        $templateMgr->display('templates/admin/confirmAccess.tpl');
+    }
+
+    /**
+     * Handle submission of form to confirm access to the Administration area of the site.
+     */
+    public function confirmAccessSubmit($args, Request $request): void
```
touhidurabir: As this evaluate for POST, need CSRF check .

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+
+        $templateMgr->setState([
+            'pageInitConfig'=>[
+                'source' => $request->getUserVar('source'),
+                'cancelUrl' => $request->getUserVar('cancelUrl'),
+                'error' =>  __('user.login.loginError'),
+            ]
+        ]);
+
+        $templateMgr->display('templates/admin/confirmAccess.tpl');
```
touhidurabir: for template display, we do not include the `template` in the template path .

locale/en/user.po
```diff
@@ -892,3 +883,11 @@ msgstr "Bio Statement"
 
 msgid "user.removeRole.roleRemainMessage"
 msgstr "You cannot remove the role. At least one role must be assigned to the user."
+
+msgid "user.confrimAccess"
+msgstr "Confirm Access"
+
+msgid "user.confrimAccess.description"
```
touhidurabir: should be `confirm` .

classes/core/PKPRequest.php
```diff
@@ -784,4 +784,20 @@ public function getPublicFilesUrl(?Context $context = null): string
...
         ]);
     }
+
+    /**
+     * Get the path of the referrer URL.
+     * @return string|null The referrer path along with any available query params, or null if the referrer is not set.
+     */
+    public function getReferrerPath(): ?string
+    {
+        $referer = $_SERVER['HTTP_REFERER'];
```
touhidurabir: should be `$_SERVER['HTTP_REFERER'] ?? null` otherwise there will be warning to access undefined index .

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+        if (!$source) {
+            $pkpPageRouter = $request->getRouter(); /** @var \PKP\core\PKPPageRouter $pkpPageRouter */
+            $pkpPageRouter->redirectHome($request);
+        }
+
+        $templateMgr->setState([
+            'pageInitConfig' => [
+                'source' => $request->getUserVar('source'),
+                'submitUrl' => $submitUrl,
+                'cancelUrl' =>  $request->getReferrerPath() ?: $request->getIndexUrl(),
```
touhidurabir: I think this can cause a redirect loop if admin comes from a another admin level page that and we get redirected to that page as that is also under the re-auth requirement

classes/core/PKPSessionGuard.php
```diff
@@ -301,4 +304,63 @@ protected function getCookieExpirationDate(array $config): DateTimeInterface|int
...
+
+        // If reauthentication is not required then Admin always has elevated access.
+        if(!Validation::isAdminReauthenticationRequired()){
+            return true;
+        }
+
+        $timeoutMinutes = (int)Config::getVar('security', 'admin_reauthentication_timeout');
+
+        $request = Application::get()->getRequest();
+        $session = $request->getSession();
```
touhidurabir: Once we make this methods instance method, we can do `$this->getSession()` and that should work unless I am missing something here .

pages/login/LoginHandler.php
```diff
@@ -501,6 +502,8 @@ public function signInAsUser($args, $request)
 
             if (isset($newUser) && $sessionGuard->getUserId() != $newUser->getId()) {
                 $request->getSessionGuard()->signInAs($newUser);
+                // Remove admin elevated session info if it exists
+                PKPSessionGuard::stopElevatedSession();
```
touhidurabir: rather than putting it here, I think we have it in the `PKPSessionGuard::signInAs()` so that we dont need to add this other places later .

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
             ]
         );
     }
+
+    /**
+     * Display the page for admin to re-authenticate and confirm access to the Administration area of the site.
+     */
+    public function confirmAccess(array $args, $request): void
```
touhidurabir: The `$request` is not type hinted .

pages/admin/AdminHandler.php
touhidurabir: `$args` not type hinted . And for `$request` we have `PKPRequest` instead of `Request` for other methods .

pages/admin/AdminHandler.php
```diff
@@ -786,4 +798,82 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
+
+        $user = Application::get()->getRequest()->getUser();
+        $userName = $user->getUsername();
+        $password = $request->getUserVar('password');
+
+        if (Validation::checkCredentials($userName, $password)) {
+            $source = str_replace('@', '', $request->getUserVar('source'));
+            PKPSessionGuard::startElevatedSession();
+
+            if (preg_match('#^/\w#', (string)$source) === 1) {
```
touhidurabir, line 908: Here should we also validated against the base url to see if this is application level or external url ?

classes/security/Validation.php
```diff
@@ -587,4 +587,13 @@ public static function canUserLoginAs(
...
+    {
+        $timeout = Config::getVar('security', 'admin_reauthentication_timeout');
```
touhidurabir: also how about calling it only `reauthentication_timeout` ? Or we can make it more aligned to laravel as `password_timeout` ?

classes/core/PKPSessionGuard.php
taslangraham: You're right. This was an oversight on my part. Thanks for catching this!

pages/admin/AdminHandler.php
taslangraham: Based on my understanding, yes it is needed.
The `SSL policy (`HttpsPolicy`) in PKPHandler::authorize` enforces HTTPS  by redirecting to HTTPS url after the request has already been received by the server. The code here ensures the form submits to an HTTPS URL from the start, instead of the unsecured HTTP url. Or am I missing something here?

touhidurabir: Probably I am misunderstanding here but if a form submit via POST/PUT/DELETE method, does not our `HttpsPolicy` redirect it from http to https if enable and then the corresponding handler method executes ? I mean thats how it should be, right ? In the implementation of HttpsPolicy, we have such an advice and deny if not possible . And policy methods do get to execute first before the handler methods . Sorry if I am bikeshedding here .

touhidurabir: Sorry, i misunderstood and mixed the `force_ssl` with `force_login_ssl` . This is correct .

pages/admin/AdminHandler.php
taslangraham: Updated to include suggested validation

taslangraham:
Hey @touhidurabir, thanks for the review. I've made updates based on your suggestions. Please take a look.

classes/user/form/ConfirmPasswordForm.php
```diff
@@ -0,0 +1,68 @@
...
+
+        parent::display($request, $template);
+    }
+
+    /**
+     * @copydoc
+     */
+    public function readInputData(): void
+    {
+        $this->readUserVars(['cancelUrl','source', 'password']);
```
touhidurabir: Here the `submitUrl` not included but in the above `display` method, it's being accessed as `$this->getData('submitUrl')` . So if I am not mistaken, in case of a faiiled validation that will re-render the from, there will be no `$submitUrl` for form action .

templates/user/confirmPassword.tpl
```diff
@@ -0,0 +1,37 @@
...
+ * Display page prompting user to confirm their password to access sensitive/restricted area of the application
+ *}
+{extends file="layouts/backend.tpl"}
+{block name="page"}
+	<h1 class="app__pageHeading text-lg-semibold text-primary">
+		{translate key="user.confirmAccess"}
+	</h1>
+	<div class="app__contentPanel">
+
+	<form class="pkp_form" id="confirmPassword" method="post" action="{$submitUrl}">
```
touhidurabir: should we include the `{include file="common/formErrors.tpl"}` to show validation errors ?

templates/user/confirmPassword.tpl
```diff
@@ -0,0 +1,37 @@
...
+		{csrf}
+		<input hidden name="cancelUrl" value="{$cancelUrl}" />
+		<input hidden name="source"  value="{$source}" />
```
touhidurabir: As this can get back as re-render from on failed validation, best to have `|escape` to avoid possible XXS .

classes/core/PKPSessionGuard.php
```diff
@@ -301,4 +307,58 @@ protected function getCookieExpirationDate(array $config): DateTimeInterface|int
...
+     * This is only applicable for site admins.
+     */
+    public function isElevatedSessionActive(): bool
+    {
+        if (!Validation::isSiteAdmin()) {
+            return false;
+        }
+
+        // If reauthentication is not required then Admin always has elevated access.
+        if(!Validation::isReauthenticationRequired()){
```
touhidurabir: we should have a space after the `if`, here and any other place .

classes/security/authorization/ReauthenticationRequiredPolicy.php
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class ReauthenticationRequiredPolicy
+ *
+ * @ingroup security_authorization
+ *
+ * @brief Policy to require to reauthenticate when accessing sensutive pages within the application.
```
touhidurabir: `sensitive`

classes/user/form/ConfirmPasswordForm.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @copydoc
+     */
+    public function display($request = null, $template = null): void
+    {
+        $templateMgr = TemplateManager::getManager($request);
+        $templateMgr->assign([
+            'pageComponent' => 'Page',
+            'cancelUrl' => $this->getData('cancelUrl'),
```
touhidurabir, line 61: question, instead of having the `cancelUrl` as part of form data, can we not just calculate it iin same way as we did in the `confirmAccess` ?

classes/security/authorization/ReauthenticationRequiredPolicy.php
```diff
@@ -0,0 +1,79 @@
...
+        return AuthorizationPolicy::AUTHORIZATION_DENY;
+    }
+
+    /**
+     * Callback when user needs to re-authenticate. Redirects user to the admin/confirmAccess page.
+     */
+    public function callOnDeny(): void
+    {
+        $dispatcher = $this->request->getDispatcher();
+        $params = ['source' => $this->request->getRequestPath()];
```
touhidurabir: I think I have stumble upon a weird but expected case during my test . So this works for GET but not for POST/PUT/DELETE etc as for in that it use GET to redirect to that page/url and failed with following response (or something similar) `{"status":false,"content":"","elementId":"0","events":[]}` . Test it with clearing the data cache which is a POST . And this is totally expected . 

During my test, following change at least does to redirect to a failed json page 
```php
$isActionRequest = $_SERVER['REQUEST_METHOD'] !== 'GET'                                                                                                         
    || $this->request->getUserVar('csrfToken');       
                                                                                                                                                                
if ($isActionRequest) {                               
    $source = $this->request->getRefererPath()
        ?? $this->request->getRequestPath();
} else {                                                                                                                                                        
    $source = $this->request->getRequestPath();
}  

$params = ['source' => $source];
```

we definitely can not make it work as POST request intended after the re-auth redirect so probably also need a short notification to let them know that they need to perform the action again . 

BTW, we have few `type="post"` instead of `POST` and better to fix those also .

classes/user/form/ConfirmPasswordForm.php
taslangraham: While this would work on the first render, subsequent requests would lose context of the correct URL. This is because `cancelUrl` is derived from the Referer header which holds the address of the page from which the current request was made.  So if we recalculate it after each request, then any re-rendering of the form triggered by an incorrect password attempt would cause it to resolve to `.../admin/confirmAccessSubmit`, since that is where the browser was when the failed request was submitted.

So the `cancelUrl` is included in the form data as to be able to carry the original Referer across multiple requests.

templates/user/confirmPassword.tpl
taslangraham: As things currently are, the form automatically displays the form errors via notification(see attached image). Including `{include file="common/formErrors.tpl"}` results in the errors being shown in two places(on the form and in the notification). Are you aware of a way to choose one place over the other for the errors to display?
image

classes/user/form/ConfirmPasswordForm.php
taslangraham: ~Since this value can be easily determined and does not need to be submitted as form data, I've updated the `ConfirmPasswordForm` class to require the `submitUrl` as a constructor argument. Therefore not needing to read as a part of the `readInputData` method.~

Moved the construction of the `submitUrl` into the `ConfirmPasswordForm` class' `display` method.

classes/security/authorization/ReauthenticationRequiredPolicy.php
taslangraham: Thanks for catching this. I like your solution. Something I'd change though is the fallback value used for `source` when `$this->request->getRefererPath()` returns `null`.  Your example uses `$this->request->getRequestPath()` as the fallback value, which would be the URL the action request was trying to hit, thus when the eventual redirect happens we would still encounter the `{"status":false,"content":"","elementId":"0","events":[]}` JSON error response.

A solution to this would be to fallback to the base page which the request belongs to. You can get that value via:  `$this->request->getRequestedPage();`

**The updated code would look something like:**

```php
        $isActionRequest = $_SERVER['REQUEST_METHOD'] !== 'GET'
            || $this->request->getUserVar('csrfToken');


        if ($isActionRequest) {
            $source = $this->request->getRefererPath()
               // If no referer then redirect to the main page of the action.
                ?? $this->request->url($this->request->getContext()?->getId(), $this->request->getRequestedPage());
        } else {
            $source = $this->request->getRequestPath();
        }

        $params = ['source' => $source];
```


**An example scenario of this in action**:

If admin is attempting to clear the template data cache, which is a `POST` request to `admin/clearTemplateCache` path, the base page would be `admin`. Therefore if no referer is present the user would be redirected back to `/admin` url instead of `admin/clearTemplateCache` which would show the JSON error.

templates/user/confirmPassword.tpl
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Display page prompting user to confirm their password to access sensitive/restricted area of the application
+ *}
+{extends file="layouts/backend.tpl"}
+{block name="page"}
+	<script type="text/javascript">
+		$(function() {ldelim}
+			$('#confirmPassword').pkpHandler('$.pkp.controllers.form.FormHandler');
+			{rdelim});
+	</script>
```
taslangraham, line 16: Helps with styling of form's html validation:

Image

taslangraham:
@touhidurabir I've made updates based on your last review. Please take a look. If everything looks good here then I'll go ahead and update the submodules on the App specific PRs.

templates/user/confirmPassword.tpl
```diff
@@ -0,0 +1,43 @@
...
+		<p class="mb-2"><span>{translate key="user.confirmAccess.description" userFullName=$currentUser->getFullName()}</span></p>
+
+		{fbvFormArea id="confirmPassword"}
+		    {fbvFormSection }
+			    {fieldLabel translate=true for=password key="user.password"}
+			    {fbvElement type="text" required=true password=true id="password" maxlength="32" size=$fbvStyles.size.MEDIUM}
+		    {/fbvFormSection}
+
+			<div>
+				<a class="pkp_button pkp_button_offset" type="button" href="{$cancelUrl}">{translate key="common.cancel"}</a>
```
touhidurabir: same as above, should add `|escape` here .

templates/user/confirmPassword.tpl
```diff
@@ -0,0 +1,43 @@
...
+	<h1 class="app__pageHeading text-lg-semibold text-primary">
+		{translate key="user.confirmAccess"}
+	</h1>
+	<div class="app__contentPanel">
+
+	<form class="pkp_form" id="confirmPassword" method="POST" action="{$submitUrl}">
+		{csrf}
+		<input hidden name="cancelUrl" value="{$cancelUrl|escape}" />
+		<input hidden name="source" value="{$source|escape}" />
+		<input hidden name="isActionRequest" value="{$isActionRequest|escape}/>
```
touhidurabir: broken HTML, missing the closing `"` for `value`

pages/admin/AdminHandler.php
```diff
@@ -786,4 +799,126 @@ protected function getAnnouncementsListPanel(string $apiUrl, PKPAnnouncementForm
...
         );
     }
+
+    /**
+     * Display the page for admin to re-authenticate and confirm access to the Administration area of the site.
+     */
+    public function confirmAccess(array $args, PKPRequest $request): void
+    {
+        $isElevatedSessionActive = Application::get()
+            ->getRequest()
```
touhidurabir: we already get the `PKPRequest $request` from method so dont think we need to do `Application::get()->getRequest()` . Here and below .

classes/user/form/ConfirmPasswordForm.php
```diff
@@ -0,0 +1,77 @@
...
+use PKP\form\validation\FormValidatorCustom;
+use PKP\form\validation\FormValidatorCSRF;
+use PKP\form\validation\FormValidatorPost;
+use PKP\security\Validation;
+
+class ConfirmPasswordForm extends Form
+{
+    /**
+     * @copydoc
+     */
```
touhidurabir, line 32: should we complete these `@copydoc`, here and below .

touhidurabir (changes requested):
@taslangraham added few more comments , once these are addressed, I think you can port the changes for other apps and ready .

taslangraham:
@touhidurabir, updated based on your last review. Thanks!

Issue #12513, opened 2026-03-31
[User invitation] | site locales needs to be consider when creating new user via invitation workflow


jardakotesovec:
### Description

Related to this one - https://github.com/pkp/pkp-lib/issues/12493

We need to make sure that the site locales are required when user details are entered, as they might differ from the context locale settings

Rule is basically that GivenName is always required in site primary locale, everything else is optional

Good to test scenario when the Site has primary locale which is not enabled on the context at all. And scenario when the context primary language for is different.

### Application Version

OJS,OMP,OPS 3.5

jardakotesovec:
@bozana @touhidurabir This is pretty much same thing as Reviewer Suggestions. Could you please review.. I know its pain.. but has to be done :-)

@Tribunal33 Up to you if you want to test after code review with testing. Its deployed as pr-5462

Tribunal33:
Not exactly sure about the steps to confirm this is working, but I tried a few different combinations.  There was one scenario which seems odd.

So I have the primary Website Locale setup as Mongo and invited a user.  Everything was smooth and shows that the forms are defaulted to primary locale of Mongo.  Given Name was entered in Mongo.  Now when I logged in as the new user and go to edit profile the given name of the context in English is not present and shows it is required.  This wasn't required before during the invitation process.  

<img width="721" height="320" alt="Image" src="https://github.com/user-attachments/assets/804d18e7-2901-433d-82cb-537ec2f367df" />

jardakotesovec:
@Tribunal33 Yeah, I think this is just coming from the unfortunate behaviour of legacy forms, where it simultaneously does not put the required language as first input and at the same time it always requires first input to have value. So it essentially requires more languages just through this behaviour. I would like to tackle that as well, but out of scope here.

jardakotesovec:
I was able to include couple of improvements in https://github.com/pkp/pkp-lib/issues/12540 to basically make it work nicely for user invitation.

- instead of requiring site locale, which I attempted here - I think better UX is just to require primary journal locale and add site locale as optional (as in legacy forms currently).  And than on server side just copy it over to site locale if its not provided by user.  


BUT the part which is still needs attention is working on migration that would address the users that has been created with invitation workflow and might be missing name in site locale as it was not required before.

jardakotesovec:
[...]

PR #12518, opened 2026-04-01
pkp/pkp-lib#12513 Require site locale for user invitations

jardakotesovec:


classes/invitation/stepTypes/AcceptInvitationStep.php
touhidurabir, line 223: this part of the code is pretty much exact in the `SendInvitationStep` . As both of them extends the abstract class InvitationStepTypes, wondering if we should just put it there ?

PR #12519, opened 2026-04-01
pkp/credit#4 update credit locale handling and translations

kaitlinnewson:


kaitlinnewson:
Hi @bozana, it looks like tests were failing as the migrations also needed to be updated for the file name changes in the credit languages files - would you mind taking a quick look at this again? Thanks!

PR #12522, opened 2026-04-02
pkp/pkp-lib#12446 Performance improvement of json data read and db insert

touhidurabir:
for pkp/pkp-lib#12446 using `fread` + `json_decode` instead of `halaxa/json-machine` . 

Benchmark is as following

#### Before (Original JsonMachine Implementation)                                                                                                                    
   
getDownloadData: 0.56s                                                                                                                                          
downloadAndExtract: 52.52s                            
processJson total: 77.17s, of which:                                                                                                                            
   - PHP record parsing (processRow): 0.68s            
   - DB batch operations: 20.22s (~0.5s per batch × 42 batches)
   - JSON streaming I/O (JsonMachine reading between records): ~56s                                                                                              
  
Peak memory: ~30-40MB (estimated, not instrumented)
Total: ~130s   
                                                                                                                                                                  
#### After (Chunked fread + Truncate-First Streaming)                                                                                                                
                                                                                                                                                                  
getDownloadData: 1.99s                                                                                                                                          
downloadAndExtract: 84.99s (network dependent, not optimizable)
processJson total: 35.59s, of which:                                                                                                                            
  - PHP record parsing (processRow): 0.18s                                                                                                                      
  - DB batch operations: 28.51s (~0.68s/batch × 42 batches)                                                                                                     
  - JSON streaming I/O (chunked fread reading between records): 6.91s                                                                                           
  
Peak memory: 48.43MB                                  
Total: 122.67s (dominated by slow Zenodo download)

#### Improvement Summary                                                                                                                                             
   
processJson:  77.17s → 35.59s  (-54%)                                                                                                                           
   - JSON I/O: 56.00s →  6.91s  (-88%)                 
   - DB ops:   20.22s → 28.51s  (same approach, run-to-run variance)
   - parsing:   0.68s →  0.18s  (-74%)                                                                                                                           
  
Peak memory: ~30-40MB → 48.43MB  (slight increase from 8MB fread chunks)                                                                                        
Data safety: preserved (upsert, no truncate)

Note: `downloadAndExtract` varies by network conditions (12-85s across runs) and is outside our control. The optimization is entirely in `processJson` — the JSON reader swap from `JsonMachine's` pure-PHP tokenizer to chunked `fread() + json_decode()` with native C functions.

### DIFF
--- a/classes/task/UpdateRorRegistryDataset.php
+++ b/classes/task/UpdateRorRegistryDataset.php
@@ -33,8 +33,6 @@ use DirectoryIterator;
 use Exception;
 use GuzzleHttp\Exception\GuzzleException;
 use Illuminate\Support\Facades\DB;
-use JsonMachine\Items;
-use JsonMachine\JsonDecoder\ExtJsonDecoder;
 use PKP\file\PrivateFileManager;
 use PKP\scheduledTask\ScheduledTask;
 use PKP\scheduledTask\ScheduledTaskHelper;
@@ -274,13 +272,13 @@ class UpdateRorRegistryDataset extends ScheduledTask
      * Uses admin.last_modified.schema_version, which reflects the file's schema format
      * (admin.created.schema_version may still reference an older version).
      *
-     * This reads only the first top-level item from the dump via JsonMachine,
+     * This reads only the first top-level item from the dump via readJsonRecords(),
      * so it stays lightweight while avoiding brittle checks against raw JSON text.
      */
     private function hasSchemaVersion(string $pathJson, int $version): bool
     {
         try {
-            foreach (Items::fromFile($pathJson, ['decoder' => new ExtJsonDecoder(assoc: true)]) as $record) {
+            foreach ($this->readJsonRecords($pathJson) as $record) {
                 $schemaVersion = $record['admin']['last_modified']['schema_version'] ?? null;
                 return (int)$schemaVersion === $version;
             }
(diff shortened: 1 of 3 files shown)


classes/task/UpdateRorRegistryDataset.php
```diff
@@ -301,15 +301,33 @@ private function hasSchemaVersion(string $pathJson, int $version): bool
...
+        // dataset, we can clear first then stream-insert both rors and settings per batch.
+        if (DB::connection()->getDriverName() === 'mysql') {
+            DB::statement('SET FOREIGN_KEY_CHECKS=0');
+            DB::table('ror_settings')->truncate();
+            DB::table('rors')->truncate();
+            DB::statement('SET FOREIGN_KEY_CHECKS=1');
+        } else {
+            // PostgreSQL: TRUNCATE with CASCADE handles FK constraints natively
+            DB::statement('TRUNCATE TABLE rors CASCADE');
+        }
```
bozana: I am not sure if we can truncate it all. If anything goes wrong we will have no data that we use and need to display to the user. Thus we would maybe need a transaction here around everything. This requires switching MySQL from TRUNCATE to DELETE FROM. However, "Transaction + DELETE + INSERT has a hidden cost on MySQL: deleting 124k+ rows inside a long-running transaction generates enormous undo log (InnoDB has  to maintain the old version for MVCC)." For a table that grows over time, this can be a problem.
So I am wondering if we should leave the DB processing as it was -- it seems the performance difference is not so big/important,  and we would have those safety risks considred? What do you think?

touhidurabir: @bozana big thanks, this is a crucial case of data not being saved at all after the truncate in case of complete failure which will leave empty ror data that I have totally missed . I know that having the `Transaction + DELETE + INSERT` is not feasible as that will accumulate a huge log and will have a major impact on other transaction and thats why i skipped that but totally did not thought about the fail case after truncate . 

In that case, I think you are right. Even if we take a bit of performance penalty, that would be a safe path to go .

PR #12529, opened 2026-04-06
pkp/pkp-lib#12493 Reviewer suggestion locale update with multilingual data and validation update

touhidurabir:
for pkp/pkp-lib#12493 .

### DIFF
--- a/api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
+++ b/api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
@@ -47,14 +47,7 @@ class AddReviewerSuggestion extends FormRequest
      */
     public function allowedLocales(): array
     {
-        $locales = Application::get()->getRequest()->getContext()->getSupportedFormLocales();
-        $sitePrimaryLocale = Application::get()->getRequest()->getSite()->getPrimaryLocale();
-
-        if (!in_array($sitePrimaryLocale, $locales)) {
-            $locales[] = $sitePrimaryLocale;
-        }
-
-        return $locales;
+        return Application::get()->getRequest()->getContext()->getSupportedFormLocales();
     }
 
     /**
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -326,6 +326,23 @@ class SettingsBuilder extends Builder
             }
         });
 
+        // Initialize missing multilingual settings props to empty array.
+        // Use array_intersect to only target props that are BOTH multilingual AND settings
+        // (excludes computed attributes like fullName, displayInitial which are not in the settings table)
+        $multilingualSettingsProps = array_intersect(
+            $this->model->getMultilingualProps(),
+            $this->model->getSettings()
+        );
+        if (!empty($multilingualSettingsProps)) {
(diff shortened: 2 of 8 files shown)


api/v1/reviewers/suggestions/formRequests/AddReviewerSuggestion.php
```diff
@@ -47,14 +47,7 @@ public function primaryLocale(): ?string
...
     {
-        $locales = Application::get()->getRequest()->getContext()->getSupportedFormLocales();
-        $sitePrimaryLocale = Application::get()->getRequest()->getSite()->getPrimaryLocale();
-
-        if (!in_array($sitePrimaryLocale, $locales)) {
-            $locales[] = $sitePrimaryLocale;
-        }
-
-        return $locales;
+        return Application::get()->getRequest()->getContext()->getSupportedFormLocales();
```
jardakotesovec, line 50: I see now, that the primary locales are merged with allowedLocales in the MultilingualInput, thats why I don't have to do it here. Its bit of implementation detail which is not intuitive, but it means reducing code here I take it..

classes/core/SettingsBuilder.php
```diff
@@ -326,6 +326,23 @@ protected function getModelWithSettings(array|string $columns = ['*']): array
...
+        $multilingualSettingsProps = array_intersect(
+            $this->model->getMultilingualProps(),
+            $this->model->getSettings()
+        );
+        if (!empty($multilingualSettingsProps)) {
+            foreach ($rows as $row) {
+                foreach ($multilingualSettingsProps as $prop) {
+                    if (!isset($row->{$prop})) {
+                        $row->{$prop} = [];
+                    }
```
jardakotesovec, line 341: @touhidurabir The idea here is just not to save empty inputs coming from the client and rather make sure we return empty object instead of null in the api for multilingual fields, right?

@asmecher @ewhanson  I will tag you here in case you have some preference, because I don't really have strong opinion, but seems like a quite significant shift how we treat the multilingual inputs/outputs in api/database. 

To provide context.

If I am for example saving contributor for submission, using multilingual fields - it sends all the fields including empty values. As result current behaviour was to save all these empty string to the database.

But now using the latest patterns - it turned out it saves only the fields which has actual value. Which I think is better behaviour as its not polluting database with empty values. And as result I was hitting the issue, where I was retrieving null value for multilingual fields, which did not get any value at all (see screenshots of request and response specifically for familyName).

I think its good idea to always return empty object from API - so there is one less variance for API consumers to worry about. So no objection for this solution - really just in case you have some suggestions or concerns here.



Image
Image

touhidurabir: > The idea here is just not to save empty inputs coming from the client and rather make sure we return empty object instead of null in the api for multilingual fields, right?

Yes . And we can do a bit the opposite also that is, if that multilingual setting props has no value, than instead of null, we can make it not available in the model instance . But I went with this approach as we are defining those as setting props so feel like they should be always available in the default model instance .

asmecher: Just to add, I agree that we should not be populating the database with empty rows (and historically we have not done that). So I support this change.

PR #12572, opened 2026-04-15
fix: categoryExistsByPath now accepts contextId to scope category path uniqueness per press

### LINKED ISSUE #12571: CategoryDAO::categoryExistsByPath() ignores $contextId parameter, causing cross-press path conflicts in OMP
Description
In OMP 3.3.0-21, https://forum.pkp.sfu.ca/t/omp-preventing-to-set-same-category-path-in-different-presses/98012/2, when trying to create a category with the same path in two different presses, the system incorrectly returns "The category path already exists" even though the path belongs to a different press/context.
Root cause
CategoryForm.inc.php calls categoryExistsByPath($path, $contextId) with two parameters, but CategoryDAO::categoryExistsByPath() only accepts $path and silently discards $contextId, so the query checks all presses instead of just the current one.
Fix
Add $contextId parameter to CategoryDAO::categoryExistsByPath() and include it in the SQL query (AND context_id = ?).
Already reported in forum: [link al post che hai condiviso]

paolorafa:
https://github.com/pkp/pkp-lib/issues/12571

Problem
When trying to create a category with the same path in two different presses, OMP incorrectly returns "The category path already exists" even though the path belongs to a different press.
Root cause
CategoryForm.inc.php calls categoryExistsByPath($path, $contextId) with two parameters, but CategoryDAO::categoryExistsByPath() only accepted $path and silently discarded $contextId, so the SQL query checked across all presses instead of just the current one.
Fix
Added $contextId parameter to CategoryDAO::categoryExistsByPath() and included it in the SQL query when provided.
Related issue: [link alla tua issue]
Tested on: OMP 3.3.0-21

### DIFF
--- a/classes/context/CategoryDAO.inc.php
+++ b/classes/context/CategoryDAO.inc.php
@@ -108,13 +108,19 @@ class CategoryDAO extends DAO {
 	 * @param $path the path for the category
 	 * @return boolean
 	 */
-	function categoryExistsByPath($path) {
-		$result = $this->retrieve(
-			'SELECT COUNT(*) AS row_count FROM categories WHERE path = ?', [$path]
-		);
-		$row = $result->current();
-		return $row ? (boolean) $row->row_count : false;
-	}
+	function categoryExistsByPath($path, $contextId = null) {
+    if ($contextId) {
+        $result = $this->retrieve(
+            'SELECT COUNT(*) AS row_count FROM categories WHERE path = ? AND context_id = ?', [$path, $contextId]
+        );
+    } else {
+        $result = $this->retrieve(
+            'SELECT COUNT(*) AS row_count FROM categories WHERE path = ?', [$path]
+        );
+    }
+    $row = $result->current();
+    return $row ? (boolean) $row->row_count : false;
+}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/context/CategoryDAO.inc.php
```diff
@@ -108,13 +108,19 @@ public function getByPublicationId($publicationId) {
...
 	 */
-	function categoryExistsByPath($path) {
-		$result = $this->retrieve(
-			'SELECT COUNT(*) AS row_count FROM categories WHERE path = ?', [$path]
-		);
-		$row = $result->current();
-		return $row ? (boolean) $row->row_count : false;
-	}
+	function categoryExistsByPath($path, $contextId = null) {
+    if ($contextId) {
```
kaitlinnewson, line 112: An additional indent is needed here (for all lines down to line 123)

classes/context/CategoryDAO.inc.php
```diff
@@ -108,13 +108,19 @@ public function getByPublicationId($publicationId) {
 	 * @param $path the path for the category
```
kaitlinnewson, line 108:
```suggestion
	 * @param $path the path for the category
	 * @param ?int $contextId
```

PR #12586, opened 2026-04-21
pkp/pkp-lib#12547 Session/cookie management update

touhidurabir:
for pkp/pkp-lib#12547 .


classes/core/PKPRequest.php
```diff
@@ -430,7 +430,7 @@ public function isGet(): bool
      */
     public function checkCSRF(): bool
     {
-        return $this->getUserVar('csrfToken') == $this->getSession()->token();
+        return $this->getUserVar('csrfToken') === $this->getSession()->token();
```
touhidurabir, line 433: add safety for PHP type juggling with == can bypass CSRF validation (e.g., 0 == "0e123..." evaluates to true)

classes/core/PKPSessionServiceProvider.php
```diff
@@ -28,14 +28,24 @@ class PKPSessionServiceProvider extends \Illuminate\Session\SessionServiceProvid
...
     public function boot()
     {
-        register_shutdown_function(function () {
+        $currentWorkingDir = getcwd();
+
+        register_shutdown_function(function () use ($currentWorkingDir) {
+
+            // Restore the current working directory
+            // @see https://www.php.net/manual/en/function.register-shutdown-function.php#refsect1-function.register-shutdown-function-notes
+            chdir($currentWorkingDir);
```
touhidurabir, line 37: missing current working dir restore .

classes/core/PKPSessionServiceProvider.php
```diff
@@ -28,14 +28,24 @@ class PKPSessionServiceProvider extends \Illuminate\Session\SessionServiceProvid
...
                 return;
             }
 
             // need to make sure that all changes to session(via pull/put) are reflected in session storage
-            Application::get()->getRequest()->getSessionGuard()->getSession()->save();
+            try {
+                Application::get()->getRequest()->getSessionGuard()->getSession()->save();
+            } catch (\Throwable $e) {
+                error_log('Session save failed: ' . $e->getMessage());
+            }
```
touhidurabir, line 48: a safety first approach with error logging if for some unforseen reason, it can not write to session driver .

classes/middleware/PKPAuthenticateSession.php
```diff
@@ -9,17 +9,42 @@
...
+        $response = parent::handle($request, $next);
+
+        if ($request->hasSession()
+            && $request->user()
+            && Config::getVar('security', 'session_check_ip')
+            && ($loginIp = $request->session()->get('login_ip'))
+            && $loginIp !== $request->ip()
+        ) {
+            $this->logout($request);
+        }
```
touhidurabir, line 43: handle the missing config check of `security.session_check_ip` , a regression from https://github.com/pkp/pkp-lib/issues/9566

classes/security/Validation.php
```diff
@@ -136,8 +136,13 @@ public static function logout()
...
 
-        $session->put('username', $user->getUsername());
-        $session->put('email', $user->getEmail());
+        // Store last login identifier in a cookie for login page auto-fill
+        $sessionLifetime = max(1, (int) Config::getVar('general', 'session_lifetime', 30));
+        $request->setCookieVar(
+            'last_login_auth',
+            $user->getEmail() ?? $user->getUsername(),
+            time() + ($sessionLifetime * 24 * 60 * 60)
+        );
```
touhidurabir, line 145: instead of generating a session row after logout just to auto fill PII, use cookie to hold that information .

pages/login/LoginHandler.php
```diff
@@ -71,10 +71,9 @@ public function index($args, $request)
         $templateMgr = TemplateManager::getManager($request);
         $templateMgr->assign([
             'loginMessage' => $request->getUserVar('loginMessage'),
-            'username' => $request->getSession()->get('email') ?? $request->getSession()->get('username'),
+            'username' => $request->getCookieVar('last_login_auth') ?? '',
             'remember' => $request->getUserVar('remember'),
             'source' => $request->getUserVar('source'),
-            'showRemember' => Config::getVar('general', 'session_lifetime') > 0,
```
touhidurabir, line 77: it seems we never use this `showRemember` in the template to control the `Keep me logged in` option's visibility . And i think this is ok as that option should be always visible and up to the user to determined if they want to keep logged in or not .

classes/user/form/RegistrationForm.php
```diff
@@ -314,9 +314,6 @@ public function execute(...$functionArgs)
             return false;
         }
 
-        $request->getSession()->put('username', $user->getUsername());
-        $request->getSessionGuard()->updateSession($user->getId());
```
touhidurabir, line 318: It seems this session set/update is redundant as for the 2 following cases after a successful registration as 
1. auto login --> this calls the Validation::login which in return calls the `Auth::attempt` which in return, if a successful login happens, it internally update the session and set session data via Validation::registerUserSession
2. if the `email.require_validation` from the config set, it does not login and need to validate the email which then leads to complete the login process which handle session set/update mechanism as no 1 .

classes/middleware/PKPEncryptCookies.php
```diff
@@ -50,28 +49,11 @@ public function handle($request, Closure $next)
...
-                        $cookie->getName(),
-                        $this->encrypter->getKey()
-                    ) . $cookie->getValue(),
-                    static::serialized($cookie->getName())
-                )
-            ));
-        }
-
-        return $response;
+        return parent::encrypt($response);
```
touhidurabir, line 57: we only need the visibility widening to handle call from `PKPSessionGuard::updateSessionCookieToResponse()` as `$pkpEncryptCookies->encrypt($response)` . So just a simple forward call to parent works but need a visibility  widening from `protected` to `public` .

classes/core/PKPContainer.php
```diff
@@ -480,14 +480,18 @@ protected function loadConfiguration(): void
         ];
 
         // Session manager
+        // session_lifetime is in days; clamp to minimum 1 day to avoid Laravel treating 0 as "immediately expired"
+        // For browser-close session expiration, use session_expire_on_close = On in config.inc.php
+        $sessionLifetime = max(1, (int) Config::getVar('general', 'session_lifetime', 30));
```
touhidurabir: setting a minimal session lifetime to `1` day to mitigate the poor `session_lifetime` to `0`  . 

Also should we allow to set less than a day . Like may be 12/6/2 hours ? in that case perhaps we should have max limit `.084` which leads to set the session max idel lifetime to 120 mins which is also laravel default .

classes/security/Validation.php
touhidurabir: reverted as not necessary .

classes/core/PKPContainer.php
taslangraham: I like the idea of being able to set this to less than a day. This would be useful for installs that want to be a bit stricter with how long a session can be left idle.

taslangraham (approved):
Looks good @touhidurabir.  I just had one comment in response to something you proposed. Thanks!

classes/core/PKPContainer.php
```diff
@@ -470,6 +475,7 @@ protected function loadConfiguration(): void
                 'web' => [
                     'driver' => 'session',
                     'provider' => 'users',
+                    'remember' => $rememberMeMinutes, // remember-me cookie lifetime in minutes
```
touhidurabir, line 478: default laravel remember me lift time is 400 days which is too big for most cases . and previous use of `session_lifetime` to control it is not possible as that moved to manage the max ideal session life time . So we need a config to control the remember me life time as well .

classes/core/PKPSessionServiceProvider.php
touhidurabir: reverted as per Alec's suggestion .

Issue #12509, opened 2026-03-30
Add Crossref support for peer review DOIs


ewhanson:
**Description:**
With the addition of internal support for peer review DOIs, support for including these in Crossref metadata exports and deposits is necessary.

It will need to be added only for public peer reviews with proper reviewer identity anonymity taken into consideration (which may already be taken care of).

Required/optional data to include can be confirmed in the Crossref metadata deposit schema [documentation](https://data.crossref.org/reports/help/schema_doc/5.3.1/index.html).

**Steps to test the changes:**
1. Take a submission through the editorial process and produce a published article with at least one public peer review.
2. Ensure peer review DOIs are enabled.
3. Export Crossref article metadata (meant for external deposit).
4. Confirm XML validity with Crossref XML metadata quality checking [tool](https://www.crossref.org/02publishers/parser.html).

**What application are you using?**
  OJS 3.6

**Additional information**
N/A

ewhanson:
@bozana, I've linked to the whole metadata schema documentation above, but if you have any specific things you know need to be included for Crossref deposits for peer reviews, could you note them above? Wondering if there's anything you know of that might not be obvious. Thanks!

taslangraham:
Hey @ewhanson @bozana

This issue description suggests that DOIs for peer review should be included in the same XML as the Crossref article metadata export. However, after reviewing the Crossref schema [documentation](https://data.crossref.org/reports/help/schema_doc/5.3.1/schema_5_3_1.html#body), it appears that a single XML export cannot contain deposits for more than one content type.

From the schema documentation:

> While it is possible to include records for multiple journals, books, conferences, or other types of content in a single submission, it is not possible to mix content types within a single DOI submission.

This is also reflected in the definition of the body element:
```                            
<xsd:element name="body">
  <xsd:annotation>
    <xsd:documentation>Container for the main body of a DOI record submission. While it is possible to include records for multiple journals, books, conferences, or other types of content in a single submission, it is not possible to mix content types within a single DOI submission.</xsd:documentation>
  </xsd:annotation>
  <xsd:complexType>
    <xsd:choice>
      <xsd:element ref="journal" maxOccurs="unbounded"/>
      <xsd:element ref="book" maxOccurs="unbounded"/>
      <xsd:element ref="conference" maxOccurs="unbounded"/>
      <xsd:element ref="sa_component" maxOccurs="unbounded"/>
      <xsd:element ref="dissertation" maxOccurs="unbounded"/>
[...]

Issue #11332, opened 2025-04-28
Add support for review-related DOIs


ewhanson:
[To be completed]

ewhanson:
Hey @bozana, would you be able to have a first look at this for OJS? Thanks!

PR #12605, opened 2026-04-23
pkp/pkp-lib#12509 Add Crossref support for peer review DOIs

taslangraham:
For pkp/pkp-lib#12509

### DIFF
--- a/api/v1/dois/PKPDoiController.php
+++ b/api/v1/dois/PKPDoiController.php
@@ -451,16 +451,6 @@ class PKPDoiController extends PKPBaseController
 
         // Invoke IDoiRegistrationAgency::exportSubmissions
         $submissionsData = $agency->exportSubmissions($submissions, $context);
-
-        $exportablePeerReviewIds = Repo::reviewAssignment()
-        ->getExportableDOIsPeerReviewIds($context->getId(), array_map(fn (Submission $submission)=>$submission->getId(), $submissions));
-
-        if (!empty($peerReview['xmlErrors'])) {
-            return response()->json([
-                'error' => __('api.dois.400.xmlExportFailed')
-            ], Response::HTTP_BAD_REQUEST);
-        }
-
         $temporaryFileIds = [];
 
         if (!empty($submissionsData)) {
@@ -471,6 +461,9 @@ class PKPDoiController extends PKPBaseController
             in_array(Repo::doi()::TYPE_PEER_REVIEW, $agency->getAllowedDoiTypes()) &&
             in_array(Repo::doi()::TYPE_PEER_REVIEW, $context->getData(Context::SETTING_ENABLED_DOI_TYPES))
         ) {
+            $exportablePeerReviewIds = Repo::reviewAssignment()
+                ->getExportableDOIsPeerReviewIds($context->getId(), array_map(fn (Submission $submission)=>$submission->getId(), $submissions));
+
             $peerReviewsData = $this->getPeerReviewExports(
                 array_map(fn($exportablePeerReviewId) => Repo::reviewAssignment()->get($exportablePeerReviewId), $exportablePeerReviewIds),
                 $context,
--- a/api/v1/dois/PKPDoiController.php
+++ b/api/v1/dois/PKPDoiController.php
@@ -419,9 +419,6 @@ class PKPDoiController extends PKPBaseController
 
         $validIds = Repo::publication()
             ->getExportableDOIsSubmissionIds($context->getId(), $context->getData(Context::SETTING_DOI_VERSIONING));
-
-        // Need a new method to get exportable publication IDs too?
-
         $invalidIds = array_diff($requestIds, $validIds);
         if (count($invalidIds)) {
             return response()->json([
@@ -451,6 +448,13 @@ class PKPDoiController extends PKPBaseController
 
         // Invoke IDoiRegistrationAgency::exportSubmissions
         $submissionsData = $agency->exportSubmissions($submissions, $context);
+
+        if (!empty($submissionsData['xmlErrors'])) {
+            return response()->json([
+                'error' => __('api.dois.400.xmlExportFailed')
+            ], Response::HTTP_BAD_REQUEST);
+        }
+
         $temporaryFileIds = [];
 
         if (!empty($submissionsData)) {
@@ -461,14 +465,14 @@ class PKPDoiController extends PKPBaseController
             in_array(Repo::doi()::TYPE_PEER_REVIEW, $agency->getAllowedDoiTypes()) &&
             in_array(Repo::doi()::TYPE_PEER_REVIEW, $context->getData(Context::SETTING_ENABLED_DOI_TYPES))
         ) {
-            $exportablePeerReviewIds = Repo::reviewAssignment()
-                ->getExportableDOIsPeerReviewIds($context->getId(), array_map(fn (Submission $submission)=>$submission->getId(), $submissions));
+            $submissionIds = array_map(fn(Submission $submission) => $submission->getId(), $submissions);
+            $peerReviewsData = $this->getPeerReviewExports($submissionIds, $context, $agency);
 
-            $peerReviewsData = $this->getPeerReviewExports(
-                array_map(fn($exportablePeerReviewId) => Repo::reviewAssignment()->get($exportablePeerReviewId), $exportablePeerReviewIds),
-                $context,
-                $agency
-            );
+            if (!empty($peerReviewsData['xmlErrors'])) {
+                return response()->json([
+                    'error' => __('api.dois.400.xmlExportFailed')
+                ], Response::HTTP_BAD_REQUEST);
+            }
 
             if (!empty($peerReviewsData)) {
                 $temporaryFileIds[] = $peerReviewsData['temporaryFileId'];
@@ -483,15 +487,15 @@ class PKPDoiController extends PKPBaseController
 
 
     /**
-     * Prepare the Peer Review XML export for a list of submission IDs.
-     * This method should be implemented at the app level if peer review exports are applicable.
-     * @param ReviewAssignment[] $peerReviews - An array of submission IDs for which the associated Peer Review's XML export is being prepared.
+     * Prepare the associated Peer Review DOI XML export for a list of submission IDs.
+     * This method should be implemented at the app level if peer review DOIs are supported.
+     * @param array $submissionIds - ID of the submissions for which the associated Peer Review DOI XML export should be generated.
      * @param Context $context - The context in which the export is being performed.
      * @param IDoiRegistrationAgency $agency - The DOI registration agency responsible for the export.
-     * @return array{ temporaryFileId: int|string,xmlErrors: array} - An array containing the temporary file ID and any XML errors encountered during export.
+     * @return array{ temporaryFileId: int|string, xmlErrors: array } - An array containing the temporary file ID and any XML errors encountered during export.
      *
      */
-    public function getPeerReviewExports(array $peerReviews, Context $context, IDoiRegistrationAgency $agency): array
+    public function getPeerReviewExports(array $submissionIds, Context $context, IDoiRegistrationAgency $agency): array
     {
         return [];
     }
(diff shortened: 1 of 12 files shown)


jobs/doi/DepositPeerReview.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->reviewId = $reviewId;
+        $this->context = $context;
+        $this->agency = $agency;
+        $this->submissionId = $submissionId;
+    }
+
+    public function handle(): void
+    {
+        $depositablePeerReviewIds = Repo::reviewAssignment()
+            ->getExportableDOIsPeerReviewIds($this->context->getId(), $this->context->getData(Context::SETTING_DOI_VERSIONING), [$this->submissionId]);
```
bozana: Should the last parameter here rather be: 
`$this->submissionId !== null ? [$this->submissionId] : null   `
?
Or rather, as said above, to always have a submissionId?

jobs/doi/DepositPeerReview.php
```diff
@@ -0,0 +1,72 @@
...
+use APP\facades\Repo;
+use APP\plugins\IDoiRegistrationAgency;
+use PKP\context\Context;
+use PKP\job\exceptions\JobException;
+use PKP\jobs\BaseJob;
+
+class DepositPeerReview extends BaseJob
+{
+    protected int $reviewId;
+    protected ?int $submissionId;
```
bozana: Should we maybe always 'require' the submissionId? -- A peer review will always have a submission... :thinking:

classes/publication/Repository.php
```diff
@@ -1176,10 +1175,10 @@ public function getReviewDoiItemsGroupedByPublication(array $publicationIds): ar
         $result = array_fill_keys($publicationIds, []);
 
         // ReviewAssignments with DOIs
+        /** @var ReviewAssignment[] $assignments */
         $assignments = Repo::reviewAssignment()
             ->getCollector()
             ->filterByReviewRoundIds($roundIds)
-            ->filterByCompleted(true)
```
bozana, line 1200: Why do we not filter by completed any more?

api/v1/dois/PKPDoiController.php
```diff
@@ -445,19 +447,59 @@ public function exportSubmissions(Request $illuminateRequest): JsonResponse
...
+        }
+
+
         return response()->json([
-            'temporaryFileId' => $responseData['temporaryFileId']
+            'temporaryFileIds' => $temporaryFileIds
         ], Response::HTTP_OK);
     }
 
+
```
bozana: Double blank line

jobs/doi/DepositSubmission.php
```diff
@@ -63,6 +63,23 @@ public function handle()
             throw new JobException(JobException::INVALID_PAYLOAD);
         }
 
-        $this->agency->depositSubmissions([$submission], $this->context);
+        $submissionDepositResults = $this->agency->depositSubmissions([$submission], $this->context);
+
+        if($submissionDepositResults['hasErrors']) {
```
bozana: empty space between `if` and `(`

jobs/doi/DepositSubmission.php
```diff
@@ -63,6 +63,23 @@ public function handle()
...
+            throw new JobException($submissionDepositResults['responseMessage']);
```
bozana: I think there is no need to throw the JobException -- it pollutes the failed jobs table with what is actually an agency-level response, while the agency-level errors are actually saved for the failed DOIs.
So maybe to add this check to the others below:
```
if (
      !$submissionDepositResults['hasErrors'] &&
      in_array(Repo::doi()::TYPE_PEER_REVIEW, $this->agency->getAllowedDoiTypes()) &&
      in_array(Repo::doi()::TYPE_PEER_REVIEW, $this->context->getData(Context::SETTING_ENABLED_DOI_TYPES))
  )
```

jobs/doi/DepositPeerReview.php
```diff
@@ -0,0 +1,72 @@
...
+        $peerReview = Repo::reviewAssignment()->get($this->reviewId);
+        if (!$peerReview) {
+            throw new JobException(JobException::INVALID_PAYLOAD);
+        }
+
+        $result = $this->agency->depositPeerReviews([$peerReview], $this->context);
+
+        if ($result['hasErrors']) {
+            throw new JobException($result['responseMessage']);
+        }
```
bozana: Here as well, I think no need to throw JobException -- this is handled by the agency

api/v1/dois/PKPDoiController.php
```diff
@@ -445,19 +447,59 @@ public function exportSubmissions(Request $illuminateRequest): JsonResponse
...
+     * @param array $submissionIds - ID of the submissions for which the associated Peer Review DOI XML export should be generated.
+     * @param Context $context - The context in which the export is being performed.
+     * @param IDoiRegistrationAgency $agency - The DOI registration agency responsible for the export.
+     * @return array{ temporaryFileId: int|string, xmlErrors: array } - An array containing the temporary file ID and any XML errors encountered during export.
+     *
+     */
+    public function getPeerReviewExports(array $submissionIds, Context $context, IDoiRegistrationAgency $agency): array
+    {
+        return [];
+    }
```
bozana, line 510: Maybe to replace this with the actual implementation currently sitting in the OJS DoiController override and the OJS DoiController override can then be deleted. This works for all apps without any override needed -- agencies that don't support peer review DOIs never reach this method (gated by the getAllowedDoiTypes() check above), and when the OMP Crossref plugin is implemented it will benefit from this automatically.
What do you think?

classes/submission/reviewAssignment/Repository.php
```diff
@@ -289,4 +289,13 @@ public function getExternalReviewerIdsByCompletedYear(int $contextId, string $ye
     {
         return $this->dao->getExternalReviewerIdsByCompletedYear($contextId, $year);
     }
+
+    /** @copydoc DAO::getExportableDOIsPeerReviewIds()
```
bozana: maybe copydoc into the next line?

bozana (commented):
Hi @taslangraham, here are my comments...

api/v1/dois/PKPDoiController.php
taslangraham: Good idea. Thanks.

classes/publication/Repository.php
taslangraham: The initial reason was tha the [`filterByCompleted` ](https://github.com/pkp/pkp-lib/blob/84fc784da306709fa6ea762d4f5df11bfc2890d8/classes/submission/reviewAssignment/Collector.php#L197)filter would return completed reviews, but only those for submissions not published. Which means that if a submission was published and has reviews with DOIs, DOI info for those reviews are not returned by the [schema](https://github.com/pkp/pkp-lib/blob/84fc784da306709fa6ea762d4f5df11bfc2890d8/classes/publication/maps/Schema.php#L181-L193) data and as such won't be shown in the DOI management UI.

taslangraham: @ewhanson The original([#11332](https://github.com/pkp/pkp-lib/issues/11332)) [logic](https://github.com/pkp/pkp-lib/blob/84fc784da306709fa6ea762d4f5df11bfc2890d8/classes/publication/Repository.php#L1167-L1222) to prepare DOI for reviews was done by you; maybe you could confirm if it is intentional and correct to only make review DOIs visible if the submission is not published.

jobs/doi/DepositPeerReview.php
taslangraham: Updated to have `submissionId` be a required argument.

jobs/doi/DepositPeerReview.php
```diff
@@ -0,0 +1,68 @@
...
+use APP\plugins\IDoiRegistrationAgency;
+use PKP\context\Context;
+use PKP\job\exceptions\JobException;
+use PKP\jobs\BaseJob;
+
+class DepositPeerReview extends BaseJob
+{
+    protected int $reviewId;
+    protected int $submissionId;
+    protected Context $context;
```
touhidurabir: @taslangraham just one suggestion , lets not pass the context object but the id only to keep the serialise data small . and we can always generate the context object in the job later. we do have few cases where we passed the context but lately trying to keep the serialise data as small as possible.

classes/publication/Repository.php
taslangraham: I've spoken with Erik, and the `filterByCompleted` filter should be replaced here with the recently introduced [`filterByIsConfirmedByEditor`](https://github.com/pkp/pkp-lib/blob/3d7499e99a1f5644d912114dffe9c0b85d538fba/classes/submission/reviewAssignment/Collector.php#L190-L194).

jobs/doi/DepositPeerReview.php
```diff
@@ -0,0 +1,73 @@
...
+ * @ingroup jobs
+ *
+ * @brief Job to deposit peer review DOI and metadata to the configured registration agency
+ */
+
+namespace PKP\jobs\doi;
+
+use APP\core\Application;
+use APP\facades\Repo;
+use APP\journal\Journal;
```
bozana: not used, can be removed

api/v1/dois/PKPDoiController.php
```diff
@@ -45,6 +46,7 @@
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
 use PKP\services\PKPSchemaService;
+use PKP\submission\reviewAssignment\ReviewAssignment;
```
bozana: not used, can be removed

jobs/doi/DepositSubmission.php
```diff
@@ -63,6 +63,20 @@ public function handle()
             throw new JobException(JobException::INVALID_PAYLOAD);
         }
 
-        $this->agency->depositSubmissions([$submission], $this->context);
+        $submissionDepositResults = $this->agency->depositSubmissions([$submission], $this->context);
+
+        // Deposit Submission's associated Peer Review if Peer Review DOIs are enabled in Context and Agency, and the Submission was successfully desposited
```
bozana: typo: deposited instead of desposited

api/v1/dois/PKPDoiController.php
```diff
@@ -437,6 +438,7 @@ public function exportSubmissions(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_NOT_FOUND);
         }
 
+
```
bozana: the empty line can be removed

classes/submission/reviewRound/ReviewRoundDAO.php
```diff
@@ -243,6 +243,16 @@ public function getByPublicationIds($publicationIds, $stageId = null, $round = n
...
+    /**
+     * Get an iterator of review round objects associated with a set of review round IDs.
+     */
+    public function getByReviewRoundIds(array $reviewRoundIds): DAOResultFactory
+    {
+        $q = DB::table('review_rounds', 'rr')
+            ->whereIn('rr.review_round_id', $reviewRoundIds);
+
+        return new DAOResultFactory($q->get(), $this, '_fromRow', [], $q);
+    }
```
bozana, line 255: a blank line could be added after this function

bozana (commented):
Hi @taslangraham, a few minor comments

api/v1/dois/PKPDoiController.php
```diff
@@ -445,19 +445,66 @@ public function exportSubmissions(Request $illuminateRequest): JsonResponse
...
+     * @param IDoiRegistrationAgency $agency - The DOI registration agency responsible for the export.
+     *
+     * @return array{ temporaryFileId: int|string, xmlErrors: array } - An array containing the temporary file ID and any XML errors encountered during export.
+     *
+     */
+    private function getPeerReviewExports(array $submissionIds, Context $context, IDoiRegistrationAgency $agency): array
+    {
+        $exportablePeerReviewIds = Repo::reviewAssignment()
+            ->getExportableDOIsPeerReviewIds($context->getId(), $context->getData(Context::SETTING_DOI_VERSIONING), $submissionIds);
+
```
bozana, line 499: I think we need to add something like this here:
```
if (empty($exportablePeerReviewIds)) {
      return [];
  }
```
Else, it will call agency's exportPeerReviews and then also create an invalid XML file -- with header but without the body peer_review, when there is no exportable review for a submission.

bozana (commented):
Hi @taslangraham, I found one problem -- if there is no exportable review for a submission, an invalid XML will be however created -- see the comment here.

Issue #12584, opened 2026-04-21
Add "Summary of Changes/Amendment" field on file revision upload


blesildaramirez:
**Description:**
- Add a field for "Summary of Changes/Amendment" when uploading a new file revision.

**Steps to test the changes:**
1. Login as an author/editor
2. Click on a submission in review stage
3. On the Revisions Uploaded section, click "Upload"
4. Fill up Upload File step (1) as necessary
5. On the Review Details step (2), there should be optional "Summary of Changes" text-area 
6. Continue to step 3

**What application are you using?**
  OJS, OMP or OPS version 3.6 (main)

**Additional information**
<img width="833" height="741" alt="Image" src="https://github.com/user-attachments/assets/0437ed9d-2067-44b5-aa01-6c84bb890349" />

bozana:
Here the Crossref vocabulary for update types: addendum, clarification, correction, corrigendum, erratum, expression_of_concern, new_edition, new_version, partial_retraction, removal, retraction, withdrawal.
Crossref itself uses the same vocabulary regardless of content type.

Primarily journal-specific (OJS):
- corrigendum — formal author correction, journal publishing term
- erratum — formal publisher correction, journal publishing term

Primarily book-specific (OMP):
- new_edition — editions are a book concept; journals don't have editions, they have versions

Primarily preprint-specific (OPS):
- withdrawal — preprints are "withdrawn" before formal publication; published articles are "retracted"

Universal (all three):
- addendum, clarification, correction, expression_of_concern, new_version, partial_retraction, removal, retraction


I am not sure if we should then separate the few per application, or rather let editor choose the right term... 🤔  -- sometimes users use our applications for different purposes, so maybe we can leave the list the same in all applications and trust the editors that they know what to chose? 
Also, the list would not be that much reduced, by max 2-3 elements, so this does not make much difference.

jardakotesovec:
@bozana Hi Bozana, could you also skim through the pkp-lib/ojs. Especially on the more backendy things. There is some magic to blend the vue.js form with the legacy Upload File dialog - you don't have to pay attention to that :-)

Blessie omitted addition of that for OMP/OPS to ship this quickly. Just advised Blessie to keep the migration / schema changes in the lib/pkp. So we just address the UI bits separately.

blesildaramirez:
Hi @bozana , thanks for the detailed explanation for each update type values.
[...]

PR #12622, opened 2026-04-29
pkp/pkp-lib#12584 Add "Summary of Changes/Amendment" field on file revision upload and update publication issues page

blesildaramirez:


templates/controllers/wizard/fileUpload/form/submissionFileMetadataForm.tpl
```diff
@@ -1,76 +1,64 @@
...
-		{fbvFormSection for="name" size=$fbvStyles.size.LARGE title="submission.form.name" required=true}
-			{fbvElement type="text" id="name" value=$submissionFile->getData('name') multilingual=true maxlength="255" required=true}
-		{/fbvFormSection}
+	<script type="text/javascript">
+		(function() {ldelim}
+			// FieldRichTextarea looks for the skin URL on pkp.tinyMCE.
+			// The main app sets it as $root state; our standalone mount
+			// doesn't see that, so populate the global fallback.
+			if (!pkp.tinyMCE) {ldelim}
+				pkp.tinyMCE = {ldelim}skinUrl: {$tinyMceSkinUrl|json_encode}{rdelim};
```
jardakotesovec: I added pkp.tinyMCE to be able to pass it to tinyMCE in storybook. Relying on Page as root component is not great solution going forward - its very unusual pattern to actually expose data to children.  I would suggest to add it to PkpTemplateManager - so its always available in pkp.tinyMCE so its universally available across all these scenarios.

schemas/publication.json
```diff
@@ -399,6 +399,14 @@
 				"no_new_line"
 			]
 		},
+		"summaryOfChanges": {
```
jardakotesovec, line 402: Should not we also have the updateType added in schema?

jardakotesovec (changes requested):
Just small things. Its mainly just to move the migration/schema change to lib/pkp and the adjustment for tinyMCE I think is also worthwhile..

templates/controllers/wizard/fileUpload/form/submissionFileMetadataForm.tpl
blesildaramirez: Noted on this, I moved the `pkp.tinyMCE` declaration to `PkpTemplateManager` so it's globally accessible. Thanks!

schemas/publication.json
blesildaramirez: The `updateType` has been added to this schema after moving the migration from ojs to pkp-lib.

blesildaramirez:
Hi @jardakotesovec , thanks for the review. I've addressed all your feedback, kindly check. Thanks!

controllers/wizard/fileUpload/form/SubmissionFilesMetadataForm.php
```diff
@@ -162,14 +171,29 @@ public function fetch($request, $template = null, $display = false)
...
             'stageId' => $this->getStageId(),
             'reviewRoundId' => $reviewRound ? $reviewRound->getId() : null,
-            'supportsDependentFiles' => Repo::submissionFile()->supportsDependentFiles($this->getSubmissionFile()),
+            'supportsDependentFiles' => Repo::submissionFile()->supportsDependentFiles($submissionFile),
             'genre' => $genre,
+            'metadataMountConfig' => [
+                'submissionFile' => $submissionFile->_data,
+                'genreCategory' => (int) $genre?->getCategory(), // will be removed once genreMetadataType is implemented by Media Files api changes
+                'supportedLocales' => $supportedLocales,
+                'primaryLocale' => $this->getDefaultFormLocale(),
```
jardakotesovec, line 193: I think here we have same problem with locales as in the UI. It should follow submission.locale as primary and something like this for supportedLocales (taking it from the submission/maps/Schema.php for metadataLocales):



collect($this->context->getSupportedSubmissionMetadataLocaleNames() + $submission->getPublicationLanguageNames())
                        ->sortKeys()
                        ->toArray();

pages/dashboard/PKPDashboardHandler.php
```diff
@@ -279,6 +279,11 @@ public function index($args, $request)
             'SUBMISSION_FILE_JATS' => SubmissionFile::SUBMISSION_FILE_JATS,
             'SUBMISSION_FILE_DEPENDENT' => SubmissionFile::SUBMISSION_FILE_DEPENDENT,
             'SUBMISSION_FILE_BODY_TEXT' => SubmissionFile::SUBMISSION_FILE_BODY_TEXT,
+
+            'GENRE_CATEGORY_DOCUMENT' => Genre::GENRE_CATEGORY_DOCUMENT,
+            'GENRE_CATEGORY_ARTWORK' => Genre::GENRE_CATEGORY_ARTWORK,
+            'GENRE_CATEGORY_SUPPLEMENTARY' => Genre::GENRE_CATEGORY_SUPPLEMENTARY,
+
```
bozana, line 286: These already exist, see lines 337–340.

pages/dashboard/PKPDashboardHandler.php
bozana: trailing space here

jardakotesovec (changes requested):
Couple adjustments needed, truly important part is the locale selection.

classes/migration/upgrade/v3_6_0/I12584_AddPublicationUpdateType.php
```diff
@@ -0,0 +1,38 @@
...
+use PKP\migration\Migration;
+use PKP\publication\enums\UpdateType;
+
+class I12584_AddPublicationUpdateType extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('publications', function (Blueprint $table) {
+            $table->enum('update_type', array_column(UpdateType::cases(), 'value'))
+                ->nullable();
```
bozana: We have just agreed (on Mattermost channel) that there should be no null values in the DB: Null values are currently not possible to add via UI, but via API.
Thus, we can remove `->nullable()` here and use `->default(UpdateType::NEW_VERSION->value)`.
This also ensures that already existing entries will have new_version per default -- the database engine fills that default into all existing rows immediately when the column is added.

schemas/publication.json
```diff
@@ -445,6 +453,15 @@
 				"nullable"
 			]
 		},
+		"updateType": {
+			"type": "string",
+			"description": "The type of update represented by this publication version (e.g. addendum, correction, retraction). One of the UpdateType enumeration values.",
+			"apiSummary": true,
+			"validation": [
+				"nullable",
```
bozana: We would need to remove `nullable` here, because we agreed not to allow null values.

Because this field is not marked as required:
Maybe to note, that In Laravel/PKP validation, without nullable and without required:
  - Field absent from request → no validation triggered, field not touched; on INSERT the DB default new_version is used, on UPDATE the existing value is preserved ✓
  - Field explicitly null → fails the below in: rule → validation error ✓
  - Field set to a valid value → passes ✓

bozana (commented):
Hi @blesildaramirez,
I left just a few commnents, above all how to ensure the DB values for the new DB table column are never null. The default value for DB also solves the migration -- the existing entries will get new_version.
Thanks a lot!

controllers/wizard/fileUpload/form/SubmissionFilesMetadataForm.php
blesildaramirez: Thank you, this is working fine per my testing.

schemas/publication.json
blesildaramirez: Thanks for sending these notes @bozana , it's really good to know.

controllers/wizard/fileUpload/form/SubmissionFilesMetadataForm.php
```diff
@@ -50,11 +50,12 @@ public function __construct($submissionFile, $stageId, $reviewRound = null, $tem
...
         }
 
-        $submissionLocale = $submissionFile->getData('submissionLocale');
-        $publicationLanguageNames = Repo::submission()->get($submissionFile->getData('submissionId'))->getPublicationLanguageNames();
+        $submission = Repo::submission()->get($submissionFile->getData('submissionId'));
+        $submissionLocale = $submission->getData('locale');
 
-        $localeNames = Application::get()->getRequest()->getContext()->getSupportedSubmissionMetadataLocaleNames() + $publicationLanguageNames + $submissionFile->getLanguageNames();
-        ksort($localeNames);
+        $localeNames = collect(Application::get()->getRequest()->getContext()->getSupportedSubmissionMetadataLocaleNames() + $submission->getPublicationLanguageNames())
```
bozana, line 56: I think we need to consider `$submissionFile->getLanguageNames()` here too -- it could be that there was a locale enabled for submission metadata and a file name or file metadata were entered in that locale -- only file metadata none publication metadata -- and then that locale is disabled. We however still have those file metadata in that disabled locale in the DB and they should be visible and editable.

controllers/wizard/fileUpload/form/SubmissionFilesMetadataForm.php
bozana: It seems defaultFormLocale is set per default to the requiredLocale in the parent Form.php class, which is submission locale. So this seems to be fine.
Just that supportedLocales are set without considering the existing file languages ($submissionFile->getLanguageNames()) as said above.

Issue #12592, opened 2026-04-21
Allow email template variables in tasks and discussions


Vitaliy-1:
**Description:**
- Compile variables when the task/discussion is saved or edited
- Data anonymization in case of double blind reviews: reviewers shouldn't be able to use variables like `authors` and `authorsShort` 
- Additional variables: due date and task owner
- Pass the list of availables variables through the task templates API

**Steps to test the changes:**
1. Create a tsk template with variables 
2. Create a task based on the template
3. Save the task
4. See that variables are compiled 
5. Important: reviewers shouldn't be able to compile `authors` and `authorsShort`

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/12630
OJS: https://github.com/pkp/ojs/pull/5527

Vitaliy-1:
@ewhanson, can you take a look at these PRs?

ewhanson:
Thanks @Vitaliy-1! Just a few small questions and a few typos, but is otherwise ready for merge.

Vitaliy-1:
Thanks, merged!

jardakotesovec:
@Vitaliy-1 Hi Vitaliy,
this end up crashing when creating submission and also for any stage transition if there is automatic discussion/task to be added for given stage.


I made some quick temporary fixes [part1](https://github.com/pkp/pkp-lib/commit/d1e56bae717d228a2a2fc5dac9f0543a84bc7e70) [part2](https://github.com/pkp/pkp-lib/commit/9d0d28af4c52dfe052393f7a12f1f6976667ac06) to prevent the crashes rather then reverting, but it will require deeper dive and consider how exactly handle the automatically added task&discussions .

Maybe good timing to the variable replacement would be when the task is being saved and having some participants? In that moment I think it could be considered to be information complete.

Also not sure how we think about the sender for automatic tasks? That will stay null for automatic tasks?

Tribunal33:
Also @Vitaliy-1 general question should the template tasks be available for incomplete submissions?  I think the tasks should only be added to the submission after the submission is finished?

<img width="882" height="629" alt="Image" src="https://github.com/user-attachments/assets/e044101e-7918-486d-8aa8-87d2911db25e" />

jardakotesovec:
Problem of automatic variable replacements in submission wizard is covered here - https://github.com/pkp/pkp-lib/issues/12761

[...]

PR #12630, opened 2026-04-30
pkp/pkp-lib#12592 Compile task template variables

Vitaliy-1:


api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -372,7 +372,7 @@ public function editTask(EditTask $illuminateRequest): JsonResponse
         $oldParticipantIds = $editTask->participants->pluck('userId')->toArray();
         $headnote = $editTask->notes->where(fn (Note $note) => $note->isHeadnote == true)->first();
         $oldFiles = Repo::submissionFile()->getCollector()
-            ->filterByAssoc(PKPApplication::ASSOC_TYPE_NOTE, [$headnote->id])
+            ->filterByAssoc(PKPApplication::ASSOC_TYPE_NOTE, [$headnote?->id])
```
ewhanson: Does this still perform as expected if the second parameter is an array with null inside of it, or should the second parameter itself be null?

classes/editorialTask/TemplateVariables.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * @file classes/mail/mailables/EditorialReminder.php
```
ewhanson: File should be `classes/editorialTask/TemplateVariables.php`

classes/mail/variables/EditorialTaskEmailVariable.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @return string Get the full name of the user responsible for the task
+     */
+    protected function getTaskOwnerName(): string
+    {
+        if ($this->editorialTask->type !== EditorialTaskType::TASK->value) {
+            return '';
+        }
+        $taskOwner = Repo::user()->get($this->editorialTask->participants->where('isResponsible', true)->pluck('userId')->first());
+        return htmlspecialchars($taskOwner->getFullName());
```
ewhanson, line 75: Can `$taskOwner` ever be null?

ewhanson (changes requested):

classes/mail/variables/EditorialTaskEmailVariable.php
Vitaliy-1: Thanks, in theory can, e.g., when task is automatically created.

PR #12639, opened 2026-05-04
pkp/pkp-lib#12547 Session/cookie management update

touhidurabir:
for pkp/pkp-lib#12547 .


classes/middleware/PKPEncryptCookies.php
```diff
@@ -50,28 +50,11 @@ public function handle($request, Closure $next)
...
-                    ) . $cookie->getValue(),
-                    static::serialized($cookie->getName())
-                )
-            ));
-        }
-
-        return $response;
+        return parent::encrypt($response);
     }
 }
```
taslangraham: Missing newline at end of file.

taslangraham (approved):
@touhidurabir Just the one comment on formatting. I've approved the PR; this change doesn't need a second look. Thanks!

classes/middleware/PKPAuthenticateSession.php
```diff
@@ -14,12 +14,37 @@
...
+        $response = parent::handle($request, $next);
+
+        if ($request->hasSession()
+            && $request->user()
+            && Config::getVar('security', 'session_check_ip')
+            && ($loginIp = $request->session()->get('login_ip'))
+            && $loginIp !== $request->ip()
+        ) {
+            $this->logout($request);
+        }
```
bozana, line 43: Shell this check maybe rather go before calling parent::handle() ?
In case we have an attacker with a stolen session cookie from a different IP... 🤔

bozana: Not related to this issue, but shouldn't our logout() throw AuthenticationException as parent logout does?

classes/core/PKPContainer.php
```diff
@@ -395,16 +400,22 @@ protected function loadConfiguration(): void
         ];
 
         // Session manager
+        // session_lifetime is in days; fractional values are allowed (e.g. 0.5 = 12 hours, 0.0833 ≈ 2 hours).
+        // It is converted to minutes and clamped to a minimum of 1 minute so Laravel never receives 0
+        // (which it would treat as "immediately expired").
+        // For browser-close session expiration, use session_expire_on_close = On in config.inc.php
+        $sessionLifetimeMinutes = max(1, (int) round((float) Config::getVar('general', 'session_lifetime', 7) * 24 * 60));
```
bozana, line 407: In your comment you said that the default value would be 1 day and not 1 minute. That seems a better approach than 1 minute, no?

bozana (commented):
Hi @touhidurabir, two questions here...
Thanks!

classes/middleware/PKPAuthenticateSession.php
touhidurabir: > Shell this check maybe rather go before calling parent::handle() ? In case we have an attacker with a stolen session cookie from a different IP... 🤔

pipeline (see Dispatcher::initSession()) doesn't route the request and fully completes — IP check + logout + a terminating redirect()/exit — before (see `Dispatcher::dispatch()` --> first `$this->initSession()` --> then at end `$router->route($request)`), the attacker is logged out and redirected to login before any protected handler runs. parent::handle() running first leaks nothing: `$next` from the parent class returns the empty placeholder, so no app logic executes; it only re-stores the password hash in the attacker's own session.

> shouldn't our logout() throw AuthenticationException as parent logout does?

Not right now. That exception converted into `redirect()->guest(route('login'))` by the `Foundation\Exceptions\Handler::unauthenticated()` but our custom container does not have that mechanism . and `Validation::redirectLogin()` follows the same mechanism as `PKPPageRouter::handleAuthorizationFailure()` .

classes/core/PKPContainer.php
touhidurabir: > In your comment you said that the default value would be 1 day and not 1 minute. That seems a better approach than 1 minute, no?

yes that was the intended approach initially . But think about it, if a installation want to tighten the idlea session lief time to more strict value like (2/6/12 hours), and if we set min to 1 --> 24 hours, then they would not be able to do it . This allow us to provide fractional value as day which will be converted to mins (we need to convert the days to mins anyway) . But if someone give a value `0` which will cause immediate logout after login so that min `1` . 

I am open to revert this if seems confusion (e.g. dropping fractional value) and only allow days .

bozana (approved):

PR #12682, opened 2026-05-07
Add default envelope sender configuration handling

vovanshu:
Add logic to assign the default sender and its name from the configuration file config.inc.php

smtp_name = "My Publisher"
default_envelope_sender = my@mail.org
force_default_envelope_sender = On

### DIFF
--- a/classes/mail/Mailable.php
+++ b/classes/mail/Mailable.php
@@ -302,6 +302,16 @@ class Mailable extends IlluminateMailable
             );
         }
 
+        $configDefaultEnvelopeSender = Config::getVar('email', 'default_envelope_sender');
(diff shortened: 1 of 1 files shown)


asmecher:
@vovanshu, the `Message::sender()` function should be getting called via the `setEnvelopeSenderDefault` function, called [here](https://github.com/pkp/pkp-lib/blob/e9b2da7294b9c95b4a1b59c7fe0516ad947c438a/classes/mail/Mailer.php#L185). Was the `sender()` function not sufficient, needing you to also call `from()`? Or was `sender()` not getting called as expected? Some more context on the fix would be helpful, thanks!

Issue #12658, opened 2026-05-05
[Eloquent Model] | Passing empty array does not removed existing setting values for eloquent based model with settings table


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

From the https://github.com/pkp/pkp-lib/pull/12397#issuecomment-4376984368

There seems to be a challenge with using Eloquent and passing an empty array to delete existing values in settings table. This was noticed by @kaitlinnewson.

Scenario:
1. Add a new funder with a custom name without a ROR
2. Edit that funder, delete the custom name and add a new one with a ROR
3. See that the ROR has been saved, but at the same time the old custom name is still there.

So it seems that when we pass an empty array for the name setting, it does not remove existing values. 

I think it should happen here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L131 but probably that is not executed when using Eloquent?

So `$deleteSettings` is set correctly but the check there inside if `is_a($this, EntityDAO::class)` prevents the deletion of being executed.

**Also additional related problem.**

`$deleteSettings` also inclused `grants`even if there is actual data set for grants. it means that if the settings deletion code would run, it would also wipe the grants data.

1. Add new funder with grants data
2. Edit the funder and log around here https://github.com/pkp/pkp-lib/blob/main/classes/core/traits/EntityUpdate.php#L74

So because `grants` has not been changed when the data is edited, it is not present in props there and ends up being subject to deletion. The only reason why it is not deleted though is that `$deleteSettings` is not passed for deletion because of the issue above.

### Steps to Reproduce

1. Store a model with settings data
2. Update the model instance with empty data for a setting 
3. See that corresponding setting value for the setting field not removed form the settings table

### Expected Result

Should removed the settings values (e.g. rows) form the settings table of the corresponding setting for which set data as empty array (`[]`) at update.

### Actual Result

Does not removed the settings values (e.g. rows) form the settings table of the corresponding setting for which set data as empty array (`[]`) at update.

### Application Version

OJS, OMP and OPS 3.5/main

### Additional Information

This also raise the question when passed null for setting data . So if we allow all setting rows to be deleted when passing empty array for setting data , say `name` in the Funder Model as \
```php
use PKP\funder\Funder;
$f = Funder::find(4);
$f->update([
    'name' => []
]);
$f->refresh();
```
which would produce result as 
```
 Before Update
PKP\funder\Funder {#1324
    funder_id: 4,
[...]

PR #12683, opened 2026-05-07
pkp/pkp-lib#12658 Multilingula setting data delete at empty array or null pass via eloquent

touhidurabir:
for pkp/pkp-lib#12658 .

### DIFF
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -271,6 +271,7 @@ class SettingsBuilder extends Builder
      * Overrides Illuminate\Database\Query\Builder to support settings in select queries
      *
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
+     * @param  mixed  $values
      * @param  string  $boolean
      *
      * @return $this
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -159,17 +159,15 @@ class SettingsBuilder extends Builder
      */
     public function delete(): int
     {
-        $id = parent::delete();
-        if (!$id) {
-            return $id;
-        }
-
+        // Delete the child settings rows first, then the main row. Child-before-parent
+        // is the correct relational-integrity discipline: it works on any schema,
+        // cascade or not, and doesn't depend on the FK constraint as a safety net.
         DB::table($this->getSettingsTable())->where(
             $this->getPrimaryKeyName(),
             $this->model->getRawOriginal($this->getPrimaryKeyName()) ?? $this->model->getKey()
         )->delete();
 
-        return $id;
+        return parent::delete();
     }
 
     /**
@@ -244,7 +242,6 @@ class SettingsBuilder extends Builder
      * Overrides Illuminate\Database\Query\Builder to support settings in select queries
      *
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
-     * @param  mixed  $values
      * @param  string  $boolean
      * @param  bool  $not
      *
@@ -274,9 +271,8 @@ class SettingsBuilder extends Builder
      * Overrides Illuminate\Database\Query\Builder to support settings in select queries
      *
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
-     * @param  mixed  $values
      * @param  string  $boolean
-     * 
+     *
      * @return $this
      */
     public function whereNotIn($column, $values, $boolean = 'and')
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -135,8 +135,14 @@ class SettingsBuilder extends Builder
             fn (mixed $value, string $key) => in_array(Str::camel($key), $this->model->getSettings())
         );
 
-
-        $id = parent::insertGetId($primaryValues->toArray(), $sequence);
+        // Default to the model's PK column when no sequence is given. Without
+        // this, parent falls back to 'id' which breaks on PostgreSQL for any
+        // model whose primary key is not literally 'id' (RETURNING "id" →
+        // undefined column). Most of the ModelWithSettings consumers use <entity>_id as PK.
+        $id = parent::insertGetId(
+            $primaryValues->toArray(),
+            $sequence ?? $this->model->getKeyName()
+        );
 
         if ($settingValues->isEmpty()) {
             return $id;
@@ -238,6 +244,7 @@ class SettingsBuilder extends Builder
      * Overrides Illuminate\Database\Query\Builder to support settings in select queries
      *
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
+     * @param  mixed  $values
      * @param  string  $boolean
      * @param  bool  $not
      *
@@ -269,6 +276,7 @@ class SettingsBuilder extends Builder
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
      * @param  mixed  $values
      * @param  string  $boolean
+     * 
      * @return $this
      */
     public function whereNotIn($column, $values, $boolean = 'and')
--- a/classes/core/SettingsBuilder.php
+++ b/classes/core/SettingsBuilder.php
@@ -308,6 +308,14 @@ class SettingsBuilder extends Builder
         return $this->model->getSchemaName();
     }
 
+    /**
+     * Checks if setting is multilingual
+     */
+    public function isMultilingual(string $settingName): bool
+    {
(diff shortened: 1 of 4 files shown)


classes/core/SettingsBuilder.php
```diff
@@ -135,8 +135,14 @@ public function insertGetId(array $values, $sequence = null)
...
 
-
-        $id = parent::insertGetId($primaryValues->toArray(), $sequence);
+        // Default to the model's PK column when no sequence is given. Without
+        // this, parent falls back to 'id' which breaks on PostgreSQL for any
+        // model whose primary key is not literally 'id' (RETURNING "id" →
+        // undefined column). Most of the ModelWithSettings consumers use <entity>_id as PK.
+        $id = parent::insertGetId(
+            $primaryValues->toArray(),
+            $sequence ?? $this->model->getKeyName()
```
touhidurabir, line 144: This particular changes as `$sequence ?? $this->model->getKeyName()` instead of only `$sequence` provide support of Postgres when doing mass assignment as `Model::query()->insertGetId([...])` . 

So basically, anyone who, in the future, writes `Model::query()->insertGetId([...])` (e.g., for bulk seeding, a custom upsert, an import path that bypasses `Model::create`) will get a silent PostgreSQL-only failure.

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,773 @@
...
+        $this->assertSame([], $model->getAttribute('title'));
+        $this->assertSame([], $model->getAttribute('subtitle'));
+    }
+
+    // NOTE: A test for select() column subsets was removed pending the fix for
+    // the documented gotcha. See claude/ENTITY.md → Gotchas & Anomalies →
+    // "select() with a setting column crashes...". Today, SettingsBuilder does
+    // not partition the select column list, so select() either crashes (setting
+    // column present) or silently drops settings (PK omitted). Re-add a real
+    // test once getModelWithSettings() learns to partition the column list.
```
bozana: What about his comment? Should it stay? Should a new issue be opened?
Also, see claude/ENTITY.md in comment which probably should not be there :-)

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,773 @@
...
+        TestSettingsSchemaModel::query()->whereKey($idA)->delete();
+
+        // Primary row gone, but settings row leaks (current bug).
+        $this->assertSame(0, DB::table('test_settings_schema_entity')->where('test_id', $idA)->count());
+        $leftoverA = DB::table('test_settings_schema_entity_settings')->where('test_id', $idA)->count();
+        $this->assertSame(
+            0,
+            $leftoverA,
+            'If this fails with leftover settings rows, delete() needs to scope the settings delete by primary key.'
+        );
```
bozana: Why do we need this test? CASCADE FK cleans them up however, no?

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,773 @@
...
+        $modelId = $this->seedSchemaModel([
+            'en' => 'English',
+            'fr_CA' => 'Français',
+            'de_DE' => 'Deutsch',
+        ]);
+        $this->assertSettingRowCount('test_settings_schema_entity_settings', $modelId, 'title', 3);
+
+        TestSettingsSchemaModel::find($modelId)->delete();
+
+        $this->assertSame(0, DB::table('test_settings_schema_entity_settings')->where('test_id', $modelId)->count());
```
bozana, line 509: given the CASCADE FK on the settings table, the DB guarantees this anyway regardless of what SettingsBuilder does, no?

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,773 @@
...
+        $count = TestSettingsSchemaModel::query()
+            ->whereKey([$idA, $idB])
+            ->update(['parent_id' => 7]);
+        $this->assertSame(2, $count);
+
+        // Hydrated-model update returns bool (Eloquent's Model::update contract).
+        // Document the bool path here so a regression to int from Model::update
+        // would also trip — both contracts matter to callers.
+        $bool = TestSettingsSchemaModel::find($idA)->update(['parentId' => 8]);
+        $this->assertTrue($bool);
```
bozana: Do we need this test -- it seems to be Eloquent's own behavior, not application code - SettingsBuilder doesn't control or override it?

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,773 @@
...
+        ]);
+        $this->assertCount(
+            0,
+            $this->getSettingRows('test_settings_schema_entity_settings', $id2, 'title'),
+            'Empty multilingual array on insert should not produce settings rows'
+        );
+    }
+
+    //
+    // delete() tests
```
bozana, line 484: It seems to me that for all delete tests the test setting table should rather not have cascade fk, only the test testDeleteIsSafeWithCascadeForeignKey?

bozana (commented):
Hi @touhidurabir, the fix looks good to me, just a few comments on the deletion tests.

tests/classes/core/SettingsBuilderTest.php
touhidurabir: Yes, I have claude to run some automated tests which based on my design tests which flags this . So making a list of other gotchas and listing those using cluade in a file which being referenced here . will remove the comment reference later once all finalised .

classes/core/SettingsBuilder.php
```diff
@@ -153,17 +159,15 @@ public function insertGetId(array $values, $sequence = null)
...
+        // Delete the child settings rows first, then the main row. Child-before-parent
+        // is the correct relational-integrity discipline: it works on any schema,
+        // cascade or not, and doesn't depend on the FK constraint as a safety net.
         DB::table($this->getSettingsTable())->where(
             $this->getPrimaryKeyName(),
             $this->model->getRawOriginal($this->getPrimaryKeyName()) ?? $this->model->getKey()
         )->delete();
 
-        return $id;
+        return parent::delete();
```
touhidurabir, line 170: This alter the delete order in reverse as `Settings Shild --> Primary row` which provide better safety check without `FK cascade`. With the call like `Model::query()->whereKey($id)->delete()` raise exception when attempt to delete the settings child after a successful primary table row which lead to orphaned child settings . Note that it used to work for code as `Model::find($id)->delete()` which used all over the codebase .

tests/classes/core/SettingsBuilderTest.php
touhidurabir: Thanks @bozana , you are correct to flag this . Also by not considering the FK cascade, it revels to another gotcha related to delete at the query path like `Model::query()->whereKey($id)->delete()` .

tests/classes/core/SettingsBuilderTest.php
touhidurabir: @bozana I agree that mock schema should not have a delete cascade on FK and by that we can test if the delete mechanism will works properly even without a FK delete cascade . I have adjusted the tests accordingly which did help to identify a problem with delete order with a FK cascade on delete as pointed out at https://github.com/pkp/pkp-lib/pull/12683#discussion_r3218760649 .

classes/core/SettingsBuilder.php
```diff
@@ -267,8 +271,8 @@ public function whereIn($column, $values, $boolean = 'and', $not = false)
      * Overrides Illuminate\Database\Query\Builder to support settings in select queries
      *
      * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
-     * @param  mixed  $values
      * @param  string  $boolean
+     *
```
kaitlinnewson, line 276: Should this be kept (`$values`) for consistency in the PHPdoc?

tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,724 @@
...
+        $this->assertSame(0, DB::table('test_settings_schema_entity')->where('test_id', $modelId)->count());
+        $this->assertSame(0, DB::table('test_settings_schema_entity_settings')->where('test_id', $modelId)->count());
+    }
+
+    public function testDeleteRemovesAllLocaleRowsForOneId(): void
+    {
+        $modelId = $this->seedSchemaModel([
+            'en' => 'English',
+            'fr_CA' => 'Français',
+            'de_DE' => 'Deutsch',
```
kaitlinnewson:
```suggestion
            'de' => 'Deutsch',
```

kaitlinnewson (approved):

PR #12704, opened 2026-05-11
pkp/pkp-lib#11914 Migrate Credit Plugin data

ajnyga:
https://github.com/pkp/pkp-lib/issues/11914

### DIFF
--- a/classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
+++ b/classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
@@ -1,0 +1,119 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I11914_MigrateCreditRoles
+ *
+ * @brief Adds migration for Credit Roles data originating from the Credit plugin
+ */
+
+namespace PKP\migration\upgrade\v3_6_0;
+
+use Illuminate\Support\Facades\DB;
+use PKP\migration\Migration;
+use PKP\install\DowngradeNotSupportedException;
+
+class I11914_MigrateCreditRoles extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        $creditRoleSettings = DB::table('author_settings')
+            ->where('setting_name', '=', 'creditRoles')
+            ->whereNotNull('setting_value')
+            ->where('locale', '=', '')
+            ->get();
+
+        if ($creditRoleSettings->isEmpty()) {
+            $this->cleanUpPlugin();
+            return;
+        }
+
+        $creditRoleIdByIdentifier = DB::table('credit_roles')
+            ->pluck('credit_role_id', 'credit_role_identifier');
+
+        $newRows = [];
+
+        foreach ($creditRoleSettings as $setting) {
+            $identifiers = json_decode($setting->setting_value, true);
+
+            if (!is_array($identifiers)) {
+                $this->_installer->log(
+                    "Invalid credit role identifiers for author_id "
+                    . "{$setting->author_id}. Expected JSON array. Skipping."
+                );
+                continue;
+            }
+
+            $identifiers = array_unique(
+                array_map(fn($id) => str_replace('http://', 'https://', $id), $identifiers)
+            );
+
+            foreach ($identifiers as $identifier) {
+                $identifier = str_replace('http://', 'https://', $identifier);
+                $creditRoleId = $creditRoleIdByIdentifier->get($identifier);
+
+                if ($creditRoleId === null) {
+                    $this->_installer->log(
+                        "Credit role identifier '{$identifier}' for author_id "
+                        . "{$setting->author_id} not found in credit_roles table. Skipping."
+                    );
+                    continue;
+                }
+
+                $newRows[] = [
+                    'contributor_id'      => $setting->author_id,
+                    'credit_role_id'      => $creditRoleId,
+                    'credit_degree'       => null,
+                    'contributor_role_id' => null,
+                ];
+            }
+        }
+
+        if (empty($newRows)) {
+            $this->cleanUpPlugin();
+            return;
+        }
+
+        DB::transaction(function () use ($newRows) {
+            foreach (array_chunk($newRows, 500) as $chunk) {
+                DB::table('credit_contributor_roles')->insert($chunk);
+            }
+
+            DB::table('author_settings')
+                ->where('setting_name', '=', 'creditRoles')
+                ->delete();
+        });
+
+        $this->cleanUpPlugin();
+    }
+
+    /**
+     * Reverse the migration.
+     */
+    public function down(): void
+    {
+        throw new DowngradeNotSupportedException();
+    }
+
+    /**
+     * Clean up plugin settings and version entries for the Credit plugin.
+     */
+    private function cleanUpPlugin(): void
+    {
+        DB::table('plugin_settings')
+            ->where('plugin_name', '=', 'creditplugin')
+            ->delete();
+
+        DB::table('versions')
+            ->where('product_class_name', '=', 'CreditPlugin')
+            ->delete();
+    }
+}


classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I11914_MigrateCreditRoles.php
```
bozana: here without `.php`

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Get all existing credit role settings for authors, 
+        // and migrate them to the new credit_contributor_roles table
+        $creditRoleSettings = DB::table('author_settings')
+            ->where('setting_name', '=', 'creditRoles')
+            ->whereNotNull('setting_value')
```
bozana, line 30: Maybe to add `->where('locale', '=', '')` to be explicit ?

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+
+        // If there are no credit role settings, we can skip this migration
+        if ($creditRoleSettings->isEmpty()) {
+            return;
+        }
+
+        DB::transaction(function () use ($creditRoleSettings) {
+
+            // Get a mapping of credit role identifiers to their IDs for quick lookup
+            $creditRoleIdByIdentifier = DB::table('credit_roles')
```
bozana: this should be moved our of DB transaction, before the transaction.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+
+                    if ($creditRoleId === null) {
+                        $this->_installer->log(
+                            "Credit role identifier '{$identifier}' for author_id "
+                            . "{$setting->author_id} not found in credit_roles table. Skipping."
+                        );
+                        continue;
+                    }
+
+                    $newRows[] = [
```
bozana: This should be also moved out of DB transaction, before the transaction.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+                if (!is_array($identifiers)) {
+                    $this->_installer->log(
+                        "Invalid credit role identifiers for author_id "
+                        . "{$setting->author_id}. Expected JSON array. Skipping."
+                    );
+                    continue;
+                }
+
+                // For each identifier, find the corresponding credit_role_id and prepare a new row for insertion
+                foreach ($identifiers as $identifier) {
```
bozana: To be sure maybe to use `foreach (array_unique($identifiers) as $identifier) {`.

Then, also: `$identifier = str_replace('http://', 'https://', $identifier);` -- see https://github.com/pkp/credit/blob/stable-3_5_0/CreditPlugin.php#L213. -- It seems old plugin installation could have `http://` which will not match the new entries in credit_roles table.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+            ->where('setting_name', '=', 'creditRoles')
+            ->whereNotNull('setting_value')
+            ->get();
+
+        // If there are no credit role settings, we can skip this migration
+        if ($creditRoleSettings->isEmpty()) {
+            return;
+        }
+
+        DB::transaction(function () use ($creditRoleSettings) {
```
bozana: The DB transaction can only contain the insertions and cleaning up the author_settings table.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+            // Clean up old settings
+            DB::table('author_settings')
+                ->where('setting_name', '=', 'creditRoles')
+                ->delete();
+
+            // Clean up plugin settings for the Credit plugin, as they are no longer needed
+            DB::table('plugin_settings')
+                ->where('plugin_name', '=', 'creditplugin')
+                ->delete();
+
```
bozana: We should also consider cleaning of the DB table versions

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+                    $newRows[] = [
+                        'contributor_id'      => $setting->author_id,
+                        'credit_role_id'      => $creditRoleId,
+                        'credit_degree'       => null,
+                        'contributor_role_id' => null,
+                    ];
+                }
+            }
+
+            foreach (array_chunk($newRows, 100) as $chunk) {
```
bozana: maybe we can do only one insert?

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+                    }
+
+                    $newRows[] = [
+                        'contributor_id'      => $setting->author_id,
+                        'credit_role_id'      => $creditRoleId,
+                        'credit_degree'       => null,
+                        'contributor_role_id' => null,
+                    ];
+                }
+            }
```
bozana, line 77: Maybe to also add the check:
```
if (empty($newRows)) {   
clean DB table plugin_setting and versions                                                                                                                       
return;                                                                                                                                     
} 
```
In case, for some reason, all of the settings were skipped.
This is then also before the DB transaction.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+        // Get all existing credit role settings for authors, 
+        // and migrate them to the new credit_contributor_roles table
+        $creditRoleSettings = DB::table('author_settings')
+            ->where('setting_name', '=', 'creditRoles')
+            ->whereNotNull('setting_value')
+            ->get();
+
+        // If there are no credit role settings, we can skip this migration
+        if ($creditRoleSettings->isEmpty()) {
+            return;
```
bozana, line 36: Maybe the plugin could be installed but not used, and in this case the plugin_settings and versions DB table should be cleaned up too.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
```diff
@@ -0,0 +1,103 @@
...
+
+        });
+    }
+
+    /**
+     * Reverse the migration.
+     */
+    public function down(): void
+    {
+        throw new \PKP\install\DowngradeNotSupportedException();
```
bozana: Add a `use` statement for this at the beginning of the file

bozana (commented):
Hi @ajnyga, a few comments and suggestions. Thanks a lot!

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
ajnyga: are you sure the versions table is not cleaned automatically in the upgrades when I plugin is removed?

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
ajnyga: I have no idea how large datasets there can be, is there a limit for a single insert?

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
ajnyga: ok, just wait for you to react to the versions table comment I made above.

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
bozana: Or maybe to use bigger chunks -- e.g. 5000  or 10000? 100 looks very small for such data inserts

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
bozana: Yes, pretty sure. I think we always have considered `versions` table -- e.g. also for funders.
Deleting the directory leaves the `versions` row (and any `plugin_settings`). 
The upgrade step Installer::addPluginVersions() iterates only the plugins that are on the filesystem and does insertVersion() for each one, but there is no DELETE of stale rows anywhere in the install/upgrade code.

ajnyga: Ok, I think we would benefit from a more constructed way of dealing with removing plugins during upgrade, but I will add this!

classes/migration/upgrade/v3_6_0/I11914_MigrateCreditRoles.php
ajnyga: I understand from your comment that you mean we should move the fetching of newRows outside the transaction and do this check there. So like this?

```php
class I11914_MigrateCreditRoles extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        $creditRoleSettings = DB::table('author_settings')
            ->where('setting_name', '=', 'creditRoles')
            ->whereNotNull('setting_value')
            ->where('locale', '=', '')
            ->get();

        if ($creditRoleSettings->isEmpty()) {
            $this->cleanUpPlugin();
            return;
        }

        $creditRoleIdByIdentifier = DB::table('credit_roles')
            ->pluck('credit_role_id', 'credit_role_identifier');

        $newRows = [];

        foreach ($creditRoleSettings as $setting) {
            $identifiers = json_decode($setting->setting_value, true);

            if (!is_array($identifiers)) {
                $this->_installer->log(
                    "Invalid credit role identifiers for author_id "
                    . "{$setting->author_id}. Expected JSON array. Skipping."
                );
                continue;
            }

            foreach (array_unique($identifiers) as $identifier) {
                $identifier = str_replace('http://', 'https://', $identifier);
                $creditRoleId = $creditRoleIdByIdentifier->get($identifier);

                if ($creditRoleId === null) {
                    $this->_installer->log(
                        "Credit role identifier '{$identifier}' for author_id "
                        . "{$setting->author_id} not found in credit_roles table. Skipping."
                    );
                    continue;
                }

                $newRows[] = [
                    'contributor_id'      => $setting->author_id,
                    'credit_role_id'      => $creditRoleId,
                    'credit_degree'       => null,
                    'contributor_role_id' => null,
                ];
            }
        }

        if (empty($newRows)) {
            $this->cleanUpPlugin();
            return;
        }

        DB::transaction(function () use ($newRows) {
            foreach (array_chunk($newRows, 500) as $chunk) {
                DB::table('credit_contributor_roles')->insert($chunk);
            }

            DB::table('author_settings')
                ->where('setting_name', '=', 'creditRoles')
                ->delete();
        });

        $this->cleanUpPlugin();
    }

    /**
     * Reverse the migration.
     */
    public function down(): void
    {
        throw new DowngradeNotSupportedException();
    }

    /**
     * Clean up plugin settings and version entries for the Credit plugin.
     */
    private function cleanUpPlugin(): void
    {
        DB::table('plugin_settings')
            ->where('plugin_name', '=', 'creditplugin')
            ->delete();

        DB::table('versions')
            ->where('product_class_name', '=', 'CreditPlugin')
            ->delete();
    }
}
```

ajnyga: @bozana did you catch this

bozana: Hi @ajnyga, sorry, I first now came to it...
Yes, this looks correct, only one thing: we should first normalize, then deduplicate -- normalization `$identifier = str_replace('http://', 'https://', $identifier);` could theoretically lead to a duplicate. Thus, maybe something like:
```
  $identifiers = array_unique(
      array_map(fn($id) => str_replace('http://', 'https://', $id), $identifiers)
  );

  foreach ($identifiers as $identifier) {
      $creditRoleId = $creditRoleIdByIdentifier->get($identifier);
      // ...
  }
```

Issue #12714, opened 2026-05-12
[Structured Citations] - fix citation jobs for rate limits and stale status code


bozana:
### Problem

The citation lookup jobs called Crossref, OpenAlex and ORCID without taking those services' rate limits into account at all. A 429 fell through the status switch's `default: return`, so the citation was dropped silently, while 408 and 504 threw, which stops the remaining jobs in the chain and leaves the citation frozen mid-pipeline. The requests also had no timeout.

`OrcidJob` additionally carried a stale `$service->statusCode` between authors, so a 404 for one author could clear the next author's ORCID iD when their request threw instead of returning.

### Result

- Each service is throttled proactively, with jittered releases: Crossref 3 req/s with a `mailto` and 1 req/s without, ORCID 11 req/s and 24,500 requests/day, OpenAlex 90 req/s.
- A 429/503 releases the job for `Retry-After` + 3s, or 60s when the service sends none.
- A 408/5xx/transport failure prepends a delayed retry to the chain — 8 attempts escalating from 5m to ~10.5h — instead of throwing.
- A lookup that exhausts its retries fails into `failed_jobs`, where it is visible and redispatchable, and the citation is marked `FAILED` instead of staying mid-stage. `FAILED = -1` sorts below every processing stage, so a redispatched citation re-runs every lookup.
- Citation lookups use a 20s request / 10s connect timeout.
- `OrcidJob` is split into a fan-out plus a per-author `OrcidAuthorJob`, so one job is one ORCID request and the daily quota can be counted.
- `CrossrefJob` skips the lookup when the citation already has a DOI.
- All citation jobs return instead of throwing when their citation no longer exists, which is what `Repository::importCitations()` causes by deleting and re-inserting citations with new IDs.
- Fixed the stale `$service->statusCode`, and the switch-case literals that compared strings (`'408'`, `'504'`) against the `int $statusCode`.

### PRs:
- https://github.com/pkp/pkp-lib/pull/12715
- https://github.com/pkp/ojs/pull/5537 (only submodule update)
- https://github.com/pkp/pkp-lib/issues/13308 — citation progress reporting in the UI, spun out

bozana:
Hi @touhidurabir, could you please review this fix?
Thanks a lot!

asmecher:
> CrossrefJob, OrcidJob, OpenAlexJob: handle HTTP 429 by releasing the job back to the queue with a 60-second delay ($this->release(60)) instead of silently returning

[...]

Issue #12355, opened 2026-02-17
Consider ROR and OpenAlex API keys


bozana:
From Feb 13 2026, OpenAlex API calls require a key. S. https://docs.openalex.org/how-to-use-the-api/rate-limits-and-authentication.
ROR introduced them as well. S. https://ror.readme.io/docs/rest-api.
Journal using RORs and structured citations would need to register -- it is free -- to get the key.
We need an input fields somewhere in the settings, where users would enter them, and this setting needs to be considered when using the APIs.

PR #12715, opened 2026-05-12
pkp/pkp-lib#12714 fix citation jobs for rate limits and stale status …

bozana:
…code

s. https://github.com/pkp/pkp-lib/issues/12714

### DIFF
--- a/classes/citation/externalServices/ExternalServicesHelper.php
+++ b/classes/citation/externalServices/ExternalServicesHelper.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/citation/externalServices/ExternalServicesHelper.php
  *
- * Copyright (c) 2025 Simon Fraser University
- * Copyright (c) 2025 John Willinsky
+ * Copyright (c) 2025-2026 Simon Fraser University
+ * Copyright (c) 2025-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class Helpers
@@ -19,6 +19,7 @@ namespace PKP\citation\externalServices;
 use APP\core\Application;
 use Exception;
 use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\Exception\RequestException;
 
 class ExternalServicesHelper
 {
@@ -49,11 +50,14 @@ class ExternalServicesHelper
      * Makes HTTP request to the API and returns the response as an array.
      *
      * @param string $url The API endpoint URL.
+     * @param array $options Guzzle request options.
+     * @param int|null &$retryAfter Set to the response's Retry-After value (in seconds), if the server sent one.
      *
      * @return array|int|null The response as an associative array, request status code or null.
      */
-    public static function apiRequest(string $url, array $options = []): array|int|null
+    public static function apiRequest(string $url, array $options = [], ?int &$retryAfter = null): array|int|null
     {
+        $retryAfter = null;
         $httpClient = Application::get()->getHttpClient();
 
         try {
@@ -70,6 +74,22 @@ class ExternalServicesHelper
             }
 
             return $result;
+
+        } catch (RequestException $e) {
+            // Guzzle throws (rather than returning a response) for 4xx/5xx statuses by default,
+            // so this is where a real 429/408/504/etc from the service actually needs to be caught.
+            $response = $e->getResponse();
+
+            if (!$response) {
+                error_log(__METHOD__ . ' ' . $e->getMessage());
+                return null;
+            }
+
+            if ($response->hasHeader('Retry-After')) {
+                $retryAfter = (int) $response->getHeaderLine('Retry-After');
+            }
+
+            return $response->getStatusCode();
 
         } catch (GuzzleException|Exception $e) {
             error_log(__METHOD__ . ' ' . $e->getMessage());
--- a/classes/citation/externalServices/ExternalServicesHelper.php
+++ b/classes/citation/externalServices/ExternalServicesHelper.php
@@ -86,7 +86,7 @@ class ExternalServicesHelper
             }
 
             if ($response->hasHeader('Retry-After')) {
-                $retryAfter = (int) $response->getHeaderLine('Retry-After');
+                $retryAfter = self::parseRetryAfter($response->getHeaderLine('Retry-After'));
             }
 
             return $response->getStatusCode();
@@ -96,5 +96,23 @@ class ExternalServicesHelper
         }
 
         return null;
+    }
+
+    /**
+     * Parses a Retry-After header value into a delay in seconds. Per RFC 9110, the value may
+     * be either a number of delay-seconds or an HTTP-date to wait until.
+     */
+    protected static function parseRetryAfter(string $value): ?int
+    {
+        if (is_numeric($value)) {
+            return max(0, (int) $value);
+        }
+
+        $timestamp = strtotime($value);
+        if ($timestamp === false) {
+            return null;
+        }
+
+        return max(0, $timestamp - time());
     }
 }
--- a/classes/citation/externalServices/openAlex/Inbound.php
+++ b/classes/citation/externalServices/openAlex/Inbound.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/citation/externalServices/openalex/Inbound.php
  *
- * Copyright (c) 2025 Simon Fraser University
- * Copyright (c) 2025 John Willinsky
+ * Copyright (c) 2025-2026 Simon Fraser University
+ * Copyright (c) 2025-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class OpenAlexInbound
@@ -28,6 +28,9 @@ class Inbound
     /** @var int Status code of external service response. */
     public int $statusCode = 200;
 
+    /** @var int|null Retry-After value (in seconds) sent by the external service, if any. */
+    public ?int $retryAfter = null;
+
     /** @var string Email address of journal contact. */
     public string $contactEmail = '';
 
@@ -43,7 +46,8 @@ class Inbound
     {
         $response = ExternalServicesHelper::apiRequest(
             $this->url . '/works/doi:' . urlencode($citation->getData('doi')),
-            ['headers' => ['mailto' => $this->contactEmail]]
+            ['headers' => ['mailto' => $this->contactEmail]],
+            $this->retryAfter
         );
 
         if (is_int($response)) {
--- a/classes/citation/externalServices/openAlex/Inbound.php
+++ b/classes/citation/externalServices/openAlex/Inbound.php
@@ -45,8 +45,8 @@ class Inbound
     public function getWork(Citation $citation): ?Citation
     {
         $response = ExternalServicesHelper::apiRequest(
-            $this->url . '/works/doi:' . urlencode($citation->getData('doi')),
-            ['headers' => ['mailto' => $this->contactEmail]],
+            $this->url . '/works/doi:' . urlencode($citation->getData('doi')) . '?mailto=' . urlencode($this->contactEmail),
+            [],
             $this->retryAfter
         );
 
--- a/classes/citation/externalServices/orcid/Inbound.php
+++ b/classes/citation/externalServices/orcid/Inbound.php
@@ -40,6 +40,8 @@ class Inbound
      */
     public function getAuthor(array $author): ?array
     {
+        $this->statusCode = 200;
+
         $response = ExternalServicesHelper::apiRequest(
             $this->url . '/' . urlencode(Orcid::removePrefix($author['orcid'])),
             ['headers' => ['mailto' => $this->contactEmail]]
--- a/classes/citation/externalServices/orcid/Inbound.php
+++ b/classes/citation/externalServices/orcid/Inbound.php
@@ -3,8 +3,8 @@
 /**
  * @file classes/citation/externalServices/orcid/Inbound.php
  *
- * Copyright (c) 2025 Simon Fraser University
- * Copyright (c) 2025 John Willinsky
+ * Copyright (c) 2025-2026 Simon Fraser University
+ * Copyright (c) 2025-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class Inbound
@@ -16,13 +16,24 @@
 
 namespace PKP\citation\externalServices\orcid;
 
+use APP\core\Application;
+use DateInterval;
+use Illuminate\Support\Facades\Cache;
 use PKP\citation\externalServices\ExternalServicesHelper;
+use PKP\context\Context;
+use PKP\orcid\OrcidManager;
 use PKP\pid\Orcid;
+use Throwable;
 
 class Inbound
 {
+    /**
+     * Maximum time (in seconds) to cache an ORCID access token.
+     */
+    protected const MAX_ACCESS_TOKEN_CACHE_SECONDS = 60 * 60 * 24;
+
     /** @var string The base URL for API requests. */
-    public string $url = 'https://pub.orcid.org/v2.1';
+    public string $url = OrcidManager::ORCID_API_URL_PUBLIC . OrcidManager::ORCID_API_VERSION_URL;
 
     /** @var int Status code of external service response. */
     public int $statusCode = 200;
@@ -33,9 +44,13 @@ class Inbound
     /** @var string Email address of journal contact. */
     public string $contactEmail = '';
 
-    public function __construct(string $contactEmail)
+    /** @var Context|null The journal, if any, whose ORCID integration settings (if configured) should be used for authenticated lookups. */
+    protected ?Context $context;
+
+    public function __construct(string $contactEmail, ?Context $context = null)
     {
         $this->contactEmail = $contactEmail;
+        $this->context = $context;
     }
 
     /**
@@ -46,11 +61,19 @@ class Inbound
         $this->statusCode = 200;
         $this->retryAfter = null;
 
-        $response = ExternalServicesHelper::apiRequest(
-            $this->url . '/' . urlencode(Orcid::removePrefix($author['orcid'])),
-            ['headers' => ['mailto' => $this->contactEmail]],
-            $this->retryAfter
-        );
+        $orcidId = urlencode(Orcid::removePrefix($author['orcid']));
+        $url = $this->url . $orcidId;
+        $options = ['headers' => ['mailto' => $this->contactEmail]];
+
+        if ($this->context !== null && OrcidManager::isEnabled($this->context)) {
+            $accessToken = $this->getAccessToken();
+            if ($accessToken !== null) {
+                $url = OrcidManager::getApiPath($this->context) . OrcidManager::ORCID_API_VERSION_URL . $orcidId;
+                $options = ['headers' => ['Authorization' => 'Bearer ' . $accessToken]];
+            }
+        }
+
+        $response = ExternalServicesHelper::apiRequest($url, $options, $this->retryAfter);
 
         if (is_int($response)) {
             $this->statusCode = $response;
@@ -71,5 +94,68 @@ class Inbound
         }
 
         return $author;
+    }
+
+    /**
+     * Clear the cached access token, e.g. after a 401 response indicates it's no longer valid.
+     */
+    public function clearCachedAccessToken(): void
+    {
+        if ($this->context !== null) {
+            Cache::forget($this->getAccessTokenCacheKey());
+        }
+    }
+
+    /**
+     * Get (and cache) a client-credentials OAuth access token for reading public ORCID
+     * records, using the journal's configured ORCID integration credentials. This grants
+     * a better rate/quota tier (Public registered or Member) than anonymous access.
+     */
+    protected function getAccessToken(): ?string
+    {
+        $cacheKey = $this->getAccessTokenCacheKey();
+
+        $cachedToken = Cache::get($cacheKey);
+        if ($cachedToken !== null) {
+            return $cachedToken;
+        }
+
+        try {
+            $response = Application::get()->getHttpClient()->request(
+                'POST',
+                OrcidManager::getApiPath($this->context) . OrcidManager::OAUTH_TOKEN_URL,
+                [
+                    'form_params' => [
+                        'grant_type' => 'client_credentials',
+                        'client_id' => OrcidManager::getClientId($this->context),
+                        'client_secret' => OrcidManager::getClientSecret($this->context),
+                        'scope' => '/read-public',
+                    ],
+                    'headers' => ['Accept' => 'application/json'],
+                ]
+            );
+        } catch (Throwable $e) {
+            error_log(__METHOD__ . ' ' . $e->getMessage());
+            return null;
+        }
+
+        $data = json_decode($response->getBody(), true);
+        if (empty($data['access_token'])) {
+            return null;
+        }
+
+        // MAX_ACCESS_TOKEN_CACHE_SECONDS always wins in practice — ORCID's real token expiry is on the order of decades.
+        $ttlSeconds = min(self::MAX_ACCESS_TOKEN_CACHE_SECONDS, max(60, (int) ($data['expires_in'] ?? 0) - 60));
+        Cache::put($cacheKey, $data['access_token'], DateInterval::createFromDateString("{$ttlSeconds} seconds"));
+
+        return $data['access_token'];
+    }
+
+    /**
+     * Cache key for this journal's ORCID client-credentials access token.
+     */
+    protected function getAccessTokenCacheKey(): string
+    {
+        return 'orcid-read-public-token-' . $this->context->getId();
     }
 }
--- a/classes/citation/externalServices/orcid/Inbound.php
+++ b/classes/citation/externalServices/orcid/Inbound.php
@@ -27,6 +27,9 @@ class Inbound
     /** @var int Status code of external service response. */
     public int $statusCode = 200;
 
+    /** @var int|null Retry-After value (in seconds) sent by the external service, if any. */
+    public ?int $retryAfter = null;
+
     /** @var string Email address of journal contact. */
     public string $contactEmail = '';
 
@@ -41,10 +44,12 @@ class Inbound
     public function getAuthor(array $author): ?array
     {
         $this->statusCode = 200;
+        $this->retryAfter = null;
 
         $response = ExternalServicesHelper::apiRequest(
             $this->url . '/' . urlencode(Orcid::removePrefix($author['orcid'])),
-            ['headers' => ['mailto' => $this->contactEmail]]
+            ['headers' => ['mailto' => $this->contactEmail]],
+            $this->retryAfter
         );
 
         if (is_int($response)) {
--- a/classes/citation/externalServices/orcid/Inbound.php
+++ b/classes/citation/externalServices/orcid/Inbound.php
@@ -65,7 +65,8 @@ class Inbound
         $url = $this->url . $orcidId;
         $options = ['headers' => ['mailto' => $this->contactEmail]];
 
-        if ($this->context !== null && OrcidManager::isEnabled($this->context)) {
+        // A sandbox-registered app's token wouldn't be honored by the production API these (real) citation authors live on.
+        if ($this->context !== null && OrcidManager::isEnabled($this->context) && !OrcidManager::isSandbox($this->context)) {
             $accessToken = $this->getAccessToken();
             if ($accessToken !== null) {
                 $url = OrcidManager::getApiPath($this->context) . OrcidManager::ORCID_API_VERSION_URL . $orcidId;
--- a/jobs/citation/OpenAlexJob.php
+++ b/jobs/citation/OpenAlexJob.php
@@ -3,8 +3,8 @@
 /**
  * @file jobs/citation/OpenAlexJob.php
  *
- * Copyright (c) 2025 Simon Fraser University
- * Copyright (c) 2025 John Willinsky
+ * Copyright (c) 2025-2026 Simon Fraser University
+ * Copyright (c) 2025-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class OpenAlexJob
@@ -59,11 +59,14 @@ class OpenAlexJob extends BaseJob
 
         if (empty($citationChanged)) {
             switch ($service->statusCode) {
-                case '408':
-                case '504':
+                case 408:
+                case 504:
                     throw new JobException(__('admin.job.failed.connection.externalService', [
                         'statusCode' => $service->statusCode]));
-                case '404':
+                case 429:
+                    $this->release(60);
+                    return;
+                case 404:
                 default:
                     return;
             }
--- a/jobs/citation/OpenAlexJob.php
+++ b/jobs/citation/OpenAlexJob.php
@@ -64,7 +64,7 @@ class OpenAlexJob extends BaseJob
                     throw new JobException(__('admin.job.failed.connection.externalService', [
                         'statusCode' => $service->statusCode]));
                 case 429:
-                    $this->release(60);
+                    $this->release($service->retryAfter !== null ? $service->retryAfter + 3 : 60);
                     return;
                 case 404:
                 default:
--- a/jobs/citation/OpenAlexJob.php
+++ b/jobs/citation/OpenAlexJob.php
@@ -24,6 +24,9 @@ use PKP\jobs\BaseJob;
 
 class OpenAlexJob extends BaseJob
 {
+    /** Rate-limit releases count as attempts too, so retry indefinitely; $maxExceptions still catches real failures. */
+    public $tries = 0;
+
     protected int $contextId;
     protected int $citationId;
     protected string $contactEmail = '';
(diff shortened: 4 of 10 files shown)


touhidurabir:
Hi @bozana you assessment is correct as mentioned at https://github.com/pkp/pkp-lib/issues/12714#issuecomment-4440154016

> This delays the individual job for a citation, that reaches the rate limit and receives the rate limit result. Maybe a note that this is a job chain for one citation. When the next citation job using the same endpoint is run and the limit is reset then it goes trough. I.e. not all existing jobs are delayed, but the one specific, for one citation, that gets the rate limit result.

The increase of attempt at each release is also a bit concern as the attempt get increased with each time jobs being pulled from queue for processing . and we have default `$tries` to `3` in BaseJob means if it get rate limited for 3 consecutive time with a max time cap of 180 (as `$this->release(60)`) , that can lead to chain failure . 

what we can do is have a retryUntil time limit instead of a hard max tries like
```php
public $tries = 0;

public function retryUntil(): \DateTimeInterface
 {
     return now()->addMinutes(30);
 }
```

`$tries = 0` is treated as "unlimited" by `Worker::markJobAsFailedIfAlreadyExceedsMaxAttempts();` only `retryUntil()` decides when to give up.

So with return `now()->addMinutes(30)` and `$tries = 0`, we will have following as 

- First dispatch → deadline frozen at T + 30 min where T is when the job was first queued.                                                                   
- Between T and T + 30 min: the job can be popped, hit 429, release(60), get re-popped, hit 429 again, release, … any number of times. No attempt cap.
- At any pop where now > T + 30 min: Worker::markJobAsFailedIfAlreadyExceedsMaxAttempts fails the job with MaxAttemptsExceededException before handle() runs.                                                                                                                                           
- That permanent failure triggers Bus::chain(...)->catch(...) in Repository::reprocessCitation() — which currently only error_logs the exception. Remaining jobs in that citation's chain (IsProcessedJob, etc.) do NOT run, so the citation stays at whatever processing_status the failed job last set.

for this case, we do not need the job rate limiting and that will not fix our problem .

bozana:
Hi @touhidurabir, I first now came to this again. Hmm... I am not sure retryUntil is better for these jobs:
An article could have 100 citations, that means 100 jobs chains will be dispatched, if then one -- usually crossref -- reaches the rate limit, that chain will be put at the end of the queue. So until 100 citations' jobs are processed lots of time can pass. So it could happen that the earlier job that reached the rate limit will not be run at all, or maybe only once. 
This is for only one article, but it could be several, especially in the multi-journal installation.
The problem we experience very concrete is the ORE import: we import 1000 of articles, some with even several versions, so there is a huge amount of citations jobs. By this amount of jobs, some will surely reach a rate limit, and with retryUntil they will never run again.
So maybe it wold be better to leave tries=3 or higher instead of retryUntil for these jobs? What do you think?
Thanks a lot!

bozana:
Hi @touhidurabir, one more thing I thought about:
Maybe a RateLimiter for these jobs (above all for crossref and orcid jobs) would also be good to have -- when rate limit happens, pause all queued jobs with the same name for 60 (or so) seconds. Because. Right now, if e.g. Crossref starts rate-limiting during a 100-citation batch, each of those 100 jobs independently discovers the 429 by making its own wasted HTTP call, then releases on its own 60s timer — a thundering herd that keeps hammering the service (e.g. Crossref or Orcid) and eventually prolonging the rate limit. With a shared limiter, one job's 429 immediately silences all the others for the pause window, with zero extra HTTP calls.
What do you think?

bozana:
Hi @touhidurabir, I did a few more changes:

- added Retry-After header support, that I now consider instead of guessing 60s
- added a job-level reate-limiter middleware, keeping our request rate under each service's documented limit, so that 429 and 503 become rare exception
- fixed a few things, like mailto for Crossref, possible case 503 for Crossref and ORCID
- considered configured ORCID credentials, to use access token for reading public ORCID records, which allows slightly better rate limits
- considered ORCID daily rate limit
- added crossref and orcid rate limiter tests

I will still need to work on this issue: https://github.com/pkp/pkp-lib/issues/12355 -- to then consider OpenAlex API key fro these lookups.

Could you please review? What do you think?

jobs/citation/OrcidJob.php
```diff
@@ -72,13 +100,22 @@ public function handle(): void
 
             if (empty($authorChanged)) {
                 switch ($service->statusCode) {
-                    case '404':
+                    case 401:
+                        // Cached access token is invalid; clear it so the next lookup fetches a fresh one.
+                        $service->clearCachedAccessToken();
+                        break;
```
touhidurabir: the comment says to do a fresh look up as `Cached access token is invalid` but it does not release back the job like for the `429/503` , is that intended ? should a similar release back to queue for next try happen here ? 

Ignore if this is intended as I am not sure of the intentions .

jobs/citation/OrcidJob.php
```diff
@@ -72,13 +100,22 @@ public function handle(): void
...
-                    case '504':
+                    case 408:
+                    case 504:
                         throw new JobException(__('admin.job.failed.connection.externalService', [
                             'statusCode' => $service->statusCode]));
+                    case 429:
+                    case 503:
+                        // ORCID returns 503 (rather than 429) specifically when the burst allowance is exceeded.
+                        $this->release($service->retryAfter !== null ? $service->retryAfter + 3 : 60);
+                        return;
```
touhidurabir, line 127: here once a 429/500 hit, we are immediately releasing the job back to queue for next try .

If I understand the mechanism correctly, so this happens 

say we got 10 authors --> make request for 1 each --> 6 resolved --> hit API rate limit at 7th --> immediate release back to queue --> lost the progress of 6 resolved . 

so if the above assumption is correct, we lost quite a bit of progress . perhaps we should update up the resolved one and then release back to queue .

jobs/citation/OrcidJob.php
touhidurabir: Here the calls flows as PKP\citation\externalServices\orcid\Inbound::getAuthor() --> PKP\citation\externalServices\orcid\Inbound::getAccessToken() --> can return a null --> check in the PKP\citation\externalServices\orcid\Inbound::getAuthor() if not null to make API request but no check if null --> No point in making an API request if access token is returned null . 

Ignore if I am misunderstanding it .

classes/citation/externalServices/openAlex/Inbound.php
```diff
@@ -43,7 +46,8 @@ public function getWork(Citation $citation): ?Citation
     {
         $response = ExternalServicesHelper::apiRequest(
             $this->url . '/works/doi:' . urlencode($citation->getData('doi')),
-            ['headers' => ['mailto' => $this->contactEmail]]
+            ['headers' => ['mailto' => $this->contactEmail]],
```
touhidurabir: for crossref, `mailto` read as query param not as HTTP header but here for open alex, it still read as HTTP header, is that intentional ?

classes/citation/externalServices/ExternalServicesHelper.php
```diff
@@ -71,6 +75,22 @@ public static function apiRequest(string $url, array $options = []): array|int|n
...
+            // so this is where a real 429/408/504/etc from the service actually needs to be caught.
+            $response = $e->getResponse();
+
+            if (!$response) {
+                error_log(__METHOD__ . ' ' . $e->getMessage());
+                return null;
+            }
+
+            if ($response->hasHeader('Retry-After')) {
+                $retryAfter = (int) $response->getHeaderLine('Retry-After');
```
touhidurabir: just a question, are we sure that the `Retry-After` always return as int value, as I know some service sent the retry after as specific timestamp(ok as int) or `DateTime` which will turn a issue with int cast ? in that case we need to use strtotime or Carbon. 

Ignore if confirmed that the retry after is always int for these services .

classes/citation/externalServices/orcid/Inbound.php
```diff
@@ -16,34 +16,64 @@
...
     {
-        $response = ExternalServicesHelper::apiRequest(
-            $this->url . '/' . urlencode(Orcid::removePrefix($author['orcid'])),
-            ['headers' => ['mailto' => $this->contactEmail]]
-        );
+        $this->statusCode = 200;
+        $this->retryAfter = null;
+
+        $orcidId = urlencode(Orcid::removePrefix($author['orcid']));
+        $url = $this->url . $orcidId;
```
touhidurabir, line 65: this directly use the production API endpoint but the later fallback via `OrcidManager::getApiPath` get the API path via context configured settings . I think we should follow one single pattern as what if the context is configured for sandbox only .

classes/citation/externalServices/orcid/Inbound.php
```diff
@@ -65,4 +95,67 @@ public function getAuthor(array $author): ?array
...
+
+        return $data['access_token'];
+    }
+
+    /**
+     * Cache key for this journal's ORCID client-credentials access token.
+     */
+    protected function getAccessTokenCacheKey(): string
+    {
+        return 'orcid-read-public-token-' . $this->context->getId();
```
touhidurabir, line 161: the cache key is same for prod or sandbox . so once a switch from prod/sandbox, the cache does not get cleared and the next call will cached access token of mismatched env(prod vs sanbox) and continue to fail . perhaps we can do something similar as OrcidManager::getApiPath() 

Not very important as switching to prod/sandbox does not happen for a production journal .

jobs/citation/OpenAlexJob.php
```diff
@@ -59,11 +59,14 @@ public function handle(): void
...
         if (empty($citationChanged)) {
             switch ($service->statusCode) {
-                case '408':
-                case '504':
+                case 408:
+                case 504:
                     throw new JobException(__('admin.job.failed.connection.externalService', [
                         'statusCode' => $service->statusCode]));
-                case '404':
+                case 429:
```
touhidurabir, line 78: unlike crossref and orcid, does not consider 503. not sure if that is intentional

touhidurabir (commented):
@bozana I added few comments, please check these . 

However we need to have a certain decision on retrying mechanism as 
 - try infinite times and only failed if real exception arise (currently 3 by `BaseJob::$maxExceptions`) via `408/504` or by other reason .
 - add a proper ceiling of tries via `$tries` (set to 3 via `BaseJob::$tries`) or via `retryUntil` which will be hard clocking ticking form the time of job dispatch . 
 
IMO, the default value of `BaseJob::$tries` is too low . Either we can make it higher like 10/12 or a much higher `retryUntil` like 12/24/48/72 hours . 

This is the most important decision to confirm here. 

On a different note, the `OrcidJob` seems to making multiple API/External requests unlike the crossref/openalex as it's making API requests in a loop . This kind of breaks the limiting caps as

`Limit::perSecond(11)` is meant to keep us under ~11 requests/sec. The limiter enforces 11 jobs/sec. But if the average citation has, say, 5 ORCID authors --> 11 jobs/sec  ×  5 authors/job  =  55 actual ORCID HTTP requests/sec --> 5× over the intended 11/sec.  Same for the daily cap.

but for the crossref/openalex we have as
 - CrossrefJob::handle()  → $service->getWork()   → 1 HTTP request to Crossref
 - OpenAlexJob::handle()  → $service->getWork()   → 1 HTTP request to OpenAlex
 One job → one external call. The limiter spends 1 hit, one request goes out. simple and accurate.

This can turn into a issue for OrcidJob . probably we can 
 - hit the limiter per author inside the loop (consume/check a hit before each getAuthor() call), so N calls spend N hits
 - restructure to one job per author (or per ORCID lookup), so the existing per-job middleware is accurate again. this is what is crossref/openalex is doing now . 
 
up to you if you want to make the changes . 

BTW, the `OrcidJob` rise the timeout question if we keep multiple request in loop as many external API requests may not be done with default 60 seconds timeout by `BaseJob::$timeout` , IMO it will need a increase .

classes/citation/externalServices/openAlex/Inbound.php
bozana: I considered it, and later with that new issue that will consider api key I will replace it with api key.

classes/citation/externalServices/orcid/Inbound.php
bozana: I am now considering the possibility of sandbox below

classes/citation/externalServices/orcid/Inbound.php
bozana: no the sandbox is not considered, see above

classes/citation/externalServices/ExternalServicesHelper.php
bozana: Hopefully I have considered it now correctly? 😅

jobs/citation/OpenAlexJob.php
bozana: added below

jobs/citation/OrcidJob.php
bozana: Oh, yes, correct! Now I tried to solve it differently: rate limiter registered once in handle() and hit once per author via reserveRateLimitSlot(); partial progress persisted via  releasePartialProgress() before any release; 401 retried once with a fresh token via fetchAuthor(); orcidLookedUp flag prevents redoing already-resolved authors on resume

jobs/citation/OrcidJob.php
bozana: fetchAuthor() now retries the same author once immediately, in-place, against a freshly-fetched token

jobs/citation/OrcidJob.php
bozana: yes, solved with releasePartialProgress and orcidLookedUp setting for each author

bozana:
Hi @touhidurabir, thanks a lot! All really good comments and findings! Some I have not thought about at all. I now solved it all like this:
- OrcidJob.php: $tries = 0, $timeout = 300; rate limiter registered once in handle() and hit once per author via reserveRateLimitSlot(); partial progress persisted via releasePartialProgress() before any release; 401 retried once with a fresh token via fetchAuthor(); orcidLookedUp flag prevents redoing already-resolved authors on resume
- CrossrefJob.php, OpenAlexJob.php: $tries = 0
- OpenAlexJob.php: 503 handled alongside 429
- ExternalServicesHelper.php: Retry-After parsing handles both delay-seconds and HTTP-date forms
- orcid/Inbound.php: skips the OAuth token path for sandbox-registered apps (token wouldn't be honored by production, where real citation-authors' data lives)
- openAlex/Inbound.php: mailto moved to query param
- schemas/citation.json: added authors[].orcidLookedUp (boolean)

Do you think $tries = 0 is safe here? Else, I would increase it, maybe to 12?
Could you please take another look?
Thanks a lot!!!

classes/citation/externalServices/ExternalServicesHelper.php
```diff
@@ -71,10 +75,44 @@ public static function apiRequest(string $url, array $options = []): array|int|n
...
+
+            if ($response->hasHeader('Retry-After')) {
+                $retryAfter = self::parseRetryAfter($response->getHeaderLine('Retry-After'));
+            }
+
+            return $response->getStatusCode();
+
         } catch (GuzzleException|Exception $e) {
             error_log(__METHOD__ . ' ' . $e->getMessage());
         }
```
touhidurabir, line 99: here for the `RequestException`, return a status code but nothing for the `GuzzleException|Exception`so return to null. So say for `OrcidJob` this cause to falls with status code 200 , will cause to survive through the OrcidJob's switch and set `orcidLookedUp` to `true` and will have citation processing status set to CitationProcessingStatus::ORCID .

jobs/citation/OrcidJob.php
```diff
@@ -58,37 +80,92 @@ public function handle(): void
...
         }
 
-        $service = new Inbound($this->contactEmail);
+        $context = Application::getContextDAO()->getById($this->contextId);
+        $service = new Inbound($this->contactEmail, $context);
+
+        RateLimiter::for(self::RATE_LIMITER_NAME, fn () => [
+            Limit::perSecond(self::PER_SECOND_LIMIT),
+            Limit::perDay(self::DAILY_LIMIT),
+        ]);
```
touhidurabir: According to https://laravel.com/docs/12.x/routing#multiple-rate-limits, we should make the multiple cache key distinct otherwise may collide with other ones .

jobs/citation/OrcidJob.php
```diff
@@ -58,37 +80,92 @@ public function handle(): void
...
+    /** @return int|null Null if a slot was reserved, otherwise the number of seconds to wait before retrying. */
+    protected function reserveRateLimitSlot(): ?int
+    {
+        foreach (Collection::wrap(RateLimiter::limiter(self::RATE_LIMITER_NAME)()) as $limit) {
+            if (RateLimiter::tooManyAttempts($limit->key, $limit->maxAttempts)) {
+                return RateLimiter::availableIn($limit->key);
+            }
+
+            RateLimiter::hit($limit->key, $limit->decaySeconds);
+        }
```
touhidurabir: same issue as https://github.com/pkp/pkp-lib/pull/12715/changes#r3701638334, we are writing un-namespaced global cache keys. making distinction by using the `by` should fix it . 

Also this is fine now as we have defined the `RateLimiter::for` set before this method call but otherwise it may return null from `RateLimiter::limiter` from unspecified ones and can break via call of `$limit->key` .

classes/citation/externalServices/orcid/Inbound.php
```diff
@@ -16,34 +16,65 @@
...
+        $orcidId = urlencode(Orcid::removePrefix($author['orcid']));
+        $url = $this->url . $orcidId;
+        $options = ['headers' => ['mailto' => $this->contactEmail]];
+
+        // A sandbox-registered app's token wouldn't be honored by the production API these (real) citation authors live on.
+        if ($this->context !== null && OrcidManager::isEnabled($this->context) && !OrcidManager::isSandbox($this->context)) {
+            $accessToken = $this->getAccessToken();
+            if ($accessToken !== null) {
+                $url = OrcidManager::getApiPath($this->context) . OrcidManager::ORCID_API_VERSION_URL . $orcidId;
+                $options = ['headers' => ['Authorization' => 'Bearer ' . $accessToken]];
```
touhidurabir: should we do merge instead of re-assign for `$options` here ? not sure of the intention, so ignore if it is the intended approach .

classes/citation/externalServices/openAlex/Inbound.php
```diff
@@ -42,8 +45,9 @@ public function __construct(string $contactEmail)
     public function getWork(Citation $citation): ?Citation
     {
         $response = ExternalServicesHelper::apiRequest(
-            $this->url . '/works/doi:' . urlencode($citation->getData('doi')),
-            ['headers' => ['mailto' => $this->contactEmail]]
+            $this->url . '/works/doi:' . urlencode($citation->getData('doi')) . '?mailto=' . urlencode($this->contactEmail),
```
touhidurabir: the `$this->contactEmail` can be empty string so do we need to consider that and not to add it to url param ? if service accept it, then it's fine .

jobs/citation/OrcidJob.php
```diff
@@ -58,37 +80,92 @@ public function handle(): void
...
+            if (empty($author['orcid']) || !empty($author['orcidLookedUp'])) {
                 $authorsChanged[] = $author;
                 continue;
             }
 
-            $authorChanged = $service->getAuthor($author);
+            // One ORCID request per author, unlike Crossref/OpenAlex's one per job, so hit the limiter per author.
+            $waitSeconds = $this->reserveRateLimitSlot();
+            if ($waitSeconds !== null) {
+                $this->releasePartialProgress($citation, $authors, $authorsChanged, $waitSeconds + 3);
```
touhidurabir, line 105: this seems to write unconditionally to DB even if the `$authorsChanged` is empty . perhaps the `if` above should be gated by `count($authorsChanged) > 0` ? that is writing only when at least one author actually changed. Ignore if I am misundestanding .

jobs/citation/OrcidJob.php
```diff
@@ -58,37 +80,92 @@ public function handle(): void
...
+        RateLimiter::for(self::RATE_LIMITER_NAME, fn () => [
+            Limit::perSecond(self::PER_SECOND_LIMIT),
+            Limit::perDay(self::DAILY_LIMIT),
+        ]);
 
         $authorsChanged = [];
 
         foreach ($authors as $author) {
-            if (empty($author['orcid'])) {
+            if (empty($author['orcid']) || !empty($author['orcidLookedUp'])) {
```
touhidurabir, line 95: just one confusion related to `!empty($author['orcidLookedUp'])` but at the `PKP\citation\maps\Schema::getCitationAuthorDataModel()` where we have 
```php
foreach (array_keys((array)$schema->properties->authors->items->properties) as $property) {
            $authorModel[$property] = '';
        }
```
e.g. unconditional assignment of empty string regardless of prop type or default(if have any) , is that going to cause issue .

touhidurabir (commented):
@bozana review added. mostly comment and questions . 

However on thing I am bit concern here is how to make sure job continue here so that it does not hit a try limit . setting `$tries` to `0` works but it make the jobs to run indefinitely and only allow to fail on real exception . But we are handling most of the API service related exception in `try...catch` so and also releasing . 

truthfully I do not have any experience without setting a hard ceiling for jobs deadline and not sure how it will behave or impact the system . so wondering if we can do following as 
```php
public $tries = 0;  // no hard attempt limit

public function retryUntil(): DateTime
{
    return now()->addHours(48);
}
```

classes/citation/externalServices/openAlex/Inbound.php
bozana: Now, I check if the contactEmail is empty and if not use I use it

classes/citation/externalServices/orcid/Inbound.php
bozana: I removed mailto -- ORCID does not use it at all, and now I am merging

classes/citation/externalServices/ExternalServicesHelper.php
bozana: Good catch. Fixed: ConnectException and the generic GuzzleException|Exception catch now return 504 instead of null. A connection failure / timeout / DNS error is reported as a gateway failure, so Crossref/OpenAlexJob throw a JobException (retried via $maxExceptions, then failed_jobs) and OrcidJob's switch hits case 504 and throws too. null now means only "empty/unparseable 2xx body".

jobs/citation/OrcidJob.php
bozana: Fixed. The two limits now get explicit namespaced keys via `->by()`: orcid-lookups:per-second and orcid-lookups:per-day. I also removed the RateLimiter::for() / RateLimiter::limiter()() round‑trip entirely — the limits are built directly in a new rateLimits() method that reserveRateLimitSlot() iterates.

jobs/citation/OrcidJob.php
bozana: Fixed. getCitationAuthorDataModel() (and the identical getDataCitationAuthorDataModel()) now pick a type‑appropriate empty value from the schema: boolean → false, array/object → [], otherwise ''. So orcidLookedUp is false, not "". (This only affected the mapped API output — OrcidJob reads orcidLookedUp from raw `$citation->getData('authors')`, never through the map, so the `!empty(...)` check was correct either way. But better to solve it correctly.)

jobs/citation/OrcidJob.php
bozana: `$authorsChanged` is actually never empty at that point — every author is pushed exactly once, and the empty‑authors case returns earlier — so `count() > 0` is always true there. But the underlying point is valid, so I added a `$performedLookup` flag: when no author needed an ORCID lookup (none have an orcid, or all were already resolved on a prior run), the job now returns without calling `Repo::citation()->edit()` at all.

jobs/citation/OrcidJob.php
bozana: Both addressed by the same change as above. reserveRateLimitSlot() now iterates `$this->rateLimits()` (limits carry explicit `->by()` keys, no empty‑string keys that could collide with other limiters). The RateLimiter::limiter(...)() call is gone and the method no longer depends on handle() having registered anything first.

bozana:
Hi @touhidurabir, thanks a lot for that great and helpful review! I think I fixed all the last comments/questions, and left the comments/explanation to each.

On the retry ceiling: I went with `$tries = 500` (there jobs mostly call third-party services so they are different from other jobs) + `$backoff = 300` on all three jobs, and not `retryUntil()`.  — In a large import the job can be for days in the queue — every job that doesn't get picked up within the window would be failed without ever running. `$tries` only decrements when the job actually runs or release()s, so queue backlog is free. 500 at ≥60s spacing is ~8h of retrying before failed_jobs — generous enough to ride out a real outage, bounded enough that a permanently‑stuck job surfaces. `$maxExceptions = 3` still fails genuinely broken jobs fast. Also added 500/502 to the fail‑fast switch (alongside 408/504), and `$backoff = 300` so a transient gateway blip gets ~10 min to clear before the job fails.

Could you please take another look at it all? Maybe especially this retry ceiling, the rate limiting, and Guzzle exceptions.
Sorry that it takes so long and so much — not easy not understand and recognize everything :confused: 
Thanks a lot!

jobs/citation/OrcidJob.php
```diff
@@ -16,14 +16,41 @@
...
+     * Just under ORCID's anonymous-tier 25k reads/day-per-IP quota (assumes a single app
+     * server/shared cache); applied uniformly since the tier isn't known at this stage.
+     */
+    protected const DAILY_LIMIT = 24500;
+
+    /**
+     * Calls a rate-limited external API and self-releases on 429/503, each release spending
+     * an attempt — hence far above BaseJob's default. $maxExceptions still bounds real errors.
+     */
+    public $tries = 500;
```
touhidurabir: what mentioned at https://github.com/pkp/pkp-lib/pull/12715#issuecomment-5527964418, I agree it . But we can not have a 500 val of the `attempts` column of `jobs` table which is a `tinyint unsigned` and max to `255` . so better to reduce it to something like 253/254.  here and other places .

jobs/citation/OrcidJob.php
```diff
@@ -58,37 +85,113 @@ public function handle(): void
...
-        $service = new Inbound($this->contactEmail);
+        $context = Application::getContextDAO()->getById($this->contextId);
+        $service = new Inbound($this->contactEmail, $context);
 
         $authorsChanged = [];
+        $performedLookup = false;
 
         foreach ($authors as $author) {
-            if (empty($author['orcid'])) {
+            if (empty($author['orcid']) || !empty($author['orcidLookedUp'])) {
```
touhidurabir: If I understand it correctly, It will introduce a different issue when re-processing happens like in the `PKPCitationController::reprocessCitation` (or any other path) as by previous processing it's set so it will skip here . or is that intended ? ignore if I am misunderstanding it .

jobs/citation/OpenAlexJob.php
```diff
@@ -59,11 +68,19 @@ public function handle(): void
...
         if (empty($citationChanged)) {
             switch ($service->statusCode) {
-                case '408':
-                case '504':
+                case 408:
+                case 500:
+                case 502:
+                case 504:
+                    // Service is unwell or unreachable: fail fast and visibly, then bulk-retry once it recovers.
                     throw new JobException(__('admin.job.failed.connection.externalService', [
```
touhidurabir: throwing here means chain will abort unless that is expected . here and other places . 

if this is due to temp service down, should we release with a delay to retry bit later with a log now that we have logging mechanism available as part of core . 

and if we do not want to increase the attempts in such case, we can use `$this->prependToChain` to dispatch fresh one .

touhidurabir (changes requested):
@bozana review added . While column size is definitely something we need to fix , I am not 100% sure of the other 2 (orcid lookup silent backout and chain abort on service down), so ignore if you think those are intentional .

jobs/citation/OrcidJob.php
bozana: I am now using 254

jobs/citation/OrcidJob.php
bozana: I have now removed orcidLookedUp fully and am prepending one new job per author. It seems like a cleaner and better solution overall.

jobs/citation/OpenAlexJob.php
bozana: yes, great catch and suggestion. now I am prepending to chain.

bozana:
Hi @touhidurabir, 

I did again a few refactoring and improvements, according to your and Claude's comments, as well as a few things I found out were wrong:
- a lookup now retries a failing service eight times over about 21 hours before giving up, and only then fails the job. I kept failed_jobs as the ending rather than returning, because a return reaches IsProcessedJob, which stamps PROCESSED on a citation nothing ever looked at. We fail only when the service never answered; "answered, nothing found" still completes the chain as before. 
- I re-read the services' documentation and corrected the limits: Crossref is 3/s polite and 1/s public, not 9 — those are the numbers for query endpoints, and query.bibliographic, which we use, is one.
- Added HTTP timeouts. Without them Guzzle waits indefinitely, the job outlives its `$timeout`, and Laravel kills the worker — leaving the job reserved rather than failed or retried. Pre-existing, unrelated to this issue.
- There is now one ORCID job per author. That removed the reason orcidLookedUp existed at all, so the flag, releasePartialProgress() and the schema property are all gone — including from the reprocess paths, which no longer have anything stale to carry.
-  ORCID is treated as the anonymous tier only, which seems to be high enough for normal usage. The limiter is registered before the request goes out and can't know which tier it will use, since a token can fail to fetch; and the daily quota is counted per IP, so assuming the higher tier would overrun a budget shared by every journal on the server.
- I introduced JitteredRateLimited, so throttled jobs don't all come back at the same instant and re-trip the limit together. It also holds a job back 30–120s on a per-second window rather than the second or two Laravel gives, because every pickup spends one of the 254 attempts a job has before it fails.
- The failure message now carries the citation id, and admin.job.failed.connection.externalService is removed.

Tested against the live services — the same 67-citation publication reprocessed with 3 workers, 8 workers and the built-in job runner. Crossref peaked at 3 req/s against its limit of 3, with no 429 in any run. 

While testing: jobs.php work was silently ignoring --force and --stop-when-empty — I fixed it in its own commit. Correct so?
I would have two more questions regarding jobs.php:
1. jobs.php total and the admin Jobs page hide reserved rows
Repo::job()->total() is nonEmptyQueue()->nonReserved()->count(), and nonReserved() is whereNull('reserved_at'). During the debugging for this issue it reported 0 while four jobs sat in the table with reserved_at set — killed workers had left them stranded, which is exactly the state you're looking for when something has gone wrong. The same call backs api/v1/jobs/PKPJobController.php:94, so the admin UI shows the same number.
Is excluding reserved jobs the intended meaning of "total" — ready-to-run rather than present — or should it count the table? 

2. jobs.php run reports work as completed while delayed jobs remain
run() counts and pops through getJobModelBuilder(), which applies isAvailable() — available_at <= now(). Delayed jobs are therefore neither counted nor run. A queue holding nothing but retries waiting on a future available_at prints "the queue is empty"; and after the loop it prints "Completed running `{$jobCount}` jobs" using the count taken before the loop, while retries sit waiting.
Is the fix just the wording, or should run report how many are waiting?'

Could you please take another look?
Huge thanks!

jobs/citation/CitationLookupJob.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        if ($this->serviceRetries >= static::MAX_SERVICE_RETRIES) {
+            $this->fail(new Exception(sprintf(
+                'Citation %d: %s abandoned after %d retries (HTTP %d)',
+                $this->citationId,
+                static::class,
+                $this->serviceRetries,
+                $statusCode
+            )));
+            return;
```
touhidurabir, line 59: this is correct but I have a concern and suggestion as what if the chain failed at CrossrefJob or OpenAlexJob and never reached to the IsProcessedJob. So the status value never get to update . which will cause the client side to continue to fire the setInterval (from citationManagerStore.js) while `processed < total` as long as the tab is open . 

so how about this 
```php
if ($this->serviceRetries >= static::MAX_SERVICE_RETRIES) {
    $citation = Repo::citation()->get($this->citationId);
    if ($citation) {
        $citation->setProcessingStatus(CitationProcessingStatus::FAILED->value);
        Repo::citation()->edit($citation, []);
    }
    $this->fail(new Exception(...));
    return;
}
```
and then in the `citationManagerStore.js` as 
```js
const citationTerminalCases = [pkp.const.citationProcessingStatus.PROCESSED,
  pkp.const.citationProcessingStatus.FAILED];
const processedCitations = computed(() =>
      structuredCitations.value.filter(c => citationTerminalCases.includes(c?.processingStatus)).length);
```

jobs/citation/CitationLookupJob.php
touhidurabir, line 1: another suggestion , how about we add a details failure to better debug as 
```php
public function failed(Throwable $e): void
{
    Log::error('Citation metadata lookup abandoned', [
        'job' => static::class,
        'contextId' => $this->contextId,
        'citationId' => $this->citationId,
        'serviceRetries' => $this->serviceRetries,
        'attempts' => $this->attempts(),
        'reason' => $e->getMessage(),
    ]);
}
```

give us bit more information about the whole failure .

touhidurabir (approved):
@bozana I think we are almost done . Just 2 suggestion but one is also a concern that only need a look . 

Anyway it's almost ready so approving, just take a look at the possibility of a dead invisible chain causing front/client polling all along .

jobs/citation/CitationLookupJob.php
bozana: Great suggestion! I have just moved that status update into the failed() function that you have suggested above.

jobs/citation/CitationLookupJob.php
bozana: yes, great suggestion!
I am also assigning the status = failed here. I also added the last citation status and publication ID to the log.

bozana:
Hi @touhidurabir, thanks a lot!
I added a new status FAILED, and considered it everywhere, also in the UI in this issue https://github.com/pkp/pkp-lib/issues/13308 that I asked Jarda to review. So, if no objection, I will merge when the tests pass...
Huge thanks!!! 
:tada:

Issue #12617, opened 2026-04-28
Fix the crossref metadata check to be a notification and not to disable the submission publishing


bozana:
If Crossref plugin is enabled, the default Automatic DOI Assignment option "Upon reaching the copyediting stage" selected,  and a submission published prior to the copyediting stage, so that no DOI has been assigned to the submission, the recently introduced Crossref metadata check will disable submission publishing, s.:

<img width="1428" height="334" alt="Image" src="https://github.com/user-attachments/assets/dfc7fd8a-014e-4d4a-a6b0-9a45a0c2be9e" />

Now that we support publishing in different, also earlier workflow stages (e.g. from the submission or review stage), we need to find a solution how to assign DOIs to those submissions versions that needs to be published in earlier workflow stages. Independent from this solution, the Crossref metadata check should not disable publishing -- it should be just a notification, and allow editor to proceed with publishing -- also because the journal/press/server might not want to assign DOIs to every submission. 

**Pull Requests**
pkp-lib: https://github.com/pkp/pkp-lib/pull/12717
ojs: https://github.com/pkp/ojs/pull/5538
ops(submodule update): https://github.com/pkp/ops/pull/1300
omp(submodule update): https://github.com/pkp/omp/pull/2350
crossref-ojs: https://github.com/pkp/crossref-ojs/pull/96

taslangraham:
@bozana, could you review the PRs listed in the issue description, please? Thanks in advance. Below is a screenshot of the crossref validation displayed as a notification instead of blocking the editor from publishing.

<img width="1462" height="772" alt="Image" src="https://github.com/user-attachments/assets/18f89848-89aa-4afd-9d78-0ebf9dd854cb" />

bozana:
Hi @taslangraham, I reviewed the PRs and left a comment -- a suggestion how to maybe solve it slightly differently. Let me know what do you think about it.
Thanks a lot!

taslangraham:
[...]

PR #12717, opened 2026-05-12
pkp/pkp-lib#12617 Don't let crossref validation block publishing

taslangraham:
For pkp/pkp-lib#12617

### DIFF
--- a/classes/publication/Repository.php
+++ b/classes/publication/Repository.php
@@ -301,6 +301,22 @@ abstract class Repository
         return $errors;
     }
 
+
+    /**
+     * Perform validations that should be treated as warnings instead of errors.
+     * @param Publication $publication
+     * @param Submission $submission
+     * @param array $allowedLocales
+     * @param string $primaryLocale
+     * @return array
+     */
+    public function validatePublishWarnings(Publication $publication, Submission $submission, array $allowedLocales, string $primaryLocale): array
+    {
+        $warnings = [];
+        Hook::call('Publication::validatePublishWarnings', [&$warnings, $publication, $submission, $allowedLocales, $primaryLocale]);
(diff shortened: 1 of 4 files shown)


classes/publication/Repository.php
```diff
@@ -301,6 +301,22 @@ public function validatePublish(Publication $publication, Submission $submission
...
     }
 
+
+    /**
+     * Perform validations that should be treated as warnings instead of errors.
+     * @param Publication $publication
+     * @param Submission $submission
+     * @param array $allowedLocales
+     * @param string $primaryLocale
+     * @return array
```
bozana: maybe no need for this -- it does not add any additional information ?

controllers/modals/publish/PublishHandler.php
```diff
@@ -99,7 +99,6 @@ public function publish($args, $request)
         $primaryLocale = $submissionContext->getPrimaryLocale();
         $allowedLocales = $submissionContext->getSupportedSubmissionLocales();
         $errors = Repo::publication()->validatePublish($this->publication, $this->submission, $allowedLocales, $primaryLocale);
```
bozana, line 101: Maybe to call here:
`$warnings = Repo::publication()->validatePublishWarnings($this->publication, $this->submission, $allowedLocales, $primaryLocale);`
and provide the $warnings to the publish template (publish.tpl) below:
`$templateMgr->assign('publishWarnings', $warnings);`
Template will then render the warnings and PublishForm does not need to be changed.
See also my comments in the OJS repo.

bozana (commented):
Hi @taslangraham, a few comments i.e. a suggestion how to maybe solve it slightly differently. What do you think?

PR #12745, opened 2026-05-14
Normalize jpeg extension in announcement image validation - fixes #12744

### LINKED ISSUE #12744: [Announcements] | Website Settings - Announcements, .jpeg files are rejected although image/jpeg MIME type is valid
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

## Description

In OJS 3.5, when uploading an announcement image with the `.jpeg` extension, the upload is rejected even though the file is a valid JPEG image.

The same image works if renamed from `.jpeg` to `.jpg`.

## Steps to reproduce

1. Go to Journal Management > Website Settings > Announcements.
2. Create or edit an announcement.
3. Upload a valid JPEG image with the `.jpeg` extension.
4. Save the form.

## Expected behavior

The image should be accepted, since `.jpeg` is a valid extension for the `image/jpeg` MIME type.

## Actual behavior

The upload is rejected. If the same file is renamed to `.jpg`, it is accepted.

## Technical details

The issue appears to be in:

`lib/pkp/classes/announcement/Announcement.php`

In the `isValidImage()` method, the original file extension is compared with the extension returned by `FileManager::getImageExtension()`.

For `image/jpeg`, `FileManager::getImageExtension()` returns `.jpg`, but the uploaded file may have the `.jpeg` extension. This causes the comparison to fail:

```php
if ($extensionFromMimeType !== '.' . $extension) {
    return false;
}

A possible fix would be to normalize jpeg to jpg before comparing:

$extension = strtolower(pathinfo($temporaryFile->getOriginalFileName(), PATHINFO_EXTENSION));

if ($extension === 'jpeg') {
    $extension = 'jpg';
}


### Steps to Reproduce

## Steps to reproduce

1. Login with User/Role `Journal Manager`.
2. Navigate to `Website Settings > Announcements`.
3. Click on `Edit` for an existing announcement.
4. Scroll down to the `Image` section.
5. Upload a valid image file with the `.jpeg` extension.
6. Select the `Save` button.
7. See the validation error in the image field.

The same image is accepted if the file extension is changed from `.jpeg` to `.jpg`.

<img width="650" height="360" alt="Image" src="https://github.com/user-attachments/assets/872ce956-7667-478a-8936-5e7c9174880c" />

### Expected Result

The system should accept valid JPEG image files using either the `.jpg` or `.jpeg` file extension when the detected MIME type is `image/jpeg`.

After selecting the image and saving the announcement form, the form should be saved successfully and the uploaded image should be displayed without validation errors.

### Actual Result

When uploading a valid JPEG image using the `.jpeg` extension, the announcement form fails to save and displays an upload error in the image field.

The issue does not occur with the same image when the file extension is changed to `.jpg`, which suggests that the validation is rejecting the `.jpeg` extension even though the detected MIME type is `image/jpeg`.

### Environment Details

## Environment Details

- OJS version: 3.5.0.4
- Server OS: Linux / Ubuntu
- PHP version: 8.3.6
- Web server: Apache/2.4.58 (Ubuntu)
- Database: MariaDB 10.11.14 with `mysqli` driver
- Locale: `pt_BR`
- Base URL: `http://localhost`
- Browser: Google Chrome 148.0.7778.96, 64-bit
- Client device: Linux desktop

### Application Version

Open Journal Systems (OJS) 3.5.0.4

### Logs

```markdown id="x2j55x"
## Logs

A related PHP warning appears in the Apache/PHP error log:

```text
PHP Warning: Attempt to read property "image" on null in /var/www/ojs35/lib/pkp/classes/announcement/Announcement.php on line 302, referer: http://localhost/index.php/rt/pt_BR/management/settings/announcements

### Additional Information

## Additional Information

I attached a screenshot showing the validation error displayed in the announcement edit form after uploading a valid `.jpeg` image file.

The same image is accepted when renamed from `.jpeg` to `.jpg`.

During local investigation, the issue appears to be related to the validation logic in:

`lib/pkp/classes/announcement/Announcement.php`

In the `isValidImage()` method, the uploaded file extension is compared with the extension returned from the MIME type. For `image/jpeg`, `FileManager::getImageExtension()` returns `.jpg`, but the uploaded file may have the `.jpeg` extension. As a result, valid `.jpeg` files are rejected.

A temporary local workaround was tested by normalizing the extension before comparison:

```php
$extension = strtolower(pathinfo($temporaryFile->getOriginalFileName(), PATHINFO_EXTENSION));

if ($extension === 'jpeg') {
    $extension = 'jpg';
}

diegojmacedo:
## Summary

This pull request fixes #12744.

Announcement image upload currently rejects valid `.jpeg` files because `image/jpeg` is mapped to `.jpg` by `FileManager::getImageExtension()`, while the uploaded file extension may be `.jpeg`.

This change normalizes the uploaded file extension by treating `jpeg` as equivalent to `jpg` before comparing it with the MIME-derived extension.

## Testing

Tested locally with OJS 3.5.0.4:

- `.jpeg` announcement image upload failed before the change.
- `.jpeg` announcement image upload succeeded after the change.
- `.jpg` files continued to work as expected.

### DIFF
--- a/classes/announcement/Announcement.php
+++ b/classes/announcement/Announcement.php
@@ -389,7 +389,7 @@ class Announcement extends Model
         $extension = pathinfo($temporaryFile->getOriginalFileName(), PATHINFO_EXTENSION);
         if ($extension === 'jpeg') {
   		$extension = 'jpg';
-	}
+	} 
 	$fileManager = new FileManager();
         $extensionFromMimeType = $fileManager->getImageExtension(
             PKPString::mime_content_type($temporaryFile->getFilePath())
--- a/classes/announcement/Announcement.php
+++ b/classes/announcement/Announcement.php
@@ -387,7 +387,10 @@ class Announcement extends Model
             return false;
         }
         $extension = pathinfo($temporaryFile->getOriginalFileName(), PATHINFO_EXTENSION);
-        $fileManager = new FileManager();
+        if ($extension === 'jpeg') {
+  		$extension = 'jpg';
+	}
+	$fileManager = new FileManager();
         $extensionFromMimeType = $fileManager->getImageExtension(
             PKPString::mime_content_type($temporaryFile->getFilePath())
         );


asmecher:
@diegojmacedo, this works around the announcement image, but there is at least one more use case that would still reject `.jpeg` files:

https://github.com/pkp/pkp-lib/blob/84ca01fd546b6fb658a5daf4c27e9b8a82f9938b/api/v1/_uploadPublicFile/PKPUploadPublicFileController.php#L191

I don't think it's actually necessary to compare the uploaded and returned extensions at all; there are plenty of checks to make sure it's a valid image, and the server-side filename is determined from the MIME type, not the uploaded file extension. I would recommend just removing the failing test in `Announcement.php` and `PKPUploadPublicFileController.php`, and forcing the server-side filename to use the extension returned from `getImageExtension` rather than the uploaded filename.

Also, we're using spaces for indentation, but your PR uses some tabs; please configure the editor to use spaces.

Thanks for contributing!

Issue #12217, opened 2026-01-19
[Submission Decisions] | Submission workflow - Submission page, reviewers & authors who have disabled status throw errors when recording a decision


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Users who have disabled status throw errors when the decision is trying to be made due to a few areas.  Namely the email area.  Not sure if that is the only place it will need to be fixed.  

I would suggest that having a few users added to the dataset that have been disabled might help flesh out the dataset.  

### Steps to Reproduce

1. Prerequisite: Must have author or reviewer with disabled status associated with a particular submission
2. Login with editor role 
3. Navigate to the submission with author or reviewer in disabled status
4. Click on any decision option that would send emails.  (Skip and send to review, send to review, move to copyediting, move to production, publish etc...).

### Expected Result

Should be able to continue with the decision.  There isn't a requirement on what it should do for emailing disabled users but I would assume no email sent to users who are disabled.

Could also warn that email is going to be sent to a user who is disabled.  

### Actual Result

500 error.  See log.

### Application Version

OJS, OMP, OPS 3.3, 3.4, 3.5 and main

### Logs

[...]

PR #12773, opened 2026-05-19
pkp/pkp-lib#12217 [MAIN] Fix errors caused by disabled users when recording submission decision 

taslangraham:
For pkp/pkp-lib#12217

### DIFF
--- a/classes/decision/Steps.php
+++ b/classes/decision/Steps.php
@@ -44,13 +44,19 @@ class Steps
      *
      * @param bool $prepend Pass true to add this step before other steps
      */
-    public function addStep(Step $step, bool $prepend = false)
+    public function addStep(Step $step, bool $prepend = false): bool
     {
+        if (!$step->isValidStep()) {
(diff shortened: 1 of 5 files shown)


classes/decision/Steps.php
```diff
@@ -86,7 +86,11 @@ public function getStageParticipants(int $roleId): array
 
         $users = [];
         foreach (array_unique($userIds) as $authorUserId) {
-            $users[] = Repo::user()->get($authorUserId);
+            $user = Repo::user()->get($authorUserId);
```
taslangraham: If the user is not found or is disabled, `null` would be returned and appended to the array. This `null` value causes errors downstream; either type errors or errors resulting from attempting to read properties on a `null` value. The code here and across the other PRs adds sanity checks to prevent such errors.

touhidurabir (approved):

kaitlinnewson (approved):

Issue #12774, opened 2026-05-20
[Themes] | Themes can't override Blade component classes like they can override templates


NateWr:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The template hierarchy allows themes to easily override built-in template files. For example, calling the `<x-layout />` component will search for a template file in the following locations:

- `plugins/themes/<active-theme>/templates/components/layout.blade`
- `plugins/themes/<parent-themes>/templates/components/layout.blade`
- `templates/components/layout.blade`
- `lib/pkp/templates/components/layout.blade`

Blade components may have a [class](https://laravel.com/docs/11.x/blade#components) which serves data to the template. Ideally, the app should load and run the first class it finds in the following locations:

- `plugins/themes/<active-theme>/classes/components/Layout.php`
- `plugins/themes/<parent-themes>/classes/components/Layout.php`
- `classes/components/Layout.php`
- `lib/pkp/classes/components/Layout.php`

However, in my testing, I have not found a way to load the correct class for a component based on the template hierarchy. The one important thing here is that the class may exist at any level of the hierarchy **or none at all**.

### Steps to Reproduce

For more details of the use case and sample code to help produce a test case, please look at this proposal for a base `Layout` component for themes:

https://github.com/NateWr/pkp-blade-theme-proposals/blob/main/layout.md

### Expected Result

A themer should be able to use `<x-layout />` with or without their own `Layout` class. A child theme should be able to extend the `Layout` class to customise the data available to the `<x-layout>` component.

### Actual Result

I am unable to figure out how to correctly set up the files to make this work. It may require adjustments to the Blade view handling.

### Application Version

main branch, OJS 3.6 pre-release

### PRs:
pkp-lib: https://github.com/pkp/pkp-lib/pull/12775
ojs (tests): https://github.com/pkp/ojs/pull/5545

### Limitations
- override works only for the components without namespaces. So this is limitation for scenario if theme plugin wants to override class component from different plugin. That would work only for blade views, not the classes. Its intentional trade off to keep things simpler. If this ends up being real use case, we can reconsider.  We intend to mostly use simple anonymous components, so might not be issue.

jardakotesovec:
@NateWr Can you try the pkp-lib PR, if it was for you as expected?

In top post I added links to PRs and limitation note.

Here is example how such classes would look like:
## 1. PKP core

The default implementation, shipped with `lib/pkp`. Extend `Component`.

```php
<?php
[...]

PR #12775, opened 2026-05-20
pkp/pkp-lib#12774 Support Blade component classes overrides

jardakotesovec:


touhidurabir (approved):
Looks good . simple and to the point .

PR #12776, opened 2026-05-20
pkp/pkp-lib#12217 Fix errors caused by disabled users when recording submission decision [stable-3.5]

taslangraham:
For pkp/pkp-lib#12217

### DIFF
--- a/classes/decision/Step.php
+++ b/classes/decision/Step.php
@@ -52,4 +52,13 @@ abstract class Step
 
         return $config;
     }
+
+    /**
+     * Indicate if a step is valid.
+     * Child classes may override this method to perform additional validation checks.
+     */
+    public function isValidStep(): bool
+    {
+        return true;
+    }
 }
--- a/classes/decision/steps/Email.php
+++ b/classes/decision/steps/Email.php
@@ -47,7 +47,7 @@ class Email extends Step
         $this->attachers = $attachers;
         $this->locales = $locales;
         $this->mailable = $mailable;
-        $this->recipients = $recipients;
+        $this->recipients = array_filter($recipients, fn ($recipient) => $recipient instanceof User && !$recipient->getDisabled());
     }
 
     /**
@@ -169,5 +169,11 @@ class Email extends Step
         }
 
         return $variables;
+    }
+
+    /** @copydoc Step::isValidStep() */
+    public function isValidStep(): bool
+    {
+        return !empty($this->recipients);
     }
 }
--- a/classes/decision/Steps.php
+++ b/classes/decision/Steps.php
@@ -44,13 +44,19 @@ class Steps
      *
      * @param bool $prepend Pass true to add this step before other steps
      */
-    public function addStep(Step $step, bool $prepend = false)
+    public function addStep(Step $step, bool $prepend = false): bool
     {
+        if (!$step->isValidStep()) {
+            return false;
+        }
+
         if ($prepend) {
             array_unshift($this->steps, $step);
         } else {
             $this->steps[$step->id] = $step;
         }
+
+        return true;
     }
 
     /**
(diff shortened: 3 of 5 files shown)


touhidurabir:
Hi @taslangraham this looks good . and I think it easier to understand this way . 

However, I have an alternative proposal , it seems the cause of issue coming from actually `PKP\decision\steps\Email` and `PKP\decision\Steps::addStep()` which add steps without checking the validity of that step addition . 

So perhaps a centralise approach can be like 
1. in the PKP\decision\Step
```php
public function isValidStep(): bool
{
    true;
}
```

2. in the PKP\decision\steps\Email, we can add filter on the prop `$recipients` in the `__construct` right before the assignment by checking where users are disable . and then set the `isValidStep` based on that is empty or not as 
```php
public function __construct(......)
{
    $this->recipients = $this->filterByActiveUsers($recipients)
}

protected function filterByActiveUsers(array $users): array
{
    // run the filtering process
}

public function isValidStep(): bool
{
    (bool) count($this->recipients);
}
```

3. in the `PKP\decision\Steps::addStep()`, we can do something like 
```php
/**
     * Add a step to the workflow
     *
     * @param bool $prepend Pass true to add this step before other steps
     */
    public function addStep(Step $step, bool $prepend = false): bool // better to change the return type to know if it added
    {
        if (!$step->isValidStep()) { return false; }

        if ($prepend) {
            array_unshift($this->steps, $step);
        } else {
            $this->steps[$step->id] = $step;
        }
        
        return true;
    }
```

this kind of allow use for future proofing for cases where a dev forgot to add all the checks added to the PRs . 

I am not confidant that this will be fully future proof (as I have not tested but something that should work theoretically), but it seems that we are adding the same checks again and again in the `PKP\decision\Steps` and other decision classed which kind of points that perhaps we are missing some sort of abstraction or centralise checking . 

can you please try and see what the result yields .

taslangraham:
@touhidurabir thanks for the review. I've applied and tested your suggestions along with a few tweaks elsewhere, and those worked. Please take a look, and if all is well, I'll port to the other apps. The null checks in `getStageParticipants`, `getDecidingEditors`, and `getReviewersFromAssignments` on [classes/decision/Steps.php](https://github.com/pkp/pkp-lib/pull/12776/changes#diff-0a7bc88d47676eb5454e65200af705e127aae612fd530fa8d6eb8cfccc582000) class are still there, as I don't think it's a good idea to have those methods still return null values in their array response.

touhidurabir (approved):
Looks good . Thanks .

taslangraham:
@touhidurabir All other PRs related to the issue have been updated. Could you give a review of those as well? PRs are listed [here](https://github.com/pkp/pkp-lib/issues/12217#issue-3828054718). Thanks in advance.

touhidurabir:
@taslangraham I have approved all the pkp-lib PRs .

taslangraham:
Thanks @touhidurabir! Could you also approve the submodule PRs, so that they will be ready for merge after I get a second review and approval from another member of the team? Thanks.

classes/decision/steps/Email.php
```diff
@@ -170,4 +170,10 @@ protected function getVariables(string $locale): array
 
         return $variables;
     }
+
+    /** @copydoc Step::isValidStep() */
+    public function isValidStep(): bool
+    {
+      return !empty($this->recipients);
```
kaitlinnewson:
```suggestion
        return !empty($this->recipients);
```

classes/decision/Step.php
```diff
@@ -52,4 +52,14 @@ public function getState(): stdClass
 
         return $config;
     }
+
+    /**
+     * Indicate if a step is valid.
+     * Child classes may override this method to perform additional validation checks.
+     * @return bool
```
kaitlinnewson: This line can be removed (return type is already indicated below)

kaitlinnewson (approved):

PR #12786, opened 2026-05-24
pkp/pkp-lib#12354 Add validation test for Data Citation identifier types

ajnyga:
https://github.com/pkp/pkp-lib/issues/12354

### DIFF
--- a/classes/pid/Pmcid.php
+++ b/classes/pid/Pmcid.php
@@ -28,7 +28,7 @@ class Pmcid extends BasePid
 
     /** @copydoc BasePid::validationRegexes */
     public const validationRegexes = [
-        '/^PMC\d+$/i'
+        '/^PMC\d+$/'
     ];
 
     /** @copydoc BasePid::prefix */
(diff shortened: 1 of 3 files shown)


classes/pid/Pmcid.php
```diff
@@ -40,7 +40,5 @@ class Pmcid extends BasePid
     /** @copydoc BasePid::alternatePrefixes */
     public const alternatePrefixes = [
         'pmcid',
-        'pmc',
-        'pmc:',
```
bozana, line 44: Maybe to leave this? This would then be the valid input: `pmc:PMC1234567`. Which is correct, no?
Removing 'pmc' does fix the bug for the valid PMC1234567, but 'pmc:' could stay, no?

bozana (commented):
Hi @ajnyga, just one comment. Thanks a lot!

ajnyga:
thanks @bozana  restarted the tests https://github.com/pkp/ojs/pull/5553

Issue #12780, opened 2026-05-21
[Login As] | Editorial workflow - Submission in Review, user with disabled state will cause a 500 error when login as feature is used


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

Originally this was discovered with a reviewer with disabled status but after further investigation this would appear to be any user who was disabled.  

So taking a look back at the how the different versions handled this
3.3 ->  Seems to have no issues with login in as a user who was disabled.  I think this is bad as I can still submit reviews and make editorial decisions based on this puppet like users who was disabled.
3.4 -> This one is a bit strange to explain.  It redirects to the login page but the session token is corrupted and hasn't been reset.  It still thinks you are logged in as the user who was disabled.  So when you login as an active user you will actually be logged in as this user with the parent user being the one who was disabled.
3.5 -> Similar to 3.4 in that the session token is corruptted but you get thrown and 500 error exception (see error code below) and the site will be broken for you.  To fix it you can go to dev window > Application > Cookies > and delete OSSID cookie.

### Steps to Reproduce

1. Preconditions: Must have a user who has been set to disabled status
2. Login with any user role with access to login as (Admin > JE > JM > ect..)
3. Navigate to Users & Roles
4. Click on three dots or (triangle for other versions)
5. Select Login as the user with disabled status
### Expected Result

Not sure what we should expect for a user who has been disabled.  But in 3.4 + we should expect the Session token OSSID cookie not to be corrupt and reset.

Here are a few ideas we could expect
- [ ] For any user who has been disabled remove (or grey out) this option.  If greyed out provide a hover over tool tip "This user has been disabled"
- [ ] Similar to 3.3 allow the Editor to manipulate the account on behalf of the user who was disabled.  <-- I do not like this one but that was what previous 3.3 LTS allows.

### Actual Result

In 3.4 you are redirected to the login screen but when you log in as an active user you actually will now be logged in as this active user and the user who is disabled will be the parent account.
In 3.5/3.6 it throws a 500 error (see log below)

### Application Version

OJS, OMP, OPS 3.4, 3.5, 3.6(main)

### Logs

`[Thu May 21 17:30:55.491016 2026] [negotiation:error] [pid 962839] [client 207.216.187.251:52534] AH00687: Negotiation: discovered file(s) matching request: /var/www/91.99.202.113.sslip.io/index (None could be negotiated)., referer: https://91.99.202.113.sslip.io/index.php
[...]

PR #12790, opened 2026-05-26
pkp/pkp-lib#12780 Fix log in as with disabled user account

asmecher:


classes/core/PKPUserProvider.php
```diff
@@ -55,25 +52,18 @@ public function __construct(
...
      */
     public function retrieveById($id)
     {
-        // If the request has already been initialized with the user object, use it.
-        if (($user = Application::get()->getRequest()->getUser()) && $user->getId() == $id) {
-            return $user;
-        }
-
-        return Repo::user()->get($id);
+        return Repo::user()->get((int) $id);
```
touhidurabir: missing second params to `Repo::user()->get()` as `true` otherwise it will be same issue .

touhidurabir (changes requested):
@asmecher just one fix is required and then it's good to merge .

Issue #12756, opened 2026-05-16
[Open Peer Review] Controls to select Review Rounds for Publication


jardakotesovec:
More to come..

PR #12808, opened 2026-05-28
pkp/pkp-lib#12756 FieldMultiSelect

jardakotesovec:


ewhanson:
Merged as part of pkp/pkp-lib#13123.

PR #12834, opened 2026-05-31
pkp/pkp-lib#12658 Multilingula setting data delete at empty array or null pass via eloquent

touhidurabir:
for pkp/pkp-lib#12658 .

### DIFF
--- a/tests/classes/core/SettingsBuilderTest.php
+++ b/tests/classes/core/SettingsBuilderTest.php
@@ -1,0 +1,727 @@
+<?php
+
+/**
+ * @file tests/classes/core/SettingsBuilderTest.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class SettingsBuilderTest
+ *
+ * @see \PKP\core\SettingsBuilder
+ * @see \PKP\core\traits\EntityUpdate
+ *
+ * @brief Unit tests for the public surface of \PKP\core\SettingsBuilder — the
+ * Eloquent query builder used by every ModelWithSettings consumer. Covers
+ * update(), insertGetId(), delete(), where()/whereIn()/whereNotIn(), getModels(),
+ * and the simple delegators, on both schema-based and pure-Eloquent models.
+ */
+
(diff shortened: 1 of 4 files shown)


tests/classes/core/SettingsBuilderTest.php
```diff
@@ -0,0 +1,727 @@
...
+        $idB = $this->seedSchemaModel(['en' => 'Banana']);
+        $idC = $this->seedSchemaModel(['en' => 'Cherry']);
+
+        $matched = TestSettingsSchemaModel::whereIn('title', ['Apple', 'Banana'])
+            ->orderBy('test_id')
+            ->get();
+
+        // On stable-3_5_0, SettingsBuilder::getModelWithSettings() keys the returned
+        // rows by primary key, so the hydrated collection is PK-keyed. Normalise with
+        // array_values() before asserting on order.
```
bozana, line 576: Hi @touhidurabir, why is it so in stable-3_5_0? Does this need to be?

touhidurabir: @bozana this is due to the PRs related to the issue https://github.com/pkp/pkp-lib/issues/12197 not being back ported to the stable-3.5.0 . I left a comment [here](https://github.com/pkp/pkp-lib/issues/12197#issuecomment-3889425313) to confirm but not sure .

bozana: Hmmm... What do you think? Because 3.5 will be LTS maybe it is worth backporting it? :thinking: 
Maybe to ask Alec? And/or open a new issue for that?
However, this issue and the PRs look fine, they can be merged. 
Thanks a lot!

touhidurabir: @bozana I think it can be back ported to stable-3.5.0 . However, the reason of this PR perhaps will not generate in the 3.5 . That PR handle a case of many to many relation and in stable branches we do not usually add new migration (for new entity or such relation, need migration) so chances are low and that may be the cause it's not considered something important to be back ported . However I think even just for the sake of keeping an identical behaviour at this early stage of these mechanism, we can/should back port .

I will ping Alec about it . 

BTW, if you think it's all ok, can you approved the PR . Thanks

bozana (approved):

kaitlinnewson (approved):

Issue #12593, opened 2026-04-21
Migration for the discussion-related email templates


Vitaliy-1:
**Description:**
Migrate discussion-related email templates to the discussion templates. This includes `DISCUSSION_NOTIFICATION_SUBMISSION`, `DISCUSSION_NOTIFICATION_REVIEW`, `DISCUSSION_NOTIFICATION_COPYEDITING`, `DISCUSSION_NOTIFICATION_PRODUCTION` and alternate templates. Correspondent mailable should be removed (`DiscussionSubmission`...)

**Steps to test the changes:**
These steps can include user interactions, database queries and API requests
1. Install the application or upgrade to 3.6
2. Go the task/discussion templates
3. Correspondent templates should be present

**What application are you using?**
  OJS, OMP and OPS main

Vitaliy-1:
The changes include

1. Migration of the default discussion-related templates
2. Migration of the customized discussion-related templates (those, which override the default ones and newly created)
3. Migration of the user group permissions

PRs

pkp-lib: https://github.com/pkp/pkp-lib/pull/12842
OJS: https://github.com/pkp/ojs/pull/5567
OMP: https://github.com/pkp/omp/pull/2423
OPS: https://github.com/pkp/ops/pull/1365

Vitaliy-1:
@touhidurabir, could you review those?

touhidurabir:
@Vitaliy-1 review added at https://github.com/pkp/pkp-lib/pull/12842#pullrequestreview-4420790289 and https://github.com/pkp/ojs/pull/5567#pullrequestreview-4421643243

Vitaliy-1:
@touhidurabir, can you take a second look? I finally made it work)

touhidurabir:
@Vitaliy-1 review added at https://github.com/pkp/pkp-lib/pull/12842#pullrequestreview-4643107535 .

Vitaliy-1:
@touhidurabir, do you have time for another review round? I've added installation for the default discussion templates.

touhidurabir:
@Vitaliy-1 review added at https://github.com/pkp/pkp-lib/pull/12842#pullrequestreview-4872532968, approved the OJS/OMP/OPS PRs .

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/13385
tests: https://github.com/pkp/ojs/pull/5858

jardakotesovec:
[...]

PR #12842, opened 2026-06-01
pkp/pkp-lib#12593 Migration for discussion-related templates

Vitaliy-1:


classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
```diff
@@ -0,0 +1,277 @@
...
+
+use APP\core\Application;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Query\Builder;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use PKP\editorialTask\enums\EditorialTaskType;
+
+class I12593_EmailToTaskTemplates extends Migration
```
touhidurabir, line 29: Our usual pattern for migration is to extend the `PKP\migration\Migration` but here we are extending the core `Illuminate\Database\Migrations\Migration`

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
```diff
@@ -0,0 +1,277 @@
...
+
+
+    /**
+     * Reverse the migration.
+     * Downgrade is not supported because the original email template rows have been deleted.
+     */
+    public function down(): void
+    {
+        // reverse the migration of the default data using one insert operation
+        if ($this->defaultData->isNotEmpty()) {
```
touhidurabir: without a way to re-setting these props, it will work at the time up full upgrade process but will fail targeted migration using the CLI tool . Also the doc block do say `Downgrade is not supported ...`, so not sure of the intention . 

Also as the up is also have destructive queries (e.g. the delete ones), do we need to have a down/recovery or we can just say it is not possible to downgrade/rollback .

touhidurabir (changes requested):
@Vitaliy-1 few comments . BTW, the tests are failing with errors .

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: Oops

Vitaliy-1:
@touhidurabir, Thanks! Yeah tests are failing and there are some unresolved workflow logic problems. Trying to figure them out  together with Jarda and Devika

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: > without a way to re-setting these props, it will work at the time up full upgrade process but will fail targeted migration using the CLI tool 

We don't have any pattern for that, right?

touhidurabir: > We don't have any pattern for that, right?

that's correct . But my concern is related to the CLI usage . if the revert command as `php lib/pkp/tools/migration.php "PKP\migration\upgrade\v3_6_0\I12593_EmailToTaskTemplates" down` can not properly rollback it, perhaps better to make it non downgradable ?

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -95,37 +95,33 @@ public function __construct($itemId, $itemType, $stageId, $template = null)
...
+                ->pluck((new UserGroup())->getPrimaryKeyName())
+                ->toArray();
+
+            $collector = Template::withContextId($context->getId())
+                ->withStageId($this->_stageId)
+                ->withType(EditorialTaskType::DISCUSSION->value);
+
+            $user->hasRole([Role::ROLE_ID_SITE_ADMIN, Role::ROLE_ID_MANAGER], $context->getId()) ?
+                $templates = $collector->get() :
+                $templates = $collector->withUserGroupsAccess($userGroupIds)->get();
```
touhidurabir, line 115: this called as `withUserGroupsAccess` but method registered as `scopeWithUserGroupAccess` , e.g `Group` vs `Groups` . this will cause crash .

classes/mail/mailables/DiscussionCopyediting.php
touhidurabir, line 1: These `Discussion*` mailables are deleted but still used by `I5716_EmailTemplateAssignments` which used for 3.4 upgrade . So a 3.3 to 3.6 direct upgrade will fail when trying to run the 3.4 upgrade migrations .

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
```diff
@@ -0,0 +1,248 @@
...
+        // Get primary locales of availables contexts
+        $contextDao = Application::getContextDAO();
+        $contextLocales = DB::table($contextDao->tableName)->get([$contextDao->primaryKeyColumn, 'primary_locale'])
+            ->mapWithKeys(fn (\stdClass $row) => [$row->{$contextDao->primaryKeyColumn} => $row->primary_locale]);
+
+        $this->taskTemplateData->each(function ($template) use ($contextLocales) {
+            DB::table('edit_task_template_settings')->insert([
+                [
+                    'edit_task_template_id' => $template->edit_task_template_id,
+                    'locale' => $contextLocales->get($template->context_id),
```
touhidurabir, line 280: here in each loop, we called `$template->context_id` which coming from above `$this->taskTemplateData` but not selected the `context_id` which should return a null . this may/will cause issue with data migration .

classes/editorialTask/Template.php
```diff
@@ -131,19 +150,35 @@ public function scopeOrderByPkDesc(Builder $query): Builder
...
-    public function scopeFilterByInclude(Builder $query, bool $include): Builder
+    public function scopeWithInclude(Builder $query, bool $include): Builder
     {
         return $query->where('include', $include);
     }
 
+    /**
+     * Identify which user groups have access to the template
+     */
+    public function scopeWithUserGroupAccess(Builder $query, array $userGroupIds): Builder
```
touhidurabir: The naming mismatch of `Group` vs `Groups` can/will cause confusion and crash . Also in the dock block of `@method` above .

classes/editorialTask/Template.php
```diff
@@ -131,19 +150,35 @@ public function scopeOrderByPkDesc(Builder $query): Builder
...
+    {
+        return $query->when(
+            !empty($userGroupIds),
+            fn ($query) =>
+            $query->whereHas(
+                'userGroups',
+                fn ($subQuery) =>
+                $subQuery->whereIn('user_group_id', $userGroupIds)
+            )->orWhere('restrict_to_user_groups', false)
```
touhidurabir: this will cause an issue as `orWhere` not grouped and run independently . Best to warp in a `where` , both the `whereIn` and `orWhere` to be fully safe .

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -95,37 +95,33 @@ public function __construct($itemId, $itemType, $stageId, $template = null)
...
+            $templateData = $templates->mapWithKeys(
+                fn (Template $template, int $key) =>
+                [$template->id => $template->getLocalizedData('title')]
+            );
         }
 
         $templateMgr = TemplateManager::getManager($request);
         $templateMgr->assign([
-            'templates' => $templates,
+            'templates' => $templateData->toArray() ?? [],
```
touhidurabir: this var `$templateData` can be unavailable if the above `if` condition does not satisfy , better to have a check

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -222,7 +220,17 @@ public function sendMessage(int $userId, Submission $submission, Request $reques
...
-        $additionalVariables = $this->getEmailVariableNames($template->getData('key'));
+        $templateKey = '';
+        foreach (EditorialTask::getTitleLocalizedStrings($context) as $key => $map) {
+            foreach ($map as $localizedTitle) {
+                if ($template->getLocalizedData('title') == $localizedTitle) {
+                    $templateKey = $key;
+                    break 2;
+                }
+            }
+        }
```
touhidurabir: it seems we have almost identical code block added in the `StageParticipantGridHandler` also , so perhaps we can centralise it in repo class ?

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
```diff
@@ -222,7 +220,17 @@ public function sendMessage(int $userId, Submission $submission, Request $reques
...
+
+        $additionalVariables = $this->getEmailVariableNames($templateKey);
```
touhidurabir, line 224: having a bit hard time to understand the logic there, so we are determining a email key based on a mapping of deprecated keys with correspondent localized name string (in the `EditorialTask::getTitleLocalizedStrings()`). So any chance template renamed by admin or locale string not being available as that may cause not found template variable ?

not 100% sure of the intention so ignore if this is not a concern .

classes/editorialTask/EditorialTask.php
```diff
@@ -496,4 +508,30 @@ protected function anonymizeAuthors(Submission $submission): bool
             ->filterByReviewMethods([ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS])
             ->getCount() > 0;
     }
+
+    /**
+     * Get the map of deprecated discussion related email template keys with correspondent localized name string
+     */
+    public static function getTitleLocalizedStrings(Context $context): array
```
touhidurabir: just wondering, perhaps it's better to move the static method to repo instead of keeping in model class as we have a repo class for editorial task ?

classes/editorialTask/EditorialTask.php
```diff
@@ -496,4 +508,30 @@ protected function anonymizeAuthors(Submission $submission): bool
...
+    {
+        $supportedLocales = $context->getSupportedLocales();
+
+        $map = [
+            'EDITOR_ASSIGN' => 'mailable.editorAssignedManual.name',
+            'COPYEDIT_REQUEST' => 'mailable.copyeditRequest.name',
+            'LAYOUT_REQUEST' => 'mailable.layoutRequest.name',
+            'LAYOUT_COMPLETE' => 'mailable.layoutComplete.name',
+            'INDEX_REQUEST' => 'mailable.indexRequest.name',
+            'INDEX_COMPLETE' => 'mailable.indexComplete.name',
```
touhidurabir: seems exactly these 2 locales are not registered anywhere , will it cause issue ?

classes/editorialTask/EditorialTask.php
```diff
@@ -496,4 +508,30 @@ protected function anonymizeAuthors(Submission $submission): bool
...
+     */
+    public static function getTitleLocalizedStrings(Context $context): array
+    {
+        $supportedLocales = $context->getSupportedLocales();
+
+        $map = [
+            'EDITOR_ASSIGN' => 'mailable.editorAssignedManual.name',
+            'COPYEDIT_REQUEST' => 'mailable.copyeditRequest.name',
+            'LAYOUT_REQUEST' => 'mailable.layoutRequest.name',
+            'LAYOUT_COMPLETE' => 'mailable.layoutComplete.name',
```
touhidurabir: another point related to https://github.com/pkp/pkp-lib/pull/12842/changes#r3535125212 , `edit_task_templates` already has an `email_template_key` column which seems to build for this purpose in migration of `I11779_UseEmailTemplateKeyOnEditTaskTemplates` but our current migration does not seems to populate that and rely on the matching translated titles . 

I am not aware of intended plans so skip this comment if that is the intended plan here or perhaps, I am misunderstanding something here .

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
```diff
@@ -0,0 +1,248 @@
...
+                    ]);
+                }
+            }
+        }
+
+        // Delete custom templates, related settings and access
+        DB::table('email_templates')->whereIn('email_id', $customIds)->delete();
+        DB::table('email_templates_settings')->whereIn('email_id', $customIds)->delete();
+        DB::table('email_templates_default_data')->whereIn('email_key', $this->getEmailKeys())->delete();
+        DB::table('email_template_user_group_access')->whereIn('email_key', $this->getEmailKeys())->delete();
```
touhidurabir: it seems we are migrating data from other 3 table before deleting that but none from the `email_template_user_group_access` , is that intentional ?

touhidurabir (changes requested):
@Vitaliy-1 added review . mostly questions but few are important like scope method naming, 3.4 upgrade migration modification and sql where/orWhere grouping .

classes/mail/mailables/DiscussionCopyediting.php
Vitaliy-1: But isn't `I5716_EmailTemplateAssignments` executed first and then only templates removal?

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: I afraid I don't understand. Context ID isn't nullable, thus should be always available, right?

touhidurabir: > Context ID isn't nullable, thus should be always available, right?

That is correct when we are not fetching specific selectable columns as 
```php
$this->taskTemplateData = DB::table('edit_task_templates')->get(['edit_task_template_id', 'title', 'description']);
```

so in the list of selectable columns as `['edit_task_template_id', 'title', 'description']` `context_id` is not available and we will not get a hydrated eloquent model object but simple stdObject collection with only object only contains given selected ones as `edit_task_template_id', 'title', 'description'` .

Vitaliy-1: Yes, thanks!

controllers/grid/users/stageParticipant/form/PKPStageParticipantNotifyForm.php
Vitaliy-1: If template is renamed, it will cause an issue in determining the right Notification type. And that requires refactoring in the future. I'll remove the part with variables, it's just an artefact.

classes/editorialTask/EditorialTask.php
Vitaliy-1: Should be used in OMP

classes/editorialTask/EditorialTask.php
Vitaliy-1: It was removed. Another approach could be to use settings. But the whole logic in determining the Notification type based on the template is flawed

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: That should be removed but I'll leave that for Taslan

classes/mail/mailables/DiscussionCopyediting.php
touhidurabir: The `APP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments` used for upgrade block `minversion="3.1.0.0" maxversion="3.3.9.9"` . Now the APP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments::getDiscussionTemplates() as following implementation as 
```php
protected function getDiscussionTemplates(): Collection
{
    return collect([
        DiscussionSubmission::getEmailTemplateKey(),
        DiscussionReview::getEmailTemplateKey(),
        DiscussionCopyediting::getEmailTemplateKey(),
        DiscussionProduction::getEmailTemplateKey(),
    ]);
}
```

But this PR deletes those `Discussion*` classes itself . So a 3.3 installation copying the codebase of 3.6 and run upgrade will try to run the migration `APP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments` and will fail/crash with fatal error , so that way seems to me . Or am I misunderstanding something there . 

However the upgrade tests points to this same issue as 
```
PHP Fatal error:  Uncaught Error: Class "PKP\mail\mailables\DiscussionSubmission" not found in /home/runner/ojs/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php:43
Stack trace:
#0 /home/runner/ojs/lib/pkp/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php(150): APP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments->getDiscussionTemplates()
#1 /home/runner/ojs/lib/pkp/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php(43): PKP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments->migrateNotificationCenterTemplates()
#2 /home/runner/ojs/lib/pkp/classes/install/Installer.php(437): PKP\migration\upgrade\v3_4_0\I5716_EmailTemplateAssignments->up()
#3 /home/runner/ojs/lib/pkp/classes/install/Installer.php(299): PKP\install\Installer->executeAction()
#4 /home/runner/ojs/lib/pkp/classes/install/Installer.php(208): PKP\install\Installer->executeInstaller()
#5 /home/runner/ojs/lib/pkp/classes/cliTool/UpgradeTool.php(96): PKP\install\Installer->execute()
#6 /home/runner/ojs/lib/pkp/classes/cliTool/UpgradeTool.php(69): PKP\cliTool\UpgradeTool->upgrade()
#7 /home/runner/ojs/tools/upgrade.php(22): PKP\cliTool\UpgradeTool->execute()
#8 {main}
  thrown in /home/runner/ojs/classes/migration/upgrade/v3_4_0/I5716_EmailTemplateAssignments.php on line 43
```
for mysql/postgresql when upgrading form 3.3 . Only mariadb upgrade test is passing as it's upgrading from 3.4 .

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
```diff
@@ -0,0 +1,267 @@
...
+        return array_keys($this->emailKeyToStageMap());
+    }
+
+    /**
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        Schema::table('edit_task_templates', function (Blueprint $table) {
+            $table->boolean('default')->default(false);
```
touhidurabir: will not cause any issue but in upgrade we set it to true below as `DB::table('edit_task_templates')->insertGetId` but adding a new context via `PKP\editorialTask\Repository::installTaskTemplates()` does not set to true and stay default false . Not sure about the intention here . 

Ignore if this is intended behaviour .

classes/editorialTask/Template.php
```diff
@@ -131,19 +151,38 @@ public function scopeOrderByPkDesc(Builder $query): Builder
...
-    public function scopeFilterByInclude(Builder $query, bool $include): Builder
+    public function scopeWithInclude(Builder $query, bool $include): Builder
     {
         return $query->where('include', $include);
     }
 
+    /**
+     * Identify which user groups have access to the template
+     */
+    public function scopeWithUserGroupsAccess(Builder $query, array $userGroupIds): Builder
```
touhidurabir, line 172: this is better now, but seems we have something similar as `scopeWithUserGroupIds` but with diff of not grouping the where/orWhere . do we need both ?

api/v1/contexts/PKPContextController.php
```diff
@@ -351,6 +352,8 @@ public function add(Request $illuminateRequest): JsonResponse
         $context = Application::getContextDAO()->newDataObject();
         $context->setAllData($params);
         $context = $contextService->add($context, $request);
+        Repo::editorialTask()->installTaskTemplates($context);
```
touhidurabir, line 355: this install the templates when creating a new context . But if a new locale added/removed, should we add/remove from the `PKP\i18n\Locale::installLocale()/uninstallLocale()` ?

touhidurabir (commented):
@Vitaliy-1 just few comments, but the new locale add/remove need a double check .

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: Thanks, good catch, that would definitely lead to a bug

api/v1/contexts/PKPContextController.php
Vitaliy-1: I was thinking about this. I decided to install always in all available site locales

touhidurabir: > decided to install always in all available site locales

then are we adding/removing when site locales updated (new added or existing removed) ?

classes/migration/upgrade/v3_6_0/I12593_EmailToTaskTemplates.php
Vitaliy-1: I've decided to remote default column. It's better just to move `key` to the primary column and make it mandatory for default templates.

Issue #11900, opened 2025-10-06
Exception while attempting to delete a submission


jonasraoni:
**Description:**
When attempting to delete a submission, the following exception is raised:

```php
PDOException: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`ore`.`publication_galleys`, CONSTRAINT `publication_galleys_submission_file_id_foreign` FOREIGN KEY (`submission_file_id`) REFERENCES `submission_files` (`submission_file_id`)) in /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php:596
--
Stack trace:
#0 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(596): PDOStatement->execute()
#1 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(812): Illuminate\Database\Connection->Illuminate\Database\{closure}()
#2 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(779): Illuminate\Database\Connection->runQueryCallback()
#3 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(584): Illuminate\Database\Connection->run()
#4 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Connection.php(548): Illuminate\Database\Connection->affectingStatement()
#5 /var/www/html/lib/pkp/lib/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(4024): Illuminate\Database\Connection->delete()
#6 /var/www/html/lib/pkp/classes/core/EntityDAO.php(233): Illuminate\Database\Query\Builder->delete()
#7 /var/www/html/lib/pkp/classes/submissionFile/DAO.php(238): PKP\core\EntityDAO->deleteById()
#8 /var/www/html/lib/pkp/classes/core/EntityDAO.php(221): PKP\submissionFile\DAO->deleteById()
#9 /var/www/html/lib/pkp/classes/submissionFile/DAO.php(219): PKP\core\EntityDAO->_delete()
#10 /var/www/html/lib/pkp/classes/submissionFile/Repository.php(488): PKP\submissionFile\DAO->delete()
[...]

PR #12846, opened 2026-06-01
pkp/pkp-lib#11990 Do not delete submission files when other galleys may refer

asmecher:


asmecher:
(Mis-tagged the issue in the commit; it should be https://github.com/pkp/pkp-lib/issues/11900)

Issue #12841, opened 2026-06-01
Add infrastructural support of laravel logging


touhidurabir:
Add the infrastructural support of [Laravel Logging](https://laravel.com/docs/12.x/logging) . 

related discussion at https://github.com/pkp/pkp-lib/discussions/7966

### Impacts and details 
 - default log channel is daily rotating log files with 30 days retention period by default . configurable .
 - if log channel set to `errorlog`, fatal error/exception will not be logged in PHP error log file to make sure no double logging happens
 - by default logs will be written in default laravel log format but support monolog formats like JOSN by setting `log_formatter` .

### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12848
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5568
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2402
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1342

touhidurabir:
@asmecher PRs at https://github.com/pkp/pkp-lib/issues/12841#issue-4562578131 ready for review . Once we finalise the PR for OJS, we can forward port it to OMP/OPS .

asmecher:
Thanks, @touhidurabir! I've added reviews:
- https://github.com/pkp/pkp-lib/pull/12848#pullrequestreview-4454230360
- https://github.com/pkp/ojs/pull/5568#pullrequestreview-4450722316

asmecher:
@touhidurabir, checking back on our original goals -- do the configuration options we're adding to `config.inc.php` allow us to configure a Monolog transport that'll e.g. send logs to a remote service? I think not at the moment.

touhidurabir:
> [@touhidurabir](https://github.com/touhidurabir), checking back on our original goals -- do the configuration options we're adding to `config.inc.php` allow us to configure a Monolog transport that'll e.g. send logs to a remote service? I think not at the moment.

@asmecher  that will be bit difficult as that would require custom handler, processor and formatter . Or compatible packages (e.g. service like sentry, logstash has laravel compatible package which provide out of box custom stuff, like handler, processor etc) . 

Probably still possible but it will be messy way to facilitate directly via the config.inc.php . cleaner approach IMO would be to implement external service via plugin and override the config.

asmecher:
[...]

Issue #12237, opened 2026-01-23
Support of laravel logging and log viewer


touhidurabir:
related discussion at https://github.com/pkp/pkp-lib/discussions/7966
log viewer package https://github.com/opcodesio/log-viewer 

### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12238
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5286

touhidurabir:
@asmecher can you take at look at the PRs at https://github.com/pkp/pkp-lib/issues/12237#issue-3846598611 . 

It has turned out better than I had initially thought and the initial issues faced. Big win is that no major modification needed to existing routing system .

touhidurabir:
@asmecher This is now at better shape . Only one thing we need a good resolution is how and from where we should serve the package assets . for now, I followed the laravel app conventions and make it served form our public folder but not sure if that is the best approach as that make it a git tracked files . Or we can make it as part of installation process to auto copy or have it as process of upgrade process .

PR #12848, opened 2026-06-02
pkp/pkp-lib#12841 Infrastructural support of laravel logging

touhidurabir:
for pkp/pkp-lib#12841 .

### DIFF
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -21,6 +21,7 @@ use DateTime;
 use DateTimeZone;
 use Exception;
 use GuzzleHttp\Client;
+use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Database\MariaDbConnection;
 use Illuminate\Database\MySqlConnection;
@@ -359,7 +360,9 @@ abstract class PKPApplication implements PKPApplicationInfoProvider
                 $dispatcher->dispatch($this->getRequest());
             } catch (\Throwable $t) {
                 if (Hook::run('PKPApplication::execute::catch', ['throwable' => $t]) !== Hook::ABORT) {
-                    // No hook handler took ownership; throw again
+                    // No hook handler took ownership; report through Laravel's
+                    // exception handler so it lands in the configured log channel.
+                    app(ExceptionHandler::class)->report($t);
                     throw $t;
                 }
             }
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -365,7 +365,7 @@ abstract class PKPApplication implements PKPApplicationInfoProvider
                     // When PHP's error_log is an active channel, the re-thrown exception
                     // below is already recorded there by PHP's native fatal handler, so reporting
                     // it here too would duplicate the entry, so skip logging in PHP's error_log
-                    if (!app()->usesErrorLogChannel()) {
+                    if (!PKPExceptionHandler::usesErrorLogChannel()) {
                         app(ExceptionHandler::class)->report($t);
                     }
                     throw $t;
--- a/classes/core/PKPApplication.php
+++ b/classes/core/PKPApplication.php
@@ -362,7 +362,12 @@ abstract class PKPApplication implements PKPApplicationInfoProvider
                 if (Hook::run('PKPApplication::execute::catch', ['throwable' => $t]) !== Hook::ABORT) {
                     // No hook handler took ownership; report through Laravel's
                     // exception handler so it lands in the configured log channel.
-                    app(ExceptionHandler::class)->report($t);
+                    // When PHP's error_log is an active channel, the re-thrown exception
+                    // below is already recorded there by PHP's native fatal handler, so reporting
+                    // it here too would duplicate the entry, so skip logging in PHP's error_log
+                    if (!app()->usesErrorLogChannel()) {
+                        app(ExceptionHandler::class)->report($t);
+                    }
                     throw $t;
                 }
             }
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -38,6 +38,11 @@ use Throwable;
 class PKPContainer extends Container
 {
     /**
+     * The directory name, under the storage (files) directory, where application logs are written
+     */
+    public const LOG_DIRECTORY = 'logs';
+
+    /**
      * Define if the app currently runing the unit test
      */
     private bool $isRunningUnitTest = false;
@@ -490,27 +495,33 @@ class PKPContainer extends Container
 
         // Logging
         $logLevel = Config::getVar('logs', 'log_level', 'error');
+        $stackChannels = Str::of(Config::getVar('logs', 'log_stacks', 'daily,errorlog'))
+            ->explode(',')
+            ->map(fn (string $channel) => trim($channel))
+            ->filter()
+            ->values()
+            ->all();
         $items['logging'] = [
             'default' => Config::getVar('logs', 'log_channel', 'stack'),
             'channels' => [
                 'stack' => [
                     'driver' => 'stack',
-                    'channels' => ['daily', 'errorlog'],
+                    'channels' => $stackChannels ?: ['daily', 'errorlog'],
                     'ignore_exceptions' => false,
                 ],
 
                 'single' => [
                     'driver' => 'single',
-                    'path' => storage_path('logs/' . $this->logFileName()),
+                    'path' => $this->logFilePath(),
                     'level' => $logLevel,
                     'replace_placeholders' => true,
                 ],
 
                 'daily' => [
                     'driver' => 'daily',
-                    'path' => storage_path('logs/' . $this->logFileName()),
+                    'path' => $this->logFilePath(),
                     'level' => $logLevel,
-                    'days' => 30, // The number of previour log files that will be retained
+                    'days' => (int) Config::getVar('logs', 'log_daily_days', 30), // The number of previous log files that will be retained
                     'replace_placeholders' => true,
                 ],
 
@@ -544,7 +555,7 @@ class PKPContainer extends Container
                 ],
 
                 'emergency' => [
-                    'path' => storage_path('logs/' . $this->logFileName()),
+                    'path' => $this->logFilePath(),
                 ],
             ],
         ];
@@ -680,6 +691,14 @@ class PKPContainer extends Container
     public function logFileName(): string
     {
         return Config::getVar('logs', 'log_file_name', Application::getName() . '.log');
+    }
+
+    /**
+     * Get the absolute path to the application log file under the storage log directory.
+     */
+    public function logFilePath(): string
+    {
+        return $this->storagePath(self::LOG_DIRECTORY . '/' . $this->logFileName());
     }
 
     /**
--- a/classes/core/PKPContainer.php
+++ b/classes/core/PKPContainer.php
@@ -577,7 +577,10 @@ class PKPContainer extends Container
                 'orcid' => [
                     'driver' => 'single',
                     'path' => $this->logFilePath(\PKP\orcid\OrcidManager::LOG_FILE),
-                    'level' => 'debug',
+                    'level' => 'debug', // intentionally to debug and let ORCID Manager filter out unwanted levels
+                    'replace_placeholders' => true,
+                    'formatter' => $logFormatter,
+                    'formatter_with' => ['includeStacktraces' => true],
                 ],
             ],
         ];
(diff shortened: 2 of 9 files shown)


classes/orcid/OrcidManager.php
```diff
@@ -335,17 +337,15 @@ public static function logError(string $message): void
...
     }
 
     /**
      * Helper method to write log message out to the configured log file.
      */
     private static function writeLog(string $message, string $level): void
     {
-        $fineStamp = date('Y-m-d H:i:s') . substr(microtime(), 1, 4);
-        $logFilePath = Config::getVar('files', 'files_dir') . '/orcid.log';
-        error_log("{$fineStamp} {$level} {$message}\n", 3, $logFilePath);
```
touhidurabir, line 348: this will make the current exiting orcid log redundant . We can move/use the same one but any previous exiting logs does not follow any format and make the log file un-ingest able/non-parseable by any service/packages .

classes/core/PKPApplication.php
```diff
@@ -359,7 +360,9 @@ public function execute(): void
                 $dispatcher->dispatch($this->getRequest());
             } catch (\Throwable $t) {
                 if (Hook::run('PKPApplication::execute::catch', ['throwable' => $t]) !== Hook::ABORT) {
-                    // No hook handler took ownership; throw again
+                    // No hook handler took ownership; report through Laravel's
+                    // exception handler so it lands in the configured log channel.
+                    app(ExceptionHandler::class)->report($t);
                     throw $t;
```
asmecher: Is calling `report` followed by throwing the exception again going to result in double processing / double logging?

classes/core/PKPContainer.php
```diff
@@ -514,9 +483,94 @@ protected function loadConfiguration(): void
...
+            ->values()
+            ->all();
+
+        // Optional Monolog formatter for the formatter-capable channels (single, daily,
+        // stderr, syslog). A null formatter falls back to Laravel's default LineFormatter
+        $logFormatter = Config::getVar('logs', 'log_formatter');
+        $logFormatter = ($logFormatter && class_exists($logFormatter)) ? $logFormatter : null;
+
+        $items['logging'] = [
+            'default' => Config::getVar('logs', 'log_channel', 'stack'),
```
asmecher: Would `syslog` rather than `stack` be a more backwards-compatible target than `stack` for those moving up from earlier versions of the software?

classes/core/PKPContainer.php
```diff
@@ -514,9 +483,94 @@ protected function loadConfiguration(): void
...
+                // ORCID writes to its own file; OrcidManager::getLogLevel() does the
+                // filtering, so this channel passes everything (level => debug).
+                'orcid' => [
+                    'driver' => 'single',
+                    'path' => $this->logFilePath(\PKP\orcid\OrcidManager::LOG_FILE),
+                    'level' => 'debug', // intentionally to debug and let ORCID Manager filter out unwanted levels
+                    'replace_placeholders' => true,
+                    'formatter' => $logFormatter,
+                    'formatter_with' => ['includeStacktraces' => true],
+                ],
```
asmecher: I'm not sure a special case for the ORCiD logging is warranted; it should probably be treated the same way as the rest of the system. @ewhanson, opinions on this?

asmecher (commented):
Thanks, @touhidurabir! Just a few questions...

classes/core/PKPApplication.php
touhidurabir: Not exactly, for any PHP app, an fatal error/exception should be registered in the default error_log . the report log in specific format to log  channel . these 2 are there for 2 different purpose. 

However, if channel configured as errorlog or in include as part of stack, it will be also logged in PHP's native error_log, but thats configurable option .

classes/core/PKPContainer.php
touhidurabir: I dont think so and in fact, that may be more confusing . as `syslog channel` → `Monolog SyslogHandler` → writes to the OS syslog daemon via PHP's `syslog()/openlog()`. That syslog also used by OS for other internal logging purpose and having it a default log channel will result in having a log file with many different format which will be hard to parse and ingest . 

basically Pre-3.6 OJS had no Laravel logging layer; fatals/exceptions surfaced through PHP's native path, i.e. error_log() (plus the old scheduled-task/error-handler writes). So the channel that preserves legacy behavior is errorlog. but using the errorlog as default make no sense as any fatal error/exception will be auto logged there by default and make it hard to parse/ingest for specific log format .

classes/core/PKPContainer.php
touhidurabir: Orcid have a custom log file implemenation so I added it here . if not needed, we can remove it and use the default channel .

classes/core/PKPApplication.php
asmecher: I'm not sure I follow 100%... If the logging configuration in `config.inc.php` is `errorlog`, will it be recorded twice?

classes/core/PKPContainer.php
asmecher: Sorry, I mean `errorlog`, not `syslog`. Generally when adding new behaviours it makes sense to preserve the old behaviour as a default in the code, so if someone doesn't add new configuration options during upgrade, the behaviour will stay the same.

touhidurabir: > I mean errorlog, not syslog

that we can do but my suggestion will be not to do that . By default any fatal error/exception will be logged as usual in the PHP's `error_log` . Other logging should be logged where it would log in a consistent format/structure which is easily parseable and can be ingested . If we keep allow to do it by default, then we will be having multiple format/structure in PHP's default error_log . This IMO kind of beat the purpose of logging infrastructure . any call to `error_log()` will do it's logging to PHP's default error log and `Log::` should log to channel that should only used for logging .

classes/core/PKPApplication.php
touhidurabir: > If the logging configuration in config.inc.php is errorlog, will it be recorded twice?

yes but in 2 different format as 

use of `Log::`
```
[09-Jun-2026 04:08:04 UTC] [2026-06-09 04:08:04] production.ERROR: testing testing {"exception":"[object] (Exception(code: 0): testing testing at /Users/abir/Sites/code/ojs-main/lib/pkp/pages/admin/AdminHandler.php:169)
[stacktrace]
#0 [internal function]: PKP\\pages\\admin\\AdminHandler->index(Array, Object(APP\\core\\Request))
#1 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPRouter.php(328): call_user_func(Array, Array, Object(APP\\core\\Request))
#2 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPPageRouter.php(249): PKP\\core\\PKPRouter->_authorizeInitializeAndCallRequest(Array, Object(APP\\core\\Request), Array, false)
#3 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/Dispatcher.php(157): PKP\\core\\PKPPageRouter->route(Object(APP\\core\\Request))
#4 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPApplication.php(360): PKP\\core\\Dispatcher->dispatch(Object(APP\\core\\Request))
#5 /Users/abir/Sites/code/ojs-main/index.php(21): PKP\\core\\PKPApplication->execute()
#6 /Users/abir/.composer/vendor/laravel/valet/server.php(110): require('/Users/abir/Sit...')
#7 {main}
"} 
```

default logging of a fatal error/exception
```
[09-Jun-2026 04:08:04 UTC] PHP Fatal error:  Uncaught Exception: testing testing in /Users/abir/Sites/code/ojs-main/lib/pkp/pages/admin/AdminHandler.php:169
Stack trace:
#0 [internal function]: PKP\pages\admin\AdminHandler->index(Array, Object(APP\core\Request))
#1 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPRouter.php(328): call_user_func(Array, Array, Object(APP\core\Request))
#2 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPPageRouter.php(249): PKP\core\PKPRouter->_authorizeInitializeAndCallRequest(Array, Object(APP\core\Request), Array, false)
#3 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/Dispatcher.php(157): PKP\core\PKPPageRouter->route(Object(APP\core\Request))
#4 /Users/abir/Sites/code/ojs-main/lib/pkp/classes/core/PKPApplication.php(360): PKP\core\Dispatcher->dispatch(Object(APP\core\Request))
#5 /Users/abir/Sites/code/ojs-main/index.php(21): PKP\core\PKPApplication->execute()
#6 /Users/abir/.composer/vendor/laravel/valet/server.php(110): require('/Users/abir/Sit...')
#7 {main}
  thrown in /Users/abir/Sites/code/ojs-main/lib/pkp/pages/admin/AdminHandler.php on line 169
```

and thats expected as thats how config will be set by the sysadmin . 

we can add a check easily there and not report via `app(ExceptionHandler::class)->report($t);` but that can also be confusing as there would be no monolog compatible log for fatal error/exception but log for other cases in monolog compatible format . 

This is same for a vanilla laravel app also .

asmecher: @touhidurabir, I'm thinking about the large set of OJS users who already know where to find their existing default PHP error log and don't want a change when they upgrade to 3.6. They won't expect errors to be logged there twice -- they already have a hard time wading through the high volume of messages there.

touhidurabir: added a check which would prevent logging the log when errorlog set as channel .

classes/core/PKPContainer.php
asmecher: @ewhanson, can you confirm whether this is necessary?

classes/core/PKPContainer.php
```diff
@@ -547,7 +603,6 @@ protected function loadConfiguration(): void
         ];
         $items['mail']['mailers']['log'] = [
             'transport' => 'log',
-            'channel' => 'errorlog',
```
touhidurabir, line 550: By removing this one one, using `log` as mail driver will respect the logs setting of default log channel . so right now, if we set the default log channel to single, it will log to the log . However setting mail driver to `log` will still be logging to the php `error_log` as defined hardcoded .

classes/core/PKPContainer.php
asmecher: @touhidurabir I think I'll just go ahead with the executive decision that the ORCiD logging does not warrant special treatment. Please change its logging behaviour to be consistent with the rest of the PR.

classes/core/PKPContainer.php
```diff
@@ -526,9 +500,91 @@ protected function loadConfiguration(): void
...
+        $logFormatter = ($logFormatter && class_exists($logFormatter)) ? $logFormatter : null;
+
+        $items['logging'] = [
+            'default' => Config::getVar('logs', 'log_channel', 'stack'),
+            'channels' => [
+                'stack' => [
+                    'driver' => 'stack',
+                    'channels' => array_values(array_filter(array_map(
+                        'trim',
+                        explode(',', Config::getVar('logs', 'log_stacks', 'single'))
```
asmecher: With `single` being the default both here and in `config.TEMPLATE.inc.php`, won't this cause most of the community to have a log file that grows without bounds? Only the `daily` log supports rotation.

classes/orcid/OrcidManager.php
asmecher: As [here](https://github.com/pkp/pkp-lib/pull/12848/changes#r3552921005), let's not preserve the special behaviour in the ORCiD code; I don't think there's any reason for it.

classes/core/PKPContainer.php
```diff
@@ -621,6 +676,76 @@ public function path(string $path = ''): string
         return $this->basePath($path);
     }
 
+    /**
+     * Get the path to the storage directory
+     */
+    public function storagePath(string $path = ''): string
```
asmecher, line 669: Are these new functions (here and below) part of a normal Laravel container, or just for our internal purposes?

classes/core/PKPContainer.php
```diff
@@ -621,6 +676,76 @@ public function path(string $path = ''): string
...
+            'errorlog',
+            array_map('trim', explode(',', Config::getVar('logs', 'log_stacks', 'single'))),
+            true
+        );
+    }
+
+    /**
+     * Get the current locale
+     */
+    public function getLocale(): string
```
asmecher: Discussed in 1:1: Maybe this can be removed?

classes/core/PKPContainer.php
asmecher: Discussed in 1:1: Let's distinguish between Laravel container extensions/overrides and our own functions using e.g. `@see` self-documentation referring to the parent implementation.

classes/orcid/OrcidManager.php
touhidurabir: Done

classes/core/PKPContainer.php
touhidurabir: removed .

classes/core/PKPContainer.php
touhidurabir: changed to `daily` as default .

classes/core/PKPContainer.php
touhidurabir: this is one we needed as logs are stored in the storage paths .

classes/core/PKPContainer.php
touhidurabir: removed this one and along with other that we dont need right now until we get back to https://github.com/pkp/pkp-lib/issues/12237 where we will have those to be needed .

PR #12849, opened 2026-06-02
Add Japanese localization fixes for PKP lib

yuk1-kondo:
## Summary

This PR updates Japanese localization coverage in PKP lib 3.5 and adds missing English source strings for common management UI labels that can otherwise appear as raw localization keys.

Changes include:

- Add missing English source strings for CAPTCHA labels, invitation/user access table headers, notification status messages, review reminder labels, discussion message mailables, and a submission author variable.
- Improve Japanese translations in shared PKP lib locale files where strings were empty or fell back to English.
- Add Japanese locale files for invitation and user access components.

## Motivation

While operating an OJS 3.5 Japanese installation, several shared PKP lib screens showed untranslated English text or raw localization keys such as `##...##`. These strings were visible in user access, invitations, review reminders, notification settings, and email-template related UI.

## Validation

- Parsed all changed `.po` files with `polib` successfully.
- Checked the translations against a live Japanese OJS 3.5 installation before preparing this upstream contribution.


asmecher:
Thanks for contributing, @yuk1-kondo!

It looks like this PR also adds some English text. At a glance, the keys I looked at aren't used in the code (though I did only a cursory search). How did you determine these were needed?

I also see some English text mixed in with the Japanese locale files; we try to avoid this when possible.

We generally recommend using Weblate for translation maintenance; see the [PKP Translation Guide](https://docs.pkp.sfu.ca/translating-guide/en/) for details. But we can manage this one as a pull request if that's easier.

yuk1-kondo:
Thanks for the quick and helpful review @asmecher!

You were right on both points. I went back and audited every English source key I had proposed adding here and in pkp/ojs#5569, and almost all of them are either obsolete in `stable-3_5_0` (already marked `#~` in upstream `.po` files) or no longer referenced anywhere in code (`*.tpl`, `*.js`, `*.php`). I originally identified them empirically because they were appearing as raw `##key##` strings in a live OJS 3.5 Japanese install, but that install had legacy plugins/templates still calling the old keys — not a problem in upstream `stable-3_5_0` itself.

I've force-pushed an updated branch that:

- Reverts every English source string addition (no changes to `locale/en/*.po` in either PR).
- Removes Japanese entries whose `msgid` is no longer in the 3.5 English source so gettext falls back cleanly.
- Replaces every Japanese `msgstr` that was still equal to the English source with an empty `msgstr` to avoid the English-in-JP mixing you flagged.

The result is a translation-only contribution. Happy to also move this to Weblate per the PKP Translation Guide if that's the preferred long-term path — please let me know if I should close this PR in favor of a Weblate upload.

asmecher (approved):

asmecher:
Merged -- thanks very much, @yuk1-kondo! I've also forward-ported this to the `main` branch, making a couple necessary adjustments.

At the moment our Weblate installation draws from `stable-3_5_0`, so these changes will automatically be updated there. If you don't mind, we'd prefer that translation changes were made there, so that they can be done a little more collaboratively and with less oversight in reviewing and merging individual PRs on our part.

Thanks for contributing!

Issue #12800, opened 2026-05-27
Tighten review round association with publications


ewhanson:
> [!IMPORTANT]
> **ORE VM manual migration step needed**
> The below code or equivalent will need to be run manually outside of a typical migration context on the ORE test VM.
> 
> ```php
> Schema::table('edit_decisions', function (Blueprint $table) {
>     $table->dropForeign('edit_decisions_publication_id');
>     $table->dropIndex('edit_decisions_publication_id');
>     $table->dropColumn('publication_id');
> });
>```
> e.g. drop foreign key constraint, index, and `publication_id` column on `edit_decisions` table.

With a post-publication review workflow where many published publications could have many different review rounds before a PMUR has been "accepted," review rounds need to be properly associated with a PMUR. This is already present in the code, but does not take into account the full flexibility of the post-publication review workflow.

This issue adds more flexibility in review round/publication creation while tightening requirements for review round/publication associations.

> [!note]
> Traditional pre-publication peer review workflows will be unaffected by these changes.

## Requirements
- PMURs require a  associated review round to be published
- Review rounds can now have a `null` publication association (previously required)
- Publication to have field-level validation check to ensure alignment with above requirements in
	- publication settings form
	- review publication details on pre-publication form
- When creating a new review round...
	- DO NOT create review round -> publication association for an already published publication version (unpublished publication versions unaffected by this)
		- E.g., review rounds created with only a published VoR 1.0 will have a `null` publication association.
	- DO associate review round with 1) latest publication that 2) has not been published (as opposed to defaulting to submission's current publication)
- WILL NOT need indicator to present unconnected review round
- DO automatically link one-off `null` publication association review rounds with newly created publication
	- If a new review round is created with no valid publication available, it will have a `null` association, then when a new publication is created, it associates the most recent `null` associated review round with the new publication

## Acceptance criteria
Acceptance criteria based on review round publication association in DB (`publication_id` column in `review_rounds` table).

NB: `RR = Review round` and `P = Publication`.

### Original criteria

- RR created with single unpublished P is associated with that P.
- RR created with single published P has `null` association.
- RR created with no valid Ps has `null` association is associated with newly created P version when P is created.
- P is created first followed by a new RR has RR associated with P.

### UI/Multiselect criteria
- If a user tries to publish a PMUR without an associated RR, they should be prompted to add one before publication is allowed (Should not happen in practice, but a fallback. If all behaves correctly, this will require manual manipulation of RR association in the database)
[...]

PR #12864, opened 2026-06-05
pkp/pkp-lib#12800 Tighten review round association with publications

ewhanson:


classes/migration/upgrade/v3_6_0/I12800_ReviewPublicationAssociationNullable.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_6_0/I12800_ReviewPublicationAssociationNullable.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
```
bozana, line 7: 2016

classes/migration/upgrade/v3_6_0/I12800_ReviewPublicationAssociationNullable.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        Schema::table('review_rounds', function (Blueprint $table) {
+            $table->bigInteger('publication_id')->nullable()->change();
+        });
+
+        Schema::table('edit_decisions', function (Blueprint $table) {
+            $table->dropForeign('edit_decisions_publication_id');
+            $table->dropIndex('edit_decisions_publication_id');
+            $table->dropColumn('publication_id');
+        });
```
bozana: If we remove PKP\migration\upgrade\v3_6_0\I12049_AddPublishReviewCurateSupport -- see my comment in the OJS PR -- we can/should remove this part of the code too.

The developers working with the main would need to get the latest dataset and maybe an extra one-time-run script for ORE test server is needed. I think it is better than having both scripts unnecessarily run on every 3.6 upgrade for each user?

classes/publication/Repository.php
```diff
@@ -1233,4 +1236,37 @@ public function getWithSourcePublicationsIds(array $publicationIds): Collection
...
+        $roundsWithoutPublicationAssociation = $reviewRounds
+            ->filter(fn (ReviewRound $reviewRound) => $reviewRound->getPublicationId() === null);
+
+        if ($roundsWithoutPublicationAssociation->count() === 0) {
+            return;
+        }
+
+        // Check how many review rounds do not have any publication associations
+        // PR_TODO: Post-first published publications can have a maximum of one review round per publication.
+        //          If there's more than one review round with a null publication association, do nothing for now
```
bozana: publication :-)

bozana: would you maybe like to open an issue for this?

classes/submission/reviewRound/ReviewRound.php
```diff
@@ -90,15 +90,16 @@ public function setSubmissionId($submissionId)
     /**
      * Get publication ID
      */
-    public function getPublicationId(): int
+    public function getPublicationId(): ?int
     {
-        return (int) $this->getData('publicationId');
+        $publicationId = $this->getData('publicationId');
+        return $publicationId === null ? null : (int) $publicationId;
```
bozana, line 96: The new possibility of having the `null` value should be considered in all other places in the code, e.g.:
api/v1/reviews/formRequests/ReviewRoundAuthorResponseCommonValidator.php:68:
```
  $publication = Repo::publication()->get($this->reviewRound->getPublicationId());
  $allAuthors = $publication->getData('authors'); // fatal if get(null) → null
```

PublicationPeerReviewResource.php:130 ($round->getPublicationId() !== $publication->getId()) 
and :169 ('originalPublicationId' => $reviewRound->getPublicationId()) 
now silently shift behavior when null. 

Are all the other places in the code correct so, do they need double checking?

bozana (commented):
Hi @ewhanson, here are my comments/questions :-)
Thanks!

classes/migration/upgrade/v3_6_0/I12800_ReviewPublicationAssociationNullable.php
ewhanson: Agreed! I've removed `PKP\migration\upgrade\v3_6_0\I12049_AddPublishReviewCurateSupport` and left instructions for the ORE test server in the issue and will flag it for the VM update.

classes/publication/Repository.php
ewhanson: Yup! Flagged it in #12888.

classes/submission/reviewRound/ReviewRound.php
ewhanson: Will do a double check again and apply to OMP/OPS as well.

Issue #12833, opened 2026-05-31
[ScheduleTask] | Fatal non-recoverable error for web based task runner at invalid request path which unable to resolve context


janevsky:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When sending notifications/emails to editorial staff during the copyediting workflow stage using the template system, a fatal NotFoundHttpException is thrown. The error occurs in the scheduled task service when attempting to load plugin registrations outside of a standard HTTP request context.

The issue is that PluginRegistry::loadAllPlugins() calls PKPApplication->getEnabledProducts(), which calls PKPRouter->getContext(). The router attempts to determine context from the request URL. However, when scheduled tasks run (particularly when sending email notifications), there is no proper HTTP request context available, causing the router to throw a 404 NotFoundHttpException.

This appears to be a context detection issue where the router cannot determine which journal/context the scheduled task is running for.

### Steps to Reproduce

1. Make sure to set the `task_runner` to `On` in the `config.inc.php`
2. Also make sure if any CLI daemon for schedule task is running, turn it off as this is specific for web based schedule task runner
3. Login as admin, go the `Administration` section and clear the data cache
4. Now in another tab, visit a broken path like `http://BASE_URL/index.php/INVALID_CONTEXT/en` but wait 60 seconds before visiting the broken path
5. see that there is exception logged in `error_log`

NOTE that this is possible to re-produce repetitively by setting the `task_runner_interval` value to lower like `5` seconds

### Expected Result

Schedule task should run or not attempt to run at all on invalid request path .

### Actual Result

A fatal PHP exception is thrown: Symfony\Component\HttpKernel\Exception\NotFoundHttpException. The email may or may not be sent (status unclear due to fatal error). The page does not refresh or show confirmation. A server 500 error occurs.

After the error, subsequent operations show warnings that RequestReviewRoundAuthorResponse mailable class does not exist and cannot be loaded.

### Environment Details

1. Server OS: Linux
2. PHP Version: 8.2
3. Database Type: MariaDB
4. Context: Email notification triggered during copyediting workflow

### Application Version

OJS 3.5.0.4

### Logs

[31-May-2026 12:04:31 Europe/Skopje] PHP Fatal error:  Uncaught Symfony\Component\HttpKernel\Exception\NotFoundHttpException in PATH/public_html/lib/pkp/classes/core/PKPRouter.php:205
Stack trace:
[...]

PR #12873, opened 2026-06-08
pkp/pkp-lib#12833 schedule task guard on non context path for web based runner

touhidurabir:
for pkp/pkp-lib#12833 .


bozana (approved):
Wow, what an implementation of the tests!

kaitlinnewson (approved):

PR #12875, opened 2026-06-08
Remove dead code

asmecher:
A quick test of phpstan identified several functions that are no longer used. Remove them.

### DIFF
--- a/classes/tombstone/DataObjectTombstoneDAO.php
+++ b/classes/tombstone/DataObjectTombstoneDAO.php
@@ -86,12 +86,6 @@ class DataObjectTombstoneDAO extends DAO
      */
     public function deleteById(int $tombstoneId, ?int $assocType = null, ?int $assocId = null): int
     {
-        $tombstone = $this->getById($tombstoneId, $assocType, $assocId);
-        if (!$tombstone) {
-            // Did not exist
-            return 0;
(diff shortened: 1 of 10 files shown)


classes/tombstone/DataObjectTombstoneDAO.php
```diff
@@ -86,12 +86,6 @@ public function _fromRow(array $row): DataObjectTombstone
      */
     public function deleteById(int $tombstoneId, ?int $assocType = null, ?int $assocId = null): int
     {
-        $tombstone = $this->getById($tombstoneId, $assocType, $assocId);
```
touhidurabir, line 89: this is fine but if possible better to remove the `$assocType` and `$assocId` from method signature unless it's hard to find out all the places this has been used .

touhidurabir (approved):
Just one comment, not that important . anyway looks good .

classes/tombstone/DataObjectTombstoneDAO.php
asmecher: Actually, this is important -- it was performing validation to ensure that the `$assocType` and `$assocId` matched before deleting the entity. I'll resolve that another way.

asmecher:
Merged manually.

Issue #12799, opened 2026-05-27
Support submission removal from workflow queues


ewhanson:
> [!important]
> This feature requires a database migration. To run it, execute the following:
> ```php
> php lib/pkp/tools/migration.php "\PKP\migration\upgrade\v3_6_0\I12799_MovePublishedSubmissionsToDone" up
> ```

With the introduction of a post-publication review workflow and the decoupling of the submission workflow stage from the publication status, it is now more ambiguous when a submission is considered "done." Submissions need to be able to moved to the published or "done" workflow queue when they are finished, not necessarily when the first publication version is published.

> [!note]
> Traditional pre-publication review workflows will be unaffected by these changes.

## Requirements
- Stay in current workflow stage until first VoR is published
	- CANNOT otherwise change to a different workflow stage while in the "done" queue
- Consolidate submission workflow summary "actions" (Library, Preview, etc.) to a "More options" dropdown
	- Include a "return to workflow"/"return to published" button that is always present, but greyed out when not applicable
- When initiated, return to workflow action move submission back to most recent non-"done" stage once it initially becomes valid for this (i.e. undo and redo)

In summary, movement to "done" workflow queue happens automatically when VoR published, but free movement to and from "done" stage is allowed after that.

## Acceptance criteria
- Library, preview, etc. actions moved to drop down
- "Return to workflow" action present in above dropdown
- "Return to workflow" option inactive before first VoR published
- Submission moves to "done" queue when first VoR published
- Cannot change workflow stage via normal means when in "done" queue
- "Return to workflow" option active post first VoR publication
- "Return to workflow" action moves submission to previous workflow stage
- "Return to done stage" action move submission back to done stage

## PRs:

- `pkp-lib`: pkp/pkp-lib#12881
- `ui-library`: pkp/ui-library#904
- `ojs`: pkp/ojs#5585
- `omp`: pkp/omp#2382
- `ops`: pkp/ops#1323

Fix PRs:
- https://github.com/pkp/pkp-lib/pull/13037
- https://github.com/pkp/ui-library/pull/928
- https://github.com/pkp/ojs/pull/5671
- https://github.com/pkp/omp/pull/2403
- https://github.com/pkp/ops/pull/1343

ewhanson:
Hey @bozana, could you have a look at these? I'll port things over for OMP and OPS afterwards. Thanks!

bozana:
[...]

PR #12881, opened 2026-06-10
pkp/pkp-lib#12799 Support submission removal from workflow queues

ewhanson:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -121,6 +122,7 @@ class PKPSubmissionController extends PKPBaseController
         'editContributor',
         'saveContributorsOrder',
         'addDecision',
+        'returnToDone',
```
bozana, line 126: Should we consider the same policies in authorize() for this endpoint as we have for addDecision? I think at least we should consider DecisionAllowedPolicy. Eventually, to be consistent, we could also add SubmissionCompletePolicy, as addDecision does.
This way, an author who can access their own submission could POST to /{id}/returnToDone, it seems.

classes/decision/types/MoveToDone.php
```diff
@@ -0,0 +1,87 @@
...
+class MoveToDone extends DecisionType
+{
+    public function getDecision(): int
+    {
+        return Decision::MOVE_TO_DONE;
+    }
+
+    public function getStageId(): int
+    {
+        return WORKFLOW_STAGE_ID_PRODUCTION;
```
bozana, line 35: getStageId() returns WORKFLOW_STAGE_ID_PRODUCTION for both MoveToDone and ReturnToDone, but neither decision is meant to be triggered via the regular /decisions endpoint — MoveToDone is system-only (listener/migration) and ReturnToDone has its own endpoint. A follow-up issue/PR could eventually make getStageId() nullable and have DecisionStageValidPolicy deny on null to make this explicit?

classes/submission/Repository.php
```diff
@@ -947,6 +947,39 @@ public function getNextAvailableVersion(Submission $submission, VersionStage $ve
...
+            ->first();
+        if ($managerId) {
+            return (int) $managerId;
+        }
+
+        // Terminal fallback: a site administrator always exists on a functioning install.
+        return (int) Repo::user()->getCollector()
+            ->filterByRoleIds([Role::ROLE_ID_SITE_ADMIN])
+            ->getIds()
+            ->first();
```
bozana, line 980: There will be probably always an admin, but how about throwing new RuntimeException() -- it makes the impossible case loud and obvious rather than letting a 0 propagate silently into edit_decisions.editor_id.
You have considered the case in the migration script as well -- the migration handles this with error_log + skip.

classes/decision/types/MoveToDone.php
bozana: if getStageId() is made nullable and DecisionStageValidPolicy updated, the custom /returnToDone endpoint could eventually be replaced by the standard /decisions endpoint with the validation moved into e.g. ReturnToDone::validate(). 🤔

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+        if (!$editorId) {
+            error_log("I12799: skipping submission {$submission->submission_id}; no editor, journal manager or site administrator could be resolved.");
+            return;
+        }
+
+        // Prefer the Version of Record's publication date so the recorded history
+        // reflects when the submission actually became "done".
+        $datePublished = DB::table('publications')
+            ->where('publication_id', '=', $submission->current_publication_id)
+            ->value('date_published');
```
bozana: This query we can also move above, to the submission select above.

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+            'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION,
+            'assocId' => (int) $submission->submission_id,
+            'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_EDITOR_DECISION,
+            'userId' => $editorId,
+            'editorName' => $this->getEditorName($editorId),
+            'message' => 'editor.submission.decision.moveToDone.log',
+            'isTranslated' => false,
+            'dateLogged' => $dateDecided,
+        ]);
+        Repo::eventLog()->add($eventLog);
```
bozana: Not only because of the performance reasons, but also our general rules for migrations scripts, this should rather be a direct insert.
We should never use repositories or models, because:
Migrations are frozen in time — they represent what needs to happen at a specific version. If repositories or models or functions or X in the code change their behavior, required fields, validation, etc. in a future version, the migrations would silently behave differently or break when run on an older database during upgrade i.e. on every change the migrations would need to be considered too.

Something like:
```php
        // Mirror the event-log entry written by the runtime decision Repository::add().
        $logId = DB::table('event_log')->insertGetId([
            'assoc_type'    => PKPApplication::ASSOC_TYPE_SUBMISSION,
            'assoc_id'      => (int) $submission->submission_id,
            'user_id'       => $editorId,
            'date_logged'   => $dateDecided,
            'event_type'    => PKPSubmissionEventLogEntry::SUBMISSION_LOG_EDITOR_DECISION,
            'message'       => 'editor.submission.decision.moveToDone.log',
            'is_translated' => 0,
        ]);

        DB::table('event_log_settings')->insert([
            'log_id'        => $logId,
            'locale'        => '',
            'setting_name'  => 'editorName',
            'setting_value' => $this->getEditorName($editorId),
        ]);

```

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+    private array $editorNameCache = [];
+
+    /**
+     * @inheritDoc
+     */
+    public function up(): void
+    {
+        // Transfer existing STATUS_PUBLISHED submissions to new WORKFLOW_STAGE_ID_DONE stage.
+        DB::table('submissions')
+            ->where('status', '=', PKPSubmission::STATUS_PUBLISHED)
```
bozana: because of the performance improvement, maybe to join with publications to get datePublished as well as editor assignments already here, something like:
```php
        DB::table('submissions as s')
            ->join('publications as p', 'p.publication_id', '=', 's.current_publication_id')
            ->leftJoinSub(
                DB::table('stage_assignments as sa')
                    ->join('user_groups as ug', 'ug.user_group_id', '=', 'sa.user_group_id')
                    ->whereIn('ug.role_id', [Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_MANAGER])
                    ->select('sa.submission_id', DB::raw('MIN(sa.stage_assignment_id) as min_assignment_id'))
                    ->groupBy('sa.submission_id'),
                'min_sa',
                'min_sa.submission_id', '=', 's.submission_id'
            )
            ->leftJoin('stage_assignments as sa_first', 'sa_first.stage_assignment_id', '=', 'min_sa.min_assignment_id')
            ->where('s.status', '=', PKPSubmission::STATUS_PUBLISHED)
            ->where('s.stage_id', '<>', WORKFLOW_STAGE_ID_DONE)
            ->select([
                DB::raw('s.submission_id as submission_id'),
                's.context_id',
                's.stage_id',
                'p.date_published',
                'sa_first.user_id as assigned_editor_id',
            ])

```

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+            ->where('stage_id', '<>', WORKFLOW_STAGE_ID_DONE)
+            ->select(['submission_id', 'context_id', 'stage_id', 'current_publication_id'])
+            ->chunkById(100, function (Collection $submissions) {
+                foreach ($submissions as $submission) {
```
bozana, line 74: Maybe also to collect the IDs that need updating here and update at the end, i.e. something like: 
```php
            ->chunkById(100, function (Collection $submissions) {
                $toUpdate = [];
                foreach ($submissions as $submission) {
                    if ($this->moveSubmissionToDone($submission)) {
                        $toUpdate[] = $submission->submission_id;
                    }
                }
                if ($toUpdate) {
                    DB::table('submissions')
                        ->whereIn('submission_id', $toUpdate)
                        ->update(['stage_id' => WORKFLOW_STAGE_ID_DONE]);
                }
            }
```

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+    private function moveSubmissionToDone(object $submission): void
+    {
+        // Skip if a MOVE_TO_DONE decision has already been recorded.
+        $alreadyMoved = DB::table('edit_decisions')
+            ->where('submission_id', '=', $submission->submission_id)
+            ->where('decision', '=', Decision::MOVE_TO_DONE)
+            ->exists();
+        if ($alreadyMoved) {
+            return;
+        }
```
bozana: I believe this check can be removed. Above we check:
```
  ->where('status', '=', PKPSubmission::STATUS_PUBLISHED)
  ->where('stage_id', '<>', WORKFLOW_STAGE_ID_DONE)
```
And theoretically on upgrade from earlier versions there should be no such decisions, right?

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+
+        // The decision is attributed to the submission's current publication, matching
+        // the runtime decision Repository::add().
+        if (empty($submission->current_publication_id)) {
+            error_log("I12799: skipping submission {$submission->submission_id}; it is published but has no current publication.");
+            return;
+        }
```
bozana: If we join earlier with publications to get date published with:
```
DB::table('submissions as s')
      ->join('publications as p', 'p.publication_id', '=', 's.current_publication_id')
      ...
```
this can be skipped too

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+
+        $editorId = $this->resolveEditorId($submission);
+        if (!$editorId) {
+            error_log("I12799: skipping submission {$submission->submission_id}; no editor, journal manager or site administrator could be resolved.");
+            return;
+        }
```
bozana, line 113: and this can be: 
```
$editorId = $submission->assigned_editor_id
            ?? $this->resolveContextManagerId((int) $submission->context_id)
            ?? $this->resolveSiteAdminId();
```
I.e. either we already have assigned_editor_id, or we call those function only when needed and they use cache.

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+            'isTranslated' => false,
+            'dateLogged' => $dateDecided,
+        ]);
+        Repo::eventLog()->add($eventLog);
+
+        // Finally, move the submission into the Done stage. Status is intentionally left
+        // untouched: a runtime move-to-Done leaves the submission STATUS_PUBLISHED.
+        DB::table('submissions')
+            ->where('submission_id', '=', $submission->submission_id)
+            ->update(['stage_id' => WORKFLOW_STAGE_ID_DONE]);
```
bozana: This we can do above and for a set of submission IDs

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Resolve the editor to attribute the system-recorded decision to, mirroring
+     * publication Repository::resolveSystemEditorId(): the earliest assigned
+     * editor/manager, then a journal manager in the context, then a site administrator.
+     */
+    private function resolveEditorId(object $submission): ?int
```
bozana: we can instead have resolveContextManagerId and resolveSiteAdminId, s. above

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
bozana: If we do not want to have the assigned editor queried above -- because it is less readable -- maybe to put it here as per-chunk pre-fetch approach:
```php
 $assignedEditors = DB::table('stage_assignments as sa')
      ->join('user_groups as ug', 'ug.user_group_id', '=', 'sa.user_group_id')
      ->whereIn('sa.submission_id', $ids)
      ->whereIn('ug.role_id', [Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_MANAGER])
      ->orderBy('sa.stage_assignment_id')
      ->get()
      ->groupBy('submission_id')
      ->map(fn($rows) => (int) $rows->first()->user_id);

```

classes/decision/types/MoveToDone.php
bozana: just thinking...

bozana (commented):
Hi @ewhanson, great work!
I have just two comments -- to consider DecisionAllowedPolicy for returnToDone, similarly how it is for addDecision.
And a few suggestions to improve the performance for the migration script, above all for huge installations that exist out there.
Thanks a lot!

classes/decision/types/MoveToDone.php
ewhanson: That would be ideal, thanks! I think it's probably best to do that as a separate PR at this point, but I was trying to think of how to make it work with the existing endpoint, but my brain ran out of energy. Will make the changes needed here for now to apply proper authorization, but will make a new follow up issue too.

classes/migration/upgrade/v3_6_0/I12799_MovePublishedSubmissionsToDone.php
ewhanson: Thanks, I'll do that. I was also a bit torn on whether we should be including an event log entry for every single published issue as it was a "new" way of conceptualizing the done state, but it seemed best for consistency.

Issue #12794, opened 2026-05-27
[Media Files] | Editorial Submission workflow - Batch/Manual linking, when linking an already linked file to another option there is an error


Tribunal33:
## Requirements

Already linked files should not have the option to link to other files, only unlinking should be a valid option. This should be the case for batch linking and manual linking.

This means that if a user wants to link already linked files, the linked files must first be unlinked before any further linking actions can be performed on them. 

## Acceptance criteria

- Cannot attempt to link already linked files to other files.
- Can unlink files from batch linking.
- Can unlink files from manual linking
- Previously documented issue below no longer applicable.

### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

During testing of #12262 there was an error that occured while trying to link previously linked files.  Manually linked will throw the error with either the low res to high res or high-res to low-res versions of a file.  Batch link can only link high-res files as the low-res are pre selected and uneditable. 

To fix this some considerations are needed.

- [ ] How are we going to handle the metadata?  I assume it continues to overwrite it like other links do.
- [ ] Also for Batch Link Media but only from web (low) version to high-res version as web version is pre selected in the table.

### Steps to Reproduce

1. Precondition: Have multiple files of the same type with multiple file versions (high and low) uploaded to a submission.  (Example will be from low-res linking another high-res file)
2. Login with a user  (Admin / JM / JE) with access to Media File section
3. Navigate to the submission with the files in the precondition 
4. Click on Publication > Media Files
5. Select the 3 dots on a low-res file > Manual Link Media 
6. From the dropdown `Select the media file to link as its counterpart` and link another high-res file from the list
7. Select Link Media

### Expected Result

No error and new link should be established.  The metadata of both files should adopt the metadata of the file that did the link.

So if the Low-res version initiated the link and selected the counter part the metadata will be from the Low-res version.
If the high-res version initiated the link and selected the counter part the metadata will be from the high-res version.

### Actual Result

Console Error and button is disabled
{"error":"The variant group already has the maximum number of files."}
<img width="1066" height="599" alt="Image" src="https://github.com/user-attachments/assets/47b3346d-5d94-4643-a2ac-e043c2d399ea" />

Batch link media results in same error but it does push the error message to the UI

<img width="1187" height="409" alt="Image" src="https://github.com/user-attachments/assets/2e46afb6-716d-466d-80dc-5e00a7069fe5" />

### Application Version

OJS, OMP, OPS 3.6(main)

---
### Development Updates
[...]

PR #12893, opened 2026-06-11
pkp/pkp-lib#12794 [Media Files] | Editorial Submission workflow - Batch/Manual linking, when linking an already linked file to another option there is an error

blesildaramirez:


ewhanson (approved):
Looks good, thanks @blesildaramirez!

PR #12896, opened 2026-06-12
fix: accept POST for bulk delete of incomplete submissions

mehedihasanhemel:
## Problem

The **Delete Incomplete Submissions** bulk action in the admin UI silently fails — clicking the button does nothing, no error is shown.

Checking the Apache access log reveals the cause:

```
POST /api/v1/_submissions?ids[]=634&... HTTP/1.1" 404
```

The frontend JavaScript sends a `POST` request to `/api/v1/_submissions`, but the backend only has `Route::delete` registered for that endpoint. The `POST` hits a 404 and is silently ignored.

## Root Cause

The frontend bulk-delete action dispatches a `POST` request (likely intending a method-override pattern), but no method-override middleware is configured in the PKP API router. The backend expects `DELETE`.

## Fix

Register `Route::post` for the same `bulkDeleteIncompleteSubmissions` handler alongside `Route::delete`, so the endpoint accepts both HTTP methods until the frontend is corrected to send `DELETE`.

## Tested

- OJS 3.5.0.4 / PHP 8.5.4 / Apache 2.4.66 / MySQL 8.4.9
- Bulk delete of incomplete submissions now works correctly through the admin UI after applying this fix.

## Note

If the frontend is later corrected to send `DELETE`, the `Route::post` alias can be safely removed — or left in place as it causes no conflict.

### DIFF
--- a/api/v1/_submissions/PKPBackendSubmissionsController.php
+++ b/api/v1/_submissions/PKPBackendSubmissionsController.php
@@ -86,6 +86,18 @@ abstract class PKPBackendSubmissionsController extends PKPBaseController
                 ]),
             ]);
 
+        // The frontend sends POST instead of DELETE for bulk deletion.
+        // Accept both methods until the frontend is corrected.
+        Route::post('', $this->bulkDeleteIncompleteSubmissions(...))
+            ->name('_submission.incomplete.delete.post')
+            ->middleware([
+                self::roleAuthorizer([
+                    Role::ROLE_ID_SITE_ADMIN,
+                    Role::ROLE_ID_MANAGER,
+                    Role::ROLE_ID_AUTHOR,
+                ]),
+            ]);
+
         Route::delete('{submissionId}', $this->delete(...))
             ->name('_submission.delete')
             ->middleware([


mehedihasanhemel:
recheck

mehedihasanhemel:
recheck

mehedihasanhemel:
recheck

mehedihasanhemel:
recheck

mehedihasanhemel:
recheck

asmecher:
@mehedihasanhemel, is your server perhaps filtering the `x-http-method-override` header before it gets to OJS? The front end is sending the `DELETE` request using that header.
image

mehedihasanhemel:
Thanks for the pointer, @asmecher. I tested this on our server:

**The `x-http-method-override` header does reach PHP** — Apache is not stripping it. I confirmed by echoing `getallheaders()` through OJS's webroot and the header arrives intact.

The issue is that **Symfony only reads `x-http-method-override` if `Request::enableHttpMethodParameterOverride()` has been called first** (see [Symfony docs](https://symfony.com/doc/current/routing.html#routing-http-method-requirements) and the note in `symfony/http-foundation/Request.php` line 1157). OJS never calls this method, so `Request::getMethod()` ignores the header entirely and returns `POST`.

So the flow is:
1. Frontend sends `POST /api/v1/_submissions?ids[]=...` with `x-http-method-override: DELETE`
2. Symfony receives it but — because `enableHttpMethodParameterOverride()` was never called — returns `POST` from `getMethod()`
3. The router finds no `Route::post` for that endpoint → 404

The fix could go either way:
- **Option A (cleanest):** Call `Request::enableHttpMethodParameterOverride()` during OJS bootstrap so Symfony honors the header and the existing `Route::delete` matches correctly
- **Option B (what this PR does):** Register `Route::post` as an alias pointing to the same handler

Happy to update the PR to Option A if that's preferred.

asmecher:
@mehedihasanhemel, I think there's still more to it -- the feature works as expected on my machine. (I tested it when I made that screenshot without any changes made, and the feature works as expected.)

mehedihasanhemel:
Thanks @asmecher. After re-reading Symfony's `getMethod()` source, I realize my earlier analysis was incorrect — Symfony reads `X-HTTP-METHOD-OVERRIDE` at line 1175 without needing `enableHttpMethodParameterOverride()`. I apologize for the wrong explanation.

The bug is clearly server-specific on our end. We're running PHP 8.5.4 (pre-release) which may be a factor. I'll investigate further to find what's actually causing the header to not be honored here.

Happy to close this PR if you'd prefer — the Route::post alias works as a local workaround for us. Could you share what PHP version you tested on? That may help narrow it down.

asmecher:
@mehedihasanhemel, we're testing OJS 3.5.0-x on PHP 8.2 and 8.3 via CI (see https://github.com/pkp/ojs/actions); 8.4 is broadly used and I think a few of us on the dev team are using it on the regular. I don't think any of us have tried an 8.5 pre-release yet. Let me know if you can determine what's happening!

I think it's better to fix the root cause rather than merging the work-around. A `POST` request shouldn't cause a `DELETE` action, and there's nothing in the URL that would lead someone to expect that behaviour.

Issue #12712, opened 2026-05-12
[Eloquent Model] ModelWithSettings: schema-derived cast silently skipped for multi-word camelCase setting props


touhidurabir:
**Description:**
For Eloquent models using `PKP\core\traits\ModelWithSettings` (every model backed by both a primary table and a settings table — `Announcement`,  `ContributorRole`, `DataCitation`, `ReviewerSuggestion`, etc.), a non-multilingual setting prop whose JSON schema name contains camelCase words (like `authorsList`, `dataReadme` -- do not have such yet but possible) does **not** get its declared type cast applied on read. The attribute comes back as the raw stored string instead of the decoded  `array`/`object`.

For example, Single-word lowercase prop names (`authors`) and snake_case prop names (`authors_list`) are unaffected — only multi-word camelCase names trip the bug. 

Multilingual setting props are also unaffected: they go through `MultilingualSettingAttribute` which doesn't depend on Eloquent's `castAttribute()` for JSON decoding.

## Impact                                              

  - **Silent failure** — no exception, no warning. Code that expects an array and gets a string will throw downstream (`Cannot use string of type ... as array`, JSON deserialization errors, etc.), or produce subtle correctness bugs (string truthiness vs. array emptiness, iteration mistakes).
  - **Affects every Eloquent + JSON-schema model** the moment a new setting prop with a multi-word camelCase name is added.                                       

**Reproduction Approach**
Using `PKP\dataCitation\DataCitation` (`authors` is currently typed `array` in `lib/pkp/schemas/dataCitation.json`):

```php
use PKP\dataCitation\DataCitation;

$c = DataCitation::find(1);

// Baseline — single-word lowercase prop works:
$c->authors;                           // ✅ array
gettype($c->authors);           // → "array"

// Rename schema prop to authorsList (multi-word camelCase), re-fetch:
$c->authorsList;                       // ❌ string (raw JSON)                                                                                                  
gettype($c->authorsList);       // → "string"

// Rename schema prop to authors_list (snake_case), re-fetch:                                                                                                   
$c->authors_list;                      // ✅ array — cast applied again                                                                                         
gettype($c->authors_list);      // → "array"

// The cast IS registered, just under a transformed key. Confirm with:

foreach ($c->getCasts() as $key => $cast) {
      echo $key . ' => ' . $cast . PHP_EOL; 
}
// For a schema with 'authorsList':                                                                                                                             
// authors_list => array     ← key was snake-cased by HasCamelCasing

[...]

PR #12902, opened 2026-06-15
pkp/pkp-lib#12712 Optimise schema deriven eloquent casting

touhidurabir:
for pkp/pkp-lib#12712 .

### DIFF
--- a/classes/core/traits/ModelWithSettings.php
+++ b/classes/core/traits/ModelWithSettings.php
@@ -221,7 +221,7 @@ trait ModelWithSettings
      * that same camelCase key, so their cast must be keyed camelCase too.
      *
      * Everything else keeps HasCamelCasing's snake_case behaviour:
-     *  - primary-table columns are stored snake_case and must snake to match;
+     *  - primary-table columns are stored snake_case and must be snake_case to match
      *  - multilingual props cast via the inbound-only MultilingualSettingAttribute,
      *    which fires on the snake-cased set path (and is not applied on read), so
      *    their cast key must stay snake_case to keep firing on write.
--- a/classes/core/traits/ModelWithSettings.php
+++ b/classes/core/traits/ModelWithSettings.php
@@ -9,23 +9,19 @@
  *
  * @class ModelWithSettings
  *
- * @ingroup core_traits
- *
  * @brief A trait for Eloquent Model classes that can be used with entities that have a settings table.
  *
  */
 
 namespace PKP\core\traits;
 
-use Exception;
 use Eloquence\Behaviours\HasCamelCasing;
-use Illuminate\Support\Str;
+use Exception;
 use Illuminate\Database\Eloquent\Casts\Attribute;
-use PKP\core\traits\LocalizedData;
+use Illuminate\Support\Str;
 use PKP\core\casts\MultilingualSettingAttribute;
 use PKP\core\maps\Schema;
 use PKP\core\SettingsBuilder;
-use PKP\facades\Locale;
 use PKP\services\PKPSchemaService;
 use stdClass;
 
@@ -79,7 +75,7 @@ trait ModelWithSettings
     public function __construct(array $attributes = [])
     {
         parent::__construct($attributes);
-        
+
         if (static::getSchemaName()) {
             $this->setSchemaData();
         } else {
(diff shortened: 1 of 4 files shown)


Vitaliy-1 (approved):

classes/core/traits/ModelWithSettings.php
```diff
@@ -221,6 +214,65 @@ public function getAttribute($key): mixed
...
     }
 
+    /**
+     * Override HasCamelCasing::getCasts() so casts for non-multilingual setting props
+     * resolve under their camelCase attribute key. Such settings are hydrated by
+     * SettingsBuilder verbatim from the camelCase setting_name and read back under
+     * that same camelCase key, so their cast must be keyed camelCase too.
+     *
+     * Everything else keeps HasCamelCasing's snake_case behaviour:
+     *  - primary-table columns are stored snake_case and must snake to match;
```
taslangraham: This sentence seems a bit off. Did you mean the following?
```suggestion
     *  - primary-table columns are stored snake_case and must be snake_case to match;
```

classes/core/traits/ModelWithSettings.php
```diff
@@ -221,6 +214,65 @@ public function getAttribute($key): mixed
...
+     */
+    public function setAttribute($key, $value)
+    {
+        $camelKey = Str::camel($key);
+
+        if (in_array($camelKey, $this->getSettings()) && !in_array($camelKey, $this->getMultilingualProps())) {
+            return parent::setAttribute($camelKey, $value);
+        }
+
+        return parent::setAttribute($this->getSnakeKey($key), $value);
```
taslangraham, line 273: Under the hood, `$this->getSnakeKey()` just returns the result from calling `Str::snake()`, can we just directly call `Str::snake()` here?

tests/classes/core/traits/InteractsWithSettingsModel.php
```diff
@@ -0,0 +1,379 @@
...
+        foreach ($familyNameByLocale as $locale => $value) {
+            DB::table('test_settings_pure_entity_settings')->insert([
+                'test_id' => $modelId,
+                'locale' => $locale,
+                'setting_name' => 'familyName',
+                'setting_value' => $value,
+            ]);
+        }
+        return $modelId;
+    }
```
taslangraham, line 216: I think I understand why these are separated into two methods, but both methods are essentially doing the same thing. This could be done via a single generic method that accepts the table name and `locale` array OR maybe keep the individual methods(separate method name makes it easier to reason about), but move the internal logic to a single generic method. What do you think?

tests/classes/core/traits/InteractsWithSettingsModel.php
```diff
@@ -0,0 +1,379 @@
...
+    {
+        $modelId = DB::table('test_settings_pure_entity')->insertGetId(['parent_id' => 1], 'test_id');
+        DB::table('test_settings_pure_entity_settings')->insert([
+            'test_id' => $modelId,
+            'locale' => $locale,
+            'setting_name' => $settingName,
+            'setting_value' => $settingValue,
+        ]);
+        return $modelId;
+    }
```
taslangraham, line 246: Similar comment to the one made [above](https://github.com/pkp/pkp-lib/pull/12902/changes#r3437476347).

taslangraham (changes requested):
@touhidurabir Overall, this looks good. I just had a few minor questions/suggestions

tests/classes/core/traits/InteractsWithSettingsModel.php
touhidurabir: @taslangraham I understand your point and this is something I initially went with . However I revert back as we moved to a trait and then use tests that can use that trait does not need to know much except just make calls as needed . but refactoring to a generic method require the tests class to pass test db name (for pure vs schema) instead of just make the call , basically tests class need to know bit more about setting up the test requirement specifically . hence this almost similar double implementation . 

I dont think for tests this is an issue and usually very normal to make stuff more simpler .

tests/classes/core/traits/InteractsWithSettingsModel.php
touhidurabir: same as [above](https://github.com/pkp/pkp-lib/pull/12902#discussion_r3440640382)

classes/core/traits/ModelWithSettings.php
touhidurabir: it's kind of intentional as for the read/write path, it went through the `HasCamelCasing` so just maintaining a symmetry in the code . so lets lets keep it there as it is now .

taslangraham (approved):

Issue #7527, opened 2021-12-03
OJS title-level metadata enhancements


AhemNason:
**Describe the problem you would like to solve**
As expressed in the ticket on NISO PIE-J recommendations towards presentation and identification of journal titles, OJS does a poor job at handling historical name changes in article-level metadata. For example, if I changed the name of my journal in OJS settings, all generated citations for my older issues would display the _new name of my journal_, not the name of the journal as it existed at publication. 

This is kind of a big deal! 

Please see: https://github.com/pkp/pkp-lib/issues/2505

Nate asked for the formation of a new ticket in this space, and some recommendations for the following: 

1. What precise pieces of metadata need to be tracked at the article level which are not being tracked now? (ie - journal title and ISSN, what else?)
2. For each piece of metadata, when should this data be "stamped" in the article?
3. When this "stamped" data needs to be changed (due to a mistake or any other reason), should it require a special operation (like creating a new version) or can any editor edit it whenever they want?
4. How will the system know when changes to stamped metadata are made, so that it can redistributed or resync metadata with third-parties?
5. How will a journal recover from a history of bad data? (ie - when a journal has already published lots of articles with the wrong title or ISSN)

I'll do my best. 

> What precise pieces of metadata need to be tracked at the article level which are not being tracked now? (ie - journal title and ISSN, what else?)

Journal title and ISSN are the big ones. Title-level metadata is very important. Any issue-level metadata would ideally be tied to an article as well. Like volume, issue, year, publication date.

> For each piece of metadata, when should this data be "stamped" in the article?
On publication. 

> When this "stamped" data needs to be changed (due to a mistake or any other reason), should it require a special operation (like creating a new version) or can any editor edit it whenever they want?

I'd like to make a case for the ability to edit the publication record for an article _from the article's metadata_ and _only as a new version of the article_. It's certainly possible someone could publish, say, their first issue with a typo in the journal title and want to fix it. I think that's ok. But, once something has been published, changing my journal-level metadata _should not_ change the metadata for my already-published works. If I _need_ to change them, I could do it from the article metadata page in the publishing workflow. 

> How will the system know when changes to stamped metadata are made, so that it can redistributed or resync metadata with third-parties?

I'm less clear on this. I know @ewhanson has been working on "stale" status for existing Crossref deposits for example. If you require a new version of an article to publish it under _a different title_, at least you can flag on the publication of a new version. 

[...]

PR #12915, opened 2026-06-17
pkp/pkp-lib#7527 Stamp context identity metadata onto publications and issues

bozana:
s. https://github.com/pkp/pkp-lib/issues/7527

### DIFF
--- a/classes/orcid/PKPOrcidWork.php
+++ b/classes/orcid/PKPOrcidWork.php
@@ -79,7 +79,7 @@ abstract class PKPOrcidWork
                 ]
             ],
             'journal-title' => [
-                'value' => $this->context->getName($publicationLocale) ?? $this->context->getName($this->context->getPrimaryLocale()),
+                'value' => $this->publication->getData('contextName', $publicationLocale) ?? $this->publication->getData('contextName', $this->context->getPrimaryLocale()),
             ],
             'short-description' => trim(strip_tags($this->publication->getLocalizedData('abstract', $publicationLocale))) ?? '',
 
@@ -215,10 +215,10 @@ abstract class PKPOrcidWork
         }
 
         // Add journal online ISSN, if it exists
-        if ($context->getData('onlineIssn')) {
+        if ($publication->getData('onlineIssn')) {
             $externalIds[] = [
                 'external-id-type' => 'issn',
-                'external-id-value' => $context->getData('onlineIssn'),
+                'external-id-value' => $publication->getData('onlineIssn'),
                 'external-id-relationship' => 'part-of'
             ];
         }
--- a/classes/orcid/PKPOrcidWork.php
+++ b/classes/orcid/PKPOrcidWork.php
@@ -15,13 +15,13 @@
 namespace PKP\orcid;
 
 use APP\author\Author;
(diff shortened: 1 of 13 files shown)


classes/orcid/PKPOrcidWork.php
```diff
@@ -215,10 +215,10 @@ private function buildOrcidExternalIds(Submission $submission, Publication $publ
         }
 
         // Add journal online ISSN, if it exists
-        if ($context->getData('onlineIssn')) {
+        if ($onlineIssn = $publication->getOnlineIssn($context)) {
```
kaitlinnewson, line 218: This is only for OJS correct? Should it be in OJS or check the application in use? It looks like `getOnlineIssn()` is only in the OJS code.

bozana: will fix it now...

classes/publication/Repository.php
```diff
@@ -604,6 +606,11 @@ public function publish(Publication $publication, false|int|null $submissionStat
 
         $itsPublished = ($newPublication->getData('status') === PKPPublication::STATUS_PUBLISHED);
 
+        // Stamp identity on first publication, only when empty so versions inherit (see version()).
+        if ($itsPublished && !$newPublication->getData('contextName')) {
+            $newPublication->stampContextIdentity();
```
kaitlinnewson, line 611: Later in this method we fetch the `$context` (Line 691) - could that be moved above this and passed to `stampContextIdentity()` here?

classes/publication/PKPPublication.php
```diff
@@ -53,6 +55,38 @@ public function getDefaultLocale(): ?string
         return $this->getData('locale');
     }
 
+    /** Get the context this publication belongs to. */
+    protected function getStampingContext(): Context
+    {
+        $submission = Repo::submission()->get((int) $this->getData('submissionId'));
```
kaitlinnewson, line 61: Could this return null, as well as the line below?

classes/publication/PKPPublication.php
```diff
@@ -53,6 +55,38 @@ public function getDefaultLocale(): ?string
         return $this->getData('locale');
     }
 
+    /** Get the context this publication belongs to. */
```
kaitlinnewson, line 58:
```suggestion
    /** 
     * Get the context this publication belongs to. 
     */
```

classes/publication/Repository.php
kaitlinnewson: And it looks like `stampContextIdentity()` -> `getStampingContext()` could potentially re-fetch the submission that is fetched here, if I am reading it correctly?

classes/publication/Repository.php
```diff
@@ -604,6 +606,11 @@ public function publish(Publication $publication, false|int|null $submissionStat
 
         $itsPublished = ($newPublication->getData('status') === PKPPublication::STATUS_PUBLISHED);
 
+        // Stamp identity on first publication, only when empty so versions inherit (see version()).
+        if ($itsPublished && !$newPublication->getData('contextName')) {
```
kaitlinnewson, line 610: Is there any risk of empty locale data rows in the DB causing problems with `!$newPublication->getData('contextName')`? I know it's not typical but seems to happen from time to time.

Issue #12913, opened 2026-06-17
Expose additional data from task-related event logs


Vitaliy-1:
**Description:**
The Task Controller's `getTasks` endpoint among other data, returns information about latest activities associated with the specific task/discussion

**Steps to test the changes:**
Example of the current data shape
```json
[
    {
        "id": 326,
        "message": "vutyag.pdf uploaded by admin on 2026-06-17",
        "type": 1342177287,
        "date": "2026-06-17 07:35:27",
        "userFullName": "admin admin",
        "userId": 1
    }
]
```
**Solution:** What the proposed solutions are and additional affected areas
Example of data shape after the change
```json
[
    {
        "id": 326,
        "message": "vutyag.pdf uploaded by admin on 2026-06-17",
        "type": 1342177287,
        "date": "2026-06-17 07:35:27",
        "userFullName": "admin admin",
        "userId": 1,
        "settings": {
            "fileId": 48,
            "filename": {
                "en": "somefile.pdf"
            },
            "stageId": 3
        }
    }
]
```

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PR: https://github.com/pkp/pkp-lib/pull/12916
OJS: https://github.com/pkp/ojs/pull/5609

Vitaliy-1:
@blesildaramirez, could you take a look at the PR regarding the data shape? Is it enough for you on the front-end?

blesildaramirez:
Hi @Vitaliy-1 , yes this will be enough to build the download URL for an event log of a file upload. Thank you!

PR #12916, opened 2026-06-17
pkp/pkp-lib#12913 Expose attached file id in the task latest activities

Vitaliy-1:


blesildaramirez (approved):
Hi @Vitaliy-1 , this looks good on my end. Thank you!

PR #12931, opened 2026-06-22
pkp/pkp-lib#11825 Build a download link for a file upload event in task and discussions

blesildaramirez:


blesildaramirez:
Hi @Vitaliy-1 , could you please review this small PR for building the download link for the frontend, this is to display a Download button for each uploaded file event on a task/discussion. I initially planned on building the URL on the frontend because there is some areas we do this on the frontend. However, there is a special url convention for download links that include `$$$call$$$` which seems to be intended for server-side resolution rather than being constructed manually on the frontend. Thanks!

Vitaliy-1 (approved):

Issue #12928, opened 2026-06-19
[Tasks and discussions] for reviewers in task participants include reviews from all review rounds


Vitaliy-1:
**Description:**
In the task controller, the `participants` endpoint includes data regarding each participant, including reviewers. The later additionally include data regarding the most recent review round. 
Include information regarding review assignments in all review rounds instead.

**Steps to test the changes:**
The current data shape:
```json
{
  "id": null,
  "userId": 7,
  "isResponsible": false,
  "fullName": "Julie Janssen",
  "username": "jjanssen",
  "roles": [
    {
      "id": 4096,
      "name": "Reviewer",
      "reviewMethod": 2
    }
  ]
}
```

**Solution:** 
Change to:
```json
{
  "id": null,
  "userId": 7,
  "isResponsible": false,
  "fullName": "Julie Janssen",
  "username": "jjanssen",
  "roles": [
    {
      "id": 4096,
      "name": "Reviewer",
      "reviews": [
        {
          "reviewId": 37,
          "reviewRound": 1,
          "reviewMethod": 2
        },
        {
          "reviewId": 40,
          "reviewRound": 2,
          "reviewMethod": 2
        }
      ]
    }
  ]
}
```

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/12935
OJS: https://github.com/pkp/ojs/pull/5622

Vitaliy-1:
@blesildaramirez, can you take a look if the new data, passed through the participants API endpoint is enough for the front-end?

blesildaramirez:
Hi @Vitaliy-1 , thank you! I tested the changes on my local and the frontend can display the necessary information with the updated data shape for the reviewers. Thanks!

PR #12935, opened 2026-06-22
pkp/pkp-lib#12928 Add data regarding review rounds in task participants

Vitaliy-1:


blesildaramirez (approved):
I tested this and it's working well with the UI for displaying the review types. Thanks a lot!

PR #12949, opened 2026-06-25
pkp/pkp-lib#12513  Add site primary locale to the invite/accept forms…

jardakotesovec:
… + copy over if empty

### DIFF
--- a/classes/invitation/invitations/userRoleAssignment/handlers/api/UserRoleAssignmentReceiveController.php
+++ b/classes/invitation/invitations/userRoleAssignment/handlers/api/UserRoleAssignmentReceiveController.php
@@ -14,6 +14,7 @@
 
 namespace PKP\invitation\invitations\userRoleAssignment\handlers\api;
 
+use APP\core\Application;
 use APP\facades\Repo;
 use Carbon\Carbon;
 use Core;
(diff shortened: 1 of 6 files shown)


bozana (approved):

classes/invitation/invitations/userRoleAssignment/handlers/api/UserRoleAssignmentReceiveController.php
```diff
@@ -168,6 +169,13 @@ public function refine(Request $illuminateRequest): JsonResponse
         $reqInput = $illuminateRequest->all();
         $payload = $reqInput['invitationData'];
 
+        // Every user needs a name in the site primary locale (the fallback used by e.g.
+        // getFullName()), but the invitee usually fills only the context primary locale. Copy the
+        // context primary name into the site primary locale when it was left empty, so the value is
+        // stored and replayed in the review step. Only empty values are filled, so a name the
+        // invitee set themselves (e.g. after going back) is never overwritten.
```
bozana: I think this comment can be removed -- the function contains it.

bozana:
Hi @jardakotesovec, all fine here.
One question maybe:
Would it be worth adding fillSiteLocaleNameFallback() also to finalize() as a defensive measure -- if someone would programmatically call `finalize` via a direct API call — skipping `refine` entirely and calling `finalize` directly?

jardakotesovec:
@bozana Thank you, I cleaned up the comment duplication and also improved the copy over logic so it does not crash if invitationData is not in the payload.

Regarding additional guards in finalise - while looking more into that I noticed (was able to reproduce) that it even does not enforce me to have firstName in journal primary locale. Thats probably because its using 'sometime' rule even for finalise step. But I would be worried to start touching the validations last minute.

I was also thinking what approach could be cleanest from clean separation of concerns (frontend vs backend). And maybe the cleanest way to approach this would be just improve the validations to enforce the primary journal&site locales to be provided on API side. And leave it up to the frontend how they will deliver it - whether it creates copy or it requests that from the user.

But I think I came to this realisation too late since want to release today. So I won't reorganise it now and just merge it, since current version is already well tested and reviewed.

Issue #12792, opened 2026-05-27
[Emails] | User invitation email: Accept/Decline buttons invisible in Outlook (white-on-white)


blesildaramirez:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The **Accept** and **Decline** links (styled as buttons) in the user-invitation email render as blank white space in Outlook (Classic) / Microsoft Exchange clients. The links are still there - you can hover and click them - but they're not visible to the user.

Issue seems to be related to the styling changes from https://github.com/pkp/pkp-lib/issues/11803. Before 11803, the `<style>` block in the email did not always make it through to the recipient - TinyMCE and the template processing would sometimes strip it out, so the links/buttons would just fall back to Outlook’s default hyperlink styling (blue underlined text, not ideal but still visible).

With #11803, we now re-inject the CSS after processing, so the style block reliably makes it into the final email. The issue now is that Outlook Desktop applies its own built-in Hyperlink style to anchors. It keeps the `color: white` from our styles, but drops the background-color on the `<a>`, which results in white text on a white background.

Related Microsoft Q&A with the same issue:
https://learn.microsoft.com/en-us/answers/questions/5617199/html-email-buttons-not-displaying-background-in-ou

From the [report](https://github.com/pkp/pkp-lib/issues/12332#issuecomment-4430029775):
<img width="763" height="354" alt="Image" src="https://github.com/user-attachments/assets/615bdbb4-1bf6-4f74-9d3d-f44220918027" />

### Steps to Reproduce

Must have Classic Outlook for testing.
1. Log in to OJS as a Journal Manager (or any user with permission to invite).
2. Go to Users & Roles -> Users.
3. Click "Invite to a role"
4. Enter any email.
5. Assign at least one role (e.g. Reviewer or Author).
6. Submit the invitation.
7. Open Mailpit (used in test environments). Download the email as raw file.
8. Open the downloaded email using Outlook.
9. Notice the links/buttons don't have proper background-color sytling.

### Expected Result

The confirmation links/buttons to accept or decline the invite are rendered correctly with green and red backgrounds.

### Actual Result

The links/buttons’ background colors are not rendering correctly and appear white, which matches the white button text, making them invisible.

### Application Version

OJS, OMP, OPS v3.5 & 3.6

---

**PRs**

**Branch stable-3_5_0**
OJS: https://github.com/pkp/ojs/pull/5646
OMP: https://github.com/pkp/omp/pull/2394
[...]

PR #12957, opened 2026-06-26
[pkp-lib][stable-3_5_0] #12792 Add bgcolor to markup for the clients that can't render style

defstat:
### **Before Fix**
image

### **After Fix**
image

## **Before Fix**
image

## **After Fix**
image


touhidurabir:
@defstat this seems to be working (I was only able to test in outlook 2019)

CleanShot 2026-06-29 at 18 01 43@2x

However, I think we need migration for cases when a default template is already stored for given locale in the `email_templates_default_data` table which was the case for me at first . So in my local setup for OJS 3.5 latest checkout, even after pulling this PR , it first render with same issue as was pulling the email body from DB which had the previous body instead of new one from the PR

Also in that case, we need to consider about the customised template also which get stored in the `email_templates_settings` .

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
```diff
@@ -0,0 +1,127 @@
...
+            ->get()
+            ->each(function ($row) {
+                $new = $this->rewriteOrWarn($row->body, "default body, locale={$row->locale}");
+                if ($new !== null) {
+                    DB::table('email_templates_default_data')
+                        ->where('email_key', self::EMAIL_KEY)
+                        ->where('locale', $row->locale)
+                        ->update(['body' => $new]);
+                }
+            });
```
touhidurabir, line 42: this works nicely . But we can also just remove the current default and re-install the default like from the `I5716_EmailTemplateAssignments::modifyEditorAssignTemplate()` . However that will require to update all the locales of `emails.userRoleAssignmentInvitationNotify.body` which I think we need to do anyway .

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
```diff
@@ -0,0 +1,127 @@
...
+     * one (the canonical .po order) and replace the pair with a single
+     * <table> that has two rows — one per button. 
+     * Localized labels and the Smarty {$...Url} variables are preserved.
+     *
+     * If the two buttons are not adjacent (hand-edited), the pattern
+     * doesn't match and rewriteOrWarn() emits a warning instead of writing
+     * something inconsistent with the .po template.
+     *
+     */
+    private function rewriteBody(string $body): string
```
touhidurabir, line 110: I think we are fine here with return type `string` but `preg_replace_callback` can return `string, array or null` as per the manual . perhaps a string cast just to be safe ?

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return preg_replace_callback(
+            '#<p>\s*<a[^>]*href=([\'"])\{\$acceptUrl\}\1[^>]*>([^<]+)</a>\s*</p>'
```
touhidurabir: tinymce like editor sometimes push some styles (not sure if we sanitize it on saves) on save so `<p>..</p>` can become `<p style='...'>...</p>` . So just to be safe we can replaces the opening `p` from `<p>` --> `<p[^>]*>`

locale/en/emails.po
```diff
@@ -613,8 +613,10 @@ msgstr ""
 "        <div>{$rolesAdded}</div>"
 "        <p>On accepting the invite, you will be redirected to {$contextName}.</p>"
 "        <p>Feel free to contact me with any questions about the process.</p>"
-"        <p><a href='{$acceptUrl}' class='btn btn-accept'>Accept Invitation</a></p>"
-"        <p><a href='{$declineUrl}' class='btn btn-decline'>Decline Invitation</a></p>"
+"        <table class='btn-wrap' role='presentation' border='0' cellpadding='0' cellspacing='10'>"
+"            <tr><td class='btn-cell btn-accept'  bgcolor='#28a745'><a class='btn' href='{$acceptUrl}'>Accept Invitation</a></td></tr>"
+"            <tr><td class='btn-cell btn-decline' bgcolor='#dc3545'><a class='btn' href='{$declineUrl}'>Decline Invitation</a></td></tr>"
+"        </table>"
```
touhidurabir, line 619: I think we need to update all other locale files to update this one also otherwise at fresh install of 3.5.0-5 with different locale say `fr_ca`, `pt` etc, it can be stored without the table structure .

touhidurabir (commented):
@defstat please see the comments . I tested and it does works nicely without any issue . 

Probably the most important part left to decide if we need to update all other locales for that email locale key or not .

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
defstat: Changed to `return (string) preg_replace_callback(`

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
defstat: Done. Applied as suggested. 

Small note: `<p[^>]*>` will also match `<para>`, `<phrase>`, etc., which is harmless here because the Smarty `{$acceptUrl}`/`{$declineUrl}` part of the pattern is unique. If you'd prefer strict `<p>` matching, we could use
`<p\b[^>]*>` (`\b` = word boundary). 

Happy either way.

locale/en/emails.po
defstat: @touhidurabir Good point - yes, if we don't update the other locale `.po` files, fresh installs and upgrades will produce different results for non-English locales:

Two options for resolving the inconsistency:

1. Update all locale `.po` files in this PR - install and upgrade would produce identical results, but would it conflict with Weblate translations, and the translation process in general? Generally I don't change other locale files except en. I have added the `#, fuzzy` keyword for `msgid "emails.userRoleAssignmentInvitationNotify.body"` for all the other locales though, to be consistent.

2. Let translators update the other locales. Even though I understand that it is not really a translation change. 

I'd lean toward option 2 (the current PR scope) unless you prefer the locale-files update. Happy to do either.

classes/migration/upgrade/v3_5_0/I12792_FixInvitationEmailButtons.php
defstat: This is related and affected by the decision here https://github.com/pkp/pkp-lib/pull/12957#discussion_r3497499139

locale/en/emails.po
touhidurabir: > but would it conflict with Weblate translations, and the translation process in general?

will it cause problem with `Weblate` ? previously i did something like that, updating all other locales with `#, fuzzy` and it got merge but I can not confirm if doing so will cause any issue or not . 

> Update all locale .po files in this PR - install and upgrade would produce identical results

IMO I think this would be better approach so that it will get fixed in one shot otherwise it will persist for other locales and need some sort of manual intervention even after the upgrade/install , please correct me if I am mistaken . 

But if issue arise with weblate, then we need to re-think . @asmecher @bozana can you confirm .

touhidurabir: Update : sorry , in last comment I mention that I prefer option 2 but it should be 1 , updated the comment

Issue #12789, opened 2026-05-25
[Task & Discussions] | Editorial View - Submission Page, 500 error thrown when adding new Task / Discussion with author who has been disabled


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

This happens for templates that are automatically added as well as existing tasks if you try to edit or view them in some way.  As soon as the participants list has to be recalled it will throw the error.

Found while testing email variables to authors who would be disabled.  But root cause found to not have to do with the Email variables found when Inserting Content into a T/D comments area. There may be additional steps to ensure that this new {$authors} and {$authorsShort} variable is not blocked by users with author role that have been disabled.

### Steps to Reproduce

1. Precondition : Must have disabled the author user associated with the submission (Users & Roles > 3 dots > disabled user)
2. Login with Admin, JM, JE role (access to T/D and Users & Roles)
3. Navigate to the editor dashboard submission with author who was disabled > select View
4. In the Tasks and Discussion Table select `Add` 

### Expected Result

1. Participants should load regardless of the users in the disabled state.
2. Those template variables referencing authors should only render those authors in the contributor list AND particpant list

### Actual Result

500 Error

<img width="1047" height="154" alt="Image" src="https://github.com/user-attachments/assets/1f7a8d90-7caf-455a-a81f-6f5b0078e74f" />

`"PKP\\API\\v1\\submissions\\tasks\\EditorialTaskController::PKP\\API\\v1\\submissions\\tasks\\{closure}(): Argument #1 ($user) must be of type PKP\\user\\User, null given"`

### Application Version

OJS, OMP, OPS 3.6(main)

### Logs

[...]

PR #12971, opened 2026-06-30
pkp/pkp-lib#12789 Fix error thrown when adding new Task/Discussion with disabled participant

taslangraham:
For pkp/pkp-lib#12789


blesildaramirez (approved):
Looks good, thanks @taslangraham !

Issue #12950, opened 2026-06-25
[OAI Metadata] Add version and relations in metadata formats exposed via OAI


kaitlinnewson:
**Description:**

For all metadata formats that are exposed via OAI, add information about the article version and relations (e.g. if an article is a version of another) to the exported metadata, following the requirements for each standard.

Relations are express from the newer version, pointing to the last sibling. Forward relations have limited usefulness, as past records are not commonly re-sent/re-harvested, and we don't want to force re-harvesting of all old records for this metadata change. Harvesters will be able to determine the structure by examining the newer record relations.

Sub-issue of #12922.

**Steps to test the changes:**

TBA (varies by standard).

**Solution:**

The updated metadata will be exposed in the OAI endpoints.

**What application are you using?**
  OJS, OMP, OPS 3.6/main

**Additional information**

N/A

## Pull Requests

- OJS main: https://github.com/pkp/ojs/pull/5656
- OPS main: https://github.com/pkp/ops/pull/1337
- OMP main: https://github.com/pkp/omp/pull/2398
- pkp-lib main: https://github.com/pkp/pkp-lib/pull/13001
- jatsTemplate main: https://github.com/pkp/jatsTemplate/pull/108

kaitlinnewson:
Hi @bozana, this is ready for your review.

bozana:
Hi @kaitlinnewson, I reviewed it, see my comments in the PRs. Thanks a lot!

kaitlinnewson:
Hi @bozana, this is ready for another review. Based on your feedback and looking at how others handle it, some of the key changes include:

- Only the newer version shows the backwards relation (not bi-directional/no forward relations, and documented in the ticket description). 
- Only the last sibling is expressed, not all of the relations (e.g. v3 shows v2, but not v1) - while these are all technically relations, this seems to be the more common design.
- Version stages are considered in determining the newer version.

bozana:
Hi @kaitlinnewson, I reviewed the PRs and left a few comments. Looks great! Thanks a lot!

kaitlinnewson:
Closing as completed and will include in testing for #12922

PR #13001, opened 2026-07-07
pkp/pkp-lib#12950 get version relations and types for publications

kaitlinnewson:


classes/publication/Repository.php
```diff
@@ -915,6 +916,100 @@ public function getVersionString(Publication $publication, ?Submission $submissi
...
+    }
+
+    /**
+     * Determine how a publication relates to a sibling version, using the DataCite
+     * version relationType vocabulary.
+     */
+    protected function getVersionRelationType(Publication $publication, Publication $related): VersionRelationType
+    {
+        $order = [$publication->getData('versionMajor'), $publication->getData('versionMinor')]
+            <=> [$related->getData('versionMajor'), $related->getData('versionMinor')];
```
bozana: Here we do not consider versionStage

classes/publication/Repository.php
```diff
@@ -915,6 +916,100 @@ public function getVersionString(Publication $publication, ?Submission $submissi
...
+                $relationType = $this->getVersionRelationType($publication, $sibling);
+                return (object) [
+                    'publicationId' => $sibling->getId(),
+                    'versionStage' => $sibling->getData('versionStage'),
+                    'versionString' => $this->getVersionString($sibling, $submission, $context),
+                    'doi' => $doiVersioning ? ($sibling->getDoi() ?: null) : null,
+                    'doiUrl' => $doiVersioning ? ($sibling->getData('doiObject')?->getResolvingUrl()) : null,
+                    'datePublished' => $sibling->getData('datePublished'),
+                    'relationType' => $relationType,
+                    'updateType' => $this->getRelationUpdateType($relationType, $publication, $sibling),
```
bozana: `updateType` is a property stored on a single publication, describing how that version amends whatever came immediately before it. However here we do not consider only the version that immediately came before, but all siblings.
Maybe to only populate `updateType` when $sibling is actually the adjacent version (immediately preceding/following in that stage's sequence), and leave it null otherwise?
🤔

bozana (commented):
Hi @kaitlinnewson, I left a few comments, but my biggest concern is whether we should have only one-directional relationships, from newer to older versions, rather  than bidirectional. At least in JATS.

Here are my thoughts:
This serves the OAI interface above all, and OAI is incrementally harvested: I believe old versions are already harvested, and a record is only reconsidered if its datestamp (last_modified) changes. When a new version is created, the datestamp of older versions is not updated. So anyone who already harvested the old record will never see a forward relation added to it later.
However, it seems OpenAIRE requires bidirectional relationships, when applicable (see https://guidelines.openaire.eu/en/latest/data/field_relatedidentifier.html#d-relatedidentifier) :-((( And F1000 uses it too. So maybe we do need it :-(((

Closely connected to this is how we'll implement record-per-version in the OAI interface — whether we key the datestamp to the publication's last_modified (which I think would be correct), or still to the submission's (which would keep returning all of a submission's versions whenever any one of them changes, which is maybe useful for this backward relationships problem). Thus, depending on what we decide here, maybe we need to update the previous versions' last_modified date, when a new version is published.

Regarding JATS specifically: older versions are typically already published and indexed elsewhere (PMC, Zenodo, DOAJ, CNKI, Portico, etc.), so I don't think we should re-index them just because of a forward relation. Interestingly, I checked F1000's actual JATS output for a versioned article, and they don't include any version related-article at all there. A backward-only relation still seems useful, I think, so that indexers can use it to connect the versions.

(Like OpenAIRE, DataCite guidelines also define both directions — but DataCite achieves that via an active deposit/update call, not passive harvesting.)

What do you think?

classes/publication/Repository.php
```diff
@@ -914,6 +916,130 @@ public function getVersionString(Publication $publication, ?Submission $submissi
...
+     */
+    public function getVersionRelation(Publication $publication, Submission $submission, Context $context): ?object
+    {
+        $doiVersioning = (bool) $context->getData(Context::SETTING_DOI_VERSIONING);
+
+        if ($doiVersioning) {
+            // getLatestMinorPublicationsForDoiDeposit() already filters to published, DOI-bearing versions.
+            $candidates = collect(
+                Repo::doi()->getLatestMinorPublicationsForDoiDeposit($submission->getData('publications') ?? collect())
+            )->flatMap(fn (array $publicationsByMajor): array => array_values($publicationsByMajor));
```
bozana: I think `Repo::doi()->getLatestMinorPublicationsForDoiDeposit() `should not be used here: in the OAI we will consider all last minor versions for a stage and major, no matter if they have DOI, or are ready for DOI deposit. So I think we should get candidates in the same way here i.e. consider all last minor versions for a stage and major.

classes/publication/Repository.php
```diff
@@ -914,6 +916,130 @@ public function getVersionString(Publication $publication, ?Submission $submissi
...
+                }
+            }
+            $candidates = collect(array_values($latestByMajor));
+        }
+
+        // Exclude the publication itself, and (with DOI versioning) any version that shares its DOI.
+        $siblings = $candidates
+            ->reject(
+                fn (Publication $sibling): bool => $sibling->getId() === $publication->getId() ||
+                    ($doiVersioning && $sibling->getDoi() && $sibling->getDoi() === $publication->getDoi())
```
bozana: The case where a candidate (last minor for a stage and major) would have the same DOI is only in case something went wrong, or editor edited a DOI so that they are the same.
However, maybe in that case, when it happens that two major versions have the same DOI, maybe we should however consider them, but use URL for the relationship rather than the DOI. I believe also in this case that version would be considered via OAI, so even if there is identifier-quality problem, it is not a predecessor-identification problem. So maybe to have this "is DOI the same" check later, wen providing doi and doiURL in the return object?

bozana (commented):
Hi @kaitlinnewson, a few comments/questions. Thanks a lot!

Issue #12821, opened 2026-05-28
[Activity log] | userGroupName field not populating


1StuJB:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description. Example of a title: [Peer Review] | Reader View - Article landing page, Missing dynamic values

### Description
Two bugs same problem.
Acceptance Criteria:

- [ ] When Assigning a participant the var {$userGroupName} in activity log notification is rendered
- [ ] When removing a participant the var {$userGroupName} in activity log notification is rendered

Please see screenshot - the reference field isn't populating properly.

<img width="1928" height="300" alt="Image" src="https://github.com/user-attachments/assets/530c86cd-1ff4-48c5-86f4-6645d39cf2da" />

### Steps to Reproduce
Assign participant scenario

1. Precondtions : Must have user added as a participant that can access Activity Log
2. Login with User/Role (JM, JE, admin)
3. Navigate to submission
4. Click on `Assign` in the Participants table
5. Select a role from dropdown > Author , Copyeditor, JE etc..
6. Select `Ok` button after filling in the rest of the email form.
7. Select Activity Log

Remove participant scenario
1. Precondtions : Must have user added as a participant that can access Activity Log
2. Login with this User/Role (JM, JE, admin)
3. Navigate to submission
4. Click on 3 dots > `Remove` in the Participants table
5. Select Activity Log

### Expected Result

The notifcaition in activity log should now show the correct value

### Actual Result

Unrendered variable name

Tribunal33:
Confirmed.  Added steps to reproduce.

Vitaliy-1:
PR: https://github.com/pkp/pkp-lib/pull/13007
tests: https://github.com/pkp/ojs/pull/5657

Vitaliy-1:
@bozana, can you take a look at the PR?

Tribunal33:
@Vitaliy-1 when closing please provide evidence that it is working as expected please.

1StuJB:
Hi @Tribunal33 and @Vitaliy-1, thanks for this, however I can still see this is throwing up errors in some article logs. Please check :)

<img width="2262" height="404" alt="Image" src="https://github.com/user-attachments/assets/c3c517fb-7498-4250-86ea-4e0ef6e3119c" />

Vitaliy-1:
That's likely the data of the instance that already had missing data during 3.6 development cycle. Is there a need for additional migration between these 3.6 snapshots?

mreiko:
[...]

PR #13007, opened 2026-07-08
pkp/pkp-lib#12821 Add localized user full name and user group to subm…

Vitaliy-1:
…ission event log

### DIFF
--- a/locale/en/submission.po
+++ b/locale/en/submission.po
@@ -858,7 +858,7 @@ msgid "submission.event.participantAdded"
 msgstr "{$userFullName} ({$username}) was assigned to this submission as a {$userGroupName}."
 
 msgid "submission.event.participantRemoved"
-msgstr "\"{$userFullName}\" ({$username}) is removed as a {$userGroupName}."
+msgstr "{$userFullName} ({$username}) was removed from this submission as a {$userGroupName}."
 
 msgid "submission.event.backToEventLog"
 msgstr "Back to Event Log"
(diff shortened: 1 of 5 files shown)


locale/en/submission.po
```diff
@@ -855,7 +855,7 @@ msgid "submission.event.participantAdded"
 msgstr "{$userFullName} ({$username}) was assigned to this submission as a {$userGroupName}."
 
 msgid "submission.event.participantRemoved"
-msgstr "\"{$userFullName}\" ({$username}) is removed as a {$userGroupName}."
+msgstr "{$userFullName} ({$username}) was removed from this submission as a {$userGroupName}."
```
bozana, line 862: the other locales need to be marked fuzzy -- i think there is a cli tool for it now

schemas/eventLog.json
```diff
@@ -304,9 +304,17 @@
...
 			]
 		},
+		"userGroupName": {
+			"type": "string",
+			"description": "The name of the role the user was assigned or unassigned to",
+			"multilingual": true,
+			"validation": [
+				"nullable"
+			]
+		},
```
bozana, line 314: Here https://github.com/pkp/pkp-lib/blob/main/api/v1/submissions/tasks/resources/TaskResource.php#L90: does the getData('userGroupName') vs getLocalizedData('userGroupName') inconsistency (array vs string) matter there? -- if it is EventLog at all 😅

bozana (commented):
Hi @Vitaliy-1 
A comment and a question (becasue I am not sure, so better to ask) :-)
Else, it can be merged.

schemas/eventLog.json
Vitaliy-1: Yeah, should be `->getLocalizedData('userGroupName')`

PR #13012, opened 2026-07-08
Add DB-based funder facet search

bozana:
s. https://github.com/pkp/pkp-lib/issues/12392


bozana:
will be merge here: https://github.com/pkp/pkp-lib/pull/13017. thus closing.

Issue #9303, opened 2023-09-13
Date format isn't properly localized in OJS 3.4


jonasraoni:
**Describe the bug**
On OJS 3.4, a function to convert the [deprecated date format](https://www.php.net/manual/en/function.strftime.php) to the [new one](https://www.php.net/manual/en/datetime.format.php) was introduced (`\PKP\core\PKPString::getStrftimeConversion()`), and it's known to affect at least the Smarty templates.

**Smarty:** Depending on some factors, the Smarty's `date_format` modifier might use internally the `strftime` (deprecated) or the `date` function, **whereas the latter isn't localized**. One of the factors is the presence of a deprecated date format (e.g. `"%B"`), which happens at OJS 3.3, but not anymore on OJS 3.4 (due to the method mentioned previously).

Once you change the language to anything else than English, the localization for the interface will change, but localizable dates pieces will be kept in English, as shown in the screenshot below.

**What application are you using?**
OJS 3.4

**Additional information**
![image](https://github.com/pkp/pkp-lib/assets/361921/eb823ba3-5ba4-441a-8a2a-f38793bdfd74)

asmecher:
Smarty currently either uses the deprecated `strftime` or non-localized `date`, and that's apparently not scheduled for fixing until Smarty adds PHP9 support; see https://github.com/smarty-php/smarty/issues/810 for details.

jonasraoni:
Yeah, we might attempt to fix it on our end, but it will require the extension they've mentioned/3rd-party package.

asmecher:
We already have Carbon in our Composer lockfile; it's probably a good candidate.

diegoabadan:
A customer with OMP 3.4 reported the same problem.

asmecher:
Here are some PRs to *partially* fix the issue (in cases where PHP's `date()` was being used to format dates):

- https://github.com/pkp/ojs/pull/4415
- https://github.com/pkp/pkp-lib/pull/10348

I didn't replace all cases of `date()`, just the ones where date formats were supplied that might include translatable content (e.g. month names). I've tested each case individually, since the PRs are for `stable-3_4_0`; they'll need cherry-picking forward to `main`.

asmecher:
@defstat, would you mind reviewing [these PRs](https://github.com/pkp/pkp-lib/issues/9303#issuecomment-2316494610)?

asmecher:
[...]

PR #13016, opened 2026-07-09
fix: preserve DateTime timezone when building monthly statistics chart labels

### LINKED ISSUE #10813: Wrong translation between actual date of data point in statistics views and date label
**Describe the bug**
In the statistics display (e.g. /stats/publications/publications, daily view, last 30 or 90 days), the displayed date range does not  fit the selected date range (see screenshots). It may be shifted by one day.

The user is given the impression that a day may be missing. May be confusing.

This might be due to a conversion error between the effective date in the data and the label of the data, e.g. because of the selected time zone. See "to reproduce".

**To Reproduce**
All statistics have been calculated up to the end of the selected date range and are available.

1. Time zone of the application has been set to Europe/Zurich in config.inc.php
2. Access logs and usage event logs record in the date/time of the selected time zone, not UTC.
1. E.g. date range is chosen 2024-12-16 - 2025-01-15 (last 30 days).
2. Abszissa shows a range of 15 December 2024 - 14 January 2025 (screenshot1).  Correct should be 16 December 2024 - 15 January 2025.
3. First data point shows a date of 15 December 2024 (screenshot2). Correct should  be 16 December 2024 
4. Last data point shows a date of 14 January 2025 (screenshot3). Correct should be 15 January 2025.
5. Add a var_dump($timeline) statement after the definition of the $timeline variable, around https://github.com/pkp/pkp-lib/blob/1b314678400fad3f9aa1d01488d817f2a3075916/pages/stats/PKPStatsHandler.php#L253
6. Inspect the dumped timeline array - this shows that the dates are indeed correct, but the labels are wrong, shifted by one day, e.g., 
`[0]=> array(3) { ["date"]=> string(10) "2024-12-16" ["label"]=> string(16) "15 December 2024" ["value"]=> string(4) "2081" } [1]=> array(3) { ["date"]=> string(10) "2024-12-17" ["label"]=> string(16) "16 December 2024" ["value"]=> string(4) "1887" }`
 See screenshot4. May this be due to a wrong conversion? 

OJS 3.4.0-8 (and previous OJS 3.4 versions)
PHP 8.2

**Additional information**
@bozana 

Image
Image
Image
Image

jomeridi:
### Problem

When OJS runs in a timezone with a positive UTC offset (e.g. `Europe/Madrid`,
UTC+2 in summer), the monthly statistics chart labels are shifted one month
backwards. June data appears labelled "May 2026", May appears as "April 2026",
and so on across all months. The data stored in the database is correct; only
the chart presentation is affected.

### Environment

- OJS: 3.5.0-4 and 3.5.0-5 (confirmed via code inspection — same line present)
- PHP: 8.3
- Server timezone: `Europe/Madrid` (UTC+2 in summer)
- OJS locale: `es_ES`

### Steps to reproduce

1. Set `time_zone = "Europe/Madrid"` (or any positive-offset timezone) in
   `config.inc.php`.
2. Navigate to **Statistics → Publications → Files**, select the **Monthly**
   interval and a date range that includes at least one complete past month.
3. Hover over any data point on the chart.

**Expected:** tooltip reads "June 2026" for data belonging to June.

**Actual:** tooltip reads "May 2026" — one month behind the actual data period.

### Root cause

In `classes/services/PKPStatsServiceTrait.php`,
`getEmptyTimelineIntervals()` builds chart labels like this (introduced by
commit `22c03902`, fix for #9303):

```php
'label' => (new \Carbon\Carbon($startDate->getTimestamp()))
    ->locale(Locale::getLocale())
    ->translatedFormat($labelFormat),
```

`$startDate` is a `DateTime` object created from a date string such as
`'2026-06-01'`. When the PHP timezone is `Europe/Madrid` (UTC+2), PHP
interprets that string as **midnight local time**, which corresponds to
`2026-05-31 22:00:00 UTC`. Calling `getTimestamp()` returns that UTC epoch
value, and `new \Carbon\Carbon(int)` constructs the instance in UTC — so
Carbon sees **31 May** and the label reads "May 2026" instead of "June 2026".

In short: `getTimestamp()` discards the timezone that `$startDate` already
carries, and the Carbon constructor does not recover it.

### Fix

`Carbon::instance()` accepts a `DateTime` object directly and preserves its
timezone, so the date is interpreted in the correct local context:

```php
'label' => \Carbon\Carbon::instance($startDate)
    ->locale(Locale::getLocale())
    ->translatedFormat($labelFormat),
```

`$startDate` is already a `DateTime` object with the correct timezone. There
is no need to round-trip through a Unix timestamp. `Carbon::instance()` is the
idiomatic way to wrap an existing `DateTime` without losing its timezone.

### Scope

A grep over the entire `lib/pkp/classes/` directory confirms this is the only
place where `getTimestamp()` is passed to a Carbon constructor. No other
occurrences of this pattern were found.

### Verification

The fix was applied temporarily in a production-equivalent environment
(`Europe/Madrid`, OJS 3.5.0-4) and validated across four chart configurations,
then reverted:

| View | Before fix | After fix |
| --- | --- | --- |
| Monthly files, Jan–Jun range | "May 2026" ✗ | "June 2026" ✓ |
| Monthly files, Jan–Jul range | "June 2026" ✗ | "July 2026" ✓ |
| Monthly abstract views | Labels off by one ✗ | Labels correct ✓ |
| Daily view | Unaffected ✓ | Unaffected ✓ |

Behaviour in UTC is unaffected: when `time_zone = UTC`, `getTimestamp()`
returns a value already in UTC, so both the old and new code produce the same
result. No regressions expected or observed in UTC environments.

### Notes on automated tests

The PKP test suite does not appear to include unit tests for
`getEmptyTimelineIntervals()`. A unit test covering the timezone case would
prevent this class of regression from recurring. Example case:

- Given: `$startDate = new DateTime('2026-06-01')` in timezone `Europe/Madrid`
- Expected label: `"June 2026"` (or locale equivalent)

If the project has a preferred location for service-layer unit tests, adding
such a test alongside this fix would be straightforward.

### Related

- Fixes #10813
- Regression introduced by commit `22c03902` (intended fix for #9303)


asmecher:
@jomeridi, it looks like this may have been submitted against the wrong branch; the changeset is huge!

Also, can I confirm that if an AI agent was used in preparation of this, that a human reviewed it before submitting? (See [this discussion](https://github.com/pkp/pkp-lib/discussions/12487) on AI and contributions.)

jomeridi:
Thanks for the review and for the heads-up about the branch.

**Wrong branch**: Apologies for that. I submitted against the wrong base branch, which is why the changeset looks huge. The actual change is a single line in PKPStatsServiceTrait.php. Could you confirm which branch I should target for a 3.5.x fix? I'll close this PR and open a new one against the correct branch with just that one-line change.
**AI disclosure**: Yes, I used an AI assistant (Claude) to help diagnose the root cause and identify the fix. I reviewed the change myself before submitting — it is a one-line substitution from new \Carbon\Carbon($startDate->getTimestamp()) to \Carbon\Carbon::instance($startDate), which preserves the DateTime object's timezone rather than converting through a UTC timestamp. The logic is straightforward and verifiable independently of any AI tooling. I confirmed the fix in our production environment before opening the PR.

asmecher:
@jomeridi Thanks, that sounds good -- I'll review the other PR when I get a chance.

Issue #13018, opened 2026-07-09
Fix ROR API to stop trusting client-submitted institution data


bozana:
POST /api/v1/rors (PKPRorController::addOrEdit()) is reachable by any user with an Author role in the context, and Author accounts are self-registered. The endpoint performs an upsert (keyed on the ror field) using the name, displayLocale, and isActive values sent in the request body, with no server-side check against the authoritative ror.org data. Only the ror field itself is regex-validated (and that regex is currently unanchored, so it doesn't reliably constrain the field to a well-formed ROR URL either).

Root cause: the endpoint was designed to let the frontend push a record it fetched from the live api.ror.org API into our local cache (so affiliation data stays current between monthly syncs), but the backend never re-verifies that content — it trusts whatever the client sends.

Proposed fix: change addOrEdit so the client can only supply the ror identifier; the server performs its own request to api.ror.org for that ID and stores only what it gets back, discarding any content fields in the client payload. Also anchor the ror validation regex and add rate limiting on the route, since the caller may be a freshly self-registered account.

---

PRs:
stable_3_5_0:
- https://github.com/pkp/pkp-lib/pull/13020
- https://github.com/pkp/ui-library/pull/927
- https://github.com/pkp/ojs/pull/5665 (only submodule update)
- https://github.com/pkp/omp/pull/2401 (only submodule update)
- https://github.com/pkp/ops/pull/1341 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/13019
- https://github.com/pkp/pkp-lib/pull/13029
- https://github.com/pkp/ui-library/pull/926
- ~~https://github.com/pkp/ojs/pull/5664~~ (only submodule update)

bozana:
Hi @asmecher and @touhidurabir, could please one of you review the changes here?
Thanks a lot!

asmecher:
Thanks, @bozana, this looks good -- but as it's an end-user-initiated transaction with an external service, it probably could use some rate limiting. Have a look at how `PKPSiteSecurityForm` does this [with the Laravel toolset](https://github.com/pkp/pkp-lib/blob/main/classes/components/forms/site/PKPSiteSecurityForm.php).

bozana:
Oh, yes, this is what I wanted to ask Touhidur, how to best solve. I will try and then lets see if correct so 😅 
Thanks!

bozana:
It seems it is only available in main, no?  What would be a solution for stable-3_5_0?

[...]

PR #13020, opened 2026-07-09
pkp/pkp-lib#13018 Fix ROR API to stop trusting client-submitted insti…

bozana:
…tution data

see https://github.com/pkp/pkp-lib/issues/13018

### DIFF
--- a/api/v1/rors/PKPRorController.php
+++ b/api/v1/rors/PKPRorController.php
@@ -20,9 +20,11 @@ namespace PKP\API\v1\rors;
 use APP\core\Application;
 use APP\facades\Repo;
 use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\Exception\RequestException;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
+use Illuminate\Support\Facades\RateLimiter;
 use Illuminate\Support\Facades\Route;
 use PKP\core\PKPBaseController;
 use PKP\core\PKPRequest;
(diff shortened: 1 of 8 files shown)


touhidurabir (approved):
@bozana look fine to me . I just have few suggestions but up to you if you want to implement as 
1. for every request we make a validation check in ror api . however what if that already exists and we want to check first . that can/may reduce the ror api hit and reduce the chances to hit the limit too quickly . seems there is `PKP\ror\DAO::getByRor` to check it
2.  The API endpoint seems to restricted behind `has.user`, so a stable authenticated identity is always available . so in that case, is it better to do rate limiting by user instead of IP as IP is maskable/changeable . 
3. we show a generic message like wait few seconds but using RateLimiter, we can provide how long to wait via `RateLimiter::availableIn($key)` . also does ROR doc provide any details about the rate limit and wait time details which we should use instead of our own custom values ? 

None of the above is blocker so up to you .

bozana:
Hi @touhidurabir, I considered your comments:
- Skipped the ror.org lookup when the ROR is already cached locally: this unfortunately removes our earlier intention to keep the ROR entries always actual. But there is a scheduled task that updates the ROR dataset monthly, so it should be OK trade-off.
- Rate limit by authenticated user instead of IP
- Surface actual retry-after time in 429 responses
- Tighten the global limit to stay under ror.org's post-2026 unauthenticated floor: at some point this year ROR will allow only 50 requests per 5 minutes if the user registration ID/key is not provided (s. https://ror.readme.io/docs/rest-api). We plan to add this setting field, s. https://github.com/pkp/pkp-lib/issues/12355, but until this is implemented we should probably assume this lower limit. I adjusted the our global limit to be less than that. I could not find any information about waiting time.

I would then merge once the tests pass. And also port it to main :-)
Thanks a lot! Great review/comments!

PR #13022, opened 2026-07-09
pkp/dev-team#280 return all peer review data linked up the publication tree via `source_publication_id`

taslangraham:
For pkp/dev-team#280

### DIFF
--- a/classes/publication/Repository.php
+++ b/classes/publication/Repository.php
@@ -36,15 +36,12 @@ use PKP\doi\exceptions\DoiException;
 use PKP\facades\Locale;
 use PKP\file\TemporaryFileManager;
 use PKP\log\event\PKPSubmissionEventLogEntry;
-use APP\decision\Decision;
 use PKP\observers\events\PublicationPublished;
 use PKP\observers\events\PublicationUnpublished;
 use PKP\orcid\OrcidManager;
 use PKP\plugins\Hook;
-use PKP\security\Role;
 use PKP\security\Validation;
 use PKP\services\PKPSchemaService;
-use PKP\stageAssignment\StageAssignment;
 use PKP\submission\Genre;
 use PKP\submission\PKPSubmission;
 use PKP\submission\reviewAssignment\ReviewAssignment;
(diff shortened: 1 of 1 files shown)


classes/publication/Repository.php
```diff
@@ -1230,16 +1231,32 @@ public function getReviewDoiItemsGroupedByPublication(array $publicationIds): ar
...
+     * it references via the `source_publication_id`.
+     * This goes up the entire tree created by the usage of `source_publication_id` on each record to ensure all associated publication IDs are returned.
      */
     public function getWithSourcePublicationsIds(array $publicationIds): Collection
     {
-        return $this->getCollector()
-            ->filterByPublicationIds($publicationIds)
-            ->filterWithSourcePublicationIds()
-            ->getIds();
+        $allIds = collect($publicationIds);
```
touhidurabir, line 1242: seems there is not array unique check so a input like [103, 103] can produce result as [103, 103, 102, 101, 100]. is that acceptable ?

touhidurabir (approved):
The method implementation is doing what the doc block has specified, thus approving . along with the unique check comment , I want to point few more stuff as\
 - the method implementation goes up to the parent but never to child . so say  publication id 22 and 23 have source to say 18 . but passing [22] will not produce back result with 23 . 
 - the comment at https://github.com/pkp/dev-team/issues/280#issuecomment-4970747514 mention of tree like structure but a mix of non related publication id will not result in a tree/chain like sequence but more set . So for example with param as `[2, 1, 22, 5]` where publication 2 has source as 1, 22 has source 18 and 5 has not source generate result as [2,1,22,5,18] . 
 
if the above example cases seems acceptable, then all ok .

classes/publication/Repository.php
taslangraham: Good catch. this should be added.

taslangraham:
> The method implementation is doing what the doc block has specified, thus approving . along with the unique check comment , I want to point few more stuff as\
> 
> * the method implementation goes up to the parent but never to child . so say  publication id 22 and 23 have source to say 18 . but passing [22] will not produce back result with 23 .
> * the comment at [pkp/dev-team#280 (comment)](https://github.com/pkp/dev-team/issues/280#issuecomment-4970747514) mention of tree like structure but a mix of non related publication id will not result in a tree/chain like sequence but more set . So for example with param as `[2, 1, 22, 5]` where publication 2 has source as 1, 22 has source 18 and 5 has not source generate result as [2,1,22,5,18] .
> 
> if the above example cases seems acceptable, then all ok .


Hi @touhidurabir, thanks for the review; yes, these are both expected and acceptable behaviours.

Issue #12922, opened 2026-06-18
Add article versions to OAI endpoints when DOI versioning is enabled, and expose AO and PMUR articles


kaitlinnewson:
## Description

In our OAI endpoints, we currently have a single record for each article. With the changes to article versions and the possibility to have a DOI for each version, we need to consider having an OAI record for each article version, as well as reflecting the versions in the metadata formats exposed via OAI if it is supported in the format (DC, JATS, MARC).

Communication will also be needed for Coalition Publica harvesting and updates to the 3.6 change orientation docs.

## Solution

- [x] Implement a record per version for OAI endpoints when DOI versioning is enabled.
- [x] Determine if a setting should be used to enable records per version, or if this should apply to all journals (e.g. will it have a negative impact on other harvesting services that use OAI?) - don't add a setting for now
- [x] Add articles in the AO and PMUR stages to the OAI records - they will either appear separately if DOI versioning is enabled, or will appear if they are the latest version of a record when DOI versioning is not enabled.
- [x] Check how each metadata format supports versioning, e.g. add an article version and indicate relations (see sub-issue)

## Note on disabling/enabling DOI versioning in an active journal

- Turning DOI versioning on does not update datestamps - harvesters must do a full re-harvest to pick up the older versions.
- Turning DOI versioning off withdraws version records silently. The consequence is that a harvester that already collected version identifiers will hold onto them indefinitely; the DOI versioning setting should be treated as one-way once it is enabled.

## What application are you using?
OJS/OMP/OPS 3.6/main

## Test setup

- A journal with DOIs enabled, OAI enabled, and at least one section.
- **Article A** with several published versions: Version of Record 1.0, then 2.0, then a minor 2.1.
- **Article B** with a single published version.
- **Article C** whose only published version is an Author Original (no VoR published yet). Optionally a fourth article with only a Published Manuscript Under Review.

Sample requests:

| Task | Request |
| --- | --- |
| Get OAI landing page (base URL for requests) | `https://{yoursite}/index.php/{journalPath}/oai` |
| List all records | `?verb=ListRecords&metadataPrefix=oai_dc` |
| List identifiers only | `?verb=ListIdentifiers&metadataPrefix=oai_dc` |
| Fetch one record | `?verb=GetRecord&metadataPrefix=oai_dc&identifier={id}` |
| Incremental harvest | `?verb=ListRecords&metadataPrefix=oai_dc&from=YYYY-MM-DD` |

[...]

PR #13028, opened 2026-07-10
pkp/pkp-lib#12922 add OAI record versions for DOI versions and cleanup OAI classes

kaitlinnewson:


classes/oai/OAI.php
```diff
@@ -111,46 +111,34 @@ public function execute()
...
-     * Return OAI record for specified identifier.
-     *
-     * @param string $identifier
+     * Return OAI record for the specified identifier.
      *
-     * @return OAIRecord (or false, if identifier is invalid)
+     * @return ?OAIRecord (null if identifier is invalid)
      */
-    abstract public function record($identifier);
+    abstract public function record(string $identifier): OAIRecord|false;
```
bozana, line 138: `@return` says `null`, the type `false`. 'false' seems to be correct.

classes/oai/PKPOAIDAO.php
```diff
@@ -98,87 +93,94 @@ public function insertToken($token)
...
      * @param array $setIds optional Objects ids that specify an OAI set,
      * in hierarchical order. If passed, will check for the data object id
      * only inside the specified set.
-     *
-     * @return bool
+     * @param int|null $publicationId optional. If passed, restrict to the
+     * record representing that specific object version.
      */
-    public function recordExists($dataObjectId, $setIds = [])
+    public function recordExists(int $dataObjectId, array $setIds = [], int $publicationId = null): bool
```
bozana: it should be `?int $publicationId`, right?

classes/oai/PKPOAIDAO.php
```diff
@@ -98,87 +93,94 @@ public function insertToken($token)
...
      * @param array $setIds optional Objects ids that specify an OAI set,
      * in hierarchical order. If passed, will check for the data object id
      * only inside the specified set.
-     *
-     * @return OAIRecord
+     * @param int|null $publicationId optional. If passed, return the record
+     * representing that specific object version.
      */
-    public function getRecord($dataObjectId, $setIds = [])
+    public function getRecord(int $dataObjectId, array $setIds = [], int $publicationId = null): ?OAIRecord
```
bozana: it should be `?int $publicationId`, right?

bozana (commented):
Hi @kaitlinnewson, just a few type comments here. Thanks a lot!

Issue #13041, opened 2026-07-14
[Schedule Task] | Web based task runner may miss task executing without last run tracking


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

As part of https://github.com/pkp/pkp-lib/issues/9678, moved the schedule task to laravel scheduler . However in the web base task runner (which runs at the end of request life cycle), it does not track the status of a task if that has not run for targeted cron time the last time unless a http request fired at the window of that specified time . This may/can cause the miss of firing of schedule task when running via task runner . 

### Steps to Reproduce

N/A

### Expected Result

Should run even when missed due to no web request at specified time .

### Actual Result

May not run even when missed due to no web request at specified time .

### Application Version

OJS/OMP/OPS 3.5

### Additional Information

see related https://forum.pkp.sfu.ca/t/daily-usage-events-log-not-loading-into-db/97376

Possible solutions : 
1. Use DB storage as system wide settings, only used by task runner
2. use cache to track the last run of task and if miss, run immediately at next run of web based task runner . 

#### Stable-3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/13043
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5673 [TEST ONLY]
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2446 [TEST ONLY]
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1386 [TEST ONLY]

Tribunal33:
Moved bug to task.  No steps to reproduce.

touhidurabir:
@asmecher can you please review the PRs at https://github.com/pkp/pkp-lib/issues/13041#issue-4880717369 .

asmecher:
Thanks, @touhidurabir, and sorry for the delay -- I've added a review [here](https://github.com/pkp/pkp-lib/pull/13043#pullrequestreview-4931969342).

asmecher:
@touhidurabir and I spoke about this today...
- I hadn't understood that the window of execution is actually just 1 minute. If a request fails to trigger the runner during that, it will be missed until the next go-around.
[...]

Issue #9678, opened 2024-01-31
Replace task scheduler by Laravel scheduler


jonasraoni:
**Describe the feature**
The task scheduler syntax isn't very friendly and has caused issues in the past.

For a next big release, it's better to discard the code and move to the [Laravel scheduler](https://laravel.com/docs/10.x/scheduling), which supports better configurations and the usage of a CRON syntax.

**Extra reasons**
If a plugin wants to run scheduled tasks, it needs to provide a `scheduledTasks.xml` file, and also attach itself to the `Acron` plugin. After installing such plugin, the administrator of a journal, which is using CRON instead of the Acron plugin, will also need to add a new entry at `CRON` to execute the plugin's `scheduledTasks.xml`.
- This is a bit counter-intuitive, the Acron shouldn't be any simpler than using CRON after both are setup.
- The "registration" should be unified, the executor just needs to have access to the list of tasks and their execution periods.

This is a follow-up of https://github.com/pkp/pkp-lib/issues/8921
Also considering https://github.com/pkp/pkp-lib/issues/7940 

**Plugins to update**
- [x] [pkp/ojs](https://github.com/pkp/ojs) · [classes/plugins/PubObjectsExportPlugin.php](https://github.com/pkp/ojs/blob/d3214bd8fd88cbb72a6b2156f0531cfbf34e4d1f/classes/plugins/PubObjectsExportPlugin.php#L94)
- [ ] [pkp/usageStats](https://github.com/pkp/usageStats) · [UsageStatsPlugin.inc.php](https://github.com/pkp/usageStats/blob/51b955e2089d5ec64eba6ba945d967f926cd1aab/UsageStatsPlugin.inc.php#L127) [NOT REQUIRED AS FUNCTIONALITY MOVED TO CORE IN 3.4.0]
- [x] [pkp/pln](https://github.com/pkp/pln) · [PlnPlugin.php](https://github.com/pkp/pln/blob/226737802d31cff7870ee6b803cdd41288c588e1/PlnPlugin.php#L107)
- [x] [pkp/crossrefReferenceLinking](https://github.com/pkp/crossrefReferenceLinking) · [CrossrefReferenceLinkingPlugin.php](https://github.com/pkp/crossrefReferenceLinking/blob/6dc8724ab5bd415dc61a4c44d5e0e414934a7b2b/CrossrefReferenceLinkingPlugin.php#L69)
[...]

PR #13043, opened 2026-07-14
pkp/pkp-lib#13041 DB based last run tracker at web based task runner

touhidurabir:
for pkp/pkp-lib#13041 .

### DIFF
--- a/classes/scheduledTask/ScheduleTaskRunner.php
+++ b/classes/scheduledTask/ScheduleTaskRunner.php
@@ -14,20 +14,22 @@
 
 namespace PKP\scheduledTask;
 
-use Throwable;
 use Carbon\Carbon;
-use PKP\core\PKPContainer;
-use Illuminate\Support\Sleep;
-use Illuminate\Support\Collection;
+use Cron\CronExpression;
+use Illuminate\Console\Events\ScheduledTaskFailed;
+use Illuminate\Console\Events\ScheduledTaskFinished;
+use Illuminate\Console\Events\ScheduledTaskSkipped;
+use Illuminate\Console\Events\ScheduledTaskStarting;
 use Illuminate\Console\Scheduling\Event;
 use Illuminate\Console\Scheduling\Schedule;
-use Illuminate\Contracts\Events\Dispatcher;
-use Illuminate\Contracts\Debug\ExceptionHandler;
-use Illuminate\Console\Events\ScheduledTaskFailed;
-use Illuminate\Console\Events\ScheduledTaskSkipped;
 use Illuminate\Contracts\Cache\Repository as Cache;
-use Illuminate\Console\Events\ScheduledTaskFinished;
-use Illuminate\Console\Events\ScheduledTaskStarting;
+use Illuminate\Contracts\Debug\ExceptionHandler;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Sleep;
+use PKP\config\Config;
+use PKP\core\PKPContainer;
+use Throwable;
 
 class ScheduleTaskRunner
 {
@@ -38,7 +40,7 @@ class ScheduleTaskRunner
 
     /**
      * Constructor
-     * 
+     *
      * @param Schedule $schedule The schedule instance.
      * @param Dispatcher $dispatcher The event dispatcher
      * @param Cache $cache The cache store implementation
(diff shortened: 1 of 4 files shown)


classes/scheduledTask/ScheduleTaskRunner.php
```diff
@@ -49,22 +51,94 @@ public function __construct(
...
 
     /**
-     * Run all the due schedule tasks
+     * Run all the due schedule tasks.
+     * 
+     * As schedule task running in web mode, no tracking possible to know if miss on web request.
+     * So use DB store based system to contains the last run mapped to each tasks name which 
+     * used to determine possible miss at next web absed task runner . This is as web request
+     * just may not fired at specified schedule task run time defined and laravel's system
+     * has not mechamism to track it.
```
asmecher:
```suggestion
     * When scheduled tasks are executed at end-of-request, a gap between requests may cause
     * the runner to miss an execution interval. To avoid this, store the last run times by task name
     * and check these to determine whether to execute a task.
```

classes/scheduledTask/ScheduleTaskRunner.php
```diff
@@ -49,22 +51,94 @@ public function __construct(
...
-                $this->dispatcher->dispatch(new ScheduledTaskSkipped($event));
+        $lastRunTimes = ScheduledTaskHelper::getLastRunTimes();
+        $lastRunTimesChanged = false;
+
+        // Frequent, exactly-due events, collected for the sub-minute repeat handling below.
+        $dueFrequentEvents = new Collection();
+
+        foreach ($this->schedule->events() as $event) {
+            // which is defined in APP/PKP scheduler via the `name` as displayable task name
+            $taskName = $event->getSummaryForDisplay();
```
asmecher, line 85: This appears to [potentially draw on the textual description](https://github.com/laravel/framework/blob/0a2201c8a207eb58bb5b6dcb54ce7e2c7aca1dcf/src/Illuminate/Console/Scheduling/Event.php#L811) of the event, which probably isn't what we want to be using/storing as an identifier. Maybe `buildCommand` instead?

asmecher (commented):
I have a couple comments suggesting small tweaks. But a bigger question... This adds a fair amount of code moving us away from a "clean" use of the Laravel scheduler, which was the original motivation behind [#9678](https://github.com/pkp/pkp-lib/issues/9678). I'd like to double-check the risk/effect of this before we deepen our tech debt in the opposite direction.

It looks like [this forum thread](https://forum.pkp.sfu.ca/t/daily-usage-events-log-not-loading-into-db/97376) is the closest we have to a motivating use case. By my understanding -- please correct me if I'm wrong! -- in order for the scheduler to cause problems with a missed interval, a few things have to be true:

- The journal must not receive a request that kicks off the scheduler during the regular interval, and
- Some characteristic of the job has to expect it to be run on every interval, or it has to be important enough to run every interval. (Otherwise it'll catch up with itself, possibly just later than expected, and meanwhile the processing might be delayed.)

Do we have confirmation that this caused the problems on the forum thread? I think it's actually much likelier that a request timeout or something similar caused the problem. I view the `register_shutdown_function` setting as a bad work-around for folks that don't have a Cron scheduler, and nothing more; the installations in the thread appear to be bigger ones, suggesting that a missed request during the scheduler window probably isn't to blame.

Statistics processing in particular is a daily task, and I'd be surprised if an active journal didn't receive a request in a day.

classes/scheduledTask/ScheduleTaskRunner.php
touhidurabir: using the `buildCommand ` will not be good idea or approach . 

What `getSummaryForDisplay()` actually returns here as for a named `CallbackEvent` it returns `$this->description`,  and we names every task  `->name(TaskClass::class)`, core and plugin alike . this can be verified by the command itself as `php lib/pkp/tools/scheduler.php list` which produce list as 
```
0 0 1  * *  PKP\task\StatisticsReport .......................................................................................... Next Due: 2 weeks from now
0 0 1  * *  PKP\task\RemoveUnvalidatedExpiredUsers ............................................................................. Next Due: 2 weeks from now
0 0 10 * *  PKP\task\UpdateIPGeoDB ............................................................................................. Next Due: 3 weeks from now
* * *  * *  PKP\task\ProcessQueueJobs ........................................................................................ Next Due: 5 seconds from now
0 0 *  * *  PKP\task\RemoveFailedJobs ......................................................................................... Next Due: 13 hours from now
0 0 *  * *  PKP\task\RemoveExpiredInvitations ................................................................................. Next Due: 13 hours from now
0 0 1  * *  PKP\task\UpdateRorRegistryDataset .................................................................................. Next Due: 2 weeks from now
0 0 *  * *  APP\plugins\importexport\doaj\DOAJInfoSender ...................................................................... Next Due: 13 hours from now
0 0 *  * *  PKP\task\ReviewReminder ........................................................................................... Next Due: 13 hours from now
0 0 *  * *  PKP\task\DepositDois .............................................................................................. Next Due: 13 hours from now
0 0 1  * *  PKP\task\EditorialReminders ........................................................................................ Next Due: 2 weeks from now
0 0 1  * *  APP\tasks\SubscriptionExpiryReminder ............................................................................... Next Due: 2 weeks from now
0 0 *  * *  APP\tasks\UsageStatsLoader ........................................................................................ Next Due: 13 hours from now
0 0 *  * *  APP\tasks\OpenAccessNotification .................................................................................. Next Due: 13 hours from now
```
here we get the name as a unique identifier of each task . and that what we are using for track . 

what use of `buildCommand` will give us something like `/dev/null' 2>&1` as event task is registered with `$schedule->call(...)`, so they're all `CallbackEvents` which falls through to `CommandBuilder::buildForegroundCommand()` (without any override and we need none) with `command = null` .

touhidurabir:
> I have a couple comments suggesting small tweaks. But a bigger question... This adds a fair amount of code moving us away from a "clean" use of the Laravel scheduler, which was the original motivation behind [#9678](https://github.com/pkp/pkp-lib/issues/9678). I'd like to double-check the risk/effect of this before we deepen our tech debt in the opposite direction.
> 
> It looks like [this forum thread](https://forum.pkp.sfu.ca/t/daily-usage-events-log-not-loading-into-db/97376) is the closest we have to a motivating use case. By my understanding -- please correct me if I'm wrong! -- in order for the scheduler to cause problems with a missed interval, a few things have to be true:
> 
> * The journal must not receive a request that kicks off the scheduler during the regular interval, and
> * Some characteristic of the job has to expect it to be run on every interval, or it has to be important enough to run every interval. (Otherwise it'll catch up with itself, possibly just later than expected, and meanwhile the processing might be delayed.)
> 
> Do we have confirmation that this caused the problems on the forum thread? I think it's actually much likelier that a request timeout or something similar caused the problem. I view the `register_shutdown_function` setting as a bad work-around for folks that don't have a Cron scheduler, and nothing more; the installations in the thread appear to be bigger ones, suggesting that a missed request during the scheduler window probably isn't to blame.
> 
> Statistics processing in particular is a daily task, and I'd be surprised if an active journal didn't receive a request in a day.

@asmecher several key factors to consider why this is not a tech debt but right approach

> [this forum thread](https://forum.pkp.sfu.ca/t/daily-usage-events-log-not-loading-into-db/97376) is the closest we have to a motivating use case

They simply did not understand the issue but there is a hint where things seems off sync . forum post points that moving task frequencey from daily or per min run task which did not run --> that is a hint . However, the case for this patch doesn't rest on it but rests on the scheduler's selection mechanism, which is checkable differently and points the gap .

> Some characteristic of the job has to expect it to be run on every interval, or it has to be important enough to run every interval. (Otherwise it'll catch up with itself, possibly just later than expected, and meanwhile the processing might be delayed.)

Queue jobs has nothing to do here . A schedule task may or not dispatch a job depending on it's requirement/implementation, and once dispatched, it's not scheduler responsibility but job runner or worker responsibility . point is if the schedule task fired or not . 

The point is the mechanism behind the laravel scheduler . It does not maintain a list or cache when or what to be fired . As by nature, it should be run via cron, ping once in every mins, see what to run and fire . and if missed, does not have any way to trace the missed one . And by design logic, it does not need to as it need to run under cron . 

But web based task runner is different as it's a non-conventional way to run schedule tasks . It depends on web request and it's life cycle . considering the once a day task `DepositDois` which is supposed to run `00:00:00–00:00:59` which is the precise behaviour for `->daily()` `(0 0 * * *)` . 

so consider a context receiving a request every in 10 minutes overnight:

| time | runner tick? | `isDue()` evaluates | outcome |
|---|---|---|---|
| 23:52:14 | yes | 23:52 ≠ 00:00 → false | skip |
| 23:53–23:58 | no — 60s cooldown | not evaluated | — |
| 23:58:31 | yes | 23:58 ≠ 00:00 → false | skip |
| **00:07:48** | **yes** | **00:07 ≠ 00:00 → false** | **skip** ← 00:00 already passed |
| 00:19:02 | yes | 00:19 ≠ 00:00 → false | skip |
| every later request that day | yes | false | skip |

The row that matters is 00:07:48. The runner **did** execute there — it loaded the schedule, examined `DepositDois`, and asked "is it 00:00 right now?". It isn't, so it skipped. And because `isDue()` takes no last-run argument, it has no way to know that 00:00 came and went seven minutes earlier with no request to observe it. The same thing happens at 00:19, 03:44, 14:22 and every remaining tick of the day.

So the task isn't deferred to the next tick but it's dropped until tomorrow's 00:00, which is the issue point and gap. There is no mechanism that could catch it up, because nothing ever recorded that a run was owed.

As cron to fire the scheduler once a min, it check against system clock but web runner has none such while depends on web request hit . So if no request within the boundary, it missed which a DB backed last run record fixed . 

and using DB backed record is durable than cache which can be cleared and track data is lost . Laravel keeps no last-run record at all, not in cache, not anywhere. What it does keep in cache is the overlap mutex, and that's TTL-bound and intentionally forgotten after the run as `CacheEventMutex::forget()`, because it handles `is a run in progress?` not `when did one last complete?`.

> I'd be surprised if an active journal didn't receive a request in a day.

By design of laravel' scheduler, the window is 60 seconds, not a day . 

> This adds a fair amount of code moving us away from a "clean" use of the Laravel scheduler, which was the original motivation behind https://github.com/pkp/pkp-lib/issues/9678. I'd like to double-check the risk/effect of this before we deepen our tech debt in the opposite direction.

It do add some code but consolidate to mainly in ScheduleTaskRunner(it handle the web based task runner) and ScheduleServiceProvider which initiate the run of web based task runner . And considering the non-conventional approach of web based task runner, it itself a tech debt . But both the web and corn now run via laravel's scheduler mechanism, so in future, removing the web based task runner is simpler without any possible errors .

asmecher (approved):

taslangraham (approved):

Issue #12702, opened 2026-05-09
[JATS XML] | Upload file - Submission > Publication > Jats XML, unauthorization issue for admin and section editor when uploading JATS files


Tribunal33:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

While testing #10436 it was noticed that uploading JATS XML document as a Admin with Author role or a section editor was being blocked.  You get a 401 unauthorized access notification and you are disconnected from the server (meaning when you refresh the page the nginx credential window pops up).  

I believe a user with other roles should still be able to upload JATS XML.

- [ ] please investigate which roles should be able to upload JATS XML to clarify if certain users with access to the Publication > JATS XML section would not be able to upload a document.

### Steps to Reproduce

1. Preconditions : Must have a user account with Admin + Author role as well as a section editor role and must have a JATS XML file to upload.  (I usually just download one from one submission and upload it to another submission or do a slight change to the downloaded one so you can see the changes when you upload)
2. Login with either user
3. Navigate to any submission > View
4. Select Publication > JATS XML 
5. Select Upload button
6. Select fJATS XML file

### Expected Result

Should be able to upload like you can with JM (without author) and no error.  Additionally when there is an issue it should not lock you out of the server.  A proper warning message would be nice but I think this should not be an issue normally.  If you can view the download button you should be able to upload.  

### Actual Result

401 error in network and you must reenter server credentials upon refresh.

### Application Version

OJS, OPS 3.5.0-4

### Logs

<img width="641" height="56" alt="Image" src="https://github.com/user-attachments/assets/97a4468b-558e-4035-8888-4de91c703dd4" />

#### Stable 3.5.0
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/12996
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5653 [TEST ONLY]

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/13054
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5677 [TEST ONLY]

asmecher:
Deferring from 3.5.0-x; this would be good to fix, but is not a show-stopper for LTS.

touhidurabir:
[...]

PR #13054, opened 2026-07-16
pkp/pkp-lib#12702 update apply of SubmissionFileStageAccessPolicy

touhidurabir:
for pkp/pkp-lib#12702 .

### DIFF
--- a/api/v1/submissions/MediaFilesController.php
+++ b/api/v1/submissions/MediaFilesController.php
@@ -37,10 +37,8 @@ use PKP\file\FileManager;
 use PKP\file\TemporaryFileManager;
 use PKP\security\authorization\ContextAccessPolicy;
 use PKP\security\authorization\internal\SubmissionFileMatchesSubmissionPolicy;
-use PKP\security\authorization\internal\SubmissionFileStageAccessPolicy;
 use PKP\security\authorization\PublicationAccessPolicy;
 use PKP\security\authorization\PublicationWritePolicy;
-use PKP\security\authorization\SubmissionFileAccessPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
 use PKP\submission\Genre;
@@ -76,6 +74,7 @@ class MediaFilesController extends PKPBaseController
      */
     public function getGroupRoutes(): void
     {
+        // Read access: authors may view media files (read-only)
         Route::middleware([
             self::roleAuthorizer([
                 Role::ROLE_ID_SITE_ADMIN,
@@ -87,7 +86,17 @@ class MediaFilesController extends PKPBaseController
         ])->group(function () {
             Route::get('', $this->getMany(...))
                 ->name('mediaFiles.getMany');
+        })->whereNumber(['submissionId', 'publicationId']);
 
+        // Write access: media is a production artifact editable by editorial roles only (no author)
+        Route::middleware([
+            self::roleAuthorizer([
+                Role::ROLE_ID_SITE_ADMIN,
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SUB_EDITOR,
+                Role::ROLE_ID_ASSISTANT,
+            ]),
+        ])->group(function () {
             Route::post('', $this->add(...))
                 ->name('mediaFiles.add');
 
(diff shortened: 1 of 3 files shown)


touhidurabir:
@ewhanson can you take a look at the PR , specially the `MediaFilesController` changes. I have removed the policy `SubmissionFileStageAccessPolicy` that can cause same issue as https://github.com/pkp/pkp-lib/issues/12702, but removing another internal policy `SubmissionFileMatchesSubmissionPolicy` seems very challenging due to how it's been used and wired to form requests. 

Note that usage of `SubmissionFileMatchesSubmissionPolicy` is not causing any issue or bug but as per the https://github.com/pkp/pkp-lib/issues/13060, we want to minimise direct use of internal policy in controller/handler `authorize` method .

api/v1/submissions/MediaFilesController.php
```diff
@@ -87,7 +86,17 @@ public function getGroupRoutes(): void
         ])->group(function () {
             Route::get('', $this->getMany(...))
                 ->name('mediaFiles.getMany');
+        })->whereNumber(['submissionId', 'publicationId']);
 
+        // Write access: media is a production artifact editable by editorial roles only (no author)
```
ewhanson, line 91: Are we sure about this? @blesildaramirez, are authors able to upload media files directly, or does it have to be handed off to an editor for upload?

api/v1/submissions/MediaFilesController.php
```diff
@@ -129,31 +138,10 @@ public function authorize(PKPRequest $request, array &$args, array $roleAssignme
...
         if (in_array($actionName, ['edit', 'delete', 'link'])) {
-            // Load the submission file to get its fileStage
             $submissionFileId = (int) static::getRequestedRoute()->parameter('submissionFileId');
-            $submissionFile = $submissionFileId ? Repo::submissionFile()->get($submissionFileId) : null;
-            $fileStage = $submissionFile ? $submissionFile->getData('fileStage') : 0;
 
             // Ensure the file belongs to the submission
             $this->addPolicy(new SubmissionFileMatchesSubmissionPolicy($request, $submissionFileId));
-
-            $this->addPolicy(
```
ewhanson, line 140: In removing these internal policies, are we not accounting for these checks anymore?

api/v1/submissions/MediaFilesController.php
```diff
@@ -129,31 +138,10 @@ public function authorize(PKPRequest $request, array &$args, array $roleAssignme
         }
 
         if (in_array($actionName, ['edit', 'delete', 'link'])) {
-            // Load the submission file to get its fileStage
             $submissionFileId = (int) static::getRequestedRoute()->parameter('submissionFileId');
-            $submissionFile = $submissionFileId ? Repo::submissionFile()->get($submissionFileId) : null;
-            $fileStage = $submissionFile ? $submissionFile->getData('fileStage') : 0;
 
             // Ensure the file belongs to the submission
             $this->addPolicy(new SubmissionFileMatchesSubmissionPolicy($request, $submissionFileId));
```
ewhanson, line 144: In working at removing the internal policies, have you explored only using `SubmissionFileAccessPolicy` directly and why or why not that does/doesn't work?

ewhanson (changes requested):
Thanks @touhidurabir. A few questions below.

ewhanson:
@touhidurabir, also to your earlier point:

> ]...] removing another internal policy SubmissionFileMatchesSubmissionPolicy seems very challenging due to how it's been used and wired to form requests.

I understand what the purpose is behind removing this being tricky (it's hard to ensure this without the check), but could you elaborate on what you mean about how it's wired to form requests making it challenging? Thanks!

api/v1/submissions/MediaFilesController.php
touhidurabir: The author dashboard does not seems to have any option to add//link etc options . also the `useMediaFileManagerConfig.js` seems to back this by specifying it in permission list as `ROLE_ID_AUTHOR --> MEDIA_FILE_LIST` only . 

@blesildaramirez perhaps you can confirm the expected behaviour

api/v1/submissions/MediaFilesController.php
touhidurabir: the use of `SubmissionFileStageAccessPolicy` seems to (or seems to me at least) trying to cover followings : 
 - Author exclusion --> now moved to route level 
 - check of file belongs to submission --> seems more like task of `SubmissionFileMatchesSubmissionPolicy`
 - Production stage scoping of editorial users --> this actually the bug that mentioned in the issue . Also does not seems that media file actions at editorial level is and should be only restricted to certain stage at editorial level

api/v1/submissions/MediaFilesController.php
touhidurabir: > have you explored only using SubmissionFileAccessPolicy directly and why or why not that does/doesn't work?

I did and it will cause more issue as it re-introduces the regression. and the `SubmissionFileAccessPolicy` is not an ownership policy but like as stage scoped/level composite checking . Here every role branch is checked on `$stageId = (int)$request->getUserVar('stageId')` and composes `SubmissionFileMatchesWorkflowStageIdPolicy + WorkflowStageAccessPolicy + AssignedStageRoleHandlerOperationPolicy` . 

Most notebale is the Media routes send no `stageId` so `0` . From PKP\submissionFile\Repository::getWorkflowStageId() , we get media files map to `PRODUCTION --> 5` . this DENY every editorial branch.

touhidurabir:
>  elaborate on what you mean about how it's wired to form requests making it challenging?

@ewhanson 
The URL file `{submissionFileId}` is obtained via `MediaFileValidationTrait::getSubmissionFile()`, which is a pure read of `ASSOC_TYPE_SUBMISSION_FILE`  which read as authorized context key fed by the `SubmissionFileMatchesSubmissionPolicy` . The the `EditMediaFile/LinkMediaFile` use that loaded one without any more validation (rely on the policy validation). Even the route `delete` which has no form request, reads `ASSOC_TYPE_SUBMISSION_FILE` straight . 

It is not impossible but the change will span to many file just to remove one single policy --> that is what I mean as challenging as does not seems worth it .

api/v1/submissions/MediaFilesController.php
blesildaramirez: Hi @ewhanson @touhidurabir , sorry for missing this question. The media files should follow the same policy as in Galleys, so yes the authors should just be able to view media files and download, with no access to uploading, editing or deleting a media file.

asmecher (approved):

asmecher:
Merging to `main`. We do have to resolve the external use of internal policies (https://github.com/pkp/pkp-lib/issues/13060), but I'd like to get this piece wrapped up and merged in the meantime. @ewhanson, let me know if more needs capturing on that issue regarding outstanding work raised here.

Issue #12994, opened 2026-07-06
[Reviews] Competing interest - Explicitly record whether was declared for reviews


jardakotesovec:
**Description:**
If following setting has content - Settings -> Workflow -> Review -> Reviewer Guidance -> Competing Interests. Than reviewer is presented with question whether he has competing interests or not (image below)

<img width="1019" height="207" alt="Image" src="https://github.com/user-attachments/assets/769ee8d4-0323-42ef-a229-e9fc54eb69f7" />

Issue is that whether in database ojs saves NULL both when he is not asked for competing interests and when he declares 'no competing interests'. Thats problematic if Journal introduces competing interests later in the process as its impossible to tell which reviews declared no competing interests and which were not asked at all and its important to differentiate to have accurate record

**Proposal**
- Add `competingInterestsDeclared` boolean to the review assignment. To capture whether he actually declared competing interest or not.

**Steps to test the changes:**
1. Check whether data are accurately saved in the database - competingInterestDeclared should become true only if reviewer was asked for competing interests (its enabled for journal in workflow -> Review -> Reviewer Guidance (some text needs to be entered to Competing interest field there)
2. Test with oreTheme whether it its displayed accurately for Readers (open peer review). When its declared it should either show actual competing interest text at the end of the review or "No competing interests were disclosed."  And if its not declared (competing interest is not enabled for given journal) - it does not show up at all. 
3. Same behaviour is on the "Review details" screen. Where before it displayed competing interests only if something was disclosed. Now it also shows "No competing interests were disclosed." if it was declared not having any CI.

**PRs:**
ojs: https://github.com/pkp/ojs/pull/5688
pkp-lib: https://github.com/pkp/pkp-lib/pull/13073
ui-library: https://github.com/pkp/ui-library/pull/935

**Migration needed**
php lib/pkp/tools/migration.php 'PKP\migration\upgrade\v3_6_0\I12994_CompetingInterestsDeclared' up

**What application are you using?**
  OJS, OMP or OPS version 3.6.0

jardakotesovec:
@kaitlinnewson Hi Kaitlin, can I ask for review?


[...]

PR #13073, opened 2026-07-21
pkp/pkp-lib#12994 competing interest declared for reviewers

jardakotesovec:


api/v1/peerReviews/resources/PublicationPeerReviewResource.php
```diff
@@ -236,6 +236,8 @@ private function getReviewAssignmentPeerReviews(Enumerable $assignments, Context
                 'reviewerRecommendationTypeLabel' => $recommendation ? $recommendationTypesTypeLabels[$recommendation->type] : null,
                 'reviewForm' => $reviewForm,
                 'reviewerComments' => $reviewerComments,
+                'competingInterestsDeclared' => $assignment->getCompetingInterestsDeclared(),
+                'competingInterests' => $assignment->getCompetingInterests(),
```
kaitlinnewson: Should this be guarded with $isReviewOpen as well?

classes/submission/reviewer/form/PKPReviewerReviewStep1Form.php
```diff
@@ -128,11 +128,16 @@ public function execute(...$functionParams)
...
 
-        // Set competing interests.
-        if ($this->getData('competingInterestOption') == 'hasCompetingInterests') {
-            $reviewAssignment->setCompetingInterests($this->request->getUserVar('reviewerCompetingInterests'));
-        } else {
-            $reviewAssignment->setCompetingInterests(null);
+        // Set competing interests. The question is only presented to the reviewer
+        // when the context has competing interests guidelines (see step1.tpl), so
+        // only then can a declaration be recorded.
+        if ($this->request->getContext()->getData('competingInterests')) {
```
kaitlinnewson, line 134: Maybe slightly out of scope, but I wonder if the reviewer competing interests should be explicitly turned on/off (like authors), rather than assuming based on any text in that field? e.g. what if they entered something like "reviewers don't need to declare competing interests", or the journal just wants to enable it.

api/v1/peerReviews/resources/PublicationPeerReviewResource.php
jardakotesovec: I was back and forward on this one.. . But you are probably right. I think anonymous public review will be rare, but if someone journal decides to use it - they probably had good reasons to keep it anonymous and therefore they would prefer to also have competing interest not exposed to not risk exposing something revealing about identity. 

We also have not really considered presentation for the anonymous review on frontend - the names are quite essential in the UI..  so will need to be eventually replaced with something like "Reviewer 1" and "Reviewer 2"..

For competing interest I won't show it on the public frontend at all for now..

classes/submission/reviewer/form/PKPReviewerReviewStep1Form.php
jardakotesovec: I was also confused that we control whether competing interest are enabled by the fact that some text is there. I suspect there is more fields working like this. Since its "one time setting" kind of thing - its not that big of the deal as other areas which are used daily.  I agree not great, but out of the scope.

PR #13075, opened 2026-07-21
I206

jardakotesovec:


api/v1/peerReviews/resources/SubmissionPeerReviewResource.php
```diff
@@ -66,120 +62,89 @@ class PublicationPeerReviewResource extends JsonResource
...
+            ->keyBy(fn (Publication $publication) => $publication->getId());
 
         /** @var ReviewRoundDAO $reviewRoundDao */
         $reviewRoundDao = DAORegistry::getDAO('ReviewRoundDAO');
-        $reviewRounds = $reviewRoundDao->getByPublicationIds($allAssociatedPublicationIds);
+        // Only rounds whose reviewed publication version is published are part of the
+        // public record; rounds of an unpublished (in-review) version stay hidden.
+        $reviewRounds = collect($reviewRoundDao->getBySubmissionId($submission->getId())->toArray())
+            ->filter(fn (ReviewRound $reviewRound) => $reviewRound->getPublicationId() !== null
+                && $publishedPublications->has((int) $reviewRound->getPublicationId()));
```
taslangraham, line 82: Just checking to confirm if you've already looked into whether we may also need to sort these review rounds by ID/publication ID to ensure the correct chronological order is maintained?

taslangraham (approved):
@jardakotesovec I just had one minor question. Approving, as everything looks good in general.

api/v1/peerReviews/resources/SubmissionPeerReviewResource.php
jardakotesovec: @taslangraham I think so, the ordering is already implicitely in the getBySubmissionId:

```
    public function getBySubmissionId($submissionId, $stageId = null, $round = null)
    {
        $q = DB::table('review_rounds', 'rr')
            ->where('rr.submission_id', '=', $submissionId)
            ->when($stageId, fn (Builder $q) => $q->where('rr.stage_id', '=', $stageId))
            ->when($round, fn (Builder $q) => $q->where('rr.round', '=', $round))
            ->orderBy('rr.stage_id')
            ->orderBy('rr.round');

    }

```

Issue #13059, opened 2026-05-21
[Integrity issue] | No transparency of actions made while proxying as someone else


1StuJB:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description. Example of a title: [Peer Review] | Reader View - Article landing page, Missing dynamic values

### Description

If you take an action, e.g. complete a review assignment, while logged into someone elses account via the account proxy mechanism, when actions are taken while proxying (e.g. a review report completion), in the activity log the proxy status is not recorded. Instead, the action taken by the account that has been proxied is logged. This is a serious integrity issue, as it means rogue editors could take action, masking things as other people without any log being recorded. Please fix this to show in cases like this, someone else was proxying as the user logged as taking the action.

### Steps to Reproduce

1. Login with User/Role 'Editor'
2. Navigate to a manuscript that needs a review report completed
3. Scroll to Reviewers, choose one and go to the menu item "Login as"
4. Complete the review while proxied as the reviewer
5. Select submit button and log back out to the initial account.

### Expected Result

If someone undertakes an action while logged in as someone else, this information should be logged in the system activity log.

### Actual Result

The action in the activity log is logged against the name of the person being proxied, with no mention of the person proxying the account who is actually taking the action.

### Additional Information

Before storing impersonation record and presenting in even log activity
<img width="3058" height="1128" alt="Image" src="https://github.com/user-attachments/assets/a2d3a4c7-4a33-4374-be05-a0003668698a" />

After storing impersonation record and presenting in even log activity
<img width="3050" height="1138" alt="Image" src="https://github.com/user-attachments/assets/05df13f0-a102-4e1d-872f-def6145cf727" />

#### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/13081
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5694
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2417
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1356

mreiko:
Hi @1StuJB This ticket is dependent on another in-progress ticket in the system so I am leaving it with Alec to add further details when available.

asmecher:
(We agreed in a 1:1 that it would be good to add a new column to the event log table for the "logged in as" user ID to be tracked.)

touhidurabir:
[...]

PR #13081, opened 2026-07-23
pkp/pkp-lib#13059 record impersonated user in event log

touhidurabir:
for pkp/pkp-lib#13059 .


api/v1/submissions/tasks/resources/TaskResource.php
```diff
@@ -119,12 +120,24 @@ public function toArray(Request $request)
                 );
             }
 
+            // Attached impersonated details to the user label if exists
+            $userLabel = $activity->getLocalizedData('userFullName');
+            if ($impersonatedAsUserId = $activity->getImpersonatedAsUserId()) {
+                $impersonatedAsName = Repo::user()->get((int) $impersonatedAsUserId, true)?->getFullName() ?? '';
+                if ($impersonatedAsName !== '' && $impersonatedAsName !== $userLabel) {
```
asmecher: Stylistic, but there's needlessly defensive code here that adds clutter...
- The `(int)` cast is unnecessary because we already have an `if` above, and the `getImpersonatedAsUserId` return is `?int`
- The `?->` is unnecessary because we have a foreign key constrant, and
- `?? ''` will be checking for a `null`, which is then tested with the exact test just below.

I also don't think we should be testing the name in order to determine whether to show impersonation or not. Someone could change their account's name to match someone else's. Here, that would result in the impersonation being hidden.

I think this is more readable:
```suggestion
                $impersonatedAsName = Repo::user()->get($impersonatedAsUserId, true)->getFullName();
                if ($impersonatedAsName != '') {
```

api/v1/submissions/tasks/EditorialTaskController.php
```diff
@@ -202,22 +203,25 @@ public function addTask(AddTask $illuminateRequest): JsonResponse
         $submission = $this->getAuthorizedContextObject(PKPApplication::ASSOC_TYPE_SUBMISSION); /** @var Submission $submission */
         $editorialTask->refresh();
 
-        // Log the action
         $currentUser = $this->getRequest()->getUser();
-        $currentUserUserGroups = $this->getCurrentUserUserGroups($submission, $currentUser, $editorialTask);
+
+        // During "Login as", attribute the log to the real actor, not the impersonated session user.
+        $actionedUser = Validation::loggedInAs() ? Repo::user()->get(Validation::loggedInAs()) : $currentUser;
```
asmecher, line 209: `$allowDisabled`? Here and elsewhere.

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
```diff
@@ -733,6 +732,7 @@ public function unconsiderReview($args, $request)
             'assocId' => $submission->getId(),
             'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_UNCONSIDERED,
             'userId' => Validation::loggedInAs() ?? $user->getId(),
+            'impersonatedAsUserId' => Validation::loggedInAs() ? $user->getId() : null,
```
asmecher: I suggest using the existing "logged in as" terminology rather than "impersonated as" which is confusing. (Just "impersonated" would make more sense, but I don't think it's necessary to introduce an additional term.)

asmecher (changes requested):
A few comments -- the "impersonated" name change is going to be annoying, I'm afraid!

classes/controllers/grid/users/reviewer/PKPReviewerGridHandler.php
asmecher: (We discussed this during the 1:1 and decided to use "impersonatedUserId" rather than "impersonatedAsUserId".)

api/v1/submissions/tasks/EditorialTaskController.php
touhidurabir: dont think it's need here as login is based on `Validation::loggedInAs() ? Repo::user()->get(Validation::loggedInAs())` . So it get the impersonating user if someone is impersonating and a disable can not perform impersonation action without login which is gated by disable flag .

locale/en/submission.po
```diff
@@ -2220,6 +2220,10 @@ msgstr "Task reassigned from {$taskOwnerOldUsername} to {$taskOwnerNewUsername}
 msgid "submission.event.task.assigned"
 msgstr "Task assigned to {$taskOwnerNewUsername} by {$username} on {$dateLogged}"
 
+#, fuzzy
```
asmecher: English-language texts shouldn't be marked as `fuzzy`; it'll prevent Weblate from showing the text for updated translations in other languages.

asmecher (changes requested):
Thanks, @touhidurabir -- just one comment and it's ready for rebase and merge. Please warn the development channel in MM about the new migration when you merge it!

Issue #13084, opened 2026-07-23
[Themes] | 404 Errors - Not possible to theme the 404 error page


NateWr:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

It is not possible for themes to style the 404 error page. When a visitor reaches one of these pages, they get an unstyled 404 error page that looks similar to a 500 Internal Server error.

I will provide PRs with suggested solutions for 3.3 and 3.5. 

For the `main` branch, I recommend that we introduce a new `404.tpl` template file that themes can customize if they want. The default template in `pkp-lib` can be a simple "non-branded" design that still looks intentional. It can provide a link to the main page, a back-link, and a link to the dashboard for logged-in users.

This work would be contributed as part of contract work for TIB.

### Steps to Reproduce

example: https://ojs33.demo.publicknowledgeproject.org/index.php/demojournal/article/view/95

### Expected Result

The error message should not look like the website broke.

### Actual Result

example: https://ojs33.demo.publicknowledgeproject.org/index.php/demojournal/article/view/95

### Application Version

OJS 3.3 and 3.5

NateWr:
PRs for 3.3 and 3.5:

- https://github.com/pkp/pkp-lib/pull/13085 (`stable-3_3_0`)
- https://github.com/pkp/pkp-lib/pull/13087 (`stable-3_5_0`)

To avoid introducing new locale keys in a stable version, I have re-used the only existing message appropriate for a 404 page. I think this is still an improvement on before and theme's can override it however they want.

For the `main` branch, I propose introducing a new template, `404.tpl`, and passing some additional information to the template, including:

- Locale keys with better 404 not found messages.
- A link to the main page (journal or site level).
- If available, the value of `PKPHandler::_isBackendPage`, along with whether or not the user is logged in and if they have a dashboard-accessing role (and a link to the dashboard).
- The requested page and op if these have been established, plus the URL requested.

If you approve, I can make a PR for the `main` branch. This work will be part of a short contract with TIB, so I will need to know in the next week or so.

mtub:
We are really interested in this to be fixed, so any update will be appreciated...

PR #13087, opened 2026-07-23
pkp/pkp-lib#13084: Use default error template to show 404 error messages

NateWr:


pages/management/ManagementHandler.php
```diff
@@ -600,7 +600,7 @@ public function editUser(array $args, $request): void
             $invitationHandler = $invitation->getInvitationUIActionRedirectController();
             $invitationHandler->createHandle($request,$userId);
         } else {
-            $request->getDispatcher()->handle404();
+            throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException();
```
NateWr, line 603: Please double-check this change. I believe it is correct.

Issue #12152, opened 2025-12-17
Consider new contributor types and roles for metadata distribution


bozana:
After the main work on different contributor types and machine readable roles is merged, this new feature needs to be considered also for metadata distribution, and in all plugins.
- [x] CitationStyleLanguage
- [ ] Crossref
- [ ] DataCite
- [ ] ...

PRs:
citationStyleLanguage
- https://github.com/pkp/citationStyleLanguage/pull/159
- https://github.com/pkp/ojs/pull/5320 (submodule update)
- https://github.com/pkp/omp/pull/2211 (submodule update)
- https://github.com/pkp/ops/pull/1182 (submodule update)

bozana:
Hi @touhidurabir, could you please review this?
Thanks a lot!

bozana:
Hi @kaitlinnewson, maybe to also ask you to review it -- Either you or Touhidur, who ever comes first to it...
Thanks a lot!

kaitlinnewson:
Hi @bozana, just a minor comment, and all approved. I think the wrong PR is linked above for the OJS submodule update so I skipped that one.

bozana:
Hi @Tribunal33, I have considered new contributor types and roles in the citations style language plugin. If you would like to test it... -- basically to see if the roles are correctly considered in the different citations outputs.
Thanks a lot!

Tribunal33:
@bozana I'm not sure what to expcet.  Could you demo this?  I tried to figure out what would be displayed but there isn't any acceptance criteria in the ticket.

So far, I've installed the plugin and added some roles to a contributor but not seeing them displayed on the article landing page where the Citation Language Pluging is being displayed.  I downloaded the 2 files as well but didn't see a role there either.  What am I missing?  



<img width="515" height="384" alt="Image" src="https://github.com/user-attachments/assets/438115ee-6e2a-443b-8e05-7c9d23f8407b" />

<img width="1092" height="393" alt="Image" src="https://github.com/user-attachments/assets/c1b33c3a-8ba3-43f0-87f0-eda925362fa1" />

Tribunal33:
I didn't see any issues when playing around with the citations or contributor roles.  This passes QA but not clear what it should be doing differently with these changes or what the existing Acceptance Criteria was for this plugin.

Here are some steps I used to test it.
1.  Install the plugin
2. Add a new contributor role under Workflow > Contributor Roles 
3. Add new contributors with anonymous or organization group type
[...]

PR #13097, opened 2026-07-27
pkp/pkp-lib#12152 Consider new contributor types and roles for metadata distribution

ajnyga:
https://github.com/pkp/pkp-lib/issues/12152


bozana:
we decided not to rely on the order here for priority, so I will close this PR

Issue #13003, opened 2026-07-07
Reader interface performance testing


asmecher:
The OJS reader interface should perform according to these requirements:

> The Platform will be developed so as to be available in full functionality at 99.9% of the time and
highly responsive to increased levels of traffic, supporting at least 3000 concurrent users with a
response time of no more than two seconds.

## Batch Entity Hydration
(See Jarda's experimental PR: https://github.com/pkp/pkp-lib/pull/13099; also refinements https://github.com/pkp/pkp-lib/pull/13203 and https://github.com/pkp/ojs/pull/5751)

Numbers taken from `asmecher/i13003-batch-load` vs. `main`, both circa 2026-08-19

### Query counts (using ORE database):
- Article landing page: `https://doi.org/10.12688/openreseurope.17205.3` (`/ore/article/view/15578`), cache warmed:
  - `main` branch: 627 queries; `asmecher/i13003-batch-load`: 247 queries (60% improvement)
- Category browse page for Engineering & Technology, 10 of 243 items (`/ore/search?frascatiBases[]=Engineering+and+technology`), cache warmed:
  - `main` branch: 1586 queries; `asmecher/i13003-batch-load`: 537 queries (66% improvement)

Known issues not addressed here (out of scope for these PRs):
- When publications are fetched from a submission, *all* publications are loaded (including sub-entities). The vast majority of interactions with a submission only require the "current" publication. I'm not sure adding more modes of loading is the best approach, when we could just reduce the number and impact of loads and keep complexity lower that way.
- When fetching submissions using `getMany`, they are still loaded one at a time. I'm not sure whether this is worth changing, or whether it'll hit a memory limit due to the number of hydrated entities.
- The `Funders` class contains static caches/dependent fetches that are not yet optimized. This results in N+1 problems.
- There is some preexisting non-standard loading on the ~~`citations`~~ / `citationsRaw` entities.
- ~~Sub-entities are hydrated when the first containing entity is hydrated. For example, when a publication is instantiated, all attached database rows for authors are fetched, regardless of whether they are ever used. Consider a fetch-on-use strategy.~~
- There are other helper classes outside the `DAO`/`Collector` pattern that cause heavy DB load, particularly around peer review data.

Pending PRs:
- Fix serialization of closures in jobs involving author objects:

Merged PRs:
- Main implementation
  - https://github.com/pkp/pkp-lib/pull/13179 (implementation notes are here)
  - https://github.com/pkp/ojs/pull/5745
[...]

PR #13099, opened 2026-07-27
Batch entity hydration to cut article page queries ~14x

jardakotesovec:
Batch-load settings and related entities in EntityDAO/DAO getMany() instead of querying per entity:

- EntityDAO: prefetchSettings() loads a batch's settings rows in one query; fromRow() falls back to a per-entity query for rows outside the prefetched batch.
- Publication DAO: prefetchRelated() batches submission locales, categories, data citations, funders, controlled vocabs and DOIs, and shares one deferred authors loader across the batch.
- Author DAO: batches affiliations and credit/contributor roles.
- ReviewAssignment DAO: batches settings, reviewers and DOIs.
- User/Citation DAOs: batch settings via prefetchSettings().
- Publication citations are now hydrated lazily (LazyCollection, mirroring authors/citationsRaw); copyCitations() accepts any iterable and the citations API no longer iterates by reference.

All prefetch maps are pre-filled for the batch ids and consumers fall back to the original per-entity queries on a missing key, so the prefetch can only reduce queries, never change hydrated data.

Article landing page: 5,356 -> 385 queries on a 728-citation, 4-version article; rendered HTML is byte-identical.

### DIFF
--- a/api/v1/citations/PKPCitationController.php
+++ b/api/v1/citations/PKPCitationController.php
@@ -347,13 +347,12 @@ class PKPCitationController extends PKPBaseController
 
         $citations = $publication->getData('citations');
         $citationsMapped = [];
-        foreach ($citations as &$citation) {
+        foreach ($citations as $citation) {
             $citation->setProcessingStatus(CitationProcessingStatus::NOT_PROCESSED->value);
             Repo::citation()->edit($citation, []);
             Repo::citation()->reprocessCitation($citation);
             $citationsMapped[] = Repo::citation()->getSchemaMap()->map($citation);
         }
-        unset($citation);
 
         return response()->json([
             'itemsMax' => count($citationsMapped),
--- a/classes/affiliation/DAO.php
+++ b/classes/affiliation/DAO.php
@@ -112,6 +112,30 @@ class DAO extends EntityDAO
         });
     }
 
+    /**
+     * Fetch the affiliations for many authors with two queries,
+     * returned as [authorId => array<Affiliation>].
+     */
+    public function getByAuthorIds(array $authorIds): array
+    {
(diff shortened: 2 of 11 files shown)


asmecher:
@jardakotesovec, this looks good!

We're "riding two horses at once" with both `EntityDAO` (submissions, publications, and many more) and Eloquent (user groups, user comments, notifications, notes, etc) storage models. On balance, I don't think the effort of maintaining two storage layers is justified; moving 100% to Eloquent will allow us to discard a lot of legacy (`DataObject`, `EntityDAO`, and a lot of related cruft) and align ourselves closer with Laravel's developer experience. So at the moment I'm wary of approaches that are going to add tech debt to EntityDAO.

This doesn't add much, but it also falls short of Eloquent's eager loading capabilities, which include [nested eager loading](https://laravel.com/docs/13.x/eloquent-relationships#nested-eager-loading) -- this would allow two-query hydration of authors when fetching publications, but not when fetching a submission, as that requires fetching of publications first.

I think this leaves us with 3 options...
1. Add this proposal; accept that nested eager loading will not be supported, and accept a slight increase in EntityDAO tech debt;
2. Do not add this proposal, and pursue other caching approaches instead;
3. Attempt to add nested eager loading in addition to this proposal, accepting significant technical debt.

I would explore either option 1 or 2. The reason I was exploring serialized entities in [my object caching PRs](https://github.com/pkp/pkp-lib/issues/13003) is because they offer the same performance improvement, and will happily work on both Eloquent and DataObject-based implementations.

Please let me know what you think!

asmecher:
(See the "Batch Entity Hydration" section at https://github.com/pkp/pkp-lib/issues/13003 for next steps)

PR #13104, opened 2026-07-28
pkp/pkp-lib#13074 Stamp submission last_modified on Author/Citation c…

### LINKED ISSUE #13074: Sub-entity saves (Author, Citation, DataCitation, Funder) don't update parent Publication/Submission `last_modified`
## Description
Editing, adding, deleting, or reordering `Author`, `Citation`, `DataCitation`, or `Funder` records never calls `stampModified()` (or `Repo::publication()->edit()` / `Repo::submission()->edit()`) on the parent `Publication`/`Submission`. Each repository (`author\Repository`, `citation\Repository`, `dataCitation\Repository`, `funder\Repository`) persists directly via its own DAO/model with no cascade.

**Impact:** `publications.last_modified` / `submissions.last_modified` don't reflect these edits, so anything relying on that timestamp as "content is fresh as of X" (OAI-PMH selective harvesting, search reindexing, feeds, caching) misses these changes entirely.

**Suggested fix:** rather than patching each repository individually (easy to miss for future sub-entities), add a generic mechanism — e.g. a shared trait/interface  implemented by these models, with one listener on their `saved`/`deleted` events that stamps the correct parent (`Publication` for Author/Citation/DataCitation, `Submission` for Funder). For bulk paths (e.g. `Citation::importCitations()`, `saveContributorsOrder()`), stamp once per operation, not once per row, to avoid redundant writes.

## PRs
- https://github.com/pkp/pkp-lib/pull/13104
- https://github.com/pkp/omp/pull/2420

monmarzia-archi:
**Draft PR — work in progress, opened for early feedback as requested by @asmecher**

Part of #13074 (Track 1 only: Author and Citation entities, targeting stable-3_5_0).
Track 2 (Funder, DataCitation, 3.6.0-only) remains open.Fixes #13074 (only for Author and Citation entities, only on v.3.5)

## Summary

Editing an Author or a Citation on a Submission's Publication does not currently update
`submissions.last_modified` on the parent Submission, affecting OAI-PMH harvesting and any
other consumer relying on that field.

This PR introduces a `StampSubmissionModified` listener on the existing `MetadataChanged`
event, and adds dispatch points in `Author\Repository` (add, edit, delete, setAuthorsOrder)
and `CitationDAO` (insert, update, delete/deleteById, deleteByPublicationId, importCitations).

## Testing done so far

- Manual testing on OJS: add/edit/delete/reorder Author, add/edit/delete Citation on a
  published Submission, confirmed `last_modified` on the parent Submission updates as
  expected.
- Same manual testing repeated on OMP (shared library).
- Confirmed existing `MetadataChangedListener` (search reindexing) is unaffected, since
  both listeners hook the same event independently.
- Not yet verified: actual propagation to an OAI-PMH harvest response. This PR changes
  `last_modified`, which the OAI-PMH datestamp query already reads from; harvest-level
  verification would be useful from reviewers with an OAI-PMH client set up.

## Still to do before this leaves draft status

- [X] Run existing pkp-lib automated test suite (Author, Citation, Submission, observers/listeners)
- [ ] Write unit/functional tests for the new listener and for each dispatch point
- [X] Confirm `MetadataChangedJobTest` / `BatchMetadataChangedJobTest` remain green

## Open questions for review

1. Is it possible, in the current OJS/OMP editorial workflow, to edit an Author or
   Citation on a non-current Publication version? If so, the approach of stamping
   the Submission directly may need to be revisited.

## AI-assisted development disclosure

Portions of this PR's code analysis and drafting were assisted by an AI tool (Claude).
All technical claims and code have been independently reviewed and verified by me
prior to submission, per PKP's AI contribution policy.


classes/author/Repository.php
```diff
@@ -320,5 +327,34 @@ public function setAuthorsOrder(int $publicationId, array $authors)
...
+
+    /**
+     * Dispatch MetadataChanged for the submission owning the given publication.
+     * Silently no-ops if the publication or submission can't be resolved.
+     *
+     * @see https://github.com/pkp/pkp-lib/issues/13074
+     */
+    private function dispatchMetadataChanged(int $publicationId): void
+    {
+        $publication = Repo::publication()->get($publicationId);
```
asmecher, line 349: @monmarzia-archi, there are a couple of downsides to rehydrating the publication from its ID...
1. It incurs database activity when that currently isn't necessary for consumers of the event, and
2. Generally speaking we risk having several versions of the object in memory and risk having interleaved updates of each clobbering fields.

The first isn't really a problem, since insert/update/delete operations are relatively rare. We'll probably also need some kind of rehydration in order to cascade the update events, i.e. so that an update to an author can cascade to the publication and from there to the submission.

Since the only thing we're doing is updating the "last modified" date, I'd rather we did that as an atomic database operation without re-persisting the whole object. If we do run into a problem with interleaved updates, it makes it clear what part of the code needs to be resolved.

I'm open to counter-proposals on this, but I think it's probably best to change the `MetadataChanged` event so that it's applicable to changes on any supported entity -- not just submissions but also authors, publications, etc.

So in summary, I think loading the "parent" entity from DB is OK, but
- it'll be necessary to then also emit a new event on that, and
- it's probably better to add an atomic update function on the last modified date in the DAO class.

classes/author/Repository.php
```diff
@@ -320,5 +327,34 @@ public function setAuthorsOrder(int $publicationId, array $authors)
...
+     * methods (add, edit, delete, setAuthorsOrder). No existing line is removed;
+     * each block below shows the method with the new line(s) marked.
+     *
+     */
+
+    /**
+     * Dispatch MetadataChanged for the submission owning the given publication.
+     * Silently no-ops if the publication or submission can't be resolved.
+     *
+     * @see https://github.com/pkp/pkp-lib/issues/13074
```
asmecher, line 345: Generally we don't refer to issue numbers from the codebase unless there's an obscure line of code that needs explanation; we can find this from `git blame` if needed. (Here and elsewhere.) Ideally the code is self-explanatory :)

asmecher (commented):
Just some early comments -- thanks, @monmarzia-archi!

classes/author/Repository.php
monmarzia-archi: Got it! Besides the descriptive comments, I've left a few comment lines referencing #13074 to find them quickly — let me know if I should remove them too.

classes/author/Repository.php
monmarzia-archi: Thanks, this gives me a much clearer picture, especially the point about the atomic update. That part is already sketched out on my end (direct DAO-level update, no full entity re-save), pending final testing before I commit it.

On generalizing the event, since you mentioned being open to counter-proposals: I'd suggest introducing new concrete events per entity instead (AuthorMetadataChanged, PublicationMetadataChanged), each with its own listener that resolves the parent and dispatches the next event in the chain, leaving MetadataChanged and its existing listener untouched. One consideration that occurred to me, more as a thought to share than a firm position: since MetadataChanged already has a listener in production, and possibly third-party plugins depending on it too, leaving it untouched means nothing already relying on it today is affected by this change. I'm not sure how much weight that should carry against the benefit of a single generic event, you'd have a much better sense of that than I would, but it seemed worth mentioning given the project's general caution around breaking changes.

My scope so far covers Author, Citation, and Publication, the entities already in 3.5, leaving Funder/DataCitation for later since they're 3.6.0-only. I'd like to know whether this concrete, additive approach covers what you had in mind, or whether there's a reason you'd want MetadataChanged itself made generic beyond this cascade. Either way, is this blocking for this PR, or can we land the current scope (atomic update plus this event chain) and treat a fully generic architecture as a separate follow-up? Part of why I'm asking: I think this fix is foundational to a caching proposal we've been discussing on the forum, whose tag-based invalidation (Section 3.3) depends on entities reliably reporting when their content changed. Getting this right at a contained scope now matters more to me than expanding the redesign, since that's what the caching work will build on.

monmarzia-archi:
@asmecher Back with the reworked cascade. Thanks for the feedback on the call, it made the design a lot clearer.

How the cascade works now

There are three new events, one per entity: AuthorMetadataChanged, CitationMetadataChanged and PublicationMetadataChanged. Each one has a small listener. All three listeners queue the same shared job.

The cascade moves one step at a time. An author change emits an author event. The job handles it, finds the publication, and emits a publication event. A second job handles that one and stamps the submission.

I chose one step at a time so that each hop is its own unit of work, with its own retry semantics if the connection ever changes. It also leaves a public event at the publication level, in case anything else needs to react to publication changes later.

A change from the version you saw

The earlier version emitted MetadataChanged from the author dispatch points. That event already has a listener, MetadataChangedListener, which reindexes the search index. So editing an author also triggered a reindex. Nobody asked for that, and it went against the "purely additive" argument I gave for concrete events.

The new events fix this. MetadataChanged is no longer emitted from any new point. Editing an author no longer reindexes. Every author test asserts this.

Targeted update

You asked for an update that only touches last_modified. SubmissionDAO::updateLastModified() does that. It writes one column, instead of going through EntityDAO::_update(), which writes every mapped column and could overwrite concurrent changes to other fields.

One consequence: Repository::stampModified() no longer goes through edit(), so it does not call Hook::call('Submission::edit', ...). This is on purpose. Going through edit() would fire that hook in situations where it never fired before, and third-party plugins would see that as a change in behaviour.

Why the job runs synchronously

The shared job sets the sync connection.

The reason: default_connection is database. If job_runner is off and no worker is running, a queued job never runs, and last_modified is never written. The config template discourages job_runner on high-traffic sites, so this can happen in practice.

Let me know if you would rather have it go through the normal queue.

Finding the parent entity

This used to be repeated at each dispatch point. It is now in one place, in the shared job, as a match on the event origin.

I looked at building on getParentColumn(), which several DAOs already implement. I chose the match instead, to keep the change small: CitationDAO is a legacy DAO and does not implement it, and adding it there felt like a bigger change than this fix needs. Happy to do it the other way if you prefer.

What gets stamped

Only submissions.last_modified. The intermediate events route the cascade, they do not stamp their own level. That is enough for the OAI-PMH datestamp, which uses GREATEST(submissions, issues, publications).

One thing I noticed while testing: on the citation path publications.last_modified changes anyway, because saving citations goes through Repo::publication()->edit(). That is existing behaviour, not something this change adds.

Bulk operations

setAuthorsOrder() and deleteByPublicationId() emit one event per operation, not one per row. There are tests for the count.

One side effect worth mentioning. importCitations() calls deleteByPublicationId() and then insertObject() for each entry, so a bulk import emits one event per operation it performs internally: replacing 15 references with 20 emits 21. I did not add a dispatch in importCitations() itself, since the inner calls already cover it. Suppressing them would need a flag through the DAO, which seemed more than this fix needs. Happy to add the suppression if you think it's worth it.

Tests

Thirteen tests in three files: the shared job, the author repository, the citation DAO. I checked that each one fails when I break it on purpose.

importCitations() has unit tests only. It takes a raw citation list and is not reachable from the editorial workflow, so I could not test it by hand.

LibraryClasses and LibraryJobs pass locally. The only failures are two in FilterDAOTest, which fail because a backup dump is missing in my environment.

Issue #13062, opened 2026-07-17
Add DataCite support for peer review DOIs


taslangraham:
> [!IMPORTANT]
> Datacite does not currently support assigning unique DOIs to article versions. When support is added, the peer review export will need to be updated accordingly.

Related to:

- [x] https://github.com/pkp/pkp-lib/issues/11332

With the addition of internal support for peer review DOIs, support for including these in Crossref metadata exports and deposits is necessary. It will need to be added only for public peer reviews with proper reviewer identity anonymity taken into consideration

DataCite schema: https://schema.datacite.org/meta/kernel-4/metadata.xsd
DataCite schema documentation: https://datacite-metadata-schema.readthedocs.io/en/4.7/
Guide on XML validation: https://support.datacite.org/docs/how-do-i-validate-doi-metadata

**Steps to test the changes:**
Prerequisites:

- Ensure Datacite plugin is enabled, selected as DOI registration agency, and configured with test credentials
---
1. Take a submission through the editorial process and produce a published article with at least one public peer review.
2. Ensure peer review DOIs are enabled.
3. Export Datacite article metadata (meant for external deposit). This will produce files depending on the following scenarios:
     - If DOIs for multiple submissions are selected for export, a zip file will be generated containing an XML file for each selected submission. However, if only a single submission's DOI is to be exported, then a single XML file will be generated and downloaded. This behaviour predates this ticket.
     - If there are multiple public peer reviews to be exported (i.e a submission has multiple public peer reviews or multiple submissions with public peer reviews were selected), then a zip file will be generated containing an XML for each of the peer reviews. However, if only a single peer review is to be exported, then a single XML file will be generated and downloaded. 
4. Confirm peer review XML validity by following the guide at https://support.datacite.org/docs/how-do-i-validate-doi-metadata.
5. Deposit Datacite article metadata. Observe the change in status for the article DOI and Peer Review DOI(s). 

**What application are you using?**
  OJS 3.6

**Additional information**
N/A

**Pull Requests**
ojs: https://github.com/pkp/ojs/pull/5715
ops (submodule update): https://github.com/pkp/ops/pull/1363
omp (submodule update): https://github.com/pkp/omp/pull/2422
pkp-lib: https://github.com/pkp/pkp-lib/pull/13119

taslangraham:
[...]

PR #13119, opened 2026-07-30
pkp/pkp-lib#13062 Add support for peer review DOI export & deposit in Datacite plugin

taslangraham:
For pkp/pkp-lib#13062


jobs/doi/DepositSubmission.php
bozana, line 70: `$submissionDepositResults['hasErrors']` reflects the whole batch (article + all galleys, later also all versions), since `exportAndDeposit()` uses one shared error flag across every item it deposits separately. So a single failing galley in DataCite blocks peer-review deposit here too, even when the article itself succeeded — Crossref isn't affected, since it deposits all versions within one XML. Maybe then to check the publication's own DOI status per-review (in the foreach loop below) instead?

Issue #13116, opened 2026-07-30
[Editable Reviews] API Additions /{reviewAssignmentId}/consider


jardakotesovec:
## Use cases

- When the 'Read review' is opened - it updates review assignment to be 'viewed' (using considered column). As we are moving it to vue.js there needs to be API to trigger the 'viewed'
- Once the reviewer wants to confirm the review - clicking on 'Confirm' needs to update considered&considered_date columns.
Proposed endpoint

```
PUT  submissions/{submissionId}/reviewAssignments/{reviewAssignmentId}/consider
{ "considered": "viewed" | "considered" | "unconsidered" }
```

## Notes

- Important for this feature are 'viewed' and 'considered' state transitions - other (revert, thank you) still will be handled via grid actions.
In terms of the implementation details, it would be following current logic, just moving it to API endpoints from the grid actions.
- Only difference is that in the new UI - once its confirmed (considered) - than we don't let them to do it. So API should return validation error.

**Pull Requests**
pkp-lib: https://github.com/pkp/pkp-lib/pull/13138
ojs (submodule update): https://github.com/pkp/ojs/pull/5726
ops (submodule update): https://github.com/pkp/ops/pull/1375
omp (submodule update): https://github.com/pkp/omp/pull/2434

taslangraham:
@touhidurabir Could you review the PRs listed [here](https://github.com/pkp/pkp-lib/issues/13116#issue-5019341198)?

These PRs introduce a new endpoint, `[PUT] submissions/{submissionId}/reviewAssignments/{reviewAssignmentId}/consider`, which allows editors to update the considered status of a review assignment. The endpoint supports three states: `viewed`, `considered`, and `unconsidered`.

The `considered` field accepts one of `ReviewAssignment::REVIEW_ASSIGNMENT_VIEWED`, `ReviewAssignment::REVIEW_ASSIGNMENT_CONSIDERED`, or `ReviewAssignment::REVIEW_ASSIGNMENT_UNCONSIDERED`. I opted to use the existing integer constants instead of strings(which are suggested in the issue description), since that is the more common pattern used throughout the codebase.

The logic to handle each state change was taken directly from `PKPReviewerGridHandler`. The corresponding methods are:

[...]

PR #13138, opened 2026-08-04
pkp/pkp-lib#13116 Add API endpoint to update the `considered` status of a review assignment

taslangraham:
For pkp/pkp-lib#13116

### DIFF
--- a/api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
+++ b/api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
@@ -17,20 +17,26 @@ namespace PKP\API\v1\submissions\reviewAssignments;
 
 use APP\core\Application;
 use APP\facades\Repo;
+use APP\orcid\actions\SendReviewToOrcid;
 use APP\submission\Submission;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Route;
 use PKP\core\Core;
+use PKP\core\PKPApplication;
 use PKP\core\PKPBaseController;
 use PKP\core\PKPRequest;
+use PKP\log\event\PKPSubmissionEventLogEntry;
+use PKP\notification\Notification;
 use PKP\security\authorization\ContextAccessPolicy;
 use PKP\security\authorization\SubmissionAccessPolicy;
 use PKP\security\authorization\UserRolesRequiredPolicy;
 use PKP\security\Role;
+use PKP\security\Validation;
 use PKP\services\PKPSchemaService;
 use PKP\stageAssignment\StageAssignment;
+use PKP\submission\reviewAssignment\ReviewAssignment;
 
 class ReviewAssignmentController extends PKPBaseController
 {
@@ -67,8 +73,13 @@ class ReviewAssignmentController extends PKPBaseController
             Route::get('{reviewAssignmentId}', $this->get(...))
                 ->name('submission.reviewAssignment.get')
                 ->whereNumber('reviewAssignmentId');
+
             Route::put('{reviewAssignmentId}', $this->edit(...))
                 ->name('submission.reviewAssignment.edit')
+                ->whereNumber('reviewAssignmentId');
+
+            Route::put('{reviewAssignmentId}/consider', $this->consider(...))
+                ->name('submission.reviewAssignment.consider')
                 ->whereNumber('reviewAssignmentId');
         });
     }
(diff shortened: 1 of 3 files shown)


api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+            'dateConsidered' => Core::getCurrentDate(),
+        ];
+
+        if (!$reviewAssignment->getDateCompleted()) {
+            // Editor completes the review.
+            $newReviewData['dateConfirmed'] = $newReviewData['dateCompleted'] = Core::getCurrentDate();
+        }
+
+        // Trigger an update of the review round status
+        Repo::reviewAssignment()->edit($reviewAssignment, $newReviewData);
```
touhidurabir, line 284: we need to re-fetch the review assignment after edit as pre-fetch `$reviewAssignment` still does up have proper data assignment after edit . so below the `$reviewAssignment->isRead()` will not reflect the proper update and may not fire/store the event log details .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+        $isAssignedSubEditor = StageAssignment::withSubmissionIds([$submission->getId()])
+            ->withStageIds([$reviewAssignment->getStageId()])
+            ->withRoleIds([Role::ROLE_ID_SUB_EDITOR])
+            ->withUserId($user->getId());
+
+        if (!$isManager && !$isAssignedSubEditor->exists()) {
+            return response()->json([
+                'error' => __('api.403.unauthorized'),
+            ], Response::HTTP_FORBIDDEN);
+        }
```
touhidurabir: we have this check for every controller method, so can we centralise it . can the `StageRolePolicy` help us here ?

Also do we allow a recommendation only editor to update it ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+        $reviewAssignment = Repo::reviewAssignment()->get($reviewAssignmentId, $submission->getId());
+
+        if (!$reviewAssignment) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $user = $this->getRequest()->getUser();
+        $isManager = $user->hasRole([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $submission->getData('contextId'));
```
touhidurabir: if we are including the `ROLE_ID_SITE_ADMIN` here and above, why not just make it part of `roleAuthorizer` middleware of route groups ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+    /**
+     * Update the considered status of a review assignment.
+     * Accepts the following values for the `considered` field:
+     * - ReviewAssignment::REVIEW_ASSIGNMENT_VIEWED
+     * - ReviewAssignment::REVIEW_ASSIGNMENT_CONSIDERED
+     * - ReviewAssignment::REVIEW_ASSIGNMENT_UNCONSIDERED
+     */
+    public function consider(Request $illuminateRequest): JsonResponse
+    {
+        $submission = $this->getAuthorizedContextObject(Application::ASSOC_TYPE_SUBMISSION); /** @var Submission $submission */
```
touhidurabir, line 202: should submission have check if it is still in review stage or moved to copy editing or production stage ? do we plan to allow editing at those stages ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+        }
+
+        // A review assignment that an editor has already considered cannot be considered again.
+        if ($considered === ReviewAssignment::REVIEW_ASSIGNMENT_CONSIDERED && $reviewAssignment->isRead()) {
+            return response()->json([
+                'error' => __('api.review.assignments.alreadyConsidered'),
+            ], Response::HTTP_CONFLICT);
+        }
+
+        match ($considered) {
```
touhidurabir: just a suggestion , how about we return re-fetched review assignment form the match methods ? so less query if we need to re-fetch anyway in any of the match methods and reuse the refetch one .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +205,161 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+                'error' => __('api.403.unauthorized'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $considered = (int)$illuminateRequest->input('considered');
+        if (!in_array($considered, [ReviewAssignment::REVIEW_ASSIGNMENT_VIEWED, ReviewAssignment::REVIEW_ASSIGNMENT_CONSIDERED, ReviewAssignment::REVIEW_ASSIGNMENT_UNCONSIDERED])) {
+            return response()->json([
+                'error' => __('api.reviews.assignments.422.missingOrInvalidConsideredValue'),
+            ], Response::HTTP_UNPROCESSABLE_ENTITY);
+        }
```
touhidurabir, line 223: suggestion : we can use simple `$request->validate` as https://laravel.com/docs/12.x/validation#quick-writing-the-validation-logic . up to you if want to update .

touhidurabir (changes requested):
@taslangraham please see the following points . 

Along with these, wondering if this is possible to centralise the 3 actions (e.g. read/consider/unconsider) as now have almost similar logic in PKPReviewerGridHandler . so should we centralise it and is that worth of the time/effort .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: Currently, editors are allowed to edit review information even after the submission has moved beyond the review stage.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: > Also do we allow a recommendation only editor to update it ?

Existing behaviour allows recommend only editors to make changes to review. However, they can only recommend editorial decisions (accept submission, decline submission, etc). This issue doesn't change any of the existing behaviour; it only exposes certain functionality via the API.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: Since it's just a simple check, I'll keep it as is.

taslangraham:
Thanks for the review @touhidurabir. I've made changes based on your suggestions. These can be reviewed at this commit: https://github.com/pkp/pkp-lib/pull/13138/changes/21d5428565f29b9327257602eff3f5aeb6f6f957. Please take a look.


To answer your question, namely:
> wondering if this is possible to centralise the 3 actions (e.g. read/consider/unconsider) as now have almost similar logic in PKPReviewerGridHandler . so should we centralise it and is that worth of the time/effort .

The goal of this endpoint is to have the UI call it whenever the user changes the consideration status of a review. Once that is in place, the corresponding logic in `PKPReviewerGridHandler` is expected be removed. Since these will soon be removed, I don't think it's worth centralizing the logic and refactoring `PKPReviewerGridHandler` to use it, as that would only add changes to code which will soon become obsolete.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: Usage of `StageRolePolicy` was added.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -194,4 +178,157 @@ public function edit(Request $illuminateRequest): JsonResponse
...
+        ];
+
+        if (!$reviewAssignment->getDateCompleted()) {
+            // Editor completes the review.
+            $newReviewData['dateConfirmed'] = $newReviewData['dateCompleted'] = Core::getCurrentDate();
+        }
+
+        // Trigger an update of the review round status
+        Repo::reviewAssignment()->edit($reviewAssignment, $newReviewData);
+        $reviewAssignment = Repo::reviewAssignment()->get($reviewAssignment->getId(), $reviewAssignment->getsubmissionId());
```
touhidurabir: `getsubmissionId` is wrong, should be `getSubmissionId`, capital `S` . But perhaps best to use `->getData('submissionId')` . Here and other places .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -81,6 +93,7 @@ public function authorize(PKPRequest $request, array &$args, array $roleAssignme
         $this->addPolicy(new UserRolesRequiredPolicy($request), true);
         $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
         $this->addPolicy(new SubmissionAccessPolicy($request, $args, $roleAssignments));
+        $this->addPolicy(new StageRolePolicy([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SUB_EDITOR]));
```
touhidurabir: this looks better, few points to double test/check 
 - should the `ROLE_ID_SITE_ADMIN` be included ? 
 - we pass not `stageId` so it determine from the submission. should we pass the stage id as review stage so that a a editor who does not have access to copyediting/production stage but have access to review stage can still make the API calls ?
 - what is the impact of OMP specific internal review stage is there is any to concern with

locale/en/api.po
```diff
@@ -505,3 +505,9 @@ msgstr "The \"{$field}\" field is required."
 
 msgid "api.submissions.reviews.422.uneditableFields"
 msgstr "Your request included fields that are not editable through this endpoint."
+
+msgid "api.review.assignments.alreadyConsidered"
+msgstr "The review assignment has already been considered and cannot be considered again."
+
+msgid "api.reviews.assignments.422.missingOrInvalidConsideredValue"
```
touhidurabir: these 2 locale keys has `review vs reviews`, better to use one single format

touhidurabir (changes requested):
@taslangraham please see the review comments below .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: > should the ROLE_ID_SITE_ADMIN be included 

Internally, the policy grants access to Admins, but I guess it could be included for completeness.


> we pass not stageId so it determine from the submission. should we pass the stage id as review stage so that a a editor who does not have access to copyediting/production stage but have access to review stage can still make the API calls ?

 Default behaviour is that section editors have access to all stages, but I see that those stage access can be toggled off in `Users & Roles` settings. I think the best approach here is to simply accept `stageId` as a query param since the review object is not be available at this point. 

We could also call the `ReviewAssignmentRequiredPolicy` policy before the `StageRolePolicy` then use `$this->getAuthorizedContextObject(Application::ASSOC_TYPE_REVIEW_ASSIGNMENT)` to get the review object in the `authorize`, but I don't see any existing code that uses `getAuthorizedContextObject` in the `authorize` method, so I'm not sure that's a good approach.


> what is the impact of OMP specific internal review stage is there is any to concern with

Since the initial idea was that the editor has access to all stages, there wouldn't have been an OMP specific concern. However, whatever solution we use from the previous bullet point should handle this.

taslangraham: Thinking more about this, I think the `stageId` should be a required path in the endpoint. So this endpoint would become `api/v1/submissions/{submissionId}/reviewAssignments/{reviewId}/{stageId}/review`. What do you think @touhidurabir?

touhidurabir: > Default behaviour is that section editors have access to all stages, but I see that those stage access can be toggled off in Users & Roles settings. I think the best approach here is to simply accept stageId as a query param since the review object is not be available at this point.

that can work and probably the only way to do it. I point is the editor or section editor making the API request, do they have the permission of that stage ? and review stage id is 3(external) and 2(internal, OMP only) . for OJS, I think it's simple as just to pass 3 but OMP make it difficult as we dont know what stage this assignment is associated with without fetching the object from DB. And determining from submission which has progressed to copy editing or production and not having permission for Editor/Section Editor of those stage will fail via `StageRolePolicy`  . 

> We could also call the ReviewAssignmentRequiredPolicy policy before the StageRolePolicy then use $this->getAuthorizedContextObject(Application::ASSOC_TYPE_REVIEW_ASSIGNMENT) to get the review object in the authorize, but I don't see any existing code that uses getAuthorizedContextObject in the authorize method, so I'm not sure that's a good approach.

lets not do that as `ReviewAssignmentRequiredPolicy` is an internal policy and should not be directly used in controller/handler authorize method .

taslangraham:
@touhidurabir Thanks for the review. I've made changes(commit: https://github.com/pkp/pkp-lib/pull/13138/changes/2eba4c985226fa2fb67fb1e5234e4115e824452f) based on your feedback; please take a look.  Regarding this [discussion](https://github.com/pkp/pkp-lib/pull/13138#discussion_r3746844170), specifically that around your 3rd bullet point, I've updated the endpoints to include the `stageId` as part of the route. This ensures that the needed `stageId` is always present to do the proper stage access checks via `StageRolePolicy`.

touhidurabir:
> @touhidurabir Thanks for the review. I've made changes(commit: [2eba4c9](https://github.com/pkp/pkp-lib/commit/2eba4c985226fa2fb67fb1e5234e4115e824452f)) based on your feedback; please take a look. Regarding this [discussion](https://github.com/pkp/pkp-lib/pull/13138#discussion_r3746844170), specifically that around your 3rd bullet point, I've updated the endpoints to include the `stageId` as part of the route. This ensures that the needed `stageId` is always present to do the proper stage access checks via `StageRolePolicy`.

@taslangraham sorry about the suggestion to use the `StageRolePolicy`  but I am going to suggest not to use it after thinking about a bit more as 
 - API is forcing the request maker to know the exact stage id of resource which is require to check the access level which is not responsibility API request 
 - one can pass a stage id of copy editing or production but have no access to review stage and it will pass and allow to modify the resource. we can have check if resource stage and passed stage id is equal after the authorization of `StageRolePolicy ` but making authoritative check in 2 layer (API authorization and execution)
 
seems we do not have a good policy available here for this purpose so 
 1. either we build a new policy
 2. use simple custom check as you previously had like 
 ```php
 protected function canAccessReviewAssignment(Submission $submission, ReviewAssignment $reviewAssignment): bool
 {
     $userRoles = (array) $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);

     // Managers and site admins shoule be able to  always act (USER_ROLES folds in the site-level admin group).
     if (array_intersect([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $userRoles)) {
         return true;
     }

     // Otherwise the user must be a sub-editor assigned to the review assignment's OWN stage.
     return StageAssignment::withSubmissionIds([$submission->getId()])
         ->withStageIds([$reviewAssignment->getStageId()])
         ->withRoleIds([Role::ROLE_ID_SUB_EDITOR])
         ->withUserId($this->getRequest()->getUser()->getId())
         ->exists();
 }

// Call it in each handler where needed

 if (!$this->canAccessReviewAssignment($submission, $reviewAssignment)) {
     return response()->json(['error' => __('api.403.unauthorized')], Response::HTTP_FORBIDDEN);
 }
 ```
(NOTE : non tested code, something like that)

I think no 2 is easier for us and if we need more like this kind of check in other controller, then we can refactor this to a policy for multiple controller/handler usage. Up to you which path you want to take

after this change, no more review require and should be good to merge . 

sorry for the wrong suggestion .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -62,14 +68,23 @@ public function getGroupRoutes(): void
             self::roleAuthorizer([
                 Role::ROLE_ID_MANAGER,
                 Role::ROLE_ID_SUB_EDITOR,
+                Role::ROLE_ID_SITE_ADMIN,
```
touhidurabir, line 71: @taslangraham another question, does the `ROLE_ID_ASSISTANT` be able to modify it  as the `PKPReviewerGridHandler` allow that access ?

taslangraham: I missed that in my initial review of `PKPReviewerGridHandler`. Once assistant role is given Review stage access, then an assistant can view and edit reviews. So yes, `ROLE_ID_ASSISTANT` should be included.

taslangraham:
> > @touhidurabir Thanks for the review. I've made changes(commit: [2eba4c9](https://github.com/pkp/pkp-lib/commit/2eba4c985226fa2fb67fb1e5234e4115e824452f)) based on your feedback; please take a look. Regarding this [discussion](https://github.com/pkp/pkp-lib/pull/13138#discussion_r3746844170), specifically that around your 3rd bullet point, I've updated the endpoints to include the `stageId` as part of the route. This ensures that the needed `stageId` is always present to do the proper stage access checks via `StageRolePolicy`.
> 
> @taslangraham sorry about the suggestion to use the `StageRolePolicy` but I am going to suggest not to use it after thinking about a bit more as
> 
> * API is forcing the request maker to know the exact stage id of resource which is require to check the access level which is not responsibility API request
> * one can pass a stage id of copy editing or production but have no access to review stage and it will pass and allow to modify the resource. we can have check if resource stage and passed stage id is equal after the authorization of `StageRolePolicy ` but making authoritative check in 2 layer (API authorization and execution)
> 
> seems we do not have a good policy available here for this purpose so
> 
> 1. either we build a new policy
> 2. use simple custom check as you previously had like
> 
> ```
> protected function canAccessReviewAssignment(Submission $submission, ReviewAssignment $reviewAssignment): bool
> {
>     $userRoles = (array) $this->getAuthorizedContextObject(Application::ASSOC_TYPE_USER_ROLES);
> 
>     // Managers and site admins shoule be able to  always act (USER_ROLES folds in the site-level admin group).
>     if (array_intersect([Role::ROLE_ID_MANAGER, Role::ROLE_ID_SITE_ADMIN], $userRoles)) {
>         return true;
>     }
> 
>     // Otherwise the user must be a sub-editor assigned to the review assignment's OWN stage.
>     return StageAssignment::withSubmissionIds([$submission->getId()])
>         ->withStageIds([$reviewAssignment->getStageId()])
>         ->withRoleIds([Role::ROLE_ID_SUB_EDITOR])
>         ->withUserId($this->getRequest()->getUser()->getId())
>         ->exists();
> }
> 
> // Call it in each handler where needed
> 
> if (!$this->canAccessReviewAssignment($submission, $reviewAssignment)) {
>     return response()->json(['error' => __('api.403.unauthorized')], Response::HTTP_FORBIDDEN);
> }
> ```
> 
> (NOTE : non tested code, something like that)
> 
> I think no 2 is easier for us and if we need more like this kind of check in other controller, then we can refactor this to a policy for multiple controller/handler usage. Up to you which path you want to take
> 
> after this change, no more review require and should be good to merge .
> 
> sorry for the wrong suggestion .

I reintroduced the initial custom check I had.

taslangraham:
Updated based on your latest review and merged, @touhidurabir. Thanks for the review!

Issue #13135, opened 2026-08-03
[Localization] | Bootstrap - Locale::lookup() returns null when locale list contains "und" (PHP 8.4.24 / 8.5.9)


hacsnemail:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

After upgrading PHP to versions 8.4.24 and 8.5.9, our OJS installation started failing during application bootstrap.

The issue causes:
plugin registration failures during scheduler execution,
frontend pages rendering without CSS because application initialization fails before the page is fully generated.

The exception originates from the following code in `lib/pkp/classes/i18n/Locale.php`:
```
$locales = array_keys($this->getWeblateLocaleNames());
\Locale::setDefault(\Locale::lookup($locales, $locale, true));
```

Diagnostic logging confirmed that:
the requested locale was cs,
the primary locale was also cs,
`Locale::lookup()` returned `null`.

Further investigation showed that the locale list returned by `getWeblateLocaleNames()` contains the locale identifier `und`. When this identifier is present, `Locale::lookup()` returns `null`, even though the requested locale (cs) exists in the list.

### Steps to Reproduce

1. Install OJS 3.5.0-4.
2. Upgrade PHP to version 8.4.24 or 8.5.9.
3. Run the scheduler:
`php lib/pkp/tools/scheduler.php run`
or open any frontend page.

4. OJS fails during bootstrap with a TypeError.

The behaviour can also be reproduced outside OJS:

```
<?php

declare(strict_types=1);

var_dump(\Locale::lookup(['cs'], 'cs', true));
// string(2) "cs"

var_dump(\Locale::lookup(['cs', 'und'], 'cs', true));
// NULL
```

### Expected Result

OJS should initialize successfully regardless of the presence of `und` in the Weblate locale list.

If `Locale::lookup()` cannot resolve a locale, OJS should gracefully fall back to the primary locale instead of terminating with a fatal error.

### Actual Result

`Locale::lookup()` returns `null`, which is passed directly to `Locale::setDefault()`. 

This results in:
`TypeError: Locale::setDefault(): Argument #1 ($locale) must be of type string, null given`

Application bootstrap fails, plugins cannot be registered, scheduled tasks report errors, and frontend pages are not rendered correctly.

### Environment Details

Server OS: Debian 12
PHP: 8.4.24 and 8.5.9
Database: PostgreSQL
Web server: Apache
Issue reproduced from both CLI (scheduler.php) and the web interface.

### Application Version

[...]

PR #13140, opened 2026-08-04
Fix Intl locale initialization on PHP 8.4

### LINKED ISSUE #13139: Locale::lookup() receives more than 100 language tags and causes TypeError on PHP 8.4
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

OJS 3.5.0-5 fails to initialize theme plugins under PHP 8.4 because `PKP\i18n\Locale::setLocale()` passes 799 language tags to PHP's `\Locale::lookup()`.

PHP documents a maximum of 100 items for the `$languageTag` parameter:

see https://www.php.net/manual/en/locale.lookup.php

With PHP 8.4, `\Locale::lookup()` returns `null` in this case. The result is passed directly to `\Locale::setDefault()`, which requires a string and throws a `TypeError`.

This prevents the default theme from being initialized. In our installation, the frontend was rendered without theme CSS and JavaScript and installed themes did not appear in the theme selection.

### Steps to Reproduce

The following diagnostic is intended to be run from the root directory of a packaged OJS installation. `languages.json` is present in the OJS release package at `lib/pkp/lib/weblateLanguages/languages.json`, although it is not tracked at that path in the standalone `pkp/pkp-lib` source repository.

```php
<?php

$file = __DIR__ . '/lib/pkp/lib/weblateLanguages/languages.json';
$data = json_decode(file_get_contents($file), true);
$locales = array_keys($data);

var_dump([
    'php' => PHP_VERSION,
    'intl' => phpversion('intl'),
    'icu' => INTL_ICU_VERSION,
    'count' => count($locales),
    'contains_en' => in_array('en', $locales, true),
    'lookup' => Locale::lookup($locales, 'en', true),
]);
```

The language file is valid JSON, contains 799 entries and includes `en`.

### Expected Result

`PKP\i18n\Locale::setLocale()` should always pass a valid locale string to `\Locale::setDefault()`.

Loading the default theme should complete successfully, and the frontend should include the theme's CSS and JavaScript assets.

The implementation should not pass more than the documented maximum of 100 language tags to `\Locale::lookup()`.

### Actual Result

The same diagnostic produces different return values under PHP 8.3 and PHP 8.4
while using the same ICU version.

### PHP 8.3.33

```text
php: 8.3.33
intl: 8.3.33
icu: 70.1
count: 799
contains_en: true
lookup: ""
```

### PHP 8.4.24

```text
php: 8.4.24
intl: 8.4.24
icu: 70.1
count: 799
contains_en: true
lookup: NULL
```

PHP 8.3 returns an empty string, which allows Locale::setDefault() to be called without a type error. PHP 8.4 returns null, exposing the problem.

The following error prevents the theme plugin from being initialized: `TypeError: Locale::setDefault(): Argument #1 ($locale) must be of type string, null given`

Relevant stack trace:

```text
#0 lib/pkp/classes/i18n/Locale.php(163): Locale::setDefault(NULL)
#1 lib/pkp/classes/i18n/Locale.php(143): PKP\i18n\Locale->setLocale('en')
#2 lib/pkp/classes/i18n/Locale.php(475): PKP\i18n\Locale->getLocale()
#3 lib/pkp/classes/i18n/Locale.php(114): PKP\i18n\Locale->translate(...)
#4 plugins/themes/default/DefaultThemePlugin.php(45): __(...)
#5 lib/pkp/classes/plugins/ThemePlugin.php(131): DefaultThemePlugin->init()
```

### Environment Details

- OJS 3.5.0-5
- PHP 8.4.24
- PHP SAPI: CGI/FastCGI
- Intl: 8.4.24
- ICU: 70.1
- ICU data: 70.1

### Application Version

OJS 3.5.0-5

### Logs

_No response_

### Additional Information

PHP documents a maximum of 100 items for the `$languageTag` parameter of `Locale::lookup()`: [https://www.php.net/manual/en/locale.lookup.php](see documentation)

The current implementation also passes the nullable return value directly to `\Locale::setDefault()`, which requires a string.

A minimal defensive fix would be:

```php
$intlLocale = \Locale::lookup($locales, $locale, true);
\Locale::setDefault($intlLocale ?: $locale);
```

However, the implementation should ideally also avoid passing more than 100 language tags to `Locale::lookup()`.

### PRs
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/13165
- https://github.com/pkp/ojs/pull/5738 (only submodule update)
- https://github.com/pkp/omp/pull/2433 (only submodule update)
- https://github.com/pkp/ops/pull/1374 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/13164
- ~~https://github.com/pkp/ojs/pull/5735~~ (only submodule update)

marianpg:
Fixes #13139

### Summary

- Pass the already validated and supported OJS locale directly to `\Locale::setDefault()`.
- Remove the redundant `\Locale::lookup()` call, which received 799 language tags despite PHP documenting a maximum of 100.
- Add a test that verifies that the PKP locale is propagated to PHP Intl.

### Why

`setLocale()` already validates the locale against the Weblate locale list and the locales supported by the application. A second lookup across all Weblate locales is therefore unnecessary.

Under PHP 8.4, the oversized `\Locale::lookup()` call returns `null`. Passing that value to `\Locale::setDefault()` causes a `TypeError` and prevents theme plugins from being initialized.

### Testing

Tested with:

- PHP 8.4.24
- Intl 8.4.24
- ICU 76.1
- PHPUnit 11.5.11

Targeted test:

```text
php lib/vendor/bin/phpunit --configuration tests/phpunit.xml classes/i18n/LocaleTest.php
```

Result after the fix: `Tests: 7, Assertions: 18`

The test was also run against the previous implementation and reproduced the reported `Locale::setDefault() TypeError`.

### DIFF
--- a/classes/i18n/Locale.php
+++ b/classes/i18n/Locale.php
@@ -159,8 +159,7 @@ class Locale implements LocaleInterface
 
         $this->locale = $locale;
         setlocale(LC_ALL, 'C.utf8', 'C');
-        $locales = array_keys($this->getWeblateLocaleNames());
-        \Locale::setDefault(\Locale::lookup($locales, $locale, true));
+        \Locale::setDefault($locale);
     }
 
     /**
--- a/tests/classes/i18n/LocaleTest.php
+++ b/tests/classes/i18n/LocaleTest.php
@@ -89,6 +89,19 @@ class LocaleTest extends PKPTestCase
             ->getMock();
     }
 
+    public function testSetLocaleSetsIntlDefault()
+    {
+        $defaultLocale = \Locale::getDefault();
+
+        try {
+            Locale::setLocale('pt_BR');
+
+            self::assertSame('pt_BR', \Locale::getDefault());
+        } finally {
+            \Locale::setDefault($defaultLocale);
+        }
+    }
+
     public function testIsLocaleComplete()
     {
         self::assertTrue(Locale::getMetadata('en')->isComplete());


edersotto:
I hit this same bug independently on another stack and did a fair amount of measuring, so let me contribute data rather than an opinion. Short version: **I think the change is right**, but the stated rationale is not the actual cause, and there is a behaviour change that isn't declared yet. Both are fixable in the description.

Environment for everything below: PHP 8.4.24, intl 8.4.24, **ICU 77.1**.

### 1. The 100-tag premise doesn't hold, and it would be a shame for the PR to be judged on it

> Remove the redundant `\Locale::lookup()` call, which received 799 language tags despite PHP documenting a maximum of 100.

I couldn't reproduce the failure as a function of list size:

| list | count | `Locale::lookup($list, 'en', true)` | ICU error |
|---|---|---|---|
| full Weblate list | 783 | `NULL` | `unable to canonicalize lang_tag: U_ILLEGAL_ARGUMENT_ERROR` |
| same list minus `und` | 782 | `'en'` | `U_ZERO_ERROR` |
| `ResourceBundle::getLocales('')` — what 3.4 passes | **881** | `'en'` | `U_ZERO_ERROR` |
| 500 valid tags, no `und` | 500 | `'en'` | `U_ZERO_ERROR` |
| **`['cs','und']`, asking `cs`** | **2** | **`NULL`** | `U_ILLEGAL_ARGUMENT_ERROR` |

Two tags reproduce it; 881 tags don't. The failing element is `und`, which ICU cannot canonicalize *as a language range* — it poisons the entire lookup regardless of list length or position. That's the same conclusion @hacsnemail reached in #13135. I've posted the full measurements in #13139.

This doesn't invalidate the patch — it just means the justification should be "the lookup is redundant and the list can contain non-range tags", not "the list is too long". As written, a reviewer who tests the 100-tag premise will find it false and may reject a correct patch for the wrong reason.

### 2. @asmecher's concern about "OK for all locales" — I think the data answers it

> I suspect it needs work to ensure it'll be OK for all locales

I compared both code paths across the real 3.5 UI locales: the value `lookup()` would have returned versus the raw locale this PR passes, each fed to `\Locale::setDefault()` and then to `IntlDateFormatter(null, LONG, NONE)`:

| OJS locale | `lookup()` returned | via `lookup()` | via `setDefault($locale)` | identical? |
|---|---|---|---|---|
| `en` | `en` | March 15, 2026 | March 15, 2026 | yes |
| `pt_BR` | `pt_br` | 15 de março de 2026 | 15 de março de 2026 | yes |
| `pt` | `pt` | 15 de março de 2026 | 15 de março de 2026 | yes |
| `es` | `es` | 15 de marzo de 2026 | 15 de marzo de 2026 | yes |
| `fr_CA` | `fr_ca` | 15 mars 2026 | 15 mars 2026 | yes |
| `de` | `de` | 15. März 2026 | 15. März 2026 | yes |
| `ca` | `ca` | 15 de març del 2026 | 15 de març del 2026 | yes |
| `be` | `be` | 15 сакавіка 2026 г. | 15 сакавіка 2026 г. | yes |
| `nb_NO` | `nb_no` | 15. mars 2026 | 15. mars 2026 | yes |
| `sr_Cyrl` | `sr_cyrl` | 15. март 2026. | 15. март 2026. | yes |
| `sr_Latn` | `sr_latn` | 15. mart 2026. | 15. mart 2026. | yes |
| `bs_Latn` | `bs_latn` | 15. mart 2026. | 15. mart 2026. | yes |
| `uz_Latn` | `uz_latn` | 15-mart, 2026 | 15-mart, 2026 | yes |
| `zh_Hans` | `zh_hans` | 2026年3月15日 | 2026年3月15日 | yes |
| `en_US` | `en_us` | March 15, 2026 | March 15, 2026 | yes |
| `xx_YY` (not a locale) | `''` | March 15, 2026 | March 15, 2026 | yes (ICU root) |

Identical in every case — `lookup()` only lowercases the tag, and ICU treats `pt_BR` and `pt_br` the same. The locales the CRAFT OA alignment introduced (`sr_Cyrl`, `zh_Hans`, `uz_Latn`, `bs_Latn`) all resolve correctly, and an unknown tag degrades to ICU root rather than erroring. `\Locale::setDefault()` also round-trips exactly (`setDefault('pt_BR')` → `getDefault() === 'pt_BR'`), so the new assertion in `LocaleTest` holds.

Worth stating explicitly in the description, since it's the strongest argument for the patch: by the time this line runs, `setLocale()` has already validated `$locale` through `isLocaleValid()` and `isSupported()` and fallen back to the primary locale otherwise. The lookup really is redundant.

### 3. There is an undeclared behaviour change, and it's a real one

This PR doesn't just remove dead work — it makes `\Locale::setDefault()` effective for the first time. On **every** current 3.5 installation the lookup fails, so the ICU default locale is silently `en_US_POSIX`, whatever the journal's locale is (on PHP ≤ 8.4.23 that's the `''` path; on ≥ 8.4.24 it's the `TypeError` path):

| what gets set | `Locale::getDefault()` | `IntlDateFormatter::LONG` | `NumberFormatter::DECIMAL` |
|---|---|---|---|
| `''` (today, every 3.5 install) | `en_US_POSIX` | March 15, 2026 | 1234.56 |
| `pt_BR` (this PR) | `pt_BR` | 15 de março de 2026 | 1.234,56 |

Application code shouldn't notice — `grep -rE "new (IntlDateFormatter|NumberFormatter|Collator)\(" lib/pkp/classes classes` returns nothing, so the core never relies on the implicit ICU default. The exposure is vendor code (Carbon/Laravel) and third-party plugins that format via the implicit default. I think that's the correct direction and matches the original intent of #9302, but it is a change in observable output and reviewers will want it called out rather than discovered.

### 4. Two smaller notes

**Target branch.** This is against `stable-3_5_0`, but `main` carries the same line (`classes/i18n/Locale.php:162-163`) and currently has no PR.

**Optional belt-and-braces.** If you'd rather not remove the lookup outright, the minimal version that fixes all supported PHP versions with no behaviour surprise is:

```php
$locales = array_diff(array_keys($this->getWeblateLocaleNames()), ['und']);
\Locale::setDefault(\Locale::lookup($locales, $locale, true) ?: $locale);
```

`?:` rather than `??` matters: `lookup()` legitimately returns `''` (with `U_ZERO_ERROR`) when nothing matches, and `''` is what PHP 8.2/8.3 return in the error case too — `??` would fix 8.4.24 and leave 8.2/8.3 still setting `en_US_POSIX`. I'm running this variant in production as a hotfix and it's forward-compatible with this PR: once the lookup is gone the guard is simply inert. Either way, `und` should stay in `getWeblateLocaleNames()` itself — `isLocaleValid()` and `getSubmissionLocaleDisplayNames()` legitimately expose it as "Unknown language" in the submission language selector.

Worth adding to the description as well: this same `TypeError` aborts `tools/upgrade.php` mid-migration on PHP ≥ 8.4.24, so the fix is on the 3.4 → 3.5 upgrade path, not only on the frontend. Details in #13139.

Happy to test any revision on 8.4.24.

bozana:
I will close this PR in favor of this one: https://github.com/pkp/pkp-lib/pull/13165

bozana:
Thanks a lot for the contribution!!!

PR #13164, opened 2026-08-10
pkp/pkp-lib#13139 Fix Locale::lookup() canonicalization failure crash…

bozana:
…ing Locale::setLocale()

s. https://github.com/pkp/pkp-lib/issues/13139

### DIFF
--- a/classes/i18n/Locale.php
+++ b/classes/i18n/Locale.php
@@ -44,6 +44,7 @@ use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use RecursiveRegexIterator;
 use RegexIterator;
+use ResourceBundle;
 use Sokil\IsoCodes\Database\Countries;
 use Sokil\IsoCodes\Database\Currencies;
 use Sokil\IsoCodes\Database\LanguagesInterface;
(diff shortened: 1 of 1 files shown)


classes/i18n/Locale.php
```diff
@@ -159,8 +160,11 @@ public function setLocale($locale): void
 
         $this->locale = $locale;
         setlocale(LC_ALL, 'C.utf8', 'C');
-        $locales = array_keys($this->getWeblateLocaleNames());
-        \Locale::setDefault(\Locale::lookup($locales, $locale, true));
+        // ICU's own list (not Weblate's) avoids "und" and similar tags breaking Locale::lookup().
+        // Cast guards against null on canonicalization failure (PHP 8.4.24+); setDefault() requires a string.
+        // Kept as a backup for any future/third-party code relying on this implicit default - nothing in the
+        // codebase currently reads Locale::getDefault().
+        \Locale::setDefault((string) \Locale::lookup(ResourceBundle::getLocales(''), $locale, true));
```
touhidurabir: should be it `\Locale::setDefault(\Locale::lookup(ResourceBundle::getLocales(''), $locale, true) ?: $locale);` to make sure we do not set something that result to return as '' from `\Locale::lookup` not as `''` or default `en_US_POSIX` which would be set a default one . 

Or our `$this->getPrimaryLocale()` is guaranteed to be always that match with ICU ? 

if it is guaranteed, then we will never set something like `''` or `en_US_POSIX` but what we get from `$locale = $this->getPrimaryLocale();` as fallback if the check `if (!$this->isLocaleValid($locale) || !$this->isSupported($locale))` failed .

touhidurabir (approved):
@bozana just one question to consider but seems ok to me .

classes/i18n/Locale.php
bozana: Maybe to use:
`\Locale::setDefault(\Locale::lookup(ResourceBundle::getLocales(''), $locale, true) ?: LocaleInterface::DEFAULT_LOCAL
`
Primary locale does not necessarily match with ICU -- e.g. 'crn'. The same applies to `$locale`.

bozana:
@touhidurabir, could you take a look at this new solution, to fall back to LocaleInterface::DEFAULT_LOCALE which is 'en' and matches with ICU.

classes/i18n/Locale.php
touhidurabir: @bozana I think this is better .

Issue #13162, opened 2026-08-10
[Main Feature] | Review Form Edit sends email with incorrect due dates


Vitaliy-1:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When review edit form is saved with a modification of the review due dates, the system, send email with the old (before edit) dates.

### Steps to Reproduce

- Login with User/Role 'Editor'
- Navigate to the manuscript, then Edit
- Update a date
- Check the activity log

### Expected Result

Dates in the email to reviewer should be updated accordingly. 

### Actual Result

The date is wrong

### Application Version

OJS, OMP starting from 3.4

Vitaliy-1:
PR main: https://github.com/pkp/pkp-lib/pull/13163
tests main: https://github.com/pkp/ojs/pull/5734

PR stable 3.5.0: https://github.com/pkp/pkp-lib/pull/13166
test stable 3.5.0: https://github.com/pkp/ojs/pull/5739

PR stable 3.4.0: https://github.com/pkp/pkp-lib/pull/13167

Vitaliy-1:
@bozana, can you take a look at these PRs?

PR #13167, opened 2026-08-11
pkp/pkp-lib#13162 Fix due dates in edit review form email

Vitaliy-1:


controllers/grid/users/reviewer/form/EditReviewForm.php
```diff
@@ -236,6 +198,45 @@ public function execute(...$functionArgs)
         }
 
         $reviewAssignmentDao->updateObject($reviewAssignment);
+
```
bozana, line 201: This way the email is always sent, not only when something changes, as above within the `if` block, right?
So, maybe to: capture the boolean result of the comparison above, before mutating, then move the notification-creation + email-sending logic (using that cached boolean) here. Something like:
```
$notify = strtotime($reviewAssignment->getDateDue()) != strtotime($this->getData('reviewDueDate'))
    || strtotime($reviewAssignment->getDateResponseDue()) != strtotime($this->getData('responseDueDate'))
    || $reviewAssignment->getReviewMethod() != $this->getData('reviewMethod');
...
...
...
if ($notify) {
   ... 
   this code
}
```

bozana: I see you have solved it in main and 3.5 PR differently, so maybe to here then do the same, to consider `if (isset($notification && ...` below?

bozana (commented):
Hi @Vitaliy-1, I think one change is needed here.

Issue #13154, opened 2026-08-07
[API] | Apply consistent anonymization in submissions API


blesildaramirez:
**Description:**
The same `$anonymize` check is implemented in multiple places, and the submissions API is not consistent in how publication and contributor responses are built.

Review and consolidate the anonymization logic to ensure it is applied consistently across submission-related responses. This is also an opportunity to review which properties should be covered by the anonymization process.

**What application are you using?**
OJS, OMP or OPS version 3.5 / 3.6 (main)

### 3.6

- `pkp-lib`: pkp/pkp-lib#13173
- `ojs`: pkp/ojs#5742
- `omp`: pkp/omp#2436
- `ops`: pkp/ops#1377

### 3.5

- `pkp-lib`: pkp/pkp-lib#13224
- `ojs`: pkp/ojs#5762
- `omp`: pkp/omp#2450
- `ops`: pkp/ops#1390

### 3.4

- `pkp-lib`: pkp/pkp-lib#13362
- `ojs`: pkp/ojs#5842
- `omp`: pkp/omp#2465
- `ops`: pkp/ops#1405

ewhanson:
Hey @blesildaramirez, would you be able to have a look at this one? Thanks!

blesildaramirez:
Hi @ewhanson , could we also port these fixes to 3.5, please? Thanks!

ewhanson:
@blesildaramirez, the port got lost in the shuffle with intermittent test failures. Re-running them now to wrap up porting. Thanks!

ewhanson:
Hey @blesildaramirez, 3.5 has been merged now!

blesildaramirez:
Hi @ewhanson , thank you! Reviewed the changes and I confirm all of them are ported to 3.5. This is working fine when I tested, closing this one.

ewhanson:
Hey @asmecher, I've put together a set of fixes for 3.4 as well. Could you have a look at them. The only changes are in pkp-lib, the applications are just the submodule commits. Thanks!

asmecher:
@ewhanson Thanks, I've reviewed and approved the PR by inspection, but *have not tested it*. If you're satisfied, please merge once the tests pass.

ewhanson:
Thanks @asmecher, I've tested and confirmed it locally already. Will merge once tests are passing.

ewhanson:
Merged, thanks @asmecher!

PR #13173, opened 2026-08-12
pkp/pkp-lib#13154 Apply consistent anonymization in submissions API

ewhanson:


api/v1/submissions/PKPSubmissionController.php
```diff
@@ -1602,6 +1621,11 @@ public function getContributor(Request $illuminateRequest): JsonResponse
             ], Response::HTTP_NOT_FOUND);
         }
 
+        $submissionsToAnonymizeByAuthor = $this->submissionsToAnonymizeByAuthor($submission);
+        if (in_array($submission->getId(), $submissionsToAnonymizeByAuthor)) {
+            return response()->json([], Response::HTTP_OK);
```
blesildaramirez: The default return type for this endpoint is object, I think we should also use that here instead of an empty array.

```suggestion
            return response()->json({}, Response::HTTP_OK);
```
Or maybe we could return HTTP_FORBIDDEN status instead?

classes/submission/maps/Schema.php
```diff
@@ -134,20 +134,22 @@ protected function getSubmissionsListProps(): array
...
+        Enumerable      $userGroups,
+        array           $genres,
+        array           $userRoles,
+        ?Enumerable     $reviewAssignments = null,
+        ?Enumerable     $stageAssignments = null,
+        ?Enumerable     $decisions = null,
+        array           $reviewsToAnonymize = [],
+        ?Enumerable     $reviewerSuggestions = null,
+        ?Enumerable     $stageFiles = null,
+        array $submissionsToAnonymizeByAuthor = [],
```
blesildaramirez: I see that we fixed the spacing in other parameter declarations here, would be nice to do the same here:

```suggestion
        array                  $submissionsToAnonymizeByAuthor = [],
```

blesildaramirez (commented):
Hi Erik, as discussed offline, there are some other remaining submission metadata yet to be anonymized. And then adding here other minor feedback based on my review. Thanks!

blesildaramirez (approved):
I tested the latest changes and the anonymization works for the affected apis. Thanks!

Issue #13117, opened 2026-07-30
[Editable Reviews] API  Additions - {reviewAssignmentId}/review


jardakotesovec:
> [!IMPORTANT]
> This issue introduces a new migration that must be executed:
`php lib/pkp/tools/migration.php "PKP\migration\upgrade\v3_6_0\I13117_AddReviewAssignmentLastModifiedBy" up`

## Use cases

- editor/section editor to be able to see and update the review.
once we update the reviewer submission process to vue.js - it would be used also by reviewers
- author needs to be able read the review if review is opened **Update:** realised that the 'reading review' screen is completely independent for authors. So lets keep this out of the scope and replace this screen separately.
- this will be also used by reviewers who are assigned to the review assignment, but can be deferred as moving the reviewer submission to API will be separate project

## Important considerations
- privateComment should not be revealed to author not even in open review
- for reviewerForms - there is also per field configuration which can be shared with authors, which needs to be considered based on the role
- if the review is changed by editor for the first time (dateCompleted is null) - than consider it as review was completed by that step and fill the `dateCompleted` with current time

## Event logging details
- In general intention is to record any review change from editor. One event per changed field, capturing before and after. Wording from figma:  `The following field was modified in this review: {field name}. Select "View changes" to see a detailed summary of all modifications`. Where value before and after would be saved in settings table.

## Proposed endpoints

Scenario A — no review form (free-text review)

```
GET submissions/{submissionId}/reviewAssignments/{reviewAssignmentId}/review

{
  "reviewAssignmentId": 123,
  "reviewFormId": null,

  "reviewerRecommendationId": 2,
  "comments": "<p>The methodology is sound, but section 3…</p>",
  "privateComments": "<p>For the editor only…</p>"
}

PUT submissions/{submissionId}/reviewAssignments/{reviewAssignmentId}/review

{
  "reviewerRecommendationId": 5,
  "comments": "<p>Edited…</p>",
  "privateComments": ""
}

```

Scenario B — custom review form

```
GET submissions/{submissionId}/reviewAssignments/{reviewAssignmentId}/review

{
  "reviewAssignmentId": 123,
  "reviewFormId": 7,

  "reviewerRecommendationId": 2,

  "reviewFormResponses": {
    "11": "<p>The sample size is too small…</p>",
    "12": 1,
    "13": [0, 2]
  },

  "reviewFormConfig": {
    "id": "reviewContent",
    "fields": [
      {
        "component": "field-rich-textarea",
[...]

Issue #6353, opened 2020-11-18
Review Detail Confirmation button


pmangahis:
**Describe the problem you would like to solve**
A hosted client has pointed out that once we click on the review details button under the referee name, a popup window opens and the referee reports are showing.

If the referee didn't submit the report the same window opens and once click on the confirm button the review reports are recorded.If by mistake once we click on the confirm button and the referee even didn't respond the review is recorded.  

![image](https://user-images.githubusercontent.com/50342731/99570189-6dd58b80-2986-11eb-80b2-3da29b4f3c11.png)

**Describe the solution you'd like**
They would like to have the confirmation button disabled if the referee has not completed the review. In other words, activate the confirm button only after the review has been submitted.

**Who is asking for this feature?**
Editors

**Additional information**
The issue is present in OJS 3.2.1.1

librariam:
+1 from PKP|PS, confirming the same behaviour as reported on 3.1.2.4

pmangahis:
+1 from PKP|PS, confirming the same behaviour in 3.3.0.7

luizbgomide:
I opened a forum question related to this: https://forum.pkp.sfu.ca/t/how-to-revert-a-review-submitted-state/77402

I consider that a warning here (and possibly a way to revert this action) is really important.

asmecher:
All, I think this is behaving as intended -- it's possible for the editor to enter the reviewer's recommendation on their behalf, in case they e.g. email the editor rather than going through the system.

If I understand the issue correctly, it's that editors are mistakenly clicking "Confirm" when they just want to dismiss the modal, correct?

asmecher:
@Devika008, can you think of a quick and easy UX intervention that would help prevent this?

Tribunal33:
Can you remove this `Review details` button if the review hasn't been accepted yet?  That way it would force the editor to log in as the reviewer and manaully accept it?

PR #13176, opened 2026-08-13
pkp/pkp-lib#13117 Allow editors to edit a submitted review

taslangraham:
For pkp/pkp-lib#13117


schemas/reviewAssignment.json
```diff
@@ -334,6 +334,23 @@
...
 			"apiSummary": true
+		},
+		"lastModifiedById": {
+			"type": "integer",
+			"description": "The ID of the user who last made modifications to the submitted review.",
+			"apiSummary": true,
+			"validation": ["nullable"]
+		},
+		"lastModifiedBy": {
+			"type": "object",
```
touhidurabir, line 345: `"type": "object"` defined 2 times here and below .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,202 @@
...
+    public function rules(): array
+    {
+        return [
+            'reviewerRecommendationId' => [
+                'required',
+                'integer',
+                Rule::exists('reviewer_recommendations', 'reviewer_recommendation_id')
+                    ->where(function (Builder $query) {
+                        $query->where('context_id', Application::get()->getRequest()->getContext()->getId());
+                    })
```
touhidurabir: should we consider if the recommendation is enable that is check the status ?

classes/submission/reviewAssignment/Repository.php
```diff
@@ -343,4 +349,140 @@ public function getExportableDOIsPeerReviewIds(int $contextId, bool $doiVersioni
...
+    }
+
+    /**
+     * Save or update a review comment for a review assignment.
+     */
+    public function saveReviewComment(ReviewAssignment $reviewAssignment, string $comments, bool $viewable): SubmissionComment
+    {
+        $submissionCommentDao = DAORegistry::getDAO('SubmissionCommentDAO');
+        /** @var SubmissionCommentDAO $submissionCommentDao */
+        $submissionComments = $submissionCommentDao->getReviewerCommentsByReviewerId($reviewAssignment->getSubmissionId(), $reviewAssignment->getReviewerId(), $reviewAssignment->getId(), true);
```
touhidurabir: instead of considering the `$viewable`, here `true` passed unconditionally to `SubmissionCommentDAO::getReviewerCommentsByReviewerId` .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,202 @@
...
+        ];
+    }
+
+    /** @copydoc FormRequest::validated() */
+    public function validated($key = null, $default = null)
+    {
+        $request = $this->validator->validated();
+        $request['reviewAssignment'] = $this->reviewAssignment;
+
+        return $request;
```
touhidurabir: suggestion : instead of accessing like this, we can do something like 
```php
return array_merge(parent::validated(), ['reviewAssignment' => $this->reviewAssignment])
```
up to you if want to refactor .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+        $reviewAssignment = $validated['reviewAssignment'];
+
+        if (!$this->canAccessReviewAssignment($submission, $reviewAssignment, [Role::ROLE_ID_SUB_EDITOR])) {
+            return response()->json([
+                'error' => __('api.403.unauthorized'),
+            ], Response::HTTP_FORBIDDEN);
+        }
+
+        $submittedReviewerRecommendationId = (int)$illuminateRequest->input('reviewerRecommendationId');
+        $commentsSubmitted = $illuminateRequest->input('comments') ?? '';
```
touhidurabir: we should get the `comments` from the validated and should never use the direct `->input` on laravel request object as that will give us raw request input, not validated one.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+
+        if ($reviewAssignment->getReviewFormId()) {
+            $submittedReviewFormResponses = $illuminateRequest->input('reviewFormResponses') ?? [];
```
touhidurabir: as same above https://github.com/pkp/pkp-lib/pull/13176/changes#r3783531569 for `reviewFormResponses` .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                        sort($existingSorted);
+                        if ($existingSorted == $submittedSorted) {
+                            continue;
+                        }
+                    } elseif ($allExistingFormResponses[$reviewFormElementId] == $reviewFormResponseValue) {
+                        continue;
+                    }
+                }
+
+                if ($reviewFormResponseValue) {
```
touhidurabir: if I understand correctly, this checks for existence but a `0` fail it as the possible response indices can be `0` based . perhaps a better check can be like `$v === null || $v === '' || (is_array($v) && !count($v))` . 

ignore if I am misunderstanding .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                    $oldReviewFormResponses[$reviewFormElementId] = Repo::reviewAssignment()->formatReviewFormElementResponseForLogEntry($reviewFormElement, $reviewFormResponseValue);
+                }
+
+                /** @var ReviewFormResponseDAO $reviewFormResponseDao */
+                $reviewFormResponseDao = DAORegistry::getDAO('ReviewFormResponseDAO');
+                $elementsNotSubmittedId = array_diff(array_keys($allExistingFormResponses), array_keys($submittedReviewFormResponses));
+
+                // If a response for the form element was not submitted, then the existing response for the element will be deleted.
+                // Ensure these are deleted before preparing log value for new form responses so that they are not included in that log.
+                $reviewResponsesToDeleteId = array_merge($reviewResponsesToDeleteId, $elementsNotSubmittedId);
```
touhidurabir: bit confused here as `$formFieldsUpdated` is only set by the submitted keys which continues on unchanged values . So if an editor drops a field and every other submitted value matches what is stored, nothing is deleted. The same payload produces different persisted state depending on unrelated fields . feels to me it contradict what method doc block mentioned as `omitting a non-required field or submitting an empty value will be considered as an indication that existing responses for those fields are to be deleted` . 

ignore is this is correct and I am misunderstanding .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                    }
+                    $updatedReviewFormResponses[$reviewFormElementId] = Repo::reviewAssignment()->formatReviewFormElementResponseForLogEntry($reviewFormElement, $reviewFormResponseValue);
+                }
+
+                // Log the form changes to event log
+                $eventLog = Repo::eventLog()->newDataObject([
+                    'assocType' => PKPApplication::ASSOC_TYPE_REVIEW_ASSIGNMENT,
+                    'assocId' => $reviewAssignment->getId(),
+                    'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_REVIEWER_FORM_RESPONSE_MODIFIED,
+                    'userId' => $this->getRequest()->getUser()->getId(),
```
touhidurabir: need to consider if someone impersonating as another user via logged in feature . also need to add details of impersonated user as `impersonatedUserId`, see https://github.com/pkp/pkp-lib/issues/13059 . 

Here and other places .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                    $reviewAssignment->getId(),
+                    true
+                )->next();
+
+                $oldComments = $existingComment?->getData('comments');
+                if ($commentsSubmitted !== $oldComments) {
+                    $savedComment = Repo::reviewAssignment()->saveReviewComment($reviewAssignment, $commentsSubmitted, true);
+                    // Log changes to the event log
+                    $eventLog = Repo::eventLog()->newDataObject([
+                        'assocType' => PKPApplication::ASSOC_TYPE_SUBMISSION_REVIEW_COMMENT,
```
touhidurabir, line 588: question : is this right usage of `ASSOC_TYPE_SUBMISSION_REVIEW_COMMENT` ? wondering is a single/direct usage of `Repo::eventLog()->getCollector()->filterByAssoc` will be able to retrived all details for UI ? 

Ignore if this has been considered for future scope .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                    'assocType' => PKPApplication::ASSOC_TYPE_REVIEW_ASSIGNMENT,
+                    'assocId' => $reviewAssignment->getId(),
+                    'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_REVIEWER_FORM_RESPONSE_MODIFIED,
+                    'userId' => $this->getRequest()->getUser()->getId(),
+                    'message' => 'submission.event.review.field.modified',
+                    'isTranslated' => false,
+                    'dateLogged' => Core::getCurrentDate(),
+                    'reviewFormResponseOld' => json_encode($oldReviewFormResponses),
+                    'reviewFormResponseNew' => json_encode($updatedReviewFormResponses),
+                    'fieldName' => __('submission.event.review.fieldName.reviewFormResponses'),
```
touhidurabir: question : should it be translated or should store the key only so at render time it get translated as per the selected locale?

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    /**
+     * The validation rules.
+     */
+    public function rules(): array
+    {
+        return [
+            'reviewerRecommendationId' => [
```
touhidurabir, line 69: question : is this only for OJS as OMP does not have it, so should we check it via `Application::hasCustomizableReviewerRecommendation()` ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+        $reviewAssignmentId = (int)$illuminateRequest->route('reviewAssignmentId');
+        $reviewAssignment = Repo::reviewAssignment()->get($reviewAssignmentId, $submission->getId());
+
+        if (!$reviewAssignment) {
+            return response()->json([
+                'error' => __('api.404.resourceNotFound'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        if (!$this->canAccessReviewAssignment($submission, $reviewAssignment, [Role::ROLE_ID_SUB_EDITOR, Role::ROLE_ID_REVIEWER, Role::ROLE_ID_ASSISTANT])) {
```
touhidurabir: here we are including the `ROLE_ID_REVIEWER` but the route group does not include it, does it not make it dead role/path ?

classes/submission/reviewAssignment/Repository.php
```diff
@@ -343,4 +349,140 @@ public function getExportableDOIsPeerReviewIds(int $contextId, bool $doiVersioni
...
+
+        $comment->setCommentType(SubmissionComment::COMMENT_TYPE_PEER_REVIEW);
+        $comment->setRoleId(Role::ROLE_ID_REVIEWER);
+        $comment->setAssocId($reviewAssignment->getId());
+        $comment->setSubmissionId($reviewAssignment->getSubmissionId());
+        $comment->setAuthorId($reviewAssignment->getReviewerId());
+        $comment->setComments($comments);
+        $comment->setCommentTitle('');
+        $comment->setViewable($viewable);
+        $comment->setDatePosted(Core::getCurrentDate());
```
touhidurabir: updating the date posted on every edit, should it only at first create and later date modified ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+            /** @var ReviewFormResponseDAO $reviewFormResponseDao */
+            $reviewFormResponseDao = DAORegistry::getDAO('ReviewFormResponseDAO');
+
+            // Tracks whether any form responses changed as a result of this update
+            $formFieldsUpdated = false;
+            $allExistingFormResponses = $reviewFormResponseDao->getReviewReviewFormResponseValues($reviewAssignment->getId());
+
+            $reviewResponsesToDeleteId = [];
+            foreach ($submittedReviewFormResponses as $reviewFormElementId => $reviewFormResponseValue) {
+                $hasExistingResponse = key_exists($reviewFormElementId, $allExistingFormResponses);
```
touhidurabir: should use `array_key_exists` .

locale/en/api.po
```diff
@@ -508,3 +508,21 @@ msgstr "The review assignment has already been considered and cannot be consider
 
 msgid "api.reviews.assignments.422.missingOrInvalidConsideredValue"
 msgstr "The `considered` field is missing or contains an invalid value."
+
+msgid "api.submissions.reviews.422.invalidRecommendation"
```
touhidurabir, line 512: this one seems never used , should it be used with validation rule ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+        }
+
+        $newAssignmentData = [
+            'reviewerRecommendationId' => $submittedReviewerRecommendationId,
+            'lastModifiedById' => $this->getRequest()->getUser()->getId(),
+        ];
+
+        if (!$reviewAssignment->getDateCompleted()) {
+            $newAssignmentData['dateCompleted'] = Core::getCurrentDate();
+        }
```
touhidurabir: I understand this is as per the issue mentioned that `if the review is changed by editor for the first time (dateCompleted is null) - than consider it as review was completed by that step and fill the dateCompleted with current time` --> but that is my biggest question --> why ? is it not possible to update without completing it ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -363,8 +387,237 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+     *
+     * This updates the review's comments or form responses, along with the reviewer's recommendation.
+     * All required review fields must be submitted. That is:
+     * - `reviewerRecommendationId` must be present
+     * - If a review has a form, then form fields marked as 'required' must be submitted in the `reviewFormResponses` field.
+     *       The submitted form fields in `reviewFormResponses` will completely overwrite existing values for those fields.
+     *       Therefore, omitting a non-required field or submitting an empty value will be considered as an indication that existing responses for those fields are to be deleted.
+     *       If the review does not have a form, then comments are allowed. Currently, comment fields are optional even when a form is not present.
+     */
+    public function editReview(EditReview $illuminateRequest): JsonResponse
```
touhidurabir, line 434: Few general question : 
 - do we need any check for assignment when it's declined or canceled ?
 - if I am not mistaken we unconditionally update the review assignment with updated `lastModifiedById` even if there is nothing to update, is that correct ?

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,202 @@
...
+                    if (!is_array($submittedReviewFormResponses)) {
+                        // Parse the submitted responses from JSON into an associative array keyed by review form element ID.
+                        $submittedReviewFormResponses = json_decode($submittedReviewFormResponses, true);
+                    }
+
+                    /** @var ReviewFormElementDAO $reviewFormElementDao */
+                    $reviewFormElementDao = DAORegistry::getDAO('ReviewFormElementDAO');
+                    $reviewFormElements = $reviewFormElementDao->getByReviewFormId($this->reviewAssignment->getReviewFormId());
+
+                    while ($reviewFormElement = $reviewFormElements->next()) {
```
touhidurabir, line 145: if I understand correctly, the 2 check validate if that if for each element in the form, is the submitted value acceptable. but does not if the form element is part of this review form . 

then the controller loop over it and try to write as `ReviewAssignmentController::editReview[$submittedReviewFormResponses as $reviewFormElementId  => $reviewFormResponseValue]` --> Repository::saveReviewFormResponse() --> ReviewFormElementDAO::getById() which can cause as 
 - non exists in DB, internal error/exception
 - exists but not related to this form, slient write to another's related . 
 
ignore if I am misunderstanding .

touhidurabir (changes requested):
@taslangraham see below questions/suggestions and few needed changes .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
taslangraham: You're right; there should also be a check to ensure that the element is related to this assignment's form. Thanks for catching that.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: That was accidentally left in.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: The latter.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: I introduced `ASSOC_TYPE_SUBMISSION_REVIEW_COMMENT` specifically to represent a log entry that is for a comment. As things are, a single `Repo::eventLog()->getCollector()->filterByAssoc` would return all that's needed. And if we require more granular inspection on the log entry when preparing for display, I think additional filtering using `eventType` should suffice.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: Maybe @jardakotesovec or @asmecher can explain why that approach was chosen?

locale/en/api.po
taslangraham: Should be used as a custom validation message in `EditReview`. Thanks for catching that.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
asmecher: @touhidurabir and @touhidurabir This is the action triggered in the past by the "Confirm" button on a peer review, right?
image
Previously, this button served several purposes:
1. To allow the Editor to confirm acceptance and record the reviewer's recommendation, if the reviewer e.g. emailed it directly to the editor
2. To flag the review as "considered"

This has been a source of occasional confusion (e.g. https://github.com/pkp/pkp-lib/issues/6353), and now that we're allowing the review to be edited, it makes sense to me to separate/clarify these functions.

taslangraham: @asmecher, we are separating the various functions of the button mentioned above. 

The endpoints introduced in https://github.com/pkp/pkp-lib/issues/13116 handle the changing of the considered (considered, viewed, unconsidered) status of a review. Marking a review as "considered" here is the same as confirming the review. The "confirm" button will now also be separated for the screen where the editor can make changes to the reviewer recommendation, etc.


This current PR is allowing the editor to record the reviewer's recommendation along with any comments or answers to the review form. The question here that @touhidurabir is raising, though, is whether an editor submitting a reviewer's recommendation along with their comments/form responses for a review that is yet to be completed should be interpreted as the review being "completed". I think yes, it should be because currently, a review is automatically marked as completed once a reviewer submits a response(recommendation along with any comments or answers to the review form) themself. If the editor is submitting these on behalf of the reviewer, then they are essentially "completing" the review for the reviewer.

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
taslangraham: Updated to require `reviewerRecommendationId ` and perform the associated checks only when the application supports customizable reviewer recommendations. If the `reviewerRecommendationId` is present in the request and the app does not support customizable reviewer recommendations, a validation error will be thrown.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: > do we need any check for assignment when it's declined or canceled ?

Yes, we should check for those. I've updated the validation checks to ensure that a review is not cancelled or declined. 


> if I am not mistaken we unconditionally update the review assignment with updated lastModifiedById even if there is nothing to update, is that correct ?

I've made changes to ensure that the `lastModifiedById` and other review assignment fields are updated only when an update is made.

taslangraham:
Hi @touhidurabir, thanks for the review. I've made updates(commit: https://github.com/pkp/pkp-lib/pull/13176/changes/af7d20e9755cf97dae32b43c7042f33a2d341f1f) to fix all the concerns/issues you highlighted in your review. Please take a look. THanks!

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -364,8 +390,261 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                        }
+                    } elseif ($allExistingFormResponses[$reviewFormElementId] == $reviewFormResponseValue) {
+                        continue;
+                    }
+                }
+
+                if (
+                    $reviewFormResponseValue !== null &&
+                    $reviewFormResponseValue !== '' &&
+                    (is_array($reviewFormResponseValue) && count($reviewFormResponseValue) !== 0)
```
touhidurabir: should it be `||` instead of `&&` as `is_array($v)` is `false` for every scalar, so the
 whole condition is false and control falls into the delete branch. perhaps it should be `$v !== null && $v !== '' && (!is_array($v) || count($v) !== 0)` ?

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+                    }
+
+                    if ($this->input('comments')) {
+                        $fail(__('api.submissions.reviews.422.commentsNotAllowed'));
+                    }
+                },
+            ],
+            // Only allow `reviewFormResponses` if the review has a form
+            'reviewFormResponses' => [
+                'sometimes',
```
touhidurabir: bit concern about it using the `sometimes`. using `bail` good as it will not other validation rules if not present. so if none present, by default it becoming `[]` as per the default from `validated` . so then in the controller, as per the `$elementsNotSubmittedId = array_diff(array_keys(...`, it will be diffing against `[]`, wiping out the previously saved from response . 

so the point is if the API only allow to pass non-form data like comments and if from data being set previously, then will be wiped out . 

Ignore if I am misunderstanding

locale/en/submission.po
```diff
@@ -3931,3 +3931,15 @@ msgstr "The following issues were found, but will not prevent publishing"
 
 msgid "reviewer.competingInterests"
 msgstr "Competing Interests"
+
+msgid "submission.event.review.field.modified"
+msgstr "The following field was modified in this review: {$fieldName}. Select \"View changes\" to see a detailed summary of all modifications."
```
touhidurabir: I see that in event log store, this moved to `fieldName` --> `fieldNameKey` but not here.

locale/en/submission.po
```diff
@@ -3931,3 +3931,15 @@ msgstr "The following issues were found, but will not prevent publishing"
...
+
+msgid "submission.event.review.fieldName.comments"
+msgstr "comments"
+
+msgid "submission.event.review.fieldName.reviewerRecommendationId"
+msgstr "reviewerRecommendationId"
+
+msgid "submission.event.review.fieldName.reviewFormResponses"
+msgstr "reviewFormResponses"
```
touhidurabir: if I understand correctly, these are user facing translated information, so should these be `"Comments", "Recommendation", "Review form"` ?

classes/submission/reviewAssignment/Repository.php
```diff
@@ -343,4 +349,147 @@ public function getExportableDOIsPeerReviewIds(int $contextId, bool $doiVersioni
...
+        $comment->setAuthorId($reviewAssignment->getReviewerId());
+        $comment->setComments($comments);
+        $comment->setCommentTitle('');
+        $comment->setViewable($viewable);
+
+        $commentId = $comment->getId();
+        // Save or update
+        if ($comment->getId() != null) {
+            $submissionCommentDao->updateObject($comment);
+            $comment->setDateModified(Core::getCurrentDate());
```
touhidurabir: I think these 2 lines need to swap position as we are updating date modified after data object update which will not persist the changes .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+
+                        if (
+                            isset($submittedReviewFormResponses[$reviewFormElementId])
+                            && !$this->isValidReviewFormFieldResponse($reviewFormElement, $submittedReviewFormResponses[$reviewFormElementId])
+                        ) {
+                            $fail(__('api.submissions.reviews.422.invalidReviewFormResponse', ['elementId' => $reviewFormElementId]));
+                        }
+                    }
+
+                    foreach (array_keys($submittedReviewFormResponses) as $submittedReviewElementId) {
```
touhidurabir, line 165: does it handle scaler json like `{ "reviewFormResponses": "5" }` as `array_keys` strictly require first params to be array otherwise internal exception .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -364,8 +390,261 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+                    'reviewFormResponseOld' => json_encode($oldReviewFormResponses),
+                    'reviewFormResponseNew' => json_encode($updatedReviewFormResponses),
+                    'fieldNameKey' => 'submission.event.review.fieldName.reviewFormResponses',
+                    'impersonatedUserId' => Validation::loggedInAs() ? $user->getId() : null,
+                ]);
+                Repo::eventLog()->add($eventLog);
+            }
+        } else {
+            // if comments were submitted, update the review comments with the new values
+            if ($commentsSubmitted) {
```
touhidurabir: what if `$commentsSubmitted` is empty string when it's not required and at update, editor want to clear out previous wrong comment to nothing ?

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+    public function validated($key = null, $default = null)
+    {
+        return array_merge(
+            parent::validated(),
+            [
+                'reviewAssignment' => $this->reviewAssignment,
+                'comments' => $this->input('comments'),
+                'reviewFormResponses' => $this->input('reviewFormResponses') ?? [],
+                // Will default to null in cases where the app does not support reviewer recommendations (e.g., in OMP).
+                'reviewerRecommendationId' => (int)$this->input('reviewerRecommendationId') ?? null,
```
touhidurabir: this would never have null due to int cast and may slip a `0` (for OMP) .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * The validation rules.
+     */
+    public function rules(): array
+    {
+        return [
+            'reviewerRecommendationId' => [
+                // Only allow `reviewerRecommendationId` in the request if the application supports having reviewer recommendations.
+                Rule::when(
+                    Application::get()->hasCustomizableReviewerRecommendation(),
```
touhidurabir, line 72: this is correct but need to update the OMP and OPS main for return value `false` which I have mistakenly put `true` . This is Important .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+                            isset($submittedReviewFormResponses[$reviewFormElementId])
+                            && !$this->isValidReviewFormFieldResponse($reviewFormElement, $submittedReviewFormResponses[$reviewFormElementId])
+                        ) {
+                            $fail(__('api.submissions.reviews.422.invalidReviewFormResponse', ['elementId' => $reviewFormElementId]));
+                        }
+                    }
+
+                    foreach (array_keys($submittedReviewFormResponses) as $submittedReviewElementId) {
+                        if (!in_array($submittedReviewElementId, $reviewAssignmentFormElementIds)) {
+                            $fail(__('api.submissions.reviews.422.invalidReviewFormElementSubmitted'));
```
touhidurabir: the `{$elementId}` params is missing .

classes/submission/reviewAssignment/Repository.php
```diff
@@ -343,4 +349,147 @@ public function getExportableDOIsPeerReviewIds(int $contextId, bool $doiVersioni
...
+        $reviewFormResponseDao = DAORegistry::getDAO('ReviewFormResponseDAO');
+
+        $reviewFormResponse = $reviewFormResponseDao->getReviewFormResponse($reviewAssignment->getId(), $reviewFormElementId);
+        if (!isset($reviewFormResponse)) {
+            $reviewFormResponse = new ReviewFormResponse();
+        }
+
+        /** @var ReviewFormElementDAO $reviewFormElementDao */
+        $reviewFormElementDao = DAORegistry::getDAO('ReviewFormElementDAO');
+        $reviewFormElement = $reviewFormElementDao->getById($reviewFormElementId);
```
touhidurabir, line 371: `$reviewFormElement` can be null, right ? so do we need a null check before proceed next ?

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
```diff
@@ -0,0 +1,239 @@
...
+                // Only allow `reviewerRecommendationId` in the request if the application supports having reviewer recommendations.
+                Rule::when(
+                    Application::get()->hasCustomizableReviewerRecommendation(),
+                    [
+                        'required',
+                        'integer',
+                        Rule::exists('reviewer_recommendations', 'reviewer_recommendation_id')
+                            ->where(function (Builder $query) {
+                                $query->where('context_id', Application::get()->getRequest()->getContext()->getId())
+                                    ->where('status', RecommendationOption::ACTIVE->criteria());
```
touhidurabir: this is correct but what if by the next update, the previously selected recommendation had been disabled ? and Editor/JM only updating comment or form . I think we need to consider it . see similar logic in `Repo::reviewerRecommendation()->getRecommendationOptions` .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -364,8 +390,261 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+        $newAssignmentData = [];
+        $oldReviewerRecommendationId = $reviewAssignment->getReviewerRecommendationId();
+
+        if ($submittedReviewerRecommendationId !== $oldReviewerRecommendationId) {
+            $newAssignmentData['reviewerRecommendationId'] = $submittedReviewerRecommendationId;
+            $isReviewUpdated = true;
+        }
+
+        if ($isReviewUpdated) {
+            $newAssignmentData['lastModifiedById'] = validation::loggedInAs() ?? $user->getId();
```
touhidurabir: Validation, capital `V` .

schemas/eventLog.json
```diff
@@ -345,6 +345,54 @@
...
+			]
+		},
+		"reviewerCommentsOld": {
+			"type": "string",
+			"description": "The previous comments of the reviewer, if the reviewer comments were changed.",
+			"validation": [
+				"nullable"
+			]
+		},
+		"reviewerCommentsNew": {
```
touhidurabir, line 372: this one does not have a validation rule, does it means it's always required ?

api/v1/submissions/reviewAssignments/resources/ReviewResource.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Map a review form element to the matching form field component.
+     */
+    protected function mapReviewFormElementToFormField(ReviewFormElement $reviewFormElement): Field
+    {
+        $name = (string) $reviewFormElement->getId();
+        $args = [
+            'label' => PKPString::html2text($reviewFormElement->getLocalizedQuestion()),
+            'isRequired' => $reviewFormElement->getRequired(),
+            'description' => PKPString::html2text($reviewFormElement->getLocalizedDescription() ?: ''),
```
touhidurabir: I understand running label via `PKPString::html2text` but it will flatten the description if passed in rich text, is that intentional ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -364,8 +390,261 @@ protected function canAccessReviewAssignment(Submission $submission, ReviewAssig
...
+            if (!is_array($submittedReviewFormResponses)) {
+                // Parse the submitted review form responses from JSON string to an associative array.
+                // Each key is a review form element ID, and the value is the submitted response.
+                $submittedReviewFormResponses = json_decode($submittedReviewFormResponses, true);
+            }
+
+            /** @var ReviewFormElementDAO $reviewFormElementDao */
+            $reviewFormElementDao = DAORegistry::getDAO('ReviewFormElementDAO');
+            /** @var ReviewFormResponseDAO $reviewFormResponseDao */
+            $reviewFormResponseDao = DAORegistry::getDAO('ReviewFormResponseDAO');
```
touhidurabir, line 464: this has been resolved here and below, 2 times , I think only once is ok . also `getById` being called multiple time which is too many DB calls, can we optimise it if possible . ignore if this is not important .

schemas/reviewAssignment.json
```diff
@@ -334,6 +334,22 @@
...
+			"description": "The user who last made modifications to the submitted review.",
+			"apiSummary": true,
```
touhidurabir, line 347: this true means it will always be resolved for dashboard but we already have `lastModifiedById` as `"apiSummary": true,` so do we need it to be true as well ?

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
touhidurabir: so I have another general question/concern here, if the review assignment completed by a Editor/JM, can the assigned reviewer still edit/update it ? I think not due to added route level role/permission guard so we are good here . 

But how the notification being handled . Say reviewer got the assignment notification but never respond to it .And Editor complete it. Then the reviewer got back to it which is now slate notification , should we make sure we remove such notifications at completed by Editor/JM/Admin ?

touhidurabir (changes requested):
@taslangraham second round review added . A reminder to set the return value of `Application::hasCustomizableReviewerRecommendation` to `false` for OMP and OPS when it's merged .

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
touhidurabir: Thanks for the explanation @asmecher and @taslangraham , this make sense now .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
taslangraham: Use of `sometimes` should, in fact, be omitted. For the other point you made, the issue here is that the `validated` defaults the missing `reviewFormResponses` to `[]`. It should instead default to `null`.

api/v1/submissions/reviewAssignments/resources/ReviewResource.php
taslangraham: Should be removed.

api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
taslangraham: This was an oversight on my part. The notification should be deleted, as that is what currently happens when an editor confirms a review(and submits the recommendation), which marks it as completed, before the reviewer submits their review.

api/v1/submissions/reviewAssignments/resources/ReviewResource.php
touhidurabir: > Should be removed.

ok, but perhaps it should run sanitization ? or sanitization will always run at client side rendering (I think that is the case) ? but be sure of whatever path you take .

api/v1/submissions/reviewAssignments/formRequests/EditReview.php
taslangraham: The check(Lines 114 to 125) in the [form validator](https://github.com/pkp/pkp-lib/pull/13176/changes#diff-20766800f237fc9ca01d8391669cf7ea2a072e95d07e5c865fb68776b1bb3905) would reject such a value:

```php
function (string $attribute, mixed $value, Closure $fail) {
    if (is_array($value)) {
        return;
    }

    if (is_string($value)) {
        $decodedValue = json_decode($value, true);
        if (json_last_error() === JSON_ERROR_NONE && is_array($decodedValue)) {
            return;
        }
    $fail(__('api.submissions.reviews.422.invalidReviewFormSubmitted'));
    }
```

PR #13179, opened 2026-08-14
pkp/pkp-lib#13003 Add support for batch loading

asmecher:
Add support for batch loading using `getMany` to resolve N+1 query issues.

Strategy:
- Built support for `fromRow` to have a cache that persists for the duration of a batch load (e.g. a fetch via `getMany` )
- Centralized `getMany` into `EntityDAO` and removed it from all the subclasses (except for the user DAO, where there's a slight extension to conditionally fetch reviewer data based on the collector configuration)
- ~~Add support in the `fromRow` function for callers to (optionally) specify a "populator" callable, which is used to specify different strategies for populating an object:~~
  - ~~Entities can extend `fromRow` (as always) to add elements for any strategy (e.g. the submission locale in `lib/pkp/classes/publication/DAO.php`)~~
  - ~~Entities can extend `individualPopulator` (the default) to add elements that will be executed for each fetch (as before)~~
  - ~~Entities can implement and specify a closure (e.g. in `getMany` in the examples in this PR) allowing a set to be loaded once per `getMany` call and used as needed by `fromRow`.~~
- Minor optimization: remove extra query to get submission locale when fetching publication https://github.com/pkp/pkp-lib/pull/13179/changes/a4ceac020eb736306e454c158b5e04491f2cfe49
- Generalize a few fetch functions that previously only accepted a single ID to accept an array instead (to avoid having to double up the getter functions). Example: `filterByAuthorIds` in `classes/affiliation/Collector.php`
- Removed some unnecessary code duplication in `classes/log/event/DAO.php`
- Some rewriting of `getQueryBuilder` in `classes/ror/Collector.php` and `classes/citation/Collector.php` to use Conditionables

### DIFF
--- a/api/v1/dataCitations/PKPDataCitationController.php
+++ b/api/v1/dataCitations/PKPDataCitationController.php
@@ -12,7 +12,7 @@
  * @ingroup api_v1_data_citations
  *
  * @brief Controller class to handle API requests for data citation operations.
- * 
+ *
  */
 
 namespace pkp\api\v1\dataCitations;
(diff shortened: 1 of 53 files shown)


asmecher:
Merged manually.

Issue #13192, opened 2026-08-18
[Submission] | Activity Log - Display the activity logs for changes made to reviews by editors


taslangraham:
### Description

The API endpoints introduced in [#13117](https://github.com/pkp/pkp-lib/issues/13117) allow editors to modify reviews, including:

* Changing the reviewer recommendation
* Adding or updating reviewer comments
* Adding or updating review form responses

Each modification to a review is logged. The log records both the **previous value** and the **new value** for each change.

This ticket is to display these logged review changes in the submission's **Activity Log**, allowing editors to see what was changed, including the previous and updated values.

### Steps to test changes

pre-requisites: [UI portion for editable reviews](https://github.com/pkp/pkp-lib/issues/13156) should be in place.

- Log in as an editor
- Select a submission that is in the review stage
- Assign reviewers to the submission
- As Editor, use the UI mentioned above to submit changes to the review (i.e, submit comments, change reviewer recommendation, or submit form responses)
- Click `Activity Log`
- Observe the log for the change made. Expand the log and click `View changes` to see log details
- A side modal should open with details about the log

### Examples of activity log entries
1. **Activity log for change in reviewer recommendation**

<img width="1140" height="138" alt="Image" src="https://github.com/user-attachments/assets/fe71c644-f0ef-4c69-bf25-2496a1b68082" />

<img width="1133" height="246" alt="Image" src="https://github.com/user-attachments/assets/67ab58be-147a-4db6-8f4f-9e551bc36147" />

2. **Activity log for change in review comments**

<img width="1113" height="97" alt="Image" src="https://github.com/user-attachments/assets/abba46ee-9b5f-4214-89e9-e6ac5db00799" />

<img width="1140" height="318" alt="Image" src="https://github.com/user-attachments/assets/8c3bf260-65ba-43b0-aec9-ff7163bfe938" />

3. **Activity for change in review form responses**

<img width="1170" height="179" alt="Image" src="https://github.com/user-attachments/assets/7883174b-c8f6-486a-9915-913a072c5dae" />

<img width="1895" height="1266" alt="Image" src="https://github.com/user-attachments/assets/2a7d0c93-2292-492b-ac91-06506599551d" />

**Pull Request**
pkp-lib: https://github.com/pkp/pkp-lib/pull/13197
osj (submodule update): https://github.com/pkp/ojs/pull/5749

taslangraham:
[...]

PR #13197, opened 2026-08-19
pkp/pkp-lib#13192 Display review logs in submission activity logs

taslangraham:
For pkp/pkp-lib#13192

### DIFF
--- a/api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
+++ b/api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
@@ -550,17 +550,18 @@ class ReviewAssignmentController extends PKPBaseController
                 //     },
                 //     ...
                 // }
+                ksort($oldReviewFormResponses, SORT_NUMERIC);
+                ksort($updatedReviewFormResponses, SORT_NUMERIC);
                 $eventLog = Repo::eventLog()->newDataObject([
                     'assocType' => PKPApplication::ASSOC_TYPE_REVIEW_ASSIGNMENT,
                     'assocId' => $reviewAssignment->getId(),
                     'eventType' => PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_REVIEWER_FORM_RESPONSE_MODIFIED,
                     'userId' => Validation::loggedInAs() ?? $user->getId(),
-                    'message' => 'submission.event.review.field.modified',
+                    'message' => 'submission.event.review.field.modified.comments.formResponse',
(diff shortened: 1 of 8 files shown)


api/v1/submissions/reviewAssignments/ReviewAssignmentController.php
```diff
@@ -464,6 +464,12 @@ public function editReview(EditReview $illuminateRequest): JsonResponse
             // Tracks whether any form responses changed as a result of this update
             $formFieldsUpdated = false;
             $allExistingFormResponses = $reviewFormResponseDao->getReviewReviewFormResponseValues($reviewAssignment->getId());
+            $allReviewFormElementsFactory = $reviewFormElementDao->getByReviewFormId($reviewAssignment->getReviewFormId());
```
ewhanson, line 467: Not a huge deal, but `DAOResultFactory` does have a helper `toAssociativeArray()` that does the same thing you've done below.

controllers/grid/eventLog/SubmissionEventLogGridHandler.php
```diff
@@ -264,4 +270,61 @@ public function _formatEmail(EmailLogEntry $emailLogEntry)
...
+            ->filterByEventType(PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_REVIEWER_FORM_RESPONSE_MODIFIED)
+            ->getMany() : [];
+
+        $recommendationLogEntries = $reviewsWithCommentsIds ? Repo::eventLog()->getCollector()
+            ->filterByAssoc(PKPApplication::ASSOC_TYPE_REVIEW_ASSIGNMENT, array_merge($reviewsWithCommentsIds, $reviewsWithFormIds))
+            ->filterByEventType(PKPSubmissionEventLogEntry::SUBMISSION_LOG_REVIEW_REVIEWER_RECOMMENDATION_MODIFIED)
+            ->getMany() : [];
+
+        return array_merge(
+            $commentLogEntries->toArray(),
```
ewhanson: These three will already be empty arrays per the ternary operators above, so maybe it would be better to do the `toArray()` part tacked onto the end of the `getMany()` so you don't have to worry about trying to call `toArray()` on an array.

ewhanson (changes requested):
Thanks @taslangraham, just a few minor changes. I didn't fully test it with the other, unmerged changes, but from what I saw, it was working as expected.

taslangraham:
Thanks for the review, @ewhanson. I've made the suggested changes and merged.

PR #13206, opened 2026-08-21
pkp/dev-team#322 [Edtiorial UI] | Review UI - "Author Response" button leads to page without an email form

blesildaramirez:
Request Author Response fails to load when a journal's `supportedLocales` is stored as a JSON object (`{"1":"en"}`) instead of an array. The composer gets locales as an object, so `this.locales.filter()` throws an error.

The reason why `supportedLocales` can sometimes be stored as a JSON object instead of an array of strings is because of disabling/removing a locale site-wide that is still used by a journal. Removing a locale at the site level filters the remaining locales using `array_diff()` / `array_intersect()` / `unset()`, which preserve the keys. That's why the indexes end up being stored as keys in a JSON object.

Using `array_values()` around the values will reset the indexes and ensure it's stored as an array of strings.

### DIFF
--- a/classes/components/RequestReviewResponsePage.php
+++ b/classes/components/RequestReviewResponsePage.php
@@ -180,13 +180,14 @@ class RequestReviewResponsePage
      */
     private function getLocales(): array
     {
-        return array_map(
+        // Re-index so the locales encode as a JSON array, not an object
+        return array_values(array_map(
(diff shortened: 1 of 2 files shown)


ewhanson (approved):
Looks, good. Thanks @blesildaramirez!

Issue #13205, opened 2026-08-20
Consolidate peer review data handling architecture


ewhanson:
<details><summary>Original notes on peer review data optimization (2026-08-24)</summary>
<p>
From @asmecher:

> The article view is running a lot of queries on peer review data, and it looks like it's doubled because of OpenReviewComponent sequentially invoking SubmissionPeerReviewResource and SubmissionPeerReviewSummaryResource, each of which fetches everything from the DB in what looks like identical ways.

Benchmark (as of 2026-08-21):

Test data included sample with two complete review rounds with a third one initiated (three reviewers per round) and three publications.

**Number of queries**:
| Item | Before | After |
|--------|--------|--------|
| `SubmissionPeerReviewDataCache` | _ | 71 |
| `SubmissionPeerReviewResource` | 73 | 44 |
| `SubmissionPeerReviewSummaryResource` | 16 | 0 |
| **Total** | 100 | 71 |
</p>
</details> 

## PRs:

- `pkp-lib`: pkp/pkp-lib#13210
- `ojs`: pkp/ojs#5758

ewhanson:
Hey @asmecher, could you have a look at these and give it a test with the dataset you've been working with. It removes the duplicate calls for the same data, but if we need to squeeze out even more performance, we probably could, but it won't be quite as quick of a win. Let me know!

asmecher:
Thanks, @ewhanson! I've added a review [with a single specific suggestion here](https://github.com/pkp/pkp-lib/pull/13210#pullrequestreview-5013291996). More broadly, 71 queries is still a lot, and at the same time a savings of 29 is not all that much. I suspect we need to apply some batch loading tools here as well, but the fact of having a caching class heirarchy that's different from our usual data model loading stuff is probably going to make that more complicated. I'll follow up via MM with some options.

ewhanson:
Thanks @asmecher! As we discussed offline, I'm going to co-opt this issue for the rejigging of the peer review data handling.

asmecher:
@ewhanson, not a huge priority, but just a reminder that I think there are some open next steps here.

ewhanson:
Yes, thanks @asmecher! 

Been plugging away at this and also tested things as they were here with the `_fromRow()` improvements, and it did reduce the number of queries, but still nothing massive. Hope to have the consolidation work done soon.

PR #13210, opened 2026-08-21
pkp/pkp-lib#13205 Consolidate peer review data handling architecture

ewhanson:


api/v1/peerReviews/resources/SubmissionPeerReviewDataCache.php
```diff
@@ -0,0 +1,101 @@
...
+    /** @var Collection<int, ReviewAssignment> $publicAcceptedReviewAssignments  */
+    public Collection $publicAcceptedReviewAssignments;
+
+    /** @var Collection<int, ReviewerRecommendation> $availableRecommendationTypes  */
+    public Collection $availableRecommendationTypes;
+
+    public function __construct(
+        private Submission $submission
+    ) {
+        $this->context = Application::getContextDAO()->getById($this->submission->getData('contextId'));
```
asmecher, line 47: I wonder if this could be passed down the chain from the calling code. We've got dozens of context fetches from the DB, and while it's not a lot of weight compared to a more complex object, it does add up.

asmecher (commented):
One small specific suggestion!

Issue #13213, opened 2026-08-23
[Editorial Tasks] | Foreign key constraint for templates


Vitaliy-1:
**Description:**
`edit_task` table has the column `edit_task_template_id, which should reference to the primary key of the `edit_task_templates` table

**Solution:** What the proposed solutions are and additional affected areas
Add foreign key constraint to the installation and migration script

**What application are you using?**
  OJS, OMP and OPS main branch

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/13214
OJS: https://github.com/pkp/ojs/pull/5759
OMP: https://github.com/pkp/omp/pull/2457
OPS: https://github.com/pkp/ops/pull/1396

Vitaliy-1:
@touhidurabir, can you review this small PR? If it looks good, I'll port it to OMP and OPS

touhidurabir:
@Vitaliy-1 added review at https://github.com/pkp/pkp-lib/pull/13214#pullrequestreview-5037586469 . Once you consider those, it's good to port to OMP/OPS and ready to merge . Thanks .

Vitaliy-1:
PRs
pkp-lib: https://github.com/pkp/pkp-lib/pull/13357
tests: https://github.com/pkp/ojs/pull/5840

Vitaliy-1:
@taslangraham, can you you review this small PR?

taslangraham:
Approved @Vitaliy-1

PR #13214, opened 2026-08-23
pkp/pkp-lib#13213 Add foreign key constraint

Vitaliy-1:


classes/migration/install/SubmissionsMigration.php
touhidurabir, line 400: By adding a FK, this will cause issue at migration down as child is dropped later . need to move the drop of `edit_task` first , e.g. need to move `Schema::drop('edit_task_templates')` to after `Schema::drop('edit_tasks')`

classes/migration/install/SubmissionsMigration.php
```diff
@@ -347,6 +347,13 @@ public function up(): void
             $table->timestamps();
         });
 
+        Schema::table('edit_tasks', function (Blueprint $table) {
+            $table->foreign('edit_task_template_id')
```
touhidurabir, line 351: suggestion : perhaps better to add a given FK constrain name than auto generated one from laravel which make it easy in `I13213`'s down with the usage of `$this->hasForeignKey` , easier to use the named reference than auto generated one .

classes/editorialTask/EditorialTask.php
```diff
@@ -189,6 +191,14 @@ public function participants(): HasMany
         return $this->hasMany(Participant::class, 'edit_task_id', 'edit_task_id');
     }
 
+    /**
+     * Relationship to source task template if task was auto-created from one.
+     */
+    public function template(): BelongsTo
+    {
+        return $this->belongsTo(Template::class, 'edit_task_template_id', 'edit_task_template_id');
+    }
```
touhidurabir, line 200: Suggestion : as we add this relation, better to add the inverse relation in the Template model as well like 
```php
public function tasks(): HasMany
 {
     return $this->hasMany(EditorialTask::class, 'edit_task_template_id', 'edit_task_template_id');
 }
```
then we can use the relational load at both end like in the `\PKP\editorialTask\Repository::taskAlreadyCreatedFromTemplate()` or use it directly to optimise query in the `\PKP\editorialTask\Repository::autoCreateFromTemplates()` like 
```php
$templates = Template::query()
         ->withContextId((int) $submission->getData('contextId'))
         ->withStageId($stageId)
         ->withInclude(true)
         ->whereDoesntHave('tasks', fn (Builder $query) => $query
             ->withAssoc(PKPApplication::ASSOC_TYPE_SUBMISSION, $submission->getId()))
         ->get();
```
and drop the `\PKP\editorialTask\Repository::taskAlreadyCreatedFromTemplate()` .

touhidurabir (changes requested):
@Vitaliy-1 added few comments , mostly suggestions so up to you if you want to adapt but the migration down ordering seems like a needed one .

PR #13215, opened 2026-08-23
Fix insecure file extension/MIME validation allowing disguised executable uploads

BayuFirmansyaah:
## 🔴 CRITICAL SECURITY FIX — Arbitrary File Upload / Remote Code Execution

**Severity: Critical.** This vulnerability is not theoretical — it has already been
exploited against real, production OJS installations. **Multiple universities,
including several in Indonesia, have had their journal websites compromised
through this exact weakness and hijacked to host online gambling ("judi
online" / "judol") spam pages.** These gambling pages get indexed by search
engines under the institution's own domain, causing serious reputational and
SEO damage that is difficult to reverse.

### The problem

`FileManager::parseFileExtension()` in `classes/file/FileManager.php` is
responsible for deciding what file extension an uploaded file is allowed to
keep. For years it has carried the comment `// FIXME Check for evil`, and its
actual check was this:

```php
if (!isset($fileExtension) || stristr($fileExtension, 'php') || strlen($fileExtension) > 6 || !preg_match('/^\w+$/', $fileExtension)) {
    $fileExtension = 'txt';
}

In plain terms, this only blocks a file extension if it contains the letters
"php" somewhere in it. That is easy to get around:

- .phtml, .pht, .phar, .phps — all still executed as PHP by many
  default web server configurations, and none of them contain "php" as a
  substring, so they were never blocked.
- Other server-executable extensions such as .cgi, .pl, .py, .sh,
  .jsp, .asp, .htaccess were not checked at all.
- Most importantly: the check never looks at the actual content of the
  uploaded file. It only ever looks at the filename the uploader typed in.
  A PHP webshell renamed to something harmless-looking like photo.jpg or
  cover-letter.docx was accepted without question.

Any user role that can upload a file (submission files, temporary files,
media files, etc.) could exploit this to plant an executable webshell on the
server, leading to full remote code execution — which is exactly the
technique used to deface OJS sites with gambling spam.

The fix

1. Replaced the substring check with a real denylist. Extensions are now
   compared exactly (case-insensitive) against FileManager::DANGEROUS_EXTENSIONS,
   a list covering PHP variants, CGI/Perl/Python/Ruby scripts, ASP/JSP
   variants, .htaccess/.htpasswd, native executables, and more.
2. Added real, content-based MIME type verification — the same approach
   Laravel's UploadedFile uses. parseFileExtension() now accepts an
   optional file path; when given, it uses PHP's finfo (via
   PKPString::mime_content_type()) to inspect the file's actual binary
   content, not the filename or the client-supplied Content-Type header.
   If the real content looks like a script or executable
   (FileManager::DANGEROUS_MIME_TYPES), the file is forced to a safe
   .txt extension no matter what name the uploader gave it.
3. Updated every upload code path (submission files, temporary files,
   media files, JATS imports, file versioning) to pass the real uploaded
   file path through, so the new content check is actually applied
   everywhere, not just extension parsing.
4. The change is backward compatible — the new parameter is optional, so
   nothing breaks if a call site isn't updated yet, it just still benefits
   from the stronger extension denylist.

Test plan

- [x] Added tests/classes/file/FileManagerTest.php:
  - Legitimate files (.docx, .pdf, .tar.gz, no extension) keep working
    as before.
  - Extensions that previously bypassed the check (.phtml, .pht,
    .phar, .phps, .cgi, .pl, .py, .sh, .jsp, .asp,
    .htaccess, .exe, etc.) are now rejected.
  - A PHP payload disguised with an innocent filename (e.g.
    vacation-photo.jpg) is caught by real content inspection.
  - Genuine files with matching content/extension still pass.
  - No file path provided falls back safely to the extension-only check.
- [x] Verified manually that the previously-bypassable extensions are
  blocked and normal document uploads are unaffected.
- [ ] Full suite to be run inside a complete OJS/OMP/OPS checkout with
  lib/pkp pinned to this branch.

Why this needs urgent attention

The weakness has sat in the codebase, publicly visible, marked with a
FIXME comment, for years — and it is already being actively exploited
against live journals. We'd ask that this be reviewed and merged with
priority, and considered for backport to all currently supported release
branches, given active exploitation is already occurring in the wild.

### DIFF
--- a/api/v1/submissions/MediaFilesController.php
+++ b/api/v1/submissions/MediaFilesController.php
@@ -255,7 +255,7 @@ class MediaFilesController extends PKPBaseController
                         throw new \Exception(__('api.404.resourceNotFound'));
                     }
 
-                    $extension = $fileManager->parseFileExtension($temporaryFile->getOriginalFileName());
+                    $extension = $fileManager->parseFileExtension($temporaryFile->getOriginalFileName(), $temporaryFile->getFilePath());
                     $fileId = app()->get('file')->add(
                         $temporaryFile->getFilePath(),
                         $submissionDir . '/' . uniqid() . '.' . $extension
--- a/api/v1/submissions/PKPSubmissionFileController.php
+++ b/api/v1/submissions/PKPSubmissionFileController.php
@@ -330,7 +330,7 @@ class PKPSubmissionFileController extends PKPBaseController
         }
 
         $fileManager = new FileManager();
-        $extension = $fileManager->parseFileExtension($_FILES['file']['name']);
+        $extension = $fileManager->parseFileExtension($_FILES['file']['name'], $_FILES['file']['tmp_name']);
 
         $submissionDir = Repo::submissionFile()
             ->getSubmissionDir(
@@ -479,7 +479,7 @@ class PKPSubmissionFileController extends PKPBaseController
             }
 
             $fileManager = new FileManager();
-            $extension = $fileManager->parseFileExtension($_FILES['file']['name']);
+            $extension = $fileManager->parseFileExtension($_FILES['file']['name'], $_FILES['file']['tmp_name']);
             $submissionDir = Repo::submissionFile()
                 ->getSubmissionDir(
                     $request->getContext()->getId(),
--- a/classes/file/FileManager.php
+++ b/classes/file/FileManager.php
@@ -52,6 +52,40 @@ class FileManager
     public const DOCUMENT_TYPE_URL = 'url';
 
     /**
+     * File extensions that common web server configurations (Apache/PHP,
+     * IIS/ASP, CGI handlers, etc.) may treat as executable/interpretable
+     * content rather than inert data. Files claiming these extensions are
+     * never allowed to keep them, regardless of client-supplied filename.
+     */
+    public const DANGEROUS_EXTENSIONS = [
+        'php', 'php1', 'php2', 'php3', 'php4', 'php5', 'php6', 'php7', 'php8',
+        'phtml', 'phtm', 'pht', 'phar', 'phps', 'pgif', 'inc',
+        'cgi', 'fcgi', 'pl', 'py', 'pyc', 'pyo', 'rb',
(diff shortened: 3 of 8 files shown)


asmecher:
@BayuFirmansyaah, the cause of this problem is not a flaw in the code -- it is an unsafe deployment whereby the "files directory" is placed within the web root where uploaded files can be invoked or downloaded by guessing the server-side path and filename.

For a safe deployment, **the files directory must be placed outside the web root, or protected from direct access via `.htaccess` or similar**. This is noted in the installation documentation, on the installation form, in the configuration file, and elsewhere.

This proposal does not fix the underlying problem, which is that submission files can be accessed or invoked directly. Even if it was possible to know and block any server-side executable -- and it's not, as these depend on the web server configuration and may vary -- it would still be possible to download author submission files without going through any of OJS's access controls. And it would still leave the server a target for spam and abuse, as third-party files can be uploaded and hosted without the journal's intention.

BayuFirmansyaah:
@asmecher I understand that the recommended deployment is to place the files directory outside the web root or restrict direct access to it.

However, I believe this should also be addressed at the application/code level as a security defense-in-depth measure. Relying solely on server configuration such as `.htaccess` or deployment practices does not fully mitigate the underlying vulnerability.

The application should ensure that uploaded submission files cannot be directly accessed, executed, or downloaded without passing through OJS's authorization and access-control mechanisms. At minimum, uploaded files should be stored and served in a way that prevents direct invocation, regardless of the web server configuration.

So, in my opinion, the proper fix should include both:

1. A secure deployment recommendation (files directory outside the web root or properly protected).
2. An application-level security fix to prevent unauthorized direct access, execution, or download of uploaded files.

This would provide defense in depth and ensure that the security of submission files does not depend entirely on the server's configuration.

asmecher:
> An application-level security fix to prevent unauthorized direct access, execution, or download of uploaded files.

This is not possible when the files directory is inside the web root. When it's inside the web root, access to the contents of that directory happens directly via the web browser, and not through OJS. OJS can't prevent it.

Issue #13218, opened 2026-08-24
Editiorial Masthead/History pages show no users — getSortedMastheadUserGroups() keyed by index, not user group ID


bozana:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

The Editorial Masthead and Editorial History pages (`AboutContextHandler::editorialMasthead()` / `editorialHistory()`) do not display users.

AC:

1. Editorial Masthead page (about/editorialMasthead) displays users under their respective masthead-visible role, for any journal with masthead-opted-in active users.
2. Editorial History page (about/editorialHistory) displays past-service users under their respective role, same as above but for ended assignments.
3. Roles with a saved order in journal setting display in that saved order.
4. New/unordered masthead roles (not yet present in the saved order) sort after all ordered roles — secondarily ordered by role ID among themselves.

`AboutContextHandler::getSortedMastheadUserGroups()` returns a `Collection` keyed by the default sequential index (`0, 1, 2, ...`):

editorialMasthead()/editorialHistory() assume it's keyed by user group ID:

```
foreach ($mastheadRoles as $userGroupId => $mastheadUserGroup) {
    foreach ($allUsersIdsGroupedByUserGroupId[$userGroupId] ?? [] as $userId) {
```

`$userGroupId` here is actually the sequential index (0, 1, 2...), not a real user group ID

Also: new masthead roles sort to the front, not the end.

### Steps to Reproduce

1. Have a journal with at least one masthead-visible user group with active/masthead-opted-in users.
2. Visit the journal's Editorial Masthead page (about/editorialMasthead).

### Expected Result

Users assigned to masthead-visible roles are listed under their respective role.

### Actual Result

No users are shown under any role — the page renders with empty role sections.

### Application Version

OJS, OMP, OPS 3.5 and main

### Additional Information

Introduced by: #10506 ("Refactor UserGroup to use Eloquent", commit 714d5d5aa49
Also related and introduced by the same refactor to Eloquent commit: new masthead roles sort to the front, not the end

<hr>

PRs:
stable-3_5_0:
- https://github.com/pkp/pkp-lib/pull/13219
- https://github.com/pkp/ojs/pull/5761 (only submodule update)
- https://github.com/pkp/omp/pull/2438 (only submodule update)
- https://github.com/pkp/ops/pull/1379 (only submodule update)

main:
- https://github.com/pkp/pkp-lib/pull/13220

bozana:
[...]

PR #13219, opened 2026-08-24
pkp/pkp-lib#13218 fix getSortedMastheadUserGroups() to be keyed by us…

bozana:
…er group ID, fix new masthead role position

see https://github.com/pkp/pkp-lib/issues/13218

### DIFF
--- a/pages/about/AboutContextHandler.php
+++ b/pages/about/AboutContextHandler.php
@@ -21,8 +21,6 @@ use APP\facades\Repo;
 use APP\handler\Handler;
 use APP\template\TemplateManager;
 use DateTime;
-use Illuminate\Support\Collection;
-use PKP\context\Context;
 use PKP\facades\Locale;
 use PKP\orcid\OrcidManager;
 use PKP\plugins\Hook;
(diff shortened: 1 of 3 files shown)


pages/about/AboutContextHandler.php
touhidurabir: unlike the implementation of `PKPAppearanceMastheadForm::__construct`, this one does not have a `->orderByRoleId()` . This can/will cause unordered roles deterministically go last, their order among themselves is unspecified DB order on the public page but role-id order in the settings preview, which do seems like a bit different.

touhidurabir: it seems this method and what is being done in the `PKPAppearanceMastheadForm::__construct` is almost identical so do you think it's better to move it to `PKP\userGroup\Repository` as 
```php
public function getSortedMastheadUserGroups(Context $context): Collection                                               
{
    $savedOrder = (array) $context->getData('mastheadUserGroupIds');

    return UserGroup::withContextIds([$context->getId()])
        ->masthead(true)
        ->excludeRoleIds([Role::ROLE_ID_REVIEWER])
        ->orderByRoleId()
        ->get()
        ->sortBy(function (UserGroup $userGroup) use ($savedOrder) {
            $position = array_search($userGroup->id, $savedOrder);
            return $position === false ? PHP_INT_MAX : $position;
        })
        ->keyBy('id');
}
```

then we can have one single source of same implementation.

touhidurabir (approved):
@bozana just 2 comments but not that impacting, up to you if you want to update . Approving .

pages/about/AboutContextHandler.php
bozana: I will move it. Good suggestion!

bozana: I moved it all into one function...

PR #13239, opened 2026-08-27
Support openid hooks

defstat:


asmecher (approved):
Looks good, @defstat. These are deprecated areas of code, but we'll have them for a while, so no objections to accommodating what we need to.

PR #13241, opened 2026-08-28
pkp/pkp-lib#13003 normalise funder name retrieve without static cache

touhidurabir:
pkp/pkp-lib#13003 funder name normalisation by dropping static cache


asmecher:
Thanks, @touhidurabir! However, this will only scale if there's something to replace the fetch of the context and submission, which only exist to get the list of possible locales. Otherwise we add at least 4 queries per funder plus all the overhead of rehydrating the entities, as each is repeatedly fetched on the article view page. I think it should be possible to rewrite `resolvedPublicationLanguages` to use a single raw DB query; even if this is repeated once per funder, that's acceptable IMO.

touhidurabir:
@asmecher this is not done yet but WIP . That is why I haven't yet tagged you . Hopefully will get it ready for review by Sunday.

PR #13244, opened 2026-08-28
pkp/pkp-lib#13003 Fix attempted serialization of closures on authors

asmecher:


asmecher:
Fixed by https://github.com/pkp/pkp-lib/pull/13257

Issue #13242, opened 2026-08-28
[Main Feature] | Theme - Add a new theme using Blade templates


NateWr:
### Description

This is a placeholder issue for the new Blade theme. I will update with more details once I have prepared the PRs.

### Steps to Test

1. Login as Admin/Editor.
2. Navigate to Website Settings > Plugins.
3. Enable the theme "Eidos".
4. Reload the page, navigate to Website Settings > Appearance.
5. Select the Eidos theme as the active theme and save.

NateWr:
The following PRs add the theme to OJS:

- https://github.com/pkp/ui-library/pull/972 (ui-library)
- https://github.com/pkp/pkp-lib/pull/13260 (pkp-lib)
- https://github.com/pkp/ojs/pull/5784 (ojs)
- https://github.com/pkp/crossref-ojs/pull/107 (crossref-ojs)

NateWr:
The PRs above implement all the features that I believe are necessary for a merge. However, there are many issues that can be addressed over time. In this comment, I will provide details on everything that I have tracked from my own testing and feedback from our Eidos working group.

### Code Refactors
There are still some architectural decisions to be made and other areas where code cleanup would be useful in the long-run.

- Decide on and implement improved pagination based on [this discussion](https://github.com/pkp/pkp-lib/discussions/12991#discussioncomment-17879818).
- I had to do some [special handling](https://github.com/pkp/pkp-lib/issues/13151) to pass the JSON to Vue components. Touhidur couldn't reproduce, but I haven't been able to investigate further.
- I have removed all class-based Blade components and moved these helper functions into the Blade templates. However, we may want to migrate the more complex code into helper functions based on [this discussion](https://github.com/pkp/pkp-lib/discussions/12991#discussioncomment-18080241).
- Metadata and Homepage Blocks -- there are two issues with the load order. The main problem arises because we need to "register" blocks early but "load" them late. Loading a block will run its callback function where it does data fetching so we want to do that only when absolutely necessary.
  - Plugins can not register their own blocks: see [details](https://github.com/NateWr/pkp-blade-theme-proposals/blob/main/theme-plugin-init-order.md).
  - Themes can not use the full list of blocks in the theme options. This happens because a theme only registers its blocks when it is active. So when a theme is first activated, its own  blocks are not part of the theme options. To reproduce:
    - 1. Enable the default theme and select it as the active theme.
    - 2. Reload the page.
    - 3. Go to Website > Settings > Appearance.
[...]

PR #13260, opened 2026-09-01
pkp/pkp-lib#13242 Add support for Blade theme

NateWr:


api/v1/comments/formRequests/AddComment.php
```diff
@@ -30,10 +30,10 @@ class AddComment extends FormRequest
     public function rules(): array
     {
         return [
-            'publicationId' => [
+            'submissionId' => [
```
NateWr, line 33: Comments are now loaded and added by `submissionId`. Commenting on old versions is not allowed, so a new comment is always attached to the current version.

api/v1/reviews/resources/ReviewRoundAuthorResponseResource.php
```diff
@@ -36,7 +36,7 @@ public function toArray(Request $request): array
         return [
             'id' => $response->id,
             'reviewRoundId' => $response->reviewRoundId,
-            'response' => $response->authorResponse,
+            'response' => $response->getLocalizedData('authorResponse'),
```
NateWr, line 39: We don't have a good way to get localized data at the Blade template level, yet. Maybe something like `ViewHelper::getLocalizedData($data)`?

I tried using the `LocalizedData` trait, but this doesn't work because `ViewHelper` is called statically. Need some way to use the trait's method statically.

classes/components/UserCommentComponent.php
```diff
@@ -138,7 +150,7 @@ public function getLoginUrl()
                 'login',
                 null,
                 null,
-                ['source' => $this->request->getRequestPath() . '#public-comments']
+                ['source' => $this->request->getRequestPath() . '?section=comments']
```
NateWr, line 153: I'm using a `?` instead of a hash `#'` because I need to save the `#` for jumping to content on the page (eg - link to a single comment or jump to author details).

Issue #13267, opened 2026-09-02
[Reviewer] | "View All Submission Details" shows "Array, Array" for Keywords, Subjects and Discipline, Supporting Agencies don't render


blesildaramirez:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

When a reviewer opens "View All Submission Details" on review step 1, the Keywords row shows Array, Array instead of the keywords. Same for Subjects and Disciplines. Agencies never shows up at all, although it was included in the controller to map its value but it's using a wrong identifier to get its data.

### Steps to Reproduce

1. Add keywords, subjects, disciplines and supporting agencies to a submission.
2. Assign a reviewer.
3. Log in as the reviewer, open the review, click "View All Submission Details" on step 1.
4. Keywords, Subjects and Disciplines show Array, Array. Agencies is missing.

### Expected Result

The correct values should render properly on the table. 

### Actual Result

The values for each metadata (controlled vocab: Keywords, Subjects, Disciplines) are showing Array, array.

### Application Version

OJS, OMP v3.5

### Additional Information

<img width="1897" height="720" alt="Image" src="https://github.com/user-attachments/assets/c163366c-19f0-44d0-b7e1-445f19237a06" />

---
### Development Updates
Pull requests:

- pkp-lib: https://github.com/pkp/pkp-lib/pull/13268
- ojs: https://github.com/pkp/ojs/pull/5788
- omp: https://github.com/pkp/omp/pull/2448
- ops: https://github.com/pkp/ops/pull/1388

blesildaramirez:
Hi @touhidurabir , @kaitlinnewson , another code review request from me - could you take a look when you get the chance? Appreciate your time on this, thanks!

touhidurabir:
@blesildaramirez all approved .

kaitlinnewson:
@blesildaramirez all approved now!

blesildaramirez:
Thanks, @touhidurabir @kaitlinnewson !
This is now merged to 3.5.

Proof of tests:
OJS:
- Double Anonymous
  <img width="1793" height="540" alt="Image" src="https://github.com/user-attachments/assets/b48da9f5-1040-48ca-bd9a-20e9df0b86eb" />
 - Disclosed Author / Open
   <img width="1802" height="627" alt="Image" src="https://github.com/user-attachments/assets/6531e770-736a-4b38-af0a-f594e5faba98" />

OMP:
- External Review
  - Double Anonymous
[...]

PR #13268, opened 2026-09-02
pkp/pkp-lib#13267 [Reviewer] | "View All Submission Details" shows "Array, Array" for keywords, subjects and discipline

blesildaramirez:


touhidurabir (approved):

controllers/modals/submission/ViewSubmissionMetadataHandler.php
```diff
@@ -67,18 +67,23 @@ public function display($args, $request)
 
         $templateMgr->assign('publication', $publication);
 
+        $controlledVocabs = [
+            'keywords' => 'common.keywords',
+            'subjects' => 'common.subjects',
+            'disciplines' => 'common.discipline',
+            'supportingAgencies' => 'submission.agencies',
```
kaitlinnewson: I think the supporting agencies metadata should actually be moved up into the `($reviewAssignment->getReviewMethod() != ReviewAssignment::SUBMISSION_REVIEW_METHOD_DOUBLEANONYMOUS)` block above - I know it's not frequently used, but it could contain identifying information about the author.

kaitlinnewson: Also the locale key could use `submission.supportingAgencies` instead (for `en` that is Supporting Agencies instead of Agencies).

blesildaramirez: Hi @kaitlinnewson , thank you for raising this. I updated this to check for double anonymous review methods, and used the suggested locale key for Supporting Agencies. Could you please check this again? Thanks!

kaitlinnewson (approved):

Issue #13109, opened 2026-07-29
[Metadata Editing] | Author should be able to edit metadata if the publication is not published


jardakotesovec:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description Issue 1
This is follow up ticket for the https://github.com/pkp/pkp-lib/issues/10263, which introduced following logic 

> If any publication of the submission is published, we don't allow authors to make any changes to any publication of that submission, even if there are other unpublished publications.

This is problematic for any follow-up versions, especially in OPS where its much more common for authors to be involved in editing and also adding galleys. 

Therefore this ticket tracks the adjustment to make it based just on one publication status, not considering other publications.

Note that Authors can edit publications metadata if they have explicitly that permission in participants settings.

### Description Issue 2

**Update (29-07-2026):** https://github.com/pkp/pkp-lib/issues/10263 also introduced behaviour, where when the publication is published it changes `canEditMetadata` flag to false for authors.  That also should be adjusted. There is no reason to change authors settings. They should be able to edit any unpublished publication if they have that permission.
 

### Steps to Reproduce

1. Login with User/Role Editor
2. Navigate to Workflow page for submission which has already one publication published
3. Create new publication and make sure author has 'Can Edit publication Metadata' permission in the participants
4. Login as author and try to edit metadata
5. Its not allowed

### Expected Result

Editing metadata should be allowed.

### Actual Result

Editing metadata is not allowed

### Application Version

OJS,OMP, OPS 3.5

Vitaliy-1:
I see that the new workflow stage was added recently, `WORKFLOW_STAGE_ID_DONE`, and this breaks the access of authors (and maybe other roles), to the metadata of the `published`, in our authorization policies. E.g., in the `[StageRolePolicy](https://github.com/pkp/pkp-lib/blob/main/classes/security/authorization/StageRolePolicy.php#L69)` where that new stage isn't taken into account. Trying to go through it...

Vitaliy-1:
PRs
ui-library: https://github.com/pkp/ui-library/pull/976
pkp-lib: https://github.com/pkp/pkp-lib/pull/13273
OJS: https://github.com/pkp/ojs/pull/5793
OMP: https://github.com/pkp/omp/pull/2453
OPS: https://github.com/pkp/ops/pull/1392

Vitaliy-1:
@ewhanson, can you take a look the these PRs? They interfere with your recent changes.
[...]

PR #13273, opened 2026-09-02
pkp/pkp-lib#13109 Ability for authors to edit metadata

Vitaliy-1:


classes/migration/upgrade/v3_6_0/I13109_PermitPublishedMetadataEdit.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
```
ewhanson, line 2: Missing @file doc block.

classes/migration/upgrade/v3_6_0/I13109_PermitPublishedMetadataEdit.php
```diff
@@ -0,0 +1,82 @@
...
+use PKP\security\Role;
+
+class I13109_PermitPublishedMetadataEdit extends Migration
+{
+    /**
+     * @inheritDoc
+     */
+    public function up(): void
+    {
+        $contextDao = Application::getContextDAO();
```
ewhanson, line 17: There's nothing wrong with this approach, but thought I'd share that I had also added a helper trait to get context table name and column name in [HasContextNameHelper](https://github.com/pkp/pkp-lib/blob/main/classes/migration/HasContextNameHelper.php). It may be a good idea to decide on one house-style approach for getting these and make sure they're used across the application. I'm guessing your approach is probably the simpler of the two since everything we need for it is already present, but it does involve instantiating the `ContextDAO`, so I'm not sure if that's ever problematic within a migration.

classes/submission/Repository.php
```diff
@@ -545,32 +545,24 @@ public function canCurrentUserDelete(Submission $submission): bool
     /**
      * Check if a user can edit the publication metadata of a submission
      */
-    public function canEditPublication(int $submissionId, int $userId): bool
+    public function canEditPublication(Publication $publication, User $user): bool
     {
         // block authors can never edit a published publication even if an editor granted them canChangeMetadata
```
ewhanson, line 550: I know this isn't from your changes, but I find this comment a little confusing, and I'm not sure what it's saying. Is it saying a "block author" is a kind of author that can never edit a published publication, or saying we should "block authors" so that they can never edit a published publication?

classes/submission/maps/Schema.php
```diff
@@ -596,6 +596,8 @@ protected function mapByProperties(
 
     /**
      * Determine whether current user is able to change metadata
+     *
+     * @deprecated should be determined per publication level, see pkp/pkp-lib#13109
```
ewhanson, line 600: Given this deprecation notice here, I'm not sure if we have an agreed upon approach for marking properties deprecated in our schema JSON files, e.g. adding a corresponding deprecation notice in `submission.json` as a counterpart to the addition of `canCurrentUserChangeMetadata` in `publication.json`, but it may make sense to mark it as deprecated there as well.

ewhanson (approved):
Thanks @Vitaliy-1. Just a few questions here, but overall looks good!

Issue #13286, opened 2026-09-08
[Submission File] | Cancelling a revision upload silently fails to restore the previous file


touhidurabir:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

initially reported at https://github.com/pkp/pkp-lib/issues/12347#issuecomment-5570287256 

In the file upload wizard, pressing `Cancel` after uploading a revision is supposed to roll the submission file back to the previous revision. It instead answers `{"status": false}` and leaves the new revision in place, with no error shown to the user.

The failure condition as the submission file's `uploaderUserId` is not the user who wrote the most recent event-log entry describing the file. The most common way to reach that is a rename by a different user, but an imported file reaches it with no rename at all.

This was made visible by pkp/pkp-lib#12347, but it is `not caused` by it as the restore has always relied on an non-deterministic assumption, and `#12347` removed the accident that was covering for it. As the `#12347` fix the event log off by one bug, this has become more easily visible

### Steps to Reproduce

There are several variants to re-produce this issue in different branch as 

Variant 1 --> fails only on main (post #12347) as
1. As editor `A`, upload article.pdf to Submission Files.
2. As editor `B`, open `Update File Details` on that file and rename it to `Renamed by B.pdf`. (Renaming does not change uploader user id, it stays `A`.) 
3. As editor `A`, open the `Upload` wizard, select that file as the file being revised, and upload article-rev.pdf.
4. Press `Cancel` on `step 1`

Variant 2 — fails on `3.4`, `3.5` and `main`.  same as above, but perform step `3–4` as editor `B` instead of `A`.

### Expected Result

The submission file is restored to the previous revision's fileId, name and uploader user id, and the cancelled upload's file is deleted.

### Actual Result

POST `.../manage-file-api/cancel-file-upload` returns `{"status": false}`. The submission file keeps the new fileId and the new name.

As a result, the file is show as renamed to canceled file . Also 2 additional logs can be observed from `More Information` of that file in submission file table where second one points to canceled of as uploaded and first one as the original file showing as upload, confusing and misleading 

### Application Version

OJS/OMP/OPS main

### Additional Information

As the pkp/pkp-lib#12347 only applied to main, this will be only fixed in main branch . 

#### main
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/13288
[...]

PR #13288, opened 2026-09-08
pkp/pkp-lib#13286 Handle edge cases of canceled revision upload

touhidurabir:
for pkp/pkp-lib#13286 .

### DIFF
--- a/classes/submissionFile/Repository.php
+++ b/classes/submissionFile/Repository.php
@@ -21,6 +21,7 @@ use APP\notification\NotificationManager;
 use APP\publication\Publication;
 use APP\submission\Submission;
 use Exception;
+use Illuminate\Database\Query\JoinClause;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Mail;
@@ -349,7 +350,8 @@ abstract class Repository
     /** @copydoc DAO::update() */
     public function edit(
         SubmissionFile $submissionFile,
-        array $params
+        array $params,
+        bool $log = true
     ): void {
         $newSubmissionFile = clone $submissionFile;
         $newSubmissionFile->setAllData(array_merge($newSubmissionFile->_data, $params));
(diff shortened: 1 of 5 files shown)


Vitaliy-1 (commented):
After trying to cancel revision upload on the `confirm` stage, it still doesn't remove the revision. Is it relevant to this particular issue?

Image

touhidurabir:
> After trying to cancel revision upload on the `confirm` stage, it still doesn't remove the revision. Is it relevant to this particular issue?
> 
> Image

@Vitaliy-1 good catch -- at the 3rd/final stage cancel, it's a the same existing issue become visible by fixing the `#12347` with a bit regression of this PR (both in combined).  In 3.5 this does not impact the file grid view as it stays same but do add the details of canceled one in the file log history 

3 5-file-grid
3 5-file-log-history

so in above 2 screenshots, first one shows the the revision cancel does not change the original file or path as shown in second screenshot, the canceled one do present in the log history which is wrong .

classes/submissionFile/Repository.php
```diff
@@ -399,6 +405,35 @@ public function edit(
...
+     * Delete the event log entries that describe one revision of a submission file.
+     *
+     * Used when a revision upload is cancelled before it was confirmed which(as revision) never
+     * became part of the file's history, and its entries would otherwise survive pointing at a
+     * file that the same operation deletes, which leaves a broken download link in the event log.
+     *
+     * Matches on the `submissionFileId` and `fileId` recorded by getSubmissionFileLogData(), so it
+     * covers both the submission file and the submission scoped entries.
+     */
+    public function deleteRevisionLogEntries(SubmissionFile $submissionFile, int $fileId): void
```
Vitaliy-1: Matter of preference but I would use subqueries, something like this
```sql
SELECT DISTINCT els.log_id FROM event_log_settings as els
WHERE els.log_id IN (SELECT els.log_id FROM event_log_settings as els WHERE els.setting_name = 'submissionFileId' AND els.setting_value = ?)
AND els.log_id IN (SELECT els.log_id FROM event_log_settings as els WHERE els.setting_name = 'fileId' AND els.setting_value = ?)
```
No need to change anything)

Vitaliy-1 (approved):
Looks good!

Issue #13221, opened 2026-08-24
Implement/integrate Crossref's Cited-By feature


bozana:
Crossref's Cited-By feature needs to be integration within OJS and OPS. See https://www.crossref.org/services/cited-by/ and especially https://www.crossref.org/documentation/cited-by/retrieve-citations/.

The following tasks are needed:
- [ ] In the Crossref main plugin there should be an option to enable/disable Cited-By.
- [ ] When Cited-By enabled, the citations will be retrieved for a DOI and display on the article landing page. 
- [ ] How to display the citations should be discussed with Devika i.e. design is needed.
- [ ] The citations for a DOI should be cached. We could eventually used start date to always add citations that came after the last requesting/caching. See startDate parameter. Also to avoid timeouts.
- [ ] The display should be adapted int he oreTheme, see https://github.com/pkp/dev-team/issues/316.

PR #13292, opened 2026-09-08
pkp/pkp-lib#13221 Implement crossref's citedBy feature

taslangraham:
For pkp/pkp-lib#13221 & https://github.com/pkp/dev-team/issues/316


bozana:
Hi @taslangraham, also here, I think this all could/should go into the plugin itself, as in the example https://github.com/jardakotesovec/backendUiExample.
I am not sure about that blade file i.e. how this could be solved... :thinking: 
Also pinging @jardakotesovec, to double check.

Issue #13299, opened 2026-09-09
[Invitations] Validate user group context


defstat:
The user groups added in a userRoleAssignment Invitation should be validated against the invitation's context.

### Application Version

OJS, OMP, OPS 3.5 and main

---

**PRs**

**Branch stable-3_5_0**
OJS: https://github.com/pkp/ojs/pull/5830 **TESTS ONLY**
pkp-lib: https://github.com/pkp/pkp-lib/pull/13300

**Branch main**
OJS: https://github.com/pkp/ojs/pull/5831 **TESTS ONLY**
pkp-lib: https://github.com/pkp/pkp-lib/pull/13340

-- Reopened

**Branch stable-3_5_0**
OJS: https://github.com/pkp/ojs/pull/5844
pkp-lib: https://github.com/pkp/pkp-lib/pull/13364

**Branch main**
OJS: https://github.com/pkp/ojs/pull/5845
pkp-lib: https://github.com/pkp/pkp-lib/pull/13365

defstat:
@asmecher could you review this PR?

asmecher:
@defstat, I sent in a submodule update for testing and it failed all tests (https://github.com/asmecher/ojs/actions/runs/35024953866/job/104569632473). Did this pass tests?

defstat:
@asmecher I rebased the PR on current `stable-3_5_0` branch and all tests are passing https://github.com/pkp/ojs/pull/5830. 

The review comments are now resolved. Also ported PRs to `main` branch.

asmecher:
Resolved, thanks!

PR #13300, opened 2026-09-09
[pkp-lib][stable-3_5_0] #13299 Validate user group context

defstat:


classes/invitation/invitations/userRoleAssignment/rules/UserGroupBelongsToContextRule.php
```diff
@@ -0,0 +1,48 @@
...
+    public function passes($attribute, $value)
+    {
+        $userGroup = UserGroup::find($value);
+
+        // Existence is checked by UserGroupExistsRule.
+        if (!$userGroup) {
+            return true;
+        }
+
+        return $userGroup->contextId === $this->invitation->getContextId();
```
asmecher: Don't use an exact test on this -- some database setups will return numeric types as strings.

api/v1/invitations/InvitationController.php
```diff
@@ -129,8 +129,6 @@ public function getGroupRoutes(): void
             self::roleAuthorizer([
                 Role::ROLE_ID_SITE_ADMIN,
                 Role::ROLE_ID_MANAGER,
-                Role::ROLE_ID_SUB_EDITOR,
```
asmecher, line 132: There's a risk here that someone later will re-add these and not realize the risk to the UserRoleAssignment toolset; add a comment here to avoid that.

defstat: I added this https://github.com/pkp/pkp-lib/issues/13339 issues to resolve that in a specific invitation context

classes/invitation/invitations/userRoleAssignment/rules/UserGroupBelongsToContextRule.php
defstat: Fixed

api/v1/invitations/InvitationController.php
asmecher: I added a warning here referencing that issue.

Issue #13310, opened 2026-09-10
Submission created by disabled user cannot be viewed


asmecher:
### Valid Title

- [x] I have updated the title to accurately reflect the bug description

### Description

A submission created by a user who was subsequently disabled cannot be viewed in the Review stage (by an editorial user). This results in an error:

<img width="1465" height="457" alt="Image" src="https://github.com/user-attachments/assets/22cc3d95-f916-497f-bc40-57870ea10d0c" />

This traces to
```
lib/pkp/api/v1/submissions/tasks/resources/EditorialTaskParticipantResource.php:108
```
...where a list of users has been loaded assuming that referential integrity holds and therefore the expected ID will exist in the collection, but the user collector by default does not allow disabled users to be returned.

### Steps to Reproduce

Reconstructing from circumstances on the ORE test VM:

1. Create a submission as an author
2. Walk that submission into the peer review process (with assigned reviews, possibly a mix of complete and incomplete)
3. As Editor, disable that user's account
4. As Editor, navigate to the submission list
5. As Editor, attempt to find/browse that submission

### Expected Result

The workflow should continue to present the scholarly record regardless of the fact that the user account is disabled

### Actual Result

The workflow will not allow the submission to be discovered in the lists or viewed in the workflow (see error above).

### Application Version

OJS pre-3.6

# PRs
- https://github.com/pkp/pkp-lib/pull/13311
- https://github.com/pkp/ojs/pull/5815

asmecher:
@Vitaliy-1, can you review these PRs?

Vitaliy-1:
Thanks, @asmecher, I've missed the logic for disabled users.

asmecher:
@Vitaliy-1, I've rebased and resolved the review comments. Can you have a second look?

Vitaliy-1:
Looks good!

PR #13311, opened 2026-09-10
pkp/pkp-lib#13310 Account for disabled users when fetching workflow details

asmecher:


Vitaliy-1 (approved):
There is another similar call: https://github.com/pkp/pkp-lib/blob/b878a76ed46fb2639bc23fd0b544e954535c98f2/api/v1/submissions/tasks/EditorialTaskController.php#L1303

Also, I see that the changes are already partially applied

Vitaliy-1 (approved):

Issue #13320, opened 2026-09-14
Remove the Web based Task Runner from main (pre 3.6)


touhidurabir:
**Description:**
Drop the web based task runner which is controlled via config.inc.php file's `[schedule].task_runner` . This mechanism allow to run schedule tasks at the end of web request life cycle without needing to set up cron to run. 

Decision from https://github.com/pkp/pkp-lib/issues/13041#issuecomment-5359505277 . 

As cron is almost native to any system unlike queue jobs worker, this will reduce the tech debt significantly along with better performance to run schedule tasks via cron. 

**What application are you using?**
  OJS, OMP or OPS version main (pre 3.6)

### PRs
[pkp-lib](https://github.com/pkp/pkp-lib) --> https://github.com/pkp/pkp-lib/pull/13323
[ojs](https://github.com/pkp/ojs) --> https://github.com/pkp/ojs/pull/5819
[omp](https://github.com/pkp/omp) --> https://github.com/pkp/omp/pull/2461
[ops](https://github.com/pkp/ops) --> https://github.com/pkp/ops/pull/1400

touhidurabir:
@asmecher ready to review the PRs at https://github.com/pkp/pkp-lib/issues/13320#issue-5447616793 .

asmecher:
I added a review [here](https://github.com/pkp/pkp-lib/pull/13323#pullrequestreview-5281564274), thanks!

PR #13323, opened 2026-09-14
pkp/pkp-lib#13320 Remove web based schedule task runner

touhidurabir:
for pkp/pkp-lib#13320 .


tools/scheduler.php
```diff
@@ -129,17 +131,97 @@ protected function run(): void
 
     /**
      * List all the schedule tasks in the system
+     *
+     * Rendered here rather than delegating to Laravel's ScheduleListCommand so that the next run
+     * can be shown as an absolute date. The relative form on its own is ambiguous: Carbon reports
+     * a single unit and floors it, so a monthly task 12 days out reads as "1 week from now".
      */
```
asmecher, line 138: @touhidurabir Is the problem here that the Laravel task list gives a vague answer when presenting it to the user?

touhidurabir: @asmecher there is actually no `vague answer` . It shows as how it should be as default defined by laravel . So current CLI command `php lib/pkp/tools/scheduler.php list` shows as 

CleanShot 2026-09-23 at 13 33 51@2x

see that it says `Next Due: 1 week from now` which is correct but it will be better if we can get the exact time so the change make it as 

CleanShot 2026-09-23 at 13 34 13@2x

which now says `Next Run: Oct 1, 2026 00:00 (in 7 days)` which is much better .

asmecher: OK, but wouldn't that be better done as an upstream proposal to Laravel? I don't think it's worth re-implementing upstream functionality here where we have to maintain it.

touhidurabir: not a bad idea but that would not help us much right now as we do not have it's CLI app bootstrapping or artisan . that is why we have our own brewed command like schedule or jobs . and this is already done in 3.5 and keeping to symmetry here .

PR #13324, opened 2026-09-14
Settings perf

jardakotesovec:


asmecher (commented):
Looks good, @jardakotesovec, but could you also add it to `config.TEMPLATE.inc.php` with a warning *not* to enable it in production? If I remember right, this can easily lead to negative interactions between applications on details like time zones, charset configurations, etc. in environments using fcgi or equivalent.

PR #13334, opened 2026-09-15
Filter users with filterByStatus when fetching tasks/discussions

blesildaramirez:


api/v1/submissions/tasks/EditorialTaskController.php
taslangraham: I haven't tested the flow that hits this piece of code, but this seems like another place that might need a similar code change to handle disabled users.

taslangraham (approved):
Approved @blesildaramirez, but I've also highlighted another area in the code that might be worth checking too.

api/v1/submissions/tasks/EditorialTaskController.php
blesildaramirez: Well spotted, thanks Taslan! I applied it to the following places:
- `fromTemplate`: for consistency - not really needed since the user collection comes from `getTaskData` which already includes disabled user accounts, so no harm in adding this filter.
- `recordParticipantsAction`: needed so the t&d log for the disabled participant/user will record their name.

Merging this.

Issue #13181, opened 2026-08-15
Invitation accept/decline URLs name the wrong journal when the invitation is sent from a queued job


DusanStijovic:
**Version:** OJS 3.5.0-4

### Description

`InvitationHandler::getActionUrl()` builds the accept and decline URLs from the journal
of the *current HTTP request*, although the invitation itself knows which journal it
belongs to:

```php
// lib/pkp/pages/invitation/InvitationHandler.php:117-124
$request = Application::get()->getRequest();
$contextPath = $request->getContext() ? $request->getContext()->getPath() : null;

return $request->getDispatcher()
    ->url(
        $request,
        Application::ROUTE_PAGE,
        $contextPath,
```

`invitations.context_id` is populated, and `PKP\jobs\email\ReviewReminder` constructs
`ReviewerAccessInvite` with the correct `$context->getId()`, but that never reaches the
URL.

The URL is built while the job is running. `ReviewerAccessInvite::updateMailableWithUrl()`
computes it eagerly on its first line and only hands the finished string to
`buildViewDataUsing()`:

```php
// lib/pkp/classes/invitation/invitations/reviewerAccess/ReviewerAccessInvite.php:91-100
public function updateMailableWithUrl(Mailable $mailable): void
{
    $url = $this->getActionURL(InvitationAction::ACCEPT);

    $mailable->buildViewDataUsing(function () use ($url) {
        return [
            ReviewAssignmentEmailVariable::REVIEW_ASSIGNMENT_URL => $url
        ];
    });
}
```

`Invitation::getActionURL()` (`lib/pkp/classes/invitation/core/Invitation.php:439-442`)
delegates straight to `InvitationHandler::getActionUrl()`, and
`updateMailableWithUrl()` is itself called from `ReviewReminder::handle()`
(`lib/pkp/jobs/email/ReviewReminder.php:83`) — that is, inside the queue job, where
"the current HTTP request" is whatever request happens to be draining the queue.

This produces a different wrong URL depending on who drains the queue. All three rows
below were measured on 3.5.0-4, on a reminder for a submission in the journal `alpha`
(supported locales `en`, `sr_Latn`; site supported locales `en`, `sr_Latn`):

| Who runs the job | `$request->getContext()` | Resulting URL |
| --- | --- | --- |
| worker daemon, `lib/pkp/tools/jobs.php run`/`work`, or `tools/runScheduledTasks.php` | none | `…/index.php/index/en/invitation/accept?…` |
[...]

PR #13341, opened 2026-09-16
[pkp-lib][main] #13181 Respect Invitation's context rather than request's

defstat:


pages/user/ProfileHandler.php
```diff
@@ -69,7 +69,7 @@ public function profile($args, $request)
             [$firstContext, $secondContext] = [$workingContexts->next(), $workingContexts->next()];
             
             if ($firstContext && !$secondContext) {
-                $request->redirect($firstContext->getPath(), 'user', 'profile', null, $args);
+                $request->redirect($firstContext->getPath(), 'user', 'profile', $args);
```
asmecher, line 72: @defstat, I don't think this was incorrect, despite the name of the `$args` param making it possibly appear so. The first parameter to an old-style handler function will be the path info; in this call, it is being bounced back into the `$path` parameter, which is appropriate. After this change, these will be transformed into URL parameters, which is not what's intended.

asmecher: Actually, I was reading the change backwards -- this commit *corrects* the problem I was talking about above. Thanks, this does resolve the issue!

